PR #211, opened 2011-12-16
Bug fix: Allow loading of metadata for Proxies

sandulungu:
Covers a few use cases, mainly when working with merged documents.
A similar problem (with possibly similar fix) has been reported in Doctrine ORM: http://www.doctrine-project.org/jira/browse/DDC-1441


malarzm:
This seems to be fixed in doctrine/common now so we don't have to care :)

PR #323, opened 2012-06-08
Ability to specify field name for one/many references.

Parad0X:
This patch should allow specifying custom field name for references.

``` yaml
Post\
    referenceMany:
        relatedPosts:
            name: related_posts
            targetDocument: Post
```

My text editor automatically removed couple trailing white spaces. I can put those back if that's important.


Parad0X:
Also I'm a little confused over this test xml file:

Doctrine.ODM.MongoDB.Tests.Mapping.User.dcm.xml

``` xml
...
<reference-many target-document="Phonenumber" field="phonenumbers">
...
```

What's field? Name of the database field? Name of the model property? If it's the name of the model property then what fieldName is supposed to do for `reference-many`?

Issue #304, opened 2012-05-07
no sharding support


tecbot:
the current version doesn't support sharding.
We need a new "Shard" key and add it to all update queries of a document.

See: http://www.mongodb.org/display/DOCS/Sharding+Limits
"All (non-multi)updates, upserts, and inserts must include the current shard key. This may cause issues for anyone using a mapping library since you don't have full control of updates."

tux-rampage:
+1

malarzm:
Sharding is now supported thanks to https://github.com/doctrine/mongodb-odm/pull/1385

PR #325, opened 2012-06-11
[WIP] Feature sharding

tecbot:
Hi,

i have implemented a first version of sharding support. I added a new annotation "QueryFields" to change the core query routine for a document. 

The next step is to update the reference mappings to include the query fields in the reference object. But first I want some feedback from you.

Issue: #304

Regards
Thomas


alcaeus:
Superseded by #1385.

PR #455, opened 2012-11-30
Added support to use dot Notation 

cwerner1:
I used this technique to create a summary object of an other object.

It enables you to use `@Field(name="prop.key")` and the value will be filled in.


jwage:
In order to merge this we need a test to understand what is being added/changed. Please reopen if you want to continue working on this.

PR #462, opened 2012-12-08
Added version mapping to XML and YAML drivers

ianunruh:
Pretty small change


jmikola:
It looks like YAML already supported this, but XML supported was added in #997, which is now available in 1.0.0-BETA12.

PR #481, opened 2013-01-29
Made discriminatorMap annotation unnecessary

DenisGorbachev:
Use case:

* XParentDocument in bundle X.
* XChildDocument extends XParentDocument in bundle X.
* YChildDocument extends XParentDocument in bundle Y.

If bundle Y is homemade and bundle X is third-party, a developer can't redefine discriminatorMap on XParentDocument, and thus can't include YChildDocument in it.

This PR solves the issue by storing FQCNs in discriminatorFields instead of arbitrary keys. Also, this PR remaps discriminatorField to an array of parent FQCNs (which, in absence of discriminatorMap, allows to find all parent + children class documents in one query).

| Q             | A
| ------------- | ---
| Bug fix?      | yes (?)
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT
| Doc PR        | none (will create if needed)

One test fails, but it does so even on current mongodb-odm master.


jwage:
At this time I don't think we want to accept this PR. I would rather keep the discriminator maps and keep things explicit. This is also a rather large BC break that I don't think I would want to introduce.

Issue #501, opened 2013-02-17
Odd Behavior with Embedded Documents, Persisting and Flushing without Clearing?


swquinn:
I've encountered what I can only describe as bizarre behavior when it comes to persisting a Document that contains an Embedded Document. I discovered this as part of some scaffolding code I have written to make it easier to populate a database with values for testing. Maybe this is due to a lack of good comprehension and understanding on my part or, possibly a bug in the Doctrine library.

_First_, I am trying to ensure consistency between certain objects by explicitly giving them an ID. I do this by setting the Document's ID to a `new MongoId()`. I didn't see anything in Doctrine--or Mongo's--documentation that indicated I could not explicitly assign an ID, so I assume that this is permitted.

_Second_, if you `persist()` and `flush()` both documents (embedded first and then the parent) without calling a `clear()` the data contained in the embedded document does not appear to be persisted, _but only when you've explicitly assigned an ID to the parent object_.

I have fixed my code not to call `persist()` and `flush()` unnecessarily on embedded documents but it still seems odd that the act of persisting/flushing an embedded document and then the owning document does not always commit the data to the database.

(Also, I'm still learning Git/Github so if there is anything I'm doing that is out-of-sorts with how issues are normally reported on or operated on, please let me know so I can  become a better Githubber, thanks!)

PR #505, opened 2013-02-19
Illustration of issue noted in DB for GH501

swquinn:
Pull request for #501, per @jmikola 's request.

I think that I'm doing this right; unfortunately I hadn't created the gh501 branch previously when working on the test case to illustrate the issue. I made all my changes against my fork's master branch. I'm hoping that creating the branch now is fine. Regardless the only change should be the addition of the GH501Test file. Please let me know if there is anything else I need to do!

Just to re-state what is happening, the test(s) actually pass (IIRC) however in one of the three test cases the data is not persisted as would be expected. This may be "works as designed" and expected behavior, but I was perplexed by it none-the-less.

Thanks again!


jwage:
Can we update the test so that it checks for the data that is incorrect and fails?

malarzm:
Closing since no update was provided for over a year

PR #532, opened 2013-03-25
[Common 2.4+] added "documentNotFound" event to hook in the proxy initialization process

tecbot:
If the document not exists so a user can hook into and create a fallback objekt to prevent a exception.


jwage:
Can we rebase the PR?

alcaeus:
Superseded by #1336.

Issue #605, opened 2013-06-05
PersistentCollections cannot be copied between databases


floriansemm:
Setup: document with references

1. fetch document from db1
2. persist document in db2
3. remove document from db1

references in document are not persist in db2

malarzm:
Please see https://github.com/doctrine/mongodb-odm/pull/606#issuecomment-164594075

PR #606, opened 2013-06-05
failing test for #605

floriansemm:


jwage:
@floriansemm I think if you clone the embedded documents before trying to copy them it should work properly. Can you test that?

floriansemm:
No sorry, I can't test it. I have currently no environment with mongoDB.

malarzm:
FWIW using `clone` works fine. I'd say it's the way to go here :)

PR #645, opened 2013-07-25
allow discriminator mapping with simple refs

sorabji52:
discriminator mapping gets ignored when using simple refs...this fixes that.


jwage:
Closing. Read my comment here for why https://github.com/doctrine/mongodb-odm/pull/1038#issuecomment-87233315

PR #692, opened 2013-10-17
Revert "PersistentCollection::count() with ReferenceMany InverseSide map...

spantaleev:
...ping queries DB without collection initialization"

This reverts commit bd7d99e290159e310469b0a14aee1aafc7df2381.

Conflicts:
    lib/Doctrine/ODM/MongoDB/PersistentCollection.php

Reverting, because not initializing counted incorrectly when:

    - the collection has the isInverseSide flag set
    - the collection is uninitialized
    - has had an item added using add(), making it dirty

Here's the previous (and broken before this commit) flow:

    - the collection is inverse-side and not initially initialized
    - we add an item via add()
        - has the side-effect of marking the collection as dirty
        - but still keeps the collection uninitialized
    - if count() is called now for this uninitialized inverse-side collection:
        - a query gets sent to the database to count the objects
        - the result of counting is 0, because:
            - there's nothing in the DB yet
            (assuming there really was nothing old there for starters)
            - nobody cared that $this->coll had some items added to it
            and the collection was dirty

The easiest fix I can think of is to restore the old behavior, which
causes initialize() to happen just in case.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/PersistentCollection.php
+++ b/lib/Doctrine/ODM/MongoDB/PersistentCollection.php
@@ -483,16 +483,10 @@ class PersistentCollection implements BaseCollection
      */
     public function count()
     {
-        if ($this->mapping['isInverseSide'] && ! $this->initialized) {
-            $documentPersister = $this->uow->getDocumentPersister(get_class($this->owner));
-            $count = empty($this->mapping['repositoryMethod'])
-                ? $documentPersister->createReferenceManyInverseSideQuery($this)->count()
-                : $documentPersister->createReferenceManyWithRepositoryMethodCursor($this)->count();
-        } else {
-            $count = $this->coll->count();
+        if ($this->mapping['isInverseSide']) {
+            $this->initialize();
         }
-
-        return count($this->mongoData) + $count;
+        return count($this->mongoData) + $this->coll->count();
     }
 
     /**


jwage:
:+1: for reverting the behavior back to how it was.

alcaeus:
I just ran into a different issue related to this. Previously, the entire concept of inverse references could be used to count objects. For example, if we wanted to know how many blog posts a user has written without manually keeping track of them, we could use an inversed reference to count them because calling ```count()``` on an uninitialized ```PersistentCollection``` would actually call ```count()``` on the underlying cursor and skip hydration of all posts.
With this change merged, this no longer works and even when I'm only interested in the number of items in the collection, the entire collection is indexed, potentially consuming a bunch of time as well as memory.

The problem is that inversed collections are not tracked for changes. This means that adding an item to an inversed collection will do absolutely nothing - the item has to be persisted separately. There are still occurences where the count will still yield the wrong result:
```
public function testAddDocumentsToInverseCollection()
{
    $user = new Documents\User();
    $this->dm->persist($user);
    $this->dm->flush();
    $this->dm->clear();

    // Load document from database to get a user object with a PersistentCollection instance
    $user = $this->dm->find('Documents\User', $user->getId());
    $blogPost = new Documents\BlogPost();
    $blogPost->setUser($user);
    $user->getPosts()->add($blogPost);

    $this->dm->persist($blogPost);
    $this->dm->flush();

    $this->assertSame(1, $user->getPosts()->count());
}
```
This test would fail because ```count()``` has not been called before persisting the object, so ```initialize()``` would load the persisted blog post from the database then add the one manually added to it, resulting in two blog posts which is wrong. I think the problem is not the way we count these documents, but the way inversed collections are treated with respect to manual changes. There is no change tracking done at all, so there's no way to get correct results consistently.

Issue #687, opened 2013-10-11
[mapping type] What about to add ObjectType like ORM has?


makasim:
Something like [DBAL ObjectType](https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Types/ObjectType.php). I have to store simple serializable object to mongo. For now I created a custom ObjectType (almost same as DBAL one) in the lib. It works fine. 

Is there any change to add it to mongo-odm as standard one? I could work on it.

alcaeus:
As per previous discussion, use embedded documents. If you don't want to, serialise data and use a binary type. If you want serialisation to be taken care of for you, extend one of the binary types and update the corresponding methods. Closing here.

PR #696, opened 2013-10-23
[wip] added serialized type

ruslan-polutsygan:
It fixes issue https://github.com/doctrine/mongodb-odm/issues/687

If it looks OK for you, I'll add tests.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Types/SerializedType.php
+++ b/lib/Doctrine/ODM/MongoDB/Types/SerializedType.php
@@ -30,13 +30,19 @@ class SerializedType extends Type
         return $val;
     }
 
+    /**
+     * {@inheritDoc}
+     */
     public function closureToMongo()
     {
         return '$return = \MongoBinData(serialize($value), \MongoBinData::BYTE_ARRAY);';
     }
 
+    /**
+     * {@inheritDoc}
+     */
     public function closureToPHP()
     {
         return '$return = unserialize($value->bin);';
     }
-} 
+}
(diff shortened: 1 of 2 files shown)


jwage:
I don't think we should be storing serialized PHP data in mongodb. It should be able to store any arbitrary array structure so storing it natively should not be a problem.

makasim:
There is still a problem, I have an object which is stored in a document's property and I do not want to use embedable to store it, It is just overkill in my situation. I just want  it to be serialized and stored. That's it. Without this the library limits me which is bad. It pushes me to create one more document define its schema use it as embedable. I dont need all that stuff. IMO the library should provide different ways to solve problems and I have to choose a better solution for my case. 

And the other point: There is an inconsistency between Doctrine ORM and ODM. ORM provide such a feature but ODM does not.

makasim:
To be more concret:

Here's the document: https://github.com/Payum/Payum/blob/master/src/Payum/Core/Model/Token.php#L13

and this object https://github.com/Payum/Payum/blob/master/src/Payum/Core/Model/Identity.php I want to serialize in details property. 

Here's the currenct schema I use https://github.com/Payum/Payum/blob/master/src/Payum/Core/Bridge/Doctrine/Resources/mapping/Token.mongodb.xml#L11 

and to make it work I ask users to register a custom type ObjectType.

makasim:
@jwage ping

smolinari:
I am admittedly quite new to all this, but I am also not getting the reasoning for this. What overhead/ overkill are you talking about, when working with an embedded document (your object?)? How is that a limit? Instead of working with "details" as a special field type, you could simply embed your Identity object directly, couldn't you? 

`/** @Document */
class Token implements TokenInterface
{
    // ...

    /** @EmbedOne(targetDocument="Identity") */
    private $details;

    // ...
}

/** @EmbeddedDocument */
class Identity extends IdentityInterface
{
    // ...
}`

Isn't this possible?

Scott

PR #701, opened 2013-11-09
DocumentGenerator configuration awareness for aliases. Unnecessary FQN watch.

ephrin:
Does not work aliases behavior. Will be updated.


jwage:
@lazycommit Any update here?

ephrin:
@jwage nope. will not

PR #782, opened 2014-01-25
Remove embedded documents complexity

### LINKED ISSUE #1058: Document is upserted and then updated during one flush operation
* I have document with ```pre-generated id``` so it added to ```documentUpserts``` array after calling ```$dm->persist($document)```

* I have ```onFlush``` listener where I add some data to that document and then call ```$dm->getUnitOfWork()->recomputeSingleDocumentChangeSet($metadata, $document);```

* I call ```$dm->flush()``` and see the next picture:
```documentUpserts``` and ```documentUpdates``` has the same document to update.

It looks like the problem is [here](https://github.com/doctrine/mongodb-odm/blob/master/lib/Doctrine/ODM/MongoDB/UnitOfWork.php#L814),  in direct adding of the document to the ```documentUpdates``` array but should be through ```scheduleForUpdate``` method?

jwage:
- Remove some of the additional complexity when working with embedded documents. Treat embedded documents like normal documents in the UnitOfWork as much as possible.
- Fix issue with ClassMetadata::associationMapping not inheriting associations correctly.
- Fix issue with re-persisting removed embedded document. #719 
- Fixes #1058, #999, #1059, #1017


AronSzigetvari:
I wanted to test this with our project but got an error:
Fatal error: Class 'Doctrine\ODM\MongoDB\MongoDBInvalidArgumentException' not found in lib\Doctrine\ODM\MongoDB\UnitOfWork.php on line 942

This MongoDBInvalidArgumentException doesn't exist at all in the project. Maybe you have forgotten to git-add it?

jwage:
@AronSzigetvari I apologize. I have a little more work to do here still. I realized it after I asked people to test and didn't update the comments here.

jwage:
@AronSzigetvari I pushed up a change in c62b532a886e6a76d5668a5af19a8590fa635ba7 that fixes this. This PR originally had some other changes in it that I removed and I missed a few references.

alcaeus:
I tried this today - it works fine except for a tiny issue.
I'm using DoctrineExtensions to add translatable behavior to documents. In this case, it's an embedded document which makes things a bit complicated.
In order to switch locales, the locale field of the embedded document is changed, after which the document is refreshed. This no longer works after the refactoring, since doRefresh() will not work with embedded documents. However, refreshing the parent document will reset the (unmanaged) locale field in the embedded document, which causes an error while loading.
Now, this could be a case of "you're using it wrong", I'd have to take another look at that. Just thought I'd mention this since I promised to run it against our application.

jwage:
@alcaeus Do you think you can make a unit test that demonstrates the problem? I can see if I can make it work like it did before in that case.

alcaeus:
Will do, give me a couple of days to investigate. As I said, this could be a case of "you're using it wrong".

alcaeus:
@jwage I created the above issue to track the underlying issue with DoctrineExtensions, or at least what I think the underlying issue is. From what I can see, the only way to currently handle those behaviors in embedded documents is to refresh the embedded document which is no longer supported with this branch. Given the nature of embedded documents, I agree that they should not be refreshed directly.

blockjon:
I am a Doctrine ODM user and I have run up against a problem that sounds related to this PR. My PR https://github.com/doctrine/mongodb-odm/pull/1094 demonstrates a data consistency problem which I thought would be fixed by this branch but I just rolled a quick test and I see we still have this same problem even on your PR. How do you suggest I proceed with the resolution?

jwage:
@blockjon we should try to fix your case in this branch.

jwage:
@blockjon I take that back. This looks like it be solved by an entirely new feature. Possibly in #1066

malarzm:
@jwage any chance for rebase? Now that I have upgraded my app to BETA13 I will be able to run our test suite against this branch

malarzm:
Ok, so the branch is rebased and all my functional tests are still passing using this branch :) 

@alcaeus @blockjon @AronSzigetvari (and everyone who step by here) may as you gentelmen to test this branch?

blockjon:
@malarzm Im checking this against my repo right now...

blockjon:
My project's unit test all seem to pass with this branch.

On a slightly different topic, you are all at this point aware of my interest in pushing forward with the atomicSet feature that @malarzm has been helping to build.

As I have been thinking about Doctrine ODM, I have come to the opinion that all the whole project should use atomicSet for EmbedMany everywhere. If it were up to me, I would remove the option for other "strategies" pretty much entirely. In my view, all new coders of this codebase (or any ODM/ORM for that matter) would reasonably expect a data persistence layer to use atomic queries such that no data loss & corruption would even be possible. My app isn't even "highly concurrent" per se. At any one moment, I have at *most* 2 users trying to update a record and I have run into integrity problems.

With the way the system supports different strategies, it means that some strategies are more stable than others. For example, the atomicSet strategy currently does not support editing of documents (https://github.com/doctrine/mongodb-odm/pull/1153).

I realize that there are advantages to having smaller & more compact queries but I think data integrity and consistency supersedes all else.

malarzm:
@blockjon thanks for the test :) As for #1153 I will come back to that my tomorrow and (hopefully) it'll be the last issue, I'll try to find more embedded documents tests, double them for being atomic and see if they'll be still working. Hopefully atomic strategires will be all fixed and hardened once we reach 1.0.0, which I hope is around the corner, we can even discuss making it default one by then - soon we shall see :)

blockjon:
Thanks @malarzm

smolinari:
@blockjon I agree completely with atomicity being entirely important, however...

 >My app isn't even "highly concurrent" per se. At any one moment, I have at most 2 users trying to update a record and I have run into integrity problems.

Here you are talking about concurrency, which I believe is not completely up to the database from a user using an application perspective. Such concurrency must be up to the application, at least partially. 

From what I've learned, in an application you can have either pessimistic concurrency, where you lock out everyone from editing a document or documents, until the user who "checked out" the documents checks them back in. 

Or, you can have optimistic concurrency, where you have versioning of the data. If someone tries to update the data with an older version, they get an "Oops, user X was faster than you and made modifications to the data already. Please refresh your page and make your changes again."  

Either way is suboptimal from a user perspective and I'd say, for running a web based application, the latter with versioning is the better solution. 

To be honest though, I am not sure if or how this all fits in with the strategy(ies) used in MongoDB-ODM for atomicity. I think the two are actually different topics, aren't they? If not, I'd like to understand how they are connected. 

Scott

AronSzigetvari:
@malarzm, I can test it on Monday the earliest. I did tested the previous version but bumped into very mysterious problems. The previous version  broke some tests, and caused some type check errors in Doctrine that didn't happen when I inserted a gettype()  call (but no if,  just a gettype!)

malarzm:
@AronSzigetvari weird indeed... Obviously I'd be very interested in a failing test case if you'd run into such issues again :)

alcaeus:
Looks good to me. The issues I previously reported are gone. :+1:

malarzm:
@AronSzigetvari got any feedback for me by any chance? :)

AronSzigetvari:
@malarzm, I have made a testing for the original problem and it seems to be OK, but I still have a failing test in the application test suite that I want to simplify to a minimal test or find out if we just used it wrong and had worked until now just because of some quirk.

malarzm:
@AronSzigetvari sure, do you think you'll be able to pull it together today or tomorrow?

AronSzigetvari:
@malarzm I am on it now so I very much hope so.

malarzm:
@AronSzigetvari thank you very much! :+1:

AronSzigetvari:
@malarzm I have found the cause of our failing test, but I am not sure if it is a bug or not (a feature, or a fix of a previous bug):

Scenario: merging 2 carts. Cart is document, Item is embeddedDocument.
When we merged cart A and cart B, we added Items of cart B to cart A. Since it is the same Item object, after we call $dm->remove($cartB), we still have both items in $cartA->getItems(), but after calling $dm->flush(), the item that belongs to the removed $cartB is removed from $cartA, too.

I could make the test pass 2 different ways:
1) If I don't use the same Item object, but create a new one with the same data: $cartA($item->copy())
2) It also seems to work if we call $cartB->removeItem($item) when transferring items, so when we flush, the items doesn't belong to the removed parent document any more. (I am not sure if it is really OK, or may cause problems later)

It makes sense since remove is cascaded, but it used to work differently before, so it is a BC in this sense, but I believe it is not a bug. What do you think?

malarzm:
@AronSzigetvari I would love to be able to just reuse embedded documents by passing them around to different documents but unfortunately we're not aware of any good solution for that :( I'd suggest going down 1), I'm personally always (deep) cloning objects that are about to change owner to ensure nothing will clash or have any side effects.

malarzm:
Merged manually in https://github.com/doctrine/mongodb-odm/commit/b1089b723df8ae2230694a3899941e05d560c41f. @AronSzigetvari @blockjon @alcaeus @danez thank you very much for all your help!

AronSzigetvari:
I have made a test that creates an owner document and an embedded document inside, then replaces the embedded, creates a new one with colliding hash value and checks the document state in the UOW. This fails with BETA10 and passes after this fix.
Probably too late to commit it in, since you have already merged the fix.

malarzm:
@AronSzigetvari I've merged @danez test in https://github.com/doctrine/mongodb-odm/commit/4817908f95ff98bb59242d16ca6e4d53a9082504 but if you think it's not enough feel free to submit a PR with passing test only :)

AronSzigetvari:
@malarzm I have added a test, it is a bit similar to @danez 's, but it is a slightly different case (no detached documents were used)

Issue #802, opened 2014-02-14
Make setting of $db in DBRef objects optional


jmikola:
See: https://github.com/doctrine/mongodb-odm/issues/432#issuecomment-34008793

coudenysj:
Any progress on this?

malarzm:
@coudenysj right now we're not working on the feature but there's a PR for which @alcaeus posted list of issues that need to be resolved https://github.com/doctrine/mongodb-odm/pull/807#issuecomment-122551710 - if you want to help it'd be awesome :)

coudenysj:
Owkay! Thanks.

coudenysj:
Working on this in #1349.

malarzm:
Closing as #1349 was merged and is already available in 1.1.0

PR #807, opened 2014-02-24
Partial reference support (without $db property)

sokac:
This PR introduces new annotation property ##partial## for ReferenceOne and ReferenceMany which, when set to true, will create a reference with no $db key.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/DocumentManager.php
+++ b/lib/Doctrine/ODM/MongoDB/DocumentManager.php
@@ -701,8 +701,10 @@ class DocumentManager implements ObjectManager
         $dbRef = array(
             '$ref' => $class->getCollection(),
             '$id'  => $class->getDatabaseIdentifierValue($id),
-            '$db'  => $this->getDocumentDatabase($class->name)->getName(),
         );
+        if ( empty($referenceMapping['partial'])) {
+            $dbRef['$db'] = $this->getDocumentDatabase($class->name)->getName();
+        }
 
         if ($class->hasDiscriminator()) {
             $dbRef[$class->discriminatorField] = $class->discriminatorValue;
--- a/lib/Doctrine/ODM/MongoDB/Mapping/Annotations/ReferenceMany.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/Annotations/ReferenceMany.php
@@ -25,6 +25,7 @@ final class ReferenceMany extends AbstractField
     public $type = 'many';
     public $reference = true;
     public $simple = false;
+    public $partial = false;
     public $targetDocument;
     public $discriminatorField;
     public $discriminatorMap;
--- a/lib/Doctrine/ODM/MongoDB/Mapping/Annotations/ReferenceOne.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/Annotations/ReferenceOne.php
@@ -25,6 +25,7 @@ final class ReferenceOne extends AbstractField
     public $type = 'one';
     public $reference = true;
     public $simple = false;
+    public $partial = false;
     public $targetDocument;
     public $discriminatorField;
     public $discriminatorMap;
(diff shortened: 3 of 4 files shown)


lib/Doctrine/ODM/MongoDB/DocumentManager.php
```diff
@@ -701,8 +701,10 @@ public function createDBRef($document, array $referenceMapping = null)
         $dbRef = array(
             '$ref' => $class->getCollection(),
             '$id'  => $class->getDatabaseIdentifierValue($id),
-            '$db'  => $this->getDocumentDatabase($class->name)->getName(),
         );
+        if ( empty($referenceMapping['partial'])) {
```
jwage: Remove space before empty(

jwage:
This PR is missing support for the other mapping drivers (xml, yaml, etc.)

jwage:
Also, could you explain the reasoning behind wanting to do this?

sokac:
Sorry for the delay. The reason behind this PR is our previous code base had partial db reference (i.e. not having db in DBRef). When you issue a query with DBRef with $db it yields different result than without one. Note that $db parameter is optional (http://docs.mongodb.org/manual/reference/database-references/#dbrefs )

solocommand:
@sokac the `$db` parameter is optional at the server level for the reference to be valid, however it's currently required for the odm to support references across databases.

While I can see the advantage of allowing yourself to bypass that behavior, my approach would be to update the existing data to include the `$db` parameter in all DBRefs.

It's forward compatible if you are going to be moving forward with mongodb-odm, and should? be backwards compatible with your existing system. Not arguing for/against this PR, just pointing out another option.

sokac:
Makes sense completely. Maybe documentation could be update to reference this potential issue.

jmikola:
Sorry, just noticed this after @malarzm linked me. Indeed, this is something we'd like to do, as `$db` is basically useless for single-database applications. We have a tracking ticket for this in #802.

I think we can re-open this for later review. Thanks for the initial PR!

malarzm:
LGTM, just need to mention this in the docs. @jmikola @jwage @alcaeus thoughts?

lib/Doctrine/ODM/MongoDB/DocumentManager.php
```diff
@@ -701,8 +701,10 @@ public function createDBRef($document, array $referenceMapping = null)
...
+            $dbRef['$db'] = $this->getDocumentDatabase($class->name)->getName();
```
alcaeus: Not always setting a ```$db``` key can cause errors in ```Expr::references()``` and ```Expr::includesReferenceTo()``` when using the methods with a reference that has the ```partial``` flag enabled and no ```targetDocument``` set. I'm not sure how much of a use case this is, but either the ```Expr``` class or the mapping drivers need to make sure to handle this case.

alcaeus:
I'm not quite sure about the ```partial``` parameter, especially since we already have ```simple```. With this PR we'd have three different kinds of references:
* "simple" references that are simply just an identifier and contain no other mapping types (currently mapped using ```simple = true```),
* "non-simple" but "partial" references that create a ```DBRef``` object but don't include a ```$db``` key (currently mapped using ```partial = true```), and
* "full" references that create a complete ```DBRef``` object (currently mapped using ```simple = false``` and ```partial = false```)

In my opinion it might be smarter to deprecate the ```simple``` parameter and introduce a new parameter that takes the values ```simple```, ```partial``` and ```full``` (with ```full``` being the default value). Unfortunately, ```type``` is already taken, so we'd need to find another wording that clearly indicates what we're trying to do here.

lib/Doctrine/ODM/MongoDB/DocumentManager.php
malarzm: definitely :+1: for catching this

coudenysj:
I agree with the approach of @alcaeus.

`type` would be the obvious option, maybe we could introduce `dbRefType`?

malarzm:
I'd throw `storeAs` to the proposition pool and ultimately ask @jmikola for a word proposition/judgment :)

coudenysj:
Sounds good too! :)

alcaeus:
:+1: to using `storeAs` as it's more descriptive than `type`.

malarzm:
Superseded by #1349 which is now available in 1.1.0 :)

PR #862, opened 2014-04-24
Add test case for gh 861

szappacosta:


malarzm:
@szappacosta there were two additional issues with your tests:

1. `Doctrine\MongoDB\Event\UpdateEventArgs` is not correct class to test against ODM's `preUpdate` event
2. `Hero::videos` is inverse side therefore `preUpdate` will not be issued if the collection is changed as no update will be issued for `Hero` document

Once these two issues are resolved the test will pass if ran against #1121. Thank you for creating test case!

Issue #642, opened 2013-07-23
spl_object_hash collisions


tecbot:
Hi, today we run in an crazy issue, We found that is releated to an issue which was already reported at doctrine ORM: http://www.doctrine-project.org/jira/browse/DDC-1896

In our case, we run a Behat test of one of our features and all the time PHP returns a hash for a new document which was already used of an embedded document previously. So doctrine thinks that the document is already managed and don't insert it into our database. Only a clear between this operations helps us to prevent this issue.

IMHO we should make the document state behaviour a little bit more robust and shouldn't check only the spl object hash of a document. By the way the documents are not of the same type.

AronSzigetvari:
We have also found the same issue, I have opened a ticket, too: https://github.com/doctrine/mongodb-odm/issues/1017 and made a pull request (https://github.com/doctrine/mongodb-odm/pull/1023 ) but there was no reaction to it, so I don't know if and when it will be merged.
At least this solved the issue in our problem.

malarzm:
This is fixed in 1.0.0 by #782

AronSzigetvari:
Augusztus 10 és 19  között szabadságon leszek, legközelebb augusztus 24-én, hétfőn olvasom a leveleimet. Sürgős esetben kérem hívja a 398-0468-as telefonszámot.


Üdvözlettel:
Szigetvári Áron
vezető fejlesztő

PR #878, opened 2014-05-15
spl_object_hash collision when doing merge() twice

danez:
After spending hours of trying to figure out why our system does not save certain documents, and afterwards hours of producing a reproducable testcase, I finally managed to track it down to the merge inside odm.

##### Background:
We have a **main-document** (1) that embedds on **embedded-document** (1).
We also have an **other-document** with just an identifier.
Initially in the test we get the **main-document** from somewhere other than doctrine (cache for example).

##### Problem:
Now when calling merge on the detached **main-document** (1), the logic inside the *UnitOfWork::doMerge* creates a new **embedded-document** (2), merges it with its detached one and sets it on the managed **main-document** (2).
When calling merge a second time on the same detached **main-document** (1), its "managed version" (2) is already found in the identity map. Still a new **embedded-document** (3) is created and merged with the one from the detached **embedded-document** (1). After the **embedded-document**(3) is merged it is set back on the **main-document**(2), replacing the previous managed **embedded-document**(2). This previous document in our scenario now does not have any references (because documents without identifiers are not in the identity map) and is probably gc by php.
The newly created **other-document** that the test tries to insert afterwards is then detected as managed, as the spl_object_hash for this document is the same as for the unreferenced **embedded-document** (2).

##### Fix:
I already have a fix for that in the second commit, but I'm not sure if it introduces any side effects that I might not be aware of. But I can not see anything bad currently and all the tests still work.

The fix simply detaches the (at least inside odm) not anymore referenced embedded document.

Relates to #642

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
+++ b/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
@@ -2133,6 +2133,12 @@ class UnitOfWork implements PropertyChangedListener
             $prevClass = $this->dm->getClassMetadata(get_class($prevManagedCopy));
 
             if ($assoc['type'] === 'one') {
+                $oldManagedCopy = $prevClass->reflFields[$assocField]->getValue($prevManagedCopy);
+                if (is_object($oldManagedCopy)) {
+                    // remove the current managed copy as it gets replaced with the setValue() call
+                    $subVisited = array();
+                    $this->doDetach($oldManagedCopy, $subVisited);
+                }
                 $prevClass->reflFields[$assocField]->setValue($prevManagedCopy, $managedCopy);
             } else {
                 $prevClass->reflFields[$assocField]->getValue($prevManagedCopy)->add($managedCopy);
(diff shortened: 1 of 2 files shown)


AronSzigetvari:
We have also run into a problem with spl_object_hash collision but in our case the problem occured even without merging a detached copy, only replacing embedded documents and the old one going out of scope. I made a different fix for it (https://github.com/doctrine/mongodb-odm/pull/1023) that works in our case but now I am not sure if the solution would solve our scenario too.

danez:
By looking at your code I think it should also fix our case. But I can't test it right now.

malarzm:
@danez can you let us know if #782 fixes your problem?

danez:
Yeah, the test that is provided in this PR seems to pass on the fix branch.
https://travis-ci.org/danez/mongodb-odm/builds/67753529

I'll create a new PR with just the test and then close this one.

malarzm:
@danez thank you sir! :+1:

PR #879, opened 2014-05-16
Throw proper exception upon missing discriminator field value

ahmedmhmd:


ahmed-araby:
I confirm this bug 
I've no idea till now , how ODM doesn't add the discrimination filed automatically , the $name , is still not enough , simple logic , If the $name was what all needed , then what's the use of discrimination filed then ?!
There must be away to raise error when discrimination field is missing or add it automatically somehow

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -196,6 +196,9 @@ public function __wakeup()
     public function newInstance()
     {
         if ($this->prototype === null) {
+            if ($this->getReflectionClass()->isAbstract() && $this->discriminatorField !== null && $this->discriminatorValue === null) {
```
jmikola: > maybe we can enforce adding discriminator field to projection in the \Doctrine\ODM\MongoDB\Query\Builder?

I believe this will be addressed by #1164 /cc @malarzm

jmikola:
> how ODM doesn't add the discrimination filed automatically

I believe this was only coming up with projections. I wouldn't have expected users to use "include" mode in a projection (e.g. `{ x: 1, y: 1 }`, where all unlisted fields apart from `_id` are implicitly excluded) and still expect the data to be hydrated into a class (instead of using `hydrate(false)` in the query builder and getting raw data back).

In any event, #1164 should help with this. I also think we can still throw an exception here, though, as a fall back, since there's still the possibility that the discriminator doesn't exist in the actual BSON document in MongoDB, and we should have some nicer handling around that.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
malarzm: Looking at the PR's changeset yes, I'd personally even prefer autoincluding the field since discriminator field should be transparent IMO

Issue #891, opened 2014-05-30
RepositoryFactory


hilobok:
Add RepositoryFactory in the same way as in ORM and PHPCR. It will give the ability to inject additional services or parameters into Repository.

PR #892, opened 2014-05-30
Added RepositoryFactory [#891]

hilobok:
Added RepositoryFactory in the same way as in ORM and PHPCR. It will give the ability to inject additional services or parameters into Repository.


jwage:
:+1:

atrauzzi:
Just so I understand correctly: This would make it so that I could make a RepositoryFactory subclass that uses my framework's DI to supply the resultant instance?

malarzm:
@atrauzzi: Yes

atrauzzi:
:tada: :balloon:

PR #896, opened 2014-06-03
QueryBuilder forces refresh() if select() was used

### LINKED ISSUE #895: Fields null when querying object multiple times (using select).
Hi,

I have a problem when querying an object. 

If first query the object by Id and select only some fields (ie: a, b, c). Then i made a second query by slug on the same collection on other fields (ie: a, d, e). 
If the object is different, this works. but if the object is the same, the query return the object already return on the first query and the application don't work (because there is some fields null).

I can't select all fields because there is many fields and for performance reason i only want some fields.
As the two query have different parameters, i don't known if the document will be equals or different.

How can i resolve this problem ?

Thanks

malarzm:
Fixes #895

I have tried detecting that document managed by UoW contains less data than just fetched ``$data`` in ``UnitOfWork::getOrCreateDocument`` but it was ~0.4s slower (for whole test suite) so imo it's no way to go.


malarzm:
This was old and bad idea ;)

Issue #715, opened 2013-12-05
Embedded document with discriminatorMap does not get persisted while using Hydrator.


veewee:
# TESTCASE:

``` php
<?php

namespace Doctrine\ODM\MongoDB\Tests;

use Doctrine\Common\Collections\ArrayCollection;
use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;

class GH715Test extends BaseTest
{

    public function testEmbeddedDocumentsWithDiscriminatorMapShouldBeInserted()
    {
        $document = new GH715User();
        $document->tasks->add(new GH715DownloadTask());
        $document->tasks->add(new GH715BuildTask());

        $this->assertEmebeddedDocumentsInserted($document);
    }


    public function testHydratedRawDocumentsWithDiscriminatorMapShouldBeInserted()
    {
        $document = new GH715User();
        $data = array(
            'tasks' => array(
                array('taskType' => 'download', 'name' => 'Download name'),
                array('taskType' => 'build', 'name' => 'Build name'),
            )
        );

        $hydrator = $this->dm->getHydratorFactory()->getHydratorFor(get_class($document));
        $hydrator->hydrate($document, $data);

        $this->assertEmebeddedDocumentsInserted($document);
    }

    public function testHydratedDocumentsWithDiscriminatorMapShouldBeInserted()
    {
        $document = new GH715User();
        $data = array(
            'tasks' => array(
                new GH715DownloadTask(),
                new GH715BuildTask(),
            )
        );

        $hydrator = $this->dm->getHydratorFactory()->getHydratorFor(get_class($document));
        $hydrator->hydrate($document, $data);

        $this->assertEmebeddedDocumentsInserted($document);
    }

    protected function assertEmebeddedDocumentsInserted(GH715User $document)
    {

        $this->dm->persist($document);
        $this->dm->flush();
        $this->dm->clear();

        $query = $this->dm->createQueryBuilder(__NAMESPACE__ . '\GH715User')
            ->field('id')->equals($document->id)
            ->getQuery();

        $document = $query->getSingleResult();

        $this->assertInstanceOf('Doctrine\ODM\MongoDB\PersistentCollection', $document->tasks);
        $this->assertEquals(2, $document->tasks->count());

        $iterator = $document->tasks->getIterator();
        $this->assertInstanceOf(__NAMESPACE__ . '\GH715DownloadTask', $iterator[0]);
[...]

PR #909, opened 2014-06-24
Fix cascade persist for associations.

kimlai:
This PR is somewhat related to https://github.com/doctrine/mongodb-odm/issues/715 I think (the last comment at least).

During a cascade persist operation on a *Many* association, a snapshot of the associated collection is [taken](https://github.com/doctrine/mongodb-odm/blob/master/lib/Doctrine/ODM/MongoDB/UnitOfWork.php#L473), marking the collection as "not dirty".  So if items were added to the collection before persisting, and the collection was never initialized, the items are lost (persisted in DB but not present in the instanciated collection anymore).

By using the previously taken snapshot during the collection initialization, we can re-add the lost items after loading the collection.

That's the first time I take a look at Doctrine's internals, so I very well might be completely off with the solution, but I think the use case really is bugged. Or did I miss something ?

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/PersistentCollection.php
+++ b/lib/Doctrine/ODM/MongoDB/PersistentCollection.php
@@ -175,6 +175,10 @@ class PersistentCollection implements BaseCollection
                 // Has NEW objects added through add(). Remember them.
                 $newObjects = $this->coll->toArray();
             }
+            else if (! empty($this->snapshot)) {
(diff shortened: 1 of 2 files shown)


malarzm:
@kimlai the test you have written for this PR is passing on current dev-master. If something is still not working as expected please feel free to reopen issue :)

kimlai:
Great news, thanks !

PR #953, opened 2014-10-03
Add support for alternate Hydrator generation methods

kperry42:
This implements two new auto-gen modes for hydrators: AUTOGENERATE_FILE_NOT_EXISTS and AUTOGENERATE_EVAL, per our discussion labelled issue 889.

When I dug into the code, I discovered that \Doctrine\Common\Proxy\AbstractProxyFactory already implemented exactly the same algorithms that I wanted for Hydrators, even including eval() mode.  So I've closely patterned my HydratoryFactory mods after the code in AbstractProxyGenerator. I also put the constant definitions for the different auto-gen modes into the Configuration class, for easy reference.


lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
```diff
@@ -137,10 +138,27 @@ public function getHydratorFor($className)
 
         if ( ! class_exists($fqn, false)) {
             $fileName = $this->hydratorDir . DIRECTORY_SEPARATOR . $hydratorClassName . '.php';
-            if ($this->autoGenerate) {
-                $this->generateHydratorClass($class, $hydratorClassName, $fileName);
+            switch ($this->autoGenerate) {
```
jwage: This property needs to be updated now that it is no longer a boolean.

Here:

https://github.com/kperry42/mongodb-odm/blob/issue889/lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php#L63

and here:

https://github.com/kperry42/mongodb-odm/blob/issue889/lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php#L96

kperry42: Good point.  Done.

malarzm:
Merged in https://github.com/doctrine/mongodb-odm/commit/f04de2ed82760d0a490d0ea80e2ebfb85eea335a - thanks @kperry42!

jmikola:
Late to the party here, but if we end up adopting Common's proxy generator down the line, will this port over nicely? I don't recall if we had an outstanding TODO ticket for that, but I seem to recall talking about it with @Ocramius last year.

PR #958, opened 2014-10-09
Fix inheritance mapping when extending documents

### LINKED ISSUE #927: ClassMetadata: associationMapping is not filled for association inherited from documents
We have the following use case:
```php
/** @ODM\Document */
class ParentDocument
{
    /** @ODM\Id */
    protected $id;

    /** @ODM\ReferenceOne(targetDocument="ReferencedDocument") */
    protected $reference;
}

/** @ODM\Document */
class ChildDocument extends ParentDocument
{
}

/** @ODM\Document */
class ReferencedDocument
{
    /** @ODM\Id */
    protected $id;
}
```

When looking at the classMetadata for the child document, there is no associationMapping recorded for the reference association. When changing ParentDocument to a mapped superclass, the associationMapping is recorded correctly.

Update: Test for the error: https://github.com/VoycerAG/mongodb-odm/commit/57be9c58c21f0dbfd1ae728bdb2d6d5871d7413d

It might be enough to add the associationMapping when adding an inherited field. However, there are differences in the handling between documents and mapped superclasses, so I wasn't quite sure where to add this.

alcaeus:
This PR fixes #927.

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH927Test.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH927Test.php
@@ -1,0 +1,39 @@
+<?php
+
+namespace Doctrine\ODM\MongoDB\Tests\Functional\Ticket;
+
+use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
+
+class GH927Test extends \Doctrine\ODM\MongoDB\Tests\BaseTest
+{
+    public function testInheritedClassHasAssociationMapping()
+    {
+        $parentMetadata = $this->dm->getClassMetadata('Doctrine\ODM\MongoDB\Tests\Functional\Ticket\GH927Parent');
+        $this->assertArrayHasKey('reference', $parentMetadata->associationMappings);
+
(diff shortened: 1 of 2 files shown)


alcaeus:
Whoops. I should learn what buttons are safe to click on.

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH927Test.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
+namespace Doctrine\ODM\MongoDB\Tests\Functional\Ticket;
+
+use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
+
+class GH927Test extends \Doctrine\ODM\MongoDB\Tests\BaseTest
+{
+    public function testInheritedClassHashAssociationMapping()
```
malarzm: Hash -> Has :)

malarzm:
@alcaeus Luckily (almost) everything can be restored ;) could you fix typo and squash commits?

malarzm:
This has been merged in https://github.com/doctrine/mongodb-odm/commit/55ce2b3ad32fcf3eef1e8a733d8a479c966501d3 - thanks @alcaeus!

PR #960, opened 2014-10-09
The introduction should mention that you have to manually add the getter...

gistya:
...s and setters.

PR #961, opened 2014-10-09
Fix this demo code so it would actually work.

gistya:


jmikola:
I noticed that this builds upon #960. Manually merged in https://github.com/doctrine/mongodb-odm/commit/1343375a4edf3e2157138e442441654e25fabbec. I also made further revisions in https://github.com/doctrine/mongodb-odm/commit/eec3989a9caca9fa2a87740825284cc8cfd8c1e6, which is included in the merge. Thanks!

Issue #963, opened 2014-10-14
PersistentCollection count, isDirty and initialize


redthor:
Hi there,
I'm doing the following:
1. Adding an object to PersistentCollection
2. Saving PersistentCollection
3. Attempting to get the object out of the PersistentCollection

What I find is that `$this->pColl->count()` is returning 1, which is what I expect. Note `count()` does not call `initialize()`.

When I call `toArray()` `initialize()` is called and the `isDirty` is false, so the collection is cleared and the object I added is lost. I get zero members back and now `count()` is zero.

The PersistentCollection is being saved to the DB.

I'm using the latest "dev-master".

Any help is appreciated.
Thanks.

rubenrua:
In https://gist.github.com/rubenrua/55ed88c087311542c563 you can found a test to debug this error.

malarzm:
https://github.com/doctrine/mongodb-odm/pull/981 seems to be fixing this error as well

jmikola:
Fixed by #981, which was merged in https://github.com/doctrine/mongodb-odm/commit/b00177ce241214cfaf2eb2a01885d964a55a8ed8.

Issue #921, opened 2014-07-25
ReferenceMany Collection after flush


Juliens:
After a flush (in the same request), the count of the ReferenceMany Collection give the right count, but if we "foreach" we have the first collection data that the repository give.

```
<?php
/**
 * @ODM\Document
 */
class User
{
    /**
     * @ODM\id
     */
    private $id;

    /**
     * @ODM\String
     */
    private $name;

    public function setName($name)
    {
        $this->name = $name;
        return $this;
    }

    public function getName($name)
    {
        return $this->name;
    }

    /**
     * @ODM\ReferenceMany(targetDocument="Post")
     */
    private $posts;

    public function __construct()
    {
        $this->posts = array();
    }

    public function addPost(Post $post)
    {
        $this->posts[] = $post;
    }

    public function getPosts()
    {
        return $this->posts;
    }
}


/**
 * @ODM\Document
 */
class Post
{
    /**
     * @ODM\id
     */
    private $id;

    /**
     * @ODM\ReferenceOne(targetDocument="User")
     */
    private $user;

    public function setUser($user)
    {
        $this->user = $user;
    }

}




$user = $dm->getRepository('User')->findOneByName('smith');

$post = new Post();
$dm->persist($post);

$user->addPost($post);

$dm->persist($user);
$dm->flush();



var_dump(count($user->getPosts()));
foreach ($user->getPosts() as $post) {
    var_dump('POST');
}

$dm->refresh($user);
var_dump(count($user->getPosts()));
foreach ($user->getPosts() as $post) {
    var_dump('POST');
}
```

Output

```
int(1)
int(1)
POST
```

jmikola:
Fixed by #981, which was merged in https://github.com/doctrine/mongodb-odm/commit/b00177ce241214cfaf2eb2a01885d964a55a8ed8.

PR #964, opened 2014-10-16
preserve dirty state - refs #963

nail:
Since bd573cb7 the dirty state is reset to false when you call `PersistentCollection::initialize()` even if extra elements have been added to the collection before being called.

This addresses #963 but I'm not sure if it could have other implication (tests pass 100%)

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/PersistentCollection.php
+++ b/lib/Doctrine/ODM/MongoDB/PersistentCollection.php
@@ -194,6 +194,10 @@ class PersistentCollection implements BaseCollection
             }
         }
 
+        if ($newObjects) {
+            $this->isDirty = true;
+        }
+
         $this->mongoData = array();
         $this->initialized = true;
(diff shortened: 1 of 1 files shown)


jmikola:
@nail: #981 has since been merged in https://github.com/doctrine/mongodb-odm/commit/b00177ce241214cfaf2eb2a01885d964a55a8ed8.

I did end up reverting the `isDirty` change (an error on my part) in https://github.com/doctrine/mongodb-odm/commit/f3f289866293c8173b7e3429c7cacd3fed3ab0be and I now test for it (see: [here](https://github.com/doctrine/mongodb-odm/commit/819577f8b884b9587022eda8adcf012d5e265650#diff-267a328f038c5597f551d666461ed1e6R52)); however, this was not the complete fix for #921 and #963, which is why I merged @malarzm's PR instead.

Just the same, thanks for looking into this!

Issue #971, opened 2014-11-06
FindAndUpdate on inherited documents always create new object


slawus:
Hi!
There is problem with creating inherited documents on findAndUpdate invocation. Let's assume we have documents:
```php
/**
 * @ODM\Document
 * @ODM\InheritanceType("SINGLE_COLLECTION")
 * @ODM\DiscriminatorField(fieldName="type")
 * @ODM\DiscriminatorMap({"car"="Car", "bicycle"="Bicycle"})
 */
class Vehicle
{
    /** @ODM\Id */
    public $id;

    /** @ODM\String */
    public $name;

    /** @ODM\EmbedMany */
    public $features;
}

/**
 * @ODM\Document
 */
class Car extends Vehicle {}

/**
 * @ODM\Document
 */
class Bicycle extends Vehicle {}
```

If we for example add document using findAndUpdate, it won't contain discriminator field. E.g:
```php
        $this->dm->createQueryBuilder(__NAMESPACE__ . '\Car')
            ->findAndUpdate()
            ->upsert(true)
            ->field('name')->equals($name)
            ->sort('_id', -1)
            ->field('features')->push('engine')
            ->getQuery()->execute();
```
will create:
```json
{
    _id: ObjectId("545bafb26ece71ada2f88bb1"),
    name: "Ferrari",
    features: [
        "engine"
    ]
}
```

And now on the second attempt, becouse of lacking "type" field new document will be created, instead of updating previous one.

alcaeus:
Fix was merged in https://github.com/doctrine/mongodb-odm/commit/a5859f5f0073bd3b65f606225975d147388eb7a0.

PR #972, opened 2014-11-06
fix update of inherited documents on findAndUpdate [#971]

slawus:
Fix for error from #971 ticket. When looking for particular document, DocumentPersist use $in to query particular type of documents, that are inherited in "SINGLE_COLLECTION" way. It works great in case of looking for a file, but has no influence on document creation on upsert update. Added condition, which change "$in" into simple equation, if there is only one discriminator value.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -844,7 +844,11 @@ class DocumentPersister
          */
         if ($this->class->hasDiscriminator() && ! isset($preparedQuery[$this->class->discriminatorField])) {
             $discriminatorValues = $this->getClassDiscriminatorValues($this->class);
-            $preparedQuery[$this->class->discriminatorField] = array('$in' => $discriminatorValues);
+            if((count($discriminatorValues) == 1)) {
+                $preparedQuery[$this->class->discriminatorField] = $discriminatorValues[0];
+            } else {
+                $preparedQuery[$this->class->discriminatorField] = array('$in' => $discriminatorValues);
+            }
         }
 
         return $preparedQuery;
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH971Test.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH971Test.php
@@ -9,10 +9,10 @@ class _Test extends \Doctrine\ODM\MongoDB\Tests\BaseTest {
 
     public function testUpdateOfInheritedDocumentUsingFindAndUpdate() {
         $name = "Ferrari";
-        $features = [
+        $features = array(
             "Super Engine",
             "Huge Wheels"
-        ];
+        );
 
         //first query, create Car with name "Ferrari"
         $this->dm->createQueryBuilder(__NAMESPACE__ . '\Car')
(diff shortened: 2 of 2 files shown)


tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH971Test.php
```diff
@@ -0,0 +1,66 @@
+<?php
+
+
+namespace Doctrine\ODM\MongoDB\Tests\Functional\Ticket;
+
+use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
+
+class _Test extends \Doctrine\ODM\MongoDB\Tests\BaseTest {
```
jwage: Can you make this a more unique name?

jmikola: This should be GH971Test. If @slawus doesn't get to it, I suppose we can change it before merging this.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -844,7 +844,11 @@ public function addDiscriminatorToPreparedQuery(array $preparedQuery)
          */
         if ($this->class->hasDiscriminator() && ! isset($preparedQuery[$this->class->discriminatorField])) {
             $discriminatorValues = $this->getClassDiscriminatorValues($this->class);
-            $preparedQuery[$this->class->discriminatorField] = array('$in' => $discriminatorValues);
+            if((count($discriminatorValues) == 1)) {
+                $preparedQuery[$this->class->discriminatorField] = $discriminatorValues[0];
+            } else {
+                $preparedQuery[$this->class->discriminatorField] = array('$in' => $discriminatorValues);
+            }
```
jmikola: I think this PR needs additional checks to ensure that upserts never resort to using `$in` when dealing with discriminators. We should throw an exception that tells users they must use a specific class repository or query builder so that we do have a single discriminator to work with.

As-is, there's still an edge case where the user runs an upsert against a base class query builder and we miss including a discriminator value in the newly inserted document.

jmikola:
Bumping this for consideration in a 1.0.x patch release.

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH971Test.php
munkie: Opening braces for classes MUST go on the next line, and closing braces MUST go on the next line after the body. PSR-2

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH971Test.php
```diff
@@ -0,0 +1,66 @@
...
+
+    public function testUpdateOfInheritedDocumentUsingFindAndUpdate() {
```
munkie: Opening braces for methods MUST go on the next line, and closing braces MUST go on the next line after the body. PSR-2

slawus:
Sorry for such huge delay. I've changed tests class name. Also moved brackets, so this clases matches PSR-2.

malarzm:
@slawus welcome back! Would you be willing to work on @jmikola's remark? https://github.com/doctrine/mongodb-odm/pull/972#discussion_r34840957

slawus:
Definitely I would like to help, unfortunately I can't guarantee that I will find free time to do it :/ So feel free to help! :)

malarzm:
@slawus I've included your commit in #1319 - thank you for fixing this part!

slawus:
Great!

PR #975, opened 2014-11-13
Update annotations reference for @Timestamp clarification

### LINKED ISSUE #974: Add note to discourage use of Timestamp type
The docs currently have a brief example with `@Timestamp` used on a `$created` entity field. As-is, users may inadvertently use this type mapping for dates, when they should actually use `@Date`.

See: http://stackoverflow.com/a/26896545/162228

solocommand:
Fixes #974 

Added note that the @Timestamp annotation should not be used for storing Dates

### DIFF
--- a/docs/en/reference/annotations-reference.rst
+++ b/docs/en/reference/annotations-reference.rst
@@ -1127,6 +1127,11 @@ Defines that the annotated instance variable holds a timestamp.
 
     /** @Timestamp */
     private $created;
+    
+.. note::
+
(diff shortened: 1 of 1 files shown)


malarzm:
Turned out this was solved in https://github.com/doctrine/mongodb-odm/commit/621a4481c9fe8f71fc12fdd01ebd1f9d501f4381#diff-1457aa7b8fa8bab45fea926215c23cbaR1150 I won't merge this then but thanks anyway @solocommand!

solocommand:
@malarzm Sounds good

Issue #978, opened 2014-11-19
detach not cascading for embedded documents


fridolin-koch:
It seems that `detach()` is not working correctly for embedded documents. I also think I a have a fix for it.

This is the structure of my mapping (Only important informations):

```
abstract AbstractDocument: MappedSuperclass
    - children: EmbedMany(targetDocument="ChildDocument")
    - fields:EmbedMany(targetDocument="DocumentField")

class Document extends AbstractDocument: Document
    - transmissionDetails: EmbedOne(targetDocument="TransmissionDetails")
```

I'm detaching the objects with this code:

```php

echo PHP_EOL . '===== SIZE before => ' . $this->manager->getUnitOfWork()->size() . PHP_EOL;

$this->manager->detach($document);
$this->manager->detach($mapping);

foreach ($this->manager->getUnitOfWork()->getIdentityMap() as $name => $value) {
    echo $name . ' ==> ' . count($value) . PHP_EOL;
}

echo PHP_EOL . '===== SIZE after => ' . $this->manager->getUnitOfWork()->size() . PHP_EOL;
```

The result:

```
===== SIZE before => 821
Document\Document ==> 0
Document\ImportSource ==> 0
Document\DocumentState ==> 1
Document\File ==> 0
Document\TransmissionDetails ==> 1
Document\DocumentType ==> 2
Document\Mapping ==> 1
Document\FileTransportInterface ==> 1
Document\SftpTransportInterface ==> 1
Document\GlobalLocationNumber ==> 3
Document\DocumentField ==> 800
Document\FtpTransportInterface ==> 1
Document\ChildDocument ==> 4
Document\FileType ==> 1
===== SIZE after => 816
```

I followed the Call-Stack to `UnitOfWork::cascadeDetach()`. I'm not sure if the recursive call of `cascadeDetach` is correct. So I tried replacing the two recursive `cascadeDetach` call with `doDetach`:

```php
private function cascadeDetach($document, array &$visited)
{
    $class = $this->dm->getClassMetadata(get_class($document));
    foreach ($class->fieldMappings as $mapping) {
        if ( ! $mapping['isCascadeDetach']) {
            continue;
        }
        if (isset($mapping['embedded'])) {
            $relatedDocuments = $class->reflFields[$mapping['fieldName']]->getValue($document);
            if (($relatedDocuments instanceof Collection || is_array($relatedDocuments))) {
                if ($relatedDocuments instanceof PersistentCollection) {
[...]

PR #979, opened 2014-11-20
Fix detach cascading see #978 

fridolin-koch:
As described in #978 I think the detach cascading for embedded documents is not working correct. I also added  a small test.

I was not sure where to put the test.. I hope it is correct this way.


malarzm:
This has been merged manually in https://github.com/doctrine/mongodb-odm/commit/994783a5edfd5f2cbe29f709cf49cbb5ad40ccb7 - thanks @fridolin-koch!

jmikola:
@fridolin-koch, @malarzm: Comparable logic existed in `cascadeRefresh()`, which recurses with `cascadeRefresh()` instead of `doRefresh()` on the embedded document(s). I suppose that's alright since embedded documents would never be refreshed on their own (unlike detachment). Would you agree?

malarzm:
I agree with that :)

fridolin-koch:
I also agree :+1:

PR #980, opened 2014-11-25
@EmbeddedDocument annotation reference fix

### LINKED ISSUE #907: Remove example of embedding a Document-annotated class
A user in IRC mentioned that https://github.com/doctrine/mongodb-odm/blob/master/docs/en/reference/annotations-reference.rst#embeddeddocument included an example of embedding an object mapped with either Document or EmbeddedDocument. With Document, this would allow the object to be persisted to its own collection as well as the owning document of the embed relationship.

This doesn't sound kosher, and I don't believe there is any other reference to this in the documentation. I suggest we simply delete the Document example.

If users do want the same type of object to be useful as a top-level and embedded document, my suggestion is to define the common document as a MappedSuperclass and then extend with concrete Document and EmbeddedDocument child classes. That also avoids any undefined behavior and questions with the current example (e.g. "how is the embedded Document-mapped class tracked by UnitOfWork?").

/ping @jwage

malarzm:
Fixes #907

### DIFF
--- a/docs/en/reference/annotations-reference.rst
+++ b/docs/en/reference/annotations-reference.rst
@@ -456,20 +456,6 @@ above code to work, you should have:
     <?php
 
     /**
-     * @Document
-     */
-    class Money
-    {
-    //...
-    }
-
-or
-
-.. code-block:: php
-
-    <?php
-
-    /**
      * @EmbeddedDocument
      */
     class Money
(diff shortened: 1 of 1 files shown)


malarzm:
Closed with https://github.com/doctrine/mongodb-odm/commit/621a4481c9fe8f71fc12fdd01ebd1f9d501f4381

PR #981, opened 2014-11-26
Fix PersistentCollection init state during flush()

### LINKED ISSUE #921: ReferenceMany Collection after flush

### LINKED ISSUE #963: PersistentCollection count, isDirty and initialize

malarzm:
Fixes #921 
EDIT: Fixes #963 as well

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/CollectionPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/CollectionPersister.php
@@ -110,6 +110,7 @@ class CollectionPersister
 
             case 'addToSet':
             case 'pushAll':
+                $coll->initialize();
                 $this->deleteElements($coll, $options);
                 $this->insertElements($coll, $options);
                 break;
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/ReferencesTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/ReferencesTest.php
@@ -191,6 +191,55 @@ class ReferencesTest extends \Doctrine\ODM\MongoDB\Tests\BaseTest
         $this->assertEquals('test', $groups[0]->getName());
         $this->assertEquals(1, count($groups));
     }
+    
+    public function testFlushInitializesEmptyPersistentCollection()
+    {
+        $user = new User();
+        $this->dm->persist($user);
+        $this->dm->flush();
+        $this->dm->clear();
+        
+        $user = $this->dm->getRepository('Documents\User')->find($user->getId());
+        
+        $user->addGroup(new Group('Group 1'));
+        $user->addGroup(new Group('Group 2'));
+
+        $this->dm->persist($user);
+        $this->dm->flush();
+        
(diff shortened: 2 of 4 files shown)


rubenrua:
+1

jmikola:
Manually merged in https://github.com/doctrine/mongodb-odm/commit/b00177ce241214cfaf2eb2a01885d964a55a8ed8.

@rubenrua: I ended up replacing your test case, as I already had a test with more coverage back from when #921 was originally reported. Nevertheless, thanks for the effort on that.

@malarzm: I also reverted the removal of setting `isDirty` from https://github.com/doctrine/mongodb-odm/commit/bd573cb7def6ab22df8c465b0d79003d26e28046 and added a test for `isDirty` is preserved. That was the only functional change from that original commit, and although it doesn't seem to have caused #921 and #963 on its own (based on some quick testing), removing it was an error on my part. Thanks very much for resolving this.

malarzm:
:+1:

rubenrua:
@jmikola I updated my test with a bi-directional reference and it's broken again.

See: https://github.com/doctrine/mongodb-odm/pull/1014

PR #983, opened 2014-11-27
Do not set customRepository for EmbeddedDocument

### LINKED ISSUE #837: Do not allow custom repository class for embedded documents
Hello there,

I haven't time to actually test it yet,
But it seems that setting the document custom repository behaviour is inconsitant through yaml/xml and annotation.

In fact, by reading at the code, it seems that only "document" or "mappedSuperClass" can have a custom repository through yaml (https://github.com/doctrine/mongodb-odm/blob/master/lib/Doctrine/ODM/MongoDB/Mapping/Driver/YamlDriver.php#L64) and through xml (https://github.com/doctrine/mongodb-odm/blob/master/lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php#L57) whereas any type of document can have a custom repository through annotation (https://github.com/doctrine/mongodb-odm/blob/master/lib/Doctrine/ODM/MongoDB/Mapping/Driver/AnnotationDriver.php#L127).

Which one should it be?

malarzm:
Fixes #837

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataInfo.php
@@ -506,6 +506,10 @@ class ClassMetadataInfo implements \Doctrine\Common\Persistence\Mapping\ClassMet
      */
     public function setCustomRepositoryClass($repositoryClassName)
     {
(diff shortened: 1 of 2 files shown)


jwage:
:+1:

Issue #225, opened 2012-01-08
SchemaManager::prepareIndexes() should allow unmapped fields


jmikola:
Per my comment in dd10565b70e2a82a302264603a6b909eebc49b29:

[Multikey](http://www.mongodb.org/display/DOCS/Multikeys) indexes on fields within an array of embedded objects is no longer supported. For instance, it's not possible to apply an index to `tags.name` for the following document schema:

``` json
{ tags: [ { name: 'foo', weight: 1 }, { name: 'bar', weight: 2 } ] }
```

I understand @tecbot's intentions beyond the series of commits adding field name preparation in various places, but the trade-off has been causing ODM to throw exceptions whenever unmapped fields are referenced. A previous problem (since fixed in #216) was indexing discriminator fields.

Slightly off-topic: IMO this problem extends beyond SchemaManager. I'd love the option to use raw field names in queries, indexes, etc. Does anyone else see value in this?

PR #984, opened 2014-11-27
Consider discriminated Embeds in index generation

malarzm:
Related discussion: https://github.com/doctrine/mongodb-odm/issues/225#issuecomment-63811609

Long story short ``SchemaManager`` needs ``targetDocument`` specified to go deeper while creating indexes. With this modification documents specified in ``discriminatorMap`` will also be checked for any indexes defined there.

cc @bensquire

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/SchemaManager.php
+++ b/lib/Doctrine/ODM/MongoDB/SchemaManager.php
@@ -164,21 +164,29 @@ class SchemaManager
 
         // Add indexes from embedded & referenced documents
         foreach ($class->fieldMappings as $fieldMapping) {
-            if (isset($fieldMapping['embedded']) && isset($fieldMapping['targetDocument'])) {
-                if (isset($embeddedDocumentIndexes[$fieldMapping['targetDocument']])) {
-                    $embeddedIndexes = $embeddedDocumentIndexes[$fieldMapping['targetDocument']];
+            if (isset($fieldMapping['embedded'])) {
+                if (isset($fieldMapping['targetDocument'])) {
+                    $possibleEmbeds = array($fieldMapping['targetDocument']);
+                } elseif (isset($fieldMapping['discriminatorMap'])) {
+                    $possibleEmbeds = array_unique($fieldMapping['discriminatorMap']);
                 } else {
-                    $embeddedIndexes = $this->doGetDocumentIndexes($fieldMapping['targetDocument'], $visited);
-                    $embeddedDocumentIndexes[$fieldMapping['targetDocument']] = $embeddedIndexes;
+                    continue;
                 }
-                foreach ($embeddedIndexes as $embeddedIndex) {
-                    foreach ($embeddedIndex['keys'] as $key => $value) {
-                        $embeddedIndex['keys'][$fieldMapping['name'] . '.' . $key] = $value;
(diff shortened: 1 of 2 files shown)


tests/Doctrine/ODM/MongoDB/Tests/Functional/IndexesTest.php
```diff
@@ -337,4 +353,27 @@ class EmbeddedManyDocumentWithIndexes
...
+
+/** @ODM\EmbeddedDocument */
+class YetAnotherEmbeddedDocumentWithIndex
+{
+    /** @ODM\String @ODM\Index */
+    public $value;
+}
+
+/** @ODM\Document */
+class DocumentWithIndexInDiscrimantedEmbeds
```
jmikola: Type: "Discriminated"

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -163,16 +163,24 @@ private function doGetDocumentIndexes($documentName, array &$visited)
...
-                $embeddedIndexes = $this->doGetDocumentIndexes($fieldMapping['targetDocument'], $visited);
-                foreach ($embeddedIndexes as $embeddedIndex) {
-                    foreach ($embeddedIndex['keys'] as $key => $value) {
-                        $embeddedIndex['keys'][$fieldMapping['name'] . '.' . $key] = $value;
-                        unset($embeddedIndex['keys'][$key]);
+            if (isset($fieldMapping['embedded'])) {
+                if (isset($fieldMapping['targetDocument'])) {
+                    $possibleEmbeds = array($fieldMapping['targetDocument']);
+                } elseif (isset($fieldMapping['discriminatorMap'])) {
+                    $possibleEmbeds = $fieldMapping['discriminatorMap'];
```
jmikola: Maybe a stupid question, but the embedded class is going to be a FQCN thanks to behavior in `ClassMetadataInfo::setDiscriminatorMap()`, correct? This would be similar to how `ClassMetadataInfo::mapField()` prefixes the parent class' namespace on the target document if it doesn't have one.

jmikola:
@malarzm: I think this is good to merge once the typo is fixed.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
malarzm: Yes, both methods seems to be in line when it comes to namespace resolution. However ``ClassMetadataInfo::setDiscriminatorMap()`` is also populating ``ClassMetadataInfo::subClasses`` and ``ClassMetadataInfo::mapField()`` is not. By the way this could be made more DRY I think :)

malarzm:
@jmikola typo is fixed

lib/Doctrine/ODM/MongoDB/SchemaManager.php
jmikola: One more thought: if there's any chance `discriminatorMap` could contain duplicate class names (perhaps the same class is mapped under two different discriminator values), I think we should use `array_unique()` on it.

jmikola:
LGTM once the following is addressed: 

 * [ ] Consider `array_unique($fieldMapping['discriminatorMap'])` if classes could be duplicated

malarzm:
``array_unique`` is added and PR is rebased

Issue #883, opened 2014-05-20
Events not dispatched when changes are made to a ReferenceMany field


solocommand:
When making a change to a ReferenceMany field, no update/remove events are dispatched. It looks like this is intentional, as the events are dispatched in the executeUpdates method of the UoW, but changes to collections do not use this method (instead using the collectionPersister after the document updates are executed).

It doesn't look like it will be easy to change without reworking how refmany collection changes are handled.

If this is intentional, can it be documented somewhere that events will not be fired when modifying only a ReferenceMany field?

jmikola:
#985 is merged, so I believe this can be closed.

PR #985, opened 2014-11-28
Updates to document's PersistentCollection now trigger update

### LINKED ISSUE #922: Version number does not increment when you remove an embedded collection
I have a Doctrine ODM document that uses the `@Version` feature. One of the properties on my document is an embedded collection.

I notice that when I make additive updates to my document, the version number is incremented via this Doctrine code:

https://github.com/doctrine/mongodb-odm/blob/master/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php#L348-L362

However, I also noticed that when I remove from my document the embedded collection value, the version does not increment.

This is likely due to the differences deletions are triggered via this code:

https://github.com/doctrine/mongodb-odm/blob/master/lib/Doctrine/ODM/MongoDB/Persisters/CollectionPersister.php#L73-L88

I would like to be able to have my version number increment when items are removed from my document even when the item is an embedded collection.

How do I make my version number increment when I remove my embedded collection from my document?

Example test which demonstrates the problem:
https://gist.github.com/blockjon/857d7cba7cdb4c865cf9

malarzm:
Fixes #922, #883, and #1039.

Would be great to know if this is good approach, especially if assumption in UoW:2750 is correct :)

cc @blockjon, @solocommand could you take a look if this indeed fixes your issues?

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -337,7 +337,7 @@ class DocumentPersister
         $id = $this->uow->getDocumentIdentifier($document);
         $update = $this->pb->prepareUpdateData($document);
 
-        if ( ! empty($update)) {
+        if ( ! empty($update) || $this->uow->hasCollectionScheduledForCommiting($document)) {
 
             $id = $this->class->getDatabaseIdentifierValue($id);
             $query = array('_id' => $id);
@@ -359,6 +359,12 @@ class DocumentPersister
                     $query[$versionMapping['name']] = new \MongoDate($currentVersion->getTimestamp());
                     $this->class->reflFields[$this->class->versionField]->setValue($document, $nextVersion);
                 }
+            }
+            
+            // We got here because $document has related PersistentCollections to be 
+            // commited later but if it's not versioned then there's nothing left to do
+            if (empty($update)) {
+                return;
             }
 
             // Include locking logic so that if the document object in memory is currently
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -337,7 +337,7 @@ class DocumentPersister
         $id = $this->uow->getDocumentIdentifier($document);
         $update = $this->pb->prepareUpdateData($document);
 
-        if ( ! empty($update) || $this->uow->hasCollectionScheduledForCommiting($document)) {
+        if ( ! empty($update) || $this->uow->hasScheduledCollections($document)) {
 
             $id = $this->class->getDatabaseIdentifierValue($id);
             $query = array('_id' => $id);
--- a/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
+++ b/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
@@ -2743,8 +2743,10 @@ class UnitOfWork implements PropertyChangedListener
     private function collectionScheduledForCommiting(PersistentCollection $coll)
     {
         $document = $coll->getOwner();
-        if (!$document || !$this->isInIdentityMap($document)) {
-            return;
+        $class = $this->dm->getClassMetadata(get_class($document));
+        while ($class->isEmbeddedDocument) {
+            list(, $document, ) = $this->getParentAssociation($document);
+            $class = $this->dm->getClassMetadata(get_class($document));
         }
         $this->hasScheduledCollections[spl_object_hash($document)] = true;
         // If owner is not scheduled for any action then most probably we're
--- a/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
+++ b/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
@@ -189,7 +189,7 @@ class UnitOfWork implements PropertyChangedListener
      * 
      * @var array
      */
-    private $hasCollectionScheduledForCommiting = array();
+    private $hasScheduledCollections = array();
 
     /**
      * List of collections visited during changeset calculation on a commit-phase of a UnitOfWork.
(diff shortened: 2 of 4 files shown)


lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -2698,6 +2721,38 @@ public function isCollectionScheduledForUpdate(PersistentCollection $coll)
...
+     */
+    private function collectionScheduledForCommiting(PersistentCollection $coll)
+    {
+        $document = $coll->getOwner();
+        if (!$document || !$this->isInIdentityMap($document)) {
+            return;
+        }
+        $this->hasCollectionScheduledForCommiting[spl_object_hash($document)] = true;
+        // If owner is not scheduled for any action then most probably we're
+        // dealing with document that will have only PersistentCollection updated
```
malarzm: If I'm not mistaken then document is also managed if we are here

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -360,6 +360,12 @@ public function update($document, array $options = array())
                     $this->class->reflFields[$this->class->versionField]->setValue($document, $nextVersion);
                 }
             }
+            
+            // We got here because $document has related PersistentCollections to be 
+            // commited later but if it's not versioned then there's nothing left to do
+            if (empty($update)) {
+                return;
+            }
```
jmikola: I'm curious how collection updates interact with versioning and locking. I suppose since updates are executed by UnitOfWork (via DocumentPersister) before CollectionPersister does its business, that any LockExceptions would prevent scheduled collection updates from happening.

It strikes me as odd that subsequent Collection updates don't use the lock, though (see: [`CollectionPersister::executeQuery()`](https://github.com/doctrine/mongodb-odm/blob/30f42aa01ff4aa289c1090d6cc7fd79d1ac8afe1/lib/Doctrine/ODM/MongoDB/Persisters/CollectionPersister.php#L284)). It's like unrelated to this PR, but I think that warrants further investigation and we should open a ticket to revisit down the line.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -2698,6 +2721,38 @@ public function isCollectionScheduledForUpdate(PersistentCollection $coll)
...
+     * Marks that owner of PersistentCollection has PersistentCollection
+     * that will be somehow updated
+     * 
+     * @param PersistentCollection $coll
+     */
+    private function collectionScheduledForCommiting(PersistentCollection $coll)
+    {
+        $document = $coll->getOwner();
+        if (!$document || !$this->isInIdentityMap($document)) {
+            return;
```
jmikola: Wouldn't you expect this to be an exception? I don't think we should be scheduling a collection if its owner isn't managed.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
jmikola: I assume if a collection is scheduled for update or deletion, the document would already be managed. And I don't imagine a document *and* its PersistentCollection collection would be scheduled for deletion at the same time (I could certainly be wrong there, though).

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
jmikola: Wouldn't you expect this to be an exception? I don't think we should be scheduling a collection if its owner isn't managed (per comments below). We could use `getDocumentState()` instead of `isInIdentityMap()` in that case.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -2698,6 +2721,38 @@ public function isCollectionScheduledForUpdate(PersistentCollection $coll)
...
     }
+    
+    /**
+     * Checks whether Document has any related PersistentCollection
+     * scheduled for update or removal
+     * 
+     * @param object $document
+     * @return boolean
+     */
+    public function hasCollectionScheduledForCommiting($document)
```
jmikola: I would rename this to `hasScheduledCollections()`. The property can also be `$hasScheduledCollections`.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
malarzm: I'll check tomorrow but I think that one of tests was breaking without this, can't recall what was that exactly though

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
malarzm: Naming things is always problematic, no problem at all :+1:

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
malarzm: I've just created #1046 to track it

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
malarzm: If I'm not mistaken than calling ``clear()`` on PersistentCollection and later removing document before commiting UnitOfWork would result in such situation... but I'm not sure if anything else won't be broken then, I'll try to investigate it tomorrow

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
malarzm: Turns out to be more interesting: this ``if`` statement was suppressing possible bug in implementation as UoW wouldn't track changes in deeply embedded documents, working on a fix now

malarzm: Indeed it was a bug, fixed and added test for it

solocommand:
@malarzm Looks like this has resolved #883. Nice!

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
jmikola: :+1:

jmikola:
Merged in https://github.com/doctrine/mongodb-odm/commit/c538f99fbd2d3016e9b0e7601bc63ae0ceb0c453. Thanks!

PR #986, opened 2014-12-01
Allow usage of capped collections via XML driver

### LINKED ISSUE #545: XMLDriver does not support capped collection options
The XMLDriver calls `ClassMetadataInfo::setCollection()` with a string, so it is not possible to pass arguments for a capped collection.

malarzm:
Fixes #545


jwage:
:+1:

PR #987, opened 2014-12-04
Throw meaningful exception while referencing non-object

### LINKED ISSUE #850: Default values on a property with ReferenceOne annotation throws Exception on persisting update, initial persist works fine.
Not sure if that is a bug (it was clearly not intended to work this way) but it throws an unexpected exception when persisting, and maybe failing earlier would be a better approach here.

```php
    /** @ODM\ReferenceOne(targetDocument="Updatemi\LibraryBundle\Document\Woeid")  */
    private $location = ''; // broken
```

This fails to persist (but only when updating a loaded document) with following exception:

```
[InvalidArgumentException]                            
  Cannot create a DBRef, the document is not an object
Exception trace:
 () at /mnt/updatemi/service/vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/DocumentManager.php:689
 Doctrine\ODM\MongoDB\DocumentManager->createDBRef() at /mnt/updatemi/service/vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php:325
 Doctrine\ODM\MongoDB\Persisters\PersistenceBuilder->prepareReferencedDocumentValue() at /mnt/updatemi/service/vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php:217
 Doctrine\ODM\MongoDB\Persisters\PersistenceBuilder->prepareUpdateData() at /mnt/updatemi/service/vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php:338
...

```

It seems like the database persists the default value given and cannot load an object for it when updating. This problem can be solved by just removing the default value, so maybe its just a documentation issue, but maybe this could be catched earlier.

```php
    /** @ODM\ReferenceOne(targetDocument="Updatemi\LibraryBundle\Document\Woeid")  */
    private $location; // works
```

I will try to provide a test case in the evening.

malarzm:
Fixes #850

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
@@ -311,7 +311,13 @@ class PersistenceBuilder
      */
     public function prepareReferencedDocumentValue(array $referenceMapping, $document)
     {
-        return $this->dm->createDBRef($document, $referenceMapping);
+        try {
+            return $this->dm->createDBRef($document, $referenceMapping);
+        } catch (\InvalidArgumentException $e) {
+            throw new \InvalidArgumentException(
+                sprintf('You are trying to reference a non-object in %s field, "%s" given', $referenceMapping['name'], $document)
+            );
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


jwage:
:+1:

PR #990, opened 2014-12-09
[Docs] Added @Version to annotations reference

### LINKED ISSUE #429: Document versioning and locking strategies
I see in the repository that their is a Version annotation. I was wondering what the functionality of this annotation is and why it is not listed in http://docs.doctrine-project.org/projects/doctrine-mongodb-odm/en/latest/reference/annotations-reference.html

### LINKED ISSUE #982: @Version seems to be missing in the docs in annotations reference

malarzm:
Fixes #982 and also resolves #429 which I've just found.

The entry is modeled after ORM's definition. Also it would be good to port http://doctrine-orm.readthedocs.org/en/latest/reference/transactions-and-concurrency.html#locking-support to docs (or at least link if no one has spare time?)

### DIFF
--- a/docs/en/reference/annotations-reference.rst
+++ b/docs/en/reference/annotations-reference.rst
@@ -1141,3 +1141,15 @@ Defines a unique index on the given document.
     private $email;
 
 .. _`DBRef`: http://docs.mongodb.org/manual/reference/database-references/#dbref
+
+@Version
+--------
+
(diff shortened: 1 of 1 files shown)


jmikola:
Manually merged (with some rewording) in https://github.com/doctrine/mongodb-odm/commit/613be17a85b1a64aaa457c857a745edf0c821bb5. Thanks.

PR #992, opened 2014-12-10
Fix embedded documents on inherited referenced documents

### LINKED ISSUE #872: Embedded documents on inherited referenced documents are not loaded correctly
It seems that Doctrine does not load embedded documents correctly behind referenced documents that are part of inheritance tree. Failing test case coming in a minute.

This resembles somehow the issue #562

malarzm:
Fixes #872, closes #873  

/cc @vmattila

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH872/GH872Test.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH872/GH872Test.php
@@ -1,0 +1,140 @@
+<?php
+
+namespace Doctrine\ODM\MongoDB\Tests\Functional\Ticket\GH872;
+
+use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
+
+use Doctrine\Common\Collections\ArrayCollection;
+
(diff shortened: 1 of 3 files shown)


tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH872/GH872Test.php
```diff
@@ -0,0 +1,145 @@
...
+        
+        $address = $location->getAddress();
+        $this->assertNotNull($address);
+
+        $this->assertEquals('Some address', $address->getStreetName());
+    }
+}
+
+/** @ODM\Document */
+class Event
```
jmikola: @malarzm: I assume this file is nested under a `GH872/` directory in order to be consistent with the namespace? And the namespace was added because these class names might conflict with others in the general `Functional\Ticket` namespace?

malarzm: I've merged @vmattila commit into this PR and that's how he's written test and since there was additional namespace and no chance for classname conflict I've left that :)

PR #994, opened 2014-12-11
Set isDirty back to true after call to takeSnapshot()

deedpolloffice:
This commit removed a line of code setting isDirty to true:
https://github.com/doctrine/mongodb-odm/commit/bd573cb7def6ab22df8c465b0d79003d26e28046
I think this line was necessary because takeSnapshot() always sets isDirty to false.  Therefore, if there have been objects added, it needs to be set to its original value.


jwage:
@deedpolloffice can you rebase?

malarzm:
https://github.com/doctrine/mongodb-odm/commit/f3f289866293c8173b7e3429c7cacd3fed3ab0be reintroduced missing `$this->isDirty = true;` so I'm closing the PR

PR #995, opened 2014-12-12
Hydrator updates EmbedOne instead of replacing (if possible)

### LINKED ISSUE #968: Embedded document and Query::HINT_REFRESH
It looks like refresh queries using ``Query::HINT_REFRESH`` are creating a new embedded document instead of updating values of existing one.

For example:

```php
$embed = $document->getEmbed();
// ... do query with HINT_REFRESH
echo $embed === $document->getEmbed() ? "Same" : "Not";
```
Outputs: **Not** :cry:

malarzm:
Fixes #968 

/cc @vkartaviy

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
+++ b/lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
@@ -318,7 +318,12 @@ EOF
             \$embeddedDocument = \$data['%1\$s'];
             \$className = \$this->unitOfWork->getClassNameForAssociation(\$this->class->fieldMappings['%2\$s'], \$embeddedDocument);
             \$embeddedMetadata = \$this->dm->getClassMetadata(\$className);
-            \$return = \$embeddedMetadata->newInstance();
+            \$current = \$this->class->reflFields['%2\$s']->getValue(\$document);
+            if (\$current instanceof \$className) {
+                \$return = \$current;
+            } else {
+                \$return = \$embeddedMetadata->newInstance();
+            }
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
```diff
@@ -319,6 +319,10 @@ private function generateHydratorClass(ClassMetadata $class, $hydratorClassName,
             \$className = \$this->unitOfWork->getClassNameForAssociation(\$this->class->fieldMappings['%2\$s'], \$embeddedDocument);
             \$embeddedMetadata = \$this->dm->getClassMetadata(\$className);
             \$return = \$embeddedMetadata->newInstance();
```
vkartaviy: It seems the new instance creation should be moved to the ```else``` block of ```instanceof``` check statement.

vkartaviy:
Btw - What about ```EmbedMany``` case?

lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
malarzm: Right, always a saving

malarzm:
@vkartaviy I'll see what I can do about ``EmbedMany``

malarzm:
Handling ``EmbedMany`` is far from being trivial and I believe there's no easy way to refresh objects there like proposed with ``EmbedOne``. After discussion with @jwage we think that being consistent is better therefore we're not merging this in

PR #996, opened 2014-12-16
require the yaml driver hard dependency

fabiocarneiro:
I know this was removed long time ago in this PR https://github.com/doctrine/mongodb-odm/commit/c17ee1c30c869763e327de905f10f92b8b37e2c3#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780, but i guess since its a hard dependency of the yaml driver, it should be downloaded no matter if you're not using the yaml driver or not.

If the desired behavior is to download it only if you'll use the yaml driver, then the driver should be moved to another library, and then be optional in your app composer.json

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,6 +13,7 @@
     ],
     "require": {
         "php": ">=5.3.2",
+        "symfony/yaml": "~2.0",
         "symfony/console": "~2.0",
         "doctrine/annotations": "~1.0",
         "doctrine/collections": "~1.1",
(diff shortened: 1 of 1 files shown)


fabiocarneiro:
@malarzm the symfony/yaml is not a dev requirement if you're using the yaml driver in production. Unless you move the driver itself to another package, it will only work if you have it as dependency. The fact that you might not use it at all, doesn't make it less important for the driver. If the driver is part of this package, the package MUST provide its requirements.

malarzm:
Well it suggests that if you need (optional) YAML driver then you should install ``symfony/yaml`` - require-dev is for tests IMO (so development, but related to ODM and not project using it). For instance [Gaufrette's current composer.json](https://github.com/KnpLabs/Gaufrette/blob/master/composer.json) is doing similar thing (require-dev & suggest).

fabiocarneiro:
Well, IMHO, suggest should be used when you have something that could improve the module performance, or provide an aditional feature that would not affect the default behavior, usually based on events, like logging, sending messages, etc... 

The yaml driver is REQUIRED to this driver to work, so it MUST stay in required session.

jwage:
I think we want to keep this in line with the ORM and other packages right now.

PR #997, opened 2014-12-17
Version bugfix

jaymecd:
This PR provided:

- [x] missing xml driver mapping for version/lock
- [x] set initial version if document is saved in upsert mode
- [x] write tests

### DIFF
--- a/doctrine-mongo-mapping.xsd
+++ b/doctrine-mongo-mapping.xsd
@@ -65,6 +65,8 @@
     <xs:attribute name="file" type="xs:boolean" />
     <xs:attribute name="distance" type="xs:boolean" />
     <xs:attribute name="reference" type="xs:boolean" />
+    <xs:attribute name="version" type="xs:boolean" />
+    <xs:attribute name="lock" type="xs:boolean" />
     <!-- index options -->
     <xs:attribute name="background" type="xs:boolean" />
     <xs:attribute name="drop-dups" type="xs:boolean" />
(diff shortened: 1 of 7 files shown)


jaymecd:
@jmikola, could u give some advice how to provide better tests for this fix? tnx

jmikola:
@jaymecd: I believe the only tests for versions and locking is in `LockTest.php`. You appear to be fixing two code paths in this PR (insert and upsert).

Your fix for insert is initializing an empty (zero when cast to an integer) `$version` property. For that, you should be able to trigger the bug by changing the following in `LockTest.php`:

```php
/** @ODM\Document */
class LockInt extends AbstractVersionBase
{
    /** @ODM\Version @ODM\Int */
    public $version = 1;
}
```

The upsert path is ultimatey decided in `UnitOfWork::persistNew()` when we examine the persisted document. If it contains a value in its `$id` field, we'll upsert it. Since all of the properties in the `LockTest.php` models are public, it should be easy to add some new tests there which set `$id` to a new MongoId before persisting (i.e. generating the ObjectId yourself instead of letting ODM do it).

jaymecd:
@jmikola done + updated xml driver mapping test for version

jmikola:
Merged in https://github.com/doctrine/mongodb-odm/commit/b37c4c3dd913d707c743e78e15c39e5dbba2f193.

In the future, feel free to leave the commits split if it makes logical sense. I'm not sure if you rebased this, but the bug fix for initializing on upsert and XML metadata support could have been separated.

In any event, thanks!

PR #948, opened 2014-09-22
Allow DocumentRepository::matching() to accept Criteria w/o "where"

MidnightDesign:

PR #998, opened 2014-12-22
update DocumentRepository to use where expression only when not null

fabiocarneiro:
This allows the document repository to be used directly with the selectable adapter when you dont set any where expression.


jmikola:
Thanks, but it looks like #948 (recently merged) already addressed this.

PR #999, opened 2014-12-23
Fix changeset handling for new documents

alcaeus:
This fixes an issue that occurs when modifying a newly persisted document in an onFlush() event handler.
The issue causes the insertion to be treated as an update even though only an insert needs to be done. Thus, postUpdate listeners are called which is not correct.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
+++ b/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
@@ -803,7 +803,9 @@ class UnitOfWork implements PropertyChangedListener
                     : $changeSet;
 
                 $this->originalDocumentData[$oid] = $actualData;
-                $this->documentUpdates[$oid] = $document;
+                if ( ! isset($this->documentInsertions[$oid])) {
+                    $this->documentUpdates[$oid] = $document;
+                }
             }
         }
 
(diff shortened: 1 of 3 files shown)


alcaeus:
I fixed the failing tests - however, considering they were written to test exactly the kind of behavior this PR aims at fixing, I'm not sure what the consequences would be. I'd really appreciate some feedback on this.

malarzm:
The test origin is https://github.com/doctrine/mongodb-odm/pull/246 and I believe it was about detecting change in array, not sure if it was intended to trigger post update then or just used the fact that it does instead of checking change set... @jwage @jmikola @tecbot ?

alcaeus:
I've removed the WIP label since there was no feedback on the old tests.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -803,7 +803,9 @@ private function computeOrRecomputeChangeSet(ClassMetadata $class, $document, $r
                     : $changeSet;
 
                 $this->originalDocumentData[$oid] = $actualData;
-                $this->documentUpdates[$oid] = $document;
+                if ( ! isset($this->documentInsertions[$oid])) {
```
malarzm: Should we check ``$this->documentUpserts[$oid]`` as well here?

jwage:
This is fixed in #782

alcaeus:
@jwage Is #782 going to be merged anytime soon? It has been open for a while, and while I appreciate a good refactoring, sometimes I'd rather see bugs fixed, especially if they are hard to work around in software ;)

jwage:
@alcaeus I understand, but I think patching the one off bugs here and there will end up making things much worse and harder or impossible to fix in the future without BC breaks. I finished the last known issues last night and got the tests passing. I need others to test the patch. I am going to be testing it against my application.

alcaeus:
I agree. I'll give it a test run next week. Could you add the unit test contained in this PR to yours? It would check against inserts triggering an update event.

jwage:
@alcaeus Yes, I copied the test case over to my branch.

alcaeus:
:+1:

PR #1007, opened 2015-01-06
fixed missing inherited association mappings

data219:


gruberro:
This PR duplicates the already existing PR #958 related to the same issue.

PR #1010, opened 2015-01-12
Fix PHP Notice due to undefined _attributes property

vkartaviy:
PHP Notice:  Undefined property: Doctrine\ODM\MongoDB\Configuration::$_attributes in ~/vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/Configuration.php on line 319


malarzm:
Merged in https://github.com/doctrine/mongodb-odm/commit/700b6ddabb6d1dc682946deb0770027025b2dc80, thanks @vkartaviy!

PR #1014, opened 2015-01-22
Add improved test for issue #963  

rubenrua:
See https://github.com/doctrine/mongodb-odm/pull/981 conversation:

>  I updated my test with a bi-directional reference and it's broken again.


malarzm:
Attached test case is passing on current dev-master so I'm closing the issue.

rubenrua:
Good!!!  Thx.

PR #1015, opened 2015-01-22
Fix test for PHP 5.4

malarzm:


jmikola:
Thanks for catching this.

PR #1016, opened 2015-01-22
Implement "also load methods" in Yaml and XML drivers.

scottaubrey:
The Yaml and XML drivers currently do not implements the also load methods mapping. I personally only wanted the feature for the Yaml driver, but I've created a pull request implementing it for both Drivers, with updates to XML schema too. I did both at the same time for simplicity in adding the test below.

I've also extended the current AbstractDriverTest unit test and supporting fixtures to add in the new elements, and test they get set on the resulting ClassMetaData. I was unsure if this was the correct place to be testing this.

As a side note also, I think the current TestDocuments.User.dcm.xml fixture does not validate as correct XML in the tool I used, I think because the embed* and reference* elements are out of order compared to the schema. I didn't want to "fix" it, because a) I have no idea what I'm talking about with XML and XML schemas, and b) incase there is a reason for it being this way.

### DIFF
--- a/doctrine-mongo-mapping.xsd
+++ b/doctrine-mongo-mapping.xsd
@@ -40,6 +40,7 @@
       <xs:element name="discriminator-field" type="odm:discriminator-field" minOccurs="0"/>
       <xs:element name="discriminator-map" type="odm:discriminator-map" minOccurs="0"/>
       <xs:element name="lifecycle-callbacks" type="odm:lifecycle-callbacks" minOccurs="0"/>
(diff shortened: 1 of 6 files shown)


scottaubrey:
the travis failure is unrelated to my changes :)

malarzm:
Travis is re-run and says it's ok now ;) I've marked this for next release (looks good to me, just want to get back soon to schema issue you've pointed out)

scottaubrey:
fantastic, thanks

jwage:
:+1:

PR #1022, opened 2015-01-30
Allow to use uptodate versions of Doctrine Common

stof:
Doctrine Common 2.5 is available, and the ODM should allow using it. Otherwise it makes it impossible to install it alongside the ORM 2.5 (which requires new features of Common 2.5)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
         "symfony/console": "~2.0",
         "doctrine/annotations": "~1.0",
         "doctrine/collections": "~1.1",
-        "doctrine/common": "2.4.*",
+        "doctrine/common": "~2.4",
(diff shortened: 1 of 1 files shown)


stof:
@jmikola looks like the Doctrine library is not compatible with ext-mongo 1.6.0, which breaks the build

jmikola:
> looks like the Doctrine library is not compatible with ext-mongo 1.6.0

Should be fixed with 1.1.7 and 1.0.10: https://github.com/doctrine/mongodb/releases

jmikola:
@stof: Just realized doctrine/mongodb needs its doctrine/common deps relaxed. Going to tag 1.1.8 and 1.0.11 to fix that.

Issue #1017, opened 2015-01-23
UnitOfWork problem due to spl_object_hash collision


AronSzigetvari:
I have run into a very strange problem at flushing, getting the following warnings and exceptions:
 * array_map(): An error occurred while invoking the map callback in  doctrine\mongodb-odm\lib\Doctrine\ODM\MongoDB\Persisters\CollectionPersister.php on line 219
 * Cannot create a DBRef without an identifier. `UnitOfWork::getDocumentIdentifier()` did not return an identifier for class

I have tracked down the problem using bisect to the commit when `DocumentManager::createDBRef()` started to throw an exception if `$this->unitOfWork::getDocumentIdentifier()` doesn't return with an identifier. The problem was there before that, but it was silent.

Finally I could find the root cause of this. UnitOfWork uses an array `documentIdentifiers` to store document ID's indexed by the result of function `spl_object_hash`. There is another array `documentStates`, also indexed by the same values. However the *`spl_object_hash` values are not completely unique*: once an object is not referenced anywhere and gets garbage-collected, the same hash value is reused, this is clearly stated in the documentation of the function (see  http://php.net/manual/en/function.spl-object-hash.php)

Probably it is less of a problem with full documents, but EmbeddedDocuments don't have an ID and therefore not stored in `identityMap`, but their managed state is still stored. See the following scenario (an actual one that happened in my case):
 1. An EmbeddedDocument is persisted 
 2. The EmbeddedDocument is replaced by another instance by a setter in its parent document, 
 3. The old instance is destroyed by garbage collection, but the value `STATE_MANAGED` is still there in `documentStates`.
 4. A new document is persisted, unfortunately its `spl_object_hash` value is the same as the object hash of the already deleted EmbeddedDocument. `UnitOfWork` finds the status STATE_MANAGED for it and thinks that the new document is already managed, its identifier is not written into `documentIdentifiers` causing inconsistency. 
 5. `DocumentManager::getDocumentIdentifier()` is called, unitOfWork doesn't return a document identifier for it, so an Exception is thrown
 6. If documentIdentifier is referenced in a referenceMany association, the map callback fails because of the exception.

The solution I can think of is preventing garbage collection of the documents whose state is stored. 
[...]

PR #1023, opened 2015-02-02
Avoiding spl_object_hash collision problem Gh 1017

AronSzigetvari:
This PR is about https://github.com/doctrine/mongodb-odm/issues/1017 
The problem it is intended to solve is described in the issue above. 
I am explaining it in a few words here too. UnitOfWork uses the spl_object_hash() values for documents as an internal identifier and stores document status and ID values in arrays indexed with these hash values.
Unfortunately, if EmbeddedDocuments are removed by garbage collection, their hash value is reused, so these internal metadata will be mixed up when a new document with matching hash is persisted.

This PR adds a new array to store new embedded documents until the UnitOfWork is cleared, so garbage collection doesn't destroy them, and the hash value won't be reused.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
+++ b/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
@@ -128,7 +128,7 @@ class UnitOfWork implements PropertyChangedListener
      * object ids by PHP for new objects after garbage collection
      * Keys are object ids (spl_object_hash).
      *
-     * @var type
+     * @var array
      */
     private $embeddedDocumentsKnown = array();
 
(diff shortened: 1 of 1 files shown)


aequasi:
:+1:

jwage:
Can you add a test for the problem you are trying to fix? I would like to see if it is fixed with #782. I don't feel like this is the right fix for the problem.

AronSzigetvari:
It is not trivial because it depends on when PHP wants to do a garbage collection and when does it reuse the same hash. But it has happened in our project quite similarly all the time so there is some chance that it is reproducible in a test.
Personally I didn't like my solution that much, because it locks in otherwise disposable the object in the memory just to avoid being their hash reused, but I didn't find other way. 
I will try to create that test and let's see that #782 solves it.

malarzm:
@AronSzigetvari could you test #782 once again? I've rebased it to latest dev-master

malarzm:
Moving discussion to #782

PR #691, opened 2013-10-17
Sharding support implemented

lotar:

PR #1025, opened 2015-02-04
Yet another sharding support implementation

clslrns:
This one is inspired by #691.

Brief summary:
* `shardKey` property was added to `ClassMetadataInfo`, mapping support were added to all drivers. It's a hash with `keys` and `options`. `keys` was originally called `fields`, but I decided to conform to the indexes definition.
* `shardKey` is class-based, since we need to keep the order of the keys to ensure compound index.
* `ensureSharding` method were added to `SchemaManager`. It's similar to `ensureIndex`. For non-empty collections it will run `ensureIndex` and retry `shardCollection` again.
* `DocumentPersister` uses `shardKey` to generate query part for `upsert`, `update`, `remove`, `findOne`.

I tried to cover all the new code with both unit and functional tests. Tests were run against sharded clusters of MongoDB 2.2.3, 2.4.9 and 2.6.5. Functional tests were tagged to avoid running against non-sharded setup.

Now I'm testing it on production. I'd like to receive some feedback from you guys to improve or change something.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -277,25 +277,8 @@ class DocumentPersister
         }
 
         foreach ($this->queuedUpserts as $oid => $document) {
-            $data = $this->pb->prepareUpsertData($document);
-
-            // Set the initial version for each upsert
-            if ($this->class->isVersioned) {
-                $versionMapping = $this->class->fieldMappings[$this->class->versionField];
-                if ($versionMapping['type'] === 'int') {
-                    $nextVersion = max(1, (int) $this->class->reflFields[$this->class->versionField]->getValue($document));
-                    $this->class->reflFields[$this->class->versionField]->setValue($document, $nextVersion);
-                } elseif ($versionMapping['type'] === 'date') {
-                    $nextVersionDateTime = new \DateTime();
-                    $nextVersion = new \MongoDate($nextVersionDateTime->getTimestamp());
-                    $this->class->reflFields[$this->class->versionField]->setValue($document, $nextVersionDateTime);
-                }
-                $data['$set'][$versionMapping['name']] = $nextVersion;
-            }
-
             try {
-                $shardKeyQueryPart = $this->getShardKeyQuery($document, $options);
-                $this->executeUpsert($data, $options, $shardKeyQueryPart);
+                $this->executeUpsert($document, $options);
                 $this->handleCollections($document, $options);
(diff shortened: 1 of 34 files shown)


clslrns:
It seems, tests failed because of long MongoDB initialization. Would be great if somebody re-ran them.

jmikola:
@clslrns: I will attempt to get to this before the next beta tag.

clslrns:
Btw, wouldn't it be better to squash all this commits?

notrix:
:+1:

alcaeus:
@clslrns sorry for the long time without updates. Could you please rebase the branch to see if it still works against the current master?

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -422,13 +429,12 @@ public function delete($document, array $options = array())
     /**
      * Refreshes a managed document.
      *
-     * @param array $id The identifier of the document.
      * @param object $document The document to refresh.
      */
-    public function refresh($id, $document)
+    public function refresh($document)
```
alcaeus: I'm afraid this is a BC break. As useless as the ID parameter has become, it has to stay. Feel free to deprecate it.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -348,14 +352,19 @@ private function executeUpsert(array $data, array $options)
...
-        $id = $this->uow->getDocumentIdentifier($document);
         $update = $this->pb->prepareUpdateData($document);
 
-        if ( ! empty($update)) {
+        $query = $this->getQueryForDocument($document);
 
-            $id = $this->class->getDatabaseIdentifierValue($id);
-            $query = array('_id' => $id);
+        foreach (array_keys($query) as $field) {
+            unset($update['$set'][$field]);
```
alcaeus: From the looks of it, this could pose issues where a field with an ```@increment``` mapping is part of the shard key. The field would not be part of a ```$set``` operator but an ```$inc``` operator, meaning that we'd send an invalid query to the database.
In this case, however, I think we should set limits on what kind of mapping can be used for fields that are part of the shard key. Having an ```@increment``` mapping makes no sense since the field should not be updated anyways, so I suggest guarding against this on the ClassMetadata level.

clslrns:
@alcaeus, thanks for the feedback.
I have rebased it and fixed two issues you pointed to. I hope, in a few days I'll test rebased version against the app.

darceee2:
:+1:

alcaeus:
Superseded by #1385.

PR #1026, opened 2015-02-10
Add gitignore ide and File-based project format files

guidroid:
Added some files that should not be in the repository.


jmikola:
Thanks for the suggestion, but I feel these exclusions are better left to developers' global gitignore files -- so they can customize them for their own OS and development environment(s).

I realize `.idea/` is in our current gitignore, but I've been meaning to remove that (it snuck in through a PR some time ago).

Issue #901, opened 2014-06-05
@Date fields do not support milliseconds


Bilge:
Mongo's `ISODate` supports milliseconds but PHP's `DateTime` does not. Since @Date fields marshal dates via `DateTime` structures, millisecond data cannot be transferred either in or out of Mongo using these fields.

malarzm:
#1063 was merged implementing feature.

PR #1028, opened 2015-02-12
Add millisecond support for Date mapping type

gerryvdm:
Both MongoDB and the PHP DateTime class support milliseconds, but currently this information is lost in the date mapping type.

This patch will transfer the millisecond information between PHP and MongoDB.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Types/DateType.php
+++ b/lib/Doctrine/ODM/MongoDB/Types/DateType.php
@@ -58,6 +58,7 @@ class DateType extends Type
         }
         if ($value instanceof \MongoDate) {
             $date = \DateTime::createFromFormat('U.u', sprintf('%d.%06d', $value->sec, $value->usec));
+            $date->setTimezone(new \DateTimeZone(date_default_timezone_get()));
         } elseif (is_numeric($value)) {
             $date = new \DateTime();
             $date->setTimestamp($value);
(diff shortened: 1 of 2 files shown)


malarzm:
@gerryvdm is Travis failure related to your changes? Also I think adding a test for this PR would be great :)

gerryvdm:
@malarzm Not sure, the error seems a bit exotic. Will add a test!

gerryvdm:
This should also fix #901.

malarzm:
It's a bit blind guess but there's field mapped as date and as not saved in https://github.com/doctrine/mongodb-odm/blob/master/tests/Documents/File.php#L29 which is referenced by https://github.com/doctrine/mongodb-odm/blob/master/tests/Documents/Profile.php#L21 which should be the class with hydration problem - maybe that's it?

gerryvdm:
I think the problem is related to timezones. Will need to have a further look.

gerryvdm:
Well that was harder than I imagined, but I think the problem is solved. Needed some workaround for allowing pre-unix era timestamps...

malarzm:
:+1:

gerryvdm:
ping @jmikola. Any chance to review?

lavoiesl:
Fixes #1061

lavoiesl:
I hadn’t seen your PR and I created one as well: #1063.

The differences: 
- The closures are much smaller through the use of a public static function.
- Microseconds specified with a float (either float or string) are also supported.
- Microseconds specified with a DateTime string (i.e.: `'2000-01-01 00:00:00.123'`) are also supported.
- More tests for corner cases.

In the end, we could merge those two PRs, but I feel mine is more complete.

jwage:
:+1:

PR #1031, opened 2015-02-22
Allow symfony 3.0 components

nicolas-grekas:
Tests should tell if any deprecated interfaces of Symfony are used. If not, then the bundle is defacto compatible with 3.0


nicolas-grekas:
Just added minimum stability: dev, allowing you to test with upcoming versions of your deps

jwage:
:+1:

PR #1032, opened 2015-02-25
[Docs] Fix @EmbedOne annotation reference

malarzm:
As far as I remember ``strategy`` has no effect for ``@EmbedOne`` annotation (and it doesn't exist in corresponding Annotation definition) and example code was copied from ``@EmbedMany`` while its explaination was refering to totally different code


jwage:
:+1:

PR #1036, opened 2015-02-27
Fixed index settings for YAML

### LINKED ISSUE #1033: Issues with sparse unique indexes using YAML mapping
Hi ,
First of all thanks for this awesome ODM !
I'm having a trouble when creating unique indexes using yml mapping files, for instance if I want a unique index on "sid" with sparse option active I only get the unique index without sparse. My "fields" section look like this
```
fields:
      id:
        type : id
        id:  true
        strategy: INCREMENT
      sid:
        type: string
        unique: true
        sparse: true
```
After updating the schema and if I check the indexes the "sparse" option is never added :
```
> db.getCollection('Messaging\\Message').getIndexes();
[
	{
		"v" : 1,
		"key" : {
			"_id" : 1
		},
		"name" : "_id_",
		"ns" : "mydb.Messaging\\Message"
	},
	{
		"v" : 1,
		"unique" : true,
		"key" : {
			"sid" : 1
		},
		"name" : "sid_1",
		"ns" : "mydb.Messaging\\Message",
		"w" : 1
	}
]
```
This is not a mongodb limitation because I can create unique sparse indexes with command line : 
```
> db.getCollection('Messaging\\Message').ensureIndex({'sid':1},{sparse:1,unique:1});
> db.getCollection('Messaging\\Message').getIndexes();
[
	{
		"v" : 1,
		"key" : {
			"_id" : 1
		},
		"name" : "_id_",
		"ns" : "mydb.Messaging\\Message"
	},
	{
		"v" : 1,
		"key" : {
			"sid" : 1
		},
		"name" : "sid_1",
		"ns" : "mydb.Messaging\\Message",
		"sparse" : true,
		"w" : 1
	}
]
```
Also adding the keyword "index:true" to the "sid" field will just create a nonsense index with neither "unique" nor "sparse" options active...

I'm not sure about the other mapping types, but for yml this is not working.
Thanks for the help !

malarzm:
Fixes #1033

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Mapping/Driver/YamlDriver.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/Driver/YamlDriver.php
@@ -183,11 +183,13 @@ class YamlDriver extends FileDriver
 
         if (isset($mapping['index'])) {
             $options = is_array($mapping['index']) ? $mapping['index'] : array();
-        } elseif (isset($mapping['unique'])) {
-            $options = is_array($mapping['unique']) ? $mapping['unique'] : array();
+        } 
+        if (isset($mapping['unique'])) {
+            $options += is_array($mapping['unique']) ? $mapping['unique'] : array();
             $options['unique'] = true;
-        } elseif (isset($mapping['sparse'])) {
-            $options = is_array($mapping['sparse']) ? $mapping['sparse'] : array();
+        }
+        if (isset($mapping['sparse'])) {
+            $options += is_array($mapping['sparse']) ? $mapping['sparse'] : array();
             $options['sparse'] = true;
         }
 
--- a/tests/Doctrine/ODM/MongoDB/Tests/Mapping/Driver/AbstractDriverTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Mapping/Driver/AbstractDriverTest.php
@@ -56,8 +56,17 @@ abstract class AbstractDriverTest extends \PHPUnit_Framework_TestCase
             'isCascadeRemove' => false,
             'isInverseSide' => false,
             'isOwningSide' => true,
-            'nullable' => false
+            'nullable' => false,
+            'unique' => true,
+            'sparse' => true
         ), $classMetadata->fieldMappings['username']);
+        
+        $this->assertEquals(array(
+            array(
+                'keys' => array('username' => 1),
+                'options' => array('unique' => true, 'sparse' => true)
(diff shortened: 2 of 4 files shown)


malarzm:
cc @AliHichem

AliHichem:
Awesome !

AliHichem:
When do you think I could get this update by composer ? mine is configured as recomended in the doc "1.0.*@dev".

malarzm:
As soon as @jmikola merges this (if PR is fine for him in first place)

AliHichem:
Excellent ! Thank you very much.

lib/Doctrine/ODM/MongoDB/Mapping/Driver/YamlDriver.php
```diff
@@ -183,11 +183,13 @@ private function addFieldMapping(ClassMetadataInfo $class, $mapping)
 
         if (isset($mapping['index'])) {
             $options = is_array($mapping['index']) ? $mapping['index'] : array();
-        } elseif (isset($mapping['unique'])) {
-            $options = is_array($mapping['unique']) ? $mapping['unique'] : array();
+        } 
+        if (isset($mapping['unique'])) {
```
jmikola: The logic for this and `sparse` below still seems wrong. We're only checking if it's set, but ignoring that it might be false. Contrast to what we do for such boolean options in [XmlDriver](https://github.com/doctrine/mongodb-odm/blob/7fcf703966986bd74f00edb7157452015792a659/lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php#L338).

We appear to allow index specifications under `unique` and `sparse` for legacy reasons (looking at the [YAML docs](http://docs.doctrine-project.org/projects/doctrine-mongodb-odm/en/latest/reference/yml-mapping.html)), so we can't eliminate this entirely; however, I think we should throw a MappingException if more than one of `index`, `unique`, and `sparse` have an array. I don't think it makes sense to allow a single index specifications (e.g. not simply boolean values) to be spread out over multiple fields.

@malarzm: I'll open a PR against your branch tomorrow with a proposed fix.

jmikola:
@malarzm: Please take a look at https://github.com/malarzm/mongodb-odm/pull/2. I didn't have time to add manual tests for it (like I did in my last XmlDriver change), but I believe it implements what we're looking for.

jmikola:
Merged in 30f42aa01ff4aa289c1090d6cc7fd79d1ac8afe1.

PR #1037, opened 2015-02-27
Lock document only if found

notrix:
I have issue that some documents expire in mongodb and references return null.
Then $repository->find($invalidId, LockMode::OPTIMISTIC); throws exceptions when UnitOfWork tries to lock NULL.


jwage:
:+1:

jwage:
Thanks for the fix!

men232:
What about setLockMode in Doctrine\ODM\MongoDB\Query\Query like Doctrine\ORM\Query?

malarzm:
@men232 feel free to open new issue or submit a PR implementing that :)

Issue #432, opened 2012-10-13
Support simple reference with discriminator maps


Bilge:
Although not documented, currently the following attributes are mutually exclusive.

* discriminatorMap
* simple=true

Trying to use them together causes errors.

A typical discriminatorMap field must therefore be similar to the following.
```
"foo" : {
    "$ref" : "Bar",
    "$id" : ObjectId("123"),
    "$db" : "MyDb",
    "type" : 'bar'
}
```

However, Doctrine only needs to know the class and ID. Everything else it can infer from the class's metadata. Therefore there is no reason not to support simple references on discriminator fields.

```
"foo" : {
    "$id" : ObjectId("123"),
    "type" : 'bar'
}
```

malarzm:
We'll use #802 to to track stop storing $db

PR #1038, opened 2015-03-03
Discriminator map support for simple references

clslrns:
I'm a big fan of simple references since they allow to map the existing documents without any migration and to stop using ODM whenever needed. The fact that the Doctrine does not allow to use the discriminator map combined with simple references disappointed me. So I tried my best to implement that support and provide consistent behaviour for it.

This PR allows you to use references like so:
```php
/**
 * @ODM\Document
 * @ODM\DiscriminatorField(name="type")
 * @ODM\DiscriminatorMap({"advanced"="AdvancedQuestion", "basic"="BasicQuestion"})
 */
class BasicQuestion
{
    /** @ODM\Id */
    public $id;
    /** @ODM\String */
    public $text;
    /** @ODM\ReferenceOne(targetDocument="Answer", simple=true) */
    public $answer;
}

/** @ODM\Document */
class AdvancedQuestion extends BasicQuestion
{}

/** @ODM\Document */
class Quiz
{
    /** @ODM\Id */
    public $id;
    /** @ODM\ReferenceMany(targetDocument="BasicQuestion", simple=true) */
    public $questions;
}
```

During the initialization, the `questions` collection would be populated with:
* Uninitialized `BasicQuestion` proxies, if there were no loaded documents in the identity map. After the proxy initialization you will get all the data loaded to `BasicQuestion` proxies, but unfortunately not the object of resolved type.
* Documents of the proper type or its proxies, if they were loaded to the identity map.

You could always use priming to get the documents of the proper type.

Also, this PR is changing the way of collections initialization. Now collections are being filled with uninitialized proxies in all cases, except sorted references. It allows to iterate over the collection and use referenced document's IDs without issuing additional queries.

I would be grateful for any feedback.

Related issue: #432, PR with the test and the similar proposal: #645.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -309,7 +309,7 @@ class DocumentPersister
             unset($data['$set']);
         }
 
-        /* If there are no modifiers remaining, we're upserting a document with 
+        /* If there are no modifiers remaining, we're upserting a document with
          * an identifier as its only field. Since a document with the identifier
          * may already exist, the desired behavior is "insert if not exists" and
          * NOOP otherwise. MongoDB 2.6+ does not allow empty modifiers, so $set
@@ -648,7 +648,6 @@ class DocumentPersister
 
     private function loadReferenceManyCollectionOwningSide(PersistentCollection $collection)
     {
-        $hints = $collection->getHints();
         $mapping = $collection->getMapping();
         $groupedIds = array();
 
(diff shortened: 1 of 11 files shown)


jwage:
1.) I think we should throw an exception if you combine simple => true and a discriminator. That type of relationship is inherently not simple and storing the reference as an array is critical to be able to store the discriminator value so we can construct the proper class during hydration.

2.) While I see the benefit of being able to iterate over the proxies to get the identifier without loading them, I think it is a big performance hit to create the proxies eagerly when they won't be used in most cases.

3.) Your method to find the class here https://github.com/doctrine/mongodb-odm/pull/1038/files#diff-ae2c9f4b58c612dce08a0f60ff683aecR1805 while it works I don't really like the fix. I think we should keep things explicit via mapping configuration.

alcaeus:
@clslrns Just a quick note about this:
> I'm a big fan of simple references since they allow to map the existing documents without any migration and to stop using ODM whenever needed.

References in MongoDB are always what Doctrine would call non-simple references. Using DBRef in MongoDB is not a problem should you ever decide to stop using the ODM.

@jwage Fully agree on all three of your points.

clslrns:
@jwage, @alcaeus, thanks for the review.

A small follow-up proposal about 2. Does it worth to implement optional proxies creation based on `@ODM\ReferenceMany`' parameter? Or even lazy proxies creation? The inability to use loaded data without issuing an additional query looks strange to me.

clslrns:
@alcaeus 
> References in MongoDB are always what Doctrine would call non-simple references. Using DBRef in MongoDB is not a problem should you ever decide to stop using the ODM.

But what about [manual refs](http://docs.mongodb.org/manual/reference/database-references/#document-references)? You should migrate them to `DBRefs` in case you are using ODM. That's the problem I'm trying to solve.

@jwage
> 1.) I think we should throw an exception if you combine simple => true and a discriminator. That type of relationship is inherently not simple and storing the reference as an array is critical to be able to store the discriminator value so we can construct the proper class during hydration.

I still think it's better to be able to use abstract proxies. There is no way to injure yourself using when your classes are SOLID enough.

> 3.) ... I think we should keep things explicit via mapping configuration.

I agree that my solution is not quite elegant, but the discrimination map is specified via mapping configuration. And I found that `simple=true` + discrimination map is explicit enough to the end-users.

alcaeus:
@clslrns A manual ref is not a reference per se - MongoDB knows nothing about it except that you're storing an ```ObjectId```. With ```DBRef``` objects MongoDB itself knows what collection the reference is pointing at.

Either way, there's no way to store the discriminator value in a simple ref, which is (and I'm just assuming here) why discriminators were not implemented for simple refs in the first place. If you want discriminators, use ```DBRef```. If you stop using the ODM, everything will work normally.

clslrns:
@alcaeus Yep, for sure. But if you have a huge DB with manual refs, it's impossible to start using simple references + discrimination map. You should migrate all the docs first. And after dropping ODM in that case there will be the refs overhead.

PR #1047, opened 2015-03-12
[Docs] Amend priming references

malarzm:
I wasn't entirely sure if I'm allowed to prime such field after reading docs and was wondering how this will be fitted in one query ;)


jmikola:
The code example on this page is based on a ReferenceMany example. Is the question about how it works with discriminated fields where various concrete classes could live in different collections?

malarzm:
Yes, that's exactly the thing I was wondering about. Example was based on ReferenceMany with targetDocument specified and later note was mentioning single collection inheritance. Thought it's worth adding that it does work with classes living in different collections

jmikola:
Merged in https://github.com/doctrine/mongodb-odm/commit/ab0137a30f3ffd440cd32788e5b0126d838c58cf with some revision. Thanks!

Issue #1048, opened 2015-03-12
ReferenceMany with repositoryMethod and eagerCursor causes fatal error


alcaeus:
When using referenceMany in combination with repositoryMethod, returning an eagerCursor instead of a regular cursor results in a fatal error in ```Doctrine\ODM\MongoDB\Persisters \DocumentPersister::createReferenceManyWithRepositoryMethodCursor```. This is because the eagerCursor does not support ```sort()```, ```limit()``` or ```skip()```. As such, repositoryMethods may not return eager cursors. There are two possibilities:

1. have the DocumentPersister unwrap the original cursor from the EagerCursor object and apply sort, limit and skip to that.
2. Have ```Doctrine\ODM\MongoDB\EagerCursor``` support sorting out of the box by wrapping the original methods from ```Doctrine\ODM\MongoDB\Cursor```. Note: this may depend on doctrine/mongodb#208.

I can provide a PR for both - just need to know what the better course of action is.

malarzm:
Option number 1 seems to be good quickfix for the case.

As for 2nd option I'd additionally consider some common interface for these two to have something to rely on :)

alcaeus:
Sounds good. I have a working draft prepared, I'll clean it up, add tests and push it in the coming days.

malarzm:
:+1:

malarzm:
#1049 has been merged

PR #1049, opened 2015-03-17
Unwrap EagerCursor from repositoryMethod

alcaeus:
This PR is a fix for #1048. If a repositoryMethod returns an ```EagerCursor```, the ```DocumentPersister``` unwraps the nested ```Cursor``` instance and applies the ```sort()```, ```skip()``` and ```limit()``` to that cursor.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -23,6 +23,7 @@ use Doctrine\Common\EventManager;
 use Doctrine\MongoDB\Cursor as BaseCursor;
 use Doctrine\ODM\MongoDB\Cursor;
 use Doctrine\ODM\MongoDB\DocumentManager;
+use Doctrine\ODM\MongoDB\EagerCursor;
 use Doctrine\ODM\MongoDB\Hydrator\HydratorFactory;
 use Doctrine\ODM\MongoDB\LockException;
 use Doctrine\ODM\MongoDB\LockMode;
@@ -782,7 +783,7 @@ class DocumentPersister
     /**
      * @param PersistentCollection $collection
      *
-     * @return Cursor
+     * @return Cursor|EagerCursor
      */
     public function createReferenceManyWithRepositoryMethodCursor(PersistentCollection $collection)
     {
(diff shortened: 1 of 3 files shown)


lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -791,20 +792,25 @@ public function createReferenceManyWithRepositoryMethodCursor(PersistentCollecti
         $cursor = $this->dm->getRepository($mapping['targetDocument'])
             ->$mapping['repositoryMethod']($collection->getOwner());
 
+        $wrappedCursor = $cursor;
+        if ($cursor instanceof EagerCursor) {
+            $wrappedCursor = $cursor->getCursor();
```
malarzm: I think I'd rather have ``$cursor = $cursor->getCursor();`` as at the end we're returning it and docblock says we should return ``Cursor`` not ``EagerCursor``

alcaeus: Actually, that's against the entire point of ```EagerCursor```. If we return the wrapped cursor, it's as if the repositoryMethod had not returned an ```EagerCursor``` in the first place.
I'd suggest changing the return annotation to ```* @return Cursor|EagerCursor``` and returning the cursor as it was returned by the ```repositoryMethod```. That would make it consistent with the ```getIterator()``` method in the ```Query``` class.

malarzm: Ok, I guess that it will be up to developer to know what he's returning from ``repositoryMethod`` then (if one would want to use this method since its public but I guess it would be rare :) ) so I'm ok with that. What do you think about throwing exception if EagerCursor was already initialized (for any reason)? If I understand correctly in such case it's too late to alter wrapped cursor?

alcaeus: From http://php.net/manual/en/mongocursor.sort.php:
> Throws MongoCursorException if this cursor has started iterating.
Since the ```Cursor``` passes the ```sort()``` call down to the ```MongoCursor``` object, I don't think we need to check for an opened cursor - ```MongoCursor``` will do this for us.

malarzm: Cool, sorry for not checking that, got million things on my mind at the moment

alcaeus: I can imagine...no worries!

malarzm:
Merged in https://github.com/doctrine/mongodb-odm/commit/a95e87158d9e224abf2c587f02d62a924a8af90f, thanks @alcaeus!

PR #1052, opened 2015-03-20
DocumentRepository->findBy doesn't work when array is passed

malarzm:
Failing test case for https://github.com/symfony/symfony/issues/13961


malarzm:
On a second thought current flow makes perfect sense as Document can contain array and user can search for document by it, closing this then

PR #1053, opened 2015-03-23
Fixed wrong MappedSuperclass Annotation on odm:generate:documents

ElectroLutz:
I fixed a typo/error in annotation generation for MappedSuperclass. It now matches the one used in the tests - so I guess this one is right.


malarzm:
Merged in https://github.com/doctrine/mongodb-odm/commit/803251505666c06150994eea065ab37329f2c2ff, thanks @ElectroLutz!

Issue #1060, opened 2015-03-28
DateTest::testOldDate() fails on 32-bit platforms


jmikola:
@jwage mentioned that https://github.com/doctrine/mongodb-odm/commit/c36c0606fb0a2f933e843cf68963f3cf4e8f207f failed for him on a 32-bit test environment:

```
1) Doctrine\ODM\MongoDB\Tests\Functional\DateTest::testOldDate
InvalidArgumentException: Could not convert "1900-01-01" to a date value
```

This is due to the the test date exceeding the bounds for a negative 32-bit int. From `strtotime()`'s documentation:

> The valid range of a timestamp is typically from Fri, 13 Dec 1901 20:45:54 UTC to Tue, 19 Jan 2038 03:14:07 UTC. (These are the dates that correspond to the minimum and maximum values for a 32-bit signed integer.) 

Interestingly, DateTime doesn't seem to have this problem:

```
vagrant@precise32:/vagrant/data/mongodb-odm$ php -r "echo (new \DateTime('1900-01-01'))->format('c');"
1900-01-01T00:00:00+00

vagrant@precise32:/vagrant/data/mongodb-odm$ php -r "var_dump((new \DateTime('1900-01-01'))->format('U'));"
string(11) "-2208988800"

vagrant@precise32:/vagrant/data/mongodb-odm$ php -r "var_export(strtotime('1900-01-01'));"
false
```

This is mainly because it prints a `long long` directly to a string buffer (see: [here](https://github.com/php/php-src/blob/a04ab4b54b36786e81d71b2398c227492df9058e/ext/date/php_date.c#L1208)), which it returns. Meanwhile, `strtotime()` returns a platform-dependent `long` (see: [here](https://github.com/php/php-src/blob/a04ab4b54b36786e81d71b2398c227492df9058e/ext/date/php_date.c#L1506)).

---

There is no clear solution for ODM, since [`MongoDate::__construct()`](https://github.com/mongodb/mongo-php-driver-legacy/blob/05db7766158f0d0d24c4bb903adbe8d52d078a46/types/date.c#L71) takes `long` values for its `sec` and `usec` parameters. Ultimately, MongoDate would need to accept a string as its first argument, and parse that directly into an `int64_t` (reported and tracked by [PHP-1424](https://jira.mongodb.org/browse/PHP-1424)).

malarzm:
@jmikola related JIRA ticket was made for legacy driver, is this something new one is tackling/will tackle?

[...]

PR #1063, opened 2015-03-28
Added support for dates with milliseconds

### LINKED ISSUE #1061: DateType should support and preserve millisecond precision
[DateType](https://github.com/doctrine/mongodb-odm/blob/master/lib/Doctrine/ODM/MongoDB/Types/DateType.php) currently operates at second precision, while BSON dates (i.e. MongoDate) supports millisecond precision. Note that although MongoDate's second constructor argument is microseconds, it internally truncates down to milliseconds when creating its internal `int64_t` value used for BSON serialization (see: [here](https://github.com/mongodb/mongo-php-driver/blob/05db7766158f0d0d24c4bb903adbe8d52d078a46/types/date.c#L52)).

PHP's DateTime type does support microseconds, and we can output those values via the `u` format string:

> Microseconds (added in PHP 5.2.2). Note that `date()` will always generate 000000 since it takes an integer parameter, whereas `DateTime::format()` does support microseconds.

DateType should include this value when creating a MongoDate.

Note: this issue closely relates to #1060, and would also benefit from MongoDate supporting a string `$sec` constructor argument to overcome limitations on 32-bit platforms (see: [PHP-1424](https://jira.mongodb.org/browse/PHP-1424)).

lavoiesl:
Fixes #1061 

Affects #1060 : now will overflow and the loaded date will be in 2036. Note that problem only appears when saving to and loading from Mongo, not when simply converting values.

Duplicate of #1028 (sorry, didn’t see).

Microseconds passed as MongoDate, DateTime, float or DateTime string (`'2000-01-01 00:00:00.123'`) are all handled.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Types/DateType.php
+++ b/lib/Doctrine/ODM/MongoDB/Types/DateType.php
@@ -28,26 +28,68 @@ namespace Doctrine\ODM\MongoDB\Types;
  */
 class DateType extends Type
 {
+    /**
+     * Converts a value to a DateTime.
+     * Supports microseconds
+     *
+     * @throws InvalidArgumentException if $value is invalid
+     * @param  mixed $value \DateTime|\MongoDate|int|float
+     * @return \DateTime
+     */
+    public static function getDateTime($value)
+    {
+        $datetime = false;
+        $exception = null;
+
+        if ($value instanceof \DateTime) {
+            return $value;
+        } elseif ($value instanceof \MongoDate) {
+            $datetime = self::craftDateTime($value->sec, $value->usec);
+        } elseif (is_numeric($value)) {
+            $seconds = $value;
+            $microseconds = 0;
+
+            if (false !== strpos($value, '.')) {
+                list($seconds, $microseconds) = explode('.', $value);
+                $microseconds = (int) str_pad((int) $microseconds, 6, '0'); // ensure microseconds
+            }
+
+            $datetime = self::craftDateTime($seconds, $microseconds);
+        } elseif (is_string($value)) {
+            try {
+                $datetime = new \DateTime($value);
+            } catch (\Exception $e) {
+                $exception = $e;
+            }
+        }
+
+        if ($datetime === false) {
+            throw new \InvalidArgumentException(sprintf('Could not convert %s to a date value', is_scalar($value) ? '"'.$value.'"' : gettype($value)), 0, $exception);
+        }
+
+        return $datetime;
+    }
+
+    private static function craftDateTime($seconds, $microseconds = 0)
+    {
+        $datetime = new \DateTime();
+        $datetime->setTimestamp($seconds);
+        if ($microseconds > 0) {
+            $datetime = \DateTime::createFromFormat('Y-m-d H:i:s.u', $datetime->format('Y-m-d H:i:s') . '.' . $microseconds);
+        }
+
+        return $datetime;
+    }
+
     public function convertToDatabaseValue($value)
     {
-        if ($value === null) {
-            return null;
-        }
-        if ($value instanceof \MongoDate) {
+        if ($value === null || $value instanceof \MongoDate) {
             return $value;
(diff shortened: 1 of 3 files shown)


lib/Doctrine/ODM/MongoDB/Types/DateType.php
```diff
@@ -28,26 +28,53 @@
...
-    public function convertToDatabaseValue($value)
+    /**
+     * Converts a value to a DateTime.
+     * Supports microseconds
+     *
+     * @throws InvalidArgumentException if $value is invalid
+     * @param  mixed $value \DateTime|\MongoDate|int|float
+     * @return \DateTime
+     */
+    public static function getDateTime($value)
```
lavoiesl: The DateTime conversion is getting too heavy to nice fit in the "closures" and there would be too much code duplication otherwise, hence the public static function. I am open to alternatives.

lib/Doctrine/ODM/MongoDB/Types/DateType.php
```diff
@@ -28,26 +28,53 @@
...
-        }
-        $timestamp = false;
+        $datetime = false;
+        $exception = null;
+
         if ($value instanceof \DateTime) {
-            $timestamp = $value->format('U');
+            return $value;
+        } elseif ($value instanceof \MongoDate) {
+            $datetime = \DateTime::createFromFormat('U.u', $value->sec . '.' . $value->usec);
```
lavoiesl: Invalid dates passed to `createFromFormat` will make it return false, without exception, hence this code organization.

tests/Doctrine/ODM/MongoDB/Tests/Functional/DateTest.php
```diff
@@ -56,6 +56,9 @@ public function provideEquivalentDates()
             array(new \DateTime('1985-09-01 00:00:00'), new \DateTime('1985-09-01 00:00:00')),
             array(new \DateTime('2012-07-11T14:55:14-04:00'), new \DateTime('2012-07-11T19:55:14+01:00')),
             array(new \DateTime('@1342033881'), new \MongoDate(1342033881)),
+            array(\DateTime::createFromFormat('U.u', '100000000.123456'), new \MongoDate(100000000, 123000)),
```
lavoiesl: Tests heavily for the truncation of microseconds.

tests/Doctrine/ODM/MongoDB/Tests/Mapping/Types/DateTypeTest.php
```diff
@@ -41,6 +43,7 @@ public function provideInvalidDateValues()
             'string' => array('whatever'),
             'bool'   => array(false),
             'object' => array(new \stdClass()),
+            'invalid string' => array('foo'),
```
lavoiesl: Not strictly related to this PR, but I figured it was missing.

lib/Doctrine/ODM/MongoDB/Types/DateType.php
```diff
@@ -28,26 +28,53 @@
...
         } elseif (is_numeric($value)) {
-            $timestamp = $value;
+            if (false !== strpos($value, '.')) {
```
lavoiesl: `\DateTime::createFromFormat('U.u', $value);` will fail if $value is an integer.
Using `strpos` because testing for string integer/float separately is quite a hassle.

lavoiesl:
The Travis issues are related to truncation, I would need to address them.

lavoiesl:
I decided to remove tests for truncation because Mongo is not consistent across its version.

I could have enforce the truncation in PHP, but it would significantly slow the process and I don’t think it’s worth it, microseconds should not matter to end user anyway since Mongo doesn’t support it.

lavoiesl:
Travis errors now unrelated, good to merge.

jwage:
thumbs_walter

Bilge:
Are you Jesus?

PR #1066, opened 2015-03-28
[WIP] Implementing atomic many updates in CollectionPersister.

jwage:
TODO:
- Add test for ReferenceMany. Same code handles both embed and reference so it should be the same.
- Updates to embed many documents using the atomic strategy should also be atomic and should not update based on the positional key.

This cannot be fully supported until we have this https://jira.mongodb.org/browse/SERVER-831

Given the following test document:

```
> db.test.insert({"test":[{"_id":ObjectId()}]})
> db.test.find().pretty()
{
    "_id" : ObjectId("5516e2f249bf305dfce8ca7c"),
    "test" : [
        {
            "_id" : ObjectId("5516e2f249bf305dfce8ca7b")
        }
    ]
}
```

With the atomic strategy the queries to remove a many document change to this:

```
> db.test.update({"_id" : ObjectId("5516e2f249bf305dfce8ca7c")}, {$pull:{"test":{"_id":ObjectId("5516e2f249bf305dfce8ca7b")}}})
```

From this:

```
> db.test.update({"_id" : ObjectId("5516d39235447e556557c694")}, {$unset:{"test.0":true}})
> db.test.update({"_id" : ObjectId("5516d39235447e556557c694")}, {$pull:{"test":null}})
```

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/CollectionPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/CollectionPersister.php
@@ -116,6 +116,12 @@ class CollectionPersister
                 $this->insertElements($coll, $options);
                 break;
 
+            case 'atomic':
+                $coll->initialize();
+                $this->atomicDeleteElements($coll, $options);
+                $this->insertElements($coll, $options);
+                break;
+
             default:
                 throw new \UnexpectedValueException('Unsupported collection strategy: ' . $mapping['strategy']);
         }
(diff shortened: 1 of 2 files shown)


malarzm:
Jusr for the record as discussed on IRC yesterday: I'd suggest using additional ``atomic=true`` flag for many mapping (which will for now throw exception if used along with set/setArray strategies) or at least renaming the strategy (``atomicPull``?) as ``atomic`` itself is too generic and can be misleading in future :)

blockjon:
@malarzm I'm in a bit of a jam and I need the solution you are proposing to solve for the PR I filed today https://github.com/doctrine/mongodb-odm/pull/1094 Do you have any sense of the time frame for releasing this fix?

jwage:
@blockjon The best bet is for you to implement what you need and submit it upstream. I wouldn't rely on someone here doing what you need in the timeframe you need it in.

malarzm:
@blockjon I'll try to look into it tomorrow or weekend the latest and let you know

jwage:
Or maybe @malarzm can do it :)

blockjon:
I just opened a pull request against this branch/PR and I see the flaw I've exposed is still a problem even in this branch.

That being said, I still need to hack a solution in the near term. My codebase is currently using the BETA11 tag of the ODM. From which branch or tag should I try to code my solution? I'm assuming the master line.

malarzm:
@blockjon the master branch, transition from BETA11 to dev-master should be smooth as far as I remember.

alcaeus:
@malarzm is this still relevant now that we have the `atomicSet` and `atomic` strategies?

malarzm:
IIRC idea behind this was to assign some kind of identifiers to embedded documents and use only that for pull. Personally I think that what we have now is sufficient especially given this can't be fully implemented without change in Mongo itself. And for that there is an open ticket since 2010 and now people are even taking bets there:

> I am taking bets. I put $20 that this will be here in 2019, open. Takers?

All in all I'll close this

PR #1067, opened 2015-03-28
Run PHP CS Fixer.

jwage:
php-cs-fixer fix lib --fixers=blankline_after_open_tag,concat_without_spaces,double_arrow_multiline_whitespaces,duplicate_semicolon,empty_return,extra_empty_lines,include,multiline_array_trailing_comma,ordered_use,no_blank_lines_before_namespace,unused_use,whitespacy_lines,default

    php-cs-fixer fix tests --fixers=blankline_after_open_tag,concat_without_spaces,double_arrow_multiline_whitespaces,duplicate_semicolon,empty_return,extra_empty_lines,include,multiline_array_trailing_comma,ordered_use,no_blank_lines_before_namespace,unused_use,whitespacy_lines,default


malarzm:
Two things from me:

1. ``__NAMESPACE__.'\GH597Post'`` is not as readable as ``__NAMESPACE__ . '\GH597Post'`` - can we live without ``concat_without_spaces``
2. ``empty_return`` can we use explicit ``return null;`` instead?

jmikola:
> empty_return can we use explicit `return null;` instead?

I'm not familiar with how `empty_return` is enforced in the CS fixer (does it convert `return null` to `return`?), but I wouldn't want to give up the ability to have simple `return` statements for void functions.

jwage:
Ya, it removed stuff like that. I didn't like it either. The default set of fixers I think is good for us.

Issue #1065, opened 2015-03-28
ReferenceMany with repositoryMethod vs. priming references


alcaeus:
We use a bunch of ```ReferenceMany``` mappings with ```repositoryMethod``` in our software, sometimes including ```sort``` and ```limit``` fields. To reduce the number of database queries, we tried adding reference priming in some of the queries. Since reference priming is implemented in ```Doctrine\ODM\MongoDB\Query\Query``` (https://github.com/doctrine/mongodb-odm/blob/master/lib/Doctrine/ODM/MongoDB/Query/Query.php#L260..L268), there were two big problems with this:

1. Using ```limit``` or ```sort``` in the mapping causes exceptions. This is due to the fact that the query loops over the results of the queries to collect the objects to be primed. This opens the MongoDB cursor object which means that no limit or sort can be applied to the cursor anymore.
2. When using ```count()``` on the field, the query would be very slow since the execute() method of the query (to return the cursor) would always prime references, even when we're not even interested in the objects themselves.

To fix this, reference priming could be moved to ```Doctrine\ODM\MongoDB\Cursor``` and only prime references once the first object is to be hydrated. This would resolve both issues:
1. The cursor would definitely have to be opened when returning the first result of the query, so we're not breaking any future limit or sort operations since they were impossible anyways.
2. Using ```count()``` would not cause objects to be loaded and thus not trigger reference priming.

I started working on implementing this into the ```Cursor``` class (and into ```EagerCursor``` as well) but wanted to get some feedback on this before finishing up the job. What do you guys think?

acrozes:
I have an error in link with this case, but without using limit() or sort():

Documents:
```
class Account
{
    /**
     * @MongoDB\ReferenceMany(
     *  targetDocument="State",
     *  mappedBy="account",
     *  cascade={"remove"}
     * )
     */
    protected $states = [];

    /**
     * @MongoDB\ReferenceMany(
     *  targetDocument="State",
     *  mappedBy="account",
     *  sort={"createdAt"="desc"},
     *  limit=480
     * )
     */
    protected $lastStates = [];

    /**
     * @MongoDB\ReferenceMany(
     *  targetDocument="State",
     *  mappedBy="account",
     *  repositoryMethod="getLastDailyStates"
     * )
     */
    protected $lastDailyStates = [];
}

class State
{
    /**
     * @MongoDB\ReferenceOne(targetDocument="Account", inversedBy="states")
     */
    protected $account;
}

```
And the repo:
[...]

Issue #520, opened 2013-03-05
prime and getSingleResult don't play nice together


superdweebie:
The following query will raise exception

```
$result = $dm->createQueryBuilder()
    ->find($myClass)
    ->field($myField)->prime(true)
    ->getQuery()
    ->getSingleResult();
```

`cannot modify cursor after beginning iteration` raised in `Doctrinne\MongoDB\Cursor.php` line 271

It will work if `execute()` is used instead of `getSingleResult()` - but then you have to muck about with the returned cursor to grab the single result you wanted.

PR #1068, opened 2015-03-31
Move reference priming to Cursor

alcaeus:
This PR moves reference priming from the ```Query::execute``` method to the cursor which iterates over the results (as described in #1065). This increases performance when iterating over part of a large query result (e.g. by setting ```skip``` and ```limit``` on the returned cursor before iterating) as it will not prime references for the entire result set.
Note that using ```prime``` for ```findAndUpdate``` and ```findAndRemove``` will prime references when executing the query since it will not return a cursor.
This also fixes a side effect of #520 where using ```getSingleResult()``` on a query without ```limit``` in combination with ```prime``` would prime references for all documents returned in the query instead of just the one returned by ```getSingleResult()```.

Feedback is greatly appreciated, I'd really appreciate it if some people using reference priming could run their test suite against the branch and post some results.

NB: I'm not too happy with having to duplicate functionality in ```Cursor``` and ```EagerCursor```, unfortunately I can't think of another way since the two classes don't share an interface.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Cursor.php
+++ b/lib/Doctrine/ODM/MongoDB/Cursor.php
@@ -24,6 +24,7 @@ use Doctrine\MongoDB\Connection;
 use Doctrine\MongoDB\Cursor as BaseCursor;
 use Doctrine\ODM\MongoDB\Mapping\ClassMetadata;
 use Doctrine\ODM\MongoDB\Query\Query;
+use Doctrine\ODM\MongoDB\Query\ReferencePrimer;
 
 /**
  * Wrapper for the Doctrine\MongoDB\Cursor class.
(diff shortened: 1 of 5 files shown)


jwage:
Nice. Looks good to me! :+1:

smolinari:
:+1: Thanks for the improvement!

Scott

blockjon:
Could this PR have caused https://github.com/doctrine/mongodb-odm/issues/1108 ?

jmikola:
@blockjon: I just looked at this PR diff and I suspect that the new call to `primeReferences()` from within `current()` ([here](https://github.com/doctrine/mongodb-odm/pull/1068/files#diff-8ccf0601339e49515848e80a1db39dc0R241)) may be related to #1108. Since MongoCursor iteration uses the standard Iterator interface methods, the order of execution will be `rewind()`, `next()`, `valid()`, and then `current()`. Here, it looks like we're invoking `rewind()` once again on each first call to `current()`. So even if no reference priming is requested, it looks as if the query gets re-executed at least one additional time (after the `foreach` on configured primers, each of which rewind on their own, too!).

Even considering the original priming code, I don't see any reason that this should be re-executing the cursor for each primer. There is no guarantee that the query is going to return the same results on successive executions. Although it would come at an expense of memory, I think that priming would best be restricted to EagerCursors, where all of the documents are cached in an array and no re-execution takes place (`rewind()` simply resets the array pointer).

@jwage, @alcaeus, @malarzm: Thoughts?

malarzm:
@jmikola sounds reasonable. So in `QueryBuilder` ODM will enforce creating query returning EagerCursor` if `prime()` is called, right?

jmikola:
I think it'd be reasonable to say that `prime(true)` implies `eagerCursor(true)`.

PR #1072, opened 2015-04-03
Add defaultDiscriminatorValue to mapping

alcaeus:
When working with large databases, refactoring your documents can be quite difficult. When I tried to split up a large document into multiple smaller once using single_collection inheritance and discriminator maps I realized that I'd have to update every document and, even worse, every single reference to those documents and add the correct discriminator to all documents and references.

This PR introduces a ```defaultDiscriminatorValue``` mapping which can be used to specify a default discriminator value that will be used if a document or association has no discriminator set. This makes it unnecessary to update the database and instead lets the ODM do the lifting.
It works by adding ```null``` to the ```discriminatorValues``` in the query builder if a ```defaultDIscriminatorValue``` has been set and it is in the list of discriminators being queried. This allows queries to find documents without discriminator and treats them as documents of the chosen type.

The mapping is similar to the rest of the discriminator mappings:
* Annotations:
```
/**
 * @ODM\Document(collection="cms_users")
 * @ODM\DiscriminatorField(fieldName="discr")
 * @ODM\DiscriminatorMap({"default"="Doctrine\ODM\MongoDB\Tests\Mapping\AbstractMappingDriverUser"})
 * @ODM\DefaultDiscriminatorValue("default")
 */
```
* YAML:
```
Doctrine\ODM\MongoDB\Tests\Mapping\AbstractMappingDriverUser:
  type: document
  collection: cms_users
  discriminatorField:
    fieldName: discr
  discriminatorMap:
    default: Doctrine\ODM\MongoDB\Tests\Mapping\AbstractMappingDriverUser
  defaultDiscriminatorValue: default
```
* XML:
```
<document name="Doctrine\ODM\MongoDB\Tests\Mapping\AbstractMappingDriverUser" collection="cms_users">
    <discriminator-field name="discr" />
    <discriminator-map>
        <discriminator-mapping value="default" class="Doctrine\ODM\MongoDB\Tests\Mapping\AbstractMappingDriverUser" />
    </discriminator-map>
    <default-discriminator-value value="default" />
</document>
```

What's missing?
* ~~Only the annotation reference has been written so far. The feature still needs to be documented in the respective sections about document inheritance and references.~~
* ~~More test cases. The test cases for discriminators are a bit spread out, so adding test cases is a bit cumbersome. A test to check that a single document without discriminator value is found has been added, but all of the association mapping stuff is still untested. I'd add that once I've gotten some feedback on the whole thing.~~

### DIFF
--- a/docs/en/reference/annotations-reference.rst
+++ b/docs/en/reference/annotations-reference.rst
@@ -178,6 +178,29 @@ in MongoDB. The property will be a DateTime when loaded from the database.
     /** @Date */
     private $createdAt;
 
+@DefaultDiscriminatorValue
+--------------------------
+
+This annotation can be used when using `@DiscriminatorField`_. It will be used
+as a fallback value if a document has no discriminator field set. This must
+correspond to a value from a discriminator map.
+
+.. code-block:: php
+
+    <?php
+
+    /**
+     * @Document
+     * @InheritanceType("SINGLE_COLLECTION")
+     * @DiscriminatorField("type")
+     * @DiscriminatorMap({"person" = "Person", "employee" = "Employee"})
+     * @DefaultDiscriminatorValue("person")
+     */
+    class Person
+    {
+        // ...
+    }
+
 @DiscriminatorField
 -------------------
 
@@ -344,6 +367,9 @@ Optional attributes:
 -
     discriminatorMap - Map of discriminator values to class names.
 -
+    defaultDiscriminatorValue - A default value for discriminatorField if no value
+    has been set in the embedded document.
+-
     strategy - The strategy used to persist changes to the collection. Possible
     values are ``addToSet``, ``pushAll``, ``set``, and ``setArray``. ``pushAll``
     is the default. See :ref:`collection_strategies` for more information.
@@ -359,7 +385,8 @@ Optional attributes:
      *     discriminatorMap={
      *         "book"="Documents\BookTag",
      *         "song"="Documents\SongTag"
-     *     }
+     *     },
+     *     defaultDiscriminatorValue="book"
      * )
      */
     private $tags = array();
@@ -394,6 +421,9 @@ Optional attributes:
     value within the embedded document.
 -
     discriminatorMap - Map of discriminator values to class names.
+-
+    defaultDiscriminatorValue - A default value for discriminatorField if no value
+    has been set in the embedded document.
 
 .. code-block:: php
 
@@ -405,7 +435,8 @@ Optional attributes:
      *     discriminatorMap={
      *         "user"="Documents\User",
      *         "author"="Documents\Author"
-     *     }
+     *     },
+     *     defaultDiscriminatorValue="user"
      * )
      */
     private $creator;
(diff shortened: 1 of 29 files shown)


gruberro:
:+1:

jwage:
:+1: I could definitely use this. I have implemented default discriminator values in my project using the `preLoad($&data)` event. Can you rebase the PR?

smolinari:
I agree, this is a good addition, since using a document type discriminator (I would have thought), would actually be a best practice?  <- It's a question. :-)

Scott

alcaeus:
@smolinari I've only added discriminators when refactoring documents to add single_collection inheritance. When you only store a single document type in a collection or reference there's no need for discriminators. Is that what you meant?

smolinari:
What I was asking is, despite only having one document type, at first, shouldn't you still use a discriminator field, just in case things change later.

Scott

alcaeus:
@smolinari That is the kind of foresight that, in hindsight, everybody should have. However, when designing classes I usually don't go around thinking "Hm, I might refactor this later, let's add a discriminator I don't really need." ;)

smolinari:
Yes, obviously you wouldn't do that, but if it is "standard" then you'd just have to do it. It might be overhead you'll never need too, but it always leaves you some options open for later. It is sort of like thinking about your shard key before you really need it (and even if you never need it):)

Scott

alcaeus:
Yes, but if you require a defaultDiscriminatorValue for every Document mapping you're also going to need a discriminatorMap. Discriminators are only used when using ```single_collection``` inheritance - so if you're using ```collection_per_class``` it's of no use and you really should not have to define it. So in the end it would end up as being a best practice that is written down somewhere in the docs where people just wouldn't bother because it's overhead and you're not going to need it for a while. Then you're stuck in the same place as now where people refactor it and realize that even though they've worked with MongoDB and ODM for quite some time now they've been stupid once again and forgotten to think of it (and I'm guilty as charged) :)

smolinari:
LOL! Ok. We'll leave it at that.

Scott

malarzm:
@alcaeus is there something more you want to add to docs? Asking since

> Only the annotation reference has been written so far. The feature still needs to be documented in the respective sections about document inheritance and references.

is not crossed out but [WIP] is removed :)

alcaeus:
Since there is no documentation on how to add discriminators later on and how to deal with the consequences, I decided against it. I can always try to write something down if you think it's necessary. Otherwise I'd leave it up to people to discover it in the annotation reference, along with many other features ;)

alcaeus:
@malarzm You convinced me - I've added a small section to the documentation on document inheritance as well as examples on how to use this feature for embedded/reference mapping. What do you think?

malarzm:
@alcaeus awesome for me ;) Merged manually in https://github.com/doctrine/mongodb-odm/commit/669789d5f49d8e422ae632a92fa06a248025f2c5

smolinari:
:+1: 

The discriminator is something I'll be using a lot. Thanks for the attention to it.

Scott

PR #1073, opened 2015-04-03
[DX] Adding bool and integer types and annotations

malarzm:
I keep using ``@ODM\Bool`` annotation and later swearing it doesn't exist and changing to full ``@ODM\Boolean``. Here are some pros of adding this:

1. We have ``int`` type, not ``integer``
2. [Scalar Type Hints](https://wiki.php.net/rfc/scalar_type_hints_v5) are in favor of ``bool``
3. ``bool`` is common in all languages

The only downside of this I see is that if somebody would have already done that on his own via ``Type::addType`` he will get exception now.


gerryvdm:
Shouldn't we, by the same reasoning, also add an "Integer" type? :)

malarzm:
@gerryvdm If we end up merging this I'm totally fine with having ``@ODM\Integer`` especially if it will save somebody a swear, DX is good thing ;)

alcaeus:
:+1: for bool
:+1: for integer
:+1: for adding the same DX to the XML and YAML mapping

lavoiesl:
:+1:, and Integer should exist too.

jmikola:
I assume the "alias" classes, which are really for annotations, will just extend the existing one. Is that correct?

I've no objections to creating both "Bool" and "Integer". Is this DX suggestion relevant to ORM?

malarzm:
@jmikola can't really extend existing one since it's final and I didn't want to change that. Also by registering normal type it should work for YAML/XML mappings out of the box

Haven't used ORM in quite a while but IIRC they had something like ``@ORM\Column(type="boolean")`` so I can submit similar thing there later

smolinari:
I have to ask a totally noob question, but what does DX mean?

Scott

malarzm:
@smolinari  Developer eXperience :)

lavoiesl:
@malarzm Could you update your PR to support Integer as well?

malarzm:
@lavoiesl I'll do that later today or tomorrow, don't worry ;)

smolinari:
@malarzm Oh, duh! Thanks for reminding me. I actually just read a few weeks ago [the blog about the push Symfony is making on DX](http://symfony.com/blog/making-the-symfony-experience-exceptional). I should have guessed that. I just couldn't put it together in context to the discussion here. Enough excuses. I am simply a noob. LOL!

Scott

malarzm:
``Integer`` type and annotation are added

jmikola:
@jwage: Spoke in IRC about this. DBAL has "boolean" and "integer" mappings, while we only had "boolean" and "int". @malarzm raises a point that this would break anyone with custom "integer" or "bool" types defined, but I don't think that's likely to affect very many users (why would you have a custom "bool" type that differs significantly from the stock "boolean" type?).

PR #1074, opened 2015-04-05
pass file content instead of file path to Yaml parser

adam187:
Parsing file paths in Yaml is deprecated
https://github.com/symfony/symfony/blob/2.7/src/Symfony/Component/Yaml/Yaml.php#L58


malarzm:
Thanks @adam187!

Issue #1078, opened 2015-04-16
Problem with serializing uninitialized PersistentCollection


ofetisov:
Hi guys! I serialize my entire document which has some embedded array-fields (which are fetched as PersistentCollection) and if they are uninitialized (and they really are, right after fetch and hydration) I've got unrepresentative string which doesn't tells anything about it's content. It seems quite bad for me doing initialization of every "array" of my document before serialization, so i came up with idea: what do you think about adding  'mongoData' string into array that returns from "__sleep" method? I've tested and it seems not ruining any tests, and in general it looks like serialization is not used at all. So I think this is not big deal. Thanks in advance!

malarzm:
Won't it break after unserialization under certain conditions? I mean that ``PersistentCollection`` needs ``UnitOfWork`` to [initialize itself](https://github.com/doctrine/mongodb-odm/blob/683728eb0b4723dd5551686addc724f8f1b13afe/lib/Doctrine/ODM/MongoDB/PersistentCollection.php#L185) which might not be available after ``unserialize($document);`` if you won't merge document first.

ofetisov:
@malarzm and won't it break now? I think no, cause unserialized object will have no "mapping" and will fail on this [line](https://github.com/doctrine/mongodb-odm/blob/683728eb0b4723dd5551686addc724f8f1b13afe/lib/Doctrine/ODM/MongoDB/PersistentCollection.php#L173). And I can't see how my improvement will influence on this.

malarzm:
Oh sorry for oversight I haven't looked at second condition there :) If no tests are failing and everything is working properly I'm personally ok with addition :+1:

ofetisov:
Should i make some PR or we need some more thoughts about that?

malarzm:
We can always have a discussion in a PR :)

malarzm:
Closing along with PR, see https://github.com/doctrine/mongodb-odm/pull/1079#issuecomment-172271018

PR #1079, opened 2015-04-17
PersistentCollection serialization fix

ofetisov:
Serialization fix to show representative string if collection is uninitialized. Related to [issue 1078](https://github.com/doctrine/mongodb-odm/issues/1078).

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/PersistentCollection.php
+++ b/lib/Doctrine/ODM/MongoDB/PersistentCollection.php
@@ -613,7 +613,7 @@ class PersistentCollection implements BaseCollection
      */
     public function __sleep()
     {
-        return array('coll', 'initialized');
+        return array('coll', 'initialized', 'mongoData');
     }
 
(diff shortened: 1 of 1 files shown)


malarzm:
@ofetisov can you add tests for this?

ofetisov:
@malarzm this should be smt like
```PHP
    $pCollectionFull = new PersistentCollection($collection, $dm, $uow);
    $pCollectionFull->setMongoData([1, 2, 3, 4, 5]);

    $pCollectionEmpty = new PersistentCollection($collection, $dm, $uow);

    $this->assertNotEquals(serialize($pCollectionFull), serialize($pCollectionEmpty));
```
or something more complicated?

malarzm:
I'd rather have functional test for this - so make a document with ref/embed many field, serialize it, unserialize it and check if unserialized instance will work as expected

malarzm:
IIRC we were discussing this on Skype and came to conclusion that reasons for making the change are not so good :)

PR #1080, opened 2015-04-17
Add missing serialization of MetadataClass properties

gruberro:
In case a collection is marked as capped the related properties ```collectionCapped```, ```collectionSize``` and ```collectionMax``` are not serialized what results in wrong collection creations if the metadata has already been cached.


malarzm:
Thanks @gruberro!

gruberro:
:+1: for your speed!!

Issue #288, opened 2012-04-08
Reference/Embbeded column with nullable=true couldn't be set to null by setMethod


kusmierz:
I think for sth like

```php
/**
 * 
 * @MongoDB\ReferenceOne(targetDocument="Sth", nullable=true)
 */
protected $test;
```

should produce

```php
public function setTest(Sth $test = null)
    {
        $this->test = $test;
    }
```

instead of 

```php
public function setTest(Sth $test)
    {
        $this->test = $test;
    }
```

According to documentation (http://php.net/manual/en/language.oop5.typehinting.php) this is the only way for accepting "null" as parameter when type hinting. Of course if there is "nullable=false" generator could create method as now.

PR #1081, opened 2015-04-20
Fix missing "nullable" option support

ebuildy:
If I want an optional document reference (with nullable=true), this will not set the setter argument type.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Tools/DocumentGenerator.php
+++ b/lib/Doctrine/ODM/MongoDB/Tools/DocumentGenerator.php
@@ -764,11 +764,12 @@ public function <methodName>()
             return;
         }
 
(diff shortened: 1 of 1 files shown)


malarzm:
@ebuildy isn't it better to generate code with typehint and add ``= null`` to method signature? Also to cross reference there's an [old issue](https://github.com/doctrine/mongodb-odm/issues/288) requesting this functionality :)

ebuildy:
You are right, it was a very quick fix so my team can continue to work ^^. Then I took a look to how EntityGenerator is dealing with this, and they append " = null".

I am changing the PR to copy as much as possible the EntityGenerator.

malarzm:
Cool, feel free to ping me once you finish :)

PR #1084, opened 2015-04-22
Fix #288 nullable option not working for reference

### LINKED ISSUE #288: Reference/Embbeded column with nullable=true couldn't be set to null by setMethod

ebuildy:
Reference/Embbeded column with nullable=true couldn't be set to null by setMethod

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Tools/DocumentGenerator.php
+++ b/lib/Doctrine/ODM/MongoDB/Tools/DocumentGenerator.php
@@ -109,7 +109,7 @@ public function <methodName>()
  * @param <variableType>$<variableName>
  * @return self
  */
-public function <methodName>(<methodTypeHint>$<variableName>)
+public function <methodName>(<methodTypeHint>$<variableName><variableDefault>)
 {
 <spaces>$this-><fieldName> = $<variableName>;
 <spaces>return $this;
@@ -461,7 +461,7 @@ public function <methodName>()
     {
         if ($this->extendsClass()) {
             // don't generate method if its already on the base class.
-            $reflClass = new \ReflectionClass($this->getClassToExtend());
+            $reflClass = new \ReflectionClass($this->getClassTogenerateDocumentStubMethodExtend());
             if ($reflClass->hasMethod($method)) {
                 return true;
             }
@@ -667,6 +667,7 @@ public function <methodName>()
                     $methods[] = $code;
                 }
             } elseif ($fieldMapping['type'] === ClassMetadataInfo::ONE) {
+                $nullable = $this->isAssociationIsNullable($fieldMapping) ? 'null' : null;
                 if ($code = $this->generateDocumentStubMethod($metadata, 'set', $fieldMapping['fieldName'], isset($fieldMapping['targetDocument']) ? $fieldMapping['targetDocument'] : null)) {
                     $methods[] = $code;
                 }
@@ -687,6 +688,16 @@ public function <methodName>()
         }
 
         return implode("\n\n", $methods);
+    }
+    
+    /**
+     * @param array $fieldMapping
+     *
+     * @return bool
+     */
+    protected function isAssociationIsNullable($fieldMapping)
+    {
+        return isset($fieldMapping['nullable']) && $fieldMapping['nullable'];
     }
 
     private function generateDocumentLifecycleCallbackMethods(ClassMetadataInfo $metadata)
@@ -750,7 +761,7 @@ public function <methodName>()
         return implode("\n", $lines);
     }
 
-    private function generateDocumentStubMethod(ClassMetadataInfo $metadata, $type, $fieldName, $typeHint = null)
+    private function generateDocumentStubMethod(ClassMetadataInfo $metadata, $type, $fieldName, $typeHint = null, $defaultValue = null)
     {
         // Add/remove methods should use the singular form of the field name
         $formattedFieldName = in_array($type, array('add', 'remove'))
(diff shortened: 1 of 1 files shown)


lib/Doctrine/ODM/MongoDB/Tools/DocumentGenerator.php
```diff
@@ -461,7 +461,7 @@ private function hasMethod($method, ClassMetadataInfo $metadata)
     {
         if ($this->extendsClass()) {
             // don't generate method if its already on the base class.
-            $reflClass = new \ReflectionClass($this->getClassToExtend());
+            $reflClass = new \ReflectionClass($this->getClassTogenerateDocumentStubMethodExtend());
```
malarzm: Does this method exist?

lib/Doctrine/ODM/MongoDB/Tools/DocumentGenerator.php
```diff
@@ -667,6 +667,7 @@ private function generateDocumentStubMethods(ClassMetadataInfo $metadata)
                     $methods[] = $code;
                 }
             } elseif ($fieldMapping['type'] === ClassMetadataInfo::ONE) {
+                $nullable = $this->isAssociationIsNullable($fieldMapping) ? 'null' : null;
                 if ($code = $this->generateDocumentStubMethod($metadata, 'set', $fieldMapping['fieldName'], isset($fieldMapping['targetDocument']) ? $fieldMapping['targetDocument'] : null)) {
```
malarzm: It seems like you haven't passed ``$nullable`` here

lib/Doctrine/ODM/MongoDB/Tools/DocumentGenerator.php
```diff
@@ -688,6 +689,16 @@ private function generateDocumentStubMethods(ClassMetadataInfo $metadata)
 
         return implode("\n\n", $methods);
     }
+    
+    /**
+     * @param array $fieldMapping
+     *
+     * @return bool
+     */
+    protected function isAssociationIsNullable($fieldMapping)
```
malarzm: there are 2 "is" in method name, also I believe this should be private

lib/Doctrine/ODM/MongoDB/Tools/DocumentGenerator.php
```diff
@@ -771,12 +782,13 @@ private function generateDocumentStubMethod(ClassMetadataInfo $metadata, $type,
...
-            '<variableName>'   => $variableName,
-            '<methodName>'     => $methodName,
-            '<fieldName>'      => $fieldName,
+            '<description>'         => $description,
+            '<methodTypeHint>'      => $methodTypeHint,
+            '<variableType>'        => $variableType,
+            '<variableName>'        => $variableName,
+            '<methodName>'          => $methodName,
+            '<fieldName>'           => $fieldName,
+            '<variableDefault>'     => ($defaultValue !== null ) ? (' = '.$defaultValue) : '',
```
malarzm: Can you please change concat to use spaces so ``' = ' . $defaultValue``?

lib/Doctrine/ODM/MongoDB/Tools/DocumentGenerator.php
ebuildy: True, I copied the declaration from https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/Tools/EntityGenerator.php#L1225, but no problem to remove the extra is and make it private!

ebuildy:
Thanks you for your feedbacks, I am very glad to help in a such important project!

lib/Doctrine/ODM/MongoDB/Tools/DocumentGenerator.php
```diff
@@ -667,7 +667,8 @@ private function generateDocumentStubMethods(ClassMetadataInfo $metadata)
                     $methods[] = $code;
                 }
             } elseif ($fieldMapping['type'] === ClassMetadataInfo::ONE) {
-                if ($code = $this->generateDocumentStubMethod($metadata, 'set', $fieldMapping['fieldName'], isset($fieldMapping['targetDocument']) ? $fieldMapping['targetDocument'] : null)) {
+                $nullable = $this->isAssociationNullable($fieldMapping) ? 'null' : null;
+                if ($code = $this->generateDocumentStubMethod($metadata, 'set', $fieldMapping['fieldName'], isset($fieldMapping['targetDocument']) ? $fieldMapping['targetDocument'] : null), $nullable) {
```
malarzm: ``$nullable`` is still in wrong place, take a look at failing tests in Travis. You can also run tests locally which eases life a lot :)

malarzm:
@ebuildy thanks for the fixed, there's still one thing left as Travis is failing. Could you also squash your commits later?

lib/Doctrine/ODM/MongoDB/Tools/DocumentGenerator.php
ebuildy: I am really sorry, I am not at home nowadays, and work from my wife computer ;-( without Php of course, am fixing it

malarzm: Thanks! Also props for strong will for working without PHP ;)

ebuildy: Github is a good place for that! And with good unit tests of course ;-)

Sorry for the loooonngg PR ^^ I will be more attentive next time.

malarzm:
Merged manually in https://github.com/doctrine/mongodb-odm/commit/0de70ab260a7d0e5b872bb65e79e8ebc1c033b85 - thanks @ebuildy!

PR #1085, opened 2015-04-22
[docs] Fix 'raw_type' to 'raw'.

amyboyd:
As seen here, the type should be `raw`: https://github.com/doctrine/mongodb-odm/blob/94bc60445c95f627228fa7e72cc9958f70c033eb/lib/Doctrine/ODM/MongoDB/Types/Type.php#L57


malarzm:
Thanks @amyboyd!

PR #1091, opened 2015-05-12
Add proper return code to indicate errors in schema console commands

alcaeus:
This adds a proper return code in case a schema command throws an error. Without this, automated scripts (e.g. capifony) can't detect the error and act accordingly.


malarzm:
:+1:

sharpner:
:+1:

gruberro:
:+1:

PR #1092, opened 2015-05-12
[Travis] PHP 5.3 run all, 5.4 to 5.6 only stable driver

malarzm:
Travis matrix is quite big, as discussed with @jmikola this configuration will be enough for ODM


jmikola:
:+1:

PR #1093, opened 2015-05-14
Fixed events.rst Lifecycle Callbacks example

smoya:
Because we know that `m` means month and not minutes.


malarzm:
Definitely you're right :) Thanks!

PR #1094, opened 2015-05-14
Functional test to demonstrate Doctrine ODM queries load inconsistent data from the database while a concurrent request is writing to the same record

blockjon:
The ODM does not save database records with embedded documents in an atomic fashion.

Even with ODM versioning, a concurrent PHP request which loads a Mongo record while another PHP request is writing the same record will load an incomplete and inconsistent version of the Mongo record into a Doctrine ODM model. This has been a repeated cause of data loss and corruption.

Repro steps:

Run this unit test:

```
phpunit --filter testDataConsistencyUsingEmbeddedDocuments ./tests/Doctrine/ODM/MongoDB/Tests/Functional/VersionTest.php
```

Result:
```
PHPUnit 4.1.6 by Sebastian Bergmann.

Configuration read from /srv/repos/instant-server/extrashared/mongodb-odm/phpunit.xml

F

Time: 9.38 seconds, Memory: 3.75Mb

There was 1 failure:

1) Doctrine\ODM\MongoDB\Tests\Functional\VersionTest::testDataConsistencyUsingEmbeddedDocuments
Version #2 of the versioned user object should have a phone number embedded document. However, 0 phone number embedded documents were found
Failed asserting that 0 matches expected 1.

/srv/repos/instant-server/extrashared/mongodb-odm/tests/Doctrine/ODM/MongoDB/Tests/Functional/VersionTest.php:87

FAILURES!
Tests: 1, Assertions: 3, Failures: 1.
```

Issue #437, opened 2012-10-25
Add support for atomic updates


jgornick:
The ODM doesn't support a single document update for all fields and embedded documents. The current implementation will run different queries for updating a single document and it's embedded documents.

This would certainly help remove queries between the ODM and Mongo and also remove the need to implement rollback functionality if an embedded document fails to insert.

I've created a repository with some scenarios @ https://github.com/jgornick/doctrine-atomic-test.

After speaking with @jwage and @jmikola, it would be nice to be able to specify an atomicUpdate option in the relationship annotation:

```
/** @ODM\EmbedMany(…, atomicUpdates=true) */
```

malarzm:
Closing thanks to https://github.com/doctrine/mongodb-odm/pull/1096

Issue #1011, opened 2015-01-13
Superfluous $unset query when using "set" collection strategy 


seb0zz:
Hi,

I am using the Symfony2 bundle I have a document which has some embedded documents. For that I use the EmbedMany annotation with the collection strategy "set". The strange thing is that every time I update this field of the document Doctrine automatically triggers an $unset query before the query that contains the "$set". 

It doesn't seem to be a problem with the Symfony2 bundle.

Does this have a reason or is it a bug?

malarzm:
I can confirm this is happening but I think that problem is elsewhere, consider this passing test:

```
<?php

namespace Doctrine\ODM\MongoDB\Tests\Functional\Ticket;

use Doctrine\Common\Collections\ArrayCollection;
use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;

class GH1011Test extends \Doctrine\ODM\MongoDB\Tests\BaseTest
{
    public function testSchedulingCollections()
    {
        $doc = new GH1011Documnet();
        $doc->embeds->add(new GH1011Embedded('test1'));
        $this->dm->persist($doc);
        $this->dm->flush();
        $this->dm->clear();
        $test = $this->dm->getRepository(get_class($doc))->find($doc->id);
        $test->embeds->add(new GH1011Embedded('test2'));
        $this->uow->computeChangeSets();
        $this->assertTrue($this->uow->isCollectionScheduledForUpdate($test->embeds));
        $this->assertFalse($this->uow->isCollectionScheduledForDeletion($test->embeds));
    }
}

/** @ODM\Document */
class GH1011Documnet
{
    /** @ODM\Id */
    public $id;

    /** @ODM\EmbedMany(targetDocument="GH1011Embedded") */
    public $embeds;

    public function __construct()
    {
        $this->embeds = new ArrayCollection();
    }
}

/** @ODM\EmbeddedDocument */
class GH1011Embedded
{
    /** @ODM\String */
    public $name;
    
    public function __construct($name)
    {
        $this->name = $name;
    }
}
```

``CollectionPersister::delete`` which triggers ``$unset`` is ran by ``UoW`` only if collection is scheduled for deletion.

[...]

PR #1096, opened 2015-05-15
Adding atomicSet and atomicSetArray collection strategies

### LINKED ISSUE #1009: Doctrine take too much time to save/update a document
When document class contains a lot of properties, Doctrine save/update each property one by one this take too much time for response (in my case taken more than 2 minutes) throwing a time out error

**Document class definition**
```php
<?php

namespace MyNameSpace\Document;

use Doctrine\Bundle\MongoDBBundle\Validator\Constraints\Unique;
use Gedmo\Mapping\Annotation as Gedmo;
use Gedmo\Timestampable\Traits\TimestampableDocument;
use JMS\Serializer\Annotation as Serializer;
use Symfony\Component\Validator\Constraints as Assert;

/**
 * Profile
 *
 * @ODM\Document(
 *     indexes={
 *         @ODM\Index(keys={"user"="DES"}, options={"unique"=true})
 *     },
 *     repositoryClass="MyNameSpace\Repository\ProfileRepository"
 * )
 * @ODM\HasLifecycleCallbacks
 * @Unique(
 *     fields={"user"}
 * )
 *
 * @Serializer\XmlRoot("profile")
 */
class Profile {
    /**
     * @ODM\Id(strategy="none")
     * @Serializer\XmlAttribute
     * @Serializer\Groups({"simple"})
     */
    private $id;

    /**
     * @var \MyNameSpace\Document\User
     *
     * @ODM\ReferenceOne(targetDocument="User", simple=true)
     * @Assert\NotNull
     * @Assert\Type(type="object")
     * @Serializer\Groups({"simple"})
     */
    private $user;

    /**
     * @var string
     *
     * @ODM\String
     * @Serializer\Groups({"simple"})
     */
    private $companyName;

    /**
     * @var boolean
     *
     * @ODM\Boolean
     * @Serializer\Groups({"simple"})
     */
    private $isCompany = false;

    /**
     * @var string
     *
     * @ODM\String
     * @Serializer\Groups({"simple"})
     */
    private $title;

    /**
     * @var \MyNameSpace\Document\Country
     *
     * @ODM\ReferenceOne(targetDocument="Country", simple=true)
     * @Assert\NotNull
     */
    private $country;

    /**
     * @var \MyNameSpace\Document\City
     *
     * @ODM\ReferenceOne(targetDocument="City", simple=true)
     * @Assert\NotNull
     */
    private $city;

    /**
     * @var string
     *
     * @ODM\String
     * @Assert\Length(min=0, max=1000)
     */
    private $aboutMe;

    /**
     * @var \MyNameSpace\Document\Interest
     *
     * @ODM\EmbedOne(targetDocument="Interest")
     */
    private $interest;

    /**
     * @var \MyNameSpace\Document\Tag
     *
     * @ODM\EmbedMany(targetDocument="Tag")
     * @Assert\Count(min="0", max="10")
     */
    private $keywords;

    /**
     * @var boolean
     *
     * @ODM\Boolean
     */
    private $located = false;

    /**
     * @var \MyNameSpace\Document\Coordinate
     *
     * @ODM\EmbedOne(targetDocument="Coordinate")
     */
    private $hotspot;

    /**
     * @var float
     *
     * @ODM\Distance
     */
    private $distance;

    /**
     * @var string
     *
     * @ODM\String
     * @Assert\Url
     */
    private $website;

    /**
     * @var string
     *
     * @ODM\String
     * @Assert\Url
     */
    private $vimeo;

    /**
     * @var string
     *
     * @ODM\String
     * @Assert\Url
     */
    private $twitter;

    /**
     * @var string
     *
     * @ODM\String
     * @Assert\Url
     */
    private $imdb;

    /**
     * @var string
     *
     * @ODM\String
     * @Assert\Url
     */
    private $facebook;

    /**
     * @var string
     *
     * @ODM\String
     * @Assert\Url
     */
    private $youtube;

    /**
     * @var string
     *
     * @ODM\String
     * @Assert\Url
     */
    private $linkedin;

    /**
     * @var \MyNameSpace\Document\Cover
     *
     * @ODM\EmbedMany(targetDocument="Cover")
     * @Assert\Count(min="0", max="3")
     */
    private $covers;

    /**
     * @var \MyNameSpace\Document\Recomendation
     *
     * @ODM\EmbedMany(targetDocument="Recomendation")
     */
    private $recomendations;

    /**
     * @var \MyNameSpace\Document\Project
     *
     * @ODM\ReferenceMany(targetDocument="Project", simple=true)s
     */
    private $projects;

    /**
     * @var \MyNameSpace\Document\Profile
     *
     * @ODM\ReferenceMany(targetDocument="Profile", simple=true)
     */
    private $following;

    /**
     * @var \MyNameSpace\Document\Profile
     *
     * @ODM\ReferenceMany(targetDocument="Profile", simple=true)
     */
    private $followers;

    /**
     * @var \MyNameSpace\Document\Project
     *
     * @ODM\ReferenceMany(targetDocument="Project", simple=true)
     */
    private $watching;

    /**
     * Hook timestampable behavior
     * updates createdAt, updatedAt fields
     */
    use TimestampableDocument;


    public function __construct()
    {
        $this->keywords       = new \Doctrine\Common\Collections\ArrayCollection();
        $this->covers         = new \Doctrine\Common\Collections\ArrayCollection();
        $this->recomendations = new \Doctrine\Common\Collections\ArrayCollection();
        $this->projects       = new \Doctrine\Common\Collections\ArrayCollection();
        $this->following      = new \Doctrine\Common\Collections\ArrayCollection();
        $this->followers      = new \Doctrine\Common\Collections\ArrayCollection();
        $this->watching       = new \Doctrine\Common\Collections\ArrayCollection();
    }

   // Setters & Getters

    /**
     * @ODM\PreFlush
     */
    public function prepare() {
        if(!$this->located) {
            $this->hotspot = null;
        }
        if(count($this->keywords) < 1) {
            $this->keywords = null;
        }
        if(count($this->covers) < 1) {
            $this->covers = null;
        }
        if(count($this->recomendations) < 1) {
            $this->recomendations = null;
        }
        if(count($this->projects) < 1) {
            $this->projects = null;
        }
        if(count($this->following) < 1) {
            $this->following = null;
        }
        if(count($this->followers) < 1) {
            $this->followers = null;
        }
        if(count($this->watching) < 1) {
            $this->watching = null;
        }
    }
}
```

**Documet Manager**
```php
<?php

namespace MyNameSpace\Model;

/**
 * DocumentManager
 */
class DocumentManager {

    /**
     * MongoDB Document Manager
     *
     * @var \Doctrine\ODM\MongoDB\DocumentManager
     */
    private $dm;

    /**
     * MongoDB Document Repository
     *
     * @var \Doctrine\ODM\MongoDB\DocumentRepository
     */
    private $repository;

    /**
     *
     * @param \Doctrine\ODM\MongoDB\DocumentManager $dm Doctrine  MongoDB instance
     * @param string                                $documentName name of document
     */
    public function __construct(\Doctrine\ODM\MongoDB\DocumentManager $dm, $documentName) {
        $this->dm = $dm;
        if ($dm) {
            $this->repository = $dm->getRepository($documentName);
        }
    }

    /**
     * Fetch all document in a collection
     *
     * @param int $limit How many documents to return.
     * @param int $offset Offset from which to start listing documents.
     *
     * @return \Doctrine\MongoDB\Collection
     */
    public function fetch($limit = null, $offset = null) {
        return $this->repository->findBy(array(), null, $limit, $offset);
    }

    /**
     * Find one document in a collection
     *
     * @param string $id Document identifier
     *
     * @return object MongoDB document
     */
    public function get($id) {
        return $this->repository->findOneById($id);
    }

    /**
     * Find documents according a criteria
     *
     * @param  array  $criteria
     * @return \Doctrine\MongoDB\Collection
     */
    public function find($criteria = array(), $limit = null, $offset = null)
    {
        return $this->repository->findBy($criteria, null, $limit, $offset);
    }

    /**
     * Save document to MongoDB
     *
     * @param object $document MongoDB document
     */
    public function save($document) {
        $this->dm->persist($document);
        $this->dm->flush();
    }

    /**
     * Update document to MongoDB
     *
     * @param object $document MongoDB document
     */
    public function update() {
        $this->dm->flush();
    }

    /**
     * Delete document from MongoDB
     *
     * @param  object $document MongoDB document
     * @return boolean true when succes
     */
    public function remove($document) {
        $this->dm->remove($document);
        $this->dm->flush();

        return true;
    }

    /**
     * Get Document Repository
     *
     * @return \Doctrine\ODM\MongoDB\DocumentRepository
     */
    public function getRepository() {
        return $this->repository;
    }

    /**
     * Find one document in a collection
     *
     * @param  \Doctrine\ODM\MongoDB\PersistentCollection $collection
     * @param  string                                     $id         Document id
     * @return object                                     $document   MongoDB document
     */
    public function findOneInCollection(\Doctrine\ODM\MongoDB\PersistentCollection $collection, $id)
    {
        $document = $collection->filter(
            function($item) use ($id) {
                return $item->getId() == $id;
            }
        );
        return $document->first();
    }
}
```

**Save/update action on controller**
```php
<?php

namespace MyNameSpace\Controller;

use FOS\RestBundle\Controller\Annotations as REST;
use FOS\RestBundle\Request\ParamFetcherInterface;
use FOS\RestBundle\Util\Codes;
use Nelmio\ApiDocBundle\Annotation\ApiDoc;
use MyNameSpace\Document\Profile;
use MyNameSpace\Form\ProfileType;
use Symfony\Component\HttpFoundation\Request;

/**
 * Profile API
 *
 * @REST\RouteResource("Profile")
 */
class ProfileController extends BaseController
{
// Some actions

    /**
     * Creates a new profile from the submitted data.
     *
     * @param Request $request the request object
     *
     * @return \Symfony\Component\Form\FormTypeInterface|\FOS\RestBundle\View\View
     *
     * @ApiDoc(
     *     resource    = true,
     *     description = "Create profile",
     *     input       = "MyNameSpace\Form\ProfileType",
     *     output      = "MyNameSpace\Document\Profile",
     *     statusCodes = {
     *         201 = "Returned when a new profile is created",
     *         400 = "Returned when error data was sended",
     *         403 = "Returned when is not authorized"
     *     }
     * )
     *
     * @REST\View(
     *     templateVar = "form",
     *     statusCode  = Codes::HTTP_BAD_REQUEST
     * )
     */
    public function postAction(Request $request) {
        $document = new Profile();
        $form     = $this->createForm(new ProfileType(), $document);

        $form->submit($request);
        if ($form->isValid()) {
            $this->get('rest_api.dm.profile')->save($document);

            $view = $this->view($document, 201)
                // set the `Location` header only when creating new resources
                ->setLocation($this->generateUrl(
                    'nu_get_profile', array('id' => $document->getId()),
                    true // absolute
                ));

            return $view;
        }

        return $form;
    }

    /**
     * Update existing profile or create new from the submitted data.
     *
     * @param Request $request the request object
     * @param int     $id      the profile id
     *
     * @return \Symfony\Component\Form\FormTypeInterface|\FOS\RestBundle\View\View
     *
     * @ApiDoc(
     *     resource     = true,
     *     description  = "Update profile",
     *     input        = "MyNameSpace\Form\ProfileType",
     *     output       = "MyNameSpace\Document\Profile",
     *     statusCodes  = {
     *         201 = "Returned when a new resource is created",
     *         204 = "Returned when successful",
     *         400 = "Returned when the data has errors",
     *         403 = "Returned when is not authorized"
     *     },
     *     requirements = {
     *         {
     *             "name"        = "id",
     *             "dataType"    = "string",
     *             "requirement" = "[a-f0-9]{24}",
     *             "description" = "Profile identifier"
     *         }
     *    }
     * )
     *
     * @REST\View(
     *     templateVar = "form",
     *     statusCode  = Codes::HTTP_BAD_REQUEST
     * )
     */
    public function putAction(Request $request, $id) {
        $document   = $this->get('rest_api.dm.profile')->get($id);
        $statusCode = Codes::HTTP_NO_CONTENT;

        if (!$document) {
            $document   = new Profile();
            $statusCode = Codes::HTTP_CREATED;
        }

        $form = $this->createForm(new ProfileType(), $document);

        $form->submit($request);
        if ($form->isValid()) {
            if($statusCode === Codes::HTTP_CREATED)
                $this->get('rest_api.dm.profile')->save($document);
            else
                $this->get('rest_api.dm.profile')->update();

            $view = $this->view($document, $statusCode)
                ->setRoute('nu_get_profile')
                ->setRouteParameters(array('id' => $document->getId()));

            return $view;
        }

        return $form;
    }
// Other actions
}
```

Is possible to save/update all fields at once? and how can I configure this?

malarzm:
This is an early implementation of how ``atomicSet`` and ``atomicSetArray`` could work in my opinion. There are some things that needs to be done

- [x] ~~maybe add `atomic=true` flag to mapping instead of new strategy (or change existing `$set` behaviour)~~
- [x] inserts
- [x] updates
- [x] upserts
- [x] removing collection
- [x] ~~collection within collection with different strategies~~ throw exception if used in embedded documents
- [x] more tests
- [x] docs

cc @jwage @jmikola @alcaeus @lavoiesl @blockjon

If this PR will reach its final form it will resolve #1009, #437, #1095 and #1094

### DIFF
--- a/docs/en/reference/collection-strategies.rst
+++ b/docs/en/reference/collection-strategies.rst
@@ -45,12 +45,38 @@ strategy, but will first numerically reindex the collection to ensure that it is
 stored as a BSON array.
 
 ``pushAll``
-------------
+-----------
 
 The ``pushAll`` strategy uses MongoDB's `$pushAll`_ operator to insert
 elements into the array. MongoDB does not allow elements to be added and removed
 from an array in a single operation, so this strategy relies on multiple update
 queries to remove and insert elements (in that order).
+
+``atomicSet``
+-------------
+
+The ``atomicSet`` strategy uses MongoDB's `$set`_ operator to update the entire
+collection with a single update query. Unlike with ``set`` strategy there will
+be only one query for updating both parent document and collection itself. This
+strategy can be especially useful when dealing with high concurrency and 
+:ref:`versioned documents <annotations_reference_version>`.
+
+.. note::
+
+    The ``atomicSet`` and ``atomicSetArray`` strategies may only be used for 
+    collections mapped directly in a top-level document.
+
+``atomicSetArray``
+------------------
+
+The ``atomicSetArray`` strategy works exactly like ``atomicSet`` strategy,  but 
+will first numerically reindex the collection to ensure that it is stored as a 
+BSON array.
+
+.. note::
+
+    The ``atomicSet`` and ``atomicSetArray`` strategies may only be used for 
+    collections mapped directly in a top-level document.
 
 .. _`$addToSet`: http://docs.mongodb.org/manual/reference/operator/addToSet/
 .. _`$pushAll`: http://docs.mongodb.org/manual/reference/operator/pushAll/
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -73,13 +73,6 @@ class DocumentPersister
     private $uow;
 
     /**
-     * The Hydrator instance
-     *
-     * @var HydratorInterface
-     */
-    private $hydrator;
-
-    /**
      * The ClassMetadata instance for the document type being persisted.
      *
      * @var ClassMetadata
@@ -235,6 +228,11 @@ class DocumentPersister
                 }
                 $data[$versionMapping['name']] = $nextVersion;
             }
+            
+            $atomicCollectionQuery = $this->getAtomicCollectionUpdateQuery($document);
+            if (!empty($atomicCollectionQuery['$set'])) {
+                $data = array_merge($data, $atomicCollectionQuery['$set']);
+            }
 
             $inserts[$oid] = $data;
         }
@@ -281,6 +279,11 @@ class DocumentPersister
                     $this->class->reflFields[$this->class->versionField]->setValue($document, $nextVersionDateTime);
                 }
                 $data['$set'][$versionMapping['name']] = $nextVersion;
+            }
+            
+            $atomicCollectionQuery = $this->getAtomicCollectionUpdateQuery($document);
+            if ( ! empty($atomicCollectionQuery)) {
+                $data = array_merge_recursive($data, $atomicCollectionQuery);
             }
 
             try {
@@ -374,6 +377,11 @@ class DocumentPersister
                     $query[$versionMapping['name']] = new \MongoDate($currentVersion->getTimestamp());
                     $this->class->reflFields[$this->class->versionField]->setValue($document, $nextVersion);
                 }
+            }
+            
+            $atomicCollectionQuery = $this->getAtomicCollectionUpdateQuery($document);
+            if ( ! empty($atomicCollectionQuery)) {
+                $update = array_merge_recursive($update, $atomicCollectionQuery);
             }
 
             /* We got here because the document has one or more related
@@ -646,7 +654,7 @@ class DocumentPersister
 
                 $this->uow->registerManaged($embeddedDocumentObject, $id, $data);
                 $this->uow->setParentAssociation($embeddedDocumentObject, $mapping, $owner, $mapping['name'] . '.' . $key);
-                if ($mapping['strategy'] === 'set') {
+                if ($mapping['strategy'] === 'set' || $mapping['strategy'] === 'atomicSet') {
                     $collection->set($key, $embeddedDocumentObject);
                 } else {
                     $collection->add($embeddedDocumentObject);
@@ -1235,5 +1243,34 @@ class DocumentPersister
         }
 
         return $discriminatorValues;
+    }
+    
+    private function getAtomicCollectionUpdateQuery($document)
+    {
+        $update = array();
+        $collections = $this->uow->getScheduledCollections($document);
+        $collPersister = $this->uow->getCollectionPersister();
+        foreach ($collections as $coll) {
+            /* @var $coll PersistentCollection */
+            $mapping = $coll->getMapping();
+            if ($mapping['strategy'] !== "atomicSet" && $mapping['strategy'] !== "atomicSetArray") {
+                continue;
+            }
+            if ($this->uow->isCollectionScheduledForUpdate($coll)) {
+                $update = array_merge_recursive($update, $collPersister->prepareSetQuery($coll));
+                $this->uow->unscheduleCollectionUpdate($coll);
+                /* TODO:
+                 * Collection can be set for both deletion and update if
+                 * PersistentCollection instance was changed. Since we're dealing
+                 * with collection update in one query we won't need the $unset.
+                 * Line can be removed once the issue is fixed.
+                 */
+                $this->uow->unscheduleCollectionDeletion($coll);
+            } elseif ($this->uow->isCollectionScheduledForDeletion($coll)) {
+                $update = array_merge_recursive($update, $collPersister->prepareDeleteQuery($coll));
+                $this->uow->unscheduleCollectionDeletion($coll);
+            }
+        }
+        return $update;
     }
 }
--- a/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
+++ b/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
@@ -937,7 +937,7 @@ class UnitOfWork implements PropertyChangedListener
             $state = $this->getDocumentState($entry, self::STATE_NEW);
 
             // Handle "set" strategy for multi-level hierarchy
-            $pathKey = $mapping['strategy'] !== 'set' ? $count : $key;
+            $pathKey = ($mapping['strategy'] !== 'set' && $mapping['strategy'] !== 'atomicSet') ? $count : $key;
             $path = $mapping['type'] === 'many' ? $mapping['name'] . '.' . $pathKey : $mapping['name'];
 
             $count++;
@@ -2698,6 +2698,22 @@ class UnitOfWork implements PropertyChangedListener
     {
         return in_array($coll, $this->collectionDeletions, true);
     }
+    
+    /**
+     * INTERNAL:
+     * Unschedules a collection from being updated deleted when this UnitOfWork commits.
+     * Effectively this is used for atomicSet and atomicSetArray update strategies.
+     * The method doesn't remove $coll from $this->hasScheduledCollections because
+     * it is called only from DocumentPersister resolving that very document
+     * 
+     * @param \Doctrine\ODM\MongoDB\PersistentCollection $coll
+     */
+    public function unscheduleCollectionDeletion(PersistentCollection $coll)
+    {
+        if (($key = array_search($coll, $this->collectionDeletions, true)) !== false) {
+            unset($this->collectionDeletions[$key]);
+        }
+    }
 
     /**
      * INTERNAL:
@@ -2712,6 +2728,22 @@ class UnitOfWork implements PropertyChangedListener
     }
     
     /**
+     * INTERNAL:
+     * Unschedules a collection from being updated update when this UnitOfWork commits.
+     * Effectively this is used for atomicSet and atomicSetArray update strategies.
+     * The method doesn't remove $coll from $this->hasScheduledCollections because
+     * it is called only from DocumentPersister resolving that very document
+     * 
+     * @param \Doctrine\ODM\MongoDB\PersistentCollection $coll
+     */
+    public function unscheduleCollectionUpdate(PersistentCollection $coll)
+    {
+        if (($key = array_search($coll, $this->collectionUpdates, true)) !== false) {
+            unset($this->collectionUpdates[$key]);
+        }
+    }
+    
+    /**
      * Checks whether a PersistentCollection is scheduled for update.
      *
      * @param PersistentCollection $coll
@@ -2720,6 +2752,21 @@ class UnitOfWork implements PropertyChangedListener
     public function isCollectionScheduledForUpdate(PersistentCollection $coll)
     {
         return in_array($coll, $this->collectionUpdates, true);
+    }
+    
+    /**
+     * INTERNAL:
+     * Gets PersistentCollections that are scheduled to update and related to $document
+     * 
+     * @param object $document
+     * @return array
+     */
+    public function getScheduledCollections($document)
+    {
+        $oid = spl_object_hash($document);
+        return isset($this->hasScheduledCollections[$oid]) 
+                ? $this->hasScheduledCollections[$oid]
+                : array();
     }
     
     /**
@@ -2752,8 +2799,9 @@ class UnitOfWork implements PropertyChangedListener
             list(, $document, ) = $this->getParentAssociation($document);
             $class = $this->dm->getClassMetadata(get_class($document));
         }
+        $oid = spl_object_hash($document);
 
-        $this->hasScheduledCollections[spl_object_hash($document)] = true;
+        $this->hasScheduledCollections[$oid][] = $coll;
 
         if ( ! $this->isDocumentScheduled($document)) {
             $this->scheduleForUpdate($document);
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/AtomicSetTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/AtomicSetTest.php
@@ -1,0 +1,124 @@
+<?php
+
+namespace Doctrine\ODM\MongoDB\Tests\Functional;
+
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
+use Documents\Phonebook;
+use Documents\Phonenumber;
+
+/**
+ * CollectionPersister will throw exception when collection with atomicSet
+ * or atomicSetArray should be handled by it. If no exception was thrown it
+ * means that collection update was handled by DocumentPersister.
+ */
+class AtomicSetTest extends \Doctrine\ODM\MongoDB\Tests\BaseTest
+{
+    public function testAtomicInsertAndUpdate()
+    {
+        $user = new AtomicUser('Maciej');
+        $user->phonenumbers['home'] = new Phonenumber('12345678');
+        $this->dm->persist($user);
+        $this->dm->flush();
+        $user->surname = "Malarz";
+        $user->phonenumbers['work'] = new Phonenumber('87654321');
+        $this->dm->flush();
+        $this->dm->clear();
+        $newUser = $this->dm->getRepository(get_class($user))->find($user->id);
+        $this->assertEquals('Maciej', $newUser->name);
+        $this->assertEquals('Malarz', $newUser->surname);
+        $this->assertCount(2, $newUser->phonenumbers);
+        $this->assertNotNull($newUser->phonenumbers->get('home'));
+        $this->assertNotNull($newUser->phonenumbers->get('work'));
+    }
+    
+    public function testAtomicUpsert()
+    {
+        $user = new AtomicUser('Maciej');
+        $user->id = new \MongoId();
+        $user->phonenumbers[] = new Phonenumber('12345678');
+        $this->dm->persist($user);
+        $this->dm->flush();
+        $this->dm->clear();
+        $newUser = $this->dm->getRepository(get_class($user))->find($user->id);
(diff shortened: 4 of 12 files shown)


blockjon:
@malarzm I was thinking that the ODM should *always* perform all changes to a record in one atomic query. Do you agree, or do you do you feel as though this behavior should be parameterized?

malarzm:
ODM can't do all changes within one query, for instance (citing @jmikola from https://github.com/doctrine/mongodb-odm/issues/1009#issuecomment-78718247)

> Collections are currently handled separately out of necessity for the base case where elements are being added and removed from the same array, since MongoDB doesn't allow that in the same write operation. 

but of course this in not the case with `$set`. For now I have made this parametrized as I wanted to have current tests pass and only new ones I'm writing fail but changing this will not be hard if we come to conclusion that we want to change existing behaviour.

malarzm:
I'm afraid we won't be able to change existing `$set` behaviour since I'm not able to make atomic update from within ``DocumentPersister`` to collection nested in another collection with `pushAll` strategy (especially not when embedded object containing that collection is yet to be created). Right now I'm thinking about limiting the feature to be available only in top level document (or deeply embedded but only within collections with `atomic*` strategy).

I've been also thinking that `atomic` flag in `set` and `setArray` could adapt and make atomic updates when possible and fall back to old behaviour but I think this can be at least confusing. That said I think that introducing new strategies is ok and even wanted given said limitations.

malarzm:
@blockjon could you run your test against this PR?

blockjon:
Having just quickly read through the code, it appears as if this might do the trick. I will try this today and report back my findings.

blockjon:
@malarzm Looks like your changes fix my problem as seen here: https://github.com/malarzm/mongodb-odm/pull/3

docs/en/reference/collection-strategies.rst
```diff
@@ -45,13 +45,34 @@ strategy, but will first numerically reindex the collection to ensure that it is
...
+
+The ``atomicSet`` strategy uses MongoDB's `$set`_ operator to update the entire
+collection with a single update query. Unlike with ``set`` strategy there will
+be only one query for updating both parent document and collection itself. This
+strategy can be especially useful when dealing with high concurrency and 
+:ref:`versioned documents <annotations_reference_version>`.
+
+.. note::
+
+    Only top level document can contain collection with ``atomicSet`` or
```
jmikola: I'd rephrase as "The `atomicSet` and `atomicSetArray` strategies may only be used for collections on a top-level document."

tests/Doctrine/ODM/MongoDB/Tests/Functional/AtomicSetTest.php
```diff
@@ -0,0 +1,84 @@
...
+    public $surname;
+    
+    /** @ODM\EmbedMany(strategy="atomicSet",targetDocument="Documents\Phonenumber") */
+    public $phonenumbers;
+    
+    public function __construct($name)
+    {
+        $this->name = $name;
+        $this->phonenumbers = new ArrayCollection();
+    }
```
jmikola: I think we can add an additional test here with a nested embed-many relationship. Perhaps: AtomicUser embeds many PhoneBooks, which each embed many Phonenumbers. The top-level PhoneBooks relationship will use atomicSet, and the nested Phonenumbers relationship can use the default strategy.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -235,6 +228,11 @@ public function executeInserts(array $options = array())
                 }
                 $data[$versionMapping['name']] = $nextVersion;
             }
+            
+            $atomicCollectionQuery = $this->getAtomicCollectionUpdateQuery($document);
+            if (!empty($atomicCollectionQuery['$set'])) {
+                $data = array_merge_recursive($data, $atomicCollectionQuery['$set']);
```
jmikola: Why a recursive merge here? I'd like to think that the collection's field name wouldn't conflict with any other mapped fields on the top-level document, so the array keys won't conflict. And if they did conflict (which I'd consider a logical mapping error), a recursive merge might lead to some odd behavior where we're merging data within the collection from two different sources.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -282,6 +280,8 @@ public function executeUpserts(array $options = array())
                 }
                 $data['$set'][$versionMapping['name']] = $nextVersion;
             }
+            
+            $data = array_merge_recursive($data, $this->getAtomicCollectionUpdateQuery($document));
```
jmikola: I suppose the recursive merge here is because `getAtomicCollectionUpdateQuery()` may return `$set` or `$unset` operations, and it keeps the code simpler than doing a non-recursive merge on `$data['$set']` or `$data['$unset']` (and having to ensure that `$data` has those array keys, too)?

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
jmikola: Per my [comment below](https://github.com/doctrine/mongodb-odm/pull/1096/files#r30641740) on `executeUpserts()`, it occurred to me that `getAtomicCollectionUpdateQuery()` might populate `$unset`. Should we have a sanity check on that and raise an exception if so, or is it not worth the trouble?

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
jmikola: I would suggest optimizing this to avoid the unnecessary function call, similar to what you did for `executeInserts()`:

```
$atomicCollectionQuery = $this->getAtomicCollectionUpdateQuery($document);
if ( ! empty($atomicCollectionQuery) {
    $data = array_merge_recursive($data, $atomicCollectionQuery);
}
```

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -375,6 +375,8 @@ public function update($document, array $options = array())
                     $this->class->reflFields[$this->class->versionField]->setValue($document, $nextVersion);
                 }
             }
+            
+            $update = array_merge_recursive($update, $this->getAtomicCollectionUpdateQuery($document));
```
jmikola: Same suggested optimization as https://github.com/doctrine/mongodb-odm/pull/1096/files#r30642646.

blockjon:
Friendly question for you all - This branch fixes a critical problem I'm faced with and I would love to use it sooner rather than later.

Since I'm new to this project, I'd like to get a sense of when you feel as though it would be in a state that you feel I should attempt to integrate into my project. Should I want until this gets merged to master, or until you tag a beta 13, for example?

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -1230,4 +1232,31 @@ private function getClassDiscriminatorValues(ClassMetadata $metadata)
...
+            $collPersister = $this->uow->getCollectionPersister();
+            if ($this->uow->isCollectionScheduledForUpdate($coll)) {
+                $update = array_merge_recursive($update, $collPersister->prepareSetQuery($coll));
+                $this->uow->unscheduleCollectionUpdate($coll);
+                /* Collection can be set for both deletion and update if
+                 * PersistentCollection instance was changed. Since we're dealing
+                 * with collection update in one query we won't need the $unset
+                 */
+                $this->uow->unscheduleCollectionDeletion($coll);
+            } else {
```
jmikola: Would it be safer swap out the `else` for a separate `isCollectionScheduledForDeletion()` check here? I realize `UnitOfWork::scheduleCollectionOwner()`, which populates the array returned by `getScheduledCollections()`, is currently only called from `scheduleCollectionDeletion()` and `scheduleCollectionUpdate()`; however, if that changes in the future, we're looking at a very destructive `else` condition :)

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -1230,4 +1232,31 @@ private function getClassDiscriminatorValues(ClassMetadata $metadata)
...
+            }
+            $collPersister = $this->uow->getCollectionPersister();
+            if ($this->uow->isCollectionScheduledForUpdate($coll)) {
+                $update = array_merge_recursive($update, $collPersister->prepareSetQuery($coll));
+                $this->uow->unscheduleCollectionUpdate($coll);
+                /* Collection can be set for both deletion and update if
+                 * PersistentCollection instance was changed. Since we're dealing
+                 * with collection update in one query we won't need the $unset
+                 */
+                $this->uow->unscheduleCollectionDeletion($coll);
```
jmikola: Is this a redundant safety check? Would it be safer would a collection be scheduled for both update and deletion?

Why would a collection be scheduled for both update and deletion? If that a logic error we should instead check in UnitOfWork? `scheduleCollectionDeletion()` and `scheduleCollectionUpdate()` don't seem to care, but perhaps they should throw if the collection is already scheduled, or one should take precedence. For instance, if something schedules a collection for deletion, perhaps any update schedule should be ignored.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -2752,8 +2799,12 @@ private function scheduleCollectionOwner(PersistentCollection $coll)
             list(, $document, ) = $this->getParentAssociation($document);
             $class = $this->dm->getClassMetadata(get_class($document));
         }
+        $oid = spl_object_hash($document);
 
-        $this->hasScheduledCollections[spl_object_hash($document)] = true;
+        if (!isset($this->hasScheduledCollections[$oid])) {
+            $this->hasScheduledCollections[$oid] = array();
+        }
```
jmikola: Was this conditional required? if `$this->hasScheduledCollections` is only ever null or an array, there should be no problem using the `[]` append operator. See: http://3v4l.org/5VikE

malarzm:
@blockjon idea is to have this feature merged and tag BETA-13 then

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
jmikola: @malarzm mentioned this was due to #1011, which appears to be a separate issue. Perhaps we should add a TODO here to remove this once that issue is resolved, then. Long-term, I don't think ODM should be issuing superfluous `$unset` queries.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
malarzm: The `elseif` was there before but I'll bring it back ;)

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
malarzm: For the record that's exactly reason for doing recursive merge :+1: Will add optimizations

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
jmikola: @jwage: If we make this an `elseif` again, how do you feel about a final `else` condition throwing a LogicException?

jmikola:
LGTM pending test results. @jwage: Do you have a chance to take a look at this within the next day?

Ideally, we'd like to get it included in a new 1.0.0-BETA13 tag this week (to help with doctrine/DoctrineMongoDBBundle#298).

jwage:
:+1:

blockjon:
I AM EXCITED!!!!!!

smolinari:
Nice work gentlemen. :+1: 

Scott

BernardoSilva:
:+1: Nice work.

malarzm:
@blockjon BETA13 is now released :)

blockjon:
@malarzm This is fantastic news thank you!!!!

daniphp:
@malarzm - I think you need to implement atomicSet as default where possible. I just found a bug in production that happened in only 1/100.000 cases and it was driving me nuts for weeks before I found the problem. Using pushAll as default is really something unexpected for me as a developer.

blockjon:
@daniphp +1 on your comments there... I had the same feedback here https://github.com/doctrine/mongodb-odm/pull/782#issuecomment-113642286

PR #1097, opened 2015-05-16
Add syntax checker for contributors.

BernardoSilva:
Did not run the tool against whole project to avoid creating conflict with all existing Pull requests.
But from now on all contributors can have their code automatically formatted according to PSR2 standard.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ tests/Proxies/*
 tests/Hydrators/*
 phpunit.xml
 vendor/
+/bin/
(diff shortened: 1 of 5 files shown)


alcaeus:
Using phpcs to enforce code quality is a good idea, but I don't like the idea of forcing a pre-commit hook on people on the client side. Besides that, doctrine has their own coding standard (see https://github.com/doctrine/coding-standard) which requires a space before a negation operator, so we should use that.

I believe we should use a more complete tool like scrutinizer to monitor these things and make pull requests fail if they introduce errors instead of relying client-side validation.

PR #1099, opened 2015-05-20
[Tests] Added typehints to $this->dm and $this->uow

malarzm:
For the sake of auto completion :)


BernardoSilva:
:+1: That's helpfull! 
I've now started to extend all my project tests that require `$dm` from this class. Do you recommend doing that?

malarzm:
@BernardoSilva it depends :) for instance I usually roll my own parent class for tests where I have helper methods mocking various stuff but not setting global `setUp` as I don't really need `DocumentManager` in lots of tests

alcaeus:
:+1:

malarzm:
Merged manually in https://github.com/doctrine/mongodb-odm/commit/9ab9301dd48692b81c9d7b08140a55eeafe6691e

PR #1100, opened 2015-05-20
Fix handling of nested collections for atomic strategies

jmikola:
Related to new functionality in #1096.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/CollectionPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/CollectionPersister.php
@@ -162,7 +162,7 @@ class CollectionPersister
         $pb = $this->pb;
 
         $callback = isset($mapping['embedded'])
-            ? function($v) use ($pb, $mapping) { return $pb->prepareEmbeddedDocumentValue($mapping, $v, in_array($mapping['strategy'], array('atomicSet', 'atomicSetArray'))); }
+            ? function($v) use ($pb, $mapping) { return $pb->prepareEmbeddedDocumentValue($mapping, $v, ($mapping['strategy'] === 'atomicSet' || $mapping['strategy'] === 'atomicSetArray')); }
             : function($v) use ($pb, $mapping) { return $pb->prepareReferencedDocumentValue($mapping, $v); };
 
         $setData = $coll->map($callback)->toArray();
--- a/lib/Doctrine/ODM/MongoDB/Persisters/CollectionPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/CollectionPersister.php
@@ -160,9 +160,9 @@ class CollectionPersister
         list($propertyPath, ) = $this->getPathAndParent($coll);
 
         $pb = $this->pb;
-
+        
         $callback = isset($mapping['embedded'])
-            ? function($v) use ($pb, $mapping) { return $pb->prepareEmbeddedDocumentValue($mapping, $v); }
+            ? function($v) use ($pb, $mapping) { return $pb->prepareEmbeddedDocumentValue($mapping, $v, in_array($mapping['strategy'], array('atomicSet', 'atomicSetArray'))); }
             : function($v) use ($pb, $mapping) { return $pb->prepareReferencedDocumentValue($mapping, $v); };
 
         $setData = $coll->map($callback)->toArray();
--- a/lib/Doctrine/ODM/MongoDB/Persisters/CollectionPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/CollectionPersister.php
@@ -160,7 +160,7 @@ class CollectionPersister
         list($propertyPath, ) = $this->getPathAndParent($coll);
 
         $pb = $this->pb;
-        
+
         $callback = isset($mapping['embedded'])
             ? function($v) use ($pb, $mapping) { return $pb->prepareEmbeddedDocumentValue($mapping, $v, in_array($mapping['strategy'], array('atomicSet', 'atomicSetArray'))); }
             : function($v) use ($pb, $mapping) { return $pb->prepareReferencedDocumentValue($mapping, $v); };
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -685,7 +685,7 @@ class DocumentPersister
 
             // no custom sort so add the references right now in the order they are embedded
             if ( ! $sorted) {
-                if ($mapping['strategy'] === 'set') {
+                if ($mapping['strategy'] === 'set' || $mapping['strategy'] === 'atomicSet') {
                     $collection->set($key, $reference);
                 } else {
                     $collection->add($reference);
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -383,7 +383,6 @@ class DocumentPersister
             if ( ! empty($atomicCollectionQuery)) {
                 $update = array_merge_recursive($update, $atomicCollectionQuery);
             }
-
             /* We got here because the document has one or more related
              * PersistentCollections to be committed later; however, if the
              * document is not versioned then there is nothing left to do.
--- a/lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
@@ -449,12 +449,14 @@ class PersistenceBuilder
      *
      * @param array $mapping
      * @param object $document
+     * @param boolean $includeNestedCollections
      * @return array|object|null
+     * @throws \InvalidArgumentException if the mapping is neither embedded nor reference
      */
-    public function prepareAssociatedDocumentValue(array $mapping, $document)
+    public function prepareAssociatedDocumentValue(array $mapping, $document, $includeNestedCollections = false)
     {
         if (isset($mapping['embedded'])) {
-            return $this->prepareEmbeddedDocumentValue($mapping, $document);
+            return $this->prepareEmbeddedDocumentValue($mapping, $document, $includeNestedCollections);
         }
 
         if (isset($mapping['reference'])) {
--- a/lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
@@ -332,7 +332,7 @@ class PersistenceBuilder
      * @param object $embeddedDocument
      * @return array|object
      */
-    public function prepareEmbeddedDocumentValue(array $embeddedMapping, $embeddedDocument)
+    public function prepareEmbeddedDocumentValue(array $embeddedMapping, $embeddedDocument, $includeNestedCollections = false)
     {
         $embeddedDocumentValue = array();
         $class = $this->dm->getClassMetadata(get_class($embeddedDocument));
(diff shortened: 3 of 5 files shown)


jmikola:
Executing the modified test, ODM inserts the following on the first flush:

```json
{
    "000000001f8298960000000023533066": {
        "_id": {
            "$id": "555cea55e84df15b498b4569"
        },
        "name": "Maciej",
        "phonebooks": {
            "private": {
                "title": "Private",
                "phonenumbers": [
                    {
                        "phonenumber": "12345678"
                    }
                ]
            }
        }
    }
}
```

The second flush (after modification) results in the following update:

```json
{
    "$set": {
        "phonebooks": {
            "private": {
                "title": "Private"
            },
            "public": {
                "title": "Public",
                "phonenumbers": [
                    {
                        "phonenumber": "10203040"
                    }
                ]
            }
        }
    }
}
```

lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
```diff
@@ -363,15 +363,21 @@ public function prepareEmbeddedDocumentValue(array $embeddedMapping, $embeddedDo
...
+                        if ($includeNestedCollections && $rawValue instanceof PersistentCollection) {
+                            $this->uow->unscheduleCollectionDeletion($rawValue);
+                            $this->uow->unscheduleCollectionUpdate($rawValue);
+                        }
 
                         $pb = $this;
-                        $value = $rawValue->map(function($v) use ($pb, $mapping) {
-                            return $pb->prepareAssociatedDocumentValue($mapping, $v);
+                        $value = $rawValue->map(function($v) use ($pb, $mapping, $includeNestedCollections) {
+                            return $pb->prepareAssociatedDocumentValue($mapping, $v, $includeNestedCollections);
```
jmikola: @malarzm: I don't believe `prepareAssociatedDocumentValue()` was updated to take a boolean argument. I assume it should also take `$includeNestedCollections` and then conditionally pass it to `prepareEmbeddedDocumentValue()` (but not `prepareReferencedDocumentValue()`).

lib/Doctrine/ODM/MongoDB/Persisters/CollectionPersister.php
```diff
@@ -162,7 +162,7 @@ public function prepareSetQuery(PersistentCollection $coll)
         $pb = $this->pb;
 
         $callback = isset($mapping['embedded'])
-            ? function($v) use ($pb, $mapping) { return $pb->prepareEmbeddedDocumentValue($mapping, $v); }
+            ? function($v) use ($pb, $mapping) { return $pb->prepareEmbeddedDocumentValue($mapping, $v, ($mapping['strategy'] === 'atomicSet' || $mapping['strategy'] === 'atomicSetArray')); }
```
jmikola: @malarzm: I swapped out your `in_array()` for two equality checks to avoid an extra function call (and since two checks were not that much more verbose than `in_array()` -- unlike times where we're checking 4+ values).

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -383,7 +383,6 @@ public function update($document, array $options = array())
             if ( ! empty($atomicCollectionQuery)) {
                 $update = array_merge_recursive($update, $atomicCollectionQuery);
             }
-
```
jmikola: @malarzm: Might want to check if your editor is leaving trailing whitespace behind. This showed up a bit :)

lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
```diff
@@ -328,11 +328,19 @@ public function prepareReferencedDocumentValue(array $referenceMapping, $documen
      * returned if the document is empty, to ensure that a BSON object will be
      * stored in lieu of an array.
      *
+     * If $includeNestedCollections is true, nested collections will be included
+     * in this prepared value and the option will cascade to all embedded
+     * associations. If any nested PersistentCollections (embed or reference)
+     * within this value were previously scheduled for deletion or update, they
+     * will also be unscheduled.
```
jmikola: This is my understanding of the logic. Let me know if I'm mistaken, though.

lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
```diff
@@ -357,21 +365,29 @@ public function prepareEmbeddedDocumentValue(array $embeddedMapping, $embeddedDo
 
                     case ClassMetadata::EMBED_ONE:
                     case ClassMetadata::REFERENCE_ONE:
-                        $value = $this->prepareAssociatedDocumentValue($mapping, $rawValue);
+                        // Nested collections should only be included for embedded relationships
+                        $value = $this->prepareAssociatedDocumentValue($mapping, $rawValue, $includeNestedCollections && isset($mapping['embedded']));
```
jmikola: Per my commit message, I think we want to relay `$includeNestedCollections` for the following case: an atomic, top-level embed-many association contains docs, which in turn contain an embed-one of a doc, which in turn has an embed-many association, and so on. Before this addition, `$includeNestedCollections` wouldn't make it past the second-level embed-one association.

 * "phones" is a top-level embed-many collection 
 * "

lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
```diff
@@ -357,21 +365,29 @@ public function prepareEmbeddedDocumentValue(array $embeddedMapping, $embeddedDo
...
                         break;
 
                     case ClassMetadata::EMBED_MANY:
                     case ClassMetadata::REFERENCE_MANY:
                         // Skip PersistentCollections already scheduled for deletion/update
-                        if ($rawValue instanceof PersistentCollection &&
+                        if ( ! $includeNestedCollections && $rawValue instanceof PersistentCollection &&
```
jmikola: I realize it's not PSR-2 compliant (it will likely be addressed in a future PSR), but we typically wrap leading negations with spaces for readability.

lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
```diff
@@ -357,21 +365,29 @@ public function prepareEmbeddedDocumentValue(array $embeddedMapping, $embeddedDo
...
+                            $this->uow->unscheduleCollectionDeletion($rawValue);
+                            $this->uow->unscheduleCollectionUpdate($rawValue);
+                        }
+
                         $pb = $this;
-                        $value = $rawValue->map(function($v) use ($pb, $mapping) {
-                            return $pb->prepareAssociatedDocumentValue($mapping, $v);
+                        $value = $rawValue->map(function($v) use ($pb, $mapping, $includeNestedCollections) {
+                            // Nested collections should only be included for embedded relationships
+                            return $pb->prepareAssociatedDocumentValue($mapping, $v, $includeNestedCollections && isset($mapping['embedded']));
```
jmikola: Similar to the above embed-one case (and per my commit message), I think we only want to relay the option when dealing with embedded associations. `prepareAssociatedDocumentValue()` doesn't handle `$includeNestedCollections` for references anyway, and we wouldn't want to process them deeper even if it did (since that data lives in another collection).

lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
```diff
@@ -443,12 +459,14 @@ public function prepareEmbeddedDocumentValue(array $embeddedMapping, $embeddedDo
...
+     * @param boolean $includeNestedCollections
      * @return array|object|null
+     * @throws \InvalidArgumentException if the mapping is neither embedded nor reference
      */
-    public function prepareAssociatedDocumentValue(array $mapping, $document)
+    public function prepareAssociatedDocumentValue(array $mapping, $document, $includeNestedCollections = false)
     {
         if (isset($mapping['embedded'])) {
-            return $this->prepareEmbeddedDocumentValue($mapping, $document);
+            return $this->prepareEmbeddedDocumentValue($mapping, $document, $includeNestedCollections);
```
jmikola: @malarzm: Added this as I think it was just an oversight in your previous commit where you were passing `$includeNestedCollections` to this method from `prepareEmbeddedDocumentValue()`.

lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
jmikola: Perhaps we want a test case for this embed-many/embed-one/embed-many sequence? The current test is just embed-many/embed-many.

lib/Doctrine/ODM/MongoDB/Persisters/CollectionPersister.php
malarzm: :+1: btw I wanted to introduce some strategy helper as right now adding new strategies is cumbersome (and some if statement would be more verbose then) but that would be another static function call again, what do you think?

lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
malarzm: In line with intents :)

lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
malarzm: Good catch, I'll add a test for that

lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
malarzm: Sure, will keep that in mind :+1:

malarzm:
@jmikola added test for deeply embedded embeds, nothing leaks to `CollectionPersister` after your `embedOne` changes :+1:

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
malarzm: Found one and turned on, thanks for noticing ;)

lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
BernardoSilva: @jmikola does that mean Doctrine has a custom code styling instead of using PSR2?
Is this not useful then? https://github.com/doctrine/mongodb-odm/pull/1097

lib/Doctrine/ODM/MongoDB/Persisters/CollectionPersister.php
malarzm: To clarify intent: [this if](https://github.com/doctrine/mongodb-odm/blob/5a202c532c981292613633ea95f228e6e51404b0/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php#L657), [this one I forgot to update before](https://github.com/doctrine/mongodb-odm/blob/5a202c532c981292613633ea95f228e6e51404b0/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php#L690) and [this one](https://github.com/doctrine/mongodb-odm/blob/5a202c532c981292613633ea95f228e6e51404b0/lib/Doctrine/ODM/MongoDB/PersistentCollection.php#L191) needs to be in line for everything to work correctly and using `CollectionStrategyHelper::handlesAssociative($strategy)` would be way easier than remembering to change all three places.

lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
jmikola: @BernardoSilva: we're mostly PSR-2, but there are some differences. I tried creating a [PHP Coding Standards Fixer](http://cs.sensiolabs.org/) profile a while back but didn't have luck. Not sure if @beberlei or @guilhermeblanco have anything in place for the ORM library.

beberlei: We don't to be honest. @deeky666 worked on something before, but we haven't fully rolled it out across the organization.

jmikola:
@malarzm: New test case LGTM!

PR #1104, opened 2015-05-23
Added failing test case for nested collection update

malarzm:
Nested collection with default strategy fails to update when used in collection using `set` or `setArray` strategy.

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/NestedCollectionsTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/NestedCollectionsTest.php
@@ -1,0 +1,102 @@
+<?php
+
+namespace Doctrine\ODM\MongoDB\Tests\Functional;
+
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
+use Documents\Phonebook;
+use Documents\Phonenumber;
+
+class NestedCollectionsTest extends \Doctrine\ODM\MongoDB\Tests\BaseTest
+{
+    /**
+     * @dataProvider provideStrategy
+     */
+    public function testStrategy($field)
+    {
+        $doc = new DocWithNestedCollections();
+        $privateBook = new Phonebook('Private');
+        $privateBook->addPhonenumber(new Phonenumber('12345678'));
+        $doc->{$field}[] = $privateBook;
+        $this->dm->persist($doc);
+        $this->dm->flush();
+        $this->dm->clear();
+
+        $doc = $this->dm->getRepository(get_class($doc))->find($doc->id);
+        $this->assertCount(1, $doc->{$field});
+        $privateBook = $doc->{$field}[0];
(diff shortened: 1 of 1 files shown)


malarzm:
The problem lies in the way collection with strategy `set` or `setArray` is updated as existing collections are excluded from `$set` query (as in one with title "Private" which at the moment of update had `phonenumbers` array with 1 element. Entry with title "Public" is just being created so the query is correct):

```
{
    "$set": {
        "set": [
            { "title": "Private" },
            { "title": "Public", "phonenumbers": [ { "phonenumber": "10203040" } ] }
        ]
    }
}
{
    "$pushAll": {
        "set.0.phonenumbers": [ { "phonenumber": "87654321" } ] 
    } 
}
```

Things that can be done that come to my mind (not sorted in any way)

* include nested collection's snapshot in the `$set` query
* use dot notation to update fields (this would include changing way the collection is saved)
* forbid using nested collections with strategies other than `set` or `setArray` within collections using `set` or `setArray` strategies

jmikola:
@malarzm: Was this a regression introduced in 1.0.0-BETA13 (for the non-atomic collections)? Is there a related bug report, or did you come across this on your own?

malarzm:
@jmikola this is not a regression, I was checking older tags as well and the issue exists. I came across this when fixing your bug report about nested collection atomic update, I think I was even mentioning this on IRC back then :)

jmikola:
> include nested collection's snapshot in the $set query

If a parent collection is using a set strategy, we're effectively overwriting all of its internal data, so I don't see any reason to use push on any nested collections. I would be in favor of this.

> use dot notation to update fields (this would include changing way the collection is saved)

I don't follow that entirely, but it sounds like it'd just be splitting a single `$set` into multiple writes, which just exposes us to a race condition if the array offsets change between those two writes. I think this is inferior to a single set.

> forbid using nested collections with strategies other than set or setArray within collections using set or setArray strategies

I don't think we'd want to go that far, since someone may have a shared embedded document class and want to use the push strategy by default. If a class within a collection using the set strategy happens to embed that document, we shouldn't blow up. That's also not something we can reliable catch before runtime, since the user could be using discriminators to store some dynamic class name (e.g. an embed-many of arbitrary embedded document classes).

----

@jwage: Now that we have `atomicSet`, is there any practical reason why users would want to use the classic `set` strategy? The older `set` strategy still writes the *entire* collection out, does it not? Or am I misunderstanding and does it only update individual array positions via the dot syntax.

jmikola:
@malarzm: I believe you answered somewhere in #1155 (or perhaps IRC) that the original `set` strategy still served a purpose since it could be used within nested collections, where some a parent collection was using `pushAll` or `addToSet`. That probably answers the question I had for @jwage above.

malarzm:
Yes I did, it's this one: https://github.com/doctrine/mongodb-odm/pull/1155#discussion_r33877234 :)

malarzm:
#1155 was merged manually in https://github.com/doctrine/mongodb-odm/commit/5b0ff1c5d4a19594b778352e9bb0fc6a54ed2152 (and test from here was included in it)

PR #1105, opened 2015-05-26
Cache documents with single type inheritance to array by parent classes

notrix:
The issue was that if I get repository using abstract class name of inherited classes and call method ->find($id) the result gets cached by inherited class name, not by abstract. So doing this multiple times persister always queries db for result. Later it finds that it has requested object, but request is made.


tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH788Test.php
```diff
@@ -20,14 +20,6 @@ public function testDocumentWithDiscriminatorMap()
         $this->dm->flush();
         $this->dm->clear();
 
-        $doc = $this->dm->find(get_class($listed), $listed->id);
```
malarzm: @notrix why this is moved?

tests/Doctrine/ODM/MongoDB/Tests/Functional/DiscriminatorsArrayCacheTest.php
```diff
@@ -0,0 +1,137 @@
...
+        $this->assertNotEmpty($result2);
+        $this->assertEquals(1, $this->listener->findCount);
+    }
+
+    public function testFindDocumentWithPerClassCollectionInheritance()
+    {
+        $id = 123;
+
+        $this->dm->persist(new DocumentChildPerClass1($id));
+        $this->dm->persist(new DocumentChildPerClass2($id));
```
malarzm: Why are we creating different object with same id as previous one?

tests/Doctrine/ODM/MongoDB/Tests/Functional/DiscriminatorsArrayCacheTest.php
```diff
@@ -0,0 +1,137 @@
...
+use Doctrine\ODM\MongoDB\DocumentManager;
+use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
+
+/**
+ * @author Vaidas Lažauskas <vaidas@notrix.lt>
+ */
+class DiscriminatorsArrayCacheTest extends \Doctrine\ODM\MongoDB\Tests\BaseTest
+{
+    /** @var DocumentManager */
+    protected $dm;
```
malarzm: I believe there is no need to bring `$dm` once again since it's already protected in parent class?

tests/Doctrine/ODM/MongoDB/Tests/Functional/DiscriminatorsArrayCacheTest.php
notrix: Maybe it's not necessary here, but the idea was to make sure the inheritance is "per class" type and allows to save same id.

tests/Doctrine/ODM/MongoDB/Tests/Functional/DiscriminatorsArrayCacheTest.php
notrix: I have added dm here for php doc to enable auto-compleete

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH788Test.php
notrix: I have moved this code to fix failing test as the logic changes and old test got cached result. Changing code order has fixed failing test.

tests/Doctrine/ODM/MongoDB/Tests/Functional/DiscriminatorsArrayCacheTest.php
malarzm: Ah sorry for an oversight, it indeed has `@ODM\InheritanceType("COLLECTION_PER_CLASS")` I was mislead by first `SINGLE_COLLECTION` :)

tests/Doctrine/ODM/MongoDB/Tests/Functional/DiscriminatorsArrayCacheTest.php
malarzm: This can be removed now then, php docs are already available in parent class :)

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH788Test.php
malarzm: Sorry for long response time, I was out of town. It doesn't look right that `$doc = $this->dm->find(get_class($listed), $unlisted->id);` suddenly started finding document, cached or not, it shouldn't be there, do you agree?

malarzm:
ping @notrix

Issue #1106, opened 2015-05-26
getTypeClass not working on Collection


wodka:
Notice: Undefined property: Doctrine\ODM\MongoDB\PersistentCollection::$typeClass

malarzm:
@wodka how did you manage to get such an error?

wodka:
For SonataAdmin Bundle there is a Pull Request that contained a reference to Doctrine 2.4 code:
https://github.com/sonata-project/SonataAdminBundle/pull/2985/files#diff-5ce9b658fbcaa128b4326ce1cc07bd14R123

While I was trying to fix that to work with odm I realised the same funktion exists there:

**Doctrine\ODM\MongoDB\PersistentCollection:** *typeClass is never set*
```php
    public function getTypeClass()
    {
        return $this->typeClass;
    }
```

**Doctrine\ORM\PersistentCollection:** *works*
```php

    /**
     * Creates a new persistent collection.
     *
     * @param EntityManagerInterface $em         The EntityManager the collection will be associated with.
     * @param ClassMetadata          $class      The class descriptor of the entity type of this collection.
     * @param Collection             $collection The collection elements.
     */
    public function __construct(EntityManagerInterface $em, $class, Collection $collection)
    {
        $this->collection  = $collection;
        $this->em          = $em;
        $this->typeClass   = $class;
        $this->initialized = true;
    }

    /**
     * @return Mapping\ClassMetadata
     */
    public function getTypeClass()
    {
        return $this->typeClass;
    }
```

malarzm:
For the quickest solution: is it possible to add check for ODM's PersistentCollection in Sonata and use `getMapping()` method instead? Or it would be cool if you'd be able to send PR here fixing the issue :)

wodka:
Hi, 

I'll try to fix it - I already have a solution based on the SonataAdmin Configuration (working)

can you point me in a direction where the persistent collection is usually created?
thx :)

malarzm:
@wodka I'd check `UnitOfWork` and `HydratorFactory` for sure :)

malarzm:
PR: #1109

malarzm:
#1109 is merged manually in https://github.com/doctrine/mongodb-odm/commit/5b30e0f7a677c8a7494d82176bb8a747f4e82fb9

PR #1109, opened 2015-05-29
implementation for https://github.com/doctrine/mongodb-odm/issues/1106

wodka:
Set typeClass once owner information is passed to collection

-> this does not require any change outside of the Persistent Collection
-> also fixed usage of ODM in https://github.com/sonata-project/SonataAdminBundle/pull/2985

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/PersistentCollection.php
+++ b/lib/Doctrine/ODM/MongoDB/PersistentCollection.php
@@ -260,7 +260,10 @@ class PersistentCollection implements BaseCollection
     {
         $this->owner = $document;
         $this->mapping = $mapping;
-        $this->typeClass = $this->dm->getClassMetadata($this->mapping['targetDocument']);
+
+        if (!empty($this->mapping['targetDocument'])) {
+            $this->typeClass = $this->dm->getClassMetadata($this->mapping['targetDocument']);
+        }
     }
 
     /**
@@ -346,9 +349,14 @@ class PersistentCollection implements BaseCollection
 
     /**
      * @return ClassMetadata
+     * @throws MongoDBException
      */
     public function getTypeClass()
     {
+        if (empty($this->typeClass)) {
+            throw new MongoDBException('targetDocument option is required for typeClass');
+        }
+
         return $this->typeClass;
     }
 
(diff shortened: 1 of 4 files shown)


malarzm:
@wodka: Good approach! You need to check if `targetClass` is set as ODM doesn't require that property, I think it's best to throw an exception saying that `getTypeClass()` is not available for mixed document references. Given that I'd also add a test :)

wodka:
well, looks also better on travis now :)

lib/Doctrine/ODM/MongoDB/PersistentCollection.php
```diff
@@ -255,6 +260,10 @@ public function setOwner($document, array $mapping)
     {
         $this->owner = $document;
         $this->mapping = $mapping;
+
+        if (!empty($this->mapping['targetDocument'])) {
```
malarzm: can you change this to `if ( ! empty($this->mapping['targetDocument'])) {` (for the sake of whitespaces)?

lib/Doctrine/ODM/MongoDB/PersistentCollection.php
```diff
@@ -330,13 +339,24 @@ public function getOwner()
...
     }
 
+    /**
+     * @return ClassMetadata
+     * @throws MongoDBException
+     */
     public function getTypeClass()
     {
+        if (empty($this->typeClass)) {
+            throw new MongoDBException('targetDocument option is required for typeClass');
```
malarzm: Can we rephrase this to "Specifying targetDocument is required for the ClassMetadata to be obtained."?

malarzm:
@wodka I was about to write about my notes but I see you've already adjusted that :) Could you also add a test case if the exception is thrown when `targetDocument` is not set?

wodka:
Done :)

malarzm:
Merged manually in https://github.com/doctrine/mongodb-odm/commit/5b30e0f7a677c8a7494d82176bb8a747f4e82fb9 - thanks @wodka!

jmikola:
@wodka, @malarzm: Was this method present in an interface, or is it just something in ORM that we're implementing for compatibility?

Since this is a new feature, it may be worth adding a changelog entry for it.

jmikola:
Scratch that. I thought this was adding a new method for some reason. It is definitely just a bug fix :)

PR #1112, opened 2015-05-31
Prime references only from eagerCursor

alcaeus:
This PR fixes a potential performance issue introduced with #1068. When using reference primers with non-eager cursors, the query would be executed multiple times, sometimes producing inconsistent results.
Using reference priming now implies using an eager cursor. Using eagerCursor = false and reference primers will result in an error.

Unfortunately, this re-introduces an issue described in #520, where calling ```getSingleResult()``` on a cursor with reference primers will always prime the entire result set.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/EagerCursor.php
+++ b/lib/Doctrine/ODM/MongoDB/EagerCursor.php
@@ -232,14 +232,13 @@ class EagerCursor extends BaseEagerCursor
      */
     protected function primeReferences()
     {
-        if ($this->referencesPrimed || !$this->hydrate) {
+        if ($this->referencesPrimed || !$this->hydrate || empty($this->primers)) {
             return;
         }
 
         $this->referencesPrimed = true;
 
         foreach ($this->primers as $fieldName => $primer) {
-            $this->rewind();
             $primer = is_callable($primer) ? $primer : null;
             $this->referencePrimer->primeReferences($this->class, $this, $fieldName, $this->unitOfWorkHints, $primer);
         }
--- a/lib/Doctrine/ODM/MongoDB/Query/Builder.php
+++ b/lib/Doctrine/ODM/MongoDB/Query/Builder.php
@@ -147,9 +147,27 @@ class Builder extends \Doctrine\MongoDB\Query\Builder
             throw new \InvalidArgumentException('$primer is not a boolean or callable');
         }
 
+        if (array_key_exists('eagerCursor', $this->query) && !$this->query['eagerCursor']) {
+            throw new \BadMethodCallException("Can't call prime() when setting eagerCursor to false");
+        }
+
+        $this->eagerCursor(true);
         $this->primers[$this->currentField] = $primer;
         return $this;
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function eagerCursor($bool = true)
+    {
+        if ( ! $bool && ! empty($this->primers)) {
+            throw new \BadMethodCallException("Can't set eagerCursor to false when using reference primers");
+        }
+
+        return parent::eagerCursor($bool);
+    }
+
 
     /**
      * @param bool $bool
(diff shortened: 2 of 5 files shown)


lib/Doctrine/ODM/MongoDB/Query/Builder.php
```diff
@@ -147,11 +147,29 @@ public function prime($primer = true)
...
     /**
+     * {@inheritdoc}
+     */
+    public function eagerCursor($bool = true)
+    {
+        if ( ! $bool && $this->primers !== array()) {
+            throw new \BadMethodCallException("Can't set eagerCursor to false when using reference primers");
+        }
+
+        return parent::eagerCursor($bool);
```
malarzm: Shouldn't this be `parent::eagerCursor($bool); return $this;` ?

jmikola:
Making a note that we should follow up on #1112 later on to see if this actually resolves the increased query load.

lib/Doctrine/ODM/MongoDB/Query/Builder.php
alcaeus: ```parent::eagerCursor``` already has return $this, so I can just return that again since it's the same object. If you'd rather have it different for legibility I'll change it.

malarzm: Ah nvm, just realized we're only extending original builder :+1:

lib/Doctrine/ODM/MongoDB/Query/Builder.php
```diff
@@ -147,11 +147,29 @@ public function prime($primer = true)
...
         $this->primers[$this->currentField] = $primer;
         return $this;
     }
 
     /**
+     * {@inheritdoc}
+     */
+    public function eagerCursor($bool = true)
+    {
+        if ( ! $bool && $this->primers !== array()) {
```
jmikola: Splitting hairs, but `! empty($this->primers)` may be more readable here.

lib/Doctrine/ODM/MongoDB/EagerCursor.php
```diff
@@ -232,14 +232,13 @@ public function enableReferencePriming(array $primers, ReferencePrimer $referenc
      */
     protected function primeReferences()
     {
-        if ($this->referencesPrimed || !$this->hydrate) {
+        if ($this->referencesPrimed || !$this->hydrate || $this->primers === array()) {
```
jmikola: `empty($this->primers)` for readability.

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH520Test.php
```diff
@@ -66,7 +66,7 @@ public function testPrimeWithGetSingleResultWillPrimeEntireResultSet()
         $result = $query->getSingleResult();
 
         $this->assertContains($document2->id, $primedIds);
-        $this->assertNotContains($document4->id, $primedIds, 'Only the dataset being fetched should be primed');
+        $this->assertContains($document4->id, $primedIds);
```
jmikola: I think the comment above needs some rewriting as well. It currently says:

> If priming is configured, references among all documents in the result set will be primed -- not simply those in the first document.

Based on these changes (priming implying EagerCursor), we should rewrite that as:

> Since priming implies an EagerCursor, all documents will be fetched and all of their references will be primed (not simply the first document and its references). As a result, users that wish to combine `getSingleResult()` with priming should also specify a limit on their queries.

jmikola:
LGTM with the changes requested above.

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH520Test.php
jmikola: Also, it'd be a good idea to invert the assertion comment here instead of removing it entirely :)

PR #1113, opened 2015-06-02
Functional test to exhibit bug: strategy="atomicSet" does not atomically update a database record

blockjon:
Beta13 introduced @malarzm 's https://github.com/doctrine/mongodb-odm/pull/1096 which helps make Doctrine ODM atomically insert and update records.

I was just checking my application and in which I have been experiencing data loss & corruption but I see the problem is still not fixed, even with atomicSet enabled using beta13.

I believe the problem may be that the solution introduced in https://github.com/doctrine/mongodb-odm/pull/1096 does not correctly handle embedMany's that are nested inside each other when changes are made to the embedded collection which is 2 levels down.

In this PR, I demonstrate this problem by creating a Book object which has 1 Chapter .. and where the chapter has 1 Page.  (book->chapters->pages) When I insert the object using strategy="atomicSet", everything is saved in a single atomic operation - this is very very good.

However, when I load the object from the database and try to add a new Page to the chapter, Doctrine executes two separate queries. For me, this means continued data loss & corruption.

Here's how to reproduce the failing test:
```
phpunit --filter testAtomicSetWorksAsExpected
```
output:
```
PHPUnit 4.1.6 by Sebastian Bergmann.

Configuration read from /mongodb-odm/phpunit.xml

F

Time: 1.44 seconds, Memory: 21.25Mb

There was 1 failure:

1) Doctrine\ODM\MongoDB\Tests\Functional\VersionTest::testAtomicSetWorksAsExpected
Despite use of atomicSet, 2 queries were detected when flushing this change. We expected only one query.
Failed asserting that 2 matches expected 1.

/mongodb-odm/tests/Doctrine/ODM/MongoDB/Tests/Functional/VersionTest.php:77

FAILURES!
Tests: 1, Assertions: 2, Failures: 1.

```

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/VersionTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/VersionTest.php
@@ -8,37 +8,37 @@ use Doctrine\Common\Collections\ArrayCollection as ArrayCollection;
 
 class VersionTest extends \Doctrine\ODM\MongoDB\Tests\BaseTest
 {
-//    public function testVersioningWhenManipulatingEmbedMany()
-//    {
-//        $expectedVersion = 1;
-//        $doc = new VersionedDocument();
-//        $doc->name = 'test';
-//        $doc->embedMany[] = new VersionedEmbeddedDocument('embed 1');
-//        $doc->embedMany[] = new VersionedEmbeddedDocument('embed 2');
-//        $this->dm->persist($doc);
-//        $this->dm->flush();
-//        $this->assertEquals($expectedVersion++, $doc->version);
-//
-//        $doc->embedMany[] = new VersionedEmbeddedDocument('embed 3');
-//        $this->dm->flush();
-//        $this->assertEquals($expectedVersion++, $doc->version);
-//
-//        $doc->embedMany[0]->embedMany[] = new VersionedEmbeddedDocument('deeply embed 1');
-//        $this->dm->flush();
-//        $this->assertEquals($expectedVersion++, $doc->version);
-//
-//        unset($doc->embedMany[1]);
-//        $this->dm->flush();
-//        $this->assertEquals($expectedVersion++, $doc->version);
-//
-//        $doc->embedMany->clear();
-//        $this->dm->flush();
-//        $this->assertEquals($expectedVersion++, $doc->version);
-//
-//        $doc->embedMany = null;
-//        $this->dm->flush();
-//        $this->assertEquals($expectedVersion++, $doc->version);
-//    }
+    public function testVersioningWhenManipulatingEmbedMany()
+    {
+        $expectedVersion = 1;
+        $doc = new VersionedDocument();
+        $doc->name = 'test';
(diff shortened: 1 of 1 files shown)


tests/Doctrine/ODM/MongoDB/Tests/Functional/VersionTest.php
```diff
@@ -20,23 +22,76 @@ public function testVersioningWhenManipulatingEmbedMany()
...
+
+        $id = $book->id;
+
+        // Simulate another PHP request which loads this record and tries to add an embedded document two levels deep...
+        $dm->clear();
+        $book = $dm->getRepository('Doctrine\ODM\MongoDB\Tests\Functional\Book')->findOneBy(array('_id' => $id));
+
+        // Now we add a new "page" to the only chapter in this book.
+        $firstChapter = $book->getChapters()->first();
+        $firstChapter->addPage(new Page(2));
```
blockjon: @malarzm Here is the example of how adding a new element to an embedded collection which is 2 levels down leads to a non-atomic update. Do you have any ideas on how I can avoid the non-atomic update that I am seeing?

malarzm:
@blockjon for reasons unknown to me yet your test passes if the parent document is not versioned

malarzm:
Ok, I have found what goes wrong but I'm still wondering what versioning has to do with that

malarzm:
@blockjon So summing up yes, this is my oversight that deeply nested collection is excluded from update. Most probably I will be able to fix it by tomorrow, or if you are willing to do this on your own the problem lies [here](https://github.com/doctrine/mongodb-odm/blob/5b30e0f7a677c8a7494d82176bb8a747f4e82fb9/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php#L1254-L1257) as we should check all the way up to top level document if the collection isn't deeply embedded within atomic one.

blockjon:
@malarzm Wow thank you for responding so quickly to this. Since you know the ins and outs of this far better than me, I would prefer to wait and have you triage if that's ok.

malarzm:
@blockjon you can also try calling `$this->uow->scheduleCollectionUpdate($book->getChapters());` by yourself if you know that collection within atomic collection neds to be updated. It's dirty but should be working, at least the test is passing :)

blockjon:
@malarzm Thanks for your idea above.. the app I'm hacking on has too many cases where that would need to be done... I think I would prefer to opt for something that handles this automagically.

malarzm:
@blockjon I'm sorry but I'm still fighting with dragons and releases and I'm not sure if I will be able to fix this today (and because I'm out of town tomorrow, Friday is first possible day I will have a computer nearby)

blockjon:
@malarzm Thanks for the update. I've waited this long I can wait until Monday for sure. Basically I'm seeing about 1 terrible data corruption per week at my current levels. So, worst case I get one this weekend but if you can possibly get this fixed, you will instantly make my life better after Monday. Thank you for all of you attention with this so far!!!

jmikola:
@blockjon: We copied your test case into #1118, which is now fixed and merged.

PR #1116, opened 2015-06-03
Call initialize() when trying to add() to a PersistentCollection in order to avoid data loss

blockjon:
Doctrine ODM will lose existing elements in an EmbedMany when you attempt to add() a new element to an embedded collection (aka "PersistentCollection") if it was loaded from the database and not yet initialized. This seems to be the case in at least BETA12, BETA13 and master HEAD. Very scary stuff.

The included unit test shows a very simple sequence of steps where we attempt to try to add() a new element to an embedded collection. This exact same simplistic test will fail when attempted against the master line of Doctrine ODM.

I had to add an optional initialize param to the add() function. This seems to have required that the memory_limit for PHP be set much higher.


jmikola:
Will attempt to look into this Sunday afternoon.

stof:
isn't it possible to fix this bug without altering the public API of the class ?

stof:
OK, just saw #1119 doing exactly that, which is much better IMO

malarzm:
@stof :+1:

Issue #1114, opened 2015-06-02
Still at risk of data loss & corruption with strategy="atomicSet"


blockjon:
I'm getting 2 update queries instead of 1 update query even when I am using @malarzm 's new atomicSet. I'm not sure if that means there's more work to do to fix the problem, or if perhaps I'm doing something wrong. I'm probably doing something wrong.

Here's a functional test which exhibits this behavior:
https://github.com/doctrine/mongodb-odm/pull/1113

malarzm:
I'm afraid that probably I'm the one who did something wrong :)

jmikola:
Fixed by #1118.

PR #1118, opened 2015-06-05
Ensure atomic collection updates use only one query

jmikola:
Incorporates tests from PR #1113 (issue #1114) and query logging added in @malarzm's own branch.


malarzm:
LGTM :+1:

Issue #1117, opened 2015-06-03
Adding an embedded item to an existing record results in data loss


blockjon:
Given a book with a chapter, this sequence of events fails:
```
// imagine a book already exists with a chapter in it...
$book = $dm->findOneBy(array('_id' => $id));
// Add a second chapter to this book object which already has a chapter
$book->chapters->add(new Chapter('Chapter Two'));
$dm->flush();
$dm->clear();

$book = $dm->findOneBy(array('_id' => $id));
this->assertEquals(2, $book->chapters->count()); // <- Fails!!!!
```
Is it just me, or does this seem *really* bad?

Here's my take on fixing it:
https://github.com/doctrine/mongodb-odm/pull/1116

malarzm:
Indeed this is happening but if I'm seeing correctly only when `set` or `atomicSet` is used since they're merging new objects using `set()` which may overwrite object using same index (which is very likely when using `add()` on uninitialized collection)

blockjon:
@malarzm To answer your question, it seems that initialize() must be called manually if the strategy is "set" or "atomicSet". However, initialize() does not need to be called if you are using "pushAll", "setArray", or "atomicSetArray".

In order to simplify, it would be awesome if the ODM's API always worked the same way so that the developer does not need to manually call initialize() ever.

malarzm:
@blockjon I'm hiding manual initialization and forcing ODM to think about this for developer in #1119

blockjon:
i like https://github.com/doctrine/mongodb-odm/pull/1119

PR #1119, opened 2015-06-06
Fix PersistentCollection::add usage with set strategy

malarzm:
This is alternative solution for #1117 (closes #1116) which runs `initialize()` while `PersistentCollection::add` is subject to data loss because of data merging strategy

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/PersistentCollection.php
+++ b/lib/Doctrine/ODM/MongoDB/PersistentCollection.php
@@ -524,6 +524,13 @@ class PersistentCollection implements BaseCollection
      */
     public function add($value)
     {
+        /* Initialize the collection before calling add() so this append operation
+         * uses the appropriate key. Otherwise, we risk overwriting original data
+         * when $newObjects are re-added in a later call to initialize().
+         */
+        if (isset($this->mapping['strategy']) && ($this->mapping['strategy'] === 'set' || $this->mapping['strategy'] === 'atomicSet')) {
+            $this->initialize();
+        }
         $this->coll->add($value);
         $this->changed();
         return true;
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1117Test.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1117Test.php
@@ -1,0 +1,59 @@
+<?php
+
+namespace Doctrine\ODM\MongoDB\Tests\Functional\Ticket;
+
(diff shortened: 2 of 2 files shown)


blockjon:
LGTM thank you!

stof:
:+1:

lib/Doctrine/ODM/MongoDB/PersistentCollection.php
```diff
@@ -524,6 +524,9 @@ public function set($key, $value)
      */
     public function add($value)
     {
+        if (isset($this->mapping['strategy']) && ($this->mapping['strategy'] === 'set' || $this->mapping['strategy'] === 'atomicSet')) {
+            $this->initialize();
+        }
```
jmikola: @malarzm: @blockjon mentioned in #1117 that:

> To answer your question, it seems that initialize() must be called manually if the strategy is "set" or "atomicSet". However, initialize() does not need to be called if you are using "pushAll", "setArray", or "atomicSetArray".

Why might that be?

malarzm: "set" and "atomicSet" are using `->set()` in `initialize()` and therefore thing that was `->add()` to uninitialized PersistentCollection will get index 0 and later overwrite thing at index 0 from database. Other strategies are using `->add()` so nothing will be overwritten

jmikola: Got it. I'd suggest adding a comment above this conditional for future reference:

> Initialize the collection before calling add() so this append operation uses the appropriate key. Otherwise, we risk overwriting original data when $newObjects are re-added in a later call to initialize().

jmikola: FWIW, we could probably do something clever by calculating the largest numeric key in the original document data (i.e. `$mongoData`) and turn this first `add()` into a `set()` (future `add()` calls wouldn't need it); however, calling `initialize()` manually here is certainly more straightforward.

Also, `CollectionPersister::prepareSetQuery()` is going to `initialize()` anyway :)

jmikola:
LGTM once a comment is added :+1:

lib/Doctrine/ODM/MongoDB/PersistentCollection.php
malarzm: I think I'd rather keep this simple, as noted the collection will be probably initialized anyway ;)

malarzm:
Merged manually in https://github.com/doctrine/mongodb-odm/commit/8f4362f86bf04389364a6d4bf20489f3883918c2

PR #1120, opened 2015-06-07
Do not schedule collection for deletion if it is to be $set later

### LINKED ISSUE #434: Using strategy set on EmbedMany seems to call an unset then set
Been trying to debug some really weird race condition issues that happen when a bunch of AJAX actions are fired off. The AJAX call passes in a new array of recipients to add onto an RFP object. We setup a logger on the Doctrine mongo connection. The expected action is that it would simply call _set_, what seems to happen is that it first calls _unset_ on the field and then call _set_ with the correct data. Since mongo isn't transactional, this opens up the (very real) possibility that another thread will read the data in between the _unset_ and _set_ calls (where the data is missing). The default pushAll strategy does the same - the issue remains the same, but removing the unset would break the expected behavior (even doing a diff and setting/unsetting array elements would be subject to issues where the incorrect elements could be unset if the document had been updated in between the read and flush events).

From looking at the code, the set strategy seems to be the only one that would be safe from a race condition (although, we are still seeing multiple set queries for the same object within the same flush), as long as the unset is dropped. Trying to understand the logic behind including it to make sure that we're not going to break something else by removing it.

Also, the set and pushAll strategies seem to save the EmbedMany ArrayCollections in different formats - pushAll as an array and set as a hash (with integers stored as strings for keys). We're currently using the default pushAll strategy but plan to convert to the set strategy to (partially) solve this race condition.

```php
Array
(
    [update] => 1
    [query] => Array
        (
            [_id] => MongoId Object
                (
                    [$id] => 5080abcfd8ad73b13a000000
                )

        )

    [newObj] => Array
        (
            [$unset] => Array
                (
                    [recipients] => 1
                )

        )

    [options] => Array
        (
            [safe] => 1
            [fsync] => 1
        )

    [db] => test
    [collection] => rfps
)
```

```php
Array
(
    [update] => 1
    [query] => Array
        (
            [_id] => MongoId Object
                (
                    [$id] => 5080abcfd8ad73b13a000000
                )

        )

    [newObj] => Array
        (
            [$set] => Array
                (
                    [recipients] => Array
                        (
                            [0] => Array
                                (
                                    [name] =>  
                                    [email] => jeff@smith.com
                                    [contact] => Array
                                        (
                                            [$ref] => contacts
                                            [$id] => MongoId Object
                                                (
                                                    [$id] => 507ebffb4cd844af6e6873bc
                                                )

                                            [$db] => adstruc
                                        )

                                    [created] => MongoDate Object
                                        (
                                            [sec] => 1350611713
                                            [usec] => 0
                                        )

                                )

                            [1] => Array
                                (
                                    [name] =>  
                                    [email] => bob@bobdole.net
                                    [contact] => Array
                                        (
                                            [$ref] => contacts
                                            [$id] => MongoId Object
                                                (
                                                    [$id] => 507ebffb4cd844af6e6873bd
                                                )

                                            [$db] => adstruc
                                        )

                                    [created] => MongoDate Object
                                        (
                                            [sec] => 1350611713
                                            [usec] => 0
                                        )

                                )

                        )

                )

        )

    [options] => Array
        (
            [safe] => 1
            [fsync] => 1
        )

    [db] => adstruc
    [collection] => rfps
)
```

### LINKED ISSUE #1011: Superfluous $unset query when using "set" collection strategy

malarzm:
Fixes #1011 and closes #434 and closes #1133. It also takes care for similar situation we faced with `atomic` collections, two tests added in `AtomicSetTest` are relevant to removed lines in `DocumentPersister`

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
+++ b/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
@@ -785,6 +785,12 @@ class UnitOfWork implements PropertyChangedListener
                 // if embed-many or reference-many relationship
                 if (isset($class->fieldMappings[$propName]['type']) && $class->fieldMappings[$propName]['type'] === 'many') {
                     $changeSet[$propName] = array($orgValue, $actualValue);
+                    /* If original collection was exchanged with a non-empty value
+                     * and $set will be issued, there is no need to $unset it first
+                     */
+                    if ($actualValue && $actualValue->isDirty() && in_array($class->fieldMappings[$propName]['strategy'], array('set', 'setArray', 'atomicSet', 'atomicSetArray'))) {
+                        continue;
+                    }
                     if ($orgValue instanceof PersistentCollection) {
                         $this->scheduleCollectionDeletion($orgValue);
                     }
@@ -2727,6 +2733,13 @@ class UnitOfWork implements PropertyChangedListener
      */
     public function scheduleCollectionUpdate(PersistentCollection $coll)
     {
+        $mapping = $coll->getMapping();
+        if (isset($mapping['strategy']) && in_array($mapping['strategy'], array('set', 'setArray', 'atomicSet', 'atomicSetArray'))) {
+            /* There is no need to $unset collection if it will be $set later
+             * This is NOP if collection is not scheduled for deletion
+             */
+            $this->unscheduleCollectionDeletion($coll);
+        }
         $this->collectionUpdates[] = $coll;
         $this->scheduleCollectionOwner($coll);
     }
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/AtomicSetTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/AtomicSetTest.php
@@ -77,7 +77,10 @@ class AtomicSetTest extends \Doctrine\ODM\MongoDB\Tests\BaseTest
         $this->assertEquals('12345678', $user->phonenumbers[0]->getPhonenumber());
     }
 
-    public function testAtomicCollectionUnset()
+    /**
+     * @dataProvider provideAtomicCollectionUnset
+     */
+    public function testAtomicCollectionUnset($clearWith)
     {
         $user = new AtomicSetUser('Maciej');
         $user->phonenumbers[] = new Phonenumber('12345678');
@@ -92,7 +95,7 @@ class AtomicSetTest extends \Doctrine\ODM\MongoDB\Tests\BaseTest
         $this->assertEquals('12345678', $user->phonenumbers[0]->getPhonenumber());
 
         $user->surname = "Malarz";
-        $user->phonenumbers = null;
+        $user->phonenumbers = $clearWith;
         $this->ql->clear();
         $this->dm->flush();
         $this->assertCount(1, $this->ql, 'Updating a document and unsetting its embed-many collection requires one query');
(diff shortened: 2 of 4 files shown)


tests/Doctrine/ODM/MongoDB/Tests/Functional/AtomicSetTest.php
```diff
@@ -104,6 +104,52 @@ public function testAtomicCollectionUnset()
...
+        $this->assertCount(1, $this->ql, 'Updating emptied collection requires one query');
+        $this->dm->clear();
+
+        $user = $this->dm->getRepository(get_class($user))->find($user->id);
+        $this->assertEquals('Maciej', $user->name);
+        $this->assertCount(1, $user->phonenumbers);
+        $this->assertEquals('87654321', $user->phonenumbers[0]->getPhonenumber());
+    }
+
+    public function testAtomicCollectionSubstitued()
```
jmikola: Typo: substituted

jmikola: Maybe rename this to "testAtomicCollectionSubstitutedAndUpdate" or "ReplacementAndUpdate"?

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1011Test.php
```diff
@@ -0,0 +1,66 @@
...
+        $this->dm->flush();
+        $doc->embeds = new ArrayCollection();
+        $doc->embeds->add(new GH1011Embedded('test2'));
+        $this->uow->computeChangeSets();
+        $this->assertTrue($this->uow->isCollectionScheduledForUpdate($doc->embeds));
+        $r = new \ReflectionObject($this->uow);
+        $rp = $r->getProperty('collectionDeletions');
+        $rp->setAccessible(true);
+        $collectionDeletions = $rp->getValue($this->uow);
+        $this->assertCount(0, $collectionDeletions);
```
jmikola: Would it make more sense to use `UnitOfWork::isCollectionScheduledForDeletion()` here instead of dealing with reflection? I suppose you could save a reference to the original `$doc->embeds` (before you replace it).

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1011Test.php
```diff
@@ -0,0 +1,66 @@
...
+    }
+}
+
+/** @ODM\Document */
+class GH1011Documnet
```
jmikola: Type: Document

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1011Test.php
```diff
@@ -0,0 +1,66 @@
...
+        $this->dm->persist($doc);
+        $this->dm->flush();
+        $doc->embeds->clear();
+        $doc->embeds->add(new GH1011Embedded('test2'));
+        $this->uow->computeChangeSets();
+        $this->assertTrue($this->uow->isCollectionScheduledForUpdate($doc->embeds));
+        $this->assertFalse($this->uow->isCollectionScheduledForDeletion($doc->embeds));
+    }
+
+    public function testSubstituteCollection()
```
jmikola: Similar to the other test, "Replace" might be a better word here.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -785,6 +785,12 @@ private function computeOrRecomputeChangeSet(ClassMetadata $class, $document, $r
                 // if embed-many or reference-many relationship
                 if (isset($class->fieldMappings[$propName]['type']) && $class->fieldMappings[$propName]['type'] === 'many') {
                     $changeSet[$propName] = array($orgValue, $actualValue);
+                    /* If original collection was exchanged with something else
+                     * and $set will be issued, there is no need to $unset it first
+                     */
```
jmikola: Should we say "exchanged with a non-empty value" here? I think that's why you include `$actualValue` in the conditional below, since `null` or an empty array would not result in any new value being set and ODM expects to delete the original PersistentCollection (which we'll do below).

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -785,6 +785,12 @@ private function computeOrRecomputeChangeSet(ClassMetadata $class, $document, $r
...
+                    if ($actualValue && in_array($class->fieldMappings[$propName]['strategy'], array('set', 'setArray', 'atomicSet', 'atomicSetArray'))) {
+                        continue;
+                    }
```
jmikola: Actually, this might have a problem with `AtomicSetTest::testAtomicCollectionUnset`. If you change `$user->phonenumbers = null;` with an empty array or new ArrayCollection instance, the original data is not cleared. Perhaps this is due to logic elsewhere where ODM is not including empty collection data in a `$set` operation?

jmikola: While we're at it, it would be good to add additional test cases for setting collections to `array()` and `new ArrayCollection()`, in addition to the current test for `null`.

malarzm: Yup, added and working on a fix :+1:

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -785,6 +785,12 @@ private function computeOrRecomputeChangeSet(ClassMetadata $class, $document, $r
                 // if embed-many or reference-many relationship
                 if (isset($class->fieldMappings[$propName]['type']) && $class->fieldMappings[$propName]['type'] === 'many') {
                     $changeSet[$propName] = array($orgValue, $actualValue);
+                    /* If original collection was exchanged with a non-empty value
+                     * and $set will be issued, there is no need to $unset it first
+                     */
+                    if ($actualValue && $actualValue->isDirty() && in_array($class->fieldMappings[$propName]['strategy'], array('set', 'setArray', 'atomicSet', 'atomicSetArray'))) {
```
jmikola: This might break if `$actualValue` isn't a PersistentCollection, since ArrayCollection doesn't have an `isDirty()` method.

Additionally, if the user stores some other invalid value here (e.g. `5`), we'd error out instead of getting an exception down the line.

malarzm: `UnitOfWork::getDocumentActualData` ensures we get PersistentCollection here for everything that's "many" type

malarzm: And is not `null`

jmikola: Got it. The "getting an exception down the line" that I saw earlier when testing this probably happened *earlier* during `getDocumentActualData()` then :)

jmikola:
:+1:

malarzm:
Merged manually in https://github.com/doctrine/mongodb-odm/commit/c06e8d40a3fd77a9c24481247890863ea858355c

PR #1121, opened 2015-06-08
preUpdate event is dispatched when only refmany collection is updated

### LINKED ISSUE #861: Document modified on event listener not flushed correctly

malarzm:
Fixes #861, closes #862


malarzm:
Merged manually in https://github.com/doctrine/mongodb-odm/commit/c055530e236b12ef29a22937e9ebc3d14ef0959a

PR #1125, opened 2015-06-08
Two functional tests demonstrate how atomicSet is not compatible with HasLifecycleCallbacks and causes data corruption

blockjon:
I have discovered that I still am at risk for data loss and corruption even on the master line when I integrate with my project despite recent improvements made in https://github.com/doctrine/mongodb-odm/pull/1118.

The current defect seems to manifest when atomicSet is used in conjunction with an embedded document *that has a LifecycleCallback*. In this PR, I use the book->chapter->page object hierarchy to demonstrate two problems that I have discovered. The problems are as follows:

- The update to the database is not atomic and is performed in two steps
- The data saved to the database is not correct. My unit test tries to create a second "page" embedded object within the chapter object, but the database actually gets a third "page" object saved into it.... something I have additionally verified by looking at my MongoDB record by hand.

Here is how to run the problem case on this PR:

```
phpunit tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH111XTest.php
```
this comes back:
```
PHPUnit 4.1.6-9-gbe3530d by Sebastian Bergmann.

Configuration read from /mongodb-odm/phpunit.xml

FF

Time: 367 ms, Memory: 3.75Mb

There were 2 failures:

1) Doctrine\ODM\MongoDB\Tests\Functional\Ticket\GH1113Test::testAtomicSetStrategySendsJustOneUpdateStatementToMongoDbWithLifecycleCallbacksInEmbeddedDocument
Adding a new page object should only cause a single update when using atomicSet.
Failed asserting that actual size 2 matches expected size 1.

/mongodb-odm/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH111XTest.php:53

2) Doctrine\ODM\MongoDB\Tests\Functional\Ticket\GH1113Test::testAtomicSetStrategyAddsExpectedNumberOfEmbeddedDocumentsOnUpdate
Two page objects are expected in the first chapter of the book.
Failed asserting that 3 matches expected 2.

/mongodb-odm/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH111XTest.php:85

FAILURES!
Tests: 2, Assertions: 3, Failures: 2.
```

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GHXXXXTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GHXXXXTest.php
@@ -1,0 +1,156 @@
+<?php
+
+namespace Doctrine\ODM\MongoDB\Tests\Functional\Ticket;
+
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
+use Doctrine\ODM\MongoDB\Tests\QueryLogger;
+
+class GH1113Test extends \Doctrine\ODM\MongoDB\Tests\BaseTest
+{
+    private $ql;
+
+    protected function getConfiguration()
+    {
+        if ( ! isset($this->ql)) {
+            $this->ql = new QueryLogger();
+        }
+
+        $config = parent::getConfiguration();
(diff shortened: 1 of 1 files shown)


malarzm:
@blockjon I've included your test case in #1128 :)

Issue #1126, opened 2015-06-08
atomicSet is not compatible with Lifecycle callbacks


blockjon:
I have created https://github.com/doctrine/mongodb-odm/pull/1125 to exhibit this problem.

jmikola:
@blockjon: If you happen to open a PR with a failing test first, feel free to skip a separate issue. That will make it easier to consolidate discussion (even if the test simply ends up being cherry-picked in lieu of merging the original PR).

blockjon:
ok.. I will close this issue out but leave the PR open in order for it to be triaged there

Issue #1046, opened 2015-03-11
Investigate Collection update and locking document


malarzm:
As per [this](https://github.com/doctrine/mongodb-odm/pull/985#discussion_r26247231) comment:

> It strikes me as odd that subsequent Collection updates don't use the lock, though (see: [CollectionPersister::executeQuery()](https://github.com/doctrine/mongodb-odm/blob/30f42aa01ff4aa289c1090d6cc7fd79d1ac8afe1/lib/Doctrine/ODM/MongoDB/Persisters/CollectionPersister.php#L284)). It's like unrelated to this PR, but I think that warrants further investigation and we should open a ticket to revisit down the line.

jmikola:
We haven't investigated this further, but #1096 introduced atomic collection strategies, which may aid with updating versioned documents.

PR #1128, opened 2015-06-09
Create snapshot of collection after querying db

malarzm:
Atomic collections and these nested inside were still considered dirty even after update query was issued and they were scheduled for update again if `UoW::recomputeSingleDocumentChangeSet` was called for instance by `postUpdate`.

@blockjon's test from #1126 is incorporated in this PR

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -211,6 +211,7 @@ class DocumentPersister
             return;
         }
 
+        $allVisitedCollections = array();
         $inserts = array();
         foreach ($this->queuedInserts as $oid => $document) {
             $data = $this->pb->prepareInsertData($document);
@@ -228,11 +229,12 @@ class DocumentPersister
                 }
                 $data[$versionMapping['name']] = $nextVersion;
             }
-            
-            $atomicCollectionQuery = $this->getAtomicCollectionUpdateQuery($document);
-            if (!empty($atomicCollectionQuery['$set'])) {
+
+            list($visitedCollections, $atomicCollectionQuery) = $this->getAtomicCollectionUpdateQuery($document);
+            if ( ! empty($atomicCollectionQuery['$set'])) {
                 $data = array_merge($data, $atomicCollectionQuery['$set']);
             }
+            $allVisitedCollections = array_merge($allVisitedCollections, $visitedCollections);
 
             $inserts[$oid] = $data;
         }
@@ -246,6 +248,9 @@ class DocumentPersister
             }
         }
 
+        foreach ($allVisitedCollections as $coll) {
+            $coll->takeSnapshot();
+        }
         $this->queuedInserts = array();
     }
 
@@ -281,7 +286,7 @@ class DocumentPersister
                 $data['$set'][$versionMapping['name']] = $nextVersion;
             }
             
-            $atomicCollectionQuery = $this->getAtomicCollectionUpdateQuery($document);
+            list($visitedCollections, $atomicCollectionQuery) = $this->getAtomicCollectionUpdateQuery($document);
             if ( ! empty($atomicCollectionQuery)) {
                 $data = array_merge_recursive($data, $atomicCollectionQuery);
             }
@@ -292,6 +297,10 @@ class DocumentPersister
             } catch (\MongoException $e) {
                 unset($this->queuedUpserts[$oid]);
                 throw $e;
+            }
+
+            foreach ($visitedCollections as $coll) {
+                $coll->takeSnapshot();
             }
         }
     }
@@ -379,7 +388,7 @@ class DocumentPersister
                 }
             }
             
-            $atomicCollectionQuery = $this->getAtomicCollectionUpdateQuery($document);
+            list($visitedCollections, $atomicCollectionQuery) = $this->getAtomicCollectionUpdateQuery($document);
             if ( ! empty($atomicCollectionQuery)) {
                 $update = array_merge_recursive($update, $atomicCollectionQuery);
             }
@@ -408,6 +417,10 @@ class DocumentPersister
 
             if (($this->class->isVersioned || $this->class->isLockable) && ! $result['n']) {
                 throw LockException::lockFailed($document);
+            }
+
+            foreach ($visitedCollections as $coll) {
+                $coll->takeSnapshot();
             }
         }
     }
@@ -1243,12 +1256,23 @@ class DocumentPersister
 
         return $discriminatorValues;
     }
-    
+
+    /**
+     * Prepares query for saving atomic collections and its children.
+     *
+     * <code>
+     * list($visitedCollections, $query) = $this->getAtomicCollectionUpdateQuery($document);
+     * </code>
+     *
+     * @param object $document
+     * @return array
+     */
     private function getAtomicCollectionUpdateQuery($document)
     {
         $update = array();
         $atomicCollUpdates = array();
         $atomicCollDeletes = array();
+        $visitedCollections = array();
         $collPersister = $this->uow->getCollectionPersister();
 
         /* Collect all atomic collections (top-level and nested) to be included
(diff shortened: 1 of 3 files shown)


lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -293,6 +298,10 @@ public function executeUpserts(array $options = array())
                 unset($this->queuedUpserts[$oid]);
                 throw $e;
             }
+
+            foreach ($visitedCollections as $coll) {
+                $coll->takeSnapshot();
+            }
```
jmikola: Why does `executeInserts()` collect these into a `$allVisitedCollections` array and issue all of the snapshots after iterating on `$this->queuedInserts`, while `executeUpserts()` is taking snapshots once per `$this->queuedUpserts` iteration?

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -1327,6 +1355,6 @@ private function getAtomicCollectionUpdateQuery($document)
             $this->uow->unscheduleCollectionDeletion($coll);
         }
 
-        return $update;
+        return array($visitedCollections, $update);
```
jmikola: `$visitedCollections` seems a bit misleading since we are looking at more collections than these. Is there any reason we can't take snapshots internally, and defer to the parent method to do so (`executeInserts()` or `executeUpserts()`)?

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -1299,6 +1325,8 @@ private function getAtomicCollectionUpdateQuery($document)
             if ( ! $this->uow->isCollectionScheduledForDeletion($parentColl)) {
                 $atomicCollUpdates[spl_object_hash($parentColl)] = $parentColl;
             }
+
+            $visitedCollections[] = $coll;
```
jmikola: In this case, the `$parentColl` is effectively being updated, even though it might never have been scheduled explicitly. Does it need a snapshot taken as well, or do we only care about the originally scheduled child collection?

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1126Test.php
```diff
@@ -0,0 +1,156 @@
...
+    {
+        $this->pages->add($page);
+    }
+
+    /**
+     * @ODM\PostUpdate
+     */
+    public function doThisAfterAnUpdate()
+    {
+        // (nothing)
```
jmikola: Your original post in the PR talks about `recomputeSingleDocumentChangeSet()` being called from a lifecycle listener, but we're not doing anything here. Would it be a good idea to actually have a listener make further modificatiosn and recompute the changeset?

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
malarzm: Idea is to take a snapshot only if query was successful and since `executeInserts()` uses `batchInsert()` there's no better way to do that over there

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
malarzm: Theoreticaly we don't care since they're not considered dirty but I think it's worth setting snapshots all the way for the sake of consistency and since #1134 might check it when `->clear()` is called

malarzm: `UnitOfWork::visitedCollections` will actually contain all the collections and snapshot them at the end of `commit()`, but an exception can be thrown in the middle leaving some collections unsnapshooted.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
jmikola: Ah, that makes sense. I'd suggest a comment over in `executeInserts()` to explain that for future reference. If we ever get around to doing bulk updates with the new driver API, that will be something to consider.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
jmikola: @malarzm: Nevermind. You explained that you only wanted to take snapshots once the writes succeeded.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -1243,12 +1256,23 @@ private function getClassDiscriminatorValues(ClassMetadata $metadata)
...
+     * Prepares query for saving atomic collections and its children.
+     *
+     * <code>
+     * list($visitedCollections, $query) = $this->getAtomicCollectionUpdateQuery($document);
+     * </code>
+     *
+     * @param object $document
+     * @return array
+     */
     private function getAtomicCollectionUpdateQuery($document)
```
jmikola: At risk of being pedantic, I'd change this to `getAtomicCollectionUpdateQueryAndVisitedCollections()` and have it return a tuple in the opposite order, since the update query is the primary return value.

malarzm:
@jmikola actually I think we should change whole flow of updating/removing Collections. To recap current commit flow is as follows:

1. Upserts
2. Inserts
3. Updates
4. Collection deletions
5. Collection updates
6. Deletions
7. Collection snapshots

Separating collection updates from parent document so far lead us to #1126, #1145 and #1046 (from top of my head). Now that we have introduced atomic strategies DocumentPersister will also take care of them and unschedule so CollectionPersister won't pick them up but I believe your suspection in https://github.com/malarzm/mongodb-odm/pull/4#issuecomment-111592953 is correct and altering child of atomic collection in `postUpdate` will schedule collection for update again and throw exception in CollectionPersister as atomic collections can't be handled by it.

The change I want to propose is to move all collection updates/deletions into DocumentPersister (it can call internally CollectionPersister) and take care of these collections just after parent document is resolved (before we return to UnitOfWork to dispatch events). This leads to several enhancements:

1. Document A's collections will be updated just after A which logically makes sense (a lot of :) )
2. We will be able to easily fix Version check of document while saving its collections since DocumentPersister has already calculated that and can simply pass information to CollectionPersister
3. All collections will be snapshotted just after saving so there will be no place for desynchronization of snapshot and db state (now it can happen if `DocumentPersister::remove` would throw `LockException`)
4. If update of document B will throw exception it will not affect A's collections
5. postUpdate will be correctly triggered after database query(ies) as it is expected to be
6. updating document's collection in postPersist/postUpdate will no longer trigger saving them (which might be a BC break but again, both events where meant to be fired after the database operations, so this change is totally sane)
7. Collections are handled in one place, not in two as currently is needed
8. Right now if parent document is to be deleted, we will issue additional query to its collections anyway (if these collections are scheduled for some reason) - it will no longer be an issue

What do you think? cc @jwage @alcaeus

jmikola:
> altering child of atomic collection in postUpdate will schedule collection for update again and throw exception in CollectionPersister as atomic collections can't be handled by it.

Reading that lead me to consider another point. We currently have logic to check if a scheduled collection is contained within some top-level atomic collection. In a more general sense, should we be doing the same for collections with `addToSet` or `pushAll` strategies that are contained in a parent with a `set*` strategy? AFAIK, we don't currently do this, although ODM likely issues the higher-level collection's write before the child. If a parent uses `set*`, though, I think the benefit of an `addToSet` or `pushAll` is lost, so we might as well coalesce the writes.

> 6. updating document's collection in postPersist/postUpdate will no longer trigger saving them (which might be a BC break but again, both events where meant to be fired after the database operations, so this change is totally sane)

This behavioral change will be due to the collection modification happening between the document's write and any event dispatching, correct?

That said, I'm inclined to agree with your proposal :)

malarzm:
> Reading that lead me to consider another point. We currently have logic to check if a scheduled collection is contained within some top-level atomic collection. In a more general sense, should we be doing the same for collections with addToSet or pushAll strategies that are contained in a parent with a set* strategy? AFAIK, we don't currently do this, although ODM likely issues the higher-level collection's write before the child. If a parent uses set*, though, I think the benefit of an addToSet or pushAll is lost, so we might as well coalesce the writes.

Awesome, probably it will fix #1104 :)

> This behavioral change will be due to the collection modification happening between the document's write and any event dispatching, correct?

Exactly

malarzm:
Superseded by #1155

PR #1129, opened 2015-06-09
Fix overriding id strategy

### LINKED ISSUE #1107: No id for child object with strategy="AUTO"
I've written a test which fails. Id is properly generated and persisted in database, but it is not propagate to PHP object.

```
<?php
namespace Doctrine\ODM\MongoDB\Tests\Persisters;
use Doctrine\ODM\MongoDB\Tests\BaseTest;
class IdStrategyForInheritanceTest extends BaseTest
{
    public function setUp()
    {
        parent::setUp();
    }
    public function tearDown()
    {
        parent::tearDown();
    }
    public function testOverrideIdStrategy()
    {
        $childObj = new ChildClass();
        $childObj->name = 'ChildObject';
        $this->dm->persist($childObj);
        $this->dm->flush();
        $this->assertNotNull($childObj->id);
    }
}
/**
 * @ODM\Document 
 */
class ParentClass
{
    /** @ODM\Id(strategy="NONE") */
    public $id;
    /** @ODM\String */
    public $name;
}
/** @ODM\Document */
class ChildClass extends ParentClass
{
    /** @ODM\Id(strategy="AUTO") */
    public $id;
}

```

malarzm:
Fixes #1107

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataFactory.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataFactory.php
@@ -158,7 +158,7 @@ class ClassMetadataFactory extends AbstractClassMetadataFactory
 
         $this->validateIdentifier($class);
 
-        if ($parent && $rootEntityFound) {
+        if ($parent && $rootEntityFound && $parent->generatorType === $class->generatorType) {
             if ($parent->generatorType) {
                 $class->setIdGeneratorType($parent->generatorType);
             }
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1107Test.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1107Test.php
@@ -1,0 +1,38 @@
+<?php
+
+namespace Doctrine\ODM\MongoDB\Tests\Persisters;
+
+use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
+use Doctrine\ODM\MongoDB\Tests\BaseTest;
+
+class GH1107Test extends BaseTest
+{
+    public function testOverrideIdStrategy()
+    {
+        $childObj = new GH1107ChildClass();
+        $childObj->name = 'ChildObject';
+        $this->dm->persist($childObj);
+        $this->dm->flush();
+        $this->assertNotNull($childObj->id);
+    }
+}
+
+/**
+ * @ODM\Document
+ * @ODM\InheritanceType("SINGLE_COLLECTION")
+ */
+class GH1107ParentClass
+{
+    /** @ODM\Id(strategy="NONE") */
+    public $id;
+
+    /** @ODM\String */
+    public $name;
+}
+
+/** @ODM\Document */
+class GH1107ChildClass extends GH1107ParentClass
+{
+    /** @ODM\Id(strategy="AUTO") */
+    public $id;
+}


lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataFactory.php
```diff
@@ -158,7 +158,7 @@ protected function doLoadMetadata($class, $parent, $rootEntityFound, array $nonS
 
         $this->validateIdentifier($class);
 
-        if ($parent && $rootEntityFound) {
+        if ($parent && $rootEntityFound && $parent->generatorType === $class->generatorType) {
```
jmikola: Earlier in the `if ($parent)` block, we did `$class->setIdGeneratorType($parent->generatorType);` so wouldn't this always be true if there's a parent?

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1107Test.php
```diff
@@ -0,0 +1,38 @@
...
+
+class GH1107Test extends BaseTest
+{
+    public function testOverrideIdStrategy()
+    {
+        $childObj = new GH1107ChildClass();
+        $childObj->name = 'ChildObject';
+        $this->dm->persist($childObj);
+        $this->dm->flush();
+        $this->assertNotNull($childObj->id);
```
jmikola: What are we testing here? Would the `NONE` strategy cause ODM to not construct an ObjectId at all (leaving it to the driver to inject one before we send the document across the wire)?

I assume the bug here is that the ID strategy could never be overridden? In that case, would it make sense to have a base class use something like an auto-increment sequence and then try to override that with the AUTO strategy and confirm that parent and child documents get an integer and ObjectId, respectively?

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataFactory.php
malarzm: It is unless driver loads other informations

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1107Test.php
malarzm: > What are we testing here?

Prior to this PR Id strategy of child document is ignored

> Would the NONE strategy cause ODM to not construct an ObjectId at all (leaving it to the driver to inject one before we send the document across the wire)?

Yes, but the value is not hydrated into document because of https://github.com/doctrine/mongodb-odm/commit/e2853e9454c6ad473ace421ff00d62492acaebd7

> leaving it to the driver to inject one before we send the document across the wire

As far as I understand using NONE strategy means "I want to generate id on my own, please don't do this for me". I might even say we should throw an exception if we persist a document using strategy NONE without an identifier? It may lead to sick behaviour when creating a DBRef especially that ODM doesn't care what id was generated by driver.

> I assume the bug here is that the ID strategy could never be overridden? In that case, would it make sense to have a base class use something like an auto-increment sequence and then try to override that with the AUTO strategy and confirm that parent and child documents get an integer and ObjectId, respectively?

It overriding NONE strategy will work then overwriting any strategy will work, not sure if we need to complicate things here :)

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataFactory.php
jmikola: Ah yes, I missed `loadMetadataForClass()`.

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1107Test.php
jmikola: > I might even say we should throw an exception if we persist a document using strategy NONE without an identifier?

I think it's worth creating a ticket for that. Even if the driver/server creates an ID for us, it's of no use to ODM since we wouldn't be able to properly track that document in the identity map.

jmikola:
LGTM. Feel free to link a new ticket re: missing ID for the `NONE` strategy back to this.

malarzm:
Merged manually in https://github.com/doctrine/mongodb-odm/commit/3e857760e674a93a10f022e99f52eef2fbfdba14

PR #1130, opened 2015-06-09
Reindex numerically PersistentCollection after save

### LINKED ISSUE #944: Consecutive flush operations after EmbedMany removal causes an entity desynchronization
```php
/**
 *
 * @ODM\Document
 */
class Document
{
    /** @ODM\Id(strategy="auto") */
    private $id;

    /** @ODM\EmbedMany */
    public $data;

    public function removeByText($text)
    {
        foreach ($this->data as $index => $d) {
            if ($d->text == $text) {
                unset($this->data[$index]);
            }
        }
    }
}

/**
 * @ODM\EmbeddedDocument
 */
class Embedded
{
    /** @ODM\String */
    public $text;
}

//main.php

        $d1 = new Document();

        $sd1 = new Embedded();
        $sd1->text = "1";
        $sd2 = new Embedded();
        $sd2->text = "2";
        $d1->data = array($sd1, $sd2);

        $dm->persist($d1);
        $dm->flush();
        $dm->clear();

        $r = $dm->getRepository('Document');
        $d1 = $r->findOneBy(array());
        $d1->removeByText("1");
        $dm->flush(); // this causes the problem

        $d1 = $r->findOneBy(array());
        $d1->removeByText("2");
        $dm->flush();

        $dm->clear();
        $d1 = $r->findOneBy(array());

        //expected 0, actual 1
        $this->assertEquals(0, count($d1->data));
```

The problem appears when I try to remove and flush multiple EmbedMany. A workaround would be to flush only once but that's hard to enforce. Also somebody should confirm the removal methods is correct.

After the first flush the embeded Collection gets a different id for each element from mongo, but no updates are made on the unit of work to synchronize them(I presume). So the next flush will update 'data.1' => true but there is not 'data.1' inside the entity.

malarzm:
Fixes #944 and brings `PersistentCollection` in line with what is in database although it will be a BC break if somebody relied on using old indexes.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/PersistentCollection.php
+++ b/lib/Doctrine/ODM/MongoDB/PersistentCollection.php
@@ -269,10 +269,20 @@ class PersistentCollection implements BaseCollection
 
     /**
      * INTERNAL:
-     * Tells this collection to take a snapshot of its current state.
+     * Tells this collection to take a snapshot of its current state reindexing
+     * itself numerically if using save strategy that is enforcing BSON array.
+     * Reindexing is safe as snapshot is taken only after synchronizing collection
+     * with database or clearing it.
      */
     public function takeSnapshot()
     {
+        if (CollectionHelper::isList($this->mapping['strategy'])) {
+            $array = $this->coll->toArray();
+            $this->coll->clear();
+            foreach ($array as $document) {
+                $this->coll->add($document);
+            }
+        }
         $this->snapshot = $this->coll->toArray();
         $this->isDirty = false;
     }
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH944Test.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH944Test.php
@@ -1,0 +1,75 @@
+<?php
+
+namespace Doctrine\ODM\MongoDB\Tests\Functional\Ticket;
+
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
+
+class GH944Test extends \Doctrine\ODM\MongoDB\Tests\BaseTest
+{
+    public function testIssue()
+    {
+        $d = new GH944Document();
+        $d->data->add(new GH944Embedded('1'));
+        $d->data->add(new GH944Embedded('2'));
+        $this->dm->persist($d);
+        $this->dm->flush();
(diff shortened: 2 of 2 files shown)


lib/Doctrine/ODM/MongoDB/PersistentCollection.php
```diff
@@ -268,10 +268,20 @@ public function setOwner($document, array $mapping)
...
      */
     public function takeSnapshot()
     {
+        if (in_array($this->mapping['strategy'], array('atomicSetArray', 'setArray', 'pushAll', 'addToSet'))) {
+            $array = $this->coll->toArray();
+            $this->coll->clear();
+            foreach ($array as $document) {
+                $this->coll->add($document);
+            }
+        }
```
jmikola: It kind of stinks that we can't simply filter the entire array in one shot, but PersistentCollection can't really assume we're working with ArrayCollection specifically -- just a Collection interface.

jmikola:
I'm :+1: on this. If we don't re-index, it's possible for our data to get out of sync with the database value, which means we could calculate bad changesets down the line. I assume that was what #944 was reporting.

alcaeus:
I agree with @jmikola here :+1:

lib/Doctrine/ODM/MongoDB/PersistentCollection.php
```diff
@@ -268,10 +268,20 @@ public function setOwner($document, array $mapping)
...
      * INTERNAL:
-     * Tells this collection to take a snapshot of its current state.
+     * Tells this collection to take a snapshot of its current state reindexing
+     * itself numerically if using save strategy that is enforcing BSON array.
+     * Reindexing is safe as snapshot is taken only after synchronizing collection
+     * with database or clearing it.
      */
     public function takeSnapshot()
     {
+        if (in_array($this->mapping['strategy'], array('atomicSetArray', 'setArray', 'pushAll', 'addToSet'))) {
```
malarzm: TODO: utilize CollectionHelper once #1155 is merged

jmikola: LGTM once that change is in.

malarzm: Will be done during rebase/merge party ;)

malarzm:
Merged manually in https://github.com/doctrine/mongodb-odm/commit/10af5e140ded5a7eb36ab9bd4e302156b9e54c3e

PR #1134, opened 2015-06-10
Fix usage of cloned PersistentCollection

### LINKED ISSUE #1132: Cloned PersistentCollection can't be cleared
Here's test case failing on dev-master. The error comes from `UoW::scheduleCollectionOwner` which was introduced in BETA13 BUT if you run the test on BETA12 you will see it was failing before as well and it is not a regression. (Actually test needs `Documents\User` modified with addition of `setPhonenumbers()` method)

```
<?php

namespace Doctrine\ODM\MongoDB\Tests\Functional\Ticket;

use Documents\Phonenumber;
use Documents\User;

class GH1132Test extends \Doctrine\ODM\MongoDB\Tests\BaseTest
{
    public function testOnwerlessPersistentCollectionCanBeScheduledForDeletion()
    {
        $u = new User();
        $u->getPhonenumbers()->add(new Phonenumber('123456'));
        $this->dm->persist($u);
        $this->dm->flush();

        $u2 = new User();
        $u2->setPhonenumbers(clone $u->getPhonenumbers());
        $u2->getPhonenumbers()->clear();
        $this->dm->persist($u2);
        $this->dm->flush();
        $this->dm->clear();

        $u = $this->dm->find(get_class($u), $u->id);
        $this->assertCount(1, $u->getPhonenumbers());

        $u2 = $this->dm->find(get_class($u2), $u2->id);
        $this->assertCount(0, $u2->getPhonenumbers());
    }
}
```

malarzm:
Fixes #1132

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/PersistentCollection.php
+++ b/lib/Doctrine/ODM/MongoDB/PersistentCollection.php
@@ -608,11 +608,20 @@ class PersistentCollection implements BaseCollection
 
         $this->mongoData = array();
         $this->coll->clear();
-        if ($this->mapping['isOwningSide']) {
-            $this->changed();
-            $this->uow->scheduleCollectionDeletion($this);
-            $this->takeSnapshot();
+
+        // Nothing to do for inverse-side collections
+        if ( ! $this->mapping['isOwningSide']) {
+            return;
         }
+
+        // Nothing to do if the collection was initialized but contained no data
+        if ($this->initialized && empty($this->snapshot)) {
+            return;
+        }
+
+        $this->changed();
+        $this->uow->scheduleCollectionDeletion($this);
+        $this->takeSnapshot();
     }
 
     /**
--- a/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
+++ b/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
@@ -706,6 +706,13 @@ class UnitOfWork implements PropertyChangedListener
             $this->originalDocumentData[$oid] = $actualData;
             $changeSet = array();
             foreach ($actualData as $propName => $actualValue) {
+                /* At this PersistentCollection shouldn't be here, probably it
+                 * was cloned and its ownership must be fixed
+                 */
+                if ($actualValue instanceof PersistentCollection && $actualValue->getOwner() !== $document) {
+                    $actualData[$propName] = $this->fixPersistentCollectionOwnership($actualValue, $document, $class, $propName);
+                    $actualValue = $actualData[$propName];
+                }
                 $changeSet[$propName] = array(null, $actualValue);
             }
             $this->documentChangeSets[$oid] = $changeSet;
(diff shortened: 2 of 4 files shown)


lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -706,6 +706,12 @@ private function computeOrRecomputeChangeSet(ClassMetadata $class, $document, $r
             $this->originalDocumentData[$oid] = $actualData;
             $changeSet = array();
             foreach ($actualData as $propName => $actualValue) {
+                /* At this PersistentCollection shouldn't be here, probably it
+                 * was cloned and its ownership must be fixed
+                 */
+                if ($actualValue instanceof PersistentCollection && $actualValue->getOwner() !== $document) {
+                    $actualData[$propName] = $actualValue = $this->fixPersistentCollectionOwnership($actualValue, $document, $class, $propName);
+                }
```
jmikola: Would it not be safer to unset the owner when we clone the PersistentCollection?

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -2674,6 +2670,23 @@ public function scheduleOrphanRemoval($document)
         $this->orphanRemovals[spl_object_hash($document)] = $document;
     }
 
+    private function fixPersistentCollectionOwnership(PersistentCollection $coll, $document, $class, $propName)
+    {
+        $owner = $coll->getOwner();
+        if ($owner === null) { // cloned
+            $coll->setOwner($document, $class->fieldMappings[$propName]);
+        } elseif ($owner !== $document) { // no clone, we have to fix
```
jmikola: So, PersistentCollection unsets the owner upon cloning, which seems fine.

I'm not sure about this condition, though. If the PersistentCollection is still affiliated with another owner, isn't it possible that it still belongs to that document? I would think it more proper to throw an exception here, since it really shouldn't be utilized in two different documents.

malarzm: `PersistentCollectionCloneTest::testNotCloneAndPassAroundFlush` is failing when exception is thrown :(

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -706,6 +706,12 @@ private function computeOrRecomputeChangeSet(ClassMetadata $class, $document, $r
             $this->originalDocumentData[$oid] = $actualData;
             $changeSet = array();
             foreach ($actualData as $propName => $actualValue) {
+                /* At this PersistentCollection shouldn't be here, probably it
+                 * was cloned and its ownership must be fixed
+                 */
+                if ($actualValue instanceof PersistentCollection && $actualValue->getOwner() !== $document) {
+                    $actualData[$propName] = $actualValue = $this->fixPersistentCollectionOwnership($actualValue, $document, $class, $propName);
```
jmikola: Can we split up the double assignment? Adding `$actualValue = $actualData[$propName];` on the following line would work.

lib/Doctrine/ODM/MongoDB/PersistentCollection.php
```diff
@@ -608,7 +608,7 @@ public function clear()
 
         $this->mongoData = array();
         $this->coll->clear();
-        if ($this->mapping['isOwningSide']) {
+        if ($this->mapping['isOwningSide'] && ( ! empty($this->snapshot) || ! $this->initialized)) {
```
jmikola: If would rewrite this for clarity:

```
// Nothing to do for inverse-side collections
if ( ! $this->mapping['isOwningSide']) {
    return;
}

// Nothing to do if the collection was initialized but contained no data
if ($this->initialized && empty($this->snapshot)) {
    return;
}
```

Otherwise, we can proceed with marking the collection as changed, scheduling it for deletion, and taking a snapshot.

jmikola: As you mentioned in IRC, perhaps we can use `$this->count()` to check if the collection contains no data.

malarzm: Now I think about this I was too eager with proposing `$this->count()`, I'd leave current condition

jmikola:
:+1:

PR #1137, opened 2015-06-11
Regression test for cascade persist of associations

jmikola:
@malarzm: I noticed that #909 had been fixed by some recent commits, but thought it be good to include the original test case.


malarzm:
:+1:

PR #1138, opened 2015-06-11
Modifying a document scheduled for insertion should not also schedule it for update

jmikola:
This test was sitting around in my local checkout, so I wanted to record it in a PR for further investigation before 1.0.0. I don't recall if it was ever related to a specific issue reported, or if it was just something I noticed while working on something many months ago.

The test is basically modifying a document at the time of flush (before writes are executed). The document is originally scheduled for insertion and our lifecycle event is going to modify it and recompute the changeset like a good listener should. This causes the document to also get scheduled for an update, with the very same data we're inserting (essentially a NOP, but potentially bad if another write interleaves these two).

A quick thought is that `computeOrRecomputeChangeSet()`'s logic for setting `$isNewDocument` is a bit too lax.

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH____Test.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH____Test.php
@@ -1,0 +1,82 @@
+<?php
+
+namespace Doctrine\ODM\MongoDB\Tests\Functional\Ticket;
+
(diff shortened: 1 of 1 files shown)


malarzm:
It seems similar to https://github.com/doctrine/mongodb-odm/issues/1058 and so should be fixed by @jwage's #782

malarzm:
I've cherry picked the test into #782

PR #1141, opened 2015-06-12
Functional test to demonstrate "Cannot update 'chapters.0.status' and 'chapters' at the same time" atomicSet glitch

blockjon:
Here's another blocker I have run into with strategy=atomicSet

This exact same code passes if you use strategy=set so hopefully there's a small fix that can be made to add support for this sequence of events when using atomicSet.

Repro:
```
phpunit ./tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1141Test.php
```
returns:
```
PHPUnit 4.1.6-9-gbe3530d by Sebastian Bergmann.

Configuration read from /mongodb-odm/phpunit.xml

E

Time: 332 ms, Memory: 3.25Mb

There was 1 error:

1) Doctrine\ODM\MongoDB\Tests\Functional\Ticket\GH1141Test::testReplacementOfEmbedManyElements
MongoWriteConcernException: localhost:27017: Cannot update 'chapters.0.name' and 'chapters' at the same time

/mongodb-odm/vendor/doctrine/mongodb/lib/Doctrine/MongoDB/Collection.php:1335
/mongodb-odm/vendor/doctrine/mongodb/lib/Doctrine/MongoDB/Collection.php:783
/mongodb-odm/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php:407
/mongodb-odm/lib/Doctrine/ODM/MongoDB/UnitOfWork.php:1205
/mongodb-odm/lib/Doctrine/ODM/MongoDB/UnitOfWork.php:453
/mongodb-odm/lib/Doctrine/ODM/MongoDB/DocumentManager.php:526
/mongodb-odm/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1141Test.php:34

FAILURES!
Tests: 1, Assertions: 0, Errors: 1.
```

PR #1142, opened 2015-06-13
Check if embedded document is not part of atomic set query

malarzm:
Closes #1141 (@blockjon's test is included in this PR)

For the record it was working for `set` strategy but there were 2 queries issued - first to update `chapters.0.status` from `DocumentPersister` and later to update whole `chapters` collection from `CollectionPersister`. I think it's worth fixing, but I've ran into some problems and decided it's outside of scope of this PR.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
@@ -187,7 +187,7 @@ class PersistenceBuilder
             } elseif (isset($mapping['association']) && $mapping['association'] === ClassMetadata::EMBED_MANY) {
                 if (null !== $new) {
                     foreach ($new as $key => $embeddedDoc) {
-                        if ( ! $this->uow->isScheduledForInsert($embeddedDoc)) {
+                        if ( ! $this->uow->isScheduledForInsert($embeddedDoc) && ! $this->isPartOfAtomicUpdate($embeddedDoc)) {
                             $update = $this->prepareUpdateData($embeddedDoc);
                             foreach ($update as $cmd => $values) {
                                 foreach ($values as $name => $value) {
@@ -267,7 +267,7 @@ class PersistenceBuilder
             // @EmbedMany
             } elseif (isset($mapping['association']) && $mapping['association'] === ClassMetadata::EMBED_MANY && $new) {
                 foreach ($new as $key => $embeddedDoc) {
-                    if ( ! $this->uow->isScheduledForInsert($embeddedDoc)) {
+                    if ( ! $this->uow->isScheduledForInsert($embeddedDoc) && ! $this->isPartOfAtomicUpdate($embeddeDoc)) {
                         $update = $this->prepareUpsertData($embeddedDoc);
                         foreach ($update as $cmd => $values) {
                             foreach ($values as $name => $value) {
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
```diff
@@ -485,4 +485,13 @@ private function isScheduledForInsert($document)
         return $this->uow->isScheduledForInsert($document)
             || $this->uow->getDocumentPersister(get_class($document))->isQueuedForInsert($document);
     }
+
+    private function isPartOfAtomicUpdate($embeddeDoc)
```
jmikola: Since we're also calculating this in the `DocumentPersister::getAtomicCollectionUpdateQuery()`, perhaps it makes sense to move this to a public UnitOfWork method (annotated with `@internal`)? Something like `isWithinAtomicCollection()` with an `$embeddedDocument` argument seems descriptive enough.

lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
```diff
@@ -187,7 +187,7 @@ public function prepareUpdateData($document)
             } elseif (isset($mapping['association']) && $mapping['association'] === ClassMetadata::EMBED_MANY) {
                 if (null !== $new) {
                     foreach ($new as $key => $embeddedDoc) {
-                        if ( ! $this->uow->isScheduledForInsert($embeddedDoc)) {
+                        if ( ! $this->uow->isScheduledForInsert($embeddedDoc) && ! $this->isPartOfAtomicUpdate($embeddedDoc)) {
```
jmikola: Why is this change not also required for `prepareUpsertData()`?

lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
malarzm: I was wondering about that but in `DocumentPersister::getAtomicCollectionUpdateQuery()` we want to get mapping back while here true/false is enough and I don't think `bool|array` is good idea for is-ser. Maybe method like `getTopmostCollectionMapping` returning `array|null` would be better? Sometimes I hate naming things

blockjon:
So for the first time, all of my unit and integration tests have passed.That's very exciting as I have been trying for a long time to fix our data corruption problems. I will try this on my canary server this week and see if things actually work correctly but it appears that I'm in really good shape.

lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
malarzm: Good question :) Will add condition there

malarzm:
@blockjon awesome :+1:

lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
jmikola: Ah, good point about requiring the mapping. Let's skip this for now, and optionally leave a `// TODO: ...` comment in place to refactor it later.

lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
malarzm: Added

jmikola:
LGTM :+1:

malarzm:
Merged manually in https://github.com/doctrine/mongodb-odm/commit/10cd9169d5075d97da17316b86ea9fc00c9f0cdd

PR #1143, opened 2015-06-13
Adding a note that the indexes will not be created for embedded documents

bicatu:
Adding a note that the indexes will not be created for embedded documents if you mix types unless you also use the discriminatorMap on the main document.

### DIFF
--- a/docs/en/reference/indexes.rst
+++ b/docs/en/reference/indexes.rst
@@ -317,7 +317,12 @@ database:
 Also, for your convenience you can create the indexes for your mapped documents from the
 :doc:`console <console-commands>`:
 
+..
+
     $ php mongodb.php mongodb:schema:create --index
+
+**Important note**: If you are mixing types for your embedded documents, the indexes will not be
+created unless you specify the ``@discriminatorMap`` annotation on the main document.  See more at :doc:`Embedded mapping <embedded-mapping>`.
 
 Geospatial Indexing
 -------------------


docs/en/reference/indexes.rst
```diff
@@ -317,8 +317,13 @@ database:
 Also, for your convenience you can create the indexes for your mapped documents from the
 :doc:`console <console-commands>`:
 
+..
+
     $ php mongodb.php mongodb:schema:create --index
 
+**Important note**: If you are mixing types for your embedded documents, the indexes will not be
+created unless you specify the ``@discriminatorMap`` annotation on the main document.  See more at :doc:`Embedded mapping <embedded-mapping>`.
```
jmikola: Should this instead be referring to the [Mixing Document Types](http://docs.doctrine-project.org/projects/doctrine-mongodb-odm/en/latest/reference/embedded-mapping.html#mixing-document-types) section? I assume you're referring to [this line](https://github.com/doctrine/mongodb-odm/blob/1eaba57d84c2d7702966f2fe253aba1244d76959/lib/Doctrine/ODM/MongoDB/SchemaManager.php#L170) in SchemaManager, where we attempt to collect all possible embedded documents in a discriminator map.

jmikola:
On a related note, it appears that we don't index discriminator fields by default. I believe that has come up before and the suggestion is that users can index it if they really want to, but there's not much reason to create those single-field indexes by default since they're no very selective.

For some reason, that's what I originally though this ticket was about before I realized it was about root documents collecting indexed fields from their embedded children.

malarzm:
> but there's not much reason to create those single-field indexes by default since they're no very selective.

Given that Mongo 2.6 can intersect indexes there may be some value in creating the index for discriminator field?

jmikola:
> Given that Mongo 2.6 can intersect indexes there may be some value in creating the index for discriminator field?

Alas, the index intersection isn't as useful as you might think from the name (at least not yet :)

docs/en/reference/indexes.rst
jmikola: @bicatu: Ping

bicatu: Yes, I am referring to the fact that unless I specify the discriminatorMap in the annotation the index will not be created.  I mapped to the main section.  If it is possible to link directly to the sub-section I can update.

jmikola:
Manually merged in https://github.com/doctrine/mongodb-odm/commit/8ab88fec3c04e4132c1262d3d3a4a0550879c881 with some revisions. Thanks, @bicatu!

PR #1144, opened 2015-06-13
Added type checks for Collection and Hash types

### LINKED ISSUE #967: Add type check to HashType
I was in a hurry today and accidentally put ``$this->html = new ArrayCollection()`` in my Document's ``__construct()`` having ``html`` mapped as ``Hash`` and it has ended in not very helpful exception during flush: ``zero-length keys are not allowed, did you use $ with double quotes?``. Of course it was my mistake and I should have used plain array but maybe we could add some type check to https://github.com/doctrine/mongodb-odm/blob/master/lib/Doctrine/ODM/MongoDB/Types/HashType.php#L34 ? I suppose it could throw exception on ``if ($value !== null && !is_array($value))`` or run ``$value->toArray()`` if it detects ``ArrayCollection``. Personally I'm more for exception because if I use ``ArrayCollection`` in Document then most probably I want to have it back not as plain array after fetching from db. What do you think?

malarzm:
Closes #967

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/MongoDBException.php
+++ b/lib/Doctrine/ODM/MongoDB/MongoDBException.php
@@ -27,46 +27,83 @@ namespace Doctrine\ODM\MongoDB;
  */
 class MongoDBException extends \Exception
 {
+    /**
+     * @param string $documentName
+     * @param string $fieldName
+     * @param string $method
+     * @return MongoDBException
+     */
     public static function invalidFindByCall($documentName, $fieldName, $method)
     {
         return new self(sprintf('Invalid find by call %s::$fieldName (%s)', $documentName, $fieldName, $method));
     }
 
+    /**
+     * @return MongoDBException
+     */
     public static function detachedDocumentCannotBeRemoved()
     {
         return new self('Detached document cannot be removed');
     }
 
+    /**
+     * @param string $state
+     * @return MongoDBException
+     */
     public static function invalidDocumentState($state)
     {
         return new self(sprintf('Invalid document state "%s"', $state));
     }
 
+    /**
+     * @param string $className
+     * @return MongoDBException
+     */
     public static function documentNotMappedToCollection($className)
     {
         return new self(sprintf('The "%s" document is not mapped to a MongoDB database collection.', $className));
     }
 
+    /**
+     * @return MongoDBException
+     */
     public static function documentManagerClosed()
     {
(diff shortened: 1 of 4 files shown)


lib/Doctrine/ODM/MongoDB/MongoDBException.php
```diff
@@ -75,8 +112,36 @@ public static function queryNotIndexed($className, $unindexedFields)
...
+     * @param mixed $got
+     * @return MongoDBException
+     */
+    public static function invalidValueForType($type, $expected, $got)
+    {
+        if (is_array($expected)) {
+            $expected = sprintf("%s or %s",
+                join(', ', array_slice($expected, 0, -1)),
+                end($expected)
+            );
```
jmikola: Clever :)

lib/Doctrine/ODM/MongoDB/Types/HashType.php
```diff
@@ -31,6 +33,9 @@ class HashType extends Type
 {
     public function convertToDatabaseValue($value)
     {
+        if ($value !== null && ! is_array($value)) {
+            throw MongoDBException::invalidValueForType('Hash', array('array', 'null'), $value);
```
jmikola: Is there any reason we shouldn't allow objects here, since we're just going to cast it below anyway?

lib/Doctrine/ODM/MongoDB/MongoDBException.php
```diff
@@ -27,46 +27,83 @@
  */
 class MongoDBException extends \Exception
 {
+    /**
+     * @param string $documentName
+     * @param string $fieldName
+     * @param string $method
+     * @return MongoDBException
+     */
```
jmikola: No need to split this up now, but in the future it'd probably be best to keep these changes in their own commit (e.g. "Add missing exception doc blocks").

jmikola:
Decide if Hash should allow objects on the way in (your call).

Beyond that, LGTM :+1:

malarzm:
I'd rather be strict and force to use array here, for instance passing `ArrayCollection` doesn't end well (see https://github.com/doctrine/mongodb-odm/issues/967 which was main motivation to do it :) )

lib/Doctrine/ODM/MongoDB/MongoDBException.php
malarzm: Will remember next time :+1:

PR #1146, opened 2015-06-13
UnitOfWork cleanup

malarzm:


jmikola:
You and @jwage are absolutely sure that CommitOrderCalculator is obsolete? :smile:

jwage:
We don't need it, but just noting that without it, the order in which documents are inserted and updated will change. This order could be an assumption that someone is relying on.

malarzm:
That assumption was very dangerous given that there could be more than one valid commit order :)

malarzm:
Merged manually in https://github.com/doctrine/mongodb-odm/commit/9addf539dfbdbe83799abbf90f70503eb36ef3e3

PR #1147, opened 2015-06-14
Forbid changing identifier field's type

### LINKED ISSUE #745: Do not allow child classes to override identifiers into regular fields
See GH435Test in https://github.com/doctrine/mongodb-odm/commit/afe6e92543eac6104694048fb6a1b49cd1cc7f91. It's a bit weird that we allow a child class to override a parent such that there is no longer an identifier field mapping. The child class still tracks an identifier field in its `$identifier` property and there is a generator. This could lead to undefined behavior.

There's no problem with children changing the ID mapping, but the field should remain an identifier.

malarzm:
Closes #745


alcaeus:
LGTM :+1:

jmikola:
:+1:

malarzm:
Merged manually in https://github.com/doctrine/mongodb-odm/commit/b1e83f340dc9f992c8c26946feafd2642325e11b

Issue #949, opened 2014-09-25
Cannot create a DBRef without an identifier when saving


palmabit:
Hello, i've encountered this error when i'm trying to update an existing user: Cannot create a DBRef without an identifier. UnitOfWork::getDocumentIdentifier() did not return an identifier for class Clickm3\Product\Classes\Category.

Here is my models(i've cleaned them a little bit):

```php
/** @ODM\Document(collection="user")
 * @ODM\HasLifecycleCallbacks
 * @ODM\InheritanceType("SINGLE_COLLECTION")
 * @ODM\DiscriminatorField("user_type")
 * @ODM\DiscriminatorMap({"user"="Clickm3\Authentication\Classes\User", "company"="Clickm3\Authentication\Classes\UserCompany"})
 * @ODM\DiscriminatorMap({"user"="Clickm3\Authentication\Classes\User", "visitor"="Clickm3\Authentication\Classes\UserVisitor"})
 */
class User{
  /*
   * User types
   */
  public static $USER_BASE = "user";
  public static $USER_DESIGNER = "designer";
  public static $USER_COMPANY = "company";
  public static $USER_AGENT = "agent";
  public static $USER_SPONSOR = "sponsor";
  public static $USER_ADMIN = "admin";
  public static $USER_VISITOR = "visitor";

  /*
  |--------------------------------------------------------------------------
  | Authentication data
  |--------------------------------------------------------------------------
  */

  /** @ODM\Id */
  public $id;

  protected $user_type;

  protected function __construct() {
    $this->user_type = User::$USER_BASE;
  }

  public static function create($user_type) {
    switch ($user_type) {
      case static::$USER_BASE :
        return new User;
      case static::$USER_DESIGNER :
        return new UserDesigner();
      case static::$USER_COMPANY :
        return new UserCompany();
      case static::$USER_AGENT :
        return new UserAgent();
      case static::$USER_SPONSOR :
        return new UserSponsor();
      case static::$USER_ADMIN :
        return new UserAdmin();
      case static::$USER_VISITOR :
        return new UserVisitor();
      default:
        return new UserVisitor();
        break;
    }
  }

  /**
   * @return mixed
   */
  public function getUserType() {
    return User::$USER_BASE;
  }
}
```

```php
/**
 * @ODM\Document
 */
class UserCompany extends User implements NotificableAdminInvitesInterface, NotificableAgentInvitesInterface, ProductAbleInterface {
  /*
  |--------------------------------------------------------------------------
[...]

Issue #1076, opened 2015-04-09
Setting the DocumentManager temporarily to read-only state


vmattila:
I'd like to share an idea about something that could setting the DocumentManager to read-only state for certain operations. When read-only document manager is used, each read document would be automatically detached at the end. Still we could explicitly merge a document if they need to be updated later.

This would allow us to use same code to fetch sets of documents without knowing for what purpose the documents are actually used in other parts of the application.

The interface could be something similar to ORM's `::transactional()`, for example:

```
$dm->readOnly(function($dm) {
    $documents = SomeOtherHelperClass::findDocuments();
    // Do something read-only with the docs
});
```

Detaching would happen at the end of the callback function (as similar to ORM's `transactional()` does flushing) for all documents that were unknown to DocumentManager at the time it entered the read-only state.

malarzm:
If I understand correctly then DocumentManager state would be snapshotted at the beginning and all documents added in read only mode would be detached later so DM's state will the same as in the beginning? It might be interesting to fetch a lot of objects with this and not care about penalty by using deferred implicit change policy if we don't want to change them anyway (manually detaching stuff would work as well but why not automate it) :)

vmattila:
Yes, this is exactly what I was thinking. I'd say that detaching could be done as soon as the object is read so it would be also available for gc as soon as possible.

malarzm:
While not on `DocumentManager` level, read-only mode will be possible in `Query` (and its `Builder`) after #1403 gets merged :)

PR #1151, opened 2015-06-16
Broken test for #949

solocommand:
This includes a requested broken test demonstrating the failure reported in #949.

When attempting to persist a document containing a reference that previously existed in the identity map (after a flush/clear) a `RuntimeException: Cannot create a DBRef without an identifier. UnitOfWork::getDocumentIdentifier() did not return an identifier for class x` exception is thrown.

I've run into this issue myself several times, as @vmattila pointed out it seems to stem from retaining an existing document that had been retrieved from the database (or persisted with a flush/clear afterwards), and attempting to persist a new document with the old document referenced (rather than retrieving it again after the clear in the UoW).

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH949Test.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH949Test.php
@@ -1,0 +1,59 @@
+<?php
+
+namespace Doctrine\ODM\MongoDB\Tests\Functional\Ticket;
+
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
+
+/**
+ * @author Josh Worden <solocommand@gmail.com>
+ * @since 6/16/15
+ */
+class GH949Test extends \Doctrine\ODM\MongoDB\Tests\BaseTest
+{
+    protected $group;
+
+    public function testDisconnectedReferencesAfterClear()
+    {
+        // Create our test docs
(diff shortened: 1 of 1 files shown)


malarzm:
Thanks for the test case @solocommand! I've scheduled it for 1.0.0 so we'll be back here soon

malarzm:
Or I'm even back now with question: @solocommand @vmattila @palmabit: why don't you specify `cascade="persist"` for reference mapping or `merge` document unknown to ODM into DocumentManager first? Both approaches are fixing provided test and seem to be clear.

solocommand:
I forgot cascading was a thing. :) If that resolves why don't we just add a
saner exception notifying the dev that the doc isn't managed/can't be
cascaded in this condition? Bad dbref message is rather unclear.

malarzm:
I'd be :+1: for saner exception here :)

vmattila:
This exception to happen, to be honest, is a kind of sign that the application design might be bad (at least it was in my case :). Definitely +1 for a better exception.

If we'd use `cascade="persist"` didn't this lead to re-persisting the referenced document to database, that most likely ends up to an error due duplicate `_id`?

How about ODM trying to `merge` the detached document automatically if the `_id` is known? I did recently submitted an idea #1076 that could also benefit of this behaviour.

malarzm:
@vmattila persisting object with id results in upsert so nothing bad should happen, although I believe that merging is beter ;)

malarzm:
#1163 was merged  manually in https://github.com/doctrine/mongodb-odm/commit/43a279633b2ea1230586f660a927b9fdd6c60935

PR #1152, opened 2015-06-18
Store parent association before hydrating document

alcaeus:
This change stores the parent association in UnitOfWork before hydrating data into an embedded documents. This allows postLoad subscribers to use getParentAssociation to get information about the containing object.


malarzm:
:+1:

malarzm:
But also having a test for this would be nice ;)

malarzm:
Merged manually in https://github.com/doctrine/mongodb-odm/commit/3fcc8cfbce7d29f18fbba32e192ef5713d366d1e

PR #1153, opened 2015-06-18
Functional test: atomicSet updates to embedded documents do not work

blockjon:
Making a change to the embedded document in this test fails to update the MongoDB record.

This test passes if strategy="set" is used.

repro:
```
phpunit --filter testFailureToUpdateEmbeddedDocumentWithAtomicSet
```
emits:
```
PHPUnit 4.1.6-9-gbe3530d by Sebastian Bergmann.

Configuration read from /mongodb-odm/phpunit.xml

F

Time: 1.41 seconds, Memory: 21.75Mb

There was 1 failure:

1) Doctrine\ODM\MongoDB\Tests\Functional\Ticket\GHXXXXTest::testFailureToUpdateEmbeddedDocumentWithAtomicSet
The chapter title failed to update.
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'First chapter A'
+'A'

/mongodb-odm/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GHXXXXTest.php:33

FAILURES!
Tests: 1, Assertions: 1, Failures: 1.
```

PR #1154, opened 2015-06-20
Fix updating of embedded documents if their atomic collection is not dirty

malarzm:
Fixes #1153 

For the record this PR can be considered as a hotfix for functionality that I've accidentally broken fixing #1141. I start to dislike the way we're managing atomic collection updates but I'm planning to fix it along with commit() flow I've described in https://github.com/doctrine/mongodb-odm/pull/1128#issuecomment-111864297 so I'm willing to merge this now and rewrite later.


jwage:
:+1:

PR #1155, opened 2015-06-21
Commit flow rework and fixed collection logic

### LINKED ISSUE #1046: Investigate Collection update and locking document

### LINKED ISSUE #1145: @postUpdate events trigger at an unexpected moment when using EmbedMany
My project uses strategy=set for nearly all of our EmbedMany elements and we also use @postUpdate hooks all over the place. 

Given a document (such as book) with many embedded documents (such as chapters), if you make changes to both the "surface" of the document (such as the book title) and then you also modify the chapters at the same time and then call flush(), the sequence of queries and postUpdate callbacks seems wrong.

When you flush() a change as I have described above, the sequence of events that takes place looks like this:

1) MongoDB update query to update the surface of the db record (the book title).
2) postUpdate hooks fired for *both* the surface and also for the embedded documents
3) MongoDB update query to update the embedded documents

Imagine an app where you update the title of a book and modify a chapter in one flush() such that after the chapter is edited, you email alert the publisher via a @postUpdate hook on the Chapter document.

Because of the sequence of the events, the email is sent *before* the chapter is actually saved to the database server. This means that if there's a fatal MongoDB error on step 3, the email was still sent and the user is disappointed because the application sent the email and the data was never saved.

The whole idea of a "postUpdate" hook is that the the event would fire *after* the database record was saved....not before it was saved.

malarzm:
As per my comment in https://github.com/doctrine/mongodb-odm/pull/1128#issuecomment-111864297 this PR takes care of following things:

- [x] Collections deletions/updates/snapshots should take place immediately after owning document query and before dispatching events (fixes #1145, fixes case when collection is not saved due to another document's LockException) (collections modified in postUpdate are no longer saved to database in same commit() which may be a BC break but is correct from our point of view, since postUpdate was meant to be triggered after updating document in database)
- [x] Simplify handling atomic collections by moving relevant logic to `PersistenceBuilder` (thing I've mentioned in #1154)
- [x] If it's possible include recursively collections during document's insert
- [x] Fix case when deeply nested collections weren't saved properly by recursively including collections if `set` or `setArray` strategies are used (fixes #1104)
- [x] Possible superflous queries updating/deleting collections of to be deleted document are no longer issued
- [x] Move tests for atomic collections from Ticket folder to one place, ~~increase number of tests for atomic strategies by doubling existing tests for collections~~ (which will be done after review to not blur the diff) EDIT: looking at current tests they seem to be enough combined with how atomic strategies are handled now

This PR doesn't fix #1046 but makes it bloody easy to implement, all that is needed is to pass current document's version to CollectionPersister and take it into account in update query. I can include this here if we decide it's good idea or leave it for another PR.


blockjon:
Something is wrong with this PR but I'm having trouble coming up with a functional test to exhibit the behavior.

In my application flow, there is a document with two strategy="set" EmbedMany elements. If I use the mongodb-odm master line, adding to these two EmbedMany elements works normally. When I use this branch, the elements are not appended to.

Although I can't figure out how to repro the problem in a functional test, I can show you the queries my app is firing when on the master line versus this branch:

mongodb-odm master line queries:
```
{"update":true,"query":{"_id":{"$id":"5587d3562e6be58a048b456a"},"revision":3},"newObj":{"$set":{"defaultChargeAccountId":"chargeAccount_stripe_card_user_741572580320049154_dWJ3LS2afUbo75Yg"},"$inc":{"revision":1}},"options":{"w":1},"db":"api-development","collection":"users"}
{"update":true,"query":{"_id":{"$id":"5587d3562e6be58a048b456a"}},"newObj":{"$set":{"chargeAccounts":[{"id":"chargeAccount_stripe_card_user_741572580320049154_dWJ3LS2afUbo75Yg","customerId":"cus_6TNsmtkHmzGlL5","provider":"stripe","method":"card","label":"Personal","labelType":"personal","createdTimestamp":1434964830,"source":"user","cardId":"card_6TNs3YPmftl4QF","ccLastFour":"9424","ccFingerprint":"dWJ3LS2afUbo75Yg","ccType":"Discover","ccDescription":"Consumer Credit - Rewards","ccBrand":"Discover","ccFunding":"credit","ccCountry":"US","validCvc":"pass","expMonth":12,"expYear":2022,"addressZip":"","billingType":"credit","ccIin":"601100"}]}},"options":{"w":1},"db":"api-development","collection":"users"}
{"update":true,"query":{"_id":{"$id":"5587d3562e6be58a048b456a"}},"newObj":{"$set":{"chargeAccountsHistory":[{"id":"chargeAccount_stripe_card_user_741572580320049154_dWJ3LS2afUbo75Yg","customerId":"cus_6TNsmtkHmzGlL5","provider":"stripe","method":"card","label":"Personal","labelType":"personal","createdTimestamp":1434964830,"source":"user","cardId":"card_6TNs3YPmftl4QF","ccLastFour":"9424","ccFingerprint":"dWJ3LS2afUbo75Yg","ccType":"Discover","ccDescription":"Consumer Credit - Rewards","ccBrand":"Discover","ccFunding":"credit","ccCountry":"US","validCvc":"pass","expMonth":12,"expYear":2022,"addressZip":"","billingType":"credit","ccIin":"601100"}]}},"options":{"w":1},"db":"api-development","collection":"users"}
```
the same code on this branch:
```
{"update":true,"query":{"_id":{"$id":"5587d3562e6be58a048b456a"},"revision":3},"newObj":{"$set":{"defaultChargeAccountId":"chargeAccount_stripe_card_user_741572580320049154_dWJ3LS2afUbo75Yg"},"$inc":{"revision":1}},"options":{"w":1},"db":"api-development","collection":"users"}
{"update":true,"query":{"_id":{"$id":"5587d3562e6be58a048b456a"}},"newObj":{"$set":{"chargeAccounts":[{"id":"chargeAccount_stripe_card_user_741572580320049154_dWJ3LS2afUbo75Yg","customerId":"cus_6TNsmtkHmzGlL5","provider":"stripe","method":"card","label":"Personal","labelType":"personal","createdTimestamp":1434964830,"source":"user","cardId":"card_6TNs3YPmftl4QF","ccLastFour":"9424","ccFingerprint":"dWJ3LS2afUbo75Yg","ccType":"Discover","ccDescription":"Consumer Credit - Rewards","ccBrand":"Discover","ccFunding":"credit","ccCountry":"US","validCvc":"pass","expMonth":12,"expYear":2022,"addressZip":"","billingType":"credit","ccIin":"601100"}]}},"options":{"w":1},"db":"api-development","collection":"users"}
```

blockjon:
I'm still trying to figure out how the UnitOfWork works.. but it seems like the chargeAccounts and chargeAccountsHistory elements are not considered to be scheduled collections when using this branch. That's just a guess.

malarzm:
I'll try to hack a failing test case in few hours, at a glance it seems strange as `set` collections should still be handled in old manner... maybe it's related to some `postUpdate` flow which is modifying collections?

blockjon:
It does not look like my app has any postUpdate events in the failure situation.

blockjon:
@malarzm I'm out of ideas.. I tried to debug this for 3-4 hours but I'm out of ideas.

malarzm:
@blockjon could you maybe bisect this PR? Maybe this will give a hint

blockjon:
Yes I can do this in about 10 hours since I'm about to go to sleep. :)

malarzm:
@blockjon sure thing is, have a good one ;)

blockjon:
@malarzm Bisect reveals that my problem is introduced in https://github.com/doctrine/mongodb-odm/commit/3313e652d3a640ca25bf1a0791b9f3c1e4be9581

malarzm:
@blockjon blind guess, but could you comment out `unset($this->hasScheduledCollections[spl_object_hash($topmostOwner)][$oid]);` from `Uow::unscheduleCollection(Deletion|Update)` and see if it fixes your issue?

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -2575,27 +2561,21 @@ public function scheduleCollectionDeletion(PersistentCollection $coll)
...
      */
     public function unscheduleCollectionDeletion(PersistentCollection $coll)
     {
-        if (($key = array_search($coll, $this->collectionDeletions, true)) !== false) {
-            unset($this->collectionDeletions[$key]);
-        }
+        $oid = spl_object_hash($coll);
+        $topmostOwner = $this->getOwningDocument($coll->getOwner());
+        unset($this->collectionDeletions[$oid]);
+        unset($this->hasScheduledCollections[spl_object_hash($topmostOwner)][$oid]);
```
blockjon: @malarzm When I comment this line out, the my problem goes away!!

malarzm: Oh crap I think I know what I did wrong here... I'll push a fix in few minutes, hopefully it'll be that :)

malarzm:
@blockjon can you check now?

blockjon:
@malarzm works!

blockjon:
I will now re-run all of my automatic and manual smoke tests...

malarzm:
@blockjon is no message a good message? :)

blockjon:
@malarzm I just heard back from QA.. they told me all tests pass. I am therefore interested in rolling this out to my production system. Is this in a fairly stable state now in your opinion?

malarzm:
@blockjon I do hope it is, with new way of handling atomic collections it should be way more stable than original implementation (and once again thank you for all failing test cases for that stuff) :) For a head up I am going to merge #782 tomorrow (or tonight) and then rebase this PR (but I don't expect tests to start failing because of that)

malarzm:
Ufff finally rebased

blockjon:
Did you merge? I would like to run my test suite once you think this PR is a candidate for a merge back to master.

malarzm:
@blockjon sure, will let you know :) I plan to merge this PR by the end of this week, hoping for some tests before that

malarzm:
FWIW my app's test suite has passed

blockjon:
I have been doing extensive tests against https://github.com/doctrine/mongodb-odm/commit/39ce685b40bface6b70b56fac4f2fd43dd9d70ac for a week and it works so I am just now rolling it out to production. I'll replace with something better once this PR is back on the master line.

lib/Doctrine/ODM/MongoDB/Utility/CollectionHelper.php
```diff
@@ -0,0 +1,75 @@
...
+ * <http://www.doctrine-project.org>.
+ */
+
+namespace Doctrine\ODM\MongoDB\Utility;
+
+/**
+ * Utility class used to unify checks on how collection strategies should behave.
+ *
+ * @since   1.0
+ * @author  Maciej Malarz <malarzm@gmail.com>
```
jmikola: Should we label this as `@internal`, or do you expect to allow this as a supported, public API?

malarzm:
@blockjon ah, this is already on top of #782 (and current master)

lib/Doctrine/ODM/MongoDB/Utility/CollectionHelper.php
malarzm: Yes we should :+1:

blockjon:
@malarzm What I mean is that when this PR is done done, you'll merge the PR back to the master line. At that time, I will re-run composer and I'll point at the doctrine master line.

alcaeus:
Just ran this against my test suite, it passes :+1:
Yeah!

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataInfo.php
```diff
@@ -1155,8 +1156,7 @@ public function mapField(array $mapping)
             throw MappingException::owningAndInverseReferencesRequireTargetDocument($this->name, $mapping['fieldName']);
         }
         
-        if ($this->isEmbeddedDocument && isset($mapping['strategy']) &&
-                ($mapping['strategy'] === 'atomicSet' || $mapping['strategy'] === 'atomicSetArray')) {
+        if ($this->isEmbeddedDocument && $mapping['type'] === 'many' && CollectionHelper::isAtomic($mapping['strategy'])) {
```
jmikola: Does this still need an `isset($mapping['strategy'])` or are we always sure it's set (like `$mapping['type']`)?

lib/Doctrine/ODM/MongoDB/PersistentCollection.php
```diff
@@ -193,7 +193,7 @@ public function initialize()
         // Reattach any NEW objects added through add()
         if ($newObjects) {
             foreach ($newObjects as $key => $obj) {
-                if ($this->mapping['strategy'] === 'set' || $this->mapping['strategy'] === 'atomicSet') {
+                if (Utility\CollectionHelper::isHash($this->mapping['strategy'])) {
```
jmikola: I realize we're already in the `Doctrine\ODM\MongoDB` namespace here, but would it be clearer to add the `use` statement for CollectionHelper at the top of the file anyway, so we can tell at a glance what classes outside of this namespace are being referenced?

lib/Doctrine/ODM/MongoDB/Persisters/CollectionPersister.php
```diff
@@ -256,7 +212,7 @@ private function insertElements(PersistentCollection $coll, array $options)
 
         $value = array_values(array_map($callback, $insertDiff));
 
-        if ($mapping['strategy'] !== 'pushAll') {
+        if ($mapping['strategy'] === 'addToSet') {
```
jmikola: While `$pushAll` didn't use `$each`, `$push` and `$addToSet` both use it to denote multiple arguments for insertion (otherwise the array argument itself would be pushed as a single element). Are you sure this is correct?

jmikola: Looking down at `PersistenceBuilder::prepareInsertData()`, I saw you added:

> We're excluding collections using addToSet since there is a risk of duplicated entries stored in the collection

In that case, is this `CollectionPersister::insertElements()` ever called for strategies other than `$addToSet`? Also, I'm curious if we can just get away with using `array_unique()` on the PHP side instead of relying on MongoDB itself to detect duplicates.

I assume the ideal case is being able to `$set` the collection data instead of using any array update operators (be it `$push` or `$addToSet`).

lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
```diff
@@ -107,6 +99,13 @@ public function prepareInsertData($document)
             // @EmbedOne
             } elseif (isset($mapping['association']) && $mapping['association'] === ClassMetadata::EMBED_ONE) {
                 $insertData[$mapping['name']] = $this->prepareEmbeddedDocumentValue($mapping, $new);
+
+            // @ReferenceMany, @EmbedMany
+            // We're excluding collections using addToSet since there is a risk
+            // of duplicated entries stored in the collection
```
jmikola: Cross-referencing this with my [earlier comment](https://github.com/doctrine/mongodb-odm/pull/1155/files#r33710131) on `CollectionPersister::prepareInsertData()`.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -492,8 +478,9 @@ private function computeScheduleInsertsChangeSets()
     {
         foreach ($this->documentInsertions as $document) {
             $class = $this->dm->getClassMetadata(get_class($document));
-
-            $this->computeChangeSet($class, $document);
+            if ( ! $class->isEmbeddedDocument) {
```
jmikola: Is this just a sanity check? I would think elsewhere we'd ensure that embedded documents would never show up in `$documentInsertions`.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -506,8 +493,9 @@ private function computeScheduleUpsertsChangeSets()
     {
         foreach ($this->documentUpserts as $document) {
             $class = $this->dm->getClassMetadata(get_class($document));
-
-            $this->computeChangeSet($class, $document);
+            if ( ! $class->isEmbeddedDocument) {
```
jmikola: Same comment as [`$documentInsertions` above](https://github.com/doctrine/mongodb-odm/pull/1155/files#r33710450).

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -748,7 +736,7 @@ private function computeOrRecomputeChangeSet(ClassMetadata $class, $document, $r
                     /* If original collection was exchanged with a non-empty value
                      * and $set will be issued, there is no need to $unset it first
                      */
-                    if ($actualValue && $actualValue->isDirty() && in_array($class->fieldMappings[$propName]['strategy'], array('set', 'setArray', 'atomicSet', 'atomicSetArray'))) {
+                    if ($actualValue && $actualValue->isDirty() && Utility\CollectionHelper::usesSet($class->fieldMappings[$propName]['strategy'])) {
```
jmikola: `use Doctrine\MongoDB\ODM\Utility\CollectionHelper` atop the file?

tests/Doctrine/ODM/MongoDB/Tests/Mapping/Driver/AbstractDriverTest.php
```diff
@@ -94,7 +94,6 @@ public function testDriver()
             'isInverseSide' => false,
             'isOwningSide' => true,
             'nullable' => false,
-            'strategy' => 'pushAll',
```
jmikola: I assume this was removed because `collection` types (BSON arrays of arbitrary, unmapped values in MongoDB) no longer use strategies at all and you cleaned up the mapping?

lib/Doctrine/ODM/MongoDB/Utility/CollectionHelper.php
```diff
@@ -0,0 +1,76 @@
...
+    
+    /**
+     * Returns whether strategy uses $set to update its data.
+     * 
+     * @param string $strategy
+     * @return bool
+     */
+    public static function usesSet($strategy)
+    {
+        return in_array($strategy, array('set', 'setArray', 'atomicSet', 'atomicSetArray'));
```
jmikola: Just curious, but given all of these changes, is there really any reason users might still want `set` and `setArray` over the atomic variants? Should we consider merging them for 1.0.0?

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -403,13 +398,14 @@ public function update($document, array $options = array())
                 }
             }
 
-            unset($update['$set']['_id']);
```
jmikola: I wasn't able to trace this line back to its original commit (there were numerous style-change commits while working my way back through the history), but I think this is related to 374efd4c51f646347c00468ff5a44f87df25cccc and should not be removed.

There may be an edge case where this leaves our update document as `{$set:{}}`, which means there's nothing to do at all; however, it doesn't look like the code before your PR was handling that.

malarzm: https://github.com/doctrine/mongodb-odm/blob/b1089b723df8ae2230694a3899941e05d560c41f/lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php#L139 seems to ensure we won't get _id in update

lib/Doctrine/ODM/MongoDB/Persisters/CollectionPersister.php
malarzm: > While $pushAll didn't use $each, $push and $addToSet both use it to denote multiple arguments for insertion

I think we don't have strategy using `$push`? This method is used for `pushAll` and `addToSet` basing on [this switch](https://github.com/doctrine/mongodb-odm/blob/b1089b723df8ae2230694a3899941e05d560c41f/lib/Doctrine/ODM/MongoDB/Persisters/CollectionPersister.php#L131-L132)

> In that case, is this CollectionPersister::insertElements() ever called for strategies other than $addToSet?

Yes, for updates and upserts. This optimization is made only for inserts since there's absolutely no need to rseparately insert document and then its collections

> Also, I'm curious if we can just get away with using array_unique() on the PHP side

I had high hopes for that but apparently `array_unique()` works only for strings :( I was thinking about implementing something for this case but it'd be overengineering

> I assume the ideal case is being able to $set the collection data instead of using any array update operators (be it $push or $addToSet).

I'm not sure if we really want to do that, `$set` could overwrite data that concurrent request has just wrote to db and if developer is not using document versioning he won't even know about it while with `$push` the item will be just appended to array?

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
malarzm: #782 allowed that actually, here we're just ensuring embedded document is not entry point for change set calculation (there's edge case when embedded document doesn't know it's parent yet here)

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
malarzm: Same answer as [above](https://github.com/doctrine/mongodb-odm/pull/1155#discussion_r33723718) :)

tests/Doctrine/ODM/MongoDB/Tests/Mapping/Driver/AbstractDriverTest.php
malarzm: Exactly, this even fitted PRs intentions

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataInfo.php
malarzm: Theoretically all drivers are putting default strategy to mapping if it's not set so I believe this is safe... no test was harmed during the change ;)

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
malarzm: Actually #782 has only removed the if I added back here, it was allowed for a long time and I remember it is even used in PersistenceBuilder's prepareUpdate

lib/Doctrine/ODM/MongoDB/Utility/CollectionHelper.php
malarzm: The only advantage of using `set`/`setArray` over their atomic siblings is that the old strategies can be used inside of collection that is using `pushAll`/`addToSet` strategies

lib/Doctrine/ODM/MongoDB/Persisters/CollectionPersister.php
jmikola: > I'm not sure if we really want to do that, $set could overwrite data that concurrent request has just wrote to db and if developer is not using document versioning he won't even know about it while with $push the item will be just appended to array?

Was that still a concern if we were doing an insert, though?

malarzm: I'm not sure what do you mean :baby: For inserting documents previous flow was 1) Insert documents 2) Update their collections (effectively insert all the stuff from them) and now it's just 1) Insert documents with all the collections already. Also since we're doing an insert there's no chance for overwriting anything, especially with new flow (I can imagine bad scenario for previous flow though)

jmikola: I'm a bit confused myself. This `insertElements()` method is only called when updating a collection with `addToSet` or `pushAll`, and only `addToSet` needs `$each` -- so that clears up the original question. I'm not sure why I asked, "is `CollectionPersister::insertElements()` ever called for strategies other than `$addToSet`" above.

No worries.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -355,42 +363,29 @@ public function update($document, array $options = array())
         $id = $this->uow->getDocumentIdentifier($document);
         $update = $this->pb->prepareUpdateData($document);
 
-        if ( ! empty($update) || $this->uow->hasScheduledCollections($document)) {
```
jmikola: Not sure why I didn't ask this earlier, but this seems like we might bump the version of a document even if there is no other update to apply. I realize you want to ensure `handleCollections()` is called at the bottom of this method, but those collection modifications will take place in separate write operations AFAIK.

Is the idea here that you want to verify the main document's write so we can throw a LockException before handling collections?

I believe there's still a concern about the collection updates not incorporating versioning/locking in their own write ops; however, that's not a new problem at all.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
jmikola: :+1:

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
malarzm: This is exactly the previous behaviour if only collection was to be updated (IIRC this was the very thing why `hasScheduledCollections` was introduced) - bump version and later update collections. From PR description:

> This PR doesn't fix #1046 [investigate locking and collection updates] but makes it bloody easy to implement, all that is needed is to pass current document's version to CollectionPersister and take it into account in update query.

How it will be done is yet to be determined, however now we will have both document and its collections update in one place which is way better than before :)

lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
```diff
@@ -204,10 +213,17 @@ public function prepareUpdateData($document)
                 } else {
                     $updateData['$unset'][$mapping['name']] = true;
                 }
-
-            // @ReferenceMany
-            } elseif (isset($mapping['association']) && $mapping['association'] === ClassMetadata::REFERENCE_MANY) {
-                // Do nothing right now
+            }
+        }
+        // collections that areen't dirty but could be subject to update are
```
jmikola: Typo: "aren't"

lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
```diff
@@ -184,14 +183,24 @@ public function prepareUpdateData($document)
...
-                foreach ($new as $key => $embeddedDoc) {
-                    if ( ! $this->uow->isScheduledForInsert($embeddedDoc) && ! $this->isPartOfAtomicUpdate($embeddedDoc)) {
-                        $update = $this->prepareUpdateData($embeddedDoc);
-                        foreach ($update as $cmd => $values) {
-                            foreach ($values as $name => $value) {
-                                $updateData[$cmd][$mapping['name'] . '.' . $key . '.' . $name] = $value;
+            // @ReferenceMany, @EmbedMany
+            } elseif (isset($mapping['association']) && $mapping['type'] === 'many' && $new) {
+                if (CollectionHelper::isAtomic($mapping['strategy']) && $this->uow->isCollectionScheduledForUpdate($new)) {
+                    $updateData['$set'][$mapping['name']] = $this->prepareAssociatedCollectionValue($new, true);
```
jmikola: Should we be calling `UnitOfWork::unscheduleCollectionUpdate()` here?

lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
```diff
@@ -204,10 +213,17 @@ public function prepareUpdateData($document)
...
-            } elseif (isset($mapping['association']) && $mapping['association'] === ClassMetadata::REFERENCE_MANY) {
-                // Do nothing right now
+            }
+        }
+        // collections that areen't dirty but could be subject to update are
+        // excluded from change set, let's go through them now
+        foreach ($this->uow->getScheduledCollections($document) as $coll) {
+            $mapping = $coll->getMapping();
+            if (CollectionHelper::isAtomic($mapping['strategy']) && $this->uow->isCollectionScheduledForUpdate($coll)) {
+                $updateData['$set'][$mapping['name']] = $this->prepareAssociatedCollectionValue($coll, true);
```
jmikola: Same question as https://github.com/doctrine/mongodb-odm/pull/1155/files?diff=split#r34735661. No need for `UnitOfWork::unscheduleCollectionUpdate()`?

lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
malarzm: `prepareAssociatedCollectionValue()` does that for us

lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
malarzm: Same as above: `prepareAssociatedCollectionValue()` does that for us ;)

lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
```diff
@@ -184,14 +183,24 @@ public function prepareUpdateData($document)
...
+            // @ReferenceMany, @EmbedMany
+            } elseif (isset($mapping['association']) && $mapping['type'] === 'many' && $new) {
+                if (CollectionHelper::isAtomic($mapping['strategy']) && $this->uow->isCollectionScheduledForUpdate($new)) {
+                    $updateData['$set'][$mapping['name']] = $this->prepareAssociatedCollectionValue($new, true);
+                } elseif (CollectionHelper::isAtomic($mapping['strategy']) && $this->uow->isCollectionScheduledForDeletion($new)) {
+                    $updateData['$unset'][$mapping['name']] = true;
+                    $this->uow->unscheduleCollectionDeletion($new);
+                } elseif (CollectionHelper::isAtomic($mapping['strategy']) && $this->uow->isCollectionScheduledForDeletion($old)) {
+                    $updateData['$unset'][$mapping['name']] = true;
+                    $this->uow->unscheduleCollectionDeletion($old);
```
jmikola: If I understand this correctly, we want to `$unset` here because we've already established that the "new" value is neither scheduled for update nor deletion. This means there's an old value that needs to be deleted.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
jmikola: @jwage: You originally removed this in https://github.com/doctrine/mongodb-odm/commit/c43b8dffe13a678b4b2c4e757d0c1ff3973a6be8#diff-ae2c9f4b58c612dce08a0f60ff683aecL497. Was that just clean-up to cut down on the conditional? IMO, it's reasonable to have this check.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -894,12 +883,10 @@ private function computeAssociationChanges($parentDocument, array $assoc, $value
...
             if ($topOrExistingDocument || strncmp($assoc['strategy'], 'set', 3) === 0) {
-                if ( ! $this->isCollectionScheduledForUpdate($value)) {
-                    $this->scheduleCollectionUpdate($value);
-                }
+                $this->scheduleCollectionUpdate($value);
             }
-
-            $this->visitedCollections[] = $value;
+            $topmostOwner = $this->getOwningDocument($value->getOwner());
+            $this->visitedCollections[spl_object_hash($topmostOwner)][] = $value;
```
jmikola: Looking at `UnitOfWork::getOwningDocument()`, there's an edge case where it might return null if it does not find a parent association. In that case, we won't pass an object to `spl_object_hash()` here. Perhaps `getOwningDocument()` needs to throw an exception instead?

lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
malarzm: Yes, but I can't recall right now what exactly this was fixing... Probably exchanging PersistentCollection instance with empty ArrayCollection or clearing

lib/Doctrine/ODM/MongoDB/Utility/CollectionHelper.php
jmikola: :+1:

tests/Doctrine/ODM/MongoDB/Tests/Mapping/Driver/AbstractDriverTest.php
jmikola: :+1:

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
malarzm: Right now it can't as

1. `$value->getOwner()` is a document - awesome, nothing bad can happen
2. `$value->getOwner()` is an embedded document - we might have got here only by computing that embedded document's changeset before which also takes care of parent association.

The thing that may go wrong though is manual computing change set of new embedded document which was mentioned here: https://github.com/doctrine/mongodb-odm/pull/1155/files#r33723718 and I'm not sure which path to follow: throw exception or don't allow such thing (first might be better as removing embedded docs complexity was main point of #782). If we go down the exception road I'll appreciate proposal of the message :)

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
jmikola: :+1:

lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
jmikola: :+1:

lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
jmikola: :+1:

lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
jmikola: :+1:

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -894,12 +883,10 @@ private function computeAssociationChanges($parentDocument, array $assoc, $value
 
         if ($value instanceof PersistentCollection && $value->isDirty() && $assoc['isOwningSide']) {
             if ($topOrExistingDocument || strncmp($assoc['strategy'], 'set', 3) === 0) {
-                if ( ! $this->isCollectionScheduledForUpdate($value)) {
```
jmikola: Offhand, should the `strncmp` above be using `CollectionHelper::usesSet()`, or is it purposefully ignoring atomic sets?

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -2680,14 +2679,46 @@ public function hasScheduledCollections($document)
...
+        if ( ! $this->isDocumentScheduled($document)) {
+            $this->scheduleForUpdate($document);
+        }
+    }
 
+    /**
+     * Gets topmost owning document for given embedded document. For simplicity
+     * returns given object if it happens to be a top level document.
+     *
+     * @param object $document
```
jmikola: Missing: `@return object`

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -2680,14 +2679,46 @@ public function hasScheduledCollections($document)
...
+        }
+
+        if ( ! $this->isDocumentScheduled($document)) {
+            $this->scheduleForUpdate($document);
+        }
+    }
 
+    /**
+     * Gets topmost owning document for given embedded document. For simplicity
+     * returns given object if it happens to be a top level document.
```
jmikola: We can rewrite as:

> Get the top-most owning document of a given document
>
> If a top-level document is provided, that same document will be returned. For an embedded document, we will walk through parent associations until we find a top-level document.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
jmikola: Let's also add `@throws \UnexpectedValueException when a top-level document could not be found`

Suggested message: `"Could not determine parent association for <embedded document class>"` in the case:

    if (!$parentAssociation) {
        return;
    }

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
jmikola: See: https://github.com/doctrine/mongodb-odm/pull/1155/files#r34836154

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
malarzm: For atomic sets `$topOrExistingDocument` will be `true` but changing this for consistency is good idea :+1:

malarzm:
@jmikola I'll run tomorrow over all comments once again and apply yours, thanks for all the input!

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
malarzm: Done in https://github.com/doctrine/mongodb-odm/commit/4f2713fdbf8315f8b75fa92ba1bada2f7c8b2167

malarzm:
Merged manually in https://github.com/doctrine/mongodb-odm/commit/5b0ff1c5d4a19594b778352e9bb0fc6a54ed2152

/cc @blockjon

guilhermeblanco:
http://www.reactiongifs.com/r/59c.gif

beberlei:
![](http://www.reactiongifs.com/r/59c.gif

Ocramius:
anigif_enhanced-30398-1415218613-3

Ocramius:
anigif_enhanced-30398-1415218613-3

Ocramius:
set_dan_thorn_dancing_news_sarah

drmonty:
loki

PR #1157, opened 2015-06-22
Test for #878

danez:
This is the test only for #878 which is fixed in #782.

PR is against this fix branch.


malarzm:
Thanks @danez! :+1:

PR #1160, opened 2015-06-25
Fix initializing proxies with notify change tracking policy

### LINKED ISSUE #844: Method addPropertyChangedListener load the object
Hi, 

In the proxy of my class with the NOTIFY changeTrackingPolicy, the method addPropertyChangedListener load the class.

    public function addPropertyChangedListener(\Doctrine\Common\PropertyChangedListener $listener)
    {
        $this->__load();
        return parent::addPropertyChangedListener($listener);
    }

This has as consequence that when i loaded my object that have a referenceone to another object, this another object is loaded too, even if i don't use it.

Is it possible to not load the object when adding the listener (added by the UnitOfWork) ?

Thanks,

malarzm:
Fixes #844

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
+++ b/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
@@ -1576,7 +1576,8 @@ class UnitOfWork implements PropertyChangedListener
 
         $this->identityMap[$class->name][$id] = $document;
 
-        if ($document instanceof NotifyPropertyChanged) {
+        if ($document instanceof NotifyPropertyChanged &&
+            ( ! $document instanceof Proxy || $document->__isInitialized())) {
             $document->addPropertyChangedListener($this);
         }
 
--- a/lib/Doctrine/ODM/MongoDB/Proxy/ProxyFactory.php
+++ b/lib/Doctrine/ODM/MongoDB/Proxy/ProxyFactory.php
@@ -19,6 +19,7 @@
 
 namespace Doctrine\ODM\MongoDB\Proxy;
 
+use Doctrine\Common\NotifyPropertyChanged;
 use Doctrine\Common\Proxy\AbstractProxyFactory;
 use Doctrine\Common\Proxy\ProxyDefinition;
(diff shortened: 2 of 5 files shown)


alcaeus:
:+1:

malarzm:
Ah for the record - this is not in line with ORM

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -1576,7 +1576,8 @@ public function addToIdentityMap($document)
 
         $this->identityMap[$class->name][$id] = $document;
 
-        if ($document instanceof NotifyPropertyChanged) {
+        if ($document instanceof NotifyPropertyChanged &&
+            ( ! $document instanceof Proxy || $document->__isInitialized())) {
```
jmikola: I assume this is the counterpart to https://github.com/doctrine/doctrine2/blob/1e7e8f216a899ad2ae4f38e208b5b26ce7f503ec/lib/Doctrine/ORM/UnitOfWork.php#L3104-L3106. In that case, doesn't @Ocramius' comment in https://github.com/doctrine/mongodb-odm/issues/844#issuecomment-114641431 hold true? Should the proxy initializer still be adding UnitOfWork as a property change listener on uninitialized proxies?

malarzm: I'm not entirely sure what adds UnitOfWork as a property change listener on uninitialized proxies in ORM as there's way more logic and I couldn't track it down to one exact thing, but [there is a test](https://github.com/doctrine/doctrine2/blob/900b55d16afdcdeb5100d435a7166d3a425b9873/tests/Doctrine/Tests/ORM/Functional/Ticket/DDC2230Test.php#L27) that ensures behaviour I'm implementing in this PR, I'm just putting logic in other place :)

malarzm:
Merged manually in https://github.com/doctrine/mongodb-odm/commit/e3468b2d2a8895bc2b560b69272e75772f08c815

PR #1161, opened 2015-06-27
Query's distinct field should be prepared

### LINKED ISSUE #1148: distinct() doesn't use document mapped name
I'm trying to make a distinct query to my database using queryBuilder, but I cannot use my document property name in the distinct method, instead of this I have to use the real database field name. Otherwise I have no problems using the property name in my sort or field methods.

For example:

        $qb = $this->dm->createQueryBuilder('myapp\Domain\User\Document\User');

        # I cannot use something like "$qb->field('postalCode')" or "postalCode"
        $qb->distinct('postal_code');

        # Using postalCode (property name) instead of postal_code (database name)
        $qb->addAnd($qb->expr()->field('postalCode')->equals(new \MongoRegex("/$needle/")));
        $qb->sort('postalCode', 'asc');

        $results = $qb->getQuery()->execute();

malarzm:
Fixes #1148

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Query/Builder.php
+++ b/lib/Doctrine/ODM/MongoDB/Query/Builder.php
@@ -317,6 +317,10 @@ class Builder extends \Doctrine\MongoDB\Query\Builder
 
         $query['newObj'] = $this->expr->getNewObj();
 
+        if (isset($query['distinct'])) {
+            $query['distinct'] = $documentPersister->prepareFieldName($query['distinct']);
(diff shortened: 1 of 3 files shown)


alcaeus:
LGTM :+1:

jmikola:
:+1:

malarzm:
Merged manually in https://github.com/doctrine/mongodb-odm/commit/0ddaafd35c7120ea8328cd438fdee0a8003ea896

PR #1162, opened 2015-06-27
Throw exception when simple reference is targeting discriminated document

### LINKED ISSUE #1124: Mapping exception for combining simple references and discriminators
Per @jwage's comment in https://github.com/doctrine/mongodb-odm/pull/1038#issuecomment-87233315:

> I think we should throw an exception if you combine simple => true and a discriminator. That type of relationship is inherently not simple and storing the reference as an array is critical to be able to store the discriminator value so we can construct the proper class during hydration.

As explained in the [Simple References documentation](http://docs.doctrine-project.org/projects/doctrine-mongodb-odm/en/latest/reference/reference-mapping.html#simple-references) and discussed in the above ticket, we do not have a viable solution to handle discriminated, simple references. Additionally, @clslrns discussed a caveat in that PHP limits us to creating a proxy object of the top-level class, which may be abstract, and does not allow us to cast into the more specific class after proxy initialization (solving this within the language would require some way to redeem a proxy object for its resolved instance, which conflicts with the idea of lazy initialization).

----

For the uninitiated, "simple" references are fields storing a referenced document's `_id` value (or an array of them in the case of a reference-many relationship). This is the lighter form of ODM's standard references, which are [DBRef](http://docs.mongodb.org/manual/reference/database-references/#dbref-explanation) objects.

malarzm:
Closes #1124


alcaeus:
:+1:

jmikola:
Question: is it not possible to catch this earlier (e.g. during class metadata loading), because we might not have all of the classes' metadata loaded yet? It definitely should be a mapping exception, but I realize we might not have a stage to do that. And if two discriminated based classes both referred to each other, one would logically have to be loaded first (so we couldn't validate both at load time).

Other than that, LGTM.

malarzm:
@jmikola unfortunately answer to your question is "yes". This was even my main concern that we'll throw `MappingException` too late, but there's no way to do that while loading metadata

malarzm:
Merged manually in https://github.com/doctrine/mongodb-odm/commit/ab600a72be81ab618c421212a1ddf5e9a2a06cbd

PR #1164, opened 2015-07-03
Wrong class is returned when using select() and single collection inheritance

malarzm:
For now we're manually adding discriminator field in select() but it would be nice if ODM could do that for us. Originally we were hit by this because we were searching by parent class which is abstract and instance of it could not be created.

Closes #879

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Query/Builder.php
+++ b/lib/Doctrine/ODM/MongoDB/Query/Builder.php
@@ -21,6 +21,7 @@ namespace Doctrine\ODM\MongoDB\Query;
 
 use Doctrine\ODM\MongoDB\DocumentManager;
 use Doctrine\ODM\MongoDB\Hydrator;
+use Doctrine\ODM\MongoDB\Mapping\ClassMetadataInfo;
 use Doctrine\ODM\MongoDB\Query\Expr;
 use Doctrine\ODM\MongoDB\UnitOfWork;
 
@@ -321,8 +322,15 @@ class Builder extends \Doctrine\MongoDB\Query\Builder
             $query['distinct'] = $documentPersister->prepareFieldName($query['distinct']);
         }
 
-        if (isset($query['select'])) {
+        if ( ! empty($query['select'])) {
             $query['select'] = $documentPersister->prepareSortOrProjection($query['select']);
+            if ($this->hydrate && $this->class->inheritanceType === ClassMetadataInfo::INHERITANCE_TYPE_SINGLE_COLLECTION
+                && ! isset($query['select'][$this->class->discriminatorField])) {
+                $includeMode = 0 < count(array_filter($query['select'], function($mode) { return $mode == 1; }));
+                if ($includeMode && ! isset($query['select'][$this->class->discriminatorField])) {
+                    $query['select'][$this->class->discriminatorField] = 1;
+                }
+            }
         }
 
         if (isset($query['sort'])) {
--- a/tests/Doctrine/ODM/MongoDB/Tests/QueryTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/QueryTest.php
@@ -170,6 +170,23 @@ class QueryTest extends BaseTest
         $this->assertTrue(array_key_exists('eO.eO.e1.eO.eP.pO._id', $debug));
         $this->assertEquals($mongoId, $debug['eO.eO.e1.eO.eP.pO._id']);
     }
+
+    public function testSelectVsSingleCollectionInheritance()
+    {
+        $p = new \Documents\SubProject('SubProject');
+        $this->dm->persist($p);
+        $this->dm->flush();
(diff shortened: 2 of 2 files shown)


alcaeus:
Looks good to me - maybe we could add that bit of info to the inheritance documentation?

lib/Doctrine/ODM/MongoDB/Query/Builder.php
```diff
@@ -317,8 +318,12 @@ public function getQuery(array $options = array())
 
         $query['newObj'] = $this->expr->getNewObj();
 
-        if (isset($query['select'])) {
+        if ( ! empty($query['select'])) {
             $query['select'] = $documentPersister->prepareSortOrProjection($query['select']);
+            if ($this->class->inheritanceType === ClassMetadataInfo::INHERITANCE_TYPE_SINGLE_COLLECTION
+                    && ! in_array($this->class->discriminatorField, $query['select'])) {
+                $query['select'][$this->class->discriminatorField] = 1;
```
jmikola: This might be premature if the user is using the projection to exclude fields. In that case, we can't ever mix `0` and `1` (unless `_id` is the only field being excluded). We can fix this by iterating through `$query['select']` and checking if there's a key other than `_id` that has a value of `1` -- if there is, we know we're in "include these fields, exclude everything else" mode. Also, you don't have to worry about array projections, since those don't trigger include/exclude mode:

```
> db.foo.insert({_id:1,x:1,y:[1,2,3]})
WriteResult({ "nInserted" : 1 })
> db.foo.find({_id:1},{y:{$slice:2}})
{ "_id" : 1, "x" : 1, "y" : [ 1, 2 ] }
> db.foo.find({_id:1},{y:1})
{ "_id" : 1, "y" : [ 1, 2, 3 ] }
```

----

If I understand correctly, you're trying to detect a case where the user is in "include" mode, in which case MongoDB is going to exclude all other fields (except `_id`), and force the discriminator to be included. What if the user is trying to do a [covered query](http://docs.mongodb.org/manual/core/query-optimization/#covered-query) and purposefully wants only fields in the index?

I think a good compromise would be to only do this if `$hydrate` is true (because we need the discriminator to construct the proper class). If the user is getting raw data back, they should be fine only getting the discriminator if they really want it.

malarzm: These are some valid points! I'll go with checking if both `$hydrate` is true and `0` and `1` are not mixed, I admit I haven't thought about it.

malarzm:
@jmikola I've introduced the check and added more tests

lib/Doctrine/ODM/MongoDB/Query/Builder.php
```diff
@@ -321,8 +322,14 @@ public function getQuery(array $options = array())
             $query['distinct'] = $documentPersister->prepareFieldName($query['distinct']);
         }
 
-        if (isset($query['select'])) {
+        if ( ! empty($query['select'])) {
             $query['select'] = $documentPersister->prepareSortOrProjection($query['select']);
+            if ($this->hydrate && $this->class->inheritanceType === ClassMetadataInfo::INHERITANCE_TYPE_SINGLE_COLLECTION) {
+                $includeMode = 0 < count(array_filter($query['select'], function($mode) { return $mode === 1; }));
```
jmikola: I was going to suggest adding ` || $mode === true` to the quality, or just relaxing it to ``$mode == 1`, but I realized that our own [Builder](https://github.com/doctrine/mongodb/blob/master/lib/Doctrine/MongoDB/Query/Builder.php) only uses integers for include/exclude modes. Booleans are supported by the server, but it seems we never give the user a chance to set a raw projection.

Up to you, but I'm ok with keeping this as-is :+1:

lib/Doctrine/ODM/MongoDB/Query/Builder.php
```diff
@@ -321,8 +322,14 @@ public function getQuery(array $options = array())
...
+                if ($includeMode && ! in_array($this->class->discriminatorField, $query['select'])) {
```
jmikola: `in_array($this->class->discriminatorField, $query['select'])` doesn't seem correct, as you want to search the keys. Did you mean to use `isset($query['select'][$this->class->discriminatorField]`?

Also, we can move this up to the parent conditional, as it's cheaper than the `array_filter()` and might help us avoid that. On a successful case, we'd end up doing both checks anyway.

lib/Doctrine/ODM/MongoDB/Query/Builder.php
malarzm: Relaxed to equal check

lib/Doctrine/ODM/MongoDB/Query/Builder.php
malarzm: Yup, thanks for catching this :+1:

jmikola:
LGTM.

PR #1173, opened 2015-07-18
PSR-2 Cleanup

peshi:
Small psr-2 cleanup


peshi:
If this is OK with you, i'll squash the commits.

malarzm:
We purposely add spaces before `!` so https://github.com/hyperunknown/mongodb-odm/commit/8fc2f0b7ead8c6d6f193ec6231d9c341fb118892 won't be much use. 

Anyway I'm not sure if this will be merged, there were previous attempts on running a CS Fixer but none of them was successful as the best would be enforcing standard organization wide.

By the way, is https://github.com/hyperunknown/mongodb-odm/commit/5e3bbd687a48e78a879f3ccb4076c9e0252690fb really a PSR thing?

peshi:
@malarzm 
Any particular reason for using a space before ```!``` ?
Why where previous attempts using a cs fixer not successful ?

For your question, please refer to pt. 6. Closures
http://www.php-fig.org/psr/psr-2

Let me know if you only want parts of it like eof_ending, trailing_spaces etc, or none.. then I'll close this PR.

BernardoSilva:
I would suggest merge this PR: https://github.com/doctrine/mongodb-odm/pull/1097 first, this way every contributor follow same PSR-2 rules. :)

malarzm:
@peshi it's more readable :) Also I think @guilhermeblanco likes that very much and we believe it was ommited accidentally in the PSR. Heh for closures, PSR2 always surprises me...

As for previous attempts we ended up not merging them mainly because we'd like to roll CS rules for all Doctrine projects but usually this issue is not on top of anybody's todo list ;)

peshi:
Alright, fair enough :)
In any case each commit represent the rule I used with php-cs-fixer from Sensiolabs.
It's probably better you guys choose your preferred rules to use in the future :)

Closing.

jmikola:
@peshi: thanks for splitting these up. I did look into creating fixer rules, but never got around to customizing them enough to meet @guilhermeblanco's demands.

https://github.com/doctrine/coding-standard exists and will one day address this, but it's still a work in progress.

peshi:
@jmikola No worries, I have been following the fixer commits that's why I decided to split each rule into it's own commit and only choosing a few. In any case I suspected that it might be rejected and in the process receive an explanation :)

PR #1175, opened 2015-07-20
Optimized commit's performace

malarzm:
Each execute action was unnecessarly iterating over all documents checking if there are maybe instance of given class, which wasn't very optimal given we were iterating over the same documents to get unique classes first. Roughly previous flow was O(n^2), new one is O(n) where n is number of documents for given action and also simplifies the logic a bit by taking out checking for embedded documents (where I could).


lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -451,23 +451,43 @@ public function commit($document = null, array $options = array())
...
     /**
+     * Divides given documents by their class name. Returns array of tuples that
+     * is easiest described by:
+     *
+     * list($class, $documents) = array(ClassMetadata, object[])
+     *
      * @param array $documents
-     * @return ClassMetadata[]
+     * @param bool $excludeEmbedded
+     * @return array
```
jmikola: We can use:

> `@return array Tuple of ClassMetadata and a corresponding array of objects`

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -451,23 +451,43 @@ public function commit($document = null, array $options = array())
     }
 
     /**
+     * Divides given documents by their class name. Returns array of tuples that
+     * is easiest described by:
+     *
+     * list($class, $documents) = array(ClassMetadata, object[])
+     *
      * @param array $documents
```
jmikola: Add `Scheduled documents (e.g. $this->documentInsertions)`.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -451,23 +451,43 @@ public function commit($document = null, array $options = array())
     }
 
     /**
+     * Divides given documents by their class name. Returns array of tuples that
+     * is easiest described by:
+     *
+     * list($class, $documents) = array(ClassMetadata, object[])
```
jmikola: We can rewrite this as:

> Groups a list of scheduled documents by their class.

Combined with the `@param` and `@return` changes below, I think that will sum up the method nicely.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -411,24 +411,24 @@ public function commit($document = null, array $options = array())
             $this->evm->dispatchEvent(Events::onFlush, new Event\OnFlushEventArgs($this->dm));
         }
 
-        foreach ($this->getClassesForCommitAction($this->documentUpserts) as $class) {
-            if ( ! $class->isEmbeddedDocument) {
-                $this->executeUpserts($class, $options);
-            }
+        foreach ($this->getClassesForCommitAction($this->documentUpserts) as $tuple) {
+            list($class, $documents) = $tuple;
```
jmikola: I'm not sure if I've done it elsewhere in ODM (certainly in other projects), but I'd suggest `$classAndDocuments` instead of a generic `$tuple` name here.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -451,23 +451,43 @@ public function commit($document = null, array $options = array())
...
+            }
+            $class = $this->dm->getClassMetadata($className);
+            if ($class->isEmbeddedDocument && $excludeEmbedded) {
+                $embeds[$className] = true;
+                continue;
+            }
+            if (empty($divided[$class->name])) {
+                $divided[$class->name] = array($class, array($oid => $d));
+            } else {
+                $divided[$class->name][1][$oid] = $d;
```
jmikola: Isn't this redundant with the `isset($divided[$className])` conditional above?

malarzm: Not when we're dealing with Proxies as $className !== $class->name

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -1110,21 +1125,17 @@ private function executeInserts(ClassMetadata $class, array $options = array())
...
 
-        foreach ($this->documentUpserts as $oid => $document) {
-            if (get_class($document) === $className) {
-                $persister->addUpsert($document);
-                $upsertedDocuments[] = $document;
-                unset($this->documentUpserts[$oid]);
-            }
+        foreach ($documents as $oid => $document) {
+            $persister->addUpsert($document);
+            unset($this->documentUpserts[$oid]);
```
jmikola: I noticed the original `foreach` here didn't include a conditional check to ignore embedded documents. `executeUpdates()` has it, but not `executeUpserts()` and `executeInserts()`. I assume that's because ODM was ensuring embedded documents are never scheduled for insert/upsert elsewhere?

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -1899,72 +1904,6 @@ private function doRemove($document, array &$visited)
     }
 
     /**
-     * Cascades the preRemove event to embedded documents.
-     *
-     * @param ClassMetadata $class
-     * @param object $document
-     */
-    private function cascadePreRemove(ClassMetadata $class, $document)
```
jmikola: Why was this removed? I expect there could be some third-party code depending on these events.

malarzm: Events are still there, they're just fired differently, these private method were unused

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
malarzm: For inserts they were filtered [here](https://github.com/doctrine/mongodb-odm/pull/1175/files#diff-ae2c9f4b58c612dce08a0f60ff683aecL421) and for upsert they couldn't even be scheduled as that would raise an exception [here](https://github.com/doctrine/mongodb-odm/blob/3e857760e674a93a10f022e99f52eef2fbfdba14/lib/Doctrine/ODM/MongoDB/UnitOfWork.php#L1398)

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -451,23 +451,43 @@ public function commit($document = null, array $options = array())
...
-                $class = $this->dm->getClassMetadata($className);
-                $classes[$class->name] = $class;
+            if (isset($embeds[$className])) {
+                continue;
+            }
+            if (isset($divided[$className])) {
+                $divided[$className][1][$oid] = $d;
+                continue;
+            }
+            $class = $this->dm->getClassMetadata($className);
```
jmikola: Based on [`AbstractClassMetadataFactory::getClassMetadata()`](https://github.com/doctrine/common/blob/ff72726b0876638fa7db2f28b0f26e3f1f6656ca/lib/Doctrine/Common/Persistence/Mapping/AbstractClassMetadataFactory.php#L187) and (`setMetadataFor()`](https://github.com/doctrine/common/blob/ff72726b0876638fa7db2f28b0f26e3f1f6656ca/lib/Doctrine/Common/Persistence/Mapping/AbstractClassMetadataFactory.php#L255), I assume this takes of resolving proxies to a usable class name?

I ask because we removed `$document instanceof Proxy && $document instanceof $className` in some of the execute methods below.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
jmikola: :+1:

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
malarzm: Yes, cross referencing with https://github.com/doctrine/mongodb-odm/pull/1175#discussion_r35679689 :)

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -451,23 +451,43 @@ public function commit($document = null, array $options = array())
...
+     *
+     * list($class, $documents) = array(ClassMetadata, object[])
+     *
      * @param array $documents
-     * @return ClassMetadata[]
+     * @param bool $excludeEmbedded
+     * @return array
      */
-    private function getClassesForCommitAction($documents)
+    private function getClassesForCommitAction($documents, $excludeEmbedded = true)
```
jmikola: Most other modular methods throughout Doctrine use an `$includeX` argument. Can we change this to `$includeEmbedded = false` for consistency?

I'm thinking of `$includeNestedCollections` for `prepareEmbeddedDocumentValue()` at the moment.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -1159,42 +1171,36 @@ private function executeUpdates(ClassMetadata $class, array $options = array())
...
 
-                if ( ! $class->isEmbeddedDocument && ( ! empty($this->documentChangeSets[$oid]) || $this->hasScheduledCollections($document))) {
-                    $persister->update($document, $options);
-                }
+            if ( ! empty($this->documentChangeSets[$oid]) || $this->hasScheduledCollections($document)) {
+                $persister->update($document, $options);
+            }
 
-                unset($this->documentUpdates[$oid]);
+            unset($this->documentUpdates[$oid]);
```
jmikola: Previously, `unset($this->documentUpdates[$oid]);` would still have executed for an embedded document. Is this no longer a concern?

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
malarzm: Sure thing is, will be here in second...

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
jmikola: Ah, it appears you removed them because UnitOfWork never called them. Correct?

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
jmikola: :+1:

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
malarzm: We're not relying on this anywhere and `$this->documentUpdates` is cleared at the end of commit so I think we should be fine

jmikola: :+1:

jmikola:
This gets my:

malarzm:
Merged manually in https://github.com/doctrine/mongodb-odm/commit/dbf21abf46f027d49328e13baaa6651e2e408710

PR #1177, opened 2015-07-23
Straighten removing strategy from collection field

### LINKED ISSUE #1174: Undefined index 'strategy' since the introduction of the CollectionHelper
Hi there, thanks for Doctrine!

We've been following dev-master and started getting this error in our tests:

```
exception 'Symfony\Component\Debug\Exception\ContextErrorException' with message 'Notice: Undefined index: strategy' in /var/www/site/web/vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php:1047 Stack trace: 
#0 /var/www/site/web/vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php(1047): Symfony\Component\Debug\ErrorHandler->handleError(8, 'Undefined index...', '/var/www/...', 1047, Array) 
#1 /var/www/site/web/vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php(936): Doctrine\ODM\MongoDB\Persisters\DocumentPersister->prepareQueryElement('nodes.id', '763cf7d5-0f88-4...', NULL, true) 
#2 /var/www/site/web/vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php(926): Doctrine\ODM\MongoDB\Persisters\DocumentPersister->prepareQueryOrNewObj(Array) 
#3 /var/www/site/web/vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php(470): Doctrine\ODM\MongoDB\Persisters\DocumentPersister->prepareQueryOrNewObj(Array) 
#4 /var/www/site/web/vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/DocumentRepository.php(184): Doctrine\ODM\MongoDB\Persisters\DocumentPersister->load(Array) 
```

I did a `git bisect` and tracked it down to https://github.com/doctrine/mongodb-odm/commit/dfd7d64fe71f6963e90cc49960fca32116910dbb

The fields `nodes` is `type: collection` in our mapping file. `nodes.id` is an `id` key in the array of nodes.

For the moment I've pinned mongo-odm in our composer.json:
`"doctrine/mongodb-odm": "dev-master#1522f509a511f26238dafce2798ccd91a6c627f7"`

malarzm:
Fixes #1174. This commit finishes removing "unused" strategy from `collection` field. People who are still using `strategy` combined with `collection` and annotation mapping will now start getting error because `AbstractField` no longer has that property (which was the reason why our tests were green but YML/XML mappings suffered as reported in #1174)

@redthor could you maybe run your test suite against this PR?

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Mapping/Annotations/AbstractField.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/Annotations/AbstractField.php
@@ -27,5 +27,4 @@ abstract class AbstractField extends Annotation
     public $type = 'string';
     public $nullable = false;
     public $options = array();
-    public $strategy;
 }
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -1044,7 +1044,8 @@ class DocumentPersister
             return array($fieldName, $value);
         }
 
-        if ($mapping['strategy'] === 'set' && isset($e[2])) {
+        if (isset($mapping['strategy']) && CollectionHelper::isHash($mapping['strategy'])
+                && isset($e[2])) {
             $objectProperty = $e[2];
             $objectPropertyPrefix = $e[1] . '.';
             $nextObjectProperty = implode('.', array_slice($e, 3));
--- a/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
+++ b/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
@@ -907,7 +907,7 @@ class UnitOfWork implements PropertyChangedListener
             $state = $this->getDocumentState($entry, self::STATE_NEW);
 
             // Handle "set" strategy for multi-level hierarchy
-            $pathKey = CollectionHelper::isList($assoc['strategy']) ? $count : $key;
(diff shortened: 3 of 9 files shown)


redthor:
The test still fails but neither driver contains `strategy`:

```
1) Doctrine\ODM\MongoDB\Tests\Functional\Ticket\GH1174Test::compareDrivers
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
     'isInverseSide' => false
+    'options' => Array ()
+    'value' => null
 )
```

I then checked out your branch in our dev code and it appears to be all working.

So my test case is perhaps not something we need?

Thanks for looking into this.

malarzm:
> So my test case is perhaps not something we need?

I added this assert: https://github.com/doctrine/mongodb-odm/pull/1177/files#diff-6a4e49bb800973ef639a55d44888ff43R50 I'd say the rest is out of scope of this PR. But anyway thank you very much for that test, I would spend a lot more time looking for the culprit if not for your test :)

jmikola:
LGTM. I don't have any recollection of strategy being used for the `collection` field type. Both it and `hash` are treated like basic fields where any changes are persisted with `$set`. The collection strategies only apply when dealing with EmbedMany and ReferenceMany relationships.

PR #1180, opened 2015-07-25
Add support for custom object ids (#1178)

marcospassos:
This PR adds support for custom object IDs #1178.

Issue #1178, opened 2015-07-25
Unable to use custom object id


marcospassos:
Hi folks!

I'm trying to use a custom object id without success. I've defined a custom mapping type, but seems like id fields just ignore the attribute `type`, thus the conversion is not performed.

Mapping:
```xml
<document name="Acme\User"  customId="true">
    <field fieldName="id" id="true" type="user_id" strategy="NONE" />
</document>
```

Custom type mapping:
```php
class UserIdType extends Type
{
    public function convertToDatabaseValue($value)
    {
        return $value !== null ? (string) $value : null;
    }

    public function closureToMongo()
    {
        return '$return = (string) $value;';
    }

    public function convertToPHPValue($value)
    {
        return $value !== null ? UserId::fromString($value) : null;
    }

    public function closureToPHP()
    {
        return '$return = Acme\UserId::fromString((string) $value);';
    }
}
```

Model:
```php
namespace Acme\User;

class User
{
    private $id;

    public function __construct(UserId $id)
    {
         $this->id = $id;
    }
}
```

Error:

```
zero-length keys are not allowed, did you use $ with double quotes?
```

For DDD adopters, is very common to have an id represented by an object.

If you guys agree, I would be glad to provide a PR to fix it.

malarzm:
@marcospassos indeed looks like a good thing to have, PR will be much appreciated!

malarzm:
I've scheduled this to be in 1.x but if feature will be done by the time we release 1.0 I'll happily include it :)

marcospassos:
@malarzm Nice! I'll work on this over the next days. Could you please take a look at #1179 as well?

marcospassos:
PR added including unit tests.

malarzm:
I'm in countryside so I'll take a look at all the things tomorrow ;)

Issue #1179, opened 2015-07-25
Support to non-primitive types in queries


marcospassos:
In addition to #1178, would be great to support the use of non-primitive values in queries, exactly as ORM does. Auto-converting parameters to database values makes things look more consistent and avoid a lot of boilerplate code for casting non-primitive types into primitive ones.

Currently, the execution of the following snippet results in an exception:

```php
public function findByOwner(UserId $id)
{
        return $this->createQueryBuilder()
            ->find()
            ->field('owner')
            ->equals($id)
            ->execute();
}
```

Error:
```
zero-length keys are not allowed, did you use $ with double quotes?
```

It looks weird because the field is mapped to a custom type, so why the parameter is not being converted to the specified value? 

The workaround is casting the value whenever you want to use it, what exposes some implementation details that, IMHO, should be handled by the ODM:

```php
public function findByOwner(UserId $id)
{
        return $this->createQueryBuilder()
            ->find()
            ->field('owner')
            ->equals((string) $id)
            ->execute();
}
```

Is there any possibilities to schedule it to be in 1.x release? If so, I'll try to provide a PR as well.

malarzm:
For this particular case using `$qb->references($user)` would be more adequate (however that requires having User object)

marcospassos:
I just updated the examples to you get my point.

malarzm:
Yes, I agree that now it makes more sense, however I'm not sure if this can be easily supported - it's easy in simple cases like this, but you might not know mapping of deeply embedded documents as it may vary on discriminator values or even have different types of id if documents share the same collection (single collection inheritance).

marcospassos:
Another use case:

```php
public function findBySupportId(SupportId $supportId)
{
     return $this->findBy(['supportId' => $supportId]);
}
```

dkarlovi:
If it's any consolation, ORM has exactly the same problem, https://github.com/doctrine/doctrine2/issues/4632

Both of which came up because using UUIDs as IDs.

juliusstoerrle:
What is the reason fot not converting normal fields in `DocumentPersister::prepareQueryElement` called in `DocumentPersister::prepareQueryOrNewObj()` if there is an entry in the `ClassMetadata`?

If my criteria contains the name of a property/field of a document I would expect that the ORM will automatically convert the type.

As this is a BC Break, is there any way we could change this for the 2.0 release?

alcaeus:
[...]

PR #1181, opened 2015-07-27
Add support for custom object ids (#1178)

marcospassos:
Rebased version of #1180.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
+++ b/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
@@ -2813,7 +2813,7 @@ class UnitOfWork implements PropertyChangedListener
             unset($data[$class->discriminatorField]);
         }
 
-        $id = $class->getDatabaseIdentifierValue($data['_id']);
+        $id = $data['_id'];
         $serializedId = serialize($id);
 
         if (isset($this->identityMap[$class->name][$serializedId])) {
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1178/UserId.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1178/UserId.php
@@ -1,0 +1,13 @@
+<?php
+
+namespace Doctrine\ODM\MongoDB\Tests\Functional\Ticket\GH1178;
+
+class UserId
+{
+    public $value;
+
+    public function __construct($value)
+    {
+        $this->value = $value;
+    }
+}
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1178Test.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1178Test.php
@@ -1,0 +1,126 @@
+<?php
+
+namespace Doctrine\ODM\MongoDB\Tests\Functional\Ticket;
+
+use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
+use Doctrine\ODM\MongoDB\Tests\Functional\Ticket\GH1178\UserId;
+use Doctrine\ODM\MongoDB\Types\Type;
+
+/**
+ * @author Marcos Passos <marcos@croct.com>
+ *
+ * @group GH-1178
+ *
+ * @see   https://github.com/doctrine/mongodb-odm/issues/GH1178
+ */
+class GH1178Test extends \Doctrine\ODM\MongoDB\Tests\BaseTest
+{
+    public static function setUpBeforeClass()
+    {
+        Type::addType('user_id', __NAMESPACE__ . '\UserIdType');
+    }
+
+    public function testCustomIdTypeValueConversions()
+    {
+        $user = new UserGH1178($id = new UserId(1));
+        $this->dm->persist($user);
+        $this->dm->flush();
+
+        $this->dm->clear();
+
+        $this->assertEquals($user, $this->dm->find(__NAMESPACE__ . '\UserGH1178', $id));
+    }
+
+    public function testRepositoryFindByCustomIdObject()
+    {
+        $user = new UserGH1178($id = new UserId(1));
+        $this->dm->persist($user);
+        $this->dm->flush();
+
+        $this->dm->clear();
+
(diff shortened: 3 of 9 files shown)


marcospassos:
@malarzm could you please take a look at this?

marcospassos:
Seems like it covers #1178 and #1179 ;)

marcospassos:
@malarzm could you please take a look at this? Seems like it covers #1178 and #1179 ;)

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1178/UserId.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace Doctrine\ODM\MongoDB\Tests\Functional\Ticket\GH1178;
+
+class UserId
```
malarzm: Why this class is in separate folder instead of with the test?

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1178Test.php
```diff
@@ -0,0 +1,126 @@
...
+namespace Doctrine\ODM\MongoDB\Tests\Functional\Ticket;
+
+use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
+use Doctrine\ODM\MongoDB\Tests\Functional\Ticket\GH1178\UserId;
+use Doctrine\ODM\MongoDB\Types\Type;
+
+/**
+ * @author Marcos Passos <marcos@croct.com>
+ *
+ * @group GH-1178
```
malarzm: `@group` is not needed

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1178Test.php
```diff
@@ -0,0 +1,126 @@
...
+
+    public function testRepositoryFindByCustomIdObject()
+    {
+        $user = new UserGH1178($id = new UserId(1));
+        $this->dm->persist($user);
+        $this->dm->flush();
+
+        $this->dm->clear();
+
+        $repository = $this->dm->getRepository(UserGH1178::class);
```
malarzm: Unfortunately `class` constant can't be used as we still support PHP 5.3 and 5.4

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1178/UserId.php
marcospassos: Because it needs to be autoloadable:

```php
public function closureToPHP()
{
    return '$return = new \Doctrine\ODM\MongoDB\Tests\Functional\Ticket\GH1178\UserId($value);';
}
```

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -2813,7 +2813,7 @@ public function getOrCreateDocument($className, $data, &$hints = array(), $docum
             unset($data[$class->discriminatorField]);
         }
 
-        $id = $class->getDatabaseIdentifierValue($data['_id']);
```
malarzm: Why are we giving up on this?

marcospassos: Because all values feched from database are already normalized. In that case, it is trying to convert `int(1)` into `int(1)`, [which will result in an error](https://github.com/doctrine/mongodb-odm/pull/1181/files#r35585992).

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1178/UserId.php
malarzm: Hmmm does it fail when it's placed in same file as test? If yes then sorry for bothering :)

marcospassos: Yes, it's not capable of loading the class anymore, because the file path is guessed from the fully-qualified class name.

malarzm: Ok, I thought that we might get away with this somehow given it should be needed when running along with test itself :+1:

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
marcospassos: BTW, all tests keep passing.

malarzm: :+1:

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1178Test.php
```diff
@@ -0,0 +1,144 @@
...
+            ->references($user)
+            ->getQuery();
+
+        $this->assertEquals($comment, $query->getSingleResult());
+    }
+
+    /**
+     * @expectedException \Doctrine\ODM\MongoDB\Tests\Functional\Ticket\ConversionException
+     */
+    public function testIdTypeMismatchShouldThrowAnError()
```
marcospassos: Notice that type mismatches will result in an error, because the UOW will try to convert it back to the database format.

marcospassos:
Issues fixed and one more test added.

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1178Test.php
marcospassos: It further highlights the importance of #1179 to keep the API consistent.

marcospassos:
@malarzm any thoughts?

malarzm:
I'm not sure the added tests brings any value, it barely tests if there's an exception thrown by the code from the test (I get its point though)?

marcospassos:
This tests make clear that type mismatches will result in a exception. It makes explicit that it's the expected result, not a bug.  The exception is used only make it  more explicit, but I could just match the error.

PR #1182, opened 2015-07-28
Add suport for querying using object ids (#1179)

marcospassos:
This PR aims to fix the issue #1179.

Follows the same strategy adopted by Doctrine ORM:

> The UnitOfWork internally assumes that entity identifiers are castable to string. Hence, when using custom types that map to PHP objects as IDs, such objects must implement the `__toString()` magic method.

Includes unit tests.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -978,6 +978,10 @@ class DocumentPersister
 
             // No further preparation unless we're dealing with a simple reference
             if (empty($mapping['reference']) || empty($mapping['simple'])) {
+                if (empty($mapping['embedded']) && $mapping['type'] !== Type::HASH && $mapping['type'] !== Type::RAW) {
+                    return array($fieldName, is_object($value) ? (string) $value : $value);
+                }
+
                 return array($fieldName, $value);
             }
 
--- a/tests/Doctrine/ODM/MongoDB/Tests/QueryTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/QueryTest.php
@@ -4,6 +4,7 @@ namespace Doctrine\ODM\MongoDB\Tests;
 
 use Doctrine\ODM\MongoDB\QueryBuilder;
 use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
+use Doctrine\ODM\MongoDB\Types\Type;
 
 class QueryTest extends BaseTest
 {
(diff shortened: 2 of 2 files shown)


marcospassos:
@malarzm would be awesome to have both #1178  and #1179 in 1.x! :)

Could you please take a look at this?

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -978,6 +978,10 @@ private function prepareQueryElement($fieldName, $value = null, $class = null, $
 
             // No further preparation unless we're dealing with a simple reference
             if (empty($mapping['reference']) || empty($mapping['simple'])) {
+                if (empty($mapping['embedded']) && $mapping['type'] !== Type::HASH && $mapping['type'] !== Type::RAW) {
```
malarzm: Shouldn't we exclude `Type::COLLECTION` here as well? Or add a check for `is_array` later as these shouldn't be casted to string?

marcospassos: Sure, you're right. I suggest to exclude `Type::COLLECTION` right here.

malarzm: Actually I think we can just remove `$mapping['type'] !== Type::HASH` here

malarzm: It can only be an `array` so it won't pass `is_object` check anyways

marcospassos: Removing `$mapping['type'] !== Type::HASH` results in lots of errors of trying to cast `stdClass` to string.

marcospassos: I think we should use `$mapping['type'] !== Type::HASH`, otherwise `Collection` will be casted to string by mistake.

malarzm: Ah crap, indeed Hash will become `stdClass` at some point, `Collection` is always an `array` so it won't be casted to string so we don't have to make a check, sorry for confusion, I should have give it more thorough thought before commenting in the morning on the run :+1:

malarzm:
Can you rebase?

marcospassos:
I'll submit the fixes and rebase the code by the end of the day. Would be nice to merge both #1178 and #1179 to avoid getting outdated by the constant improvements. Is it possible?

malarzm:
Maybe you can just make them 1 PR instead? Just use separate commits and I think we'll be fine :)

Issue #854, opened 2014-04-11
Ensure Cursor::toArray() specifies false for $useKeys


jmikola:
Calling `toArray()` when documents in a result set have non-scalar keys invites PHP errors. There are several places where we call `toArray()` on possible cursors:

* [x] `DocumentPersister::loadReferenceManyCollectionInverseSide()`
* [x] `DocumentPersister::loadReferenceManyCollectionOwningSide()`
* [x] `DocumentPersister::loadReferenceManyWithRepositoryMethod()`
* [x] `DocumentRepository::matching()`
* [x] `EagerCursor::initialize()`
* [x] `ReferencePrimer::__construct()`

malarzm:
Outstanding calls were fixed in https://github.com/doctrine/mongodb-odm/commit/b6aa8620e4f29d2952fffe4209b49acff39f1547

jmikola:
![](http://www.reactiongifs.com/r/vhpy.gif)

PR #1184, opened 2015-07-29
Add support for custom mapping type for object IDs

marcospassos:
Implements #1178 and #1179 (merges  #1181 and #1182 into a single PR).

Adds a note in documentation for examplaing the implemention details required for using custom types as object IDs.


marcospassos:
@malarzm I think it's ready to get merged.

malarzm:
One thing that crossed my mind and I'm not sure if it's important or not - complex IDs that shouldn't be casted to string... @jmikola @alcaeus thoughts? I haven't used them much but I do remember https://github.com/doctrine/mongodb-odm/issues/854

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -978,6 +978,10 @@ private function prepareQueryElement($fieldName, $value = null, $class = null, $
 
             // No further preparation unless we're dealing with a simple reference
             if (empty($mapping['reference']) || empty($mapping['simple'])) {
+                if (empty($mapping['embedded']) && $mapping['type'] !== Type::HASH && $mapping['type'] !== Type::RAW) {
```
marcospassos: @malarzm what do you think about this?
```php
if (empty($mapping['embedded']) && is_object($value) && 0 !== strpos(get_class(Type::getType($mapping['type'])), 'Doctrine\ODM\MongoDB\Types')) {
    return array($fieldName, (string) $value);
}

In that way, only objects mapped to custom types will be casted to string.

Comment (malarzm) [inline lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php]: That could be a solution unless somebody has a custom type that shouldn't be casted to string (some fancy Hash for instance).

Comment (marcospassos) [inline lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php]: So we can end up with the following condition (and addresses your point about complex IDs):

```php
if (empty($mapping['embedded']) && is_object($value) && 0 !== strpos(get_class(Type::getType($mapping['type'])), 'Doctrine\ODM\MongoDB\Types') && method_exists($value, '__toString')) {
    return array($fieldName, (string) $value);
} 
```

What do you think?

malarzm: I think we should wait for input from other maintainers :)

alcaeus:
I'll take a look at this PR tonight ;)

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataInfo.php
```diff
@@ -1137,7 +1137,7 @@ public function mapField(array $mapping)
                     } elseif ($generatorType === ClassMetadata::GENERATOR_TYPE_INCREMENT) {
                         $mapping['type'] = 'int_id';
                     } else {
-                        $mapping['type'] = 'custom_id';
+                        $mapping['type'] = $mapping['type']  == 'id' ? 'custom_id' : $mapping['type'];
```
alcaeus: Could you explain this change?

Guessing from the information in the tickets associated with this PR I'd guess that using ```strategy=NONE``` and type=```user_id``` results in ```type=custom_id``` since no generator type was set. To prevent this I think we should only be using custom_id when ```$generatorType !== ClassMetadata::GENERATOR_TYPE_NONE``` instead of checking for a specific field type.
Of course, one could always specify a generator type using ```id-generator-option``` to get the correct type mapping with ```strategy=NONE```:

```
<document name="Acme\User"  customId="true">
    <field fieldName="id" id="true" type="user_id" strategy="NONE">
        <id-generator-option name="type" value="user_id" />
    </field>
</document>
```

alcaeus:
@marcospassos as for #1178, my comment explains why you ran into issues - I think we can fix that with minor changes to ClassMetadata not overriding a given ID type when using the NONE strategy (or you can use the workaround I mentioned above by specifying a generator type). This should not require any other changes since it's just a mapping issue.

As for type casting in queries, casting to string is not exactly the correct way. Remember that not every type needs to resolve to a string in ODM.
When inserting documents, the type mapping is used to get the correct database representation using the ```convertToDatabaseValue()``` method of the corresponding type class. The same thing should apply when running queries: we have the mapping and thus the type, so why simply cast to string and hope for the best? In the end, simply casting to string would exclude custom types that are stored as arrays or hashes in the database.

One more thing: a user can always run ```$queryBuilder->field('stringField')->equals(true)``` - in that case I'd expect no result because we're comparing strings with a boolean value. If the bool value is cast to string based on the type mapping of ```stringField``` this no longer holds true. We should keep that in mind when finding a solution.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataInfo.php
marcospassos: @alcaeus does the following proposed solution work currently?

```xml
<document name="Acme\User"  customId="true">
    <field fieldName="id" id="true" type="user_id" strategy="NONE">
        <id-generator-option name="type" value="user_id" />
    </field>
</document>
````

Comment (alcaeus) [inline lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataInfo.php]: There is no ```customId``` attribute in the document type - my bad from copy/pasting your code. It should work without customId:
```
<document name="Acme\User">
    <field fieldName="id" id="true" strategy="NONE">
        <id-generator-option name="type" value="user_id" />
    </field>
</document>
```
You don't need to specify a type on the field mapping since the generator option will override it.

Basically, when using a generator other than AUTO the generator is in charge of specifying the type of the field. It's not entirely intuitive and also not documented, so I think we'll improve the docs on that one.

malarzm: I just happened to be implementing the very same thing for #1190. I think this is good change, it's easy to forget about `generatorOptions` at all (I did :P) and it doesn't break anything especially that `generatorOptions`' `type` still takes priority

PR #1186, opened 2015-07-30
Add PHP 7 on Travis

Soullivaneuh:
On allowed failures section as unstable version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ php:
   - 5.4
   - 5.5
   - 5.6
+  - 7.0
 
 env:
   - MONGO_VERSION=stable
@@ -17,7 +18,9 @@ matrix:
       env: MONGO_VERSION=1.3.7
     - php: 5.3
       env: MONGO_VERSION=1.4.5
-
+  allow_failures:
+    - php: 7.0
+  
 services: mongodb
 
 before_script:


Soullivaneuh:
ping

malarzm:
Right now there is no point in adding this to travis matrix given we're not supporting new Mongo driver yet.

Soullivaneuh:
> we're not supporting new Mongo driver yet.

Did you plan something about it? Any WIP PR?

malarzm:
Yes, we have plans for it, no there's no WIP PR yet unfortunately :(

Soullivaneuh:
And what it your plans? Can I help? :-)

jmikola:
An initial task (possible today) would be getting [doctrine/mongodb](https://github.com/doctrine/mongodb) and [ODM](https://github.com/doctrine/mongodb-odm) working on [mongofill](https://github.com/mongofill/mongofill). That would grant us PHP 7 and HHVM support in one shot, as was done in the [Bernhard project](https://github.com/bernardphp/bernard/blob/master/tests/travis.sh). It shouldn't require more than some Travis CI modifications, as mongofill is designed to be a drop-in replacement for the legacy driver.

The more ambitious and long-term initiative is to port things over to the new [PHP driver](https://github.com/mongodb/mongo-php-driver), which will be getting PHP 7 support in a forthcoming release (see mongodb/mongo-php-driver/pull/136). There is also an HHVM version of this new driver, which shares the very same API.

Soullivaneuh:
Ok I see, thanks for the explanation.

I think the first solution should be implemented quickly on a new minor to allow legacy projects to run on PHP7.

In this case, you will have indeed more time to decide the future of your odm. :+1:

alcaeus:
Going to close this as PHP 7 support is not going to be possible for quite some time due to the new driver API.

PR #1190, opened 2015-08-01
Introduce sanity checks for IDs and fix overriding its type

### LINKED ISSUE #819: Memory leak in detach
Memory leak test:
```
<?php

require_once '../vendor/autoload.php';
require_once 'Doc.php';

use Doctrine\MongoDB\Connection;
use Doctrine\ODM\MongoDB\DocumentManager;
use Doctrine\ODM\MongoDB\Mapping\Driver\AnnotationDriver;

AnnotationDriver::registerAnnotationClasses();

class Test
{
    const COUNT = 200;
    const ITERATIONS = 20;

    /**
     * @var DocumentManager
     */
    protected $dm;
    protected $inc = 1;

    public function test1()
    {
        $docs = [];
        for ($i = 0; $i < self::COUNT; $i++) {
            $doc = $this->createDoc();
            $this->dm->persist($doc);
            $docs[] = $doc;
        }
        $this->dm->flush();
        foreach ($docs as $doc) {
            $this->dm->detach($doc);
        }
    }

    public function test2()
    {
        for ($i = 0; $i < self::COUNT; $i++) {
            $doc = $this->createDoc();
            $this->dm->persist($doc);
        }
        $this->dm->flush();
        $this->dm->clear('Doc');
    }

    public function test3()
    {
        for ($i = 0; $i < self::COUNT; $i++) {
            $doc = $this->createDoc();
            $this->dm->persist($doc);
        }
        $this->dm->flush();
        $this->dm->clear();
    }

    protected function clear()
    {
        $config = new Configuration();
        $config->setProxyDir('/tmp/test');
        $config->setProxyNamespace('Proxies');
        $config->setHydratorDir('/tmp/test');
        $config->setMetadataDriverImpl(AnnotationDriver::create(__DIR__));

        if ($this->dm) {
            $this->dm->close();
        }
        $this->dm = DocumentManager::create(new Connection(), $config);
        $this->dm->getDocumentCollection('Doc')->remove([]);
        $this->inc = 1;
    }

    public function run()
    {
        echo "Test1\n";
        $this->clear();
        for ($i = 0; $i < self::ITERATIONS; $i++) {
            $this->test1();
            $this->mem();
        }

        echo "\n\nTest2\n";
        $this->clear();
        for ($i = 0; $i < self::ITERATIONS; $i++) {
            $this->test2();
            $this->mem();
        }

        echo "\n\nTest3\n";
        $this->clear();
        for ($i = 0; $i < self::ITERATIONS; $i++) {
            $this->test3();
            $this->mem();
        }
    }

    private function createDoc()
    {
        $doc = new Doc();
        $doc->id = $this->inc++;
        $doc->name = str_repeat(md5($this->inc), 40);
        return $doc;
    }

    private function mem()
    {
        echo sprintf("%.2fM\n", memory_get_usage() / 1024 / 1024);
    }
}

$test = new Test();
$test->run();
```

test1 and test2 has memory leaks. 
```
Test1
3.23M
3.55M
3.87M
4.19M
4.57M
4.90M
5.21M
5.53M
5.85M
6.17M
6.62M
6.94M
7.26M
7.58M
7.89M
8.21M
8.53M
8.85M
9.16M
9.48M


Test2
3.45M
3.77M
4.09M
4.40M
4.72M
5.05M
5.37M
5.68M
6.00M
6.32M
6.65M
6.97M
7.29M
7.61M
7.93M
8.25M
8.56M
8.88M
9.20M
9.52M


Test3
3.14M
3.14M
3.14M
3.14M
3.14M
3.14M
3.14M
3.14M
3.14M
3.14M
3.14M
3.14M
3.14M
3.14M
3.14M
3.14M
3.14M
3.14M
3.14M
3.14M
```

Upd: setting ODM\Id(strategy="NONE") fixes the leak

### LINKED ISSUE #857: Specifying anything other than AUTO generator strategy overrides type mapping
I'm not sure if this is expected behaviour, but I did find it very confusing. When specifying a generator strategy and a type in the mapping, the type mapping information is lost if the strategy is anything other than AUTO, eg. strategy NONE.

The reason is clear when looking at the code in Doctrine\ODM\MongoDB\Mapping\ClassMetadataInfo::mapField()

	if ($generatorType !== self::GENERATOR_TYPE_AUTO) {
	    if (isset($this->generatorOptions['type'])) {
	        $mapping['type'] = $this->generatorOptions['type'];
	    } elseif ($generatorType === ClassMetadata::GENERATOR_TYPE_INCREMENT) {
	        $mapping['type'] = 'int_id';
	    } else {
	        $mapping['type'] = 'custom_id';
	    }

	    unset($this->generatorOptions['type']);
	}

I found this very confusing, and maybe it should check first if a type was explicitly configured in the mapping, although I'm not sure how. Other possible solution is also excluding GENERATOR_TYPE_NONE from the condition?

### LINKED ISSUE #1178: Unable to use custom object id

### LINKED ISSUE #1183: Introduce sanity check for id's NONE generation strategy
Citing me from https://github.com/doctrine/mongodb-odm/pull/1129#discussion_r34840667

> As far as I understand using NONE strategy means "I want to generate id on my own, please don't do this for me". I might even say we should throw an exception if we persist a document using strategy NONE without an identifier? It may lead to sick behaviour when creating a DBRef especially that ODM doesn't care what id was generated by driver.

Are there any objections? Generating IDs in lifecycle events comes to my mind but I think `prePersist` should be the only one considered valid (and it happens before id generation so it's perfectly fine)

malarzm:
Fixes #857 (and by the way fixes #1178), closes #1183 and closes #819.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
+++ b/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
@@ -1029,7 +1029,8 @@ class UnitOfWork implements PropertyChangedListener
                 ));
             }
 
-            if ($class->generatorType === ClassMetadata::GENERATOR_TYPE_AUTO && $idValue !== null && ! \MongoId::isValid($idValue)) {
+            // \MongoId::isValid($idValue) was introduced in 1.5.0 so it's no good
+            if ($class->generatorType === ClassMetadata::GENERATOR_TYPE_AUTO && $idValue !== null && ! preg_match('/^[0-9a-f]{24}$/', $idValue)) {
                 throw new \InvalidArgumentException(sprintf(
                     "%s uses AUTO identifier generation strategy but provided identifier is not valid MongoId.",
                     get_class($document)
--- a/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
+++ b/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
@@ -1003,10 +1003,11 @@ class UnitOfWork implements PropertyChangedListener
     }
 
     /**
-     * @param $class
+     * @param ClassMetadata $class
      * @param object $document
+     * @throws \InvalidArgumentException If there is something wrong with document's identifier.
      */
-    private function persistNew($class, $document)
+    private function persistNew(ClassMetadata $class, $document)
     {
         $oid = spl_object_hash($document);
         if ( ! empty($class->lifecycleCallbacks[Events::prePersist])) {
@@ -1019,7 +1020,21 @@ class UnitOfWork implements PropertyChangedListener
         $upsert = false;
         if ($class->identifier) {
             $idValue = $class->getIdentifierValue($document);
-            $upsert = !$class->isEmbeddedDocument && $idValue !== null;
+            $upsert = ! $class->isEmbeddedDocument && $idValue !== null;
+
+            if ($class->generatorType === ClassMetadata::GENERATOR_TYPE_NONE && $idValue === null && ! $class->isEmbeddedDocument) {
+                throw new \InvalidArgumentException(sprintf(
+                    "%s uses NONE identifier generation strategy but no identifier was provided when persisting.",
+                    get_class($document)
+                ));
+            }
+
+            if ($class->generatorType === ClassMetadata::GENERATOR_TYPE_AUTO && $idValue !== null && ! \MongoId::isValid($idValue)) {
+                throw new \InvalidArgumentException(sprintf(
+                    "%s uses AUTO identifier generation strategy but provided identifier is not valid MongoId.",
+                    get_class($document)
+                ));
+            }
 
             if ($class->generatorType !== ClassMetadata::GENERATOR_TYPE_NONE && $idValue === null) {
                 $idValue = $class->idGenerator->generate($this->dm, $document);
@@ -1775,6 +1790,8 @@ class UnitOfWork implements PropertyChangedListener
      * Persists a document as part of the current unit of work.
      *
      * @param object $document The document to persist.
+     * @throws MongoDBException If trying to persist MappedSuperclass.
+     * @throws \InvalidArgumentException If there is something wrong with document's identifier.
      */
     public function persist($document)
     {
--- a/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
+++ b/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
@@ -1022,7 +1022,7 @@ class UnitOfWork implements PropertyChangedListener
             $idValue = $class->getIdentifierValue($document);
             $upsert = ! $class->isEmbeddedDocument && $idValue !== null;
 
-            if ($class->generatorType === ClassMetadata::GENERATOR_TYPE_NONE && $idValue === null && ! $class->isEmbeddedDocument) {
+            if ($class->generatorType === ClassMetadata::GENERATOR_TYPE_NONE && $idValue === null) {
                 throw new \InvalidArgumentException(sprintf(
                     "%s uses NONE identifier generation strategy but no identifier was provided when persisting.",
                     get_class($document)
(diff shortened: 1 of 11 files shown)


lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -1019,7 +1020,22 @@ private function persistNew($class, $document)
...
+
+            if ($class->generatorType === ClassMetadata::GENERATOR_TYPE_NONE && $idValue === null) {
+                throw new \InvalidArgumentException(sprintf(
+                    "%s uses NONE identifier generation strategy but no identifier was provided when persisting.",
+                    get_class($document)
+                ));
+            }
+
+            // \MongoId::isValid($idValue) was introduced in 1.5.0 so it's no good
+            if ($class->generatorType === ClassMetadata::GENERATOR_TYPE_AUTO && $idValue !== null && ! preg_match('/^[0-9a-f]{24}$/', $idValue)) {
```
alcaeus: Technically, this could happen with any generator strategy, including custom. It probably would be best if the generator itself could handle this check, including on whether to ignore it, throw an exception or quietly change the value.

alcaeus:
To be honest, I don't know why we have the generatorOptions type at all. In the end, it does exactly the same as the field mapping but it just introduces complexity and mistakes along the way. I'm not sure about BC here, but IMO the best option would be to drop generatorOptions.type, always use type from the field mapping and setting it to ```id``` or ```custom_id``` depending on the generator strategy only if no type was given. As it is now, writing a generator that generates a ```MongoId``` still requires specifying a generatorOptions.type mapping because the field mapping of ```id``` will be overwritten (which, incidentally, needed a trip through the debugger to figure out why it wouldn't work when I first tried it without generatorOptions.type).

Either way we do this, we need to write some documentation about this feature.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
malarzm: Partially I agree but since AUTO strategy is the default one its misuse is the most problematic (see #819 or tests I needed to fix in this PR like [this one](b4fe40bae403ec471d26b10e734947a0L106)), I think we can safely assume that if somebody is changing strategy he is clever enough to pass correct id (or will have his own checks in the type).

malarzm:
>  I'm not sure about BC here, but IMO the best option would be to drop generatorOptions.type

Maybe trigger deprecation?

> writing a generator that generates a MongoId

What is the point of such generator?

> Either way we do this, we need to write some documentation about this feature.

I'll create a ticket for this later

alcaeus:
> What is the point of such generator?

The use case is a bit special: we needed predictable identifiers to trigger upserts but wanted to keep ```MongoId``` values. So we assigned a specific time identifier (the first 4 bytes of ```MongoId```) and added a hash based on document settings. Weird, but it works ;)

malarzm:
@alcaeus I've commited a change we discussed on IRC :)

alcaeus:
@malarzm Looks good. This way we ensure using ```MongoId``` with AUTO strategy but don't overly complicate the mapping. I've also created #1197 to track documentation changes.

PR #1191, opened 2015-08-07
ReferenceMany's sort should not be used with some strategies

### LINKED ISSUE #623: RemoveElement on sorted collection
I have a document Whatever with a collection "users" which reference many User. 
This collection is sorted with an annotation.

    /**
    * @ODM\ReferenceMany(targetDocument="User", simple=true, sort={"lastname"="asc", "firstname"="asc"})
    */
    protected $users = array();

If I try to delete one user reference from the collection. 
```php
$whatever->getUsers()->removeElement($user);
```
The user deleted is not the one I want to delete. There is an issue with the key here because of the sort annotation. The removeElement method should remove the element at the key x based on the order in the database.

Log :
<code>doctrine.INFO: MongoDB query: {"update":true,"query":{"_id":{"$id":"51c4a13ffa4634362a000061"}},"newObj":{"$unset":{"users.2":true}},"options":{"safe":true,"fsync":false,"timeout":30000},"db":"something","collection":"Whatever"}</code>

malarzm:
Fixes #623. Strategies using `$set` should be ok as they are setting whole collection so it will just become sorted in the database, others could unset wrong index. 

I was wondering whether to revive #524 but that would be perfect if indeed all references would be unique (or using `addToSet` strategy) but I'm not sure if we can assume someone is not depending on having multiple references...

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataInfo.php
@@ -1158,6 +1158,11 @@ class ClassMetadataInfo implements \Doctrine\Common\Persistence\Mapping\ClassMet
                 (isset($mapping['mappedBy']) || isset($mapping['inversedBy']))) {
             throw MappingException::owningAndInverseReferencesRequireTargetDocument($this->name, $mapping['fieldName']);
         }
+
+        if (isset($mapping['reference']) && $mapping['type'] === 'many' && ! isset($mapping['mappedBy'])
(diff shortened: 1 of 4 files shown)


alcaeus:
Looks good to me :+1: 
Maybe we can pick up #524 again for 1.x.

malarzm:
Merged manually in https://github.com/doctrine/mongodb-odm/commit/56796fa0b74d3a4b68fc0b266bcee942b4f8b3c3

I've scheduled #524 for 1.x to come back to that :)

PR #1192, opened 2015-08-08
Mind versioning when updating collections

### LINKED ISSUE #1046: Investigate Collection update and locking document

malarzm:
Closes #1046 

~~I've got absolutely no idea how this could be tested so my word that it should work must do :baby:~~ Actually I'll write a test later, it's too hot and I forgot I can just test `CollectionPersister` ;)

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/CollectionPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/CollectionPersister.php
@@ -20,6 +20,7 @@
 namespace Doctrine\ODM\MongoDB\Persisters;
 
 use Doctrine\ODM\MongoDB\DocumentManager;
+use Doctrine\ODM\MongoDB\LockException;
 use Doctrine\ODM\MongoDB\PersistentCollection;
 use Doctrine\ODM\MongoDB\Persisters\PersistenceBuilder;
 use Doctrine\ODM\MongoDB\UnitOfWork;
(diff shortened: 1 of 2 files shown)


malarzm:
Tests are added now

malarzm:
@jmikola @alcaeus I'm wondering if collection updates should take place before updating parent document (so before version field is incremented and before lock field is unset) - any thoughts?

alcaeus:
@malarzm Collection updates should respect the version and lock fields - so yes, in this case I think moving them ahead of regular document updates should be ok (I assume the atomic* strategies are not affected by this).
Could there be any unwanted side effects?

malarzm:
> I assume the atomic* strategies are not affected by this

No, these are included in the main query that handles locking and versioning

> Could there be any unwanted side effects?

No that I can think of at the minute, tests will tell us soon :)

malarzm:
> No that I can think of at the minute, tests will tell us soon :)

2 tests started failing after changing order, I'm wondering if it's worth fixing now or should we be served better by educating users that for lockable/versioned documents atomic strategies are the way to go.

jmikola:
> I'm wondering if collection updates should take place before updating parent document (so before version field is incremented and before lock field is unset) - any thoughts?

@malarzm: I believe that makes sense, but there's an edge case where the collection update succeeds but the actual document update triggers a LockException. In that case, we have a partial update (of the collection) applied, which leaves the document in an inconsistent state. I think the proper solution would be to increment the version (and UnitOfWork's expected version) each time we issue an update operation against the document. If we think about this in terms of MongoDB replication, that would mean we're bumping the version on each update entry in the oplog.

> 2 tests started failing after changing order

What was the change and what two tests started failing?

> I'm wondering if it's worth fixing now or should we be served better by educating users that for lockable/versioned documents atomic strategies are the way to go.

I'm in favor of suggesting that folks use atomic strategies with versioning. IIRC, that concurrency issue was the reason you introduced it for @blockjon in the first place.

jmikola:
@malarzm: I think this PR still improves the situation we currently had, where collection updates ignored versioning altogether. We may want to create a separate ticket based on what I described in https://github.com/doctrine/mongodb-odm/pull/1192#issuecomment-131675560 (each update and oplog entry bumping the version).

As-is, I'm in favor of merging this for 1.0.0, though.

PR #1194, opened 2015-08-09
Make easier PSR-4 compliance

castarco:
In order to make possible a clean fix in the DoctrineMongoDBBundle project I have added an extra parameter to the DocumentRepositoryGenerator->writeDocumentRepositoryClass method.

This extra parameter allows to generate classes in directories that doesn't directly map the namespace structure.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Tools/DocumentRepositoryGenerator.php
+++ b/lib/Doctrine/ODM/MongoDB/Tools/DocumentRepositoryGenerator.php
@@ -59,12 +59,20 @@ class <className> extends DocumentRepository
         return str_replace(array_keys($variables), array_values($variables), self::$template);
     }
 
-    public function writeDocumentRepositoryClass($fullClassName, $outputDirectory)
+    public function writeDocumentRepositoryClass($fullClassName, $outputDirectory, $outputDirectoryNamespace = null)
     {
         $code = $this->generateDocumentRepositoryClass($fullClassName);
-
+        
+        if (null === $outputDirectoryNamespace) {
+            $relativeClassName = $fullClassName;
+        } else {
+            $relativeClassName = preg_replace(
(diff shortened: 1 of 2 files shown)


malarzm:
@castarco thanks for the PR! I know we currently have no tests for this tool at all, but would you mind creating one (the best would be testing both old and new behaviour you just added)?

castarco:
OK, I'll try it :)

malarzm:
You may want to take a look here: https://github.com/doctrine/mongodb-odm/blob/master/tests/Doctrine/ODM/MongoDB/Tests/Tools/DocumentGeneratorTest.php

castarco:
@malarzm Hi again, I've added two simple tests to check the behaviours affected by my change. I haven't many time to add more coverage to other not related cases.

castarco:
In the last commit I also tested that it's possible to load and instantiate the generated repository classes.

malarzm:
@castarco thank you very much! Last (I hope) thing is that https://github.com/castarco/mongodb-odm/commit/eeb78f9ecc3638e728fd4230316cd3b24ec4213d makes travis fail, could you remove it? We'll live without coverage annotations :)

castarco:
Hi @malarzm , I've reverted the commit using `git revert` because I think it's the cleanest way to do it. If you want to completely erase the mistake from the git history I can do it too.

malarzm:
@castarco yes, I was thinking about wiping it out completely as it doesn't make much sense to merge commits with a revert of one of them to main repository :)

castarco:
@malarzm  Done. By the way, I'm curious about why the PHPUnit annotations break the TravisCI build ...

malarzm:
@castarco maybe it needed FQCN? Not sure but since we're not using it at all there's no need to bother :) Thank you for taking care of this!

PR #1204, opened 2015-08-17
Add documentation about custom ID generators

### LINKED ISSUE #1197: Improve documentation for custom ID generation strategies
The documentation is currently limited to a list of available generators. Especially the CUSTOM strategy and its options should be defined, e.g. to show how to define a different type since it's not always intuitive.

alcaeus:
Closes #1197.

### DIFF
--- a/docs/en/reference/basic-mapping.rst
+++ b/docs/en/reference/basic-mapping.rst
@@ -300,7 +300,7 @@ Here is how you can configure the strategy for the different configuration forma
                                 xsi:schemaLocation="http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping
                                                     http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping.xsd">
     
-            <document name="MyPersistentClass" customId="true">
+            <document name="MyPersistentClass">
                 <field name="id" id="true" strategy="NONE" />
             </document>
         </doctrine-mongo-mapping>
@@ -308,7 +308,6 @@ Here is how you can configure the strategy for the different configuration forma
     .. code-block:: yaml
 
         MyPersistentClass:
-          customId: true
           fields:
             id:
               id: true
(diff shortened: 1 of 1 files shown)


docs/en/reference/basic-mapping.rst
```diff
@@ -337,6 +336,53 @@ Now you can retrieve the document later:
 
     $document = $dm->find('MyPersistentClass', 'my_unique_identifier');
 
+When using the ``CUSTOM`` strategy you will have to define the type returned by the generator in your mapping:
```
jmikola: I'm not sure that this is required. See [this bit of ClassMetadataInfo](https://github.com/doctrine/mongodb-odm/blob/85d46370986ed81ed1ef462d9c9819a5f9bbe28a/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataInfo.php#L1126). Therein, we do infer the generate type from the "strategy" option; however, the mapping's type will be picked up from the following, in order:

 * `GENERATOR_TYPE_AUTO` always implies the "id" type
 * "type" field from mapping's "options" array, if it exists (the "options" array is what ClassMetadataInfo refers to as "generatorOptions")
 * The mapping's own "type" option, if it exists
 * If an incrementing generator is used, the "int_id" type is used
 * Else, "custom_id" is used

Furthermore, [ClassMetadataFactory](https://github.com/doctrine/mongodb-odm/blob/85d46370986ed81ed1ef462d9c9819a5f9bbe28a/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataFactory.php#L248) seems to require a "class" generator option, which points to a class implementing [AbstractIdGenerator](https://github.com/doctrine/mongodb-odm/blob/85d46370986ed81ed1ef462d9c9819a5f9bbe28a/lib/Doctrine/ODM/MongoDB/Id/AbstractIdGenerator.php).

alcaeus: /me facepalms. I had this block prepared before @malarzm finished #1190. In this case, I'll remove the change and only leave the minor bugfix (customId in XML and YML is not necessary).

docs/en/reference/basic-mapping.rst
```diff
@@ -337,6 +336,57 @@ Now you can retrieve the document later:
...
+.. configuration-block::
+
+    .. code-block:: php
+
+        <?php
+
+        /** Document */
+        class MyPersistentClass
+        {
+            /** @Id(strategy="CUSTOM", type="string", options={"class"="Vendor\Specific\Generator"}) */
```
jmikola: Just for the record, `type` is still optional here and the default behavior will be ODM using the CustomIdType class, which does absolutely no type-conversion to/from the database.

jmikola:
LGTM.

jmikola:
Manually merged in https://github.com/doctrine/mongodb-odm/commit/5632b821d0c46a3c1fd17e899c16e092f58890ef.

PR #1205, opened 2015-08-17
Don't import annotations whose names are reserved words in PHP7+

### LINKED ISSUE #1168: Deprecate annotations that conflict with reserved words in PHP7
See:

 * https://github.com/doctrine/phpcr-odm/issues/641
 * https://github.com/doctrine/phpcr-odm/pull/643

This is not terribly urgent, since the existing [mongo](http://pecl.php.net/package/mongo) extension will not be updated for PHP7 compatibility ([mongodb](http://pecl.php.net/package/mongodb) will); however, we should probably get this change in before 1.0.0 stable.

----

Per [PHP7 reserved words](http://php.net/manual/en/reserved.other-reserved-words.php), we need to not include `Int.php`, `Bool.php`, `String.php`, and `Float.php` from `DoctrineAnnotations.php` on PHP7+.

jmikola:
Fixes #1168


malarzm:
:+1:

alcaeus:
LGTM :+1:

Soullivaneuh:
Thank you! :+1: 

Could you please provide a new release tag for this?

PR #1206, opened 2015-08-17
Refactor cursor classes to compose CursorInterface

### LINKED ISSUE #639: ODM Cursor should only compose, not extend, base Cursor class
This would build upon changes in #638, which removed state mutation from the ODM cursor. At present, we still extend the base Cursor class for BC.

Much like Doctrine MongoDB's Cursor class composes the driver's MongoCursor class, the ODM Cursor class should do the same for our base Doctrine MongoDB Cursor. This would allow us to simplify the constructor considerably.

### LINKED ISSUE #1018: getFirstResult nulls Cursor's Limit
## Issue
Running `Doctrine\ODM\MongoDB\Cursor::getSingleResult()` will clear limit from Cursor.

## Example Code
```
<?php

$objects = $myObjectRepo->createQueryBuilder()
    ->limit(10)
    ->getQuery()
    ->execute();

// Access First Result:
$firstObject = $objects->getSingleResult();

// limit has been removed from $objects.
```

## Reason

This is the implementation of `getSingleResult` from `Doctrine\MongoDB\Cursor`.

```
    public function getSingleResult()
    {
        $originalLimit = $this->limit;
        $this->reset();
        $this->limit(1);
        $result = current($this->toArray(false)) ?: null;
        $this->reset();
        $this->limit($originalLimit);
        return $result;
    }
```
It's retrieving the cursors value using `$this->limit`. Since `Doctrine\ODM\MongoDB\Cursor` extends `Doctrine\MongoDB\Cursor`, it has a `cursor` property. `Doctrine\ODM\MongoDB\Cursor` acts as decorator for `Doctrine\MongoDB\Cursor` and should return all values from `Doctrine\ODM\MongoDB\Cursor::$baseCursor`. Accessing the property breaks or ignores this design and returns the null `$limit` value of `Doctrine\ODM\MongoDB\Cursor`.  

This could be fixed by creating a protected accessor `$this->getLimit()` on both Cursors, the ODM cursor though would return the result of `$this->baseCursor->getLimit()`.

alcaeus:
Fixes #1018, Closes #639.

This change refactors the Cursor class to no longer extend Doctrine\MongoDB\Cursor and instead implement the new cursor interface.
~~Currently tests are failing due to an issue with complex id types and toArray functionality, this can be fixed by always defining useKeys = false for eager cursors after doctrine/mongodb#223 has been merged.~~

Composing CursorInterface instead of the base cursor class also allows us to deprecate the ODM\EagerCursor class since the special handling related to eager cursors is done by doctrine/mongodb and does not need any special handling.

This also fixes a long-standing issue (#520) where using reference priming in combination with ```getSingleResult``` would still prime references for the entire dataset, not only the one document being fetched from the cursor.

Last but not least, this introduces ```~1.2``` as new version requirement for doctrine/mongodb and removes special treatment for eager cursors in DocumentPersister.

~~Note: since a few methods are still missing (see doctrine/mongodb#225) and there is no stable release of doctrine/mongodb containing the interface yet, this branch temporarily uses a dev version of doctrine/mongodb.~~


lib/Doctrine/ODM/MongoDB/Cursor.php
```diff
@@ -451,4 +602,82 @@ public function timeout($ms)
...
+    {
+        return $this->baseCursor->valid();
+    }
+
+    /**
+     * @param array $primers
+     * @param ReferencePrimer $referencePrimer
+     * @return self
+     */
+    public function enableReferencePriming(array $primers, ReferencePrimer $referencePrimer)
```
jmikola: As discussed in video chat, let's have this method (and `primeReferences()`) assert that the wrapped cursor is eager. If not, we can throw an exception.

This addresses the manual case of users creating their own ODM Cursor, since the query builder should not allow this anyway.

alcaeus: Done, ```enableReferencePriming``` now throws a ```BadMethodCallException``` if the wrapped cursor is not an ```EagerCursor```.

lib/Doctrine/ODM/MongoDB/Cursor.php
```diff
@@ -319,6 +387,19 @@ public function hydrate($hydrate = true)
     }
 
     /**
+     * @param array $document
+     * @return object
```
malarzm: This seem to return `array|object|null` as for instance `getSingleResult()`?

lib/Doctrine/ODM/MongoDB/Cursor.php
```diff
@@ -451,4 +603,86 @@ public function timeout($ms)
...
+        $this->primers = $primers;
+        return $this;
+    }
+
+    /**
+     * Prime references
+     */
+    protected function primeReferences()
+    {
+        if ($this->referencesPrimed || !$this->hydrate || empty($this->primers)) {
```
malarzm: CS: ` ! $this->hydrate`, also in `primeReferencesForSingleResult()`

lib/Doctrine/ODM/MongoDB/Cursor.php
```diff
@@ -207,6 +218,20 @@ public function batchSize($num)
     }
 
     /**
+     * Wrapper method for MongoCursor::count().
+     *
+     * @see \Doctrine\MongoDB\Cursor::count()
```
jmikola: Would it make more sense for these `@see` lines to point to the CursorInterface? I believe that also has doc-blocks, and then you could also do without the full namespace prefix, too (since the class name isn't ambiguous).

lib/Doctrine/ODM/MongoDB/Cursor.php
```diff
@@ -78,15 +80,35 @@ class Cursor extends BaseCursor
...
+     */
+    private $referencesPrimed = false;
+
+    /**
      * Constructor.
      *
-     * @param BaseCursor    $baseCursor  Doctrine\MongoDB\Cursor instance being wrapped
-     * @param UnitOfWork    $unitOfWork  UnitOfWork for result hydration and query preparation
-     * @param ClassMetadata $class       ClassMetadata for the document class being queried
+     * @param CursorInterface $baseCursor  Doctrine\MongoDB\Cursor instance being wrapped
```
jmikola: Shorten to "Cursor instance being wrapped".

lib/Doctrine/ODM/MongoDB/Cursor.php
alcaeus: Yes, definitely. Will change the docblocks accordingly.

lib/Doctrine/ODM/MongoDB/Cursor.php
```diff
@@ -319,6 +405,19 @@ public function hydrate($hydrate = true)
     }
 
     /**
+     * @param array $document
+     * @return array|object|null
+     */
+    private function hydrateDocument($document)
```
jmikola: If this is the only place `$unitOfWorkHints` are used, does it make sense for ODM to return a Doctrine MongoDB cursor when not hydrating? Does this class have any other value, or do we just want to guarantee that the user always has the ODM instance to work with, since it does have some extra methods.

alcaeus: Technically there's no problem with iterating over a cursor, then changing a setting (e.g. ```hydrate```) and then iterate over it again, so I'd actually always return an ODM cursor to allow working with that.

lib/Doctrine/ODM/MongoDB/EagerCursor.php
```diff
@@ -19,230 +19,14 @@
...
-use Doctrine\ODM\MongoDB\Query\Query;
-use Doctrine\ODM\MongoDB\Query\ReferencePrimer;
-
 /**
  * EagerCursor wraps a Cursor instance and fetches all of its results upon
  * initialization.
  *
  * @since  1.0
  * @author Jonathan H. Wage <jonwage@gmail.com>
+ * @deprecated Deprecated in favor of using Cursor
```
jmikola: Can we append "; will be removed in 2.0"?

jmikola:
LGTM.

PR #1211, opened 2015-08-19
Fix regression with cursor counts

alcaeus:
This fixes a regression introduced in e2f23bc3eea11393995f0a799a02f34380201606 that causes wrong counts to be returned.
The default behavior for count() in a cursor is to return all elements, disregarding any skip/limit set.


tests/Doctrine/ODM/MongoDB/Tests/Functional/CursorTest.php
```diff
@@ -91,4 +91,26 @@ public function runningEagerQueryWrapsEagerCursor()
...
+        $this->dm->flush();
+
+        $cursor = $this->dm->createQueryBuilder('Documents\User')
+            ->sort('username', 'asc')
+            ->limit(2)
+            ->getQuery()
+            ->execute();
+
+        $this->assertEquals(5, $cursor->count());
+        $this->assertEquals(2, $cursor->count(2));
```
malarzm: `->count(true)`?

alcaeus: Forgive me. Haven't had my morning coffee yet. ;)

PR #1212, opened 2015-08-19
Do not prime references for single result when document is empty

munkie:
This PR fixes issue
Given priming is enabled in query
And query result is empty
When _Cursor::getSingleResult()_ method is called
Then php error is triggered:
```
ReflectionProperty::getValue() expects parameter 1 to be object, null given
 ../mongodb-odm/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataInfo.php:1535
 ../mongodb-odm/lib/Doctrine/ODM/MongoDB/Query/ReferencePrimer.php:140
 ../mongodb-odm/lib/Doctrine/ODM/MongoDB/Cursor.php:716
 ../mongodb-odm/lib/Doctrine/ODM/MongoDB/Cursor.php:347
 ../mongodb-odm/tests/Doctrine/ODM/MongoDB/Tests/Functional/CursorTest.php:126
```

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Cursor.php
+++ b/lib/Doctrine/ODM/MongoDB/Cursor.php
@@ -707,7 +707,7 @@ class Cursor implements CursorInterface
      */
     protected function primeReferencesForSingleResult($document)
     {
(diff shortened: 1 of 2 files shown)


alcaeus:
Thanks!

PR #1213, opened 2015-08-20
Fix incorrect behavior of EagerCursor::count() method with foundOnly false flag

munkie:
This PR fixes issue:

Given document collection has 5 elements
When i create query with limit 2
And priming is enabled
And query is executed
Then eager cursor is returned
When count method is called with foundOnly false flag
Then count 5 is expected to be returned
But count 2 is returned instead

This happens because _\Doctrine\MongoDB\EagerCursor_ that is used as base cursor of _\Doctrine\ODM\MongoDB\EagerCursor_ does not support _$foundOnly_ flag in _count()_ method

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/EagerCursor.php
+++ b/lib/Doctrine/ODM/MongoDB/EagerCursor.php
@@ -19,6 +19,8 @@
 
 namespace Doctrine\ODM\MongoDB;
 
+use Doctrine\MongoDB\EagerCursor as BaseEagerCursor;
+
 /**
  * EagerCursor wraps a Cursor instance and fetches all of its results upon
  * initialization.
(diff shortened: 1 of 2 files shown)


alcaeus:
This is a limitation of eager cursors: they fetch all data at once to immediately close the database connection and no longer run any queries for the lifetime of the cursor object. Unfortunately, this means that $foundOnly functionality is not supported. We had a discussion on how to deal with this in doctrine/mongodb#209 and decided it's best to leave functionality as is:
* The newly created CursorInterface does not include the option as it is not common to all cursors
* Regular cursors by default count all elements, including those not returned due to skip/limit options. This is consistent with the behavior of [MongoCursor::count()](https://secure.php.net/manual/en/mongocursor.count.php)
* Regular cursors can be switched to ```$foundOnly = true```, again consistent with MongoCursor::count()
* Eager cursors are designed to query the database only once, which is why they can only handle ```$foundOnly = true```.

munkie:
@alcaeus  But, why eager cursor can't save both true and false foundOnly count values on initialization and return them later in count method ?

alcaeus:
Please read the discussion in doctrine/mongodb#209, starting with [this comment](https://github.com/doctrine/mongodb/pull/209#issuecomment-116811735) as to why storing an internal count is not an option.

PR #1214, opened 2015-08-20
Allows DateType to be more easily extended.

cryode:
Uses late static binding to allow `DateType` to be extended without the need to duplicate methods unnecessarily.

In my application, I want to use [nesbot/carbon](https://github.com/briannesbitt/carbon) in place of `DateTime` objects with all of my dates on the PHP side of things. I wanted this change to be universal, so rather than implementing a totally new custom type, I chose to override the original `date` type.

In doing so, I found that the `DateType` makes too many references to itself, both via `self::` and through the output of `__CLASS__` when generating the hydrator, making extension difficult. With the proposed changes, my extension remains extremely tidy:

```php
use Carbon\Carbon;
use Doctrine\ODM\MongoDB\Types\DateType;

class CarbonDateType extends DateType
{
    public static function getDateTime($value)
    {
        return Carbon::instance(parent::getDateTime($value));
    }
}
```

`DateTypeTest` passes all assertions, only skipping 32 or 64 bit respectively. I ran into a lot of other errors running PHPUnit as a whole, looking into those further to determine if that's on my end or what, so I apologize if I missed a related test that happens to fail.

If I can improve this pull request in any way, please let me know! Congratulations on your 1.0 release recently. :beers:

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Types/DateType.php
+++ b/lib/Doctrine/ODM/MongoDB/Types/DateType.php
@@ -44,7 +44,7 @@ class DateType extends Type
         if ($value instanceof \DateTime || $value instanceof \DateTimeInterface) {
             return $value;
         } elseif ($value instanceof \MongoDate) {
-            $datetime = self::craftDateTime($value->sec, $value->usec);
+            $datetime = static::craftDateTime($value->sec, $value->usec);
         } elseif (is_numeric($value)) {
             $seconds = $value;
             $microseconds = 0;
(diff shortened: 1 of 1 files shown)


cryode:
Whoops, I failed to check for minimum PHP version requirements on the ODM, and used the `::class` syntax which is >= 5.5. Replaced with `get_class($this)`. I originally benchmarked the two and saw that `static::class` was faster, for what it's worth.

malarzm:
Merged manually in https://github.com/doctrine/mongodb-odm/commit/d1054ff1bb5a08c140820cedb16af56904102084 - thanks @cryode!

PR #1217, opened 2015-08-21
add options for filter. This closes romankonz/DoctrineMongoODMModule#147

CheesEM:
Hi folks, 

first this is my first PR, so please be patient :)

I faced the issue, that i can`t enable filters by default in my zf2 application.
So i decided to create this PR. 
If this PR will be merged, i will also create a PR for DoctrineMongoODMModule.

In this PR I also modified a little bit the FilterCollectionTest::testEnable(), cause the this test had test enable and disable. So I split it up.


malarzm:
Hi @CheesEM, welcome aboard!

I'm not sure how DoctrineMongoODMModule allows you to define filters but unfortunately you can't change public method signature as it's backward compatibility break. I suggest that Module should be fixed instead of ODM.

PR #1219, opened 2015-08-24
Custom collection classes for EmbedMany and ReferenceMany

malarzm:
# Reasoning

In my application we have an `Application` document that embeds many `Section` objects and as time goes by the main document becomes bloated with methods like `getSortedSections()`, `getEnabledSections()`, `getSearchableSections()` and so on. Semantically it would be awesome to have custom `SectionCollection` and use `$app->getSections()->getEnabledSections()` and at the same time filtering methods that needs to exist in more than one place (think references) will be in one place as well.

# It will be as simple as

```
/** @ODM\Document */
class DocumentWithCustomCollection
{
    /** @ODM\Id */
    private $id;

    /** @ODM\EmbedMany(
     *   collectionClass="MyEmbedsCollection",
     *   targetDocument="EmbeddedDocumentInCustomCollection"
     * ) 
     */
    private $embeds;

    public function __construct()
    {
        $this->embeds = new MyEmbedsCollection();
    }

    /** @return MyEmbedsCollection */
    public function getEmbeds()
    {
        return $this->embeds;
    }

    public function setEmbeds(MyEmbedsCollection $embeds)
    {
        $this->embeds = $embeds;
    }
}
``` 

`MyEmbedsCollection` has to implement `Doctrine\Common\Collections\Collection` interface.

# Behind the scenes

If user didn't specified new `collectionClass` property he will still use good ol' `ArrayCollection` wrapped by good ol' `PersistentCollection` (so no BC breaks - so far so good). If the user opts for a custom collection ODM will generate specialized `PersistentCollection` class to work with the collection. It works a bit like `Proxy` (even few generation methods are borrowed from this) so a class is generated that decorates original class ensuring that initialization takes place before working with decorated collection. Generated class also extends user's one to allow typehinting, code completion and don't tie developer to the ODM.

# TODO

- [x] Generated collections are now `eval`ed, need to implement other generation strategies (same as for Hydrators and Proxies)
- [x] Code generation should be moved out of `PersistentCollectionFactory` to its own `PersistentCollectionGenerator`
- [x] Write docs
- [x] Add more tests

Things that were in the list but no longer are:

1. Allow adding logic to traited methods - the only things that are coming to my mind are dirty (to say least) - let's put this on hold and see if people will really need (and have a good use case) for this
2. Possibly use Instantiator to create empty collections - I think it's not necessary anymore, `PersistentCollectionFactory` is small enough to roll own one if in need for additional dependencies in collection
3. Additional checks that `ProxyFactory` had seem to be guards for old PHP reflection bugs, I think there's no need to have them now in 2015

# Making it Doctrine-wide feature

So far I planned this to be ODM feature but some interfaces and/or concrete classes could be extracted to doctrine/common and used in other Doctrine projects.

/cc @doctrine/owners @lsmith77 @dbu feel free to ping more people that can be interested


dbu:
i quite like this idea and think it should be a general doctrine feature. for the instantiation of collections, maybe check what the jms serializer does - it can also unserialize objects from xml, json, ... and it allows to configure custom collections.

malarzm:
Code has been rebased onto current `dev-master` and description updated:

1. PHP 5.4 chapter is removed as we already require 5.5+
2. Known limitations have been addressed and removed

docs/en/reference/annotations-reference.rst
```diff
@@ -373,6 +373,9 @@ Optional attributes:
     strategy - The strategy used to persist changes to the collection. Possible
     values are ``addToSet``, ``pushAll``, ``set``, and ``setArray``. ``pushAll``
     is the default. See :ref:`collection_strategies` for more information.
+-
+    collectionClass - A full class name that implements ``Collection`` interface
```
jmikola: Should this be "fully qualified class name"? I think we use that elsewhere, sometimes abbreviated as FQCN.

docs/en/reference/annotations-reference.rst
```diff
@@ -373,6 +373,9 @@ Optional attributes:
...
+    and is used to hold documents. By default Doctrine's ``ArrayCollection`` is used.
```
jmikola: Either "By default, ..." or "Doctrine's ``ArrayCollection`` is used by default."

docs/en/reference/annotations-reference.rst
```diff
@@ -1117,6 +1120,9 @@ Optional attributes:
...
```
jmikola: Same as https://github.com/doctrine/mongodb-odm/pull/1219/files#r48495834

docs/en/reference/custom-collections.rst
```diff
@@ -0,0 +1,161 @@
+.. _custom_collection:
+
+Custom Collections
+==================
+
+.. note::
+    This feature was introduced in version 1.1
+
+By default Doctrine uses ``ArrayCollection`` implementation of its ``Collection``
```
jmikola: "By default, Doctrine uses the ``ArrayCollection`` implementation..."

docs/en/reference/custom-collections.rst
```diff
@@ -0,0 +1,161 @@
...
+interface to hold both embedded and referenced documents alike which are later
```
jmikola: Either "both embedded and referenced documents" or "embedded and referenced documents alike", but we don't need both terms :)

docs/en/reference/custom-collections.rst
```diff
@@ -0,0 +1,161 @@
...
+        public function __construct()
+        {
+            $this->sections = new ArrayCollection();
+        }
+
+        // ...
+    }
+
+For most cases this solution is sufficient but more sophisticated domains could use
+their own collections (e.g. ones ensuring contained objects are sorted) or to simply
```
jmikola: I'd rewrite as "(e.g. a collection that ensures its contained objects are sorted)"

docs/en/reference/custom-collections.rst
```diff
@@ -0,0 +1,161 @@
...
+add common filtering methods that otherwise would be added to owning documents.
```
jmikola: "would otherwise be added to the owning document's class."

docs/en/reference/custom-collections.rst
```diff
@@ -0,0 +1,161 @@
...
+
+Own collection classes
+----------------------
+
+Using your own ``Collection`` implementation is as simple as updating your ``@EmbedMany``
+or ``@ReferenceMany`` mapping with ``collectionClass`` parameter and changing constructor:
```
jmikola: "as simple as specifying the ``collectionClass`` parameter in the ``@EmbedMany`` or ``@ReferenceMany`` mapping and ensuring that your custom class is initialized in the owning class' constructor:"

docs/en/reference/custom-collections.rst
```diff
@@ -0,0 +1,161 @@
...
+        }
+
+        // ...
+    }
+
+For most cases this solution is sufficient but more sophisticated domains could use
+their own collections (e.g. ones ensuring contained objects are sorted) or to simply
+add common filtering methods that otherwise would be added to owning documents.
+
+Own collection classes
```
jmikola: Should this be "Defining a Custom Collection Class", "Custom Collection Classes", or something to that effect?

docs/en/reference/custom-collections.rst
```diff
@@ -0,0 +1,161 @@
...
+
+        public function __construct()
+        {
+            $this->sections = new SectionCollection();
+        }
+
+        // ...
+    }
+
+and implementing your class. If you are satisfied with ``ArrayCollection`` and only want
```
jmikola: I think we can drop "and implementing your class" here or rewrite as "The SectionCollection class must then be implemented".

docs/en/reference/custom-collections.rst
```diff
@@ -0,0 +1,161 @@
...
+to sprinkle it with some filtering methods you may just extend it:
```
jmikola: Comma between the clauses: "filtering methods, you may"

docs/en/reference/custom-collections.rst
```diff
@@ -0,0 +1,161 @@
...
+    {
+        public function getEnabled()
+        {
+            return $this->filter(function(Section $s) {
+                return $s->isEnabled();
+            });
+        }
+    }
+
+or you may want to implement whole class from scratch:
```
jmikola: Rewrite as a full sentence: "Alternatively, you may want to implement the whole class from scratch:"

docs/en/reference/custom-collections.rst
```diff
@@ -0,0 +1,161 @@
...
+
+        public function __construct(array $elements = array())
+        {
+            $this->elements = $elements;
+        }
+
+        // your implementation of all methods interface requires
+    }
+
+Taking control over collection's __construct
```
jmikola: Rewrite as: "Taking Control of the Collection's Constructor"

docs/en/reference/custom-collections.rst
```diff
@@ -0,0 +1,161 @@
...
+--------------------------------------------
+
+By default Doctrine assumes it can instantiate your collections in same manner as
```
jmikola: > By default, Doctrine assumes that it can instantiate your collections in same manner as an ``ArrayCollection`` (i.e. the only parameter is an optional PHP array); however, you may want to inject additional dependencies into your custom collection class. This will require you to create a ``PersistentCollectionFactory`` `interface <https://github.com/doctrine/mongodb-odm/blob/master/lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionFactory.php>`_ implementation (either from scratch or by extending our ``AbstractPersistentCollectionFactory``):

docs/en/reference/custom-collections.rst
```diff
@@ -0,0 +1,161 @@
...
+            switch ($collectionClass) {
+                case SectionCollection::class:
+                    return new $collectionClass(array(), $this->eventDispatcher);
+                default:
+                    return new $collectionClass;
+            }
+        }
+    }
+
+and register it in the ``Configuration``
```
jmikola: > The factory class must then be registered in the ``Configuration``:

docs/en/reference/reference-mapping.rst
```diff
@@ -18,6 +18,12 @@ constructed internally during hydration. As a developer, you should develop with
 the ``Collection`` interface in mind so that your code can operate with any
 implementation.
 
+.. note::
+
+    New in 1.1: you are no longer limited to using ``ArrayCollection`` class
```
jmikola: Either write "the ``ArrayCollection`` class" or simply "using ``ArrayCollection`` and can..."

docs/en/reference/annotations-reference.rst
malarzm: I used that because `targetDocument` few lines above had same construct, shall we change all?

malarzm:
Failing test will be fixed once we merge #1328

malarzm:
Tests are passing now but I screwed rebasing and lost commit amended on my work machine and not fetched in home, will fix that tomorrow -.-'

malarzm:
For those who'd like to test the feature but don't have a time to prepare own collections I've prepared small library: https://github.com/malarzm/collections - will cut a release somewhen along merging this PR :)

docs/en/reference/annotations-reference.rst
jmikola: I'm up for changing it across the board for consistency. Bonus points if we can get the RST to do acronym tags for us so we can just write FQCN and get the abbreviation on hover via the HTML tags.

malarzm: For tomorrow me: http://stackoverflow.com/questions/34645787/abbreviation-abbr-element-in-restructuredtext

lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -482,4 +551,53 @@ public function getRepositoryFactory()
...
+     * Get the persistent collection generator.
+     *
+     * @return DefaultPersistentCollectionGenerator
+     */
+    public function getPersistentCollectionGenerator()
+    {
+        if ( ! isset($this->attributes['persistentCollectionGenerator'])) {
+            $this->attributes['persistentCollectionGenerator'] = new DefaultPersistentCollectionGenerator(
+                $this->attributes['persistentCollectionDir'],
+                $this->attributes['persistentCollectionNamespace']
```
alcaeus: I just stumbled over this...when using a custom collection without having configured anything, there is a generic error: `Undefined index: persistentCollectionDir`. Using the getters here would change this to a more descriptive error message: `Doctrine\ODM\MongoDB\PersistentCollection\PersistentCollectionException: You must configure a PersistentCollection directory. See docs for details.`.

alcaeus:
Gave these changes a spin the last few days, works as advertised and expected. Great stuff!

Issue #1220, opened 2015-08-27
Recursive function calls using set strategy and a repositoryMethod


rubenrua:
See https://github.com/rubenrua/mongodb-odm/commit/ba8d02412e063e0ad5362e2d404f5df5b1e4d1bb

rubenrua:
Added a test and a possible fix in the pull request #1221.

malarzm:
Let's keep discussion (if there'll be one in #1221)

PR #1221, opened 2015-08-27
Recursive function calls using set strategy and a repositoryMethod #1220

rubenrua:


lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -778,9 +778,15 @@ public function createReferenceManyInverseSideQuery(PersistentCollection $collec
...
         $documents = $cursor->toArray(false);
-        foreach ($documents as $document) {
-            $collection->add($document);
+        foreach ($documents as $key => $obj) {
+            if (CollectionHelper::isHash($mapping['strategy'])) {
+                $collection->set($key, $obj);
+            } else {
+                $collection->add($obj);
+            }
+
```
malarzm: Could you maybe remove this extra line (and squash the commit)? Other than that this looks good to me (I was wondering about forcing collection into initialized status but this seems to be cleaner).

rubenrua:
@malarzm extra line removed. 
Looking forward for a 1.0.2 release.

malarzm:
Thanks @rubenrua! As for 1.0.2 I will tag it probably somewhen next week

PR #1222, opened 2015-08-29
Lifecycle callbacks should receive event arguments

### LINKED ISSUE #902: Lifecycle callbacks should receive event arguments
See [ListenersInvoker](https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/Event/ListenersInvoker.php) in ORM.

malarzm:
Closes #902 

This PR contains a tiny BC break - before `preLoad` lifecycle callback received `$data` by reference directly, now it receives a `PreLoadEventArgs` as it should BUT I wouldn't count it as BC break though - that event wasn't even documented.

### DIFF
--- a/docs/en/reference/events.rst
+++ b/docs/en/reference/events.rst
@@ -153,7 +153,11 @@ the life-time of their registered documents.
 - 
    postUpdate - The postUpdate event occurs after the database update
    operations to document data.
-- 
+-
+   preLoad - The preLoad event occurs for a document before the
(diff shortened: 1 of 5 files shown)


malarzm:
Rebased atop current `master` branch. FWIW Doctrine Extensions are not using `preLoad` event and brief googling (first 5 pages for "doctrine preload event") haven't brought any results (neither did adding "mongodb" to search query) - let's merge this once tests pass

PR #1225, opened 2015-09-03
Functional test to demonstrate that modifications to embedManys do not work as expected if @preUpdate hooks are used

blockjon:
When preUpdate hooks are used, Doctrine ODM does not handle changes to embedMany elements correctly.

The attached functional test only passes when the preUpdate hook is removed.

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1225Test.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1225Test.php
@@ -10,16 +10,15 @@ class GH1225Test extends \Doctrine\ODM\MongoDB\Tests\BaseTest
     public function testAddOnUninitializedCollection()
     {
         $doc = new GH1225Document();
-        $mySubDoc = new GH1225EmbeddedDocument('foo');
-        $doc->embeds->add($mySubDoc);
+        $doc->embeds->add(new GH1225EmbeddedDocument('foo'));
         $this->dm->persist($doc);
         $this->dm->flush();
         $this->dm->clear();
 
         $doc = $this->dm->getRepository(get_class($doc))->find($doc->id);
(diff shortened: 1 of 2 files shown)


malarzm:
@blockjon for the record I do want to fix this but I'm going on vacations next week and as usual there's like million things to take care of so I won't promise I'll be the one to fix the bug before October. I may be able to guide somebody though, just ping me on IRC before Monday

blockjon:
@malarzm Not a problem... we have a workaround where there's a very strange coding pattern that helps us avoid this problem..but once you are back I would love to see a fix so we can simply our code. Thank you!

malarzm:
FWIW these two things seem easier than weird patterns (or maybe you're even using one of these :) ).

1. `$doc->embeds->add($embeddedDoc);` -> `$doc->embeds->add(clone $embeddedDoc);`
2. `$doc->embeds = new ArrayCollection(array($doc->embeds->first()));` instead of clearing and adding

Both things above are making the test pass however I agree the fail is strange :)

malarzm:
Ah got it, now I mentioned cloning or not calling `clear()` it becomes kind of obvious: calling `clear()` schedules all [embedded documents for deletion](https://github.com/doctrine/mongodb-odm/blob/c3933ad5eb0c23e7262f60efa1ad12605484bf48/lib/Doctrine/ODM/MongoDB/PersistentCollection.php#L628-L632) and later they're not picked up by `PersistenceBuilder` because of their state. IIRC persisting document scheduled for deletion will unschedule it from deletion so probably the same thing can be done for embedded documents - if embedded documents that are scheduled for deletions are found during change set calculation their state needs to be changed.

As for `preUpdate` exhibiting the bug I believe that it's second change set calculation that makes a difference as it takes place after lifecycle callback is executed. Anyway I'll try to go down the path described before and see if nothing breaks

malarzm:
PR incoming!

alcaeus:
Fix has been merged in #1248.

PR #1230, opened 2015-09-17
Functional test to demonstrate incorrect value in PreUpdate changeset with embedded document

mxr:
In an `@ODM\PreUpdate` hook, the DM changeset's 'original value' for an embedded document is actually the changed value. The attached test fails with the following output:

```
$ vendor/bin/phpunit /Users/maxr/src/mongodb-odm/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1230Test.php
PHPUnit 4.8.7 by Sebastian Bergmann and contributors.

F

Time: 127 ms, Memory: 7.50Mb

There was 1 failure:

1) Doctrine\ODM\MongoDB\Tests\Functional\Ticket\GH1230Test::testEmbeddedDocChangesetInPreUpdateHook
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
     'stringProperty' => Array (...)
     'embeddedDoc' => Array (
         0 => Doctrine\ODM\MongoDB\Tests\Functional\Ticket\GH1230EmbeddedDocument Object (
-            'value' => 'embedded doc value'
+            'value' => 'updated embedded doc value'
         )
         1 => Doctrine\ODM\MongoDB\Tests\Functional\Ticket\GH1230EmbeddedDocument Object (...)
     )
 )

/Users/maxr/src/mongodb-odm/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1230Test.php:71
/Users/maxr/src/mongodb-odm/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataInfo.php:549
/Users/maxr/src/mongodb-odm/lib/Doctrine/ODM/MongoDB/UnitOfWork.php:1194
/Users/maxr/src/mongodb-odm/lib/Doctrine/ODM/MongoDB/UnitOfWork.php:426
/Users/maxr/src/mongodb-odm/lib/Doctrine/ODM/MongoDB/DocumentManager.php:528
/Users/maxr/src/mongodb-odm/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1230Test.php:28

FAILURES!
Tests: 1, Assertions: 1, Failures: 1.
```

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1230Test.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1230Test.php
@@ -5,19 +5,19 @@ namespace Doctrine\ODM\MongoDB\Tests\Functional\Ticket;
 use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
 use Doctrine\ODM\MongoDB\Tests\BaseTest;
 
-class GHXXXXTest extends BaseTest
+class GH1230Test extends BaseTest
 {
     public function testEmbeddedDocChangesetInPreUpdateHook()
     {
-        $doc = new GHXXXXDocument();
-        $doc->embeddedDoc = new GHXXXXEmbeddedDocument('embedded doc value');
+        $doc = new GH1230Document();
+        $doc->embeddedDoc = new GH1230EmbeddedDocument('embedded doc value');
         $doc->stringProperty = 'string property';
 
         $this->dm->persist($doc);
         $this->dm->flush();
         $this->dm->clear();
 
-        /** @var GHXXXXDocument $doc */
+        /** @var GH1230Document $doc */
         $doc = $this->dm->getRepository(get_class($doc))->find($doc->id);
         $doc->embeddedDoc->value = 'updated embedded doc value';
         $doc->stringProperty = 'updated string property';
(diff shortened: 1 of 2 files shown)


malarzm:
Technically this is correct, the embedded document has changed but since it was only modified both old value and new value are the same (since object is reused). I think you should be getting change set for the embedded document itself instead of relying on parent's one (or make your VOs immutable).

mxr:
Thanks @malarzm, that makes sense. Indeed if I update the test to look at the embedded doc's changeset then I can access both the new and old values, and the test passes:

```diff
diff --git a/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1230Test.php b/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1230Test.php
index 1e2960e..95a3bc7 100644
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1230Test.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1230Test.php
@@ -58,16 +58,12 @@ class GH1230Document
     {
         $this->test->assertEquals(
             array(
-                'stringProperty' => array(
-                    'string property',
-                    'updated string property',
-                ),
-                'embeddedDoc'    => array(
-                    new GH1230EmbeddedDocument('embedded doc value'),
-                    new GH1230EmbeddedDocument('updated embedded doc value'),
+                'value' => array(
+                    'embedded doc value',
+                    'updated embedded doc value',
                 ),
             ),
-            $this->test->getDm()->getUnitOfWork()->getDocumentChangeSet($this)
+            $this->test->getDm()->getUnitOfWork()->getDocumentChangeSet($this->embeddedDoc)
         );
     }
 }
```

cc @bkorsunsky and @blockjon as FYI

bkorsunsky:
That's great the the embedded doc's changeset is correctly represented. 

Does this mean that we should define preUpdate hooks that inspect the embeddedDocs in the embeddedDoc classes themselves? or that we should iterate through each embeddedDoc in the parent class' preUpdate hook?

malarzm:
AFAIR preUpdate [is fired](https://github.com/doctrine/mongodb-odm/blob/c3933ad5eb0c23e7262f60efa1ad12605484bf48/lib/Doctrine/ODM/MongoDB/UnitOfWork.php#L1231) for embedded documents as well?

bkorsunsky:
So in order to detect and record the changesets of embedded docs, we could use the preUpdate hook to detect that a change is registered and then recurse into embeddedDoc to get the details of the change. 

It would definitely be more convenient to surface the changeset at the top level `getDocumentChangeSet()` function, but the above approach will do for now.

PR #1236, opened 2015-09-30
Avoid null embedded documents.

aheinz-sg:
We recently uncovered some null embedded documents in our mongodb. Though it is not clear how they were introduced, their existence causes loadEmbedManyCollection to die with:

```
Symfony\Component\Debug\Exception\ContextErrorException occured doing REDACTED, message: Catchable Fatal Error: Argument 2 passed to Doctrine\ODM\MongoDB\UnitOfWork::getClassNameForAssociation() must be of the type array, null given, called in /REDACTED/vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php on line 615 and defined in /REDACTED/vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/UnitOfWork.php line 2702
```

Adding some defensive coding to avoid type hint fatal.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -612,6 +612,9 @@ class DocumentPersister
         $owner = $collection->getOwner();
         if ($embeddedDocuments) {
             foreach ($embeddedDocuments as $key => $embeddedDocument) {
+                if (!is_array($embeddedDocument)) {
+                    continue;
+                }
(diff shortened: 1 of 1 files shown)


malarzm:
I'm not sure if silently ignoring `null`s from database is the best idea as index calculation for certain strategies will be off then.

alcaeus:
I concur with @malarzm - silently ignoring `null` values shouldn't happen, as we always assume that each item in the collection is of the type indicated in the mapping. The issue you're having points to a deeper issue that shouldn't be silently ignored.

PR #1242, opened 2015-10-12
Fixing ClassMetadata to work on case sensitive filesystems

drymek:
When BsonFilter class gets to mock, it shows case sensitive problem. If any class inherits from it, the problem does not exists - the inherited class can change typed argument (ClassMetaData to ClassMetadata). The only way I know how to test it,  is to creating a mock.

### DIFF
--- a/docs/en/reference/filters.rst
+++ b/docs/en/reference/filters.rst
@@ -29,7 +29,7 @@ should be accessed via ``BsonFilter::getParameter()``.
 
     namespace Vendor\Filter;
 
-    use Doctrine\ODM\MongoDB\Mapping\ClassMetaData;
+    use Doctrine\ODM\MongoDB\Mapping\ClassMetadata;
     use Doctrine\ODM\MongoDB\Query\Filter\BsonFilter;
 
     class MyLocaleFilter extends BsonFilter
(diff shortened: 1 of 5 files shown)


malarzm:
:+1: 

@drymek was that the only appearance of `ClassMetaData`?

drymek:
Hm... didn't check that before. There are:
```sh
./lib/Doctrine/ODM/MongoDB/Query/FilterCollection.php:    use Doctrine\ODM\MongoDB\Mapping\ClassMetaData;
./tests/Doctrine/ODM/MongoDB/Tests/Query/Filter/Filter.php:    use Doctrine\ODM\MongoDB\Mapping\ClassMetaData;
./docs/en/reference/filters.rst:    use Doctrine\ODM\MongoDB\Mapping\ClassMetaData;
```
will do that.

malarzm:
@drymek thanks, ping me when you're done and I'll merge the changes :)

drymek:
```bash
grep "ClassMetaData" -R ./ | wc -l
0
```
Could not write test for *FilterCollection*. *BsonFilter*-like test (creating a mock) pass without fixing the code. It seems like only caused errors for abstract classes (?). All files are fixed but *FilterCollection* not covered by any test.

//cc @malarzm

malarzm:
I wouldn't worry too much about no tests here :) I see you did separate commit, can you squash please?

drymek:
done.

malarzm:
Thank you @drymek!

PR #1244, opened 2015-10-13
Add note concerning distinct and sort

franek:
It is not possible to combine distinct with sort().

### DIFF
--- a/docs/en/reference/query-builder-api.rst
+++ b/docs/en/reference/query-builder-api.rst
@@ -182,7 +182,7 @@ method:
 
 The above would give you an ``ArrayCollection`` of all the distinct user ages!
 
-.. note:: Please note that you can not combined distinct with ``sort()`` due to `mongo driver limitations <http://php.net/manual/fr/mongocollection.distinct.php>`_.
+.. note:: Please note that you can not combined distinct with ``sort()`` due to `mongo driver limitations <http://php.net/manual/en/mongocollection.distinct.php>`_. ``distinct()`` method only accept two parameters ($key and $query). We can't provide a sort.
 
 Refreshing Documents
 ~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


docs/en/reference/query-builder-api.rst
```diff
@@ -182,6 +182,8 @@ method:
 
 The above would give you an ``ArrayCollection`` of all the distinct user ages!
 
+.. note:: Please note that you can not combined distinct with ``sort()`` due to `mongo driver limitations <http://php.net/manual/fr/mongocollection.distinct.php>`_.
```
malarzm: 1) This should point to english manual
2) Where is the limitation described?

docs/en/reference/query-builder-api.rst
```diff
@@ -182,6 +182,8 @@ method:
 
 The above would give you an ``ArrayCollection`` of all the distinct user ages!
 
+.. note:: Please note that you can not combined distinct with ``sort()`` due to `mongo driver limitations <http://php.net/manual/en/mongocollection.distinct.php>`_. ``distinct()`` method only accept two parameters ($key and $query). We can't provide a sort.
+
```
jmikola: It never occurred to me that one would attempt to sort the result of a distinct command, but perhaps that's an SQL thing folks are used to with the ORM? In any event, I suggest rewriting this as:

> MongoDB's `distinct command <http://docs.mongodb.org/manual/reference/command/distinct/>`_ does not support sorting, so you cannot combine ``distinct()`` with ``sort()``. If you would like to sort the results of a distinct query, you will need to do so in PHP after executing the query.

Note that this also isn't a driver limitation. I think it's clearer to simply refer to the command documentation itself.

I'm a bit confused as to why one would try to sort the results of

jmikola:
Manually merged (with edits) in https://github.com/doctrine/mongodb-odm/commit/eee069b5c41f8ef27db22c5fa9ef4670059f6e03

franek:
thx. Your rewording is better.
Sorting could be done on another field different from the one sent to distinct().

PR #1248, opened 2015-10-15
Unschedule orphan removal if an element is re-added

alcaeus:
This fixes a minor issue where removing an element from a PersistentCollection with orphanRemoval enabled and then re-adding it to the same collection leaves it scheduled for removal. This can produce invalid references.
This PR also fixes an issue where using ```set()``` in a PersistentCollection with orphanRemoval enabled would not remove an element that was overwritten and is no longer part of the collection.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/PersistentCollection.php
+++ b/lib/Doctrine/ODM/MongoDB/PersistentCollection.php
@@ -187,9 +187,13 @@ class PersistentCollection implements BaseCollection
             $newObjects = $this->coll->toArray();
         }
 
+        $this->initialized = true;
+
         $this->coll->clear();
         $this->uow->loadCollection($this);
         $this->takeSnapshot();
+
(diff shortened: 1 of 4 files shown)


lib/Doctrine/ODM/MongoDB/PersistentCollection.php
```diff
@@ -551,6 +551,11 @@ public function add($value)
         }
         $this->coll->add($value);
         $this->changed();
+
+        if ($this->uow !== null && $this->isOrphanRemovalEnabled()) {
```
malarzm: Should same thing be done for `set()`?

alcaeus: Yes. There was a bug that elements were not removed when being overwritten by set() - this is done properly now.

malarzm:
For the record @gabma's issue was resolved in #1256

PR #1249, opened 2015-10-15
[Docs] Fix wrong text about DocumentManager

danizord:
- Replaces "ORM" by "ODM"
- Also removes the "2" from "Doctrine 2" because Doctrine Mongo ODM is currently in 1.X

PS: Github added EOL EOF automatically.


malarzm:
Thank you @danizord!

PR #1251, opened 2015-10-19
Using repositoryMethod should imply inverseSide

malarzm:
Lately I omitted on purpose `mappedBy` on `ReferenceMany` using `repositoryMethod` since there was no owning side and using `count()` on such uninitialized `PersistentCollection` was stubbornly returning `0`.


malarzm:
Unless it makes sense to persist/remove documents via such collections (using `cascade`) - it surely is possible (tested that)? In that case `PersistentCollection::count` should be fixed (https://github.com/doctrine/mongodb-odm/blob/df142796394681d0f77416ec92c66972ce0ba14b/lib/Doctrine/ODM/MongoDB/PersistentCollection.php#L522)

jmikola:
IMO, `repositoryMethod` collections are inverse-side, and should probably be considered read-only if we ever get around to implement that (I know we discussed it for standard inverse-side collections).

This issue seems like a fix for a previous oversight, so :+1: from me.

alcaeus:
I'm also :+1: on this. As for making inverse collections read-only, there was no consensus on this yet ;)

PR #1252, opened 2015-10-21
Allow reusing embedded documents and collections

malarzm:
So this is the thing that was popping out lately a lot. I must admit that first direction I took was throwing exception when such embedded document was found BUT since we allowed reusing/cloning persistent collection loong long time ago I couldn't. ~~So I'd say make reusing embedded documents easier for now and forbid both things in 2.0?~~

Thoughts? If we'd get around to merge this I'll write test ~~(and hopefully create a ticket scheduled for 2.0)~~ EDIT: actually I'm leaning towards keeping this, indeed it might be better for the developer to not care

/cc @alcaeus @jmikola @jwage

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
+++ b/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
@@ -745,7 +745,7 @@ class UnitOfWork implements PropertyChangedListener
                 // created one. This can only mean it was cloned and replaced
                 // on another document.
                 if ($actualValue instanceof PersistentCollection && $actualValue->getOwner() !== $document) {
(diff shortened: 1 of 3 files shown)


alcaeus:
I like the idea. This way we don't change the behavior people would expect (meaning that they can just copy them around as they please) to an error. :+1:

jmikola:
This seems reasonable. Any reason you'd want to remove the cloning behavior in 2.0?

malarzm:
> Any reason you'd want to remove the cloning behavior in 2.0?

To force proper usage :) Anyway I'll try to finish this today

alcaeus:
Merged manually.

malarzm:
:+1:

PR #1253, opened 2015-10-23
Update LICENSE

kevinyien:
The license was pretty out of date :)


malarzm:
Indeed it was, thank you!

Issue #1149, opened 2015-06-15
Can't flush only specific documents


dorongutman:
I'm trying to call flush([$documentOne, $documentThree]) so it'll only flush these two documents and not others which I have not specified.

However, the ODM flushes all the documents that are pending.
Specifically, it flushes a scheduled "insert" which is a new document that's not referenced by any other document.

malarzm:
After some internal discussion it turned out that `flush()` with documents specified (or entities in ORM's case, where behaviour is the same) was not really meant to be "we will save just these and nothing else" restriction. This is documented [here](https://github.com/doctrine/mongodb-odm/blob/1eaba57d84c2d7702966f2fe253aba1244d76959/lib/Doctrine/ODM/MongoDB/UnitOfWork.php#L515-L519) but I think it could be exposed better than on UoW's private method :)

dorongutman:
@malarzm oh wow, that's very much confusing.. http://doctrine-mongodb-odm.readthedocs.org/en/latest/reference/working-with-objects.html#flushing-single-documents

Anyways - is there any way of flushing only a single document ? Maybe with the querybuilder or something  ?

malarzm:
@dorongutman heh yes, you're right, we need to write a disclaimer there as well... To be honest I was a bit surprised as well, but this predates my work here very much ;) Anyway right now I see two possible ways:

1. As you suggest QueryBuilder might be the way to go (but there will be no events and preparation)
2. Change your flow to not schedule unwanted documents or detach them first

Also there's a chance we may diverge from ORM here and actually allow flushing only specified documents. One of main obstacles would be that current commit saves all collections BUT #1155 changes that to update collections only when their owner is updated (so outstanding things will be orphan removal and updating UoW internal state). I've marked this yesterday as an idea to allow discussion whether we want to introduce the change or not

alcaeus:
#1707 tracks the removal of `flush($document)` in 2.0 - the correct solution is to use change tracking policies.

PR #1257, opened 2015-10-29
NOTIFY change tracking policy fixes

banrobert:
This PR fixes some issues with documents using the NOTIFY change tracking policy, namely:
* Committing a document of this kind multiple times raised an `Undefined index` notice as reported in #740. Fixed the error and created a test case for it.
* Committing only a single document made the UnitOfWork lose changes of all other managed documents with NOTIFY change tracking policy. Fixed this too and also created a test case for it.
* Additionally I found that the current test case for the this change tracking policy (namely: `UnitOfWorkTest::testChangeTrackingNotify`) used a document without the required `@ODM\ChangeTrackingPolicy("NOTIFY")` annotation as its test subject, so the test didn't test what it supposed to test. I added the missing annotation to this dummy document.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
+++ b/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
@@ -690,7 +690,7 @@ class UnitOfWork implements PropertyChangedListener
             // and we have a copy of the original data
             $originalData = $this->originalDocumentData[$oid];
             $isChangeTrackingNotify = $class->isChangeTrackingNotify();
-            if ($isChangeTrackingNotify && ! $recompute) {
+            if ($isChangeTrackingNotify && ! $recompute && isset($this->documentChangeSets[$oid])) {
                 $changeSet = $this->documentChangeSets[$oid];
             } else {
                 $changeSet = array();
--- a/tests/Doctrine/ODM/MongoDB/Tests/UnitOfWorkTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/UnitOfWorkTest.php
@@ -253,9 +253,18 @@ class UnitOfWorkTest extends \Doctrine\ODM\MongoDB\Tests\BaseTest
         $this->uow->persist($entity);
 
         $this->uow->commit($entity);
-        @$this->uow->commit($entity);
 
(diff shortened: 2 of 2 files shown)


malarzm:
@banrobert thank you for PR! Before I get to review could you please take a look at failing test on Travis?

banrobert:
Updated the test case that failed previously, it should work now.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -441,13 +441,32 @@ public function commit($document = null, array $options = array())
...
+        // if not the whole unit of work is committed,
+        // then don't remove all the change sets
+        if (is_object($document)) {
+            $class = $this->dm->getClassMetadata(get_class($document));
+            $oid   = spl_object_hash($document);
+
+            unset($this->documentChangeSets[$oid]);
+            unset($this->scheduledForDirtyCheck[$class->name][$oid]);
+        } elseif (is_array($document)) {
+            foreach ($document as $object) {
```
malarzm: This will not be enough as inserted documents should have their change sets cleared as well (and embedded documents that were committed etc), see #1149

malarzm: Having said that I think the PR will be better without this at all and instead we should work on #1149 for 1.x

banrobert: You might be right. Until a decision is made on #1149 it's not certain that this fix is what needed here and as you said it's not even complete. I'll remove this from the PR.

malarzm:
@banrobert cool thanks, could you remove that two commits (434045a8c092b2610e4fb0eec3d9e5ae3507c30a and 0d06d666fdb319c14c6d47a0312c9b23ca678aa7) at all? Once that is done PR's good to be merged.

As for #1149 I'll poke rest of the team for thoughts, FWIW I think it does make sense to implement it.

banrobert:
Removed the commits, never rebased before, didn't know this was possible, cool feature.

And regarding #1149 i would also say that implementing the single document flush differently than the ORM but making it clearer by not flushing anything else that is not related to the document in question is the way to go. Hope the team agrees!

malarzm:
Merged, thank you @banrobert!

PR #1259, opened 2015-10-29
Set parentAssociation on persist and fix embeds by then

### LINKED ISSUE #509: Deep cloning a document and it's embedded documents
I have recently encountered some strange behavior regarding cloning a document object that contains embedded documents. The use case here is to create a copy of the original document with new identity, but exactly same details.

It seems that `$newDocument = clone $originalDocument` does not clone the embedded documents. `$documentManager->persist($newDocument);` works find and also embedded documents are copied at that point. However, if an embedded document of `$newDocument` is modified before (or after persisting), the change is coupled also to the `$originalDocument`.

Consider the following test case (added to be a part of EmbeddedTest.php)

```php
	public function testCloneParentAndModifyEmbeddedWithoutChangeInOriginal()
    {
        $address = new Address();
        $address->setAddress('20780 Kaarina');

        $user = new User();
        $user->setUsername('vmattila');
        $user->setAddress($address);

        $this->dm->persist($user);
        $this->dm->flush();
        $this->dm->clear();

        $user = $this->dm->createQueryBuilder('Documents\User')
            ->field('id')->equals($user->getId())
            ->getQuery()
            ->getSingleResult();
        $this->assertNotNull($user);
        
		// Cloning user
		$newUser = clone $user;
		$newUser->setId(null); // Nulling identity of the user
		$this->assertEquals('20780 Kaarina', $newUser->getAddress()->getAddress());
		
		// Changing address for the new user
		$newUser->getAddress()->setAddress('20100 Turku');
		$this->dm->persist($newUser);
        $this->dm->flush();
		
		$this->assertEquals('20100 Turku', $newUser->getAddress()->getAddress());
		$this->assertEquals('20780 Kaarina', $user->getAddress()->getAddress());
    }
```

The test fails, because `$newUser->getAddress()->setAddress('20100 Turku');` will couple the change also to `$user->getAddress()->setAddress('20100 Turku');`.

This behaviour might be expected (is it?) - but a little bit confusing. It would be useful if Doctrine would provide helpers to make a "deep copy" of the document.

malarzm:
Follow up of #1252 which also takes care of fixing reused embedded documents (and collections) as soon as document is persisted. As a side effect (heh) parent association is now available just after the persist (which is neat since it'll be available in `prePersist` event - now I wonder if tests for this should be added).

Also closes #509 as we won't get any nearer behaviour that was wanted there.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
+++ b/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
@@ -2332,7 +2332,6 @@ class UnitOfWork implements PropertyChangedListener
      *
      * @param object $document
      * @param array $visited
-     * @param array $insertNow
      */
     private function cascadePersist($document, array &$visited)
     {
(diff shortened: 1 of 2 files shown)


alcaeus:
Nice! :+1:

PR #1261, opened 2015-10-30
No backup option

kevinmlong:
add no-backup option. default is false. if you want to have backups, need to specify --generate-backup=true on command line.

For future consideration: Update documentation to specific that options where false is an option must be specified as <option>=0, since false is recognized as a string and thus resolves to TRUE in php.

Thanks!


malarzm:
Can we be in line with ORM's command? See https://github.com/doctrine/doctrine2/blob/b1451cda5253f34bf2eca146bbfabd7c6448da1c/lib/Doctrine/ORM/Tools/Console/Command/GenerateEntitiesCommand.php#L84-L87

kevinmlong:
We could, I was trying to keep it consistent with what it means to have it on the command line. I'll submit a new PR.

malarzm:
@kevinmlong appreciate the intention however for the sake of consistency of tools I'd rather be with ORM here :)

malarzm:
Thanks for the update, could you squash the commits please?

kevinmlong:
@malarzm - done!

malarzm:
Thank you @kevinmlong!

PR #1263, opened 2015-10-31
[WIP] Add support for hydrated aggregation results

alcaeus:
~~~Note: this PR is on hold until #1269 is merged as that will remove support for driver versions that don't support returning command cursor for aggregation queries.~~~

This PR builds on aggregation builder added in doctrine/mongodb#213 and adds hydrated aggregation results to ODM.

Similar to ```createQueryBuilder```, one now has ```createAggregationBuilder``` in DocumentManager and DocumentRepository classes. This will create an ODM aggregation builder that adds support for ```references``` and ```includesReferenceTo``` in match stages as well as type handling in those stages as well (type handling for other stages to be added later). The aggregation builder itself is extended to provide optional hydration for aggregation result documents via a new AggregationResultDocument type.

This new mapping type was necessary to handle the special cases for aggregation results:
* aggregation result documents may have complex IDs (think ```$group``` stages where the id might actually be an embedded document)
* aggregation result documents should not be managed by UnitOfWork
* aggregation result documents should not be persisted to the database

Note: hydrated aggregation will only be available for MongoDB 2.6 and later. MongoDB 2.2 and 2.4 do not support the cursor option for aggregation commands so no hydration results for those versions.

TODO list:
* [x] Add builder with hydration support
* [x] Ensure type handling and field name translation in match stage
* [x] Ensure type handling in other stages
* [x] Ensure field name translation in `$sort` and `$unwind` stages
* [ ] Make `$lookup` stage work with referenced documents (TBD)
* [ ] Ensure discriminator fields work for aggregation results
* [ ] Ensure filters are applied to aggregation queries
* [ ] Write tons of tests for above cases
* [ ] Write documentation


alcaeus:
A big thing about ODM aggregation builder is going to be name and type conversion in pipelines. Just a quick primer:
* **name conversion** means that field names in queries are always changed to their database names. Let's take the following example:
```
/** @Field(name="blogTitle") */
public $title;
```
In a query builder, these two statements produce the exact same database query:
```
$qb->field('title')->equals('Something');
$qb->field('blogTitle')->equals('Something');
```
The result will always be `{'blogTitle': 'Something'}`.
* **type conversion** means that the value in a query will be converted to the type appropriate for the field in question. Again, an example:
```
/** @Field(type="date") */
public $createDate;
```
In a query builder, any object given is converted to a `MongoDate` object:
```
$qb->field('createDate')->lte(new \DateTime());
```

Now, in theory all of this holds true for an aggregation builder as well, **as long as we haven't modified the source document**. Once it has been modified, we lose all previously defined mapping information. A few examples:
* New fields can be defined in the pipeline, so when encountering a new field named `title` in a `$project` stage we can't change the name to `blogTitle` because it might be used as `title` in the result document
* Continuing the example above, when a following `$match` stage contains a condition on a `title` field, we don't know whether to apply it to the original `blogTitle` field or whether to apply it to the newly defined field
* Even further, we don't have type mapping for newly defined fields, so when encountering a field in a `$match` stage that has been defined previously, we don't know whether to do any type conversion or not
* Type conversion might even apply when there is no field context involved at all since we can do arbitrary expressions in a pipeline. For example, the following expression is perfectly valid but we don't have any mapping information whatsoever: `{age: { $subtract: [new ISODate(), { $max: [ '$createDate', '$updateDate' ] } } }`. Handling expressions like these would require way too much logic in the aggregation builder.

For **type conversion**, we could add type discovery to all existing types. This means that a query/aggregation builder can take an arbitrary value and ask the `Doctrine\ODM\MongoDB\Types\Type` class which specific type handles the given value. This way we'd always know how to convert non-scalar database values. We might lose some type conversion (e.g. converting a string to int because the field mapping tells us it's an int field) but I think the cons of trying to follow fields through a pipeline greatly outweigh the pros.

For **name conversion**, I'm not sure what to do. Obviously we'd want to have it for compatibility reasons (people might not understand why they have to use different names in query and aggregation builders) but I'm afraid doing so might require us to look at the pipeline more than necessary (e.g. do name conversion only if the field has not been defined by a previous stage but comes from the source document).

@jmikola @malarzm et al, I'd appreciate feedback on this one.

malarzm:
How about we remove `->hydrate()` method and instead allow developer optionally pass mapping to each stage? Exceptions would be:

1. Last stage MUST specify mapping as results will be hydrated into these objects
2. Stages like `$limit`, `$skip`, `$sort` or other alike could inherit mapping from previous stage

If mapping is passed we'll do everything we can to ease developer's life and in case it's not we'll be like `Query\Builder` which doesn't do all conversions (even if it could, but that's for a different PR as we discussed on IRC some time ago) and generally assume developer knows what is he doing.

alcaeus:
@malarzm I'm not sure how we would allow developers to pass mapping information inline (you wouldn't want to create files for every aggregation stage in the pipeline) in a way that doesn't make complex pipelines (much) more difficult to read.

As for the last stage specifying mapping: this is optional as well, since a developer could want unhydrated results. In fact, that's the default and a developer explicitely has to turn it on using the `hydrate` method (which works opposite from the same method in the query builder and thus might be renamed).

malarzm:
> you wouldn't want to create files for every aggregation stage in the pipeline

I wouldn't but maybe someone would like to to get type casting et al, writing class with a bunch of public properties that are annotated isn't that big deal

alcaeus:
I'm not worried about type conversion that much since we can at least cast whatever is involved to something that makes sense (e.g. cast `DateTime` objects to `MongoDate`). When doing `field('foo')->equals(new \DateTime)` the resulting `MongoDate` value might not be of the same type as the mapping for `foo` specifies which is something I can live with.

speedytwenty:
There's probably a better place to ask this, but I'm curious to know if mapReduce can be used with this Aggregation Builder.

I have tested this PR and it is exactly what I need but I also need be able to use mapReduce as well. Any insight is much appreciated!

alcaeus:
The aggregation result document type could be used for hydrating map reduce queries but I haven't thought of that yet. Maybe I'll need to rename this to "QueryResultDocument" to make it more generic

speedytwenty:
Thanks for the quick response! Guess I'll go ahead and describe what I am trying to do as I suspect it would be a popular use-case.

Essentially, we are using embedded documents heavily. Our embedded documents have unique ids and we need to fetch one or more embedded documents that reside one or more levels deeps in the document. I'm hoping to be able to use this aggregationBuilder to query embedded documents without loading the entire parent document, while returning the hydrated embedded documents.

I may be missing something, but it looks like the AggregationResultDocument doesn't provide any additional functionality yet. But very eager to see what you might have in mind for this!

alcaeus:
Ah, projection in hydrated queries. Yes, that's a classic. Aggregation builder won't completely solve your problem. It will allow you to only fetch certain embedded documents but you won't be able to persist changes to them as aggregation result documents are treated as read-only in UnitOfWork. So, any changes you make to them won't be persisted to the database since ODM won't know where to put them.

What you'd need are partially hydrated proxy documents which then trigger a database load when an unhydrated property is being accessed. This is far more complicated but it is in the back of my head for an upcoming release. Of course, I can't give you a guarantee that we'll ever see it happen.

speedytwenty:
I certainly understand nothing is guaranteed in software until it's done! Just wanted to share my use case. I anticipated the read-only limitation but it's not a problem in my case. Thanks for the response, and good luck. If I can help with tests or anything let me know!

alcaeus:
Closed in favor of #1448. Please direct all discussion there.

PR #1266, opened 2015-11-04
Inverse side of RefOne is no longer included in change set

### LINKED ISSUE #817: PostUpdate event occurs even if database operation is not performed in bi-directional references on the inverse side
[In this example](http://docs.doctrine-project.org/projects/doctrine-mongodb-odm/en/latest/reference/bidirectional-references.html#one-to-one) if I have a Customer and link it to a Cart instance, a postUpdate event generated on Customer too.

Documentation says: "The postUpdate event occurs _after the **database update**_ operations to document data."

But database update occurred only on Cart (owning side, holding the reference), not on Customer (inverse side).

Is it a bug or the documentation is wrong? Events should reflect document or database operations?

malarzm:
Fixes #817

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
@@ -93,7 +93,7 @@ class PersistenceBuilder
                 $insertData[$mapping['name']] = Type::getType($mapping['type'])->convertToDatabaseValue($new);
 
             // @ReferenceOne
-            } elseif (isset($mapping['association']) && $mapping['association'] === ClassMetadata::REFERENCE_ONE && $mapping['isOwningSide']) {
+            } elseif (isset($mapping['association']) && $mapping['association'] === ClassMetadata::REFERENCE_ONE) {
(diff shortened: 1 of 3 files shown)


malarzm:
Marked as WIP because there are 2 strange things happening if

1. inverse side checks are removed from `PersistenceBuilder`
2. inverse side checks are not done by `UnitOfWork::computeAssociationChanges`

and I think it's worth ironing this out

malarzm:
1. Fixed
2. This is happening because [this test](https://github.com/doctrine/mongodb-odm/blob/c0c7220de7aad8638e0389508e401cc4da245f2a/tests/Doctrine/ODM/MongoDB/Tests/Functional/OwningAndInverseReferencesTest.php#L198) uses [this inverse side collection](https://github.com/doctrine/mongodb-odm/blob/c0c7220de7aad8638e0389508e401cc4da245f2a/tests/Documents/BlogPost.php#L20-L21) to cascade persist. Even if we'd think this shouldn't be allowed I wouldn't change that in a bugfix release

alcaeus:
Merged manually into 1.0.x branch in https://github.com/doctrine/mongodb-odm/commit/790110e33b638fb1e16bed617056bf507499e7fb.

PR #1268, opened 2015-11-06
Update to Method Generation

kevinmlong:
Added capability for collection types, so that the the entire collection can be set. 

**An Example:** 

``` php
/**
* Set widgets
*
* @param array $widgets
* @return self
*/
public function setDevices(array $widgets = array()) {
        $this->widgets = $widgets;
        return $this;
}

...

$object->setWidgets($arrayOfWidgets)
```

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Tools/DocumentGenerator.php
+++ b/lib/Doctrine/ODM/MongoDB/Tools/DocumentGenerator.php
@@ -693,6 +693,9 @@ public function <methodName>()
                 if ($code = $this->generateDocumentStubMethod($metadata, 'get', $fieldMapping['fieldName'], '\Doctrine\Common\Collections\Collection')) {
                     $methods[] = $code;
                 }
+                if ($code = $this->generateDocumentStubMethod($metadata, 'set', $fieldMapping['fieldName'], 'Doctrine\Common\Collections\Collection')) {
+                    $methods[] = $code;
+                }
             }
         }
 


malarzm:
`\Doctrine\Common\Collections\Collection` is more correct typehint

malarzm:
Also thank you @kevinmlong for the PR!

kevinmlong:
I'll update it and push the changes later today. Thanks!

kevinmlong:
@malarzm - change has been made and pushed

malarzm:
@kevinmlong something is wrong, please see Travis

kevinmlong:
@malarzm - I saw the problem. I changed the type hint, but forgot to change the default value. A default of `array()` was not obviously not correct. For now, I deleted the default value.

kevinmlong:
@malarzm - All tests have passed.

lib/Doctrine/ODM/MongoDB/Tools/DocumentGenerator.php
```diff
@@ -788,6 +791,9 @@ private function generateDocumentStubMethod(ClassMetadataInfo $metadata, $type,
 
         $types = Type::getTypesMap();
         $methodTypeHint = $typeHint && ! isset($types[$typeHint]) ? '\\' . $typeHint . ' ' : null;
+        if($methodTypeHint == "\array "){
```
malarzm: I think this is not needed anymore?

docs/en/_theme
```diff
@@ -1 +0,0 @@
-Subproject commit dc294be1dbcf9abde82d24aea1902898d6ee6314
```
malarzm: Why is this change here?

kevinmlong:
@malarzm - Fixes made and tests are running. I cleaned up the commit log. Something funky happened with that subproject. Should be good now.

lib/Doctrine/ODM/MongoDB/Tools/DocumentGenerator.php
```diff
@@ -693,6 +693,9 @@ private function generateDocumentStubMethods(ClassMetadataInfo $metadata)
                 if ($code = $this->generateDocumentStubMethod($metadata, 'get', $fieldMapping['fieldName'], '\Doctrine\Common\Collections\Collection')) {
                     $methods[] = $code;
                 }
+                if ($code = $this->generateDocumentStubMethod($metadata, 'set', $fieldMapping['fieldName'], 'Doctrine\Common\Collections\Collection')) {
```
malarzm: @kevinmlong sorry for the delay. Last thing that struck me odd is that there's no leading `\` here while there is one in getter - is this correct?

kevinmlong: @malarzm I believe it is. When I had the preceding slash there before, the generation resulted in there being the slashes '\\Doctrine\Common\...'. For now, I would say leave it as is. I'll research this more when I have a free block to focus on it. Thanks!

malarzm:
@kevinmlong any update on trailing `\` investigation? :)

alcaeus:
I have to ask: do we want `setSomeCollection` methods in all generated documents? We have `add`, `remove` and `get`, which is what you should be doing with a collection, but replacing the entire collection at once is hardly a standard use case, so I'd think that one should be added on demand by the developers themselves.

alcaeus:
After a long delay and some internal discussion, I'm closing this pull request.

For one, replacing an entire collection is not a use-case I'd consider common enough to warrant adding a method by default - remember, you may still add one if you need it. I think our set of `add`, `remove` (for updates) and `get` for reads should be sufficient for most use-cases, which is what the document generator should try to cover.

There's another reason though: the document generator is not something to be used often. It was derived from the entity generator, which again was designed to quickly generate entities for legacy systems (where you already have a mapping since the database structure already exists). In a regular development lifecycle, you'd design your entities **before** mapping fields to the database, so there's no need to use the document generator at all. That (and the decision by the ORM team) is one of the reasons why we're considering removing the document generator for 2.0.

PR #1269, opened 2015-11-07
Bump PHP requirement to 5.5

malarzm:
As warned in the [1.0.3 release blogpost](http://www.doctrine-project.org/2015/11/03/doctrine-mongodb-odm-release-1-0-3.html) this is happening :)


malarzm:
Note to myself: remove all internal checks for driver version (and my `preg_match` for `MongoId` validation could now be changed as well)

PR #1276, opened 2015-11-13
Use fieldName in mapping instead of name when persisting

monteiro:
This is a bug that happens when the name annotation of an embed document is different of the field name itself.

Related: https://github.com/doctrine/mongodb-odm/commit/85d91b74b64057d9f812d030d6d15d51e9fa278d


malarzm:
Thank you @monteiro! All looks good to me, I'll get to merging today evening or tomorrow

monteiro:
Thanks!

malarzm:
Also backported into `1.0.x` branch in https://github.com/doctrine/mongodb-odm/commit/03793a23049afcad53114323529ce9383e598351. Thanks again @monteiro!

Issue #1275, opened 2015-11-13
Sorting collections on 1.0.3 leads to data loss


sharpner:
We are currently sorting persistent collections using something like this (pseudo code)
```
$currentPosition = $this->collection->indexOf($item);

$newPosition = $currentPosition + $move;
if ($newPosition < 0) {
    $newPosition = 0;
} elseif ($newPosition >= $this->collection->count()) {
    $newPosition = $this->collection->count() - 1;
}

if ($move < 0) {
    for ($index = $currentPosition; $index > $newPosition; $index--) {
        $this->collection->set($index, $this->collection->get($index - 1));
    }
} else {
    for ($index = $currentPosition; $index < $newPosition; $index++) {
        $this->collection->set($index, $this->collection->get($index + 1));
    }
}

$this->collection->set($newPosition, $item);
```

But since 1.0.3 sorting leads to missing elements between requests. 

I reproduced this writing a test in our software using persist/flush after each sort, which fails on 1.0.3 but succeeds on 1.0.2.

I currently trying to write a test in the mongodb-odm but I need to get into that first...

Any hints/ideas?

malarzm:
Which collection strategy are you using?

sharpner:
```
* @ODM\ReferenceMany(
     *     targetDocument="Item",
     *     cascade={"refresh","persist"},
     *     orphanRemoval="true",
     *     strategy="atomicSet"
     * )
```

malarzm:
Ah crap, no idea then at a glance :/ failing test case will be appreciated

sharpner:
K will give it a try :) thanks for the fast reply.

malarzm:
@sharpner if there is any chance for a failing test case today I'd be able to spend some time tomorrow on fixing it (unless you'd do that for me as well ;) )

sharpner:
I'm on it ;) first time fiddling in the odm, takes me so me time :p

sharpner:
Okay I'm lost. I migrated the test to the odm project, but it succeeds. 

Logically that means the Bug must be in our code, but, a downgrade to 1.0.2 fixes it for us.... Please don't start to work on it, I will have to look into it further on monday. 

Thanks for the help so far.

malarzm:
@sharpner sure, thanks for the heads up!

sharpner:
Okay we've found the culprit. We are using the Gedmo translatable extension. 

[...]

PR #1277, opened 2015-11-13
Add test to reproduce behaviour

sharpner:
#1275 

We are totally brainfucked, but this should reproduce our confusing somewhat.

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1275Test.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1275Test.php
@@ -34,7 +34,7 @@ class GH41275Test extends \Doctrine\ODM\MongoDB\Tests\BaseTest
         $container->add($itemThree);
 
         $this->assertSame(
-            ['Number One', 'Number Two', 'Number Three'],
+            array('Number One', 'Number Two', 'Number Three'),
             array_map($getNameCallback, $container->items->toArray())
         );
 
@@ -46,7 +46,7 @@ class GH41275Test extends \Doctrine\ODM\MongoDB\Tests\BaseTest
         $this->dm->refresh($container);
 
         $this->assertSame(
-            ['Number One','Number Three', 'Number Two'],
+            array('Number One','Number Three', 'Number Two'),
             array_map($getNameCallback, $container->items->toArray())
         );
     }
@@ -77,7 +77,7 @@ class GH41275Test extends \Doctrine\ODM\MongoDB\Tests\BaseTest
         $container->add($itemThree);
 
         $this->assertSame(
-            ['Number One', 'Number Two', 'Number Three'],
+            array('Number One', 'Number Two', 'Number Three'),
             array_map($getNameCallback, $container->items->toArray())
         );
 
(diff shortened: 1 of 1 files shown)


tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1275Test.php
```diff
@@ -0,0 +1,239 @@
...
+        $container->add($itemOne);
+        $container->add($itemTwo);
+        $container->add($itemThree);
+
+        $this->assertSame(
+            array('Number One', 'Number Two', 'Number Three'),
+            array_map($getNameCallback, $container->items->toArray())
+        );
+
+        $this->dm->refresh($container);
```
alcaeus: This will revert the changes made to ```$container``` in L75..L77 since they haven't been flushed to the database.

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1275Test.php
```diff
@@ -0,0 +1,350 @@
...
+
+        $this->assertSame(
+            array('one', 'two', 'three'),
+            array_map($getNameCallback, $container->$strategy->toArray())
+        );
+
+        $two = $container->$strategy->get(1);
+        $three = $container->$strategy->get(2);
+        $container->items->set(1, $three);
```
malarzm: I lost some hair due to `$container->items` here :D

gruberro: Oh damn. I've reviewed the code so often... I'm so sorry :-(

malarzm: No biggie, don't worry ;)

gruberro:
Great... I was able to reproduce the error I'm getting within our application by writing a failing but broken test case. I'll invest some time tomorrow to debug our application stack as the error must be somewhere there. Maybe some broken event subscriber or something similar - will see. @malarzm: do you wan't me to remove the test case?

malarzm:
@gruberro I've included all tests from here in #1278 so whatever suits you :)

gruberro:
I finally found the "making it buggy" trigger (https://github.com/doctrine/mongodb-odm/commit/ab89a77dd3a50707e348ba26b2e8235cc9acd629). By using a ```preUpdate``` lifecycle callback (even if it's doing nothing) elements are removed from the collection. I lost some hair either in the meantime ;)

malarzm:
I suppose that'll be because UnitOfWork recomputes change sets for document that had `preUpdate` callback - I remember that the same trigger was reported when reusing embedded document that was already scheduled for orphan removal.

sharpner:
So is this a bug or expected behaviour?

malarzm:
Unfortunate but expected (as of now) behaviour I'd say.

sharpner:
Even though it loses data? with 1.0.2 it works, and with 1.0.3 we lose items...

malarzm:
@alcaeus wrote full explaination on why is this happening: https://github.com/doctrine/mongodb-odm/issues/1275#issuecomment-156501140. If we don't merge #1278 for whatever reasons (I for one don't like making it O(n)) you may add items in new collection and replace the old one entirely.

alcaeus:
@sharpner @gruberro Closing this to consolidate discussion into #1278.

We'll take a look at what exactly is going on with preUpdate and see if we can work around it.

PR #1278, opened 2015-11-15
Check for duplicates before scheduling orphanRemoval

### LINKED ISSUE #1275: Sorting collections on 1.0.3 leads to data loss

malarzm:
Fixes #1275 and closes #1277 BUT increases complexity of `set` and `removeElement` when `orphanRemoval` is enabled

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/PersistentCollection.php
+++ b/lib/Doctrine/ODM/MongoDB/PersistentCollection.php
@@ -441,7 +441,7 @@ class PersistentCollection implements BaseCollection
 
         $this->changed();
 
-        if ($this->isOrphanRemovalEnabled()) {
+        if ($this->isOrphanRemovalEnabled() && ! $this->contains($element)) {
             $this->uow->scheduleOrphanRemoval($element);
         }
 
@@ -539,7 +539,7 @@ class PersistentCollection implements BaseCollection
 
         // Handle orphanRemoval
         if ($this->uow !== null && $this->isOrphanRemovalEnabled() && $oldValue !== $value) {
-            if ($oldValue !== null) {
+            if ($oldValue !== null && ! $this->contains($oldValue)) {
                 $this->uow->scheduleOrphanRemoval($oldValue);
             }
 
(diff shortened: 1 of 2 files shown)


alcaeus:
I took another look at the whole embedding issue and it is the same problem as with referenced documents: orphanRemoval (which is implied for embedded collections).

For referenced documents, the sorting bug (when switching two elements) manifests itself by removing the referenced document but leaving the reference in the collection (thus leading to a proxy initialization failed error).
For embedded documents however the document would not be removed directly from the database **until the next changeset computation**. This is why everything seems fine until after the preUpdate lifecycle callback: after the callback is invoked, changeset computation sees that the element in question was removed (documents scheduled for orphanRemoval are scheduled for removal before inserts and updates are executed) and also removes it from the collection.

Since both problems boil down to orphanRemoval, this PR effectively fixes that weird behavior as well.

malarzm:
How about we stop using `orphanRemoval` for embeds? If firing lifecycle events is relying on this we should be able to employ `getDeleteDiff` I believe?

malarzm:
Superseded by #1281

PR #1279, opened 2015-11-18
Changed XML mapping examples

joenivl:
The XML examples of the embedded mapping are using the attribute 'fieldName'. It should be the attribute 'field' so the example is the same for PHP and YAML. Also the attribute 'field' is required for the embedded elements.


malarzm:
Merged manually in https://github.com/doctrine/mongodb-odm/commit/4b4967870034be830d2db0854d6b5ccd0b1187d7 - thank you @joenivl and sorry for not doing this sooner!

joenivl:
@malarzm no prob, and thank you for making awesome software!!

PR #1280, opened 2015-11-19
Rework scheduling for orphan removal

malarzm:
Supersedes #1278 and personally I like this one better. Also I was thinking if it's worth removing embedded documents out of orphan removal's interest but probably that would mean duplicating logic for cascading.


malarzm:
Opened from wrong branch

PR #1281, opened 2015-11-19
Moved logic for scheduling orphanRemoval

malarzm:
Supersedes #1278 and personally I like this one better. Also I was thinking if it's worth removing embedded documents out of orphan removal's interest but probably that would mean duplicating logic for cascading.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
+++ b/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
@@ -207,7 +207,7 @@ class UnitOfWork implements PropertyChangedListener
     private $evm;
 
     /**
-     * Embedded documents that are scheduled for removal.
+     * Additional documents that are scheduled for removal.
      *
      * @var array
      */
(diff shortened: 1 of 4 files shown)


alcaeus:
:+1: on this.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -898,12 +898,18 @@ private function computeAssociationChanges($parentDocument, array $assoc, $value
...
 
-        if ($value instanceof PersistentCollection && $value->isDirty() && $assoc['isOwningSide']) {
+        if ($value instanceof PersistentCollection && $value->isDirty() && ($assoc['isOwningSide'] || isset($assoc['embedded']))) {
             if ($topOrExistingDocument || CollectionHelper::usesSet($assoc['strategy'])) {
                 $this->scheduleCollectionUpdate($value);
             }
             $topmostOwner = $this->getOwningDocument($value->getOwner());
             $this->visitedCollections[spl_object_hash($topmostOwner)][] = $value;
+            if ( ! empty($assoc['orphanRemoval']) || isset($assoc['embedded'])) {
+                $value->initialize();
```
alcaeus: Since both remove and set (which are the only methods that can remove elements) initialize the collection before making changes I'd argue that we could actually just skip uninitialized collections here. That could save a few seconds when having a large uninitialized collection sitting around.

malarzm: `set` doesn't really initialize collection when used: https://github.com/doctrine/mongodb-odm/pull/1281/files#diff-23b2d33965df3f6d7b558d98a301a34aL538

malarzm: For the record discussed behaviour was changed for a moment (https://github.com/malarzm/mongodb-odm/pull/6) but as agreed on IRC it's not going to make it into ODM

PR #1284, opened 2015-11-24
Correct the getFilterCollection method call

coudenysj:


malarzm:
Merged manually in https://github.com/doctrine/mongodb-odm/commit/02d0de3341a15a040366eaa35638dec268ef1cb6 - thank you @coudenysj and sorry for  for not doing this sooner!

PR #1289, opened 2015-12-01
Rename `$scheduledForDirtyCheck` to `$scheduledForSynchronization`

malukenho:


alcaeus:
Nice, thanks. How about renaming adding ```isScheduledForSynchronization``` and ```scheduleForSynchronization``` and marking ```isScheduledForDirtyCheck``` and ```scheduleForDirtyCheck``` as deprecated? That would bring the method names inline with the property name.

malukenho:
@alcaeus done.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -1506,10 +1505,18 @@ public function isScheduledForUpdate($document)
...
-    public function isScheduledForDirtyCheck($document)
+    public function isScheduledForSynchronization($document)
     {
         $class = $this->dm->getClassMetadata(get_class($document));
-        return isset($this->scheduledForDirtyCheck[$class->name][spl_object_hash($document)]);
+        return isset($this->scheduledForSynchronization[$class->name][spl_object_hash($document)]);
+    }
+
+    /**
+     * @deprecated
```
malarzm: Should we add something like "since 1.1 to be removed in 2.0" ?

malarzm:
Just for the record, renaming things for the sake of renaming things is annoying especially if it involves deprecations :P

malukenho:
@malarzm done. And thanks for the tip.

alcaeus:
After discussion in IRC we decided not to merge this. Doctrine2 has renamed the [property](https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/UnitOfWork.php#L155) but left the [methods](https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/UnitOfWork.php#L2970-L2972) with their old names. In this case I'd opt for consistency first and move this back to the 2.0 milestone.

PR #1290, opened 2015-12-01
Align phpdoc

malukenho:


alcaeus:
If you look closely you will see that everything below this line is indented 5 characters instead of 4. Note sure we want to change this as it will make merging existing pull requests that change UoW below this line a real pain. Thoughts, @malarzm and @jmikola?

malukenho:
@alcaeus I guess old PR's can be rebased.

alcaeus:
I apologise, apparently my IDE tricked me into thinking indentation is wrong. You are correct, the comment line is the only one that is incorrect. :+1:

malukenho:
@alcaeus thanks :D

alcaeus:
Thanks!

PR #1296, opened 2015-12-03
Fix return type hint

MidnightDesign:


malarzm:
Thank you @MidnightDesign !

Issue #1294, opened 2015-12-03
QueryBuilder API: \MongoRegex instances on string typed identifier fields get discarded


narcoticfresh:
Hi guys

There is a proprietary behavior when a user issues an `equals()` on an identifier field using `QueryBuilder` as opposed to a 'normal' field ([see DocumentPersister](https://github.com/doctrine/mongodb-odm/blob/1.0.3/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php#L1004)).

If I do this:

```php
$builder->field("id")->equals(new \MongoRegex("/^ad.*$/i"));
```

Then check `$builder->getQueryArray()` I get:

```
array (size=1)
  '_id' => string '/^ad.*$/i' (length=9)
``` 

As you can see, the `MongoRegex` instance is lost and the query doesn't work as expected.

I don't know if this is a feature, a bug or wrong usage - so I first created a StackOverflow question detailing this issue:
http://stackoverflow.com/questions/34061566/doctrine-2-odm-querying-id-field-with-mongoregex-like

I'm creating this issue in the hopes to get some answer why doctrine-odm behaves that way as it does or to get an answer how one can do a Regex-based search on an identifier field..

IMHO Doctrine should not intervene in a feature that is possible using MongoDB shell. As all our IDs are strings and not `MongoID`s, I see no reason why my `MongoRegex` gets converted to a plain string. What am I missing?

Thanks so much!

malarzm:
I agree that ODM should let you do so :) Tagged it as a bug and hopefully I'll get back to it shortly

narcoticfresh:
@malarzm thanks very much; I'll update the StackOverflow question..

I'm not into active development here, is there anything I can help? Would it be a quickfix in `DocumentPersister` or is it correcting the return of `getDatabaseIdentifierValue()` and then correct that in each `Type`?

malarzm:
From top of my head I'd suggest the DocumentPersister and additional check.
If you'd like to work on the fix it'd be great :)

narcoticfresh:
@malarzm 
i started on creating the unit test for this bug.. and that yields interesting things..
apparently, this bug only applies to identifier fields of type `string` - not `id` or `custom_id`. 

so this bug apparently only appears if:

* the `UUID` strategy is used
* the type is forced to `string`, meaning that `StringType` is internally used instead of `IdType`/`CustomidType`.

### what works

let's consider this annotation class:

```php
/** @ODM\Document */
[...]

PR #1298, opened 2015-12-07
Preserve \MongoRegex on string typed identifier

narcoticfresh:
This is a possible fix for #1294 and a unit test to avoid regression.

I'm not sure if this should be handled in the `Type` class. It makes kinda sense as it's the culprit. But maybe it's a too much narrow check (topic wise) in the `Type`. But I'm not sure where it could be handled elsewhere as `convertToDatabaseValue()` will be called anyway and the `\MongoRegex` instance would be lost(?)

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Types/StringType.php
+++ b/lib/Doctrine/ODM/MongoDB/Types/StringType.php
@@ -30,7 +30,7 @@ class StringType extends Type
 {
     public function convertToDatabaseValue($value)
     {
-        return $value !== null ? (string) $value : null;
+        return ($value === null || $value instanceof \MongoRegex) ? $value : (string) $value;
     }
 
     public function convertToPHPValue($value)
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1294Test.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1294Test.php
@@ -1,0 +1,52 @@
+<?php
+
+namespace Doctrine\ODM\MongoDB\Tests\Functional\Ticket;
+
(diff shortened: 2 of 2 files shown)


lib/Doctrine/ODM/MongoDB/Types/StringType.php
```diff
@@ -30,7 +30,7 @@ class StringType extends Type
 {
     public function convertToDatabaseValue($value)
     {
-        return $value !== null ? (string) $value : null;
+        return $value !== null ? $value instanceof \MongoRegex ? $value : (string) $value : null;
```
malarzm: Can we put this new check in normal `if`? As much as I love ternary operator it'll more readable :)

narcoticfresh:
@malarzm agreed, it looked nasty.. i changed it a bit, looks way for simple now ;-)

lib/Doctrine/ODM/MongoDB/Types/StringType.php
```diff
@@ -30,7 +30,7 @@ class StringType extends Type
 {
     public function convertToDatabaseValue($value)
     {
-        return $value !== null ? (string) $value : null;
+        return (is_null($value) || $value instanceof \MongoRegex) ? $value : (string) $value;
```
malarzm: Nitpicking a bit maybe, but could we use `$value === null`? I don't think we use function checks anywhere

malarzm:
I've left one more comment and could you also please squash the commits?

narcoticfresh:
@malarzm ok, changed and squashed ;-)

alcaeus:
Manually merged in https://github.com/doctrine/mongodb-odm/commit/17be5e28f6e773c499e30c8fc2a9c33b57a49695. Thanks!

PR #1300, opened 2015-12-08
Do not duplicate properties and methods on children classes

### LINKED ISSUE #1299: Document generator bug with extended classes
Suppose you have these 2 classes:

```php
abstract class AbstractUser
{
    /** @ODM\Id */
    protected $id;
    
    protected $name;
    
    public function getId()
    {
        return $this->id;
    }
    
    public function getName()
    {
        return $this->name;
    }
    
    public function setName($name)
    {
        $this->name = $name;
    }
}
```

and:

```php
class User extends AbstractUser
{
    protected $lastname;
}
```

Then when generating documents, the fields, getters and setters from `AbstractUser` are duplicated to the `User` child class, which should not be the case.

A PR will be opened in the next couple of minutes.

michaelperrin:
Fixes #1299 .

This is inspired by the PR made on the ORM: https://github.com/doctrine/doctrine2/pull/1098

Please note that some extra whitespaces have been removed for code style as well. This can been done in an other PR if needed, but I think that's alright as there are not many.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Tools/DocumentGenerator.php
+++ b/lib/Doctrine/ODM/MongoDB/Tools/DocumentGenerator.php
@@ -438,20 +438,21 @@ public function <methodName>()
 
     private function hasProperty($property, ClassMetadataInfo $metadata)
     {
-        if ($this->extendsClass()) {
+        if ($this->extendsClass() || class_exists($metadata->name)) {
             // don't generate property if its already on the base class.
-            $reflClass = new \ReflectionClass($this->getClassToExtend());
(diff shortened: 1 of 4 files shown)


malarzm:
@michaelperrin thank you for the PR! I'll try to get back to it around weekend if that's ok

michaelperrin:
@malarzm Great! Thanks!

malarzm:
Merged manually in https://github.com/doctrine/mongodb-odm/commit/490ca30235f1bc2b1f85b15389a5ba43a47e7a05 and will be merged into `master` soon - thank you @michaelperrin !

PR #1303, opened 2015-12-11
Add support for partial indexes

alcaeus:
MongoDB 3.2 adds support for [partial indexes](https://docs.mongodb.org/manual/core/index-partial/). This PR adds support for adding partialIndexExpression options to the index specification.

Examples:
- YAML:
```
  indexes:
    index1:
      keys:
        username: desc
      options:
        unique: true
        dropDups: false
        partialFilterExpression:
          version: { $gt: 1 }
          discr: { $eq: 'default' }
```
- XML:
```
<indexes>
    <index unique="true">
        <key name="username" order="desc" />
        <option name="dropDups" value="false" />
        <partialFilterExpression>
            <field name="version" value="1" operator="gt" />
            <field name="discr" operator="eq" value="default" />
        </partialFilterExpression>
    </index>
</indexes>
```
- Annotations:
```
/**
 * @ODM\Indexes(
 *     @ODM\Index(keys={"username"="asc"},dropDups=true,partialFilterExpression={"counter"={"$gt"=5}})
 * )
 */
```

List of TODOs:
- [x] Figure out how to properly map $and in XML
- [x] Ensure SchemaManager::isMongoIndexEquivalentToDocumentIndex() works properly
- [x] You can always add more tests

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Mapping/Driver/AnnotationDriver.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/Driver/AnnotationDriver.php
@@ -235,6 +235,9 @@ class AnnotationDriver extends AbstractAnnotationDriver
                 $options[$name] = $index->$name;
             }
         }
+        if (! empty($index->partialFilterExpression)) {
+            $options['partialFilterExpression'] = $index->partialFilterExpression;
+        }
         $options = array_merge($options, $index->options);
         $class->addIndex($keys, $options);
     }
--- a/lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
@@ -373,22 +373,67 @@ class XmlDriver extends FileDriver
             }
         }
 
-        if (isset($xmlIndex->{'partialFilterExpression'})) {
-            foreach ($xmlIndex->{'partialFilterExpression'} as $partialFilterExpression) {
-                $operator = (string) $partialFilterExpression['comparison'] ?: 'eq';
-                $value = (string) $partialFilterExpression['value'];
-                if ($value === 'true') {
-                    $value = true;
-                } elseif ($value === 'false') {
-                    $value = false;
-                } elseif (is_numeric($value)) {
-                    $value = preg_match('/^[-]?\d+$/', $value) ? (integer) $value : (float) $value;
+        if (isset($xmlIndex->{'partial-filter-expression'})) {
+            $partialFilterExpressionMapping = $xmlIndex->{'partial-filter-expression'};
+
+            if (isset($partialFilterExpressionMapping->and)) {
+                foreach ($partialFilterExpressionMapping->and as $and) {
+                    if (! isset($and->field)) {
+                        continue;
+                    }
+
+                    $partialFilterExpression = $this->getPartialFilterExpression($and->field);
+                    if (! $partialFilterExpression) {
+                        continue;
+                    }
+
+                    $options['partialFilterExpression']['$and'][] = $partialFilterExpression;
                 }
-                $options['partialFilterExpression'][(string) $partialFilterExpression['field']] = array('$' . $operator => $value);
+            } elseif (isset($partialFilterExpressionMapping->field)) {
+                $partialFilterExpression = $this->getPartialFilterExpression($partialFilterExpressionMapping->field);
+
+                if ($partialFilterExpression) {
+                    $options['partialFilterExpression'] = $partialFilterExpression;
+                }
             }
         }
 
         $class->addIndex($keys, $options);
+    }
+
+    private function getPartialFilterExpression(\SimpleXMLElement $fields)
+    {
+        $partialFilterExpression = [];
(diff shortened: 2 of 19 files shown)


alcaeus:
Long time no see, I've updated the XML mapping. The example in the comment above outlines the straight-forward mapping. For more complicated mappings, there are additional possiblilities. You can either check properties of embedded documents (checking only those properties mentioned):
```
<partialFilterExpression>
    <field name="nested">
        <field name="count" operator="gt" value="0" />
        <field name="test" value="foo" />
    </field>
</partialFilterExpression>
```

Or check a document for equality (ensuring that a document looks exactly like what's provided):
```
<partialFilterExpression>
    <field name="object" operator="eq">
        <field name="property" value="foo" />
        <field name="count" value="bleh" />
    </field>
</partialFilterExpression>
```

With this it's also possible to use the `$and` operator. MongoDB supports this operator only at the root level:
```
<partialFilterExpression>
    <and>
        <field name="nested">
            <field name="count" operator="gt" value="0" />
            <field name="test" value="foo" />
        </field>
    </and>
    <and>
        <field name="object" operator="eq">
            <field name="property" value="foo" />
            <field name="count" value="bleh" />
        </field>
    </and>
</partialFilterExpression>
```
When using `and`, you cannot add any `field` mappings at the root level. You also can't nest the `and` operator in a `field` mapping.

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AnnotationDriver.php
```diff
@@ -237,6 +237,9 @@ private function addIndex(ClassMetadataInfo $class, $index, array $keys = array(
                 $options[$name] = $index->$name;
             }
         }
+        if (! empty($index->partialFilterExpression)) {
```
malarzm: Missing space before `!` ;)

lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
```diff
@@ -375,9 +375,69 @@ private function addIndex(ClassMetadataInfo $class, \SimpleXmlElement $xmlIndex)
             }
         }
 
+        if (isset($xmlIndex->{'partialFilterExpression'})) {
+            $partialFilterExpressionMapping = $xmlIndex->{'partialFilterExpression'};
+
+            if (isset($partialFilterExpressionMapping->and)) {
+                foreach ($partialFilterExpressionMapping->and as $and) {
+                    if (! isset($and->field)) {
```
malarzm: Here and in next `if` as well

lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
```diff
@@ -375,9 +375,69 @@ private function addIndex(ClassMetadataInfo $class, \SimpleXmlElement $xmlIndex)
             }
         }
 
+        if (isset($xmlIndex->{'partialFilterExpression'})) {
+            $partialFilterExpressionMapping = $xmlIndex->{'partialFilterExpression'};
+
+            if (isset($partialFilterExpressionMapping->and)) {
```
malarzm: Why some paths are enclosed in `'` while for instance this one isn't?

lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
malarzm: And more ahead too, will stop

lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
```diff
@@ -375,9 +375,69 @@ private function addIndex(ClassMetadataInfo $class, \SimpleXmlElement $xmlIndex)
...
+                }
+
+                $nestedExpression = $this->getPartialFilterExpression($field->field);
+                if (! $nestedExpression) {
+                    continue;
+                }
+
+                $value = $nestedExpression;
+            } else {
+                $value = (string) $field['value'];
```
malarzm: Shall we `trim` $value just in case?

malarzm:
I think we should have tests for expression with `and` since XML driver has a check on that?

lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
alcaeus: Technically it's only necessary when having elements that are invalid php names, e.g. `discriminator-map`. Through copy/paste it turns out that most of them are written like that even when they don't need to. I'd say we change this to only use `->{'tag-name'}` when it would constitute an invalid identifier.

alcaeus:
@malarzm Good idea, I'll add a few extra documents to be safe.

alcaeus:
@malarzm Added tests for various filter expressions being loaded via XML and YAML to ensure they are the same. Invalid filter expressions will only be fixed by the XML driver as YAML and annotations are schemaless and will just let the database throw a fit if the expression is invalid.

malarzm:
:+1:

PR #1309, opened 2015-12-16
Add phpunit to dev dependencies

malukenho:


malarzm:
Travis seems to be unhappy :)

malukenho:
@malarzm done.

malarzm:
@malukenho could you please update `composer.lock` instead of removing it? :)

malukenho:
@malarzm only to I know. Why we have to storage `composer.lock` on repository?

malarzm:
See this commit https://github.com/doctrine/mongodb-odm/commit/c1c55facdbe7e85e4a2bef5c215c5fb6ef4b9b93

malukenho:
@malarzm doesn't make so much sense for a library. For an application makes sense.

IMHO library should work with updated dependencies, otherwise maintainers have to keep an eyer on security issues on this old locked version.

> Note: For libraries it is not necessary to commit the lock file, see also: Libraries - Lock file.

composer.json
```diff
@@ -26,6 +26,7 @@
         "doctrine/mongodb": "~1.2"
     },
     "require-dev": {
+        "phpunit/phpunit": "~4.8",
```
Ocramius: `"~4.8|~5.0"`

Ocramius:
Indeed: the `composer.lock` shouldn't be in here :-\

malarzm:
I'll defer this to @jmikola then :P

jmikola:
If ORM has removed it, I won't object. The other changes seem to be in line with https://github.com/doctrine/doctrine2/commit/a91e05db9ea381de6b63a13df10f4f60fd88f86d and https://github.com/doctrine/doctrine2/commit/2528d0b3fc103194a00a4ef3efd2bde11f5921b0 (namely, using the PHPUnit dev dependency and self-updating Composer).

Ocramius:
@malukenho can you also gitignore it again?

malukenho:
@jmikola done.

alcaeus:
Merged manually in https://github.com/doctrine/mongodb-odm/commit/55bff8ba7ddf3040558a4c3e3eaa84637882f15c. I took the liberty of squashing a couple of commits. Thanks!

PR #1311, opened 2015-12-16
Fix parameter checking in BsonFilter

kovagoz:
Replace `isset` to `array_key_exists` because using the former throws exception if the parameter's value is null.


malarzm:
LGTM :+1:

malarzm:
Merged manually into `1.0.x` branch in https://github.com/doctrine/mongodb-odm/commit/1a2c2efcd8d03b668237f1a26868751a5f43b76e - thank you @kovagoz!

PR #1313, opened 2015-12-16
Fix querying for empty simple ref many

### LINKED ISSUE #621: Is it possible to remove documents by references field with empty object?
I have document with $groupActions reference many field.

```PHP
/**
 * @Mongo\ReferenceMany(targetDocument="Action", simple=true)
 */
private $groupActions;
```

I want to remove documents where { groupActions: {} }

I tried something like this:

```PHP
$qb->remove()->field('groupActions')->equals(new \stdClass());
$query = $qb->getQuery();
$result = $query->execute();
```

But when I looked into $query->query array I found something interesting:

$query->query

I suppose this is a bug?

malarzm:
Fixes #621

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -973,7 +973,7 @@ class DocumentPersister
             }
 
             // No further preparation unless we're dealing with a simple reference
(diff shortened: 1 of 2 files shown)


alcaeus:
Merged manually in https://github.com/doctrine/mongodb-odm/commit/8876b91f036accbe02161fa13fc41ee638f6bd7d.

Issue #1233, opened 2015-09-25
[doc] Full text search sample missing


acrozes:
I don't find any full text search sample in the mongodb ODM documentation ...

PR #1317, opened 2015-12-18
Add missing documentation

### LINKED ISSUE #180: No lazy loading of references which has the inversedBy and mappedBy option
Currently all references which was configured with the inversedBy and mappedBy option is not lazy loading. The generated hydrator class loads all references on hydration.

alcaeus:
Closes #180, #1233.

This adds documentation for text queries and adds a note about inverse `@ReferenceOne` annotations not being lazy-loaded using proxies.

### DIFF
--- a/docs/en/reference/bidirectional-references.rst
+++ b/docs/en/reference/bidirectional-references.rst
@@ -169,6 +169,14 @@ you must null it out on the ``Cart.customer`` side:
     $cart->setCustomer(null);
     $dm->flush();
 
+.. note::
+
+    When specifying inverse one-to-one relationships the referenced document is
+    loaded directly when the owning document is hydrated instead of using a
+    proxy. In the example above, loading a ``Customer`` object from the database
+    would also cause the corresponding ``Cart`` to be loaded. This can cause
+    performance issues when loading many ``Customer`` objects at once.
+
 Self-Referencing Many to Many
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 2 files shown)


docs/en/reference/bidirectional-references.rst
```diff
@@ -169,6 +169,14 @@ you must null it out on the ``Cart.customer`` side:
     $cart->setCustomer(null);
     $dm->flush();
 
+.. note::
+
+    While the inverse side of a one-to-many relationship is loaded lazily, all
```
jmikola: If I understand this and #180 correctly, this isn't an issue with one-to-many. Rather, we must eager load references on the inverse-side because we cannot reliably create proxy objects for them. Should this be rewritten to clarify that we can only lazily load (i.e. use proxy objects) for owning-side references, and that inverse-side references always require eager loading?

Also, is this only a case with reference-one relationships, since inverse-side reference-many relationships can use collections?

alcaeus: Exactly. one-to-many references are always using collections, regardless of whether they are owning or inverse. That means that they will only be initialized once the collection is used for the first time, not directly on hydration. Inverse one-to-one relationships can't take advantage of this because there is no ID (which would allow the use of proxies) and collections can't be used because they are not designed for single objects.

alcaeus: I could clarify that the collection is initialized when it is first accessed but all documents contained therein are loaded and hydrated right then as well, not as proxy objects.

malarzm:
Merged manually in https://github.com/doctrine/mongodb-odm/commit/a124bfa2cf443b88449f3df69a7f1aca394455e9

PR #1318, opened 2015-12-19
Deprecate field type annotation classes in favor of Field

### LINKED ISSUE #1207: Deprecate field type annotation classes in favor of Field
This is related to #1168 (deprecating annotations that share names with reserved words in PHP7). It is more directly related to doctrine/phpcr-odm#641 and the corresponding PR (doctrine/phpcr-odm#643).

We should encourage users to use `@Field(type="...")` instead of the plethora of named annotation classes. Deprecating this in 1.x will allow us to remove the named classes entirely in 2.0.

alcaeus:
Closes #1207.

Starting with PHP 7, some of the type annotation ODM uses are reserved keywords:
* int
* string
* bool
* float

To avoid having large inconsistencies we decided to deprecate all type annotations (e.g. those extending `AbstractField` and only overwriting the `type` property) to be removed in ODM 2.0. This affects the following annotation classes:
* Bin
* BinCustom
* BinFunc
* BinMD5
* BinUUID
* BinUUIDRFC4122
* Bool
* Boolean
* Collection
* Date
* Float
* Hash
* Increment
* Int
* Integer
* Key
* ObjectId
* Raw
* String
* Timestamp

Instead, users should switch to using `@Field(type="...")` to map fields.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Mapping/Annotations/Bin.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/Annotations/Bin.php
@@ -19,7 +19,10 @@
 
 namespace Doctrine\ODM\MongoDB\Mapping\Annotations;
 
-/** @Annotation */
+/**
+ * @Annotation
+ * @deprecated This class will be removed in ODM 2.0
+ */
 final class Bin extends AbstractField
 {
     public $type = 'bin';
--- a/lib/Doctrine/ODM/MongoDB/Mapping/Annotations/Document.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/Annotations/Document.php
@@ -19,7 +19,11 @@
 
 namespace Doctrine\ODM\MongoDB\Mapping\Annotations;
 
-/** @Annotation */
+/**
+ * Identifies a class as a document that can be stored in the database
+ *
+ * @Annotation
+ */
 final class Document extends AbstractDocument
 {
     public $db;
--- a/lib/Doctrine/ODM/MongoDB/Mapping/Annotations/EmbedMany.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/Annotations/EmbedMany.php
@@ -21,7 +21,11 @@ namespace Doctrine\ODM\MongoDB\Mapping\Annotations;
 
 use Doctrine\ODM\MongoDB\Utility\CollectionHelper;
 
-/** @Annotation */
+/**
+ * Embeds multiple documents
+ *
+ * @Annotation
+ */
 final class EmbedMany extends AbstractField
 {
     public $type = 'many';
--- a/lib/Doctrine/ODM/MongoDB/Mapping/Annotations/EmbedOne.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/Annotations/EmbedOne.php
@@ -19,7 +19,11 @@
 
 namespace Doctrine\ODM\MongoDB\Mapping\Annotations;
 
-/** @Annotation */
+/**
+ * Embeds a single document
+ *
+ * @Annotation
+ */
 final class EmbedOne extends AbstractField
 {
     public $type = 'one';
--- a/lib/Doctrine/ODM/MongoDB/Mapping/Annotations/EmbeddedDocument.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/Annotations/EmbeddedDocument.php
@@ -19,7 +19,11 @@
 
 namespace Doctrine\ODM\MongoDB\Mapping\Annotations;
 
-/** @Annotation */
+/**
+ * Identifies a class as a document that can be embedded but not stored by itself
+ *
+ * @Annotation
+ */
 final class EmbeddedDocument extends AbstractDocument
 {
     public $indexes = array();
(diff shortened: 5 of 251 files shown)


malarzm:
I think [this](https://github.com/doctrine/mongodb-odm/pull/1318/files#diff-1457aa7b8fa8bab45fea926215c23cbaR110) among the others from bigger list should have different notes? 

Beside that big :+1: for going through everything!

alcaeus:
Yes, we could expand the wording, should we also add an example of how the field show be mapped?

malarzm:
> should we also add an example of how the field show be mapped?

Was thinking about it and frankly I'm not sure... existing examples are updated for the new way and we could use that until absolute removal which will need documentation changes anyway so we could postpone writing new chapter? Unless you feel like doing that :D

alcaeus:
I'm not sure we need this. I think for 1.1 just deprecating the classes is a good first step. We can then always throw deprecation notices in `AnnotationDriver` when a deprecated annotation is used.

malarzm:
:+1:

malarzm:
Another :+1: after rewording deprecation notes :)

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/Bin.php
```diff
@@ -19,7 +19,10 @@
 
 namespace Doctrine\ODM\MongoDB\Mapping\Annotations;
 
-/** @Annotation */
+/**
+ * @Annotation
```
jmikola: Is `@Annotation` a convention from Doctrine ORM or Symfony? I'm curious if we should just replace this with a one-line class description (e.g. "Annotation for binary data field.")

jmikola:
One question above, but LGTM otherwise. Thanks for the tedious search/replace job!

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/Bin.php
alcaeus: IIRC, the `@Annotation` annotation tells the annotation reader that the class in question can be used as an annotation.
I'm actually :+1: to documenting annotation classes and their properties while we're at it.

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/Document.php
```diff
@@ -19,7 +19,11 @@
 
 namespace Doctrine\ODM\MongoDB\Mapping\Annotations;
 
-/** @Annotation */
+/**
+ * Identifies a class as document that can be stored in the database
```
jmikola: "as a document"

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/EmbeddedDocument.php
```diff
@@ -19,7 +19,11 @@
 
 namespace Doctrine\ODM\MongoDB\Mapping\Annotations;
 
-/** @Annotation */
+/**
+ * Identified a class as document that can be embedded but not stored by itself
```
jmikola: "Identifies"

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/MappedSuperclass.php
```diff
@@ -19,7 +19,11 @@
 
 namespace Doctrine\ODM\MongoDB\Mapping\Annotations;
 
-/** @Annotation */
+/**
+ * Specify class as parent class for documents containing mapping information
```
jmikola: "Specifieds a parent class that other documents may extend to inherit mapping information"

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/EmbedMany.php
```diff
@@ -21,7 +21,11 @@
 
 use Doctrine\ODM\MongoDB\Utility\CollectionHelper;
 
-/** @Annotation */
+/**
+ * Specifies a one-to-many relationship to a different document
```
jmikola: I think this is mixed up with ReferenceMany's doc block.

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/EmbedOne.php
```diff
@@ -19,7 +19,11 @@
 
 namespace Doctrine\ODM\MongoDB\Mapping\Annotations;
 
-/** @Annotation */
+/**
+ * Specifies a one-to-one relationship to a different document
```
jmikola: This also appears to be mixed up with ReferenceOne's doc block.

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/ReferenceOne.php
```diff
@@ -19,7 +19,11 @@
 
 namespace Doctrine\ODM\MongoDB\Mapping\Annotations;
 
-/** @Annotation */
+/**
+ * Embeds a single document
```
jmikola: Is this mixed up with EmbedOne's doc block?

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/ReferenceMany.php
```diff
@@ -21,7 +21,11 @@
 
 use Doctrine\ODM\MongoDB\Utility\CollectionHelper;
 
-/** @Annotation */
+/**
+ * Embeds multiple documents
```
jmikola: Is this mixed up with EmbedMany's doc block?

PR #1320, opened 2015-12-20
Let ReflectionException bubble

### LINKED ISSUE #685: [class metadata] The property mapping is skipped if fieldName is wrong
If the fieldName defined (in xml mapping for example) is not matched any property it is just silently skipped [here](https://github.com/doctrine/mongodb-odm/blob/master/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php#L82).

It is not like ORM work. It would throw an exception and force me to correct mapping:

```
ReflectionException: Property Payum\Paypal\ExpressCheckout\Nvp\Bridge\Doctrine\Entity\PaymentDetails::$tokenNotExist does not exist
```

If it is a bug I could try to fix.

malarzm:
Closes #685

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
@@ -79,11 +79,9 @@ class ClassMetadata extends ClassMetadataInfo
     {
         $mapping = parent::mapField($mapping);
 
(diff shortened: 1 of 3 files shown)


alcaeus:
Merged manually in https://github.com/doctrine/mongodb-odm/commit/58e001e5cea4b7b8fa4ec9c94c9d3e1dac686a3e.

PR #1321, opened 2015-12-22
Replace hardcoded FQCN with use of `::class` magic constant

malukenho:


lib/Doctrine/ODM/MongoDB/Mapping/Driver/AnnotationDriver.php
```diff
@@ -26,6 +26,11 @@
 use Doctrine\Common\Persistence\Mapping\Driver\AnnotationDriver as AbstractAnnotationDriver;
 use Doctrine\ODM\MongoDB\Events;
 use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
+use Doctrine\ODM\MongoDB\Mapping\Annotations\Document;
+use Doctrine\ODM\MongoDB\Mapping\Annotations\EmbeddedDocument;
+use Doctrine\ODM\MongoDB\Mapping\Annotations\HasLifecycleCallbacks;
+use Doctrine\ODM\MongoDB\Mapping\Annotations\MappedSuperclass;
```
jmikola: We still have cases where `ODM\AnnotationClassName` is used within this file (after your changes), so are these four specific `use` statements necessary?

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AnnotationDriver.php
```diff
@@ -35,13 +40,14 @@
  * @since       1.0
  * @author      Jonathan H. Wage <jonwage@gmail.com>
  * @author      Roman Borschel <roman@code-factory.org>
+ * @author      Jefersson Nathan <malukenho@phpse.net>
```
jmikola: While refactoring patches are much appreciated, I don't think it warrants adding `@author` tags. I know that I have a handful in my own name from previous classes, but they are mostly clutter since we have git history.

@malarzm, @jwage: I'd be in favor of removing them across the board if there are no objections.

malukenho: @jmikola sorry. Removed, It's not a problem :+1:

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AnnotationDriver.php
malukenho: @jmikola good catch! Done.

malarzm:
Thank you @malukenho!

jmikola:
Likewise, thanks for the PR!

PR #1322, opened 2015-12-22
Remove references to ORM

malukenho:


malarzm:
Thank you @malukenho I wanted to do this for quite some time but kept forgetting :D

PR #1324, opened 2015-12-23
Removed composer.lock from .gitignore

Bilge:
It is a well established best practice to commit the `composer.lock` file to the repository.


alcaeus:
This was done on purpose in #1309 since it's not necessary for libraries. See also the [composer documentation](https://getcomposer.org/doc/02-libraries.md#lock-file).

PR #1325, opened 2015-12-23
Use short array syntax

malukenho:


alcaeus:
Thanks for the PR! For now we've scheduled this for the 1.x milestone as we first want to get people to use short array syntax in their PRs before changing all current occurences and have to pay close attention when merging PRs or merging down bugfixes from the 1.0 branch. Let's give people some time to get used to this change.

malukenho:
@alcaeus we can also put CS checks on travis build

alcaeus:
That would be a good idea, regardless of when we merge this!

alcaeus:
Given the number of conflicts, I'll close this pull request. @malukenho thanks for the initial work, but I believe this is more suitable for just before a release of 2.0: the changes there will most likely make many pull requests conflicted anyways, so there's no harm doing it then.

PR #1331, opened 2015-12-31
Moved lifecycle callbacks/events logic

malarzm:
Possibly first PR in a serie of making UnitOfWork a smaller and friendlier place. Frankly I was expecting to make performance a little bit worse but seen a gain instead (however running tests in vagrant may not be realiable perf-wise) - for sure it's not worse :)


malarzm:
Wow locally I have 

```
vagrant@precise64:/var/www$ php -v
PHP 5.6.10-1+deb.sury.org~precise+1 (cli)
```

and everything passed o.O Will fix this after new year ;)

alcaeus:
Nice, looks good! :+1:

malarzm:
It's still on top of latest changes so I'll just merge it with GH :)

PR #1333, opened 2016-01-07
Expr::references should go through discriminator map

malarzm:
Before it was throwing an exception that field was not found. Test case is written like this only for the sake of testing, normally we use same type of references everywhere, only fields are named differently in child classes for the sake of making sense :)

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Query/Expr.php
+++ b/lib/Doctrine/ODM/MongoDB/Query/Expr.php
@@ -75,7 +75,7 @@ class Expr extends \Doctrine\MongoDB\Query\Expr
                     }
                 }
                 if ( ! isset($mapping) || $mapping === null) {
-                    throw $e;
+                    throw MappingException::mappingNotFoundInClassNorDescendants($this->class->name, $this->currentField);
                 }
             }
             $dbRef = $this->dm->createDBRef($document, $mapping);
--- a/lib/Doctrine/ODM/MongoDB/Query/Expr.php
+++ b/lib/Doctrine/ODM/MongoDB/Query/Expr.php
@@ -21,6 +21,7 @@ namespace Doctrine\ODM\MongoDB\Query;
 
 use Doctrine\ODM\MongoDB\DocumentManager;
 use Doctrine\ODM\MongoDB\Mapping\ClassMetadata;
+use Doctrine\ODM\MongoDB\Mapping\MappingException;
 
 /**
  * Query expression builder for ODM.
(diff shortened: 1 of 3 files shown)


lib/Doctrine/ODM/MongoDB/Query/Expr.php
```diff
@@ -60,7 +61,23 @@ public function setClassMetadata(ClassMetadata $class)
...
+                }
+                foreach ($this->class->discriminatorMap as $child) {
+                    $childClass = $this->dm->getClassMetadata($child);
+                    if ($childClass->hasAssociation($this->currentField)) {
+                        $mapping = $childClass->getFieldMapping($this->currentField);
+                        break;
+                    }
+                }
+                if ( ! isset($mapping) || $mapping === null) {
+                    throw $e;
```
alcaeus: This would throw an exception with the message "No mapping found for field '$fieldName' in class '$className'." - should we change that to mention that we've tried child classes as well to let people know about that?

alcaeus:
Looks good! :+1:

lib/Doctrine/ODM/MongoDB/Query/Expr.php
malarzm: "No mapping found for field '$fieldName' in class '$className' nor its descendants." ?

alcaeus: Sounds good!

malarzm: Oook, now I'll need to add more tests :P

malarzm:
Exception added and rebased atop new `master` for quick merge :)

lib/Doctrine/ODM/MongoDB/Query/Expr.php
```diff
@@ -60,7 +61,23 @@ public function setClassMetadata(ClassMetadata $class)
...
+                $mapping = $this->class->getFieldMapping($this->currentField);
+            } catch (MappingException $e) {
+                if (empty($this->class->discriminatorMap)) {
+                    throw $e;
+                }
+                foreach ($this->class->discriminatorMap as $child) {
+                    $childClass = $this->dm->getClassMetadata($child);
+                    if ($childClass->hasAssociation($this->currentField)) {
+                        $mapping = $childClass->getFieldMapping($this->currentField);
+                        break;
```
jmikola: Rather than give up after finding the first mapping, I think we should check all discriminator possibilities and throw if we find two conflicting mappings. It should be sufficient to compare the mapping arrays with `===`.

malarzm:
I just recalled that `includesReferenceTo` is a thing, will update that method as well

alcaeus:
:+1: on the changes to catch conflicting field mappings as well!

jmikola:
LGTM!

PR #1339, opened 2016-01-15
Include Embed/RefMany in changeset

malarzm:
As of this PR the collection will be included in changeset if at least one of following is true

1. at least one element was added
2. at least one element was removed
3. collection was cleared

Collection won't be in changeset if only one of its element was modified.

Thoughts? :)


alcaeus:
Nice, didn't think the changes would be so small. :+1:

PR #1342, opened 2016-01-17
Allow PHP 7 in composer.json

alcaeus:
While ODM does not officially support PHP 7 (due to the legacy MongoDB driver not being available), there are alternatives the people may use. This PR aims to allow those developers to use ODM with a polyfill for ext-mongo (e.g. mongofill, mongo-php-adapter) in PHP 7.


malarzm:
:+1:

Issue #1346, opened 2016-01-21
Undefined property $distance


zajca:
Hi, I run to this error and I'm bit confused.

```
Notice: Undefined property: MongoDBODMProxies\__CG__\MyBundle\Document\Place::$distance
```

My use case is this:

I have place document which have coordinates and 2d index (following documentation)

```
/**
 * @ODM\Document
 * @ODM\Index(keys={"coordinates"="2d"})
 */
class Place extends AbstractDocument
{

    /**
     * @ODM\Distance
     */
    public $distance;

    /**
     * @Type("Coordinates")
     * @Assert\Valid
     * @ODM\EmbedOne(targetDocument="Coordinates")
     */
    protected $coordinates;
...
```

and user document which can have multiple favourites places:

```
/**
 * @ODM\Document 
 */
class User extends AbstractUser
{
    /**
     * @var ArrayCollection<Place> $favourites
     * @ODM\ReferenceMany(targetDocument="Place")
     */
    protected $favourites;

    public function addFavourite($place)
    {
        $this->favourites->add($place);

        return $this;
    }

    public function hasFavourite($place)
    {
        return $this->favourites->contains($place);
    }
...
}
```

weird thing happens when I try to add more favourites (this newer happen when adding first)

```
       if ($user->hasFavourite($place)) {
              return ...;
        }
        $user->addFavourite($place);
        $dm->persist($user);
        $dm->flush();
...
```

In ```MongoDBODMProxies\__CG__\MyBundle\Document\Place``` class I notice in constructor:
```
unset($this->distance)
```
which is cause of error. If I set property to protected error disappear.

I'm not sure if this is a bug or I don't understand what doctrine is doing.

//I'm using symfony3 and mongodb-odm master

alcaeus:
As you've properly noticed, you need to make the property `protected` or `private` for this to work, as is noted [in our documentation](http://docs.doctrine-project.org/projects/doctrine-mongodb-odm/en/latest/reference/architecture.html).

malarzm:
@zajca as @alcaeus noted it is way better to use `protected` or `private` visibility and it's your quickest way to fix the problem. I'd discourage using public persistent properties but this should work in fact and warrants an investigation.

[...]

PR #1347, opened 2016-01-21
add test for issue #1346 public property undefined

zajca:
test should fail with:
```
Doctrine\ODM\MongoDB\Tests\Functional\Ticket\GH1346Test::testPublicDistanceProperty
Undefined property: Proxies\__CG__\Doctrine\ODM\MongoDB\Tests\Functional\Ticket\GH1346ReferencedDocument::$distance

/<path>/mongodb-odm/tests/Proxies/__CG__DoctrineODMMongoDBTestsFunctionalTicketGH1346ReferencedDocument.php:64
/<path>/mongodb-odm/lib/Doctrine/ODM/MongoDB/UnitOfWork.php:614
/<path>/mongodb-odm/lib/Doctrine/ODM/MongoDB/UnitOfWork.php:687
/<path>/mongodb-odm/lib/Doctrine/ODM/MongoDB/UnitOfWork.php:674
/<path>/mongodb-odm/lib/Doctrine/ODM/MongoDB/UnitOfWork.php:903
/<path>/mongodb-odm/lib/Doctrine/ODM/MongoDB/UnitOfWork.php:388
/<path>/mongodb-odm/lib/Doctrine/ODM/MongoDB/DocumentManager.php:526
/<path>/mongodb-odm/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1346Test.php:45

FAILURES!
Tests: 1, Assertions: 0, Errors: 1.
```

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1346Test.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1346Test.php
@@ -10,27 +10,16 @@ class GH1346Test extends \Doctrine\ODM\MongoDB\Tests\BaseTest
     /**
      * @group GH1346Test
      */
-    public function testPublicDistanceProperty()
+    public function testPublicProperty()
     {
-        $coordinates = new GH1346Coordinates();
-        $coordinates->setLatitude(10);
-        $coordinates->setLongitude(10);
 
         $refrenced1 = new GH1346ReferencedDocument();
-        $refrenced1->setCoordinates($coordinates);
-
         $refrenced2 = new GH1346ReferencedDocument();
-        $refrenced2->setCoordinates($coordinates);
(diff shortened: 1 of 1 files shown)


malarzm:
Oh right, now that we have a failing test case could you make it as small as possible to still fail? It really helps after some time if it'd start failing again. For instance there's no need to remove documents explicitly as `BaseTest` drops collections anyway or having `Index` probably doesn't bring any value to the test itself

zajca:
Test should be now simple as possible. I didn't success to replicate this with less calls.

malarzm:
Awesome this will be enough :) I personally won't be able to investigate this before weekend but if you or anyone else would like to take a stab and need help please feel free to ping me either here or on IRC

malarzm:
#1355 included provided test and has been merged

PR #1348, opened 2016-01-22
Feature proposal: DocumentManager::export() and DocumentManager::deepCompare()

blockjon:
This PR introduces two new concepts that we find essential and that we have implemented in our fork:

export()
---
After we save a model to MongoDB, we like to put a copy of it into other databases, caches and queues. By calling DocumentManager::export(), a plain associative array is returned.

deepCompare()
---
When we load an item from an "alternate" source such as a database or a cache, we like to sometimes do a comparison against what is in the database to determine if the the copy of the data was accurate. The "deepCompare()" function makes this possible.

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1348Test.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1348Test.php
@@ -1,11 +1,11 @@
 <?php
 
-namespace Doctrine\ODM\MongoDB\Tests\Functional\Ticket\GHXXXX;
+namespace Doctrine\ODM\MongoDB\Tests\Functional\Ticket\GH1348;
 
 use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
 use Doctrine\Common\Collections\ArrayCollection;
 
-class GHXXXXTest extends \Doctrine\ODM\MongoDB\Tests\BaseTest
+class GH1348Test extends \Doctrine\ODM\MongoDB\Tests\BaseTest
 {
     public function testCanSnapshotSimpleModel()
     {
@@ -43,7 +43,7 @@ class GHXXXXTest extends \Doctrine\ODM\MongoDB\Tests\BaseTest
                         'numPages' => 88,
                         'pages' => array(),
                     ),
-                    '__class' => 'Doctrine\ODM\MongoDB\Tests\Functional\Ticket\GHXXXX\DrivingManual'
+                    '__class' => 'Doctrine\ODM\MongoDB\Tests\Functional\Ticket\GH1348\DrivingManual'
                 )
             )
         );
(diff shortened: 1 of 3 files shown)


tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1348Test.php
```diff
@@ -0,0 +1,498 @@
...
+        $this->dm->clear();
+
+        // Get the model from the database.
+        $retrievedCarDocument = $this->dm->getRepository(__NAMESPACE__ . '\Car')->findOneBy(array('id' => 'a'));
+
+        // Now lets build an exact replica of that model from scratch.
+        $exactlyTheSameCar = $this->buildComplexCarModel();
+        $exactlyTheSameCar->setId('a');
+
+        $this->assertTrue($this->dm->deepCompare($retrievedCarDocument, $exactlyTheSameCar));
```
blockjon: In my code we're doing something like this... we are comparing a model which was loaded from another database to a model which was loaded by Doctrine ODM. We want to be able to verify that the models are equal based on their odm field values. (This is not necessarily easy since the objects have different spl object hashes and sometimes there are private properties added to model instances which were loaded from MongoDB)

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1348Test.php
```diff
@@ -0,0 +1,498 @@
...
+        $book->setAuthor("Jonathan Block");
+        $exported = $this->dm->export($book);
+        $this->assertTrue(is_array($exported));
+        $this->assertTrue(count($exported) == 3);
+        $this->assertTrue(array_key_exists("id", $exported));
+        $this->assertTrue(array_key_exists("title", $exported));
+        $this->assertTrue(array_key_exists("author", $exported));
+        $this->assertNull($exported["id"]);
+        $this->assertEquals("Awesome PHP", $exported["title"]);
+        $this->assertEquals("Jonathan Block", $exported["author"]);
```
blockjon: This unit test demonstrates how we can extract a simple associative array from a Doctrine model instance.

malarzm:
@blockjon nice! I had something similar on my list which at least partially resembles what you've done here so I hope we'll come to an agreement :) 

My idea was to extract `ChangeSetCalculator` from `UnitOfWork` to a separate class and let it have 2 methods:

1. calculate($document) (this probably will have UnitOfWork as well)
2. getActualData($document)

and my number 2 seems to be your `export()`. I wanted to do so to make our `UnitOfWork` a friendlier place, eventually allow rolling own calculator which `getActualData` would work in more efficient way than reflections and apparently could fit your use case here as well. Reason I'm leaning towards having this utility in UoW instead of in `DocumentManager` is not cluttering public API and not adding methods that are not in an interface there.

Thoughts? :)

blockjon:
@malarzm It has come to my attention that we would also *really* want an option to not only export() to an array, but to also subsequently call some kind of import() which would re-build our Models into that exact same state. Can you please comment on this?

blockjon:
As for the location the DocumentManager is the public place where callers can invoke and things are fairly easy to understand. UnitOfWork is the black magic of the ODM and I've tried to avoid that where possible because things get very confusing very quickly in there.

malarzm:
> subsequently call some kind of import() which would re-build our Models into that exact same state.

This is exactly what Hydrators are for :) You can get to them, again, from UnitOfWork - [this method](https://github.com/doctrine/mongodb-odm/blob/957ab471ef4b60669dc6d0e283411e33deec8c0a/lib/Doctrine/ODM/MongoDB/UnitOfWork.php#L2635) is getting or creating documents (which you could utilize if document to be imported is not managed) or in case it would be [this else block](https://github.com/doctrine/mongodb-odm/blob/957ab471ef4b60669dc6d0e283411e33deec8c0a/lib/Doctrine/ODM/MongoDB/UnitOfWork.php#L2677-L2685) will show you how to create document using Hydrators - later you could probably `merge` newly created document.

> UnitOfWork is the black magic of the ODM

Sadly this is true hence my intention of breaking it up a little :)

PR #1349, opened 2016-01-22
storeAs option for references

coudenysj:
As discusses in #807, this pull requests starts using a `storeAs` option to support different storage formats:

* `simple`: $id
* `partial`: $id, $ref
* `full`: $id, $ref, $db

With `full` being the default.


lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataInfo.php
```diff
@@ -99,6 +99,16 @@ class ClassMetadataInfo implements \Doctrine\Common\Persistence\Mapping\ClassMet
     const MANY = 'many';
     const ONE = 'one';
 
+    /**#@+
+     * The types of storeAs references
+     *
+     * TODO move this to a different class?
```
malarzm: I think it's a good place :+1:

alcaeus:
Thanks for getting started with this, this should clean up reference mapping a bit!

I know I'm the one to come up with "partial" and "full", but I've "seen some things" that make me want to change the naming to include "dbref" in the name, especially since that's the name of the structure we're actually storing. Likewise, we could rename the "simple" option to "id" as a reference of that type would only store the ID of the reference object, not a DBRef object.

I'm really looking forward to this :+1:

coudenysj:
I'm not sure if we should include the dbRef name, because the "simple" option would not use the the DBRef structure.

The 'simple' vs 'id' argument I can follow.

alcaeus:
@coudenysj the "simple" option does not store it as DBRef and thus doesn't have "dbref" in its name. If we store it as "dbref" we might as well rename the storage option to include "dbref" in the name ;)

coudenysj:
@alcaeus Are you talking about the "storeAs" name? Or the different values it can have?

alcaeus:
Sorry, I was talking about the values it can have. IMO, "dbref" is more descriptive than "full" - same goes for partial as it's actually a partial dbref object (still dislike the wording "partial" in there, don't have an alternative though)

malarzm:
Gotta love naming things

coudenysj:
So you want to make the values really descriptive.

Some more brain dumps:

* **id**: `$id`
* **dbRef**: `$id`, `$ref` (as the `$db` option is optional:  http://php.net/manual/en/mongodbref.create.php)
* **dbRef_X?_**: `$id`, `$ref`, `$db`

alcaeus:
@coudenysj what's the state of this? I could use some of this stuff for some sharding magic in references but want to hold off until we're done changing the whole simple vs. full thing. Any progress you can report?

coudenysj:
Hi @alcaeus, no further progress on this. If we can agree upon naming, I can do some more programming to finalize this.

alcaeus:
Right. How about `id`, `dbRef` and `dbRefWithDb`? Sounds silly but in the end it's the term that best explains the difference between `dbRef` and `dbRefWithDb`. What do you think?

coudenysj:
I can live with that :). Let's see if I can work on it.

coudenysj:
Owkay, so I've added the option to all drivers and scanned the code for the "simple" reference.

Is it the correct approach if I try to "redirect" the "simple" option to the "storeAs: id" option, and refactor all the "simple" checks?

alcaeus:
Yes. Within ODM we can change all checks from `simple` to `storeAs == 'ID'`. I don't think we need to keep the `simple` flag around in the mapping for BC reasons - opinions @malarzm @jmikola?

malarzm:
I think we should keep it

coudenysj:
I will try to support both (`simple=true` triggers `storeAs=id` and `storeAs=id` triggers `simple=true`). I will however refactor all "simple" checks to "storeAs" checks in the code.

coudenysj:
If this TravisCI check comes back positive, I think this pull request is ready.

coudenysj:
@alcaeus @malarzm Any suggestions of unit tests that should be added?

docs/en/reference/reference-mapping.rst
```diff
@@ -314,6 +314,8 @@ a certain class, you can optionally specify a default discriminator value:
 Simple References
 -----------------
 
+TODO: Rewrite this to reflect the new ``storeAs`` approach.
```
malarzm: Commenting on this so we don't forget to revisit before merging

tests/Doctrine/ODM/MongoDB/Tests/Mapping/ClassMetadataInfoTest.php
```diff
@@ -234,6 +234,38 @@ public function testSimpleReferenceRequiresTargetDocument()
             'simple' => true,
         ));
     }
+
+    /**
+     * @expectedException \Doctrine\ODM\MongoDB\Mapping\MappingException
+     * @expectedExceptionMessage Target document must be specified for simple reference: stdClass::assoc
```
malarzm: I think we should now change the exception message or at least leave a `@todo` to do so in 2.0

lib/Doctrine/ODM/MongoDB/Query/Expr.php
```diff
@@ -74,11 +75,19 @@ public function references($document)
             $mapping = $this->getReferenceMapping();
             $dbRef = $this->dm->createDBRef($document, $mapping);
 
-            if (isset($mapping['simple']) && $mapping['simple']) {
+            if (array_key_exists('storeAs', $mapping)
+                && $mapping['storeAs'] === ClassMetadataInfo::REFERENCE_STORE_AS_ID
```
malarzm: Nitpicking, but how about extracting `$mapping['storeAs']` to a variable to avoid such two lines ifs?

doctrine-mongo-mapping.xsd
```diff
@@ -145,7 +146,8 @@
     <xs:attribute name="target-document" type="xs:string" />
     <xs:attribute name="field" type="xs:NMTOKEN" use="required" />
     <xs:attribute name="fieldName" type="xs:NMTOKEN" />
-    <xs:attribute name="simple" type="xs:boolean" />
+    <xs:attribute name="simple" type="xs:boolean" /><!-- deprecated -->
+    <xs:attribute name="storeAs" type="xs:NMTOKEN" default="dbRefWithDb" />
```
malarzm: Could you enumerate possible values so `storeAs` will be validated to have proper value?

malarzm:
@coudenysj I've left few comments :) As for unit tests I'd definitely add one for `references` and `includesReferenceTo` as they've changed. Other than that thanks for working on this! :)

tests/Doctrine/ODM/MongoDB/Tests/Mapping/ClassMetadataInfoTest.php
coudenysj: I was thinking about changes this (and the Simple Reference documentation), but I think the "Simple Reference" label makes more sense than "Store As Id References".

malarzm: If only we mention in the docs that by simple we mean storing as id I'm ok :+1: :)

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataInfo.php
```diff
@@ -97,6 +97,14 @@ class ClassMetadataInfo implements \Doctrine\Common\Persistence\Mapping\ClassMet
     const MANY = 'many';
     const ONE = 'one';
 
+    /**#@+
+     * The types of storeAs references
+     */
+    const REFERENCE_STORE_AS_ID = 'id';
+    const REFERENCE_STORE_AS_DB_REF = 'dbRef';
+    const REFERENCE_STORE_AS_DB_REF_WITH_DB = 'dbRefWithDb';
+    /**#@-*/
```
alcaeus: Could you remove the extra comment? I assume it's related to code folding, but since each IDE has their own way of doing this I'd vote for not including any of this. Thanks.

lib/Doctrine/ODM/MongoDB/DocumentManager.php
```diff
@@ -689,9 +689,12 @@ public function createDBRef($document, array $referenceMapping = null)
         $dbRef = array(
             '$ref' => $class->getCollection(),
             '$id'  => $class->getDatabaseIdentifierValue($id),
-            '$db'  => $this->getDocumentDatabase($class->name)->getName(),
         );
 
+        if ($referenceMapping['storeAs'] === ClassMetadataInfo::REFERENCE_STORE_AS_DB_REF_WITH_DB) {
```
alcaeus: I've mentioned this in the original PR (https://github.com/doctrine/mongodb-odm/pull/807#discussion_r34948481) and I'll mention it again: `Expr::references` and `Expr::includesReferenceTo` currently rely on the `$db` option being set when no `targetDocument` is given.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataInfo.php
```diff
@@ -1181,7 +1189,20 @@ public function mapField(array $mapping)
...
+        }
+        // Remove the "simple" mapping and use "storeAs" in all further logic
+        if (isset($mapping['simple'])) {
+            unset($mapping['simple']);
+        }
+
+        if (isset($mapping['reference'])
+            && isset($mapping['storeAs'])
+            && $mapping['storeAs'] === ClassMetadataInfo::REFERENCE_STORE_AS_ID
+            && ! isset($mapping['targetDocument'])
```
alcaeus: Not having a `targetDocument` set is one case where we need to throw an exception - the other would be `targetDocument` being stored in a different database than the document storing the reference.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataInfo.php
coudenysj: It is an "official" way of group documentation (https://en.wikipedia.org/wiki/PHPDoc#DocBlock_Templates), but I will refactor it.

tests/Doctrine/ODM/MongoDB/Tests/Mapping/ClassMetadataInfoTest.php
coudenysj: @malarzm Updated the docs in 74d0ff9d3b168b (and I've added a not about the old "simple reference" term).

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataInfo.php
coudenysj: @alcaeus I'm not quite sure how I can fetch the mapping for the targetDocument in the ClassMetadataInfo class.

malarzm: Unfortunately you can't, IMO that could be detected at the runtime but I have a feeling there's way more things than references that could bite you in such cases so that wouldn't be my main concern :)

lib/Doctrine/ODM/MongoDB/Query/Expr.php
```diff
@@ -106,12 +112,18 @@ public function includesReferenceTo($document)
         if ($this->currentField) {
             $mapping = $this->getReferenceMapping();
             $dbRef = $this->dm->createDBRef($document, $mapping);
+            $storeAs = array_key_exists('storeAs', $mapping) ? $mapping['storeAs'] : null;
 
-            if (isset($mapping['simple']) && $mapping['simple']) {
+            if ($storeAs === ClassMetadataInfo::REFERENCE_STORE_AS_ID) {
+                // TODO: This creates an invalid query ($elemMatch needs an Object)
                 $this->query[$mapping['name']]['$elemMatch'] = $dbRef;
```
coudenysj: I think I discoverer a bug here, because the ``$elemMatch`` is used to search in arrays, but the simple reference contains only a ``MongoId``.

tests/Doctrine/ODM/MongoDB/Tests/Query/BuilderTest.php
```diff
@@ -77,7 +89,25 @@ public function testIncludesReferenceToGoesThroughDiscriminatorMap()
             ->field('featureSimpleMany')->includesReferenceTo($f)
             ->getQuery()->debug();
 
+        // TODO: Looks like this query is not valid ($elemMatch needs an Object)
+
         $this->assertEquals([ 'featureSimpleMany' => [ '$elemMatch' => new \MongoId($f->id) ] ], $q2['query']);
```
coudenysj: This tests the simple reference via ``$elemMatch``, but this is not a valid query (throws ``$elemMatch needs an Object`` exception)

tests/Doctrine/ODM/MongoDB/Tests/QueryTest.php
```diff
@@ -105,6 +158,68 @@ public function testIncludesReferenceTo()
...
+            'friendsSimple' => array(
+                '$elemMatch' => new \MongoId($kris->id)
+            ),
+        ), $queryArray);
+
+        //$query = $qb->getQuery();
+
+        // TODO: Looks like this is not valid:
+        // MongoException: Cannot run command count(): $elemMatch needs an Object
+        //$this->assertEquals(1, $query->count());
```
coudenysj: When running this query, I get ``MongoException: Cannot run command count(): $elemMatch needs an Object``.

coudenysj:
I think this pull request is ready. I will create a new pull request to fix the comments I gave about the ``$elemMatch needs an Object``.

malarzm:
Cool thanks, I'll try to review this once again today/tomorrow :)

lib/Doctrine/ODM/MongoDB/Query/Expr.php
alcaeus: Absolutely right - thanks for catching that!

docs/en/reference/reference-mapping.rst
```diff
@@ -331,19 +329,19 @@ Example:
         <?php
 
         /**
-         * @ReferenceOne(targetDocument="Profile", simple=true)
+         * @ReferenceOne(targetDocument="Profile", storeAs=id)
```
alcaeus: The `id` string should be quoted

lib/Doctrine/ODM/MongoDB/Query/Expr.php
```diff
@@ -115,6 +115,7 @@ public function includesReferenceTo($document)
             $storeAs = array_key_exists('storeAs', $mapping) ? $mapping['storeAs'] : null;
 
             if ($storeAs === ClassMetadataInfo::REFERENCE_STORE_AS_ID) {
+                // TODO: This creates an invalid query ($elemMatch needs an Object)
```
alcaeus: Leaving a note to remove the TODO once the PR containing a fix has been created.

tests/Doctrine/ODM/MongoDB/Tests/Query/BuilderTest.php
```diff
@@ -77,7 +89,25 @@ public function testIncludesReferenceToGoesThroughDiscriminatorMap()
             ->field('featureSimpleMany')->includesReferenceTo($f)
             ->getQuery()->debug();
 
+        // TODO: Looks like this query is not valid ($elemMatch needs an Object)
```
alcaeus: Leaving a note to remove the TODO once the PR containing a fix has been created.

coudenysj: I've created #1405 to fix this issue. Is it ok if the TODO will be removed in that pull request?

alcaeus: I see you've based #1405 on top of this branch. In this case, yes we can.

alcaeus:
#1406 has been merged into 1.0.x and master - you can rebase this PR if you want and update the fix.

coudenysj:
Processed changes from #1406, so this PR is ready for final review.

docs/en/reference/reference-mapping.rst
```diff
@@ -351,6 +351,28 @@ itself and any indexes on the reference field; however, simple references cannot
...
+
+The ``storeAs`` option has three possible values:
+
+- **dbRefWithDb**: Uses a `DBRef`_ with ``$ref``, ``$id``, and ``$db`` fields (this is the default)
+- **dbRef**: Uses a `DBRef`_ with ``$ref`` and ``$id``
+- **id**: Uses a ``MongoId``
+
+.. note::
+
+    The ``storeAs=id`` option used to call a "simple reference". The old syntax is
```
malarzm: "used to be called"

tests/Doctrine/ODM/MongoDB/Tests/Mapping/ClassMetadataInfoTest.php
malarzm: :+1:

malarzm:
LGTM, thanks @coudenysj for working on this!

alcaeus:
Thank you @coudenysj!

PR #1352, opened 2016-01-27
Replace increment type with a storage strategy

alcaeus:
This PR deprecates the `increment` type that can be used in ODM, replacing it instead with a more flexible `strategy` option for fields.

First of all, the `increment` type was supposed to handle int and float values at the same time. There was no option to ensure a value read from the database was cast to an int (which could lead to unexpected floats appearing in the application if one wasn't careful).
Furthermore, for collections there is a `strategy` option which controls how the elements in a collection are written to the database. This kind of write control is exactly what should be used for increment fields as well.

At first glance, there are no BC breaks to this. When the `ClassMetadataInfo` class encounters a mapping with type `increment`, it will automatically set the strategy to `increment`, regardless of any other strategy being set. The type is left unchanged to keep the current type cast behavior. Users are encouraged to switch their increment mappings to the new strategy field.

Open tasks:
- [x] Write proper documentation about the new strategy, add deprecation notes for the `increment` type
- [x] Collection strategies still use magic strings, these should be converted to class constants for good measure
- [x] The validation which strategy is acceptable for which mapping should probably be generalized and put in one place to ensure consistency.

@malarzm: I'd appreciate your guidance on the last two tasks since you've spent quite some time with collections ;)

### DIFF
--- a/docs/en/reference/storage-strategies.rst
+++ b/docs/en/reference/storage-strategies.rst
@@ -1,29 +1,37 @@
-.. _collection_strategies:
+.. _storage_strategies:
 
-Collection Strategies
-=====================
+Storage Strategies
+==================
 
-Doctrine MongoDB ODM implements four different strategies for persisting changes
-to collections of embedded documents or references. These strategies apply to
-the following mapping types:
+Doctrine MongoDB ODM implements several different strategies for persisting changes
+to mapped fields. These strategies apply to the following mapping types:
 
+- :ref:`int`
+- :ref:`float`
 - :ref:`embed_many`
 - :ref:`reference_many`
 
-Internally, Doctrine tracks changes via the PersistentCollection class. The
+For collections, Doctrine tracks changes via the PersistentCollection class. The
 strategies described on this page are implemented by the CollectionPersister
-class.
+class. The ``increment`` strategy cannot be used for collections.
 
-``addToSet``
-------------
+increment
+---------
+
+The ``increment`` strategy does not apply to collections but can be used for
+``int`` and ``float`` fields. When using the ``increment`` strategy, the field
+value will be updated using the `$inc`_ operator.
+
+addToSet
+--------
 
 The ``addToSet`` strategy uses MongoDB's `$addToSet`_ operator to insert
 elements into the array. This strategy is useful for ensuring that duplicate
 values will not be inserted into the collection. Like the `pushAll`_ strategy,
 elements are inserted in a separate query after removing deleted elements.
 
-``set``
--------
+set
+---
 
 The ``set`` strategy uses MongoDB's `$set`_ operator to update the entire
 collection with a single update query.
@@ -37,15 +45,15 @@ collection with a single update query.
     `setArray`_ strategy if you want to ensure that the collection is always
     stored as a BSON array.
 
-``setArray``
-------------
+setArray
+--------
 
 The ``setArray`` strategy uses MongoDB's `$set`_ operator, just like the ``set``
 strategy, but will first numerically reindex the collection to ensure that it is
 stored as a BSON array.
 
-``pushAll``
------------
+pushAll
+-------
 
 The ``pushAll`` strategy uses MongoDB's `$pushAll`_ operator to insert
 elements into the array. MongoDB does not allow elements to be added and removed
@@ -54,8 +62,8 @@ queries to remove and insert elements (in that order).
 
 .. _atomic_set:
 
-``atomicSet``
--------------
+atomicSet
+---------
 
 The ``atomicSet`` strategy uses MongoDB's `$set`_ operator to update the entire
 collection with a single update query. Unlike with ``set`` strategy there will
@@ -70,8 +78,8 @@ strategy can be especially useful when dealing with high concurrency and
 
 .. _atomic_set_array:
 
-``atomicSetArray``
-------------------
+atomicSetArray
+--------------
 
 The ``atomicSetArray`` strategy works exactly like ``atomicSet`` strategy,  but 
 will first numerically reindex the collection to ensure that it is stored as a 
@@ -83,6 +91,7 @@ BSON array.
     collections mapped directly in a top-level document.
 
 .. _`$addToSet`: http://docs.mongodb.org/manual/reference/operator/addToSet/
+.. _`$inc`: http://docs.mongodb.org/manual/reference/operator/inc/
 .. _`$pushAll`: http://docs.mongodb.org/manual/reference/operator/pushAll/
 .. _`$set`: http://docs.mongodb.org/manual/reference/operator/set/
 .. _`$unset`: http://docs.mongodb.org/manual/reference/operator/unset/
--- a/doctrine-mongo-mapping.xsd
+++ b/doctrine-mongo-mapping.xsd
@@ -64,7 +64,7 @@
     <xs:attribute name="id" type="xs:boolean" default="false" />
     <xs:attribute name="name" type="xs:NMTOKEN" />
     <xs:attribute name="type" type="xs:NMTOKEN" default="string" />
-    <xs:attribute name="strategy" type="xs:NMTOKEN" default="pushAll" />
+    <xs:attribute name="strategy" type="xs:NMTOKEN" default="set" />
     <xs:attribute name="fieldName" type="xs:NMTOKEN" />
     <xs:attribute name="embed" type="xs:boolean" />
     <xs:attribute name="file" type="xs:boolean" />
--- a/lib/Doctrine/ODM/MongoDB/Mapping/Annotations/AbstractField.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/Annotations/AbstractField.php
@@ -27,4 +27,5 @@ abstract class AbstractField extends Annotation
     public $type = 'string';
     public $nullable = false;
     public $options = array();
+    public $strategy;
 }
--- a/lib/Doctrine/ODM/MongoDB/Mapping/Annotations/Increment.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/Annotations/Increment.php
@@ -21,9 +21,11 @@ namespace Doctrine\ODM\MongoDB\Mapping\Annotations;
 
 /**
  * @Annotation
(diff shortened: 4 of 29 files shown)


doctrine-mongo-mapping.xsd
```diff
@@ -64,7 +64,7 @@
     <xs:attribute name="id" type="xs:boolean" default="false" />
     <xs:attribute name="name" type="xs:NMTOKEN" />
     <xs:attribute name="type" type="xs:NMTOKEN" default="string" />
-    <xs:attribute name="strategy" type="xs:NMTOKEN" default="pushAll" />
+    <xs:attribute name="strategy" type="xs:NMTOKEN" default="set" />
```
malarzm: Why this change?

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/Id.php
```diff
@@ -18,6 +18,7 @@
  */
 
 namespace Doctrine\ODM\MongoDB\Mapping\Annotations;
+use Doctrine\ODM\MongoDB\Mapping\ClassMetadataInfo;
```
malarzm: Not needed?

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/Increment.php
```diff
@@ -26,4 +26,5 @@
 final class Increment extends AbstractField
 {
     public $type = 'increment';
+    public $strategy = 'increment';
```
malarzm: Should we add additional note to deprecation message?

doctrine-mongo-mapping.xsd
malarzm: Ah ok, got it after looking further :) Won't it break collections though? (note: I'm n00b when it comes to XMLs and theiir stuff)

malarzm:
Like the approach and where it can go :) I'll elaborate on later today

doctrine-mongo-mapping.xsd
alcaeus: This type only applies to scalar fields, the association types have a different mapping. Technically, this field was not used to begin with ;)

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/Id.php
alcaeus: Whoopsie.

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/Increment.php
alcaeus: Yes, I'll add a bit about using increment strategy instead of increment type.

lib/Doctrine/ODM/MongoDB/Mapping/MappingException.php
```diff
@@ -275,4 +275,15 @@ public static function referenceManySortMustNotBeUsedWithNonSetCollectionStrateg
...
         return new self("ReferenceMany's sort can not be used with addToSet and pushAll strategies, $strategy used in $className::$fieldName");
     }
+
+    /**
+     * @param string $className
+     * @param string $fieldName
+     * @param string $strategy
+     * @return MappingException
+     */
+    public static function invalidStorageStrategy($className, $fieldName, $strategy)
```
malarzm: I'd pass type here as well, optionally we could also throw in allowed strategies and propose correct one basing on Levenshtein distance? But it can be too much :D

malarzm:
> Collection strategies still use magic strings, these should be converted to class constants for good measure

`ClassMetadataInfo::STORAGE_STRATEGY_ATOMIC_SET_ARRAY`? I'd pass on this :) Also it's easier to write `atomicSetArray` in mapping than calling constant I think. Also for internal usage they're pretty much hidden by [CollectionHelper](https://github.com/doctrine/mongodb-odm/blob/453ef86b8c499b05fefe9b67347b8b9c06e3f176/lib/Doctrine/ODM/MongoDB/Utility/CollectionHelper.php)

> The validation which strategy is acceptable for which mapping should probably be generalized and put in one place to ensure consistency.

:+1: on this, end of `ClassMetadataInfo::mapField` seems like a good place for this especially there are no early returns. We could put some `switch` over there that would check certain types if their strategy is allowed (and throw exception if not) and by default unset whatever was passed there.

docs/en/reference/storage-strategies.rst
```diff
@@ -1,29 +1,37 @@
...
+class. The ``increment`` strategy cannot be used for collections.
 
-``addToSet``
-------------
+increment
+---------
+
+The ``increment`` strategy does not apply to collections but can be used for
+``int`` and ``float`` fields. When using the ``increment`` strategy, the field
+value will be updated using an `$inc`_ operator.
```
jmikola: I think we can write "the `$inc` operator" here.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataInfo.php
```diff
@@ -1227,6 +1245,55 @@ public function mapField(array $mapping)
     }
 
     /**
+     * Validates the storage strategy of a mapping for consistency
+     * @param array $mapping
+     * @throws \Doctrine\ODM\MongoDB\Mapping\MappingException
+     */
+    private function validateStorageStrategy(array &$mapping)
```
jmikola: "Validate" doesn't convey that we're modifying the `$mapping` argument. Perhaps "apply" would be more appropriate?

lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
```diff
@@ -246,31 +241,28 @@ public function prepareUpsertData($document)
...
                 // If we have a new embedded document then lets set the whole thing
                 if ($new && $this->uow->isScheduledForInsert($new)) {
                     $updateData['$set'][$mapping['name']] = $this->prepareEmbeddedDocumentValue($mapping, $new);
-
-                // If we don't have a new value then do nothing on upsert
-                } elseif ( ! $new) {
-
-                // Update existing embedded document
-                } else {
+                } elseif ( $new) {
```
jmikola: Spacing seems a bit odd here.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataInfo.php
```diff
@@ -1227,6 +1245,55 @@ public function mapField(array $mapping)
...
+    {
+        if (! isset($mapping['type']) || isset($mapping['id'])) {
```
jmikola: Should the `!` be surrounded by spaces on both sides for CS?

malarzm: we lost with PSR after all :(

docs/en/reference/storage-strategies.rst
```diff
@@ -1,29 +1,37 @@
...
 
-Collection Strategies
-=====================
+Storage Strategies
+==================
 
-Doctrine MongoDB ODM implements four different strategies for persisting changes
-to collections of embedded documents or references. These strategies apply to
-the following mapping types:
+Doctrine MongoDB ODM implements seven different strategies for persisting changes
```
malarzm: How about changing seven to several? I noticed "four" only lately and haven't got to changing it to "six", I suppose it'll be the same with "seven" when we add eighth one :D

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -1043,7 +1043,7 @@ private function prepareQueryElement($fieldName, $value = null, $class = null, $
             return array($fieldName, $value);
         }
 
-        if (isset($mapping['strategy']) && CollectionHelper::isHash($mapping['strategy'])
+        if ($mapping['type'] == 'many' && isset($mapping['strategy']) && CollectionHelper::isHash($mapping['strategy'])
```
malarzm: Now that all fields have `strategy` set we could get rid of all `isset($mapping['strategy'])` checks, right? For reference I introduced few with https://github.com/doctrine/mongodb-odm/pull/1177

lib/Doctrine/ODM/MongoDB/Utility/CollectionHelper.php
```diff
@@ -70,6 +71,14 @@ public static function isList($strategy)
      */
     public static function usesSet($strategy)
     {
-        return in_array($strategy, array('set', 'setArray', 'atomicSet', 'atomicSetArray'));
+        return in_array(
+            $strategy,
+            array(
```
malarzm: Short array syntax?

tests/Doctrine/ODM/MongoDB/Tests/Mapping/Driver/AbstractDriverTest.php
```diff
@@ -57,7 +58,8 @@ public function testDriver()
             'isOwningSide' => true,
             'nullable' => false,
             'unique' => true,
-            'sparse' => true
+            'sparse' => true,
+            'strategy' => 'set',
```
malarzm: This should use constant :)

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/Increment.php
malarzm: :+1:

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/AbstractField.php
```diff
@@ -20,6 +20,7 @@
 namespace Doctrine\ODM\MongoDB\Mapping\Annotations;
 
 use Doctrine\Common\Annotations\Annotation;
+use Doctrine\ODM\MongoDB\Mapping\ClassMetadataInfo;
```
malarzm: We don't need this I think?

malarzm:
Merged manually in https://github.com/doctrine/mongodb-odm/commit/b1cbaf8d674a6fe8d1d6d0db88ce362b8e15b0ec

PR #1355, opened 2016-01-31
Fix initializing lazy properties during hydration

malarzm:
Fixes #1347 

~~Marking as WIP as for some reason PHP segfaults on my local machine, let's see what Travis will say~~ - travis is fine, my machine will need an investigation of what's going on


malarzm:
@zajca test you provided is commited here and passes, but if you could test this also in your app it'd be great :)

alcaeus:
Changes look good :+1:

PR #1358, opened 2016-02-08
Document changeTrackingPolicy annotation

alcaeus:
While change tracking policies are explained in the documentation, the annotations reference should contain a quick explanation about the respective mapping annotation for policies.


malarzm:
Merged manually in https://github.com/doctrine/mongodb-odm/commit/5ec3f1b1743f10db15caf267c7d08f2b09a6e3b3

PR #1360, opened 2016-02-11
Fix PHP 7 compatibility errors

alcaeus:
This PR doesn't bring full PHP 7 compatibility but it removes a couple of issues that will arise should one manage to get ODM running on PHP 7:
- As mentioned in the [PHP 7 Upgrade log](https://secure.php.net/manual/de/migration70.incompatible.php), changes to variable handling cause an array to string conversion in DocumentPersister
- There were a couple of deprecated annotations still being used in tests. Due to string and integer being reserved keywords in PHP 7, these annotations are not loaded and cause errors.


malarzm:
LGTM :+1:

PR #1363, opened 2016-02-14
Prepare field's value for references

### LINKED ISSUE #292: findBySomething($object) doesn't work
I've field with ReferenceOne annotate:

```php
class Article
{
// ...
    /**
     *
     * @ODM\ReferenceOne(targetDocument="Document\Category", inversedBy="articles", simple=true, cascade={"all"}, nullable=true)
     */
    protected $category;
// ...
}
```

Now in its repository I'm trying:

```php
class ArticleRepository extends DocumentRepository
{
    // ...
    public function findByCategory(Category $category)
    {
        return parent::findByCategory($category->getId());
    }
    // ...
}
// ...
```

Yes, I know, it's stupid, but because of DocumentPersister::prepareQuery I can't do it like:

```php
return parent::findByCategory($category);
```

Maybe DocumentPersister::prepareQuery should check if first param is array or not (object!) and tries to find getId method (or id attribute). What do you think, @jwage?

Then it will be possible to do annotation like:

```php
class Category
{
    // ...
    /**
     * @ODM\ReferenceMany(targetDocument="Document\Article", mappedBy="category", repositoryMethod="findByCategory", simple=true, cascade={"all"}, nullable=true)
     */
    protected $articles;
    // ...
}
```

nice, isn't it?

### LINKED ISSUE #295: Unique constraint (from DoctrineMongoDBBundle) doesn't work with embedded document
I'm trying to use the "unique validator" from the DoctrineMongoDBBundle, but  I get this error:
[BadMethodCallException] Doctrine\ODM\MongoDB\Mapping\ClassMetadataInfo::getAssociationTargetClass($assocName) is not implemented yet.

Am I doing something wrong?

I use Symfony 2.1, and latest bundles/projects versions.

malarzm:
This is a subset of ODM preparing query according to field's mapping, but as it's bigger thing here's small part that is cool enough to make its way to 1.1 already - now one can query for references without using query builder! It even doesn't matter if reference is one or many, simple or not:

```
$dm->getRepository(User::class)->findOneBy(['account' => $account]);
$dm->getRepository(User::class)->findBy(['groups' => $group]);
```

Additionally to being nice it should fix https://github.com/doctrine/DoctrineMongoDBBundle/issues/208 and close #292 :)

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -975,8 +975,7 @@ class DocumentPersister
                 return array($fieldName, $this->pb->prepareEmbeddedDocumentValue($mapping, $value));
             }
 
-            if (! empty($mapping['reference']) && $mapping['type'] === ClassMetadata::ONE
-                && is_object($value) && ! ($value instanceof \MongoId)) {
+            if (! empty($mapping['reference']) && is_object($value) && ! ($value instanceof \MongoId)) {
                 try {
                     return array($fieldName, $this->dm->createDBRef($value, $mapping));
                 } catch (MappingException $e) {
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -20,6 +20,7 @@
 namespace Doctrine\ODM\MongoDB\Persisters;
 
 use Doctrine\Common\EventManager;
+use Doctrine\Common\Persistence\Mapping\MappingException;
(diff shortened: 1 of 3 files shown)


alcaeus:
Wow, so simple one could ask why we haven't thought of this before. Does this also work in query builder or does that still have to go through `reference`?

```
$queryBuilder
    ->field('account')
    ->equals($account);
```

malarzm:
Should work already, but won't be as smart as `references` after #1333 and I'm not sure yet whether it should :)

alcaeus:
:+1: Sounds good!

malarzm:
Merged manually in https://github.com/doctrine/mongodb-odm/commit/fbbe096078943cb13941ceb013b8cc693bc8a87e

PR #1370, opened 2016-02-29
Corrected fluent interface docblock annotations.

Bilge:
In the same vein as doctrine/mongodb#251.


malarzm:
@Bilge would you mind changing bunch of `@return Builder` as well? They're all returning `$this` so let's make it consistent :)

Bilge:
@malarzm Done.

malarzm:
Thanks! :+1:

PR #1371, opened 2016-02-29
Removed unused imports from library and tests.

Bilge:
Noticed there were a lot of unused imports throughout the code base so I removed them. Also removed some trailing whitespace and fixed files missing a newline at the end of the file.


malarzm:
Tests are failing now, I think I fell into something similar with one of tests that generates classes when removed "unused" imports

Bilge:
The issue is with `GH1299User`. It extends the class `BaseUser`, which is stored in a separate file but in the same directory as `GH1299User`. Normally PHP will search for files in the same directory as the current script but clearly that is not happening in this case, so it needs the import to find `BaseUser`.

malarzm:
Merged manually in https://github.com/doctrine/mongodb-odm/commit/8f1c34f0ec338dc95393ea44e8bdceb56df0b413 - thanks!

PR #1372, opened 2016-02-29
DocumentRepository code quality improvements.

Bilge:
Made some rudimentary code and documentation quality improvements.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/DocumentRepository.php
+++ b/lib/Doctrine/ODM/MongoDB/DocumentRepository.php
@@ -20,9 +20,12 @@
 namespace Doctrine\ODM\MongoDB;
 
 use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Common\Collections\Collection;
 use Doctrine\Common\Collections\Criteria;
 use Doctrine\Common\Collections\Selectable;
+use Doctrine\Common\Inflector\Inflector;
 use Doctrine\Common\Persistence\ObjectRepository;
+use Doctrine\ODM\MongoDB\Mapping\ClassMetadata;
 use Doctrine\ODM\MongoDB\Query\QueryExpressionVisitor;
 
 /**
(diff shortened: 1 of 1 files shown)


lib/Doctrine/ODM/MongoDB/DocumentRepository.php
```diff
@@ -91,19 +93,19 @@ public function clear()
...
 
     /**
-     * Finds a document by its identifier
+     * Finds a document by the specified identifier.
      *
-     * @param string|object $id The identifier
-     * @param int $lockMode
+     * @param mixed $id Identifier.
+     * @param int $lockMode Optional. Locking mode.
      * @param int $lockVersion
```
Bilge: I don't know what this is so I couldn't document it.

malarzm: This can be used to indicate which version of document you expect to be fetched, throws exception in case of mismatch

lib/Doctrine/ODM/MongoDB/DocumentRepository.php
```diff
@@ -259,7 +260,7 @@ public function getClassName()
      *
      * @see Selectable::matching()
      * @param Criteria $criteria
-     * @return Collection
+     * @return ArrayCollection
```
malarzm: This should stay with `Collection` in case we might stop returning `ArrayCollection` in favour of some other implementation

malarzm:
@Bilge I left you two comments, please amend them into your commit (and rebase to get rid of conflicts ;) )

lib/Doctrine/ODM/MongoDB/DocumentRepository.php
Bilge: OK, in that case the import needs adding because the name was an unresolved reference before.

malarzm:
Thanks @Bilge!

PR #1373, opened 2016-02-29
Removed redundant parens and incorrect double-quote usage.

Bilge:
Threw in ternary shorthand as a bonus. This feature was introduced in PHP 5.3 so no reason not to use it.


malarzm:
Thanks @Bilge :+1:

PR #1374, opened 2016-02-29
Added missing imports.

Bilge:


lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataFactory.php
```diff
@@ -24,7 +24,13 @@
 use Doctrine\Common\Persistence\Mapping\ReflectionService;
 use Doctrine\ODM\MongoDB\Configuration;
 use Doctrine\ODM\MongoDB\DocumentManager;
+use Doctrine\ODM\MongoDB\Event\LoadClassMetadataEventArgs;
 use Doctrine\ODM\MongoDB\Events;
+use Doctrine\ODM\MongoDB\Id\AbstractIdGenerator;
+use Doctrine\ODM\MongoDB\Id\AlnumGenerator;
+use Doctrine\ODM\MongoDB\Id\AutoGenerator;
+use Doctrine\ODM\MongoDB\Id\IncrementGenerator;
+use Doctrine\ODM\MongoDB\Id\UuidGenerator;
```
malarzm: How do you feel about changing these 5 to `use Doctrine\ODM\MongoDB\Id as IdGenerators`?

Bilge: I think:

1. Partial imports are bad
2. Aliases are bad
3. You want to do 1 and 2 which is superbad :cat:

malarzm: Ok, let's keep it then, matter of taste I guess. Still this one needs rebase tho :)

Bilge: I'll take care of rebasing momentarily.

Bilge: Aliases create new names which may cause you to think it's a new class you've not seen before or cause you to fail to recognize a class you're already familiar with because it now has a different name. For this reason I opine that aliases are objectively a bad idea whenever they can be avoided. Partial imports are, however, possibly a matter of taste.

I'll take care of rebasing momentarily.

malarzm:
Thanks @Bilge!

PR #1376, opened 2016-02-29
Added missing type hints.

Bilge:


lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -147,7 +147,7 @@ public function setMetadataDriverImpl(MappingDriver $driverImpl)
      * @param array $paths
      * @return Mapping\Driver\AnnotationDriver
      */
-    public function newDefaultAnnotationDriver($paths = array())
+    public function newDefaultAnnotationDriver(array $paths = array())
```
alcaeus: I'm afraid this is a BC break and needs to stay out. Until now it is possible to pass in a string, which will be converted to an array containing that one string when `$paths` is cast to array. Also, method signature changes in public methods is a no-go in SemVer unless you do it in a major version change.

alcaeus:
@Bilge I refrained from adding redundant comments but we can't add these typehint changes. Any chance you could take the whitespace changes and integrate them into one of your other PRs?

Bilge:
I sincerely doubt anyone cares about whitespace changes. I certainly don't.

PR #1379, opened 2016-03-04
Introduce postCollectionLoad event

malarzm:
This event will effectively allow developer to implement own change tracking in collections as event is dispatched after UoW loads collection but before re-adding new elements (same time as `PersistentCollection` takes its snapshot).

- [x] dispatch event
- [x] ~~annotated callback~~ it's too hacky to introduce with current `ClassMetadata` :(
- [x] docs

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionTrait.php
+++ b/lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionTrait.php
@@ -455,7 +455,7 @@ trait PersistentCollectionTrait
                 : $documentPersister->createReferenceManyWithRepositoryMethodCursor($this)->count();
         }
 
-        return count($this->mongoData) + $count;
+        return $count + ($this->initialized ? 0 : count($this->mongoData));
     }
 
     /**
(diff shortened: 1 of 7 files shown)


lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionTrait.php
```diff
@@ -455,7 +455,7 @@ public function count()
                 : $documentPersister->createReferenceManyWithRepositoryMethodCursor($this)->count();
         }
 
-        return count($this->mongoData) + $count;
+        return $count + ($this->initialized ? 0 : count($this->mongoData));
```
alcaeus: I can see why we don't count `mongoData` after a collection has been initialized (which means all elements have been added to `coll`). Does this fix any bugs and if so, do we need to include this in 1.0.x?

malarzm: No, before there was no way to call `->count()` in the middle of initialization which can be done now. Alternatively we could clear `$this->mongoData` before firing an event but maybe somebody would it useful to have?

alcaeus: Ah, I see. Thanks!

malarzm:
FWIW I've added missing PHPDoc and note in docs about version when event was introduced

alcaeus:
Merged manually in fce7110.

PR #1382, opened 2016-03-09
Fix broken conversion of microseconds less than 100000.

jpinner-lyft:
Dates with microsecond values less than 100000 are not correctly converted to PHP objects.

When converting from MongoDate objects, the `usec` field returns an integer that is concatenated directly to a string formated date in `craftDateTime`. For example, the value 1.0001 gets split into 1 second and 100 microseconds, which then gets concatenated as "1.100".

When converting numeric values, the sub-seconds are broken into a string which is then padded to six values; however, it is casted to an int first. For example, 1.0001 gets split into a string "0001" which is cast to the integer 1 and padded to "100000", resulting in "1.100000".

In order to properly convert microseconds, they are string padded without converting to an integer.


alcaeus:
Thanks!

PR #1383, opened 2016-03-11
Allow custom collections to modify their state

malarzm:
@alcaeus this should fix your problems :)


alcaeus:
Awesome - will test this on Monday to make sure it works and report back!

alcaeus:
Merged manually in 0573f48. Thanks!

PR #1386, opened 2016-03-15
Don't persist inverse references in embedded docs

alcaeus:
This fixes an error that occurs in rare cases when a root document is also stored embedded in another document. In those cases, inverse references are no longer treated as inverse; instead they are stored directly in the database. With this PR, these references are still treated as inverse and not persisted in the database.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
@@ -346,8 +346,8 @@ class PersistenceBuilder
         $class = $this->dm->getClassMetadata(get_class($embeddedDocument));
 
         foreach ($class->fieldMappings as $mapping) {
-            // Skip notSaved fields
-            if ( ! empty($mapping['notSaved'])) {
+            // Skip notSaved fields or inversed collections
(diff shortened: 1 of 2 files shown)


malarzm:
My vote goes to finally disallowing using document-mapped objects as embedded documents. Valid way is to have common `MappedSuperclass` extended by two classes, one mapped as `Document` and another one as `EmbeddedDocument`. Generally bad things can happen :)

JakeFalcor:
Yep. Bad things happen...

I had this in my project a few months ago, not realizing using `Document` as `EmbeddedDocument` had severe side effects. It wasn't until very last commits that this popped up, when I discovered that `Document`'s collection was also updating every time `EmbeddedDocument` was persisted. I would vote for malarzm suggestion. Documents and EmbeddedDocuments should be always separate classes.

alcaeus:
After discussion in IRC we've decided that this is the side effect of doing something one shouldn't be doing at all. Since embedded documents cannot be properly referenced (e.g. there is no way to fetch a referenced object alone by a DBRef object) it also doesn't make sense to have inverse support for inverse references in an embedded document, invalidating this feature completely.

I'll be creating a new PR to throw a `MappingException` when using a document mapped using `@Document` in an embedded association.

PR #1390, opened 2016-03-22
Fix for fatal exception if persisted collection was unserialized

notrix:
UnitOfWork could set document manager again but not sure where it should check for dm presence


darceee2:
:+1:

malarzm:
@notrix thanks for the PR! Would you mind updating `getTypeClass` method to throw an exception with different message (for instance "No DocumentManager is associated with this PersistentCollection, please set one using setDocumentManager")? We could also get `typeClass` once again if it's `null` and `DocumentManager` is available.

malarzm:
@notrix I've included your commits in #1411 and added few more things from me, thanks for starting this!

Also I'm sorry for no looking into this sooner, I haven't checked hanging PRs for a while and there are no email notifications about new commits on their own.

PR #1392, opened 2016-03-23
Fixed targetDocument types to match property names and annotations

vmattila:
A test document class ``SimpleEmbedAndReference`` contained ``targetDocument`` settings that did not match property and annotations, this PR fixes them. The only test method that actually uses the document is ``FunctionalTest::testSimplerEmbedAndReference`` that only checks the association type, not actually whether it's embedded or referenced.


malarzm:
Thanks @vmattila :+1:

PR #1395, opened 2016-03-27
Add PHP 7 to the build matrix using polyfill

alcaeus:
While composer.json has allowed installing on PHP 7 for a while (given that the `ext-mongo` requirement is fulfilled), there were no tests for that platform. This PR adds alcaeus/mongo-php-adapter as polyfill to the PHP 7 version of the build matrix.

Note: due to a [bug with composer](https://github.com/composer/composer/issues/5030) the build step installing the polyfill is running with the `ignore-platform-reqs` flag. Once this bug is fixed the flag should be removed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,14 +11,22 @@ env:
     - MONGO_VERSION=1.5.8
     - MONGO_VERSION=stable
 
+matrix:
+  include:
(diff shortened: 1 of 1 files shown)


malarzm:
\o/

PR #1403, opened 2016-05-08
Add option to fetch documents as Read-Only

### LINKED ISSUE #1377: Make Hydrators more usable outside of UoW scope
I was lately playing with hydrating data into documents and, unfortunately, hydrator assumes the data that goes through should be registered in `UnitOfWork` (this is perfectly valid assumption when hydrating data from database, but not desirable when hydrating manually). Because of call to [`UnitOfWork::registerManaged`](https://github.com/doctrine/mongodb-odm/blob/8f1c34f0ec338dc95393ea44e8bdceb56df0b413/lib/Doctrine/ODM/MongoDB/UnitOfWork.php#L2800) nested embedded documents are marked as managed and their snapshot is saved which later makes up wrong change sets when preparing update query.

Since [`hydrate`](https://github.com/doctrine/mongodb-odm/blob/8f1c34f0ec338dc95393ea44e8bdceb56df0b413/lib/Doctrine/ODM/MongoDB/Hydrator/HydratorInterface.php#L37) will gladly accept any `$hints` passed to it I'd propose adding one, that would omit any calls registering objects in associated `UnitOfWork` effectively only populating objects with data, as it could be expected. Alternatively we could decouple registering objects and do that in `UoW` instead, but that'd warrant iterating over reflection once again which is not that good performance-wise (but is praiseworthy ;) ).

Thoughts?

malarzm:
This not only fixes #1377 but also exposes `readOnly` as an option for query builder. 

Read-only takes precedence over `refresh` when combined. Document fetched in read-only manner can be `merge`d back into `DocumentManager` at developer's discretion, forbiding such action is out of scope of this PR (IMHO :)

### DIFF
--- a/docs/en/reference/query-builder-api.rst
+++ b/docs/en/reference/query-builder-api.rst
@@ -216,6 +216,32 @@ changeset will be reset in the process.
 
 Refreshing is not applicable if hydration is disabled.
 
+Fetching Documents as Read-Only
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Similar to ``refresh()``, ``readOnly()`` instructs ODM to not only hydrate the
+latest data but also to create new document's instance (i.e. if found document
+would be already managed by Doctrine, new instance will be returned) and not
+register it in ``UnitOfWork``.
+
+This technique can prove especially useful when using ``select()`` with no intent
+to update fetched documents.
+
+.. code-block:: php
+
+    <?php
+
+    $user = $dm->createQueryBuilder('User')
+        ->field('username')->equals('malarzm')
+        ->readOnly()
+        ->getQuery()
+        ->getSingleResult();
+
+    // Maciej's user will have the latest data, and will not be the same object
+    // as the one that was already managed (if it was)
+
+Read-Only is not applicable if hydration is disabled.
+
 Disabling Hydration
 ~~~~~~~~~~~~~~~~~~~
 
--- a/docs/en/reference/query-builder-api.rst
+++ b/docs/en/reference/query-builder-api.rst
@@ -242,6 +242,13 @@ to update fetched documents.
 
 Read-Only is not applicable if hydration is disabled.
 
+.. note::
+
+    Read-only mode is not deep, i.e. any references (be it owning or inverse) of
+    fetched WILL be managed by Doctrine. This is a shortcoming of current
+    implementation, may change in future and will not be considered a BC break
+    (will be treated as a feature instead).
+
 Disabling Hydration
 ~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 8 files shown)


jwage:
:+1: Looks good to me.

alcaeus:
Looks good except for one tiny detail: embedded documents are fetched as read-only while references are not considered as read-only. I'm not sure if that's the best way to go since I'd expect the read-only to be deep. I've created a PR (malarzm/mongodb-odm#11) including a failing test case in case you want references to be read-only as well. This would also affect other areas, such as reference primers and proxy objects.

malarzm:
Yes, I was aware that references weren't read only but frankly I had no idea what to do with that fact and decided to leave that :) Most possibly if we'd only pass the hint further we'll be perfectly fine, I can try doing so :)

malarzm:
So I started working on read-only going deep but there's (at least) one case we can't easily handle: if you combine `ReferenceOne` with `repositoryMethod` there's no way to pass hints anywhere. I was thinking about automatic detaching of documents BUT it may not be feasible due to [this exception in `UnitOfWork::doPersist`](https://github.com/doctrine/mongodb-odm/blob/a4d5226bd7594d09db4dc2741895c185c021dffd/lib/Doctrine/ODM/MongoDB/UnitOfWork.php#L1690). These are our options:

1. Leave the PR as is but amend the docs with warning about referenced objects
2. Disable inverse sides of references when doing read-only query/hydration (not my favourite)
3. Sort out persisting detached documents

@alcaeus thoughts?

alcaeus:
Any PersistentCollection can store hints, so the hint could be added to a collection by the hydrator. When loading a collection, the DocumentPersister can retrieve these hints and act accordingly: in the case of a referenceMany annotation with repositoryMethod, the cursor created in the DocumentPersister just needs to receive the same UnitOfWorkHints that the collection had - done.

This would require us to change the type check in [DocumentPersister](https://github.com/doctrine/mongodb-odm/blob/master/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php#L812); requiring an ODM cursor (and thus allowing us to add hints) is the right choice here.

malarzm:
Yes, handling `ReferenceMany` should be easy because of having `PersistentCollection` and the fact we're operating on `Cursor`, but take a look [here](https://github.com/doctrine/mongodb-odm/blob/a4d5226bd7594d09db4dc2741895c185c021dffd/lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php#L272-L287) where we're having an inverse side ref one using repository method.

malarzm:
@alcaeus I've added a note in the documentation that references will not be fetched in read-only mode.

alcaeus:
@malarzm Sounds good - in that case, feel free to merge! 👍

PR #1411, opened 2016-05-23
Fix for fatal exception if persisted collection was unserialized

malarzm:
This builds up on @notrix PR #1390 and changes way `typeClass` is obtained to prevent duplicating code

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionTrait.php
+++ b/lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionTrait.php
@@ -103,11 +103,6 @@ trait PersistentCollectionTrait
      */
     private $hints = array();
 
-    /**
-     * @var ClassMetadata
-     */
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionTrait.php
```diff
@@ -317,11 +313,18 @@ public function getMapping()
     /** {@inheritdoc} */
     public function getTypeClass()
     {
-        if (empty($this->typeClass)) {
-            throw new MongoDBException('Specifying targetDocument is required for the ClassMetadata to be obtained.');
+        switch (true) {
+            case (! empty($this->typeClass)):
+                return $this->typeClass;
```
malarzm: Now I'm thinking about this there's no point in remembering `$this->typeClass` (and doing so may lead to edge case if somebody would call `setOwner` with other mapping), I'll remove it later today

PR #1412, opened 2016-05-23
update docs links to mongodb docs

mhor:


docs/en/reference/find-and-update.rst
```diff
@@ -74,7 +74,7 @@ You can also remove a document and return it:
         ->execute();
 
 You can read more about the find and modify functionality on the
-`MongoDB website <http://www.mongodb.org/display/DOCS/findandmodify+Command>`_.
+`MongoDB website <https://docs.mongodb.com/v3.0/reference/method/db.collection.findAndModify/>`_.
```
malarzm: There's no need to pinpoint docs to 3.0, https://docs.mongodb.com/manual/reference/method/db.collection.findAndModify/ link can be used instead

malarzm:
@mhor thanks a bunch for going through all the links! There's one change I'd like to see before merging though (please see comment above), would you mind ammending your commit with it?

docs/en/reference/find-and-update.rst
mhor: @malarzm Done :+1:

malarzm:
Thank you @mhor!

PR #1414, opened 2016-05-26
Fix bug when priming references nested in embedded documents

chapa:
Fix bug that occur when you have a schema like this :
```php
/** @ODM\Document */
class Document
{
    /** @ODM\Id */
    public $id;

    /** @ODM\EmbedOne(targetDocument="Embedded") */
    public $embeddedDocument;
}

/** @ODM\EmbeddedDocument */
class Embedded
{
    /** @ODM\ReferenceMany(targetDocument="Referenced", name="referenced_documents") */
    public $referencedDocuments;
}

/** @ODM\Document */
class Referenced
{
    /** @ODM\Id */
    public $id;
}
```

And you execute a query like this :
```php
$qb = $dm->createQueryBuilder('Document')
    ->field('embeddedDocument.referencedDocuments')->prim(true)
;
$documents = $qb->getQuery()->execute();
```

I think it also occur if you change `@ODM\EmbedOne` to `@ODM\EmbedMany` and/or `@ODM\ReferenceMany` to `@ODM\ReferenceOne`.
To be sure, I test it all in [ReferencePrimerTest.php](1414/files#diff-cc9044aa79f40c245e290f9eb2c471c6R175)

PR #1416, opened 2016-05-29
Fix bug when priming references nested in embedded documents

malarzm:
This PR backports patch from #1414 into `1.0.x` branch. Tests are not carried over because they're built on stuff that is not available in `1.0.x` branch, they will be merged with original PR into `dev-master`


alcaeus:
🚀 LGTM

Issue #1350, opened 2016-01-25
using doctrine to map "raw" data


wgallagher:
We're increasingly seeing the need to map data that's coming from backends that doctrine doesn't support.  For example, dynamodb and REST services.  I was wondering if there is any support in Doctrine for mapping raw PHP Arrays to / from models.  If not, has anyone considered adding it?

malarzm:
@wgallagher possibly there is an exact match - [a raw type](https://github.com/doctrine/mongodb-odm/blob/master/lib/Doctrine/ODM/MongoDB/Types/RawType.php) :) I'm even using it in our application to save raw json values sent from API (and return them later as it's namely for saved search criteria)

> the need to map data that's coming from backends that doctrine doesn't support

What exactly we don't support? Maybe it'll be common enough to add to core

wgallagher:
Support for a REST backend would be very useful to us as would a DynamoDB backend.

For REST, we're replacing some of our mongo mappings with calls to REST services.  So, for example, instead of looking up a mongo record by ID, we're making a rest call with an ID:

http://host:port/api/users/12345

And that returns a json object representing the record.  We'd like to be able to hydrate doctrine models with the json.  In addition, we'd like to be able to take a Doctrine model and serialize it into an array containing a changeset or a complete document.

malarzm:
You can take a look at #1348 but also to answer your questions here:

> We'd like to be able to hydrate doctrine models with the json.

You can already use Hydrators for that given json is converted to an array (https://github.com/doctrine/mongodb-odm/pull/1348#issuecomment-174084190)

> take a Doctrine model and serialize it into an array

That's actually not Doctrine's responsibility, we're persisting stuff, not serializing it :) JMSSerializer is good for that

speedytwenty:
Thought I'd go ahead and add my question to this thread as it is very similar.

I am hydrating a document from data retrieved from a REST API and attempting to persist it to the database. The documents' mappings are identical with the exception of the "_id". The rest api uses it's own id key for which I am using alsoLoadMethods to map the "source id key" to the mongo id (_id). This is working fine, I only mention this as this difference in mapping prevents me from using the UnitOfWork::getOrCreateDocument() method.

I am using the merge() method instead of persist().

The problem I am having is that embedded documents are not being saved to the database. It seems these are "lazily loaded" and not "initialized" after hydration.

[...]

PR #1418, opened 2016-05-31
Adding test for persisting embedded documents from manually hydrated …

speedytwenty:
…documents

In regards to #1350, this PR adds a (failed) test that hydrates a document from raw data, merges the document to the db, and then tests that the embedded documents were actually persisted.

The embedOne field is properly persisted, but the embedMany field is not.

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1418Test.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1418Test.php
@@ -54,63 +54,6 @@ class GH1418Test extends \Doctrine\ODM\MongoDB\Tests\BaseTest
         $this->assertEquals(1, $document->embedMany->count());
         $this->assertEquals('maciej', $document->embedMany->first()->name);
     }
-
-    public function testManualHydrateAndMergeWorkaround()
-    {
-        $document = new GH1418Document;
-        $class = $this->dm->getClassMetadata(get_class($document));
-        $this->dm->getHydratorFactory()->hydrate($document, array(
-          '_id' => 1,
-          'name' => 'maciej',
-          'embedOne' => ['name' => 'maciej'],
-          'embedMany' => [
-              ['name' => 'maciej']
-          ],
-        ), [ Query::HINT_READ_ONLY => true ]);
-
-        foreach ($document->embedMany AS $val) {
-          $this->dm->getUnitOfWork()->getDocumentState($val);
-        }
-
-        $this->dm->merge($document);
-        $this->dm->flush();
-        $this->dm->clear();
-
-        $document = $this->dm->getRepository(__NAMESPACE__.'\GH1418Document')->find(1);
-        $this->assertEquals(1, $document->id);
-        $this->assertEquals('maciej', $document->embedOne->name);
-        $this->assertEquals(1, $document->embedMany->count());
-        $this->assertEquals('maciej', $document->embedMany->first()->name);
-    }
-
-    public function testManualHydrateAndPersistWorkaround()
-    {
-        $document = new GH1418Document;
-        $class = $this->dm->getClassMetadata(get_class($document));
-        $this->dm->getHydratorFactory()->hydrate($document, array(
-          '_id' => 1,
-          'name' => 'maciej',
-          'embedOne' => ['name' => 'maciej'],
-          'embedMany' => [
-              ['name' => 'maciej']
-          ],
-        ), [ Query::HINT_READ_ONLY => true ]);
-
-        foreach ($document->embedMany AS $val) {
-          $this->dm->getUnitOfWork()->getDocumentState($val);
-        }
-
-        $this->dm->persist($document);
-        $this->dm->flush();
-        $this->dm->clear();
-
-        $document = $this->dm->getRepository(__NAMESPACE__.'\GH1418Document')->find(1);
-        $this->assertEquals(1, $document->id);
-        $this->assertEquals('maciej', $document->embedOne->name);
-        $this->assertEquals(1, $document->embedMany->count());
-        $this->assertEquals('maciej', $document->embedMany->first()->name);
-    }
-
 }
 
 /** @ODM\Document */
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1418Test.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1418Test.php
@@ -7,6 +7,29 @@ use Doctrine\ODM\MongoDB\Query\Query;
 
 class GH1418Test extends \Doctrine\ODM\MongoDB\Tests\BaseTest
 {
+    public function testManualHydrateAndPersist()
+    {
+        $class = $this->dm->getClassMetadata(__NAMESPACE__.'\GH1418Document');
+        $document = $class->newInstance();
+        $this->dm->getHydratorFactory()->hydrate($document, array(
+          '_id' => 1,
+          'name' => 'maciej',
+          'embedOne' => ['name' => 'maciej'],
+          'embedMany' => [
+              ['name' => 'maciej']
+          ],
+        ), [ Query::HINT_READ_ONLY => true ]);
+
+        $this->dm->persist($document);
+        $this->dm->flush();
+        $this->dm->clear();
+
+        $document = $this->dm->getRepository(__NAMESPACE__.'\GH1418Document')->findOneById(1);
+        $this->assertEquals(1, $document->id);
+        $this->assertEquals('maciej', $document->embedOne->name);
+        $this->assertEquals(1, $document->embedMany->count());
+    }
+
     public function testManualHydrateAndMerge()
     {
         $class = $this->dm->getClassMetadata(__NAMESPACE__.'\GH1418Document');
@@ -19,6 +42,7 @@ class GH1418Test extends \Doctrine\ODM\MongoDB\Tests\BaseTest
               ['name' => 'maciej']
           ],
         ), [ Query::HINT_READ_ONLY => true ]);
+
         $this->dm->merge($document);
         $this->dm->flush();
         $this->dm->clear();
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1418Test.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1418Test.php
@@ -60,10 +60,10 @@ class GH1418Document
     /** @ODM\Id(strategy="none") */
     public $id;
 
-    /** @ODM\EmbedOne(targetDocument="GH1418Embedded", strategy="set") */
+    /** @ODM\EmbedOne(targetDocument="GH1418Embedded") */
     public $embedOne;
 
-    /** @ODM\EmbedMany(targetDocument="GH1418Embedded", strategy="set") */
+    /** @ODM\EmbedMany(targetDocument="GH1418Embedded") */
     public $embedMany;
 }
 
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1418Test.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1418Test.php
@@ -24,7 +24,7 @@ class GH1418Test extends \Doctrine\ODM\MongoDB\Tests\BaseTest
         $this->dm->flush();
         $this->dm->clear();
 
-        $document = $this->dm->getRepository(__NAMESPACE__.'\GH1418Document')->findOneById(1);
+        $document = $this->dm->getRepository(__NAMESPACE__.'\GH1418Document')->find(1);
         $this->assertEquals(1, $document->id);
         $this->assertEquals('maciej', $document->embedOne->name);
         $this->assertEquals(1, $document->embedMany->count());
@@ -47,7 +47,7 @@ class GH1418Test extends \Doctrine\ODM\MongoDB\Tests\BaseTest
         $this->dm->flush();
         $this->dm->clear();
 
-        $document = $this->dm->getRepository(__NAMESPACE__.'\GH1418Document')->findOneById(1);
+        $document = $this->dm->getRepository(__NAMESPACE__.'\GH1418Document')->find(1);
         $this->assertEquals(1, $document->id);
         $this->assertEquals('maciej', $document->embedOne->name);
         $this->assertEquals(1, $document->embedMany->count());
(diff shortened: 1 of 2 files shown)


malarzm:
Just for the record, could you check if using `persist` instead of `merge` would work?

speedytwenty:
I have tried both merge and persist to no avail.

Should both work? I can add a second test that uses persist if you'd like.

malarzm:
I'd expect `persist` to work so yes please, add another test method with `persist` please. As for `merge` I've frankly never used it in any of my projects ;)

speedytwenty:
Okay, adding a second test for persist.

Also, this is one of the first tests I've contributed so if you notice something that could be better, any and all feedback is welcome.

speedytwenty:
Added the second test which uses persist instead of merge. Still no luck :(

But I can't tell you how much I appreciate the support. The consistent activity which I see you involved with on this project is tremendous. Thank you very much for all your contributions!

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1418Test.php
```diff
@@ -0,0 +1,75 @@
...
+
+use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
+use Doctrine\ODM\MongoDB\Query\Query;
+
+class GH1418Test extends \Doctrine\ODM\MongoDB\Tests\BaseTest
+{
+    public function testManualHydrateAndPersist()
+    {
+        $class = $this->dm->getClassMetadata(__NAMESPACE__.'\GH1418Document');
+        $document = $class->newInstance();
```
malarzm: `new GH1418Document()` is easier and (after all) it's your class so you shouldn't need Doctrine to instantiate it ;)

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1418Test.php
```diff
@@ -0,0 +1,75 @@
...
+          'embedMany' => [
+              ['name' => 'maciej']
+          ],
+        ), [ Query::HINT_READ_ONLY => true ]);
+
+        $this->dm->persist($document);
+        $this->dm->flush();
+        $this->dm->clear();
+
+        $document = $this->dm->getRepository(__NAMESPACE__.'\GH1418Document')->findOneById(1);
```
malarzm: `dev-master` branch is PHP 5.5+ so you can freely use `GH1418Document::class`. Also `findOneById` is doing exactly the same as `find` yet `find` tends to be faster because it first looks up `UnitOfWork` while `findOneById` will always fire a query (it's not the case here but it's a hint in general)

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1418Test.php
```diff
@@ -0,0 +1,75 @@
...
+    }
+}
+
+/** @ODM\Document */
+class GH1418Document
+{
+    /** @ODM\Id(strategy="none") */
+    public $id;
+
+    /** @ODM\EmbedOne(targetDocument="GH1418Embedded", strategy="set") */
```
malarzm: `strategy` has no effect here

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1418Test.php
```diff
@@ -0,0 +1,75 @@
...
+        ), [ Query::HINT_READ_ONLY => true ]);
+
+        $this->dm->persist($document);
+        $this->dm->flush();
+        $this->dm->clear();
+
+        $document = $this->dm->getRepository(__NAMESPACE__.'\GH1418Document')->findOneById(1);
+        $this->assertEquals(1, $document->id);
+        $this->assertEquals('maciej', $document->embedOne->name);
+        $this->assertEquals(1, $document->embedMany->count());
```
malarzm: I'd add an assertion for `$document->embedMany->first()->name` having an expected value

malarzm:
> Also, this is one of the first tests I've contributed so if you notice something that could be better, any and all feedback is welcome.

I've left some minor comments but generally this is a good functional test :)

> But I can't tell you how much I appreciate the support. The consistent activity which I see you involved with on this project is tremendous. Thank you very much for all your contributions!

Thank you sir for kind words, that means a lot!

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1418Test.php
speedytwenty: I had this originally, but it was triggering a Fatal error:

> Trying to get property of non-object

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1418Test.php
speedytwenty: Cool. I'll get rid of it. This was leftover from copying another test as a starting point.

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1418Test.php
malarzm: That still counts for a failing test, once the embedded documents is in place the test will be green and we'll know that beside of a collection having one element, the element is in fact the one we wanted

speedytwenty: Oh, I misread you. Should I keep the test on `..embedMany->count()` and add the test on `..first()->name` or replace the count test with the name test?

speedytwenty: I went ahead and ADDED the second assertion and kept the original assertion on count. Haven't done a lot of testing like this, but it seems that that way the fatal error is avoided. (My thought is that a fatal error in a test impedes other tests from running, which may be okay, but is probably less preferred).

speedytwenty:
Thanks for the feedback. This has been a very valuable exercise for me.

Now if I only knew how to pronounce your name, I'd be praising you out loud!

speedytwenty:
Needed to find a workaround for now, so thought I'd share what I did (in my implementation) in case it triggers any ideas on your side.

I discovered this by accident, I was debugging and later removed the debug code and it stopped persisting the embedMany.

My head is spinning from all the voodoo you do, but when I call `UnitOfWork::getDocumentState($document->embedMany)`, the embedded document then gets persisted properly (obviously not the solution, but perhaps a problem that will be encountered).

However, when the embedded document is empty, a fatal error is thrown. The error that gets thrown differs between embedMany and embedOne.

When embedOne is empty:
> spl_object_hash() expects parameter 1 to be object, null given
> /mongodb-odm/lib/Doctrine/ODM/MongoDB/UnitOfWork.php:1442

When embedMany is empty this error gets thrown:
> Notice: Array to string conversion
> /mongodb-odm/lib/Doctrine/ODM/MongoDB/UnitOfWork.php:949

This may be specific to my workaround, in which case you can disregard and I'll remove the additional test. But I suspect that it may not be specific to my workaround so thought I'd share so you can anticipate it.

I added an additional test which demonstrates my work around but with empty embedded documents which triggers the errors described above.

speedytwenty:
Alright, so I needed a workaround, and I ran into some issues with it but haven't been able to properly reproduce them in the tests, however. I did find that the workaround only works for `merge()` and not for `persist()`, so I added tests which demonstrate this in case it helps distinguish the problem and identify a solution.

With all the voodoo that you do, my head is spinning, but calling `UnitOfWork::getDocumentState($document->embedMany)` leads to a successful merge but not a successful persist.

malarzm:
> if I only knew how to pronounce your name, I'd be praising you out loud!

Haha thanks for making my morning ;) I suck at writing phonetics but [Google](https://translate.google.pl/?ion=1&espv=2&bav=on.2,or.&bvm=bv.123325700,d.bGg&biw=1920&bih=990&dpr=1&um=1&ie=UTF-8&hl=pl&client=tw-ob#pl/en/Maciej) does quite a decent job on reading it ;)

As for the workarounds you put into the tests I have no idea why they'd change anything but I won't be looking into fixing them, I'd rather have "official way" fixed :) Saying that I'm not sure if I'll get to debugging it today but I've marked the issue to be included in 1.1

speedytwenty:
Ah, thanks for pointing out Google (I was way off)! I have since proceeded with the vocal praise, although I'm the only one to hear it!

I included the "workaround" tests for the sole purpose of demonstrating (ahead of an official fix) that merge() somehow works while persist() does not.

I'll remove these work around tests soon, but I'm finally turning in after a long day.

Looking forward to 1.1! Cheers!!!

alcaeus:
Thanks for the test cases. I managed to figure out the issue: after hydration, the PersistentCollection containing the documents mapped using `embedMany` is not marked as dirty. Since it isn't modified before being persisted, `computeAssociationChanges` doesn't pick it up as changed and it is ignored.

A quick workaround would be to mark a PersistentCollection as dirty if it was hydrated using a `READ_ONLY` hint but I'm not sure that would be a good idea. What do you think @malarzm?

malarzm:
@alcaeus there's already an if in [DocumentPersister::loadEmbedManyCollection](https://github.com/doctrine/mongodb-odm/blob/f88851d2269c4b79b34271dedf4f8a0b7f72edfe/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php#L691-L693) so I'd put the logic you're talking about there

alcaeus:
It would have to be somewhere down in the hydrator where the PersistentCollection object is instantiated. Otherwise the problem would only be fixed if the collection is initialized (which in the case of the tests above it isn't since it's never accessed). All in all, the collection needs to be marked as dirty as soon as it's created. If your fine with that I'll create a PR to fix the issue.

malarzm:
Generally there'll be two places to adjust if I'm not mistaken ([this](https://github.com/doctrine/mongodb-odm/blob/f88851d2269c4b79b34271dedf4f8a0b7f72edfe/lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionTrait.php#L160) being the second one) - I'd suggest adding some test cases to check persisting initialized and not initialized collection and see what exactly needs to be changed to fix them all :)

speedytwenty:
Thanks guys for pressing on this! Very much appreciated. Going to have to figure out how to buy each of you a beer by the time 1.1 rolls out!

I've removed the "workaround" test cases. 

@malarzm 
>  I'd suggest adding some test cases to check persisting initialized and not initialized collection and see what exactly needs to be changed to fix them all :)

If you can share some pseudo code or point out some similar code that demonstrates how to perform the initialized/not initialized check, I can try to add the test cases to the functional test.

speedytwenty:
Perhaps `PersistentCollection::isInitialized()` is what you are expecting for the initialized check. If so, just need to understand what to check. Assuming a more robust document model hydrated with varying data?

malarzm:
> A quick workaround would be to mark a PersistentCollection as dirty if it was hydrated using a READ_ONLY hint but I'm not sure that would be a good idea.

@alcaeus: After some time I think this actually might be the way to go BUT what do you think about initializing `PersistentCollection` and then marking it dirty? I'm not sure if otherwise we won't shoot ourselves in the foot as dirty but uninitialized collection will still fail to persist correctly since for such collections we're considering only new elements when building data in `PersistenceBuilder`.

> If you can share some pseudo code or point out some similar code that demonstrates how to perform the initialized/not initialized check, I can try to add the test cases to the functional test.

@speedytwenty: the start for the tests would be similair so:

```
$doc = hydrate([...]);
$doc->embedMany->add(new Embed('...'));
$dm->persist($doc);
$dm->flush();
// refetch data
assertCount(2, $doc->embedMany);
assertSame('...', $doc->embedMany[1]->...);
```

2nd test would be alike but before persisting we'd modify `$doc->embedMany[0]->...` and later assert for the changed value after refetching. Roughly this should be it :)

speedytwenty:
Thanks for the instruction. I'm on the road this afternoon but will see what I can tackle tomorrow morning... approx 16 hours from now.

malarzm:
All right, thanks a bunch!

alcaeus:
I took a stab at fixing this but it would do more bad than good. The collection needs to be initialized and have its snapshot cleared in order for UoW to pick up the documents contained in 1-n associations. The problem with this is that when reading a document in read-only mode, this process would break things if that document is then persisted with the document manager: when writing the document to the database, all documents in the 1-n association will be duplicated in the database (since they will all be picked up as new).

I believe in this case the onus is on developers using this feature to generate documents through the hydrator. The hydrator won't know if it was called because a document was read from the database or because it came from somewhere else.

IMO, there are two solutions for developers:
1. Use hydrator to generate these documents, then initialize collections yourself and clear their snapshots. This would have to "go deep", e.g. needs to traverse the entire document and all embedded documents
2. Use a factory to create a document from external data source. ODM and its hydrators assume that documents being hydrated come from a database, not an external source, so there will always be unwanted side effects.

malarzm:
I think we have another way of making this right (tm): Hydrator should hydrate the data to `ArrayCollection` when in read only mode. It adds a little overhead but also do not messes with `PersistentCollection` which obviously wasn't meant to be (ab)used like this :)

alcaeus:
That still causes the same issue: a document read from the database using `READ_ONLY` would still duplicate all elements in a collection if the document is later managed again (non-persistent collections will always cause an empty snapshot and dirty flag; basically the same as the fix I tried in the hydrator)

speedytwenty:
First, let me reiterate. I had success merging documents with embedMany fields after calling `UnitOfWork::getDocumentState()` on each embedMany field before merging without duplicating data. Sure this was a work around, and I don't know what exactly was happening, but I didn't have to "go deep". I only had to "touch" the top level embedded documents with said method.

With that in mind, and considering what you guys have said, it seems this may be better served as a feature request and not as a bug.

Keep in mind that I may be way off as you guys are obviously much more familiar with the inner workings of the system, but picture this: rather than making this use case work for all hydrations, instead adding either a secondary hydration helper (or merely a functional test) that performs any necessary initialization/massaging that may be necessary and may optionally or necessarily also perform a refresh from any previously persisted data where the changes are then computed accordingly.

This may still be a work around. But I think it's worth taking a step back and considering this as a new feature. Even though I'd hate to postpone this from 1.1 milestone, I'd like to see the "official" solution for this without mucking up the core of the library.

alcaeus:
Thanks for the hint with `getDocumentState`, I'll take a look at why that would fix it later. I also think that using `merge` *could* be supported while `persist` will most likely create issues. I'm currently tracking down a different issue in `merge` but will take a look at this one afterwards.

alcaeus:
So, turns out it wasn't `getDocumentState` that was fixing this but rather the initialisation of the `PersistentCollection` instance which later fixed the behavior in `merge`. I've created a pull request to fix this issue which means you should be able to `merge` a read-only document in order to persist writes to the database.

Note: using `persist` to do this won't work due to the internal workings. Using `merge` for this purpose works and has to be used regardless of whether the document was previously read from the database or not.

speedytwenty:
Closing this as this has been addressed in #1424 

@malarzm Can you tag #1424 for the 1.1 milestone please :)

PR #1419, opened 2016-06-01
Write concern per document

### LINKED ISSUE #1409: Specify writeConcern per document
The document mapping lets you specify a readPreference (still called slaveOkay) for a document, which will be applied to read queries for a given document. However, there is no way to specify a writeConcern per document, just a global one in the DocumentManager configuration.

If specified, the writeConcern for a document would override the writeConcern defined in the global options, allowing you to send unacknowledged writes for a document:
```
/**
  * @ODM\Document(writeConcern=0)
  */
class LogEntry
{
    // ...
```

gruberro:
This PR aims to fix #1409. Please let me know if there is something missing that I can improve or there should be more tests.

I had to move the default options from the `UnitOfWork` to the `DocumentPersister` to ensure it can be determined if the given options are passed through `flush()` or are the system wide defaults.

### DIFF
--- a/docs/en/reference/annotations-reference.rst
+++ b/docs/en/reference/annotations-reference.rst
@@ -401,8 +401,8 @@ Optional attributes:
 -
    writeConcern - Specifies the write concern for this document that overwrites
    the default write concern specified in the configuration. It does not overwrite
-   a write concern given as option to the ``flush`` method when committing your
-   documents.
+   a write concern given as :ref:`option <flush_options>` to the ``flush``
+   method when committing your documents.
 
 .. code-block:: php
 
--- a/docs/en/reference/annotations-reference.rst
+++ b/docs/en/reference/annotations-reference.rst
@@ -398,6 +398,9 @@ Optional attributes:
 -
    requireIndexes - Specifies whether or not queries for this document should
    require indexes by default. This may also be specified per query.
+-
+   writeConcern - Specifies the write concern for this document that overwrites
+   the default write concern specified in the configuration.
 
 .. code-block:: php
 
--- a/docs/en/reference/annotations-reference.rst
+++ b/docs/en/reference/annotations-reference.rst
@@ -400,7 +400,9 @@ Optional attributes:
    require indexes by default. This may also be specified per query.
 -
    writeConcern - Specifies the write concern for this document that overwrites
-   the default write concern specified in the configuration.
+   the default write concern specified in the configuration. It does not overwrite
+   a write concern given as option to the ``flush`` method when committing your
+   documents.
 
 .. code-block:: php
 
--- a/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataFactory.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataFactory.php
@@ -147,6 +147,7 @@ class ClassMetadataFactory extends AbstractClassMetadataFactory
             $class->setLifecycleCallbacks($parent->lifecycleCallbacks);
             $class->setAlsoLoadMethods($parent->alsoLoadMethods);
             $class->setChangeTrackingPolicy($parent->changeTrackingPolicy);
+            $class->setWriteConcern($parent->writeConcern);
             $class->setFile($parent->getFile());
             if ($parent->isMappedSuperclass) {
                 $class->setCustomRepositoryClass($parent->customRepositoryClassName);
@@ -179,10 +180,6 @@ class ClassMetadataFactory extends AbstractClassMetadataFactory
         if ($parent && $parent->isInheritanceTypeSingleCollection()) {
             $class->setDatabase($parent->getDatabase());
             $class->setCollection($parent->getCollection());
-        }
-
-        if ($parent && $class->writeConcern === null) {
-            $class->setWriteConcern($parent->writeConcern);
         }
 
         $class->setParentClasses($nonSuperclassParents);
--- a/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataFactory.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataFactory.php
@@ -180,10 +180,9 @@ class ClassMetadataFactory extends AbstractClassMetadataFactory
             $class->setDatabase($parent->getDatabase());
             $class->setCollection($parent->getCollection());
 
-            if (! $class->writeConcern) {
+            if ($class->writeConcern === null) {
                 $class->setWriteConcern($parent->writeConcern);
             }
-
         }
 
         $class->setParentClasses($nonSuperclassParents);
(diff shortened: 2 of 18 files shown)


lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataFactory.php
```diff
@@ -179,6 +179,11 @@ protected function doLoadMetadata($class, $parent, $rootEntityFound, array $nonS
         if ($parent && $parent->isInheritanceTypeSingleCollection()) {
             $class->setDatabase($parent->getDatabase());
             $class->setCollection($parent->getCollection());
+
+            if (! $class->writeConcern) {
```
malarzm: I think you should be comparing with `null` as `writeConcern` set to `0` will be overridden by parent's setting

malarzm: Also for this setting's propagation only `$parent` is the only required thing, it doesn't need to use single collection inheritance, does it?

alcaeus: That was a suggestion I made, the reasoning behind it being that we shouldn't apply the `writeConcern` defined in the parent class to a different collection. We used `slaveOkay` as an example for this, which is not propagated to child classes either.

That being said, we should probably treat `slaveOkay` the same way as `writeConcern`, regardless of the decision.

tests/Doctrine/ODM/MongoDB/Tests/Mapping/AbstractMappingDriverTest.php
```diff
@@ -36,6 +36,17 @@ public function testDocumentCollectionNameAndInheritance($class)
      * @depends testDocumentCollectionNameAndInheritance
      * @param ClassMetadata $class
      */
+    public function testDocumentLevelWriteConcern($class)
+    {
+        $this->assertEquals(0, $class->getWriteConcern());
```
malarzm: This should be `assertSame`

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataFactory.php
malarzm: Hmmm I personally would expect write concern to be propagated (f.e. I have an abstract class that is supposed to set write concern for all its descendants). If we end leaving it as-is I'd add a note in documentation

tests/Doctrine/ODM/MongoDB/Tests/Mapping/AbstractMappingDriverTest.php
gruberro: With `assertSame` the test will fail for the xml driver (returning strings only). This would require the annotation driver to cast the values if the write concern is `0` or `1`, but not in all other cases. Do you think this is necessary?

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataFactory.php
alcaeus: Alright - in that case, make it so :)

gruberro: Thank you @malarzm for the above hint. Will fix that. Personally I'd also expect to propagate the write concern regardless of the inheritance type... so @alcaeus 1:2 😄

tests/Doctrine/ODM/MongoDB/Tests/Mapping/AbstractMappingDriverTest.php
malarzm: I wanted to avoid false positive if `$class->getWriteConcern()` returns `null`, how about we test for `1` here and ensure that `0` works as expected someplace else?

gruberro: Did your suggested changes! Thx for the advise.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataFactory.php
```diff
@@ -181,6 +181,10 @@ protected function doLoadMetadata($class, $parent, $rootEntityFound, array $nonS
             $class->setCollection($parent->getCollection());
         }
 
+        if ($parent && $class->writeConcern === null) {
```
alcaeus: Since the conditional is only here due to metadata being loaded already, we might as well move this call up before the call to `loadMetadataForClass` where all other inherited options are set on the `ClassMetadata` instance (somewhere along line 150).

tests/Doctrine/ODM/MongoDB/Tests/Functional/DocumentPersisterTest.php
```diff
@@ -344,6 +344,128 @@ public function testPrepareQueryOrNewObjWithDBRefReferenceToTargetDocumentWithHa
...
     }
+
+    /**
+     * @return array
+     */
+    public static function dataProviderTestWriteConcern()
+    {
+        return array(
+            'default' => array(
+                'className' => __NAMESPACE__ . '\DocumentPersisterTestDocument',
```
alcaeus: Since we're working with 5.6 only you can use the `::class` constant here for consistency.

alcaeus:
Looks good to me. Thanks for implementing this and even more so for doing it in time for the upcoming 1.1 release! 🎉

tests/Doctrine/ODM/MongoDB/Tests/Functional/DocumentPersisterTest.php
gruberro: I tried to be consistent with the existing tests in this `TestCase`. I'd suggest to change all occurrences at least within this class?

alcaeus: I saw you went ahead and did just that, thank you!

docs/en/reference/annotations-reference.rst
```diff
@@ -398,6 +398,11 @@ Optional attributes:
 -
    requireIndexes - Specifies whether or not queries for this document should
    require indexes by default. This may also be specified per query.
+-
+   writeConcern - Specifies the write concern for this document that overwrites
+   the default write concern specified in the configuration. It does not overwrite
+   a write concern given as option to the ``flush`` method when committing your
+   documents.
```
gruberro: Should I add also a link to http://doctrine-orm.readthedocs.io/projects/doctrine-mongodb-odm/en/latest/reference/working-with-objects.html#flush-options?

malarzm: It certainly won't hurt :)

alcaeus:
Thanks!

Issue #1408, opened 2016-05-12
$dm->flush() continueOnError not found


Identity-labs:
Hi,

I use mongodb-odm with this configuration
        php5.6
        zf2, 
        mongodb 3.0.11, 
        ext-mongodb 1.1.6

composer :
        "mongodb/mongodb": "1.0.x",
        "alcaeus/mongo-php-adapter": "1.0.x",
        "doctrine/doctrine-mongo-odm-module": "dev-master",

When I use and I have a duplicate entry, my other document are not persisted with this method :
        $db->flush(null, array('continueOnError' => true, 'ordered' => false));
And throw MongoDuplicateKeyException 

Do you have a solution to this problem ?
How to force Upsert ? (http://php.net/manual/fr/class.mongodb-driver-bulkwrite.php)
I think doctrine only insert document if _id exists or auto-generated by custom id generator,
But with persist method we can't specify insert update or upsert

Thx for your help

alcaeus:
> I think doctrine only insert document if _id exists or auto-generated by custom id generator

You are correct. If your document already has an identifier, the document is written to the database using an upsert query. If there is no ID, one is generated and the document is written to the database using an insert.

As for `continueOnError` not working, could you provide a failing test case?

Identity-labs:
Hello

I have created an repo ready to go clone and execute
https://github.com/Identity-labs/mongodb-odm-flush-issue

This exemple show continueOnError fail with the third doc,
And an _id with IdGenerator don't upsert

Identity-labs:
No one have the same issue ?

alcaeus:
Sorry, I took a quick peek and then forgot about this, my bad. I'll take a look at it this afternoon and convert it into a functional test.

alcaeus:
Alright, took a look. From what it looks like, this would be an issue with the driver or none at all, one of the driver guys (ping @jmikola) would have to answer it. It boils down to this code:

```
$client = new MongoClient();
$coll = $client->selectCollection('test', 'duplicate');
$coll->drop();
$coll->insert(['_id' => 1]);
$coll->batchInsert([['_id' => 1], ['_id' => 2]], ['continueOnError' => true]);
var_dump(iterator_to_array($coll->find()));
```

The result of this is:
```
Warning: Uncaught exception 'MongoDuplicateKeyException' with message 'localhost:27017: E11000 duplicate key error collection: test.duplicate index: _id_ dup key: { : 1 }' in php shell code:1
[...]

PR #1422, opened 2016-06-01
Add tests on IdGenerator and upsert with IdGenerator

Identity-labs:
Doctrine don't UPSERT document with _id auto-generated by custom id generator (class extends \Doctrine\ODM\MongoDB\Id\AbstractIdGenerator),

Write 2 tests on Custom Id 
```
@ODM\Id(strategy="CUSTOM", type="string", options={"class"="\Documents\IdGenerator\UserIdGenerator"})
```

- testCustomIdGenerator : OK
- testUpsertCustomIdGenerator : FAILED (duplicate error, need to upsert)

related issue : https://github.com/doctrine/mongodb-odm/issues/1408

edit (@alcaeus): Fence code block.

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/CustomIdTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/CustomIdTest.php
@@ -5,6 +5,7 @@ namespace Doctrine\ODM\MongoDB\Tests\Functional;
 use Documents\Account;
 use Documents\CustomUser;
 use Documents\User;
+use Documents\UserIdGenerator;
 
 class CustomIdTest extends \Doctrine\ODM\MongoDB\Tests\BaseTest
 {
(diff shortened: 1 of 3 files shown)


alcaeus:
As discussed in IRC, this is somewhat expected. Right now, upserts can only be triggered by **manually** setting an identifier. Identifiers that are set by generators won't qualify a document for an upsert.

To work around this, you could write a lifecycle event subscriber that hooks onto `prePersist` and generates an identifier:

```
    /**
     * @param LifecycleEventArgs $args
     */
    public function prePersist(LifecycleEventArgs $args)
    {
        $document = $args->getDocument();

        if (!$document instanceof Document\UpsertInterface) {
            return;
        }

        $documentManager = $args->getDocumentManager();
        $class = $documentManager->getClassMetadata(get_class($document));

        // Generate ID to ensure upsert for document
        if ($class->generatorType !== ClassMetadata::GENERATOR_TYPE_NONE) {
            $idValue = $class->idGenerator->generate($documentManager, $document);
            $idValue = $class->getPHPIdentifierValue($class->getDatabaseIdentifierValue($idValue));
            $class->setIdentifierValue($document, $idValue);
        }
    }
```

I've created #1425 to track this issue for the 2.0 milestone.

Issue #1297, opened 2015-12-04
EmbeddedDocument @id changes after every merge()


websirnik:
I'm seeing a strange behaviour atm. I have a document with embedded document. I want to have an id for every EmbeddedDocument. Ids get generated after the document gets flushed initially. But every time I merge() the document to update it, every EmbeddedDocument gets assigned new ids. This is not what I expect. Is it a bug? If now, how do I keep the same Ids after merge()?

```php
/**
 * @MongoDB\Document
 */
class Post
{
    /* ... */

    /**
     * @MongoDB\EmbedMany(targetDocument="Item")
     */
    protected $items = array();
}

```

```php
/**
 * @MongoDB\EmbeddedDocument
 */
class Item
{
    /**
     * @MongoDB\Id(strategy="AUTO")
     */
    protected $id;
    
    /* ... */
}
```
After this action every Item has new $id:
```php
class PostsController extends FOSRestController
{
    /**
     * [PUT] /posts/{slug} "put_post"
     *
     * @ParamConverter("post", converter="fos_rest.request_body")
     * @Rest\View()
     */
    public function putPostAction(Post $post) {
        
        $dm = $this->get('doctrine_mongodb')->getManager();
        $dm->merge($post);
        $dm->flush();

        return $post;
    }
}
```

malarzm:
@websirnik could you please create a failing test case?

malarzm:
ping @websirnik

malarzm:
This should be fixed in `1.1.0` thanks to #1424

PR #1424, opened 2016-06-02
Fix merging read-only documents with persistent collections

alcaeus:
This builds on top of the tests written in #1418.

When merging previously unmanaged documents, we need to initialize `PersistentCollection` instances if the association is configured to cascade merges. This way, ODM picks up embedded documents that were not previously stored. This can happen when using hydrators with a read-only hint to hydrate data from an external source and then using `merge` to manage the document for a later write to the database.

### DIFF
--- a/docs/en/reference/query-builder-api.rst
+++ b/docs/en/reference/query-builder-api.rst
@@ -249,6 +249,12 @@ Read-Only is not applicable if hydration is disabled.
     implementation, may change in future and will not be considered a BC break
     (will be treated as a feature instead).
 
+.. note::
+
+    To manage a document previously fetched in read-only mode, always use the
+    `merge` method of the DocumentManager. Using `persist` in these cases can
+    have unwanted side effects.
+
 Disabling Hydration
 ~~~~~~~~~~~~~~~~~~~
 
--- a/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
+++ b/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
@@ -1798,36 +1798,32 @@ class UnitOfWork implements PropertyChangedListener
                 $document->__load();
             }
 
-            // Try to look the document up in the identity map.
-            $id = $class->isEmbeddedDocument ? null : $class->getIdentifierObject($document);
+            $identifier = $class->getIdentifier();
+            // We always have one element in the identifier array but it might be null
+            $id = $identifier[0] !== null ? $class->getIdentifierObject($document) : null;
+            $managedCopy = null;
 
-            if ($id === null) {
-                // If there is no identifier, it is actually NEW.
+            // Try to fetch document from the database
+            if (! $class->isEmbeddedDocument && $id !== null) {
+                $managedCopy = $this->dm->find($class->name, $id);
+
+                // Managed copy may be removed in which case we can't merge
+                if ($managedCopy && $this->getDocumentState($managedCopy) === self::STATE_REMOVED) {
+                    throw new \InvalidArgumentException('Removed entity detected during merge. Cannot merge with a removed entity.');
+                }
+
+                if ($managedCopy instanceof Proxy && ! $managedCopy->__isInitialized__) {
+                    $managedCopy->__load();
+                }
+            }
+
+            if ($managedCopy === null) {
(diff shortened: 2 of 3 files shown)


speedytwenty:
Shoot. I tried this and it definitely picks up the embedded documents now. But it looks like the "alsoLoadMethods" are not called on the PersistentCollection(s). Double checking my work to see if it's not an issue on my end.

speedytwenty:
I added an "alsoLoad" property to the embedded document and relevant assertions. I submitted a PR to your fork. Let me know if you think supporting this is possible! In my case I am using alsoLoad to map an id field on the embedded document from the source id key. I could remove the id field and map the ids straight across if need be, but assuming that the alsoLoad stuff is desired as part of this fix.

malarzm:
LGTM :+1:

malarzm:
Actually one thing that came to my mind: if we know `persist` is flawed and `merge` is the way to go, would you mind adding a note to documentation?

alcaeus:
Will do.

speedytwenty:
Thanks again for all your work on this. Unfortunately I'm still hitting an issue and I can't tell if it's on my side or not.

Somehow, UnitOfWork::persistNew() is getting called on an empty embedded document that doesn't exist after hydration. Will see if I can reproduce this in a test and follow up.

alcaeus:
`persistNew` is being called for those documents that are not known during the merge process, as can be seen in the [UnitOfWork](https://github.com/doctrine/mongodb-odm/blob/master/lib/Doctrine/ODM/MongoDB/UnitOfWork.php#L1804..L1831). This will affect:
* All embedded documents
* All documents that don't have an identifier set
* All documents where ODM could not find a document with the same ID in the database

This should not be a problem, since the call to persistNew is a completely empty document (one that has been created using `newInstance` by the `ClassMetadata` instance). Afterwards, all properties from the document being merged (e.g. the one you hydrated yourself) are copied into this new document.

Careful though: `UnitOfWork::merge()` **returns** the managed copy of the document. So, the following will not work:
```
$dm->merge($document);
$document->doSomething();
$dm->flush();
```

All changes after the merge are ignored since you are not working on the managed copy of the document. To fix this, re-assign the result of the merge to the original document:
```
$document = $dm->merge($document);
$document->doSomething();
$dm->flush();
```

speedytwenty:
So I was able to reproduce the problem that I'm having within the tests. I submitted a PR to your fork so you can see exactly what I mean.

The key is that I am using an identifier set within the embedded documents.

**Notice** that the id is properly set after hydration and then somehow missing during `merge()` in UnitOfWork::persistNew()`.

Are identifier sets not intended to be used in embeddedDocuments? We could work around this with a unique index but hoping to be able to use an id here.

malarzm:
> Notice that the id is properly set after hydration and then somehow missing during merge() in UnitOfWork::persistNew().

This reminds me of #1297

> Are identifier sets not intended to be used in embeddedDocuments? We could work around this with a unique index but hoping to be able to use an id here.

In case of normal documents the uniqness of `_id` field (so the field mapped as identified) is guaranteed by Mongo itself but it's not the case for embedded documents - I'd suggest creating an unique index for the field :)

alcaeus:
I won't comment on the usefulness of identifiers in embedded documents and whether (and how) to enforce uniqueness on them. However, I did work that test into the current PR and fixed an issue that occured when merging embedded documents with identifiers that have the NONE strategy set (and thus don't generate their own id during `persistNew`). Thanks for providing me with the test case!

speedytwenty:
WOOHOO! This is now working beautifully for me!!! This has been a blocker and I'm very grateful to finally be past it. Thank you both for all your help on this! 

Any ideas on how I can buy you guys a beer from half way around the world? If not, I'll just donate some funds to the Doctrine project although I regret it won't be as substantial as the support you've provided.

Cheers to 1.1!!

malarzm:
> Any ideas on how I can buy you guys a beer from half way around the world?

I guess that one day we'll run into each other during some conference ;)

speedytwenty:
> I guess that one day we'll run into each other during some conference ;)

Well in that case, we won't limit it to just one beer! But if we run into each other on your side of the pond, you're buying! :P

PR #1427, opened 2016-06-06
Store simple field in mapping

alcaeus:
This fixes a minor BC break introduced in #1349. The value for the `simple` field must be provided in the mapping since software might be relying on it.


lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataInfo.php
```diff
@@ -1299,9 +1299,9 @@ public function mapField(array $mapping)
         if (isset($mapping['simple']) && ($mapping['simple'] === true || $mapping['simple'] === 'true')) {
             $mapping['storeAs'] = ClassMetadataInfo::REFERENCE_STORE_AS_ID;
         }
-        // Remove the "simple" mapping and use "storeAs" in all further logic
-        if (isset($mapping['simple'])) {
-            unset($mapping['simple']);
+        // Provide the correct value for the "simple" field for backwards compatibility
+        if (isset($mapping['storeAs'])) {
+            $mapping['simple'] = $mapping['storeAs'] == ClassMetadataInfo::REFERENCE_STORE_AS_ID;
```
malarzm: `===` :)

PR #1429, opened 2016-06-07
Failing test case of github issue

pilou:
https://github.com/doctrine/mongodb-odm/issues/1428

PR #1430, opened 2016-06-07
Prevent spl_object_hash collisions

malarzm:
This is improved version of #1023 which will remove embedded documents' instances from their "identity map" if that's possible.

@alcaeus @AronSzigetvari @amirmab would you mind running your test suites against this branch? It's based on current `dev-master` so you can run the tests with PHP 7. Offhand #1429 seems to be totally different can of worms as it still fails here, will try looking into that separately.


alcaeus:
Yes, this fixes the issues we were having. Thanks!

alcaeus:
Superseded by #1431.

PR #1434, opened 2016-06-15
Fix best-practices.rst

ryo-utsunomiya:
PHP 7 is not fully unicode-aware yet.
Like Doctrine ORM's docs( http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/best-practices.html#don-t-use-special-characters ),
we should remove a part of "(PHP 7)" .


malarzm:
Thanks @ryo-utsunomiya!

PR #1436, opened 2016-06-16
Ensure generated Document/Repository follow PSR-2

ryo-utsunomiya:
- @see: http://www.php-fig.org/psr/psr-2/#2-2-files
  "All PHP files MUST end with a single blank line."


malarzm:
Thanks @ryo-utsunomiya!

PR #1437, opened 2016-06-18
Fix changeset missing old value for exchanged associations

malarzm:
Found this bug while working on new change sets; in https://github.com/doctrine/mongodb-odm/pull/1339 I was so happy that collections are in the change set that I haven't checked what was in the change set :) By the way fixed the same thing for when embedded document is exchanged.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
+++ b/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
@@ -840,7 +840,11 @@ class UnitOfWork implements PropertyChangedListener
                 $oid2 = spl_object_hash($obj);
 
                 if (isset($this->documentChangeSets[$oid2])) {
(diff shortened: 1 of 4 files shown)


malarzm:
By the way

> Collection won't be in changeset if only one of its element was modified.

from original PR seem to no longer be true however I haven't tested that

alcaeus:
> from original PR seem to no longer be true however I haven't tested that

If that's something we want, I'd add a quick test to it, if not I'd also add a quick test to make sure it isn't so ;)

malarzm:
As much as I admire the idea I'm pretty sure we'll get into troubles soon, for example I have quite bad feelings about including `clear()`ed collection into the changeset or detecting deeper in-collection changes... I'd leave straightening this out to change set overhaul that is yet to come :)

alcaeus:
Sounds like a plan 👍

PR #1438, opened 2016-06-18
Improved Yaml ParseException exception message

speedytwenty:
Presently, if a Yaml mapping file has a parse error, it can be tough to track down as the resulting error message only tells you the line of the error and not the filename.

This PR adds a better exception error message that appends the filepath to the end of the parse exception.

The current PR catches only "\Symfony\Component\Yaml\Exception\ParseException" and perhaps it would be better to catch all exceptions, but then I don't know which exception class to "re-throw". Thoughts?

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Mapping/Driver/YamlDriver.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/Driver/YamlDriver.php
@@ -25,6 +25,7 @@ use Doctrine\ODM\MongoDB\Mapping\ClassMetadata as MappingClassMetadata;
 use Doctrine\ODM\MongoDB\Utility\CollectionHelper;
 use Doctrine\ODM\MongoDB\Mapping\ClassMetadataInfo;
 use Symfony\Component\Yaml\Yaml;
+use Symfony\Component\Yaml\Exception\ParseException;
 
 /**
  * The YamlDriver reads the mapping metadata from yaml schema files.
@@ -340,7 +341,13 @@ class YamlDriver extends FileDriver
      */
     protected function loadMappingFile($file)
     {
-        return Yaml::parse(file_get_contents($file));
+        try {
+            return Yaml::parse(file_get_contents($file));
+        }
+        catch (ParseException $e) {
+            $e->setParsedFile($file);
+            throw $e;
+        }
     }
 
     private function setShardKey(ClassMetadataInfo $class, array $shardKey)


lib/Doctrine/ODM/MongoDB/Mapping/Driver/YamlDriver.php
```diff
@@ -340,7 +340,15 @@ private function parseDiscriminatorField($discriminatorField)
...
     protected function loadMappingFile($file)
     {
-        return Yaml::parse(file_get_contents($file));
+        try
+        {
+            $return = Yaml::parse(file_get_contents($file));
+        }
+        catch (\Symfony\Component\Yaml\Exception\ParseException $e)
+        {
+            throw new \Symfony\Component\Yaml\Exception\ParseException($e->getMessage() . ' File: ' . $file);
```
malarzm: `ParseException` has quite few [other arguments](https://github.com/symfony/symfony/blob/afed2f888de3f0aef743be1cacb76c90e06ce987/src/Symfony/Component/Yaml/Exception/ParseException.php#L35) that should be carried over and also caught exception itself should be passed as `$previous`.

speedytwenty: Ah, good eye! Did you look that up or do you have all this in your head!?!
Maybe this is as easy as: 
```
$e->setParseFile($file);
throw $e;
```

malarzm: I wanted to note that caught exception should be passed as `$previous` and then looked up what's in the exception's constructor to not say something utterly stupid :)

malarzm:
@speedytwenty would you mind squashing your commits?

lib/Doctrine/ODM/MongoDB/Mapping/Driver/YamlDriver.php
```diff
@@ -340,7 +340,14 @@ private function parseDiscriminatorField($discriminatorField)
      */
     protected function loadMappingFile($file)
     {
-        return Yaml::parse(file_get_contents($file));
+        try {
+            $return = Yaml::parse(file_get_contents($file));
+        }
+        catch (\Symfony\Component\Yaml\Exception\ParseException $e) {
```
malarzm: Ah, please put the class into `use` statements, and then squash :)

speedytwenty:
Not having much luck with squashing. Seems to always lead to a merge. Trying to figure it out.

malarzm:
`git rebase -i master` while you're on your branch :)

lib/Doctrine/ODM/MongoDB/Mapping/Driver/YamlDriver.php
```diff
@@ -340,7 +341,14 @@ private function parseDiscriminatorField($discriminatorField)
      */
     protected function loadMappingFile($file)
     {
-        return Yaml::parse(file_get_contents($file));
+        try {
+            $return = Yaml::parse(file_get_contents($file));
```
malarzm: I know I'm nitpicking, but you can just `return ` this instead of assigning to variable?

speedytwenty: Good call!

speedytwenty:
I tried `git rebase -i master` but afterward when I try `git push origin feature_branch` it tells me my local branch is behind the remote branch and forces me to pull, which adds an empty merge commit. Guessing the commits are getting rebased in the wrong order or something... sorry haven't had to rebase much. But anxious to figure this out!

Steveb-p:
`git rebase` is *supposed* to rewrite history, so pushing without specifying `--force-with-lease` won't work. What you did you told git specifically - "abandon old commits, here's new ones that are valid".

See https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase for a good explanation (at least, it was clear for me) ;)

speedytwenty:
@Steveb-p - Thank you! The `--force-with-lease` was exactly what was I missing!

@malarzm Commits squashed, nit picks too :)

malarzm:
@speedytwenty thanks for all the effort! Merging since the tests are green now.

Offhand, @Steveb-p thanks for chiming in :)

PR #988, opened 2014-12-05
DocumentGenerator treating traits in line with ORM

malarzm:
Fixes #297 

This has been ported from ORM PRs:

* doctrine/doctrine2#632
* doctrine/doctrine2#763

Issue #297, opened 2012-05-01
Duplication of generated methods / attributes with Traits


khepin:
When a class is set to use a given trait as for example the one provided by
[DoctrineExtensions](https://github.com/l3pp4rd/DoctrineExtensions/blob/master/lib/Gedmo/Timestampable/Traits/TimestampableDocument.php),
the properties and methods defined in the trait file are re-created in the file 
defining the class even though they already exist in the Trait definition.

For example here's what I get for a comment thread class in a SF project if I use
the `doctrine:mongodb:generate:documents` command:

    <?php
    namespace Xxx\CommentBundle\Document;

    use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
    use Gedmo\Mapping\Annotation as Gedmo;

    /**
     * @ODM\Document
     */
    class Thread
    {
        use \Gedmo\Timestampable\Traits\TimestampableDocument;

        /**
         * @ODM\Id
         */
        private $id;

        /**
         * @ODM\ReferenceMany(targetDocument="Comment")
         */
        private $comments;

        public function __construct()
        {
            $this->comments = new \Doctrine\Common\Collections\ArrayCollection();
        }
        
        /**
         * Get id
         *
         * @return id $id
         */
        public function getId()
        {
            return $this->id;
        }

        /**
         * Add comments
         *
         * @param Xxx\CommentBundle\Document\Comment $comments
         */
        public function addComments(\Xxx\CommentBundle\Document\Comment $comments)
        {
            $this->comments[] = $comments;
        }

        /**
         * Get comments
         *
         * @return Doctrine\Common\Collections\Collection $comments
         */
        public function getComments()
        {
            return $this->comments;
        }
        /**
         * @var date $createdAt
         */
        protected $createdAt;

        /**
         * @var date $updatedAt
         */
        protected $updatedAt;

        /**
         * Set createdAt
         *
         * @param date $createdAt
         */
        public function setCreatedAt($createdAt)
        {
            $this->createdAt = $createdAt;
        }

        /**
         * Get createdAt
         *
         * @return date $createdAt
         */
        public function getCreatedAt()
        {
            return $this->createdAt;
        }

        /**
         * Set updatedAt
         *
         * @param date $updatedAt
         */
        public function setUpdatedAt($updatedAt)
        {
            $this->updatedAt = $updatedAt;
        }

        /**
         * Get updatedAt
         *
         * @return date $updatedAt
         */
        public function getUpdatedAt()
        {
[...]

PR #1439, opened 2016-06-18
Fix hasMethod() and hasProperty() when generating traits.

speedytwenty:
When generating document classes, and using a trait as a mappedSuperclass, the properties and methods defined by the trait get duplicated.

The call to class_exists() in `DocumentGenerator::hasProperty()` and `DocumentGenerator::hasMethod()` fails when using a trait, causing methods to be duplicated in a trait superclass.

This PR solved the issue for me.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Tools/DocumentGenerator.php
+++ b/lib/Doctrine/ODM/MongoDB/Tools/DocumentGenerator.php
@@ -438,7 +438,7 @@ public function <methodName>()
 
     private function hasProperty($property, ClassMetadataInfo $metadata)
     {
-        if ($this->extendsClass() || class_exists($metadata->name)) {
+        if ($this->extendsClass() || class_exists($metadata->name) || trait_exists($metadata->name)) {
             // don't generate property if its already on the base class.
             $reflClass = new \ReflectionClass($this->getClassToExtend() ?: $metadata->name);
 
@@ -461,7 +461,7 @@ public function <methodName>()
 
     private function hasMethod($method, ClassMetadataInfo $metadata)
     {
-        if ($this->extendsClass() || class_exists($metadata->name)) {
+        if ($this->extendsClass() || class_exists($metadata->name) || trait_exists($metadata->name)) {
             // don't generate method if its already on the base class.
             $reflClass = new \ReflectionClass($this->getClassToExtend() ?: $metadata->name);
 


speedytwenty:
Seems to be an issue generating hydrators using traits as mapped superclasses..

malarzm:
> When generating document classes, and using a trait as a mappedSuperclass

Do you mean something like this?

```
/** @ODM\MappedSuperclass */
trait MyTrait {  }
```

Also for the record traits should work just fine in normal scenarios, we had #297 which was later fixed in #988

speedytwenty:
Yes, except with generating from yaml mappings! 

I think this is different from what 297/988 addressed. That seems to detect properties and methods defined by traits that are used, not mappings that are traits themselves. But maybe I'm doing something wrong?

Eg.

```
Document/ThingTrait:
    type: mappedSuperclass
    fields:
        ...
```

```
<?php
namespace Document;
trait ThingTrait
{
```

malarzm:
Traits can not be mapped as anything as they do not exist in runtime, only things defined in the trait are moved into class using it. So unfortunately this isn't supported :)

speedytwenty:
I don't quite understand what you mean. The "runtime" issue seems like it would apply to annotations as well (and mappedSuperclass works with annotations?). This seems really close to working minus an issue in the hyrdator/proxy generation.

In my use case, I have a ton of fields that are re-used across multiple documents. But those documents may already be extending another mappedSuperclass, which is where mapping it as a trait becomes ideal.

I've moved back to using abstract classes, but I might reopen and pursue this as a feature request sometime!

malarzm:
Let me expand my answer a bit:

First thing, according to [docs](http://docs.doctrine-project.org/projects/doctrine-mongodb-odm/en/latest/reference/inheritance-mapping.html#mapped-superclasses) "A mapped superclass is an abstract or concrete class that provides mapping information for its subclasses" and the fact that there can be only one parent (which could be `MappedSuperclass`) is used while [building class metadata](https://github.com/doctrine/mongodb-odm/blob/df6e1ebacb82bd361794c7c691f2186f49a9022b/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataFactory.php#L134-L154). Because of this relationship trait can't be a `MappedSuperclass` as it's not a parent to any class.

Additionally to trait not being a parent, traits does not really exist when your script is executed, behind the scenes parser takes their contents and puts into class using the trait (eventually renaming things according to `use` declaration) - you can think of it as a "copy & paste" in the PHP itself. That's why I said that traits can not be mapped as anything (but properties defined by then can). Also supposing traits could be mapped and that mapping would be carried to a class using the trait, it would lead to dangerous mapping collisions. If one trait says that is `MappedSuperclass` and another says it's `Document` then what should be done?

Summing up, traits are not good candidates for `MappedSuperclass` (nor any other class level mapping). While I understand reasons for using traits with mapped properties we need to find another way of mapping them. I know it's pretty straightforward with annotations (I do that a lot personally in my work project) as you just map the fields in trait and that's all, I guess it's YAML (and possibly XML) that miss the way to re-use once defined mapping (however I have no idea how that could be properly implemented).

speedytwenty:
That helps clear up my misunderstanding. Thanks for elaborating, @malarzm!

PR #1440, opened 2016-06-20
Fix upserts for documents with YAML mapping

### LINKED ISSUE #1435: Warning: Undefined index: strategy in Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php:248
When i updated mongo-odm to 1.1.0 i have got warning "Notice Undefined index: strategy" in Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php:248

It happens when i try to update mapped object:
$this->dm->persist($object);
$this->dm->flush($object);

alcaeus:
Fixes #1435.

This also uncovered a bug where using the increment strategy for ID generation could lead to errors when upserting documents due to a name conflict with strategies.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
@@ -146,7 +146,7 @@ class PersistenceBuilder
                 if ($new === null && $mapping['nullable'] !== true) {
                     $updateData['$unset'][$mapping['name']] = true;
                 } else {
-                    if ($new !== null && $mapping['strategy'] === ClassMetadataInfo::STORAGE_STRATEGY_INCREMENT) {
+                    if ($new !== null && isset($mapping['strategy']) && $mapping['strategy'] === ClassMetadataInfo::STORAGE_STRATEGY_INCREMENT) {
                         $operator = '$inc';
                         $value = Type::getType($mapping['type'])->convertToDatabaseValue($new - $old);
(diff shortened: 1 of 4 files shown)


lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
```diff
@@ -146,7 +146,7 @@ public function prepareUpdateData($document)
                 if ($new === null && $mapping['nullable'] !== true) {
                     $updateData['$unset'][$mapping['name']] = true;
                 } else {
-                    if ($new !== null && $mapping['strategy'] === ClassMetadataInfo::STORAGE_STRATEGY_INCREMENT) {
+                    if ($new !== null && isset($mapping['strategy']) && $mapping['strategy'] === ClassMetadataInfo::STORAGE_STRATEGY_INCREMENT) {
```
malarzm: I believe this line is fixing #1435 but I don't think this is covered by tests you added?

alcaeus: Yeah. I see that now. I'll update the test, thanks!

alcaeus: Test updated - it now checks to ensure IDs are not updated, since all other fields will have a mapping set, regardless of the driver being used.

malarzm:
1 step closer to 1.1.1 :)

PR #1452, opened 2016-06-29
Do not query for referenced documents when collection is initialized

websirnik:
This is in response to the earlier SO discussion: http://stackoverflow.com/a/38100697/257815

My assumption is that referenced documents should not be fetched in the `foreach` loop when getting referenced document ids.

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/ReferencesTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/ReferencesTest.php
@@ -13,9 +13,27 @@ use Documents\Group;
 use Documents\User;
 use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
 use Doctrine\ODM\MongoDB\PersistentCollection;
+use Doctrine\ODM\MongoDB\Tests\QueryLogger;
 
 class ReferencesTest extends \Doctrine\ODM\MongoDB\Tests\BaseTest
 {
+    
+    /**
+	 * @var QueryLogger
+	 */
+	private $ql;
+
+	protected function getConfiguration() {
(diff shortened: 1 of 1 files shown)


tests/Doctrine/ODM/MongoDB/Tests/Functional/ReferencesTest.php
```diff
@@ -13,9 +13,27 @@
...
 class ReferencesTest extends \Doctrine\ODM\MongoDB\Tests\BaseTest
 {
+    
+    /**
+	 * @var QueryLogger
+	 */
+	private $ql;
+
+	protected function getConfiguration() {
+		if (!isset($this->ql)) {
```
malarzm: For the sake of possible future PRs, we're not using 4 spaces instead of tabs

malarzm:
Sorry for long time without an answer. Generally speaking there's nothing much we can do, the query is fired when collection is initialized to avoid one query per document later. Ideally we could populate the collection with proxies and load all the data only when any of them is initialized, but we can't do so right now as we're limited by how `Proxies` are implemented at the moment. I'm going to keep this open tho as a reminder for when proxies initialization would be changed. I'll also update the SO with a workaround.

alcaeus:
This is going to be a huge n+1 problem if we don't make it opt-in. The current behavior is intended - initializing a collection loads all documents in that collection to avoid single document loads as @malarzm mentioned earlier.
The only way to make this work is to have a kind of "lazy" collection, for which the hydrator would generate appropriate proxy objects and mark the collection as initialized. I'm not sure whether that should be the default case or not, but given the performance implications we should be careful changing this behavior in the future.

malarzm:
I don't believe we will get back to this, closing.

PR #1457, opened 2016-07-11
Do not refresh documents when loading sorted collection

### LINKED ISSUE #1456: Document included in sorted ReferenceMany can lose its changes
This is happening because when owning side is initialized, ODM fire a query to get documents if that [document is proxy or collection needs to be sorted](https://github.com/doctrine/mongodb-odm/blob/master/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php#L740). On its own it's not that bad (maybe despite firing a query as #1452 notes), but later fetched data [is aggressively set as current](https://github.com/doctrine/mongodb-odm/blob/master/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php#L773) effectively overwriting not only original data but also removing already calculated change set (which later may not be recalculated).

malarzm:
Fixes #1456


alcaeus:
🚀 Looks good, thanks!

malarzm:
Closing in favour of #1461

PR #1462, opened 2016-07-26
Use decorated methods instead of general ones

### LINKED ISSUE #1455: PersistentCollectionTrait does not call underlying collection for ArrayAccess members
Methods required by `ArrayAccess`, namely: `offsetExists`, `offsetGet`, `offsetSet` and `offsetUnset` do not call the underlying collection's implementation for those methods which can have unexpected consequences if the underlying collection has chosen to override any of these methods.

malarzm:
Fixes #1455 and additionally ensures `isEmpty` method is forwarded when collection is initialized.


alcaeus:
Looks good!

PR #1463, opened 2016-07-26
Use createMock where possible

malarzm:
as `getMock` was deprecated. As a bonus I changed now unnecessary long `getMockBuilder` to `createMock` too

PR #1464, opened 2016-07-27
Use createMock where possible

malarzm:
#1463 for `1.0.x`


malarzm:
PHP versions supported by `1.0.x` makes this patch impossible to apply

PR #1468, opened 2016-07-28
Ensure GridFS collections are dropped properly

gruberro:
This PR aims to ensure also a GridFS collection is dropped when calling `$dm->getSchemaManager()->dropCollections();`. Currently dropping a GridFS collection has no affect and the data remains in there.


alcaeus:
Thanks!

PR #1469, opened 2016-07-29
Update mapping documentation

alcaeus:
Removed a couple of paragraphs from the docs. Serializing private properties from proxy object seems to no longer be a problem, see the following 3v4l snippet: https://3v4l.org/1Ufld

About public properties: while not recommended to be mapped, there is nothing there that wouldn't work. Public properties in proxy objects are unset and lazy loaded through the `__get` and `__set` magic accessors, properly triggering proxy initialization whenever accessed. Thus, I also removed that section from the documentation as it's not correct.


malarzm:
:+1:

Issue #1330, opened 2015-12-31
[RFC] Changesets: array vs. object


alcaeus:
Right now, the changeset for a document is an array containing tuples for each field that was changed: `[ 'field' => ['oldValue', 'newValue'] ]`

Of course, this works well but as evidenced by #1329 it can be difficult to accomodate different objects, in this case collections. We could solve this by introducing changeset classes:
```
interface DocumentChangeset extends ArrayAccess
{
    protected $document; // The document this changeset is for
    protected $metadata; // ClassMetadata object for the document
    protected $fieldChangesets;  // Array of FieldChangeset objects

    public function getDocument();
    public function getMetadata();
    public function getChangedFields();
    public function getFieldChangesets();
    public function isFieldChanged($fieldName);
    public function getFieldChangeset($fieldName);
}

interface FieldChangeset
{
    protected $document;
    protected $metadata;
    protected $mapping;
    protected $oldValue;
    protected $newValue;

    public function getOldValue();
    public function getNewValue();
}
```

By implementing `ArrayAccess` in `DocumentChangeset` we could provide a BC layer to the old tuple format. There is another advantage to this change. Currently embedded documents are listed in the changeset if they have changed, but the old and new values are (in most instances) the same object, so one has to check the DocumentManager for the changeset of the embedded document. With object changesets we could provide a special changeset that provides the embedded document changeset:

```
interface EmbeddedDocumentChangeset extends FieldChangeset
{
    protected $changeset; // DocumentChangeset object for the embedded document in question

    public function getChangeset();
}
```

Last but not least, we could solve the issue mentioned in #1329 by having a `CollectionChangeset` class that provides the known old and new values (which in most cases is the same collection object). This class could provide various helpers to return all changesets for documents in this collection and methods to get all elements that were added or removed from the collection.

```
interface CollectionChangeset extends FieldChangeset
{
    protected $changesets; // DocumentChangeset objects for all elements in this collection

    public function getChangesets();
    public function getAddedElements();
    public function getRemovedElements();
}
```

I'm interested in opinions or ideas - the goal is to make changesets more accessible to developers and this structured way of providing them should help.

malarzm:
Huge :+1: from me as I also wanted to make changesets as objects for some time and seeing that possible in a BC way is simply amazing!

jwage:
[...]

PR #1470, opened 2016-07-30
[2.0] Revamped Changesets

malarzm:
Implements #1330 (except for having metadata available in each change set, this may be added later as might be needed for changes down the line anyway)

BC Breaks in `BCish introduction of ObjectChangeSet`
- `UnitOfWork::getDocumentChangeSet` is now returning `ObjectChangeSet|null`. This will make `empty($changeSet)` checks still valid however all `is_array` or comparing anything to an array will break.

BC Breaks in `Use ObjectChangeSet for modified embedded documents`
- change sets for embedded objects are instances of `ObjectChangeSet` which does not (yet?) provides BC for accessing `[0]` and `[1]` for old and new value respectively.

Open issues:
- [x] Should `ObjectChangeSet` allow access via `[0]` and `[1]` to get old/new embedded document or maybe we should ditch `[0]` and `[1]` access altogether? - No, removed
- [ ] Documentation (for this we'd finally need to make a switch between versions ;) )


lib/Doctrine/ODM/MongoDB/ChangeSet/ChangeSetCalculator.php
```diff
@@ -0,0 +1,38 @@
...
+{
+    /**
+     * Calculates change set for passed document basing on passed original data.
+     * 
+     * @param object $document
+     * @param ClassMetadata $class
+     * @param array $originalData
+     * @return ObjectChangeSet
+     */
+    public function calculate($document, ClassMetadata $class, $originalData = []);
```
alcaeus: How about `calculateChangeSet`?

lib/Doctrine/ODM/MongoDB/ChangeSet/CollectionChangeSet.php
```diff
@@ -0,0 +1,139 @@
...
+     *
+     * @param ObjectChangeSet $changeSet
+     */
+    public function registerChange(ObjectChangeSet $changeSet)
+    {
+        $this->changes[spl_object_hash($changeSet->getObject())] = $changeSet;
+    }
+
+    /** {@inheritdoc} */
+    public function getNewValue()
```
alcaeus: IMO, implementing `ChangedValue` may not be wrong here but it might be misleading. If I can call `getOldValue` and `getNewValue` I'd expect different results. In my opinion, we could return the snapshot stored in the `PersistentCollection` object for `getOldValue` and return the new data in `getNewValue`, but the way it is now is just confusing. Another alternative might be to do away with `getOldValue` and `getNewValue` in this class and only have a `getCollection` method which returns the collection object which in turn can allow accessing the snapshot if desired.

lib/Doctrine/ODM/MongoDB/ChangeSet/ObjectChangeSet.php
```diff
@@ -0,0 +1,160 @@
...
+    /** ChangedValue implementation */
+
+    /** {@inheritdoc} */
+    public function getNewValue()
+    {
+        return $this->object;
+    }
+
+    /** {@inheritdoc} */
+    public function getOldValue()
```
alcaeus: Similar to `CollectionChangeSet`, having `getOldValue` and `getNewValue` seems confusing here, since they will return the same value. I think doing away with these here would make sense.

lib/Doctrine/ODM/MongoDB/ChangeSet/ChangeSetCalculator.php
malarzm: IMO it's long and since interface is already called `ChangeSetCalculator` it's redundant

lib/Doctrine/ODM/MongoDB/ChangeSet/CollectionChangeSet.php
malarzm: I didn't like this much either but implementing interface removes bunch of `if`s checking `ChangedValue` vs `CollectionChangeSet` that would be scattered all over the project otherwise

lib/Doctrine/ODM/MongoDB/ChangeSet/ObjectChangeSet.php
malarzm: Same reasoning as for https://github.com/doctrine/mongodb-odm/pull/1470#discussion_r72902141

malarzm: Also I'm open to other ideas, just need them ;)

alcaeus: Gave it some thought, this would be different if replacing an entire embedded document or changing the reference itself - in that case old/new will be different objects. Question is, how does the ChangeSetCalculator handle those cases for embedded documents? Will I have `old !== new` but still get a pretty changeset?

lib/Doctrine/ODM/MongoDB/ChangeSet/CollectionChangeSet.php
alcaeus: You're right - even though in most cases an `instanceof` check would still be required in order to use the changeset methods specific to collections.

lib/Doctrine/ODM/MongoDB/ChangeSet/CollectionChangeSet.php
```diff
@@ -0,0 +1,139 @@
...
+        // 2. if change set represents references then this returns empty array
+        return array_values($this->changes);
+    }
+
+    /**
+     * Gets list of objects removed from collection.
+     *
+     * @return object[]
+     */
+    public function getDeletedObjects()
```
alcaeus: If the collection was cleared using `clear()`, this method will return an empty list even though the entire collection is removed. Do we need to expose an `isCleared` method to handle that case?

malarzm: At this moment cleared collection behaves exactly like it previously did (that is clearing snapshots). It might be feasible to turn things around however it's pretty easy to combo custom collection (like [this](https://github.com/malarzm/collections/blob/master/lib/Malarzm/Collections/Mixin/Diffable.php) one) with `postCollectionLoad` for own change-tracking collection (may do a cookbook post for the documentation about it).

lib/Doctrine/ODM/MongoDB/ChangeSet/ObjectChangeSet.php
malarzm: Right now exchanging embedded document (or whole collection) warrants a `FieldChanged` object in a change set. Pretty changeset is available upon further asking :)

alcaeus: Pretty please? Would be nice to be able to do this: `$objectChangeSet->getChange('embeddedField')->getChanges()`.

malarzm: Hah by "Pretty changeset is available upon further asking :)" I meant asking UnitOfWork but I think I may have an idea ;)

PR #1471, opened 2016-07-30
[2.0] Remove deprecated annotations

malarzm:
- [x] do we still need [DoctrineAnnotations.php](https://github.com/doctrine/mongodb-odm/blob/c515e77627e51ae10e35b6a9ba7caa57390b67f3/lib/Doctrine/ODM/MongoDB/Mapping/Annotations/DoctrineAnnotations.php) in 2016?
- [x] adjust documentation (again, we need a version switch)


alcaeus:
> do we still need DoctrineAnnotations.php?

Not sure, but at least the Bundle seems to load the annotations automatically. Personally I'd prefer if we could do without.

docs/en/tutorials/getting-started.rst
```diff
@@ -6,8 +6,8 @@ domain model in a non-interfering way. Non-relational or no-sql
 databases like MongoDB give you flexibility of building data store
 around your object model and not vise versa. You can read more on the
 initial configuration and setup in :doc:`Introduction to MongoDB Object
-Document Mapper <../reference/introduction>`. This section will give you a basic
-overview of what could be accomplished using Doctrine MongoDB ODM.
+    Document Mapper <../reference/introduction>`. This section will give you a basic
```
alcaeus: Spacing seems off here.

PR #1472, opened 2016-07-30
[1.1] Change field annotations

malarzm:
Found some leftovers when doing #1471


malarzm:
Argh this should go into `dev-1.1.x`

PR #1476, opened 2016-07-30
[2.0] Remove requireIndexes and stuff thereto related

malarzm:
Explaination why I'm doing this will be given in a follow up PR introducing deprecations, I just needed to know what will be removed as a result :) This, again, needs documentation updates


alcaeus:
+14, -655. That's the kind of PR I like. 👍

malarzm:
Now includes even more removals thanks to nuking relevant pieces of docs!

PR #812, opened 2014-02-26
[WIP] Compound Indexes

malarzm:
Fixes #761 ~~, depends on https://github.com/doctrine/mongodb/pull/154~~

**Exceptions** I have left Query::getUnindexedFields untouched and adjusted Exception Message (instead of showing unindexed fields there's a list of used ones. I was thinking about comparing existing indexes with fields from query and suggesting developer fields to add/remove to query to match index requiring least changes but finally gave up on this - however if you think it's good idea I'll implement it)

TODO:
- [x] adjust index checking
- [x] tests (without getUnindexedFields thing yet)
- [x] adjust thrown exceptions
- [x] add check for [Sort Order](http://docs.mongodb.org/manual/core/index-compound/#sort-order)
- [x] document CompoundIndex
- [x] document allowLessEfficientIndexes in @Document annotation

AFTER REVIEW:
- [x] remove ``allowLessEfficientIndexes`` feature (including docs)
- [ ] get index information from ``ClassMetadata`` instead of DB (code & docs)
- [ ] consider text and geospatial indexes
- [ ] consider index intersection (https://github.com/doctrine/mongodb-odm/pull/812#issuecomment-45268453)

cc @dwieeb

Issue #761, opened 2014-01-15
Compound index order not respected


dwieeb:
Given [this explanation of how mongo uses compound indexes](http://docs.mongodb.org/manual/core/index-compound/#prefixes), a Doctrine\ODM\MongoDB\MongoDBException should be thrown with message "Cannot execute unindexed queries on..." if I attempt to do this:

```php
$item = $dm->getRepository('Item')
    ->createQueryBuilder()
    ->field('location')->equals('something')
    ->getQuery()
    ->execute();
```

or

```php
$item = $dm->getRepository('Item')
    ->createQueryBuilder()
    ->field('stock')->equals('something')
    ->getQuery()
    ->execute();
```

or

```php
$item = $dm->getRepository('Item')
    ->createQueryBuilder()
    ->field('location')->equals('something')
    ->field('stock')->equals('something')
    ->getQuery()
    ->execute();
```

And does not.
(It does of course for fields that are not indexed.)

malarzm:
I will have to revisit #812 but this is generally something bigger to be built after 1.0 given pending changes in PR

malarzm:
For the moment being I'm more in favour of dropping requireIndexes feature in 2.0 so I'd rather declare it's broken and won't be fixed.

PR #1477, opened 2016-07-30
[1.2] Deprecate requireIndexes and stuff thereto related

malarzm:
As promised, follow up to #1476. Right now the feature is more broken than working thus giving developer false sense of security. To name few things that are wrong:

1. [Compound index order is not respected](https://github.com/doctrine/mongodb-odm/issues/761)
2. [Index intersection](https://docs.mongodb.com/manual/core/index-intersection/) added in Mongo 2.6 is not considered
3. [`$or` clauses can use own indexes](https://docs.mongodb.com/manual/reference/operator/query/or/#or-clauses-and-indexes) and that's not considered

I even took a stab at some of things listed above in #812 - it reached +1.2k and still was wrong. In my opinion all this adds unnecessary complexity and things to maintain, especially if Mongo would again introduce some new index-related feature making simple true/false switch not enough.

### DIFF
--- a/docs/en/reference/annotations-reference.rst
+++ b/docs/en/reference/annotations-reference.rst
@@ -424,6 +424,9 @@ Optional attributes:
         //...
     }
 
+.. note::
+    Requiring Indexes was deprecated in 1.2 and will be removed in 2.0.
+
 @EmbedMany
 ----------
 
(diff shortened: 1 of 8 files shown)


alcaeus:
Absolutely agree on this one 👍

PR #1478, opened 2016-07-30
[2.0] Remove DiscriminatorField's name and fieldName

malarzm:
PR deprecating `name` and fixing our test suite will be sent soon (travis will be green again after #1479 is merged and this rebased)


alcaeus:
I assume this is because the discriminator field is not hydrated anyways? I've sometimes wondered why that is and thought it would be cool if the discriminator field value would be available in the document as well if we want. What do you think?

malarzm:
I believe this was done to make discriminator transparent (to keep people away from modifying, or trying to modify, it I reckon?) - it'd be good if @jmikola or @jwage would shed some light if they recall :) . Personally I prefer `instanceof` check so I'm not sure if there's value in exposing the field :)

alcaeus:
That would make sense - it was in a previous project where it would have been helpful. Either way, let's remove it since it wasn't used anyways.

PR #1480, opened 2016-07-30
[2.0] Remove simple references leftovers

malarzm:
We need to decide whether we still want to call references using `storeAs="id"` "simple" or not (I would be in favour of doing so, feels more natural than "references stored as ids" however I may just be used to "simple references" term)


alcaeus:
I've started calling them "ID references", opposed to "DBRef references". Since there's no more mention to "simple" in the code, I'd like to avoid that term in the future since it might make it more complex to understand for newbies.

coudenysj:
I like the @alcaeus approach :)

malarzm:
Changed "simple references" occurrences to "ID references" as suggested :+!:

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH499Test.php
```diff
@@ -1,68 +1,68 @@
...
+        $a->addRef($b);
+        $a->addRef($c);
+
+        $this->dm->persist($a);
+        $this->dm->persist($b);
+        $this->dm->persist($c);
+        $this->dm->flush();
+        $this->dm->clear();
+
+        $collection = $this->dm->getDocumentCollection(__NAMESPACE__ . '\GH499Document');
```
alcaeus: You can use the `GH499Document::class` here.

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH499Test.php
```diff
@@ -1,68 +1,68 @@
...
+
+        $a = $collection->findOne(array('_id' => new \MongoId($a->getId())));
```
alcaeus: Short array syntax

tests/Doctrine/ODM/MongoDB/Tests/Mapping/Driver/fixtures/yaml/TestDocuments.User.dcm.yml
```diff
@@ -1,44 +1,44 @@
...
-    groups:
-      targetDocument: Documents\Group
-      cascade: all
-  lifecycleCallbacks:
-    prePersist: [doStuffOnPrePersist]
-    postPersist: [doStuffOnPostPersist, doOtherStuffOnPostPersist]
-  alsoLoadMethods:
-    doStuffOnAlsoLoad: unmappedField
-
+TestDocuments\User:
```
alcaeus: I see Windows line endings here.

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH499Test.php
malarzm: I'm more wondering why is this marked as an addition, it should only be mapping changed...

alcaeus: Looking at the [diff without whitespaces](https://github.com/doctrine/mongodb-odm/pull/1480/files?w=0) hides the lines, so I assume there were changes to the line endings.

malarzm: Funny, the only thing that happened to this files was a rebase, will revisit my git settings and fix it :+1:

malarzm:
@alcaeus I've fixed CRLF mess that I introduced

> I think the changes to the mapping in tests (changing simple to storeAs) could be backported to the master branch, what do you think?

I've split commits into fixing test mappings and removing leftovers for cherry picking 👍

PR #1481, opened 2016-08-01
Fix query preparation when querying for referenced object

alcaeus:
This PR fixes an issue introduced with #1363 when running queries like `$dm->getRepository(User::class)->findOneBy(['account' => $account]);`, with `$account` being a referenced document. When using the `references` method of the query builder, the query array looks like this:
```
[
    'account.$id' => MongoId(...)
]
```

When using `includesReferenceTo` which is made for `@ReferenceMany` mappings, the result looks like this:
```
[
    'account' => [
        '$elemMatch' => [
            '$id' => MongoId(...)
        ]
    ]
]
```

However, the implementation in the DocumentPersister always yielded a full DBRef object:
```
[
    'account' => [
        '$ref' => 'accounts',
        '$id' => MongoId(...)
        '$db' => 'mydatabasename'
    ]
]
```

This caused issues when forcing queries to be covered by indexes (through the `notablescan` option in MongoDB), since it would require two different kinds of indexes on a reference: one on the entire DBRef object to cover queries like created through the DocumentPersister and another index on the `$id` field of the reference to cover queries created by the query builder in combination with `references`.

This solution tries to apply the same logic applied by the query builder to the documents in question. The only exception are `@ReferenceOne` fields without a target document set. These would produce the following query array:
```
[
    'account.$ref' => 'accounts',
    'account.$id' => MongoId(...),
    'account.$db' => 'mydatabasename'
]
```

This is not possible at this time since the DocumentPersister always assumes that a single field has exactly one value. For now, the PR includes a failing test case for this, as the DocumentPersister excludes `@ReferenceOne` fields without targetDocument set.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -908,9 +908,9 @@ class DocumentPersister
      */
     public function prepareFieldName($fieldName)
     {
-        list($fieldName) = $this->prepareQueryElement($fieldName, null, null, false);
+        $fieldNames = $this->prepareQueryElement($fieldName, null, null, false);
 
-        return $fieldName;
+        return $fieldNames[0][0];
     }
 
     /**
@@ -991,11 +991,12 @@ class DocumentPersister
                 continue;
             }
 
-            list($key, $value) = $this->prepareQueryElement($key, $value, null, true);
-
-            $preparedQuery[$key] = is_array($value)
-                ? array_map('\Doctrine\ODM\MongoDB\Types\Type::convertPHPToDatabaseValue', $value)
-                : Type::convertPHPToDatabaseValue($value);
+            $preparedQueryElements = $this->prepareQueryElement($key, $value, null, true);
+            foreach ($preparedQueryElements as list($preparedKey, $preparedValue)) {
+                $preparedQuery[$preparedKey] = is_array($preparedValue)
+                    ? array_map('\Doctrine\ODM\MongoDB\Types\Type::convertPHPToDatabaseValue', $preparedValue)
+                    : Type::convertPHPToDatabaseValue($preparedValue);
+            }
         }
 
         return $preparedQuery;
@@ -1011,7 +1012,7 @@ class DocumentPersister
      * @param mixed $value
      * @param ClassMetadata $class        Defaults to $this->class
      * @param boolean $prepareValue Whether or not to prepare the value
-     * @return array        Prepared field name and value
+     * @return array An array of tuples containing prepared field names and values
      */
     private function prepareQueryElement($fieldName, $value = null, $class = null, $prepareValue = true)
     {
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -1036,7 +1036,20 @@ private function prepareQueryElement($fieldName, $value = null, $class = null, $
 
             if (! empty($mapping['reference']) && is_object($value) && ! ($value instanceof \MongoId)) {
                 try {
-                    return array($fieldName, $this->dm->createDBRef($value, $mapping));
+                    $dbRef = $this->dm->createDBRef($value, $mapping);
+
+                    if (! is_array($dbRef)) {
```
malarzm: I had to stop for a moment and think about why such a check is here, the original check on strategy is way more readable. Also it will prevent us from hitting an edge case when id might be an array

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -1036,7 +1036,20 @@ private function prepareQueryElement($fieldName, $value = null, $class = null, $
...
+                        return array($fieldName, $dbRef);
+                    } elseif ($mapping['type'] == 'many') {
+                        $keys = ['$ref' => true, '$id' => true, '$db' => true];
+                        if (isset($mapping['targetDocument'])) {
```
malarzm: Don't we need [this](https://github.com/doctrine/mongodb-odm/blob/c515e77627e51ae10e35b6a9ba7caa57390b67f3/lib/Doctrine/ODM/MongoDB/Query/Expr.php#L84) check too?

malarzm:
As a starter I'm sorry for introducing the bug, I admit that I haven't thought about indexes at all back then :)

> This is not possible at this time since the DocumentPersister always assumes that a single field has exactly one value.

If this is only issue of what's `prepareQueryElement` is returning (and how) it could be _fairly_ easy (found 3 places using the method) to rework it so it could return an array of tuples (or `["field.$id" => abc, "field.$ref" => xyz]` array) - what do you think?

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -1036,7 +1036,20 @@ private function prepareQueryElement($fieldName, $value = null, $class = null, $
 
             if (! empty($mapping['reference']) && is_object($value) && ! ($value instanceof \MongoId)) {
                 try {
-                    return array($fieldName, $this->dm->createDBRef($value, $mapping));
+                    $dbRef = $this->dm->createDBRef($value, $mapping);
+
+                    if (! is_array($dbRef)) {
+                        return array($fieldName, $dbRef);
+                    } elseif ($mapping['type'] == 'many') {
```
malarzm: This should be `===`

alcaeus:
> it could be fairly easy to rework it so it could return an array of tuples
I thought about that as well - just wasn't sure which route we want to take. I'll have to see how we can accomodate that without creating a weird return type, but if at all possible I'll do exactly that. Thanks for the input! 👍

alcaeus:
@malarzm udpated PR as per your suggestion. `prepareQueryElement` now returns an array of key-value tupels so that more than one query element can be returned.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -1189,12 +1190,16 @@ private function prepareQueryElement($fieldName, $value = null, $class = null, $
                 ? $this->dm->getClassMetadata($targetMapping['targetDocument'])
                 : null;
 
-            list($key, $value) = $this->prepareQueryElement($nextObjectProperty, $value, $nextTargetClass, $prepareValue);
+            $fieldNames = $this->prepareQueryElement($nextObjectProperty, $value, $nextTargetClass, $prepareValue);
+
+            return array_map(function ($preparedTupel) use ($fieldName) {
```
malarzm: Nitpicking, but you have a typo in var name ;)

alcaeus: It's not a typo, it's German! 😁

malarzm: So the germanization of Doctrine begins, one variable at a time? :P

malarzm:
Thanks \o/

PR #1484, opened 2016-08-05
[2.0] Bump PHP to 7.0

malarzm:
This may (and probably) will be bumped more, but for now let's save good ol' Travis some cycles


malarzm:
Uh oh almost merged this into `dev-master`

PR #1486, opened 2016-08-08
Fixed find-and-update.rst example

altitude:
Fixed example for clarity. `->field('in_progress')->set(true)` is actually part of the update, not the find.


malarzm:
Thanks @altitude!

PR #1496, opened 2016-09-26
PHP5.3 compatibility

kadala:
Fixed version doctrine/mongodb < 1.3 compatible to PHP5.3. version mongodb  1.3 is not compatible with PHP 5.3

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
         "doctrine/cache": "~1.0",
         "doctrine/inflector": "~1.0",
         "doctrine/instantiator": "~1.0.1",
-        "doctrine/mongodb": "~1.2"
+        "doctrine/mongodb": "1.2.*"
     },
     "require-dev": {
         "phpunit/phpunit": "~4.8|~5.0",


alcaeus:
This change is not necessary and counter-productive. Composer will do this kind of version resolution for you. It will always pick a set of dependencies that fill all constraints. You can see this in the travis build for the 1.0.x branch. The build for PHP 5.3 [installs doctrine/mongodb 1.2.2](https://travis-ci.org/doctrine/mongodb-odm/jobs/148221180#L206), while the build for PHP 5.6 [installs doctrine/mongodb 1.3.0](https://travis-ci.org/doctrine/mongodb-odm/jobs/148221183#L207). There is no need to arbitrarily limit the possible versions just because of a PHP version constraint. That's why we use Composer.

kadala:
thank you @alcaeus  this is resolved we change local version of php

alcaeus:
Just to understand your problem: your remote (production) system used PHP 5.3 but your development systems (on which you ran `composer update`) are running a newer PHP version?

kadala:
yes I use development php5.5 but PHP5.3 on the server production. I launch on  development  composer update in development .

Thks @alcaeus

PR #1498, opened 2016-09-27
Fix/1497

akomm:
Failing test (1. commit) and fix (2. commit)

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionTrait.php
+++ b/lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionTrait.php
@@ -540,6 +540,7 @@ trait PersistentCollectionTrait
         }
 
         if ($this->isOrphanRemovalEnabled()) {
+            $this->initialize();
             foreach ($this->coll as $element) {
                 $this->uow->scheduleOrphanRemoval($element);
             }
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/OrphanRemovalTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/OrphanRemovalTest.php
@@ -119,6 +119,31 @@ class OrphanRemovalTest extends \Doctrine\ODM\MongoDB\Tests\BaseTest
         $this->assertNull($this->getProfileRepository()->find($profile2->id), 'Profile 2 should have been removed');
     }
 
+    public function testOrphanRemovalOnReferenceManyUsingClearUninitialized()
+    {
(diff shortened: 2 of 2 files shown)


alcaeus:
Thanks for the PR! Since this bug is already present in 1.0.x, it needs to target the 1.0.x branch, not master. I could change the target branch of the PR, but due to the different branch ancestor the PR wouldn't come out clean. You can either create a new PR based on the 1.0.x branch or manually reset your branch to 1.0.x before re-applying the commits and then force-pushing to the branch, after which we can change the target branch of the PR. If this is too complicated, you can also enable this checkbox in the PR sidebar and I'll update the branch for you: 
image

lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionTrait.php
```diff
@@ -540,6 +540,7 @@ public function clear()
         }
 
         if ($this->isOrphanRemovalEnabled()) {
+            $this->initialize();
```
Steveb-p: It still won't execute due to the `$this->initialized` you pointed out earlier. At first glance I would just remove `$this->initialized` from `if`, I did not read further down the code to be sure.

malarzm: If collection is not initialized then `if` statements will evaluate to `false` thus reaching this point

Steveb-p: Damn, true ;)

akomm: I'v made two commits so you can see the test failing and then not failing any more with the fix commit. In the bug-case `$this->initialized` is false, so the code IS executed :). Removing the `$this->initialized` check & not inserting the `$this->initialize()` line cause the test to fail again. It looks like `$this->isEmpty()` is not (implicitly) triggering initialization either.

akomm:
I'v not reset & reapplied it from 1.1.x. I'v also made a PR from 1.0.x

lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionTrait.php
alcaeus: > It looks like `$this->isEmpty()` is not (implicitly) triggering initialization either.

You are correct. `isEmpty` uses the `count` method defined in `PersistentCollectionTrait` if the collection is not initialized, which again does not initialize the collection (because it doesn't need to).

PR #1504, opened 2016-10-07
Add changelog for 1.1.2

alcaeus:


malarzm:
You need to change https://github.com/doctrine/mongodb-odm/blob/master/docs/en/conf.py#L50 too :)

Issue #1509, opened 2016-10-12
Method convertToPHPValue does not get called for Custom Types.


mgonzalezbaile:
I think this is a recurrent issue (I've seen similar already opened) so I wanted to propose an approach to make the `convertToPHPValue` method work without modifying anything else in the rest of the code.

I have created the PR #1510 modifying the test class `CustomTypeTest` to show how it works.

To avoid using reflection to instantiate `Type` I have pushed PR #1513.

This allows to test the custom hydrators (custom types) properly, both hydrate (convertToPHPValue) and extract (convertToDatabaseValue) methods.

malarzm:
Closing as #1513 was merged

PR #1510, opened 2016-10-12
Add CustomType class to let use convertToPHPValue method instead of closureToPHP

mgonzalezbaile:
Issue #1509

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Types/CustomType.php
+++ b/lib/Doctrine/ODM/MongoDB/Types/CustomType.php
@@ -1,0 +1,19 @@
+<?php
+
+namespace Doctrine\ODM\MongoDB\Types;
+
+class CustomType extends Type
+{
+    /**
+     * @return string Redirects to the method convertToPHPValue from child class
+     */
+    final public function closureToPHP()
+    {
+        $fqcn = get_class($this);
+
+        return sprintf('
+            $reflection = new \ReflectionClass("%s");
+            $type = $reflection->newInstanceWithoutConstructor();
+            $return = $type->convertToPHPValue($value);', $fqcn);
+    }
+}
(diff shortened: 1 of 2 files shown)


tests/Doctrine/ODM/MongoDB/Tests/Functional/CustomTypeTest.php
```diff
@@ -62,37 +63,28 @@ public function convertToDatabaseValue($value)
...
     // Note: this method is never called
     public function closureToMongo()
     {
         return '$return = array_map(function($v) { if ($v instanceof \DateTime) { $v = $v->getTimestamp(); } else if (is_string($v)) { $v = strtotime($v); } return new \MongoDate($v); }, $value);';
     }
-
-    // Note: this code ends up in the generated hydrator class (via HydratorFactory)
-    public function closureToPHP()
-    {
-        return '$return = array_map(function($v) { if ($v instanceof \MongoDate) { $date = new \DateTime(); $date->setTimestamp($v->sec); return $date; } else { return new \DateTime($v); } }, $value);';
```
mgonzalezbaile: Just extracted the code from quotes and put it inside `convertToPHPValue` while keeping the test green.

lib/Doctrine/ODM/MongoDB/Types/CustomType.php
```diff
@@ -0,0 +1,19 @@
...
+    /**
+     * @return string Redirects to the method convertToPHPValue from child class
+     */
+    final public function closureToPHP()
+    {
+        $fqcn = get_class($this);
+
+        return sprintf('
+            $reflection = new \ReflectionClass("%s");
+            $type = $reflection->newInstanceWithoutConstructor();
```
malarzm: I think that instead of using reflection we could add a method that would get `Type` instance by its FQCN

mgonzalezbaile: If I am not missing anything, in order to get the instance of `Type` I can do it by:
- constructor: which is defined as `final private` in the base class.
- factory method: current factory methods instantiate `Type` from the custom type identifier (key in `$typesMap` array from `Type` class). I don't see a way to get the custom type identifier so I can make use of any factory method.

Am I right?

mgonzalezbaile: I have pushed PR #1513 avoiding reflection to instantiate `Type`, let me know your thoughts.

malarzm:
Closing in favour of #1513

PR #1513, opened 2016-10-15
Add CustomType class to let use convertToPHPValue method instead of closureToPHP (no reflection)

mgonzalezbaile:
Issue #1509

Notes: Per suggested by @malarzm here it is another approach from PR #1510 avoiding reflection to instantiate a `Type`.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Types/CustomType.php
+++ b/lib/Doctrine/ODM/MongoDB/Types/CustomType.php
@@ -2,14 +2,14 @@
 
 namespace Doctrine\ODM\MongoDB\Types;
 
-class CustomType extends Type
+trait CustomType
 {
     /**
      * @return string Redirects to the method convertToPHPValue from child class
      */
     final public function closureToPHP()
     {
-        $fqcn = self::class;
+        $fqcn = Type::class;
 
         return sprintf('
             $type = \%s::getType($typeIdentifier);
--- a/lib/Doctrine/ODM/MongoDB/Types/CustomType.php
+++ b/lib/Doctrine/ODM/MongoDB/Types/CustomType.php
@@ -1,4 +1,21 @@
 <?php
+/*
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * This software consists of voluntary contributions made by many individuals
+ * and is licensed under the MIT license. For more information, see
+ * <http://www.doctrine-project.org>.
+ */
 
 namespace Doctrine\ODM\MongoDB\Types;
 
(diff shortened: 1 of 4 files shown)


malarzm:
I'm personally fine with such addition (also introducing this method in a `trait` wanders through my mind), if we're collectively fine (ping @alcaeus @jmikola @jwage) there are few things to do:

1. License header is missing in new class
2. I'd add a PHPDoc for a class mentioning why is it there (and why one should consider inlining the code anyway)
3. Add documentation

Anyways thanks @mgonzalezbaile for throwing the idea :)

mgonzalezbaile:
I guess I should wait for confirmation before starting working on the task list given by @malarzm, right?

malarzm:
Pinged folks on IRC, please bear with us a little longer @mgonzalezbaile

lib/Doctrine/ODM/MongoDB/Types/CustomType.php
```diff
@@ -0,0 +1,18 @@
...
+{
+    /**
+     * @return string Redirects to the method convertToPHPValue from child class
+     */
+    final public function closureToPHP()
+    {
+        $fqcn = self::class;
+
+        return sprintf('
+            $type = \%s::getType($typeIdentifier);
```
jmikola: Would it make sense to call [`Doctrine\ODM\MongoDB\Types\Type::getType()`](https://github.com/doctrine/mongodb-odm/blob/2de3ae35e30c8dbc0d873a45c9c2fc55a2f281b3/lib/Doctrine/ODM/MongoDB/Types/Type.php#L145) directly here, or do we expect that it might be overridden (the method isn't `final`)?

lib/Doctrine/ODM/MongoDB/Types/CustomType.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+namespace Doctrine\ODM\MongoDB\Types;
+
+class CustomType extends Type
+{
+    /**
+     * @return string Redirects to the method convertToPHPValue from child class
+     */
+    final public function closureToPHP()
```
jmikola: > (also introducing this method in a trait wanders through my mind)

@malarzm: A trait may allow users more flexibility than forcing them to extend a new CustomType base class.

tests/Doctrine/ODM/MongoDB/Tests/Functional/CustomTypeTest.php
```diff
@@ -62,37 +63,28 @@ public function convertToDatabaseValue($value)
...
-
-        $converter = Type::getType('date');
-
-        $value = array_map(function($date) use ($converter) {
-            return $converter->convertToPHPValue($date);
-        }, array_values($value));
-
-        return $value;
+        return array_map(
+            function ($v) {
```
jmikola: Is there a particular reason to not rely on [`Doctrine\ODM\MongoDB\Types::convertToPHPValue()`](https://github.com/doctrine/mongodb-odm/blob/2de3ae35e30c8dbc0d873a45c9c2fc55a2f281b3/lib/Doctrine/ODM/MongoDB/Types/DateType.php#L94) here, as the original code was doing with `$converter->convertToPHPValue()`?

lib/Doctrine/ODM/MongoDB/Types/CustomType.php
malarzm: I wouldn't expect anyone to override the main method but we can't slap `final`

alcaeus: TBH, we're always using `Doctrine\ODM\MongoDB\Types\Type::getType()` directly to fetch a type, so I'd do the same here.

alcaeus:
I'm 👍 on moving this to a trait and 👍 in general as a fix to basically writing the logic twice for every type.

Since we've already talked about changing the types a bit for 2.0 (parameters for types come to mind, e.g. immutable=true/false for DateType), and I would like to add the closure handling to that list. Currently, a type closure can access all document data as well as the finished object during hydration and modify it at will. I know performance is a key issue here, but we also need to limit the scope a bit here. As it is, a type can do a lot of things during document hydration.

mgonzalezbaile:
I think I applied all your comments, let me know if something is missing :)

lib/Doctrine/ODM/MongoDB/Types/CustomType.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
```
malarzm: Please add licensing, you can copy it from any other file

lib/Doctrine/ODM/MongoDB/Types/CustomType.php
```diff
@@ -0,0 +1,18 @@
...
+    /**
+     * @return string Redirects to the method convertToPHPValue from child class
+     */
+    final public function closureToPHP()
+    {
+        $fqcn = Type::class;
+
+        return sprintf('
+            $type = \%s::getType($typeIdentifier);
+            $return = $type->convertToPHPValue($value);', $fqcn);
```
malarzm: Could you make this `return sprintf('$return = \%s::getType($typeIdentifier)->convertToPHPValue($value);', Type::class);` ?

mgonzalezbaile:
@malarzm do you think we can merge?

malarzm:
Ah crap @mgonzalezbaile I'm sorry, this fell through the cracks due to intensive season in my normal work... Anyways when I was about to hit "Merge" button last time I realized that we need to change `CustomType` trait name... At that time I (and generally after brief chat on IRC: we) had no idea how to name it better and the only good proposition came from @jmikola to name if after method it provides: `ClosureToPHP` - do you maybe have some better idea? 

Also we'll need to add a note in the docs so people will know such feature is in place, but this can be done by us in a follow up PR to speed things up.

mgonzalezbaile:
@malarzm just renamed the trait :-)

malarzm:
@mgonzalezbaile thanks! I'll merge this prob later tonight

malarzm:
Merged manually in https://github.com/doctrine/mongodb-odm/commit/44e1b92a7a18e256afc7c2926cfd69c8f0e07507 - thanks @mgonzalezbaile!

PR #1514, opened 2016-10-17
Correct Class Name in Custom Mapping Type Docs

acleon:
Docs were using an old class name. With these changes, was able to follow documentation and create an custom mapping type.


malarzm:
Thanks @acleon for the PR! I'll merge this as soon as Travis build is green

PR #1521, opened 2016-11-02
Avoid call to MongoDBs deprecated eval

gruberro:
By using the already existing `createCollection` instead of doing a fake command it's possbile to avoid calls to MongoDBs deprecated `eval`.

When using the `schema:create`-command at deploy time this will avoid MongoDB startup warnings like
```
"2016-10-21T14:05:35.870+0000 W COMMAND  [conn581654] the eval command is deprecated"
```

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/SchemaManager.php
+++ b/lib/Doctrine/ODM/MongoDB/SchemaManager.php
@@ -405,6 +405,8 @@ class SchemaManager
 
     /**
      * Create all the mapped document databases in the metadata factory.
+     *
+     * @deprecated Databases are created automatically by MongoDB (>= 3.0). Deprecated since ODM 1.2, to be removed in ODM 2.0.
      */
     public function createDatabases()
     {
@@ -421,6 +423,8 @@ class SchemaManager
      *
      * @param string $documentName
      * @throws \InvalidArgumentException
+     *
+     * @deprecated A database is created automatically by MongoDB (>= 3.0). Deprecated since ODM 1.2, to be removed in ODM 2.0.
      */
     public function createDocumentDatabase($documentName)
     {
@@ -428,6 +432,14 @@ class SchemaManager
         if ($class->isMappedSuperclass || $class->isEmbeddedDocument) {
             throw new \InvalidArgumentException('Cannot delete document indexes for mapped super classes or embedded documents.');
         }
+
+        $serverStatus = $this->dm->getDocumentDatabase($documentName)->command(['serverStatus' => true]);
+        $version = $serverStatus['version'];
+
+        if (version_compare($version, '3.0.0') >= 0) {
+            return;
+        }
+
         $this->dm->getDocumentDatabase($documentName)->execute('function() { return true; }');
     }
 
--- a/tests/Doctrine/ODM/MongoDB/Tests/SchemaManagerTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/SchemaManagerTest.php
@@ -253,19 +253,6 @@ class SchemaManagerTest extends \PHPUnit_Framework_TestCase
         $this->schemaManager->dropDocumentCollection(\Documents\CmsArticle::class);
     }
 
-    public function testCreateDocumentDatabase()
-    {
-        foreach ($this->documentDatabases as $class => $database) {
-            if ($class === \Documents\CmsArticle::class) {
-                $database->expects($this->once())->method('execute');
-            } else {
-                $database->expects($this->never())->method('execute');
-            }
-        }
-
-        $this->schemaManager->createDocumentDatabase(\Documents\CmsArticle::class);
-    }
-
     public function testDropDocumentDatabase()
     {
         foreach ($this->documentDatabases as $class => $database) {


alcaeus:
I like the change, could you backport it to the 1.0.x branch and change the base for the PR? That way we could provide this as a bugfix to the 1.0.x and 1.1.x releases as well, not only in 1.2.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -428,7 +428,7 @@ public function createDocumentDatabase($documentName)
         if ($class->isMappedSuperclass || $class->isEmbeddedDocument) {
             throw new \InvalidArgumentException('Cannot delete document indexes for mapped super classes or embedded documents.');
         }
-        $this->dm->getDocumentDatabase($documentName)->execute('function() { return true; }');
+        $this->dm->getDocumentDatabase($documentName)->createCollection($class->getCollection());
```
malarzm: Just to be sure we don't break some lowest setup when backporting - is there any test that will ensure database was in fact created? I agree that it should, just want to make sure it will :)

gruberro: After investigating some more time in testing the database creation I recognized that currently `createDocumentDatabase` is in fact not creating a database. Also creating a random collection and removing it afterwards isn't creating a database (or is creating a database and is removing the database once the last remaining collection in it is dropped). So there is no need to created databases itself. This is done by MongoDB once other commands/actions are creating collections, inserting data, etc.

tests/Doctrine/ODM/MongoDB/Tests/SchemaManagerTest.php
```diff
@@ -256,11 +256,7 @@ public function testDropDocumentCollection()
     public function testCreateDocumentDatabase()
     {
         foreach ($this->documentDatabases as $class => $database) {
-            if ($class === \Documents\CmsArticle::class) {
-                $database->expects($this->once())->method('execute');
-            } else {
-                $database->expects($this->never())->method('execute');
-            }
+            $database->expects($this->never())->method('createCollection');
```
alcaeus: I think we can do away with the tests completely - no point in checking a method that NOPs.

malarzm:
Technically speaking this removes a feature one might be relying on

alcaeus:
Yes - that is true. However:
- On MongoDB 3.2 (and I would assume this was introduced with 3.0), the database is no longer created. In fact, MongoDB drops the entire database when the last collection in it is dropped. There is no way to create just the database
- When running an operation, the database is implicitly created. Since there are no options to be configured, we can rely on this "feature"
- Not creating the database at least makes this whole thing explicit. Currently, the call to createDatabase just silently NOPs on newer MongoDB servers.

malarzm:
Just to recap conversation from IRC: I'm suggesting adding version test and let the method execute normally if running on servers older than 3.0 and make it NOP for newer ones. For the record I fully agree that we shouldn't go down the version switches path often :)

gruberro:
I've added a version check skipping the database creation for all Mongo versions >= 3.0. How to deal with the deprecation annotations?

alcaeus:
I'm still in favor of dropping these for ODM 2.0, so I suggest we leave the deprecations in 👍

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -421,13 +423,23 @@ public function createDatabases()
      *
      * @param string $documentName
      * @throws \InvalidArgumentException
+     *
+     * @deprecated document databases must not be created, databases are created automatically
```
malarzm: I'd just rephrase to say that method was deprecated in 1.2, to be removed in 2.0 and doesn't need to be worked around as databases are created automatically starting from 3.0

gruberro: ✔️

alcaeus:
Thanks @gruberro! 🍻

acrozes:
That commit  introduce a bug:

[2016-11-16 05:42:19] php_error.WARNING: Undefined index: version {"type":8,"file":"./vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/SchemaManager.php","line":435,"level":30719} 

If the user is not allow to execute the command "serverStatus":

$serverStatus = $this->dm->getDocumentDatabase($documentName)->command(['serverStatus' => true]);
$serverStatus == [
    'ok' => 0
    'errmsg' => not authorized on user to execute command { serverStatus: true }
    'code' => 13
]

alcaeus:
Could you share some details about your database in regards to authentication? Also, what version of MongoDB are you running?

acrozes:
Mongodb: 3.2.10
PHP: 7.0.13
Mongodb User:
```
use myuser
db.createUser(
  {
    user: "myuser",
    pwd: "password",
    roles: [ { role: "dbOwner", db: "myuser" } ]
  }
)
```
For connexion: 
mongo_server: mongodb://myuser:password@domain.net:27017/myuser

alcaeus:
Ah, authentication. Apparently, users need the `clusterMonitor` role to be able to execute this command: https://dba.stackexchange.com/questions/121832/db-serverstatus-got-not-authorized-on-admin-to-execute-command/121838#121838.

@malarzm So, what should we do if we can't determine the MongoDB version due to those restrictions? Normally I'd say "err on the side of caution" and try to avoid calling a non-existing (for whenever `eval` gets removed) command, but I understand your concerns about a BC break. Opinions?

malarzm:
If we can't reliably check the version then I think it's better to indeed remove the feature and add it to changelog as potential BC break that we couldn't avoid... Unless it's possible to make a command switch to opt out of creating database? Then we could advice to start using the switch, deprecate whole thing and remove it in 2.0

PR #1523, opened 2016-11-11
Add fieldTypeName to yml-generator

waldihuber:
If you implement and register a custom type there is no way to define a destination type/field name for the generated fields and methods.

With this change you can define in the yml-file:

        created_at:
            type: mycustomtype
            fieldTypeName: \DateTime

This results in:

```
    /**
     * @var \DateTime $created_at
     */
    protected $created_at;

/**
     * Get createdAt
     *
     * @return \DateTime $createdAt
     */
    public function getCreatedAt()
    {
        return $this->created_at;
    }

/**
     * Set createdAt
     *
     * @param \DateTime $createdAt
     * @return self
     */
    public function setCreatedAt( \DateTime $createdAt)
    {
        $this->created_at = $createdAt;
        return $this;
    }
```

### DIFF
--- a/docs/en/reference/basic-mapping.rst
+++ b/docs/en/reference/basic-mapping.rst
@@ -339,50 +339,48 @@ You can define your own ID generator by extending the
 ``Doctrine\ODM\MongoDB\Id\AbstractIdGenerator`` class and specifying the class
 as an option for the ``CUSTOM`` strategy:
 
-.. configuration-block::
+.. code-block:: php
 
-    .. code-block:: php
+	<?php
 
-        <?php
+	/** Document */
+	class MyPersistentClass
+	{
+		/** @Id(strategy="CUSTOM", type="string", options={"class"="Vendor\Specific\Generator"}) */
+		private $id;
 
-        /** Document */
-        class MyPersistentClass
-        {
-            /** @Id(strategy="CUSTOM", type="string", options={"class"="Vendor\Specific\Generator"}) */
-            private $id;
+		public function setId($id)
+		{
+			$this->id = $id;
+		}
 
-            public function setId($id)
-            {
-                $this->id = $id;
-            }
+		//...
+	}
 
-            //...
-        }
+.. code-block:: xml
 
-    .. code-block:: xml
+	<doctrine-mongo-mapping xmlns="http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping"
+							xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+							xsi:schemaLocation="http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping
+												http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping.xsd">
 
-        <doctrine-mongo-mapping xmlns="http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping"
-                                xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-                                xsi:schemaLocation="http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping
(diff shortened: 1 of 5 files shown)


lib/Doctrine/ODM/MongoDB/Tools/DocumentGenerator.php
```diff
@@ -661,20 +661,22 @@ private function generateDocumentStubMethods(ClassMetadataInfo $metadata)
         $methods = array();
 
         foreach ($metadata->fieldMappings as $fieldMapping) {
+            $typeHint = isset($fieldMapping['fieldTypeName']) ? $fieldMapping['fieldTypeName'] : $fieldMapping['type'];
```
malarzm: I'm not fan of this name but frankly I don't have anything good on my mind (beside `typeTypehint`) - @alcaeus @jmikola any ideas?

lib/Doctrine/ODM/MongoDB/Tools/DocumentGenerator.php
```diff
@@ -661,20 +661,22 @@ private function generateDocumentStubMethods(ClassMetadataInfo $metadata)
...
+            
             if (isset($fieldMapping['id'])) {
                 if ($metadata->generatorType == ClassMetadataInfo::GENERATOR_TYPE_NONE) {
-                    if ($code = $this->generateDocumentStubMethod($metadata, 'set', $fieldMapping['fieldName'], $fieldMapping['type'])) {
+                    if ($code = $this->generateDocumentStubMethod($metadata, 'set', $fieldMapping['fieldName'], $typeHint)) {
                         $methods[] = $code;
                     }
                 }
-                if ($code = $code = $this->generateDocumentStubMethod($metadata, 'get', $fieldMapping['fieldName'], $fieldMapping['type'])) {
+                if ($code = $code = $this->generateDocumentStubMethod($metadata, 'get', $fieldMapping['fieldName'], $typeHint)) {
```
malarzm: Could you remove the deuplicated assignment while we're here please?

lib/Doctrine/ODM/MongoDB/Tools/DocumentGenerator.php
```diff
@@ -789,7 +791,7 @@ private function generateDocumentStubMethod(ClassMetadataInfo $metadata, $type,
         $description = ucfirst($type) . ' ' . $variableName;
 
         $types = Type::getTypesMap();
-        $methodTypeHint = $typeHint && ! isset($types[$typeHint]) ? '\\' . $typeHint . ' ' : null;
+        $methodTypeHint = $typeHint && ! isset($types[$typeHint]) ? (mb_substr($typeHint, 0, 1) != '\\' ? '\\' : '') . $typeHint . ' ' : null;
```
malarzm: I think there's no need for `mb_substr`, `substr` should be sufficient. Also please use `!==`

waldihuber:
I did all the changes :)

docs/en/reference/basic-mapping.rst
```diff
@@ -339,50 +339,48 @@ You can define your own ID generator by extending the
 ``Doctrine\ODM\MongoDB\Id\AbstractIdGenerator`` class and specifying the class
 as an option for the ``CUSTOM`` strategy:
 
-.. configuration-block::
```
alcaeus: Why are you removing `configuration-block`?

waldihuber: I removed them in the hope they will then display content like other blocks in the same document. Currently you cannot see any code or configuration code at "Fields":
https://github.com/doctrine/mongodb-odm/blob/master/docs/en/reference/basic-mapping.rst

alcaeus:
Closing for lack of feedback and the fact that it's a mapping solution to a code generation problem. The document generator is something you really shouldn't use and one that might be dropped for 2.0, so IMO there's no point merging this.

PR #1524, opened 2016-11-16
Remove version check in SchemaManager

alcaeus:
This PR removes the version check in SchemaManager introduced with #1521. This is necessary since the `serverStatus` command requires extra privileges which may or may not have been given to the user being used. Instead, users should skip the creation of the database by passing the `--collection` and `--index` options, omitting the `--db` option when creating the schema using the `schema:create` command.

The deprecation of the `createDatabases` and `createDocumentDatabase` methods in `SchemaManager` stays intact; the methods will be removed with ODM 2.0.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Tools/Console/Command/Schema/CreateCommand.php
+++ b/lib/Doctrine/ODM/MongoDB/Tools/Console/Command/Schema/CreateCommand.php
@@ -45,6 +45,10 @@ class CreateCommand extends AbstractCommand
 
     protected function execute(InputInterface $input, OutputInterface $output)
     {
+        if ($input->getOption(self::DB)) {
+            @trigger_error('The ' . self::DB . ' option is deprecated and will be removed in ODM 2.0', E_USER_DEPRECATED);
+        }
+
         foreach ($this->createOrder as $option) {
             if ($input->getOption($option)) {
                 $create[] = $option;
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ODM/MongoDB/Tools/Console/Command/Schema/CreateCommand.php
```diff
@@ -45,6 +45,10 @@ protected function configure()
 
     protected function execute(InputInterface $input, OutputInterface $output)
     {
+        if ($input->getOption(self::DB)) {
+            trigger_error('The ' . self::DB . ' option is deprecated and will be removed in ODM 2.0', E_USER_DEPRECATED);
```
malarzm: This can also break builds, no? Should we use `@`?

jmikola:
You should be able to use [`isMaster`](https://docs.mongodb.com/manual/reference/command/isMaster/) to determine the server version based on `minWireVersion` and `maxWireVersion` fields. This does require authentication (see: [DRIVERS-90](https://jira.mongodb.org/browse/DRIVERS-90)), but it should not require any special user permissions since all drivers use `isMaster` for server discovery and monitoring. Here's a quick table of wire versions correlated with server versions:

| Server version | Wire version |
|---|---|
| 2.4 and earlier | 0 |
| 2.6 | 2 |
| 3.0 | 3 |
| 3.2 | 4 |
| 3.4 | 5 |

alcaeus:
@jmikola I think we'll just leave it as is. Users can decide not to create the database by passing the appropriate options to the command (or not calling `createDatabase` if they are doing it programmatically). We might revisit this if MongoDB removes the eval functionality before we get a chance to remove the deprecated methods completely, but until then I think we'll be safe.

lib/Doctrine/ODM/MongoDB/Tools/Console/Command/Schema/CreateCommand.php
alcaeus: Good point. I've changed it.

malarzm:
:+1:

PR #970, opened 2014-11-01
Embedded priming fix for Issue #624, with Tests

gistya:
This is a fix for the issue where dot syntax would not work in field-names to prime. 
https://github.com/doctrine/mongodb-odm/issues/624

With this fix, now you can prime a reference that is in an embedded document. It works no matter how many levels of embedded documents there are between the original document and the final reference, but it does not traverse across multiple references. 

The recursive method is based loosely upon the structure of the prepareQueryElement() method of Doctrine\ODM\MongoDB\Persisters\DocumentPersister, except with priming it can be much simpler since we only care if it's a reference or embedded document.

PR #1527, opened 2016-11-23
Store redundant fields in references

alcaeus:
References in MongoDB can be just IDs for a referenced document, or they can be full-fledged DBRef objects that can contain additional information. We use this functionality to add a discriminator value to properly hydrate a referenced document.

This PR allows developers to store additional fields from a referenced document within the reference, so it can be used in queries. There are two caveats at this time:
* The denormalized value won't be updated. This would require us to keep track of incoming references to a document and update all documents referencing ours when such a field is updated. Apart from the potentially long update query, this would only work within the first nesting level of documents due to [SERVER-831](https://jira.mongodb.org/browse/SERVER-831).
* The value is not yet used when creating proxy classes. This is because the current implementation of the proxy library doesn't allow for partially initialized proxies. This has been discussed in doctrine/doctrine2#1241 and I'd like to add that to ODM 2.0.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Mapping/Annotations/ReferenceMany.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/Annotations/ReferenceMany.php
@@ -42,10 +42,11 @@ final class ReferenceMany extends AbstractField
     public $inversedBy;
     public $mappedBy;
     public $repositoryMethod;
-    public $sort = array();
-    public $criteria = array();
+    public $sort = [];
+    public $criteria = [];
     public $limit;
     public $skip;
     public $strategy = CollectionHelper::DEFAULT_STRATEGY;
     public $collectionClass;
+    public $redundantFields = [];
 }
--- a/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataInfo.php
@@ -1323,6 +1323,14 @@ class ClassMetadataInfo implements \Doctrine\Common\Persistence\Mapping\ClassMet
             throw MappingException::simpleReferenceRequiresTargetDocument($this->name, $mapping['fieldName']);
         }
 
+        if (isset($mapping['reference'])
+            && isset($mapping['storeAs'])
+            && $mapping['storeAs'] === ClassMetadataInfo::REFERENCE_STORE_AS_ID
+            && ! empty($mapping['redundantFields'])
+        ) {
+            throw MappingException::simpleReferenceCannotHaveRedundantFields($this->name, $mapping['fieldName']);
+        }
+
         if (isset($mapping['reference']) && empty($mapping['targetDocument']) && empty($mapping['discriminatorMap']) &&
                 (isset($mapping['mappedBy']) || isset($mapping['inversedBy']))) {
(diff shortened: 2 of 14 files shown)


lib/Doctrine/ODM/MongoDB/Mapping/Annotations/ReferenceMany.php
```diff
@@ -48,4 +48,5 @@
     public $skip;
     public $strategy = CollectionHelper::DEFAULT_STRATEGY;
     public $collectionClass;
+    public $redundantFields = array();
```
malarzm: Please use short array syntax

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataInfo.php
```diff
@@ -1323,6 +1323,15 @@ public function mapField(array $mapping)
             throw MappingException::simpleReferenceRequiresTargetDocument($this->name, $mapping['fieldName']);
         }
 
+        if (isset($mapping['reference'])
+            && isset($mapping['storeAs'])
+            && $mapping['storeAs'] === ClassMetadataInfo::REFERENCE_STORE_AS_ID
+            && isset($mapping['redundantFields'])
+            && $mapping['redundantFields'] !== []
```
malarzm: The last two checks can be combined to `! empty($mapping['redundantFields'])`

lib/Doctrine/ODM/MongoDB/Mapping/Driver/YamlDriver.php
```diff
@@ -304,6 +304,9 @@ private function addMappingFromReference(ClassMetadataInfo $class, $fieldName, $
         if (isset($reference['criteria'])) {
             $mapping['criteria'] = $reference['criteria'];
         }
+        if (isset($reference['redundantFields'])) {
+            $mapping['redundantFields'] = $reference['redundantFields'];
```
malarzm: Should we check if what we got is an array and not scalar?

malarzm:
Looks cool, putting into 1.2 feature bucket :+1:

lib/Doctrine/ODM/MongoDB/Mapping/Driver/YamlDriver.php
alcaeus: I added a cast to array, that way scalars are accepted gracefully instead of being discarded.

malarzm:
Leaving a note to future us:

* need to check if storing association (can't recall if `getField` from class metadata can give us association) is safe
* allowing dot notation in redundant fields sounds interesting (not sure how helpful this could be but https://github.com/doctrine/mongodb-odm/pull/970/files introduced `ReferencePrimer::parseDotSyntaxForPrimer` which may or may not be useful :) )

alcaeus:
Dot notation would be a bit complicated, since that may trigger a whole bunch of database loads. There's already a potential database load involved when building a DBRef object (e.g. when the referenced document is not initialized and redundant fields are defined). This happens even when the DBRef object is not used for writing, but only for reading (e.g. by being called in `Query\Builder::references`. Allowing dot notation and thus a whole reference tree would only make the problem worse. This is something I'd like to address, but I'm not sure it can be done in a BC fashion.

alcaeus:
Closing as the PR is horribly stale: created #2252 to track this and will try to revisit this for one of the next 2.x releases.

PR #1529, opened 2016-12-01
Fix preLoad annotation event documentation

crashbtz:


alcaeus:
Interesting, I remember @malarzm saying (in #1222) that since the event wasn't really documented, it's technically not a BC break 😁 . Either way, thanks for fixing this, and sorry for the inconvenience!

crashbtz:
haha, no worries

malarzm:
Oh I wasn't aware of this place in docs :D Sorry once again!

PR #1530, opened 2016-12-02
Add PHP 7.1 and MongoDB 3.2 to the build matrix

alcaeus:
With the old matrix, we'd have 10 builds, which seems excessive. Since there's no need to test every PHP version against every MongoDB version, we only test PHP 5.6 against each MongoDB version.

Note: tests against MongoDB 3.4 are not possible at this time; I'll add them once MongoDB 3.4 is available on the repository we're using in our travis builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,24 +5,25 @@ language: php
 php:
   - 5.6
   - 7.0
+  - 7.1
 
 env:
   global:
     - DOCTRINE_MONGODB_SERVER="mongodb://localhost:27017"
     - KEY_SERVER="hkp://keyserver.ubuntu.com:80"
-    - MONGO_REPO_URI="http://repo.mongodb.com/apt/ubuntu"
+    - MONGO_REPO_URI="https://repo.mongodb.com/apt/ubuntu"
     - MONGO_REPO_TYPE="precise/mongodb-enterprise/"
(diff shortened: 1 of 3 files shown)


malarzm:
Seems we have some failing tests on 7.1 my first guess goes to http://php.net/manual/en/migration71.incompatible.php#migration71.incompatible.datetime-microseconds

alcaeus:
Yes, I'm trying to figure out who's to blame: ODM, the adapter, the driver or PHP itself ;)

alcaeus:
Works now - our tests previously assumed that the `usec` portion in the `MongoDate` objects would always be 0. The calculations in the tests are due to the fact that `MongoDate` (and `UTCDateTime` in the new MongoDB driver) only store millisecond precision, not microsecond.

caiquecastro:
A "native" support to php ^7 depends on this library or on the doctrine/mongodb abstraction layer?

alcaeus:
@caiquecastro the doctrine/mongodb abstraction layer will not be adapted to support the new MongoDB driver. Instead, we'll be relying on the official abstraction library (mongodb/mongodb).

"Native" support for the new MongoDB driver will be one of the core points of ODM 2.0 - along with support for strict types and other features included in PHP 7.0 and 7.1.

PR #1531, opened 2016-12-06
Remove suggest mongo-php-adapter

caiquecastro:
It is already suggested by doctrine/mongodb which is required by this
library.


alcaeus:
Thanks!

PR #1532, opened 2016-12-07
[PersistentCollection] Fix compatibility with PHP7 return types

Maks3w:
Solves the issue about signature mismatch due missing return type in child class.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/PersistentCollection/DefaultPersistentCollectionGenerator.php
+++ b/lib/Doctrine/ODM/MongoDB/PersistentCollection/DefaultPersistentCollectionGenerator.php
@@ -19,6 +19,7 @@
 
 namespace Doctrine\ODM\MongoDB\PersistentCollection;
 
+use Doctrine\Common\Proxy\Exception\UnexpectedValueException;
 use Doctrine\ODM\MongoDB\Configuration;
 
 /**
@@ -189,7 +190,7 @@ CODE;
     /**
      * {@inheritDoc}
      */
-    public function {$method->name}($parametersString)
+    public function {$method->name}($parametersString){$this->getMethodReturnType($method)}
     {
         \$this->initialize();
         if (\$this->needsSchedulingForDirtyCheck()) {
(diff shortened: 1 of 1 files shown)


Maks3w:
I've tested in local tuning CustomCollectionsTest successfully. I didn't commit the changes due PHP5 vs PHP7 runs.

alcaeus:
Great, I haven't even thought of that! With PHP 7.1 released, could you add support for nullable types and the void return type?

Maks3w:
@alcaeus The code is based on Proxy code present at doctrine/commons. When that code become updated this could be done as well.

alcaeus:
This was added to doctrine/common in 2.7.0: https://github.com/doctrine/common/pull/734

Maks3w:
Then the code proposed on this PR include those changes. I just copy the code present on master yesterday.

alcaeus:
Ok. I'll take a look later today.

lib/Doctrine/ODM/MongoDB/PersistentCollection/DefaultPersistentCollectionGenerator.php
```diff
@@ -297,4 +298,65 @@ function (\ReflectionParameter $parameter) {
...
+     * @param \ReflectionParameter|null $parameter
+     *
+     * @return string
+     *
+     * @see \Doctrine\Common\Proxy\ProxyGenerator::formatType()
+     */
+    private function formatType(
+        \ReflectionType $type,
+        \ReflectionMethod $method,
+        \ReflectionParameter $parameter = null
```
Maks3w: It's not my intention make something different from the code already present on doctrine commons.

If both snippets are equals then maintenance will be easy until a long term solution (DRY) become

alcaeus: I agree, we should try to keep it consistent.

lib/Doctrine/ODM/MongoDB/PersistentCollection/DefaultPersistentCollectionGenerator.php
```diff
@@ -297,4 +298,65 @@ function (\ReflectionParameter $parameter) {
...
+        $nameLower = strtolower($name);
+        if ('self' === $nameLower) {
+            $name = $method->getDeclaringClass()->getName();
+        }
+        if ('parent' === $nameLower) {
+            $name = $method->getDeclaringClass()->getParentClass()->getName();
+        }
+        if ( ! $type->isBuiltin() && ! class_exists($name) && ! interface_exists($name)) {
+            if (null !== $parameter) {
+                throw UnexpectedValueException::invalidParameterTypeHint(
```
alcaeus: This should not be a `Proxy\UnexpectedValueException`. Either create a new exception class or throw a `PersistentCollectionException`.

alcaeus:
@malarzm I was going to ask for some tests, then I saw that we currently don't have any. Would it be worth to create some tests for this class?

malarzm:
IIRC I wasn't originally writing tests for this specific class because for any of custom collection test this generator must be well and alive :)

Maks3w:
This test class cover the scenario for PHP <7 but it's heavily coupled to the whole MongoDB setup and I don't have time for to perform a refactor and decouple the tests.

Sorry for not to be more useful but don't expect more changes from my side. The branch is open so any of you can add more commits

malarzm:
> Sorry for not to be more useful but don't expect more changes from my side.

@Maks3w mind changing at least the exception to `PersistentCollectionException` so we could merge this as is and take care of tests on our own later?

malarzm:
Closing in favour of #1543. Thanks for getting us started on this @Maks3w!

PR #1540, opened 2016-12-23
Test against MongoDB 3.4 (1.1.x)

alcaeus:
This adds MongoDB 3.4 to the test suite and switches to the `mongodb-org` package instead of `mongodb-enterprise`. Since `mongodb-org` 2.6 is not available, we drop it from the build matrix to avoid having two different apt repositories in the build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,26 +11,28 @@ env:
   global:
     - DOCTRINE_MONGODB_SERVER="mongodb://localhost:27017"
     - KEY_SERVER="hkp://keyserver.ubuntu.com:80"
-    - MONGO_REPO_URI="https://repo.mongodb.com/apt/ubuntu"
-    - MONGO_REPO_TYPE="precise/mongodb-enterprise/"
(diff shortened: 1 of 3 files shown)


tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/LookupTest.php
```diff
@@ -258,4 +292,20 @@ private function requireMongoDB32()
             $this->markTestSkipped('$lookup tests require at least MongoDB 3.2.0');
         }
     }
+
+    private function skipOnMongoDB34()
```
malarzm: it seems like the number of methods will keep on growing, maybe it'd be feasible to add a method to our base test class: `->skipIfMongo($operator, $version, $message)` - what do you think?

alcaeus:
Note: I'm holding off on merging this because I'd like to backport #1530 to the 1.1.x branch and base this PR against 1.1.x as well. There's no reason not to have support and tests for PHP 7.1 in 1.1.x.

PR #1542, opened 2016-12-25
Fix embedded document parent before computing changeset

### LINKED ISSUE #1525: Embedded document isn't added in update
Sometimes an embedded document is not added to the update data in the PersistenceBuilder.

This happens when I have 2 documents and add the same embedded document to both documents. Then I perform a flush (in a loop so the documents are flushed seperately). 
The first time the unit of work sees the embedded document as new and $this->uow->isScheduledForInsert($new) returns true and the embedded document is added to the $updateData.

The second time I perform a flush the unit of work sees the document as already added, so $this->uow->isScheduledForInsert($new) returns false.  Then $this->prepareUpdateData($new); will be called, which will return nothing because the documentchangeset for the embedded document is empty.

My temporary workaround is to clear the documentmanager after each flush, but thats not really a desirable solution.

* the code I mentioned is in the PersistenceBuilder from line 161 to 178 (function prepareUpdateData)

alcaeus:
Fixes #1525.

The problem is that the changeset wrote down the old embedded document, which was then used when trying to persist changes. I don't like adding another check for the embedded document parent, maybe @malarzm has a better idea?

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
+++ b/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
@@ -746,6 +746,15 @@ class UnitOfWork implements PropertyChangedListener
                         $this->scheduleOrphanRemoval($orgValue);
                     }
 
+                    if ($actualValue !== null) {
+                        list(, $knownParent, ) = $this->getParentAssociation($actualValue);
+                        if ($knownParent && $knownParent !== $document) {
+                            $actualValue = clone $actualValue;
+                            $class->setFieldValue($document, $class->fieldMappings[$propName]['fieldName'], $actualValue);
+                            $this->setOriginalDocumentProperty(spl_object_hash($document), $class->fieldMappings[$propName]['fieldName'], $actualValue);
+                        }
+                    }
+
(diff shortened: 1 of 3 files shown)


malarzm:
The only thing that comes to my mind is removing the check in `persist` part of UoW and seeing if the new one will be sufficient for both cases (I think there was a test checking if the embedded document is changed as soon as `persist` is called but I think we could break that as we were mostly interested in having embedded document cloned eventually during `flush`, not in specific moment in time)

alcaeus:
@malarzm tried that, with the following result:

> 1) Doctrine\ODM\MongoDB\Tests\Functional\EmbeddedTest::testReusedEmbeddedDocumentsAreClonedInFact
> Failed asserting that two variables don't reference the same object.
>
> tests/Doctrine/ODM/MongoDB/Tests/Functional/EmbeddedTest.php:580

malarzm:
That exact assertion could be removed as it was checking if embedded documents were cloned as soon as `persist` :)

alcaeus:
@malarzm I removed the failing assertion, everything else works as expected.

malarzm:
Cool!

PR #1547, opened 2017-01-04
Add section about using PHP 7 to the documentation

### LINKED ISSUE #1545: php7, mongodb, windows server 2008, composer do not work =)
hello,
I Try to install mongodb with php 7.0.9 on windows serveur 2008.

- i installed php 7.0.9 with MAMP
- i install mongodb php driver on https://pecl.php.net/package/mongodb ( 1.2.2 )
- php -m  ( i have mongodb)
- i try to install my application with composer

        "doctrine/mongodb": "1.3.*@dev",
        "doctrine/mongodb-odm": "1.1.*@dev",

- I have an error 

the requested php extension mongo is missing from your system windows..

Strange no?? ... 

Thanks, =)

alcaeus:
Fixes #1545.

### DIFF
--- a/docs/en/reference/introduction.rst
+++ b/docs/en/reference/introduction.rst
@@ -457,6 +457,37 @@ The final ``bootstrap.php`` file should look like this:
 
 That is it! Your ``DocumentManager`` instance is ready to be used!
 
+Using PHP 7
+-----------
+
+You can use Doctrine MongoDB ODM with PHP 7, but there are a few extra steps during
+the installation. Since the legacy driver (referred to as ``ext-mongo``) is not
+available on PHP 7, you will need the new driver (``ext-mongodb``) installed and
+use a polyfill to provide the API of the legacy driver.
+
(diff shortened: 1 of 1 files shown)


docs/en/reference/introduction.rst
```diff
@@ -457,6 +457,37 @@ The final ``bootstrap.php`` file should look like this:
 
 That is it! Your ``DocumentManager`` instance is ready to be used!
 
+Using PHP 7
+-----------
+
+You can use Doctrine MongoDB ODM with PHP 7, but there are a few extra steps during
```
malarzm: Should we add a `..note:` that PHP 7 can be used with ODM 1.1+ ?

alcaeus: I omitted that since we no longer support 1.0. I can add it if you think it's necessary. The composer command mentioned in the documentation will automatically install 1.1 at this time anyways.

malarzm: That's true however for people having (for whatever reasons) 1.0 and upgrading PHP to 7 this might be valuable information I think

PR #1550, opened 2017-01-09
Fix more re-used embedded docs issues

malarzm:
I'll comment on changes separately but as for now failing test case @redthor has provided in #1525 is fixed

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
+++ b/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
@@ -745,16 +745,6 @@ class UnitOfWork implements PropertyChangedListener
                     if ($orgValue !== null) {
                         $this->scheduleOrphanRemoval($orgValue);
                     }
-
-                    if ($actualValue !== null) {
-                        list(, $knownParent, ) = $this->getParentAssociation($actualValue);
-                        if ($knownParent && $knownParent !== $document) {
-                            $actualValue = clone $actualValue;
-                            $class->setFieldValue($document, $class->fieldMappings[$propName]['fieldName'], $actualValue);
-                            $this->setOriginalDocumentProperty(spl_object_hash($document), $class->fieldMappings[$propName]['fieldName'], $actualValue);
-                        }
-                    }
-
                     $changeSet[$propName] = array($orgValue, $actualValue);
                     continue;
                 }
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -745,16 +745,6 @@ private function computeOrRecomputeChangeSet(ClassMetadata $class, $document, $r
                     if ($orgValue !== null) {
                         $this->scheduleOrphanRemoval($orgValue);
                     }
-
-                    if ($actualValue !== null) {
```
malarzm: This is #1542 removed because it was good only for updates (it's in the path when document is not new) while one of tests I've added was interested in fixing inserts

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -998,6 +988,10 @@ private function computeAssociationChanges($parentDocument, array $assoc, $value
                             if ($assoc['type'] === ClassMetadata::ONE) {
                                 $class->setFieldValue($parentDocument, $assoc['fieldName'], $entry);
                                 $this->setOriginalDocumentProperty(spl_object_hash($parentDocument), $assoc['fieldName'], $entry);
+                                $poid = spl_object_hash($parentDocument);
+                                if (isset($this->documentChangeSets[$poid][$assoc['fieldName']])) {
+                                    $this->documentChangeSets[$poid][$assoc['fieldName']][1] = $entry;
```
malarzm: This fixed #1542's failing test case and new one with early persist

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -2193,6 +2187,11 @@ function ($assoc) { return $assoc['isCascadePersist']; }
                 }
             } elseif ($relatedDocuments !== null) {
                 if ( ! empty($mapping['embedded'])) {
+                    list(, $knownParent, ) = $this->getParentAssociation($relatedDocuments);
+                    if ($knownParent && $knownParent !== $document) {
+                        $relatedDocuments = clone $relatedDocuments;
+                        $class->setFieldValue($document, $mapping['fieldName'], $relatedDocuments);
+                    }
```
malarzm: This however I fail to see why I need to bring back but without this change set test with late persist is failing for reason I don't see yet

redthor:
Thanks @malarzm ! I've removed the `clone()` I had added to our models and then pulled in your branch and ran the particular integration test that was failing and it passed :sunglasses:

PR #1554, opened 2017-01-24
Use reference name instead of field name in Query expression builder

ossinkine:


lib/Doctrine/ODM/MongoDB/Query/Expr.php
```diff
@@ -90,7 +90,7 @@ public function references($document)
                 }
 
                 foreach ($keys as $key => $value) {
-                    $this->query[$this->currentField . '.$' . $key] = $dbRef['$' . $key];
+                    $this->query[$mapping['name'] . '.$' . $key] = $dbRef['$' . $key];
```
malarzm: Sorry for long time that has passed, but I think this also should be fixed in `includesReferenceTo` method below, would you mind?

ossinkine: I'm not sure how this method should work properly and it's not covered by tests.
So I made this changes on your own risk.

malarzm: I'll take the responsibility :)

malarzm:
Merged manually in https://github.com/doctrine/mongodb-odm/commit/0d4c2f06f3f02583af3304e241cb5b1319992d0c, thanks @ossinkine!

PR #1556, opened 2017-02-01
Change serialization of inheritance fields

### LINKED ISSUE #1271: Broken ODM schema produces different results with and without schema caching
On dev enviroment we don't use MetadataCache, but in production we do.
The problem is that the schema genereated from yaml files seems to be okay in dev, but fails on production.
The first request is okay when there is no cache yet, but when the schema is read from cache sometimes there are errors that are not experienced without caching.
By saying "using no cache" I mean either setting ArrayCache, or setting other caching methods like Filesystem cache or Mongodb cache but clearing all entries before running the script.

In the example below you can see a schema with inheritance, with some lines commented out.
I know that they are invalid without the commented lines, but it works when caching is not enabled.
When retrieving documents from DB without cache (even if I caching is on but empty), I get the correct hydrated objects, but when using caching, ODM tries to hydrate the document as Document\Font\Family (the superclass)  

```
Document\Font\Family:
#  type: mappedSuperclass
#  inheritanceType: SINGLE_COLLECTION
  collection: FontFamily
  repositoryClass: Repository\Font\Family
  discriminatorField: _t
  discriminatorMap:
      system: Document\Font\Family\System
      google: Document\Font\Family\Google
      uploaded: Document\Font\Family\Uploaded
  fields:
    id:
      id: true
      strategy: AUTO
    name:
      type: string
      name: n
    wysiwygSafe:
      type: boolean
      name: ws

Document\Font\Family\Google:
  collection: FontFamily
#  inheritanceType: SINGLE_COLLECTION
  repositoryClass: Repository\Font\Family

Document\Font\Family\System:
  collection: FontFamily
#  inheritanceType: SINGLE_COLLECTION
  repositoryClass: Repository\Font\Family

Document\Font\Family\Uploaded:
  collection: FontFamily
#  inheritanceType: SINGLE_COLLECTION
  repositoryClass: Repository\Font\Family
  fields:
      md5:
          type: string
          name: md5
```

It is absolutely clear that the problem is with the missing attrbutes (inheritanceType, type: mappedSuperclass), but the sad thing is that we encounter the problem on production env only.

Is there a way that we validate the schema with a cli tool, or to make the caching system so that ODM produces the same results with and without caching (even if the schema is invalid)?

alcaeus:
Fixes #1271.

The previous solution would always require specifying an inheritance type, which makes sense for documents but seems weird for embedded documents. With this PR, this changes to serialize the inheritance information (inheritance type and discriminator data) if an inheritance type was specified or if the discriminator field was given. This should fix hard to find cases where data can't be stored or loaded once the metadata was loaded from cache.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
@@ -123,7 +123,7 @@ class ClassMetadata extends ClassMetadataInfo
             $serialized[] = 'customRepositoryClassName';
         }
 
-        if ($this->inheritanceType != self::INHERITANCE_TYPE_NONE) {
+        if ($this->inheritanceType != self::INHERITANCE_TYPE_NONE || $this->discriminatorField !== null) {
             $serialized[] = 'inheritanceType';
             $serialized[] = 'discriminatorField';
             $serialized[] = 'discriminatorValue';
--- a/tests/Doctrine/ODM/MongoDB/Tests/Mapping/ClassMetadataTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Mapping/ClassMetadataTest.php
@@ -4,6 +4,7 @@ namespace Doctrine\ODM\MongoDB\Tests\Mapping;
 
 use Doctrine\ODM\MongoDB\Mapping\ClassMetadata;
 use Doctrine\ODM\MongoDB\Mapping\ClassMetadataInfo;
+use Documents\CmsUser;
 
 class ClassMetadataTest extends \Doctrine\ODM\MongoDB\Tests\BaseTest
 {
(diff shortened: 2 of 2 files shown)


Bilge:
Is there any reason not to merge this?

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -123,7 +123,7 @@ public function __sleep()
             $serialized[] = 'customRepositoryClassName';
         }
 
-        if ($this->inheritanceType != self::INHERITANCE_TYPE_NONE) {
+        if ($this->inheritanceType != self::INHERITANCE_TYPE_NONE || $this->discriminatorField !== null) {
```
malarzm: wouldn't it be better to check for being embedded document's metadata? Otherwise the issue still persists when not using `discriminatorField` which I think can be `null` to default to `_doctrine_class_name` later down the line? Please take this with a grain of salt though, I'm always confused by class metadata's defaults

alcaeus: Ah, didn't think about `_doctrine_class_name` - especially since that doesn't involve a map either. Will have to investigate.

malarzm: Actually I think that map can be combined with default discriminator field since it only provides string <=> FQCN mapping ;)

alcaeus: After looking at the code, this looks safe: `_doctrine_class_name` is not used at the class level, which this PR is trying to fix. Instead it's used when an association contains no `targetDocument` and no `discriminatorField` has been set - in this case the `discriminatorField` is simply set to `_doctrine_class_name` as if it were used in the mapping. I think we should be safe merging this.

malarzm: :+1:

PR #1563, opened 2017-02-20
Fix ->mod() example, add ()->addAnd description

svencan:
Two fixes in Mongo ODM documentation for QueryBilder - Conditional Operators:
 * The example of `mod` is `field('field')->mod('field', array(10, 1))` while the signature is `public function mod($divisor, $remainder = 0)`.
 * Also added an example for `addAnd` which is useful for nested conditions.

### DIFF
--- a/docs/en/reference/query-builder-api.rst
+++ b/docs/en/reference/query-builder-api.rst
@@ -406,10 +406,15 @@ Query for active administrator users:
 .. code-block:: php
 
     <?php
-
+    
     $qb = $dm->createQueryBuilder('User')
         ->field('type')->equals('admin')
         ->field('active')->equals(true);
+    
+    // Or using addAnd()
+    $qb = $dm->createQueryBuilder('User');
+    $qb->addAnd($qb->expr()->field('type')->equals('admin'));
+    $qb->addAnd($qb->expr()->field('active')->equals(true));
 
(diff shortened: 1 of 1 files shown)


docs/en/reference/query-builder-api.rst
```diff
@@ -406,10 +406,15 @@ Query for active administrator users:
 .. code-block:: php
 
     <?php
-
+
```
alcaeus: Can you remove the trailing spaces?

docs/en/reference/query-builder-api.rst
```diff
@@ -406,10 +406,15 @@ Query for active administrator users:
 .. code-block:: php
 
     <?php
-
+    
     $qb = $dm->createQueryBuilder('User')
         ->field('type')->equals('admin')
         ->field('active')->equals(true);
+    
+    // Or using addAnd()
```
alcaeus: I don't think this is a good example for `addAnd`, since it is just a more complicated way of writing the above query. How about using `addAnd` to combine multiple `$or` clauses?

alcaeus:
@svencan could you work in the feedback I gave above or is this pull request abandoned?

alcaeus:
Pull request seems abandoned - if anyone wants to continue working on this, feel free to work in review feedback.

Issue #1560, opened 2017-02-09
Suggestion: New type of reference to deprecate usage of dbref


Nilz11:
Hi,

I have another question.

First of all some explanation:
DbRefs are more or less on the way to being deprecated by mongodb, on the page about database references in most cases it recommends to use simple references and let the reference be resolved by the application.
Also new features in mongodb do not work with DbRefs, e.g. some map-reduce functionality and $lookup. 

See here for example here: https://jira.mongodb.org/browse/SERVER-14466
"[...] we've never formally deprecated use of DBRef. The DBRef convention is **still** widely used in ODM libraries."

Also the way it is now with a discriminator field inside the dbref for references specified, it is hard to query and a lot of tools do not support modifying such documents (e.g. robomongo). To query for an array of references which point to different tables, you need to either use $elemMatch which is slow compared to an $eq match or you need to know if there is a discriminator field inside the document and build the query accordingly (equal with $id, $ref or with $id, $ref and _doctrine_class_name).

Suggestion / question:
Can't we implement a new type of reference, next to id, dbRef and dbRef with $db, which handles relations in a simple and consistent way. E.g. there could be the class name (or some short form of it) and the object id stored inside a reference object and doctrine could resolve the relations.

Then also features like $lookup and using references inside map-reduce would work, querying would be simpler and in some cases even faster (depending on implementation).

Thanks for your comments :) 
Nilz

alcaeus:
I've thought about that before, especially with the limitation about using DBRef objects in `$lookup` stages. It would most likely still be quite similar to the current DBRef object, besides the `$` prefix:
```
{
    ref: 'collection',
    id: ObjectId(...),
    discriminatorField: 'value'
}
```

I've added this to the 2.0 milestone, we can always push it forward if we want to.

Nilz11:
Would be cool to have it earlier. I can also help to make changes, but probably need deeper understanding of the Doctrine internals first.

About the structure to have it like the original DBRef makes sense for me, in order to have the real collection name for lookup stages. The problem with discriminator field inside ref mentioned above would disappear then but there could be still some more improvement inside the query building:

Method includesReferenceTo at the moment uses $elemMatch which is less effective than using an $eq search instead. $elemMatch uses a collection scan while $eq can use an index scan. Building the query using $eq just is a little bit more complicated because we need to check whether there is a discriminator field on the reference or not and do the query appropriately.

alcaeus:
[...]

PR #1566, opened 2017-03-02
[WIP] Feature/ref without dbref

Nilz11:
See #1560

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataInfo.php
@@ -103,6 +103,8 @@ class ClassMetadataInfo implements \Doctrine\Common\Persistence\Mapping\ClassMet
     const REFERENCE_STORE_AS_ID = 'id';
     const REFERENCE_STORE_AS_DB_REF = 'dbRef';
     const REFERENCE_STORE_AS_DB_REF_WITH_DB = 'dbRefWithDb';
+    const REFERENCE_STORE_AS_REF = 'ref';
+    const REFERENCE_STORE_AS_REF_WITH_DB = 'refWithDb';
 
     /* The inheritance mapping types */
     /**
@@ -465,6 +467,33 @@ class ClassMetadataInfo implements \Doctrine\Common\Persistence\Mapping\ClassMet
     {
         $this->name = $documentName;
         $this->rootDocumentName = $documentName;
+    }
+
+    /**
+     * Helper method to get reference id of ref* type references
+     * @param array  $reference
+     * @param string $storeAs
+     * @return mixed
+     * @internal
+     */
+    public static function getReferenceId($reference, $storeAs)
+    {
+        return $reference[ClassMetadataInfo::getReferencePrefix($storeAs) . 'id'];
+    }
+
+    /**
+     * Returns the reference prefix used for a referene
+     * @param string $storeAs
+     * @return string
+     * @internal
+     */
+    public static function getReferencePrefix($storeAs)
+    {
+        if ($storeAs === ClassMetadataInfo::REFERENCE_STORE_AS_REF_WITH_DB || $storeAs === ClassMetadataInfo::REFERENCE_STORE_AS_REF) {
+            return '';
+        } else {
+            return '$';
+        }
     }
 
     /**
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -722,7 +722,7 @@ class DocumentPersister
                 $mongoId = $reference;
             } else {
                 $className = $this->uow->getClassNameForAssociation($mapping, $reference);
-                $mongoId = $reference['$id'];
+                $mongoId = ClassMetadataInfo::getReferenceId($reference, $mapping['storeAs']);
             }
             $id = $this->dm->getClassMetadata($className)->getPHPIdentifierValue($mongoId);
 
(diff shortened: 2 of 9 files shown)


Nilz11:
Any comment on this so far?

alcaeus:
Sorry, I haven't gotten to it yet. I'll make some time for an initial review tonight.

Nilz11:
Ok, just want to know if the general approach is ok like that. Still want to do testing and some minor modifications.

alcaeus:
No specifics yet, just one thing I was wondering: do we really need to make a distinction between `ref` and `refWithDb`? For normal references, this isn't necessary since we'll have a `ClassMetadata` object tell us what DB the referenced object is in. The only time we'll need the `db` field in the reference is when there is no `targetDocument` set on the reference, in which case we can automatically add the `db` field as necessary. FWIW, the same would work for the `DbRef` strategy.

While I like the idea of storing references in documents that can be used in aggregation pipeline stages, I'm not entirely sure adding two new reference strategies is the way to go. Maybe we could reduce it to one, maybe we could just add `id`, `ref` and (optionally) `db` fields to the DbRef object for use in aggregation pipeline stages?

Nilz11:
Ok I can change that appropriately, didn't know that there is a db field inside the class metadata.

So if no target document is set, we have the following fields: `id`, `ref`, `db` and the discriminator or `_doctrine_class_name`. Otherwise just id and ref.

Nilz11:
Also I thought of providing a migration command using the native mongo driver, would be quite easy to implement and could migrate all relations inside database.

Nilz11:
So depending on the release version and general release strategy, if it's 2.0 DbRef could be removed completely then with the need to migrate and in 1.x it could be deprecated.

Nilz11:
One more comment about optional db field, in most cases refs are inside the same db, so if we just make one type, there would be always an unnecessary db value inside the refs without target document

malarzm:
I know that multi-tenant apps aren't that popular but here goes my 2 cents:

> For normal references, this isn't necessary since we'll have a ClassMetadata object tell us what DB the referenced object is in. The only time we'll need the db field in the reference is when there is no targetDocument set on the reference

Please don't infer `db` from `ClassMetadata` blindly, I was working on a multi-tenant application where we weren't specifying db as it was dynamic (depending on context set during request handling) and were relying on default database set for `DocumentManager` (so in practice we had our own `DocumentManagerFactory` that was producing different instances for each tenant, but all document managers were sharing documents metadata, event listeners and so on, the only difference was that default database setting in configuration).

> One more comment about optional db field, in most cases refs are inside the same db, so if we just make one type, there would be always an unnecessary db value inside the refs without target document

I see "in most cases", just wanted to let you know that my app was outside of that "most cases" :) We were utilizing `ReferenceMany` with and without `targetDocument` targeting multiple other databases.

Nilz11:
With the last comment I wanted to say, that it could make sense to have two types of refs one with db and one without. So if you have multiple databases you can use this type, otherwise just the normal one which leaves out the db field and so you do not have unnecessary information inside the database (even if you do not specify target document).

Nilz11:
In the end that's not a decision I can make. My proposal would be to make everything similar to current implementation. Changing this behaviour later is easily possible

malarzm:
I'm marking this for 1.2 so we could already deprecate DBRefs and remove them in 2.0

alcaeus:
Preparing for a merge into 1.2, so let's solve a few issues.

> Please don't infer db from ClassMetadata blindly, I was working on a multi-tenant application where we weren't specifying db as it was dynamic (depending on context set during request handling) and were relying on default database set for DocumentManager (so in practice we had our own DocumentManagerFactory that was producing different instances for each tenant, but all document managers were sharing documents metadata, event listeners and so on, the only difference was that default database setting in configuration).

We have to trust our metadata. If the db property of two metadata instances is the same we can't but assume that they are stored in the same database. That being said, I'm fine with leaving that magic behavior out for now.

> I'm marking this for 1.2 so we could already deprecate DBRefs and remove them in 2.0

Given our last hangout, wen can still deprecate DBRef, but I wouldn't remove them in 2.0 just yet, since the migration could be quite painful.

@Nilz11 could you give this branch a rebase against master? Thanks!

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataInfo.php
```diff
@@ -466,6 +468,30 @@ public function __construct($documentName)
...
 
     /**
+     * Helper method to get reference id of ref* type references
+     * @param array  $reference
+     * @param string $storeAs
+     * @return \MongoId
+     */
+    public static function getReferenceId($reference, $storeAs)
+    {
+        if ($storeAs === ClassMetadataInfo::REFERENCE_STORE_AS_REF_WITH_DB || $storeAs === ClassMetadataInfo::REFERENCE_STORE_AS_REF) {
```
alcaeus: This method should just use `getReferencePrefix`

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -1161,7 +1166,7 @@ private function prepareQueryElement($fieldName, $value = null, $class = null, $
 
         // Prepare DBRef identifiers or the mapped field's property path
         $fieldName = ($objectPropertyIsId && ! empty($mapping['reference']) && $mapping['storeAs'] !== ClassMetadataInfo::REFERENCE_STORE_AS_ID)
-            ? $e[0] . '.$id'
+            ? $e[0] . '.'.ClassMetadataInfo::getReferencePrefix($mapping['storeAs']).'id'
```
alcaeus: Spaces between strings and `.` please.

alcaeus:
@Nilz11 gave your branch a quick rebase, fixed checkstyles and squashed commits. I'll take another look in the coming days and finish this up. Thanks for your help! 🍻

alcaeus:
Superseded by #1623.

PR #1567, opened 2017-03-06
Fix invalid info about event arguments

Steveb-p:
Events were changed sometime ago and actually have arguments

### DIFF
--- a/docs/en/reference/events.rst
+++ b/docs/en/reference/events.rst
@@ -203,9 +203,10 @@ listeners:
 
 - 
    Lifecycle Callbacks are methods on the document classes that are
-   called when the event is triggered. They receive absolutely no
-   arguments and are specifically designed to allow changes inside the
-   document classes state.
+   called when the event is triggered. They receive instances
+   of ``Doctrine\ODM\MongoDB\Event\LifecycleEventArgs`` (see relevant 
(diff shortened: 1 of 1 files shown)


docs/en/reference/events.rst
```diff
@@ -203,9 +203,10 @@ listeners:
 
 - 
    Lifecycle Callbacks are methods on the document classes that are
-   called when the event is triggered. They receive absolutely no
-   arguments and are specifically designed to allow changes inside the
-   document classes state.
+   called when the event is triggered. They may receive receive instances
```
malarzm: How about

> They receive instance of ``Doctrine\ODM\MongoDB\Event\LifecycleEventArgs`` as the only argument (see relevant examples below) and are specifically [...]

IMO "may" suggests the argument might not be passed and I think we're always passing it

malarzm:
Ah, would you also mind squashing commits?

docs/en/reference/events.rst
Steveb-p: Wasn't sure if all events now receive objects. Will update & squash.

Steveb-p:
Squashed & included your comment. Or the other way around ;)

malarzm:
Thanks a bunch!

PR #1572, opened 2017-03-23
Fix counting of inverse collections with limit

alcaeus:
Fixes an issue where having an inverse reference with a limit set would return the wrong number of elements when the collection isn't initialized.

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1572Test.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1572Test.php
@@ -1,0 +1,81 @@
+<?php
+
+namespace Doctrine\ODM\MongoDB\Tests\Functional\Ticket;
+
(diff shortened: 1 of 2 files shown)


malarzm:
@alcaeus I was to click the "Merge" button but noticed that travis is not really happy, could you look into that?

alcaeus:
Yeah, I'll have to check what dependency needs a bump for this to work.

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1572Test.php
```diff
@@ -0,0 +1,81 @@
...
+    }
+}
+
+/** @ODM\Document */
+class GH1572Blog
+{
+    /** @ODM\Id */
+    public $id;
+
+    /** @ODM\ReferenceMany(targetDocument=GH1572Post::class, mappedBy="blog") */
```
malarzm: I guess that `doctrine/annotations` in lowest allowed version are not happy with this

PR #1577, opened 2017-04-20
Support priming references in inverse many relationships

alcaeus:
This PR adds support for reference priming in inverse collections. This avoids having to declare a `repositoryMethod` just to use primers, e.g.:

```php
// Document
    /**
     * @ODM\ReferenceMany(
     *     targetDocument=User\Follow::class,
     *     repositoryMethod="getFollowers",
     *     sort={"createdAt"="desc"},
     *     cascade={"remove"},
     * )
     */
    protected $followers;

// Repository
    public function getFollowers(User $user)
    {
        return $this->createQueryBuilder()
            ->field('followed')
            ->equals($user)
            ->field('follower')
            ->prime()
            ->getQuery()
            ->getIterator();
    }
```

This can now be shortened using the new `primers` property:
```php
    /**
     * @ODM\ReferenceMany(
     *     targetDocument=User\Follow::class,
     *     mappedBy="followed",
     *     primers={"follower"=true}
     *     sort={"createdAt"="desc"},
     *     cascade={"remove"},
     * )
     */
    protected $followers;
```

While primers can technically be closures, this is not supported by the mapping drivers at this time.

### DIFF
--- a/doctrine-mongo-mapping.xsd
+++ b/doctrine-mongo-mapping.xsd
@@ -157,6 +157,7 @@
       <xs:element name="default-discriminator-value" type="odm:default-discriminator-value" minOccurs="0"/>
       <xs:element name="sort" type="odm:sort-map" minOccurs="0" />
       <xs:element name="criteria" type="odm:criteria-map" minOccurs="0" />
+      <xs:element name="prime" type="odm:primers" minOccurs="0" />
     </xs:sequence>
     <xs:attribute name="target-document" type="xs:string" />
     <xs:attribute name="collection-class" type="xs:string" />
@@ -193,6 +194,16 @@
     <xs:sequence>
       <xs:element name="criteria" type="odm:criteria-type" minOccurs="1" maxOccurs="unbounded"/>
     </xs:sequence>
+  </xs:complexType>
+
+  <xs:complexType name="primers">
+    <xs:sequence>
+      <xs:element name="field" type="odm:primer-field" minOccurs="1" maxOccurs="unbounded"/>
+    </xs:sequence>
+  </xs:complexType>
+
+  <xs:complexType name="primer-field">
+    <xs:attribute name="name" type="xs:NMTOKEN" use="required"/>
   </xs:complexType>
 
   <xs:complexType name="emptyType"/>
--- a/lib/Doctrine/ODM/MongoDB/Mapping/Annotations/ReferenceMany.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/Annotations/ReferenceMany.php
@@ -42,10 +42,11 @@ final class ReferenceMany extends AbstractField
     public $inversedBy;
     public $mappedBy;
     public $repositoryMethod;
-    public $sort = array();
-    public $criteria = array();
+    public $sort = [];
+    public $criteria = [];
     public $limit;
     public $skip;
     public $strategy = CollectionHelper::DEFAULT_STRATEGY;
     public $collectionClass;
+    public $prime = [];
 }
--- a/lib/Doctrine/ODM/MongoDB/Mapping/Driver/YamlDriver.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/Driver/YamlDriver.php
@@ -270,7 +270,7 @@ class YamlDriver extends FileDriver
     {
         $defaultStrategy = $type == 'one' ? ClassMetadataInfo::STORAGE_STRATEGY_SET : CollectionHelper::DEFAULT_STRATEGY;
         $mapping = array(
-            'cascade'          => isset($reference['cascade']) ? $reference['cascade'] : null,
+            'cascade'          => isset($reference['cascade']) ? $reference['cascade'] : [],
             'orphanRemoval'    => isset($reference['orphanRemoval']) ? $reference['orphanRemoval'] : false,
             'type'             => $type,
             'reference'        => true,
@@ -285,6 +285,7 @@ class YamlDriver extends FileDriver
             'repositoryMethod' => isset($reference['repositoryMethod']) ? (string) $reference['repositoryMethod'] : null,
             'limit'            => isset($reference['limit']) ? (integer) $reference['limit'] : null,
             'skip'             => isset($reference['skip']) ? (integer) $reference['skip'] : null,
+            'prime'            => isset($reference['prime']) ? $reference['prime'] : [],
         );
         if (isset($reference['name'])) {
             $mapping['name'] = $reference['name'];
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -22,9 +22,11 @@ namespace Doctrine\ODM\MongoDB\Persisters;
 use Doctrine\Common\EventManager;
 use Doctrine\Common\Persistence\Mapping\MappingException;
 use Doctrine\MongoDB\CursorInterface;
+use Doctrine\MongoDB\EagerCursor;
 use Doctrine\ODM\MongoDB\Cursor;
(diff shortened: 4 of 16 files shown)


lib/Doctrine/ODM/MongoDB/Mapping/Annotations/ReferenceMany.php
```diff
@@ -42,10 +42,11 @@
...
     public $repositoryMethod;
-    public $sort = array();
-    public $criteria = array();
+    public $sort = [];
+    public $criteria = [];
     public $limit;
     public $skip;
     public $strategy = CollectionHelper::DEFAULT_STRATEGY;
     public $collectionClass;
+    public $primers = [];
```
malarzm: How about calling it `prime` to be on par with query builder's method?

tests/Doctrine/ODM/MongoDB/Tests/Functional/ReferencePrimerTest.php
```diff
@@ -532,7 +535,91 @@ public function testPrimeEmbeddedReferenceTwoLevelsDeep()
...
+        $this->dm->flush();
+        $this->dm->clear();
+
+        /** @var BlogPost $post */
+        $post = $this->dm->find(BlogPost::class, $post->id);
+        $this->assertInstanceOf(BlogPost::class, $post);
+
+        $comment = $post->repoCommentsEager->first();
+        $this->assertInstanceOf(Proxy::class, $comment->author);
+        $this->assertTrue($comment->author->__isInitialized());
```
malarzm: Should we also assert numbers of queries priming should generate or not really?

lib/Doctrine/ODM/MongoDB/Mapping/Driver/YamlDriver.php
```diff
@@ -270,7 +270,7 @@ private function addMappingFromReference(ClassMetadataInfo $class, $fieldName, $
     {
         $defaultStrategy = $type == 'one' ? ClassMetadataInfo::STORAGE_STRATEGY_SET : CollectionHelper::DEFAULT_STRATEGY;
         $mapping = array(
-            'cascade'          => isset($reference['cascade']) ? $reference['cascade'] : null,
+            'cascade'          => isset($reference['cascade']) ? $reference['cascade'] : [],
```
malarzm: Why this changed?

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -863,6 +868,16 @@ public function createReferenceManyWithRepositoryMethodCursor(PersistentCollecti
             throw new \BadMethodCallException("Expected repository method {$repositoryMethod} to return a CursorInterface");
         }
 
+        if (!empty($mapping['primers']) && $cursor instanceof Cursor) {
+            $referencePrimer = new ReferencePrimer($this->dm, $this->dm->getUnitOfWork());
+
+            try {
+                $cursor->enableReferencePriming(array_filter($mapping['primers']), $referencePrimer);
+            } catch (\BadMethodCallException $e) {
+                // Silence exception handling
```
malarzm: Maybe an exception here would be good?

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/ReferenceMany.php
alcaeus: The original idea was to use `primers` because you could theoretically supply a priming function for a field. I could change it to `prime`, in which case I would also change it to a simple list of field names instead of the current field => primer hash currently in use.

tests/Doctrine/ODM/MongoDB/Tests/Functional/ReferencePrimerTest.php
alcaeus: I could, not sure it's necessary though. What do you think?

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
alcaeus: The exception occurs if the `Cursor` object we're receiving doesn't support reference priming, which is the case if it's not an eager cursor. There's nothing we can do at this point and I thought it's best to silence the exception since the collection can be correctly initialized, except that the primers haven't been run. I can remove the `catch` if you think it's better to let the code fail.

lib/Doctrine/ODM/MongoDB/Mapping/Driver/YamlDriver.php
alcaeus: This messes up due to the shared parent class of `YamlDriverTest` and `XmlDriverTest`. The XML driver always passes an array for `cascade` while the Yaml driver passes null if not given in the mapping. I thought it best to fix this in the driver but can work around it in the tests if you prefer.

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/ReferenceMany.php
malarzm: I'd vote for simple list then, I don't see how current mapping could allow for closures anyhow :)

tests/Doctrine/ODM/MongoDB/Tests/Functional/ReferencePrimerTest.php
malarzm: Not sure either :) Let's skip is then

lib/Doctrine/ODM/MongoDB/Mapping/Driver/YamlDriver.php
malarzm: No need, asking since the change was unexpected

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
malarzm: I think it would be good as people who are not using eager cursors can slap `primers` to mapping and then complain it's not working as advertised... effectively this could save us invalid issues and linking to documentation from them :)

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/ReferenceMany.php
alcaeus: Makes sense, I don't think we would want to read closures from XML mapping anyways. From personal experience, supplying a custom primer closure is hardly ever necessary.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
alcaeus: Code now throws an exception, docs updated. Fewer questions and invalid issues are a good thing.

doctrine-mongo-mapping.xsd
```diff
@@ -195,6 +196,16 @@
     </xs:sequence>
   </xs:complexType>
 
+  <xs:complexType name="primers">
```
malarzm: I think this should be `prime` too

malarzm: Ah, this is only internal, up to you :)

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -863,6 +868,13 @@ public function createReferenceManyWithRepositoryMethodCursor(PersistentCollecti
             throw new \BadMethodCallException("Expected repository method {$repositoryMethod} to return a CursorInterface");
         }
 
+        if (!empty($mapping['prime']) && $cursor instanceof Cursor) {
```
malarzm: Should something happen when `$cursor instanceof Cursor` part is not true?

alcaeus: Good point - it now throws a `BadMethodCallException`.

PR #1587, opened 2017-05-01
Transactions and concurrency documentation

bmichalski:
I was looking for the ODM equivalent of http://doctrine-orm.readthedocs.org/en/latest/reference/transactions-and-concurrency.html, but could not find one.

Basically I took the ORM documentation on the subject which I adapted to the ODM, according to both the official MongoDB documentation and my own experience working with Mongo and the ODM.

### DIFF
--- a/docs/en/reference/transactions-and-concurrency.rst
+++ b/docs/en/reference/transactions-and-concurrency.rst
@@ -30,6 +30,8 @@ Concurrency
 Doctrine MongoDB ODM offers native support for pessimistic and optimistic locking strategies.
 This allows for very fine-grained control over what kind of locking is required for documents in your application.
 
+.. _transactions_and_concurrency_optimistic_locking:
+
 Optimistic Locking
 ~~~~~~~~~~~~~~~~~~
 
@@ -222,6 +224,8 @@ And the change headline action (POST Request):
     $postVersion = (int)$_POST['version'];
 
     $post = $dm->find('BlogPost', $postId, LockMode::OPTIMISTIC, $postVersion);
+
+.. _transactions_and_concurrency_pessimistic_locking:
 
 Pessimistic Locking
 ~~~~~~~~~~~~~~~~~~~
--- a/docs/en/reference/transactions-and-concurrency.rst
+++ b/docs/en/reference/transactions-and-concurrency.rst
@@ -48,6 +48,10 @@ the expected version and version increment are incorporated into the update crit
 If this results in no document being modified by the update (i.e. expected version did not match),
 a ``LockException`` is thrown, which indicates that the document was already modified by another query.
 
+.. note::
+
+    | Versioning can only be used on *root* (top-level) documents.
+
 Document Configuration
 ^^^^^^^^^^^^^^^^^^^^^^
 
--- a/docs/en/reference/transactions-and-concurrency.rst
+++ b/docs/en/reference/transactions-and-concurrency.rst
@@ -286,3 +286,20 @@ You can use pessimistic locks in two different scenarios:
    ``DocumentManager#lock($document, \Doctrine\ODM\MongoDB\LockMode::PESSIMISTIC_WRITE)``
    or
    ``DocumentManager#lock($document, \Doctrine\ODM\MongoDB\LockMode::PESSIMISTIC_READ)``
+
+.. warning::
+
+    | A few things could go wrong:
+    |
+    | If your script fails to complete, for example because of an exception, you might end up with stale locks.
+    | Said locks would then have to be manually released.
+    | Or you would have to devise a strategy to automatically do so.
+    | For PHP exceptions, to prevent the problem, you might want to try / catch your code to handle exceptions gracefully, making sure you unlock your documents before your script ends.
+    |
+    | You could also create `deadlocks <https://en.wikipedia.org/wiki/Deadlock>`_.
+    | Say that you have a P1 process that needs a R1 resource and has locked a R2 resource.
+    | You also have a P2 process that has locked the R1 resource but also needs the R2 resource.
+    | If both processes keep waiting for resources, processing is stuck.
+    | When loading a document, the ODM immediately throws an exception if it is already locked.
+    | If you try / catch the ``LockException``, you could create a deadlock by retrying to acquire the lock an infinite number of times in both processes.
+    | One way to avoid this situation would be to implement a maximum number of retries and automatically release possessed locks when script ends, which would allow one of the processes to end gracefully and the other to complete its task.
--- a/docs/en/reference/transactions-and-concurrency.rst
+++ b/docs/en/reference/transactions-and-concurrency.rst
@@ -1,0 +1,288 @@
+.. Heavily inspired by Doctrine 2 ORM documentation
+
+Transactions and Concurrency
+============================
+
+Transactions
+------------
+
+As per the `documentation <https://docs.mongodb.com/manual/core/write-operations-atomicity/#atomicity-and-transactions>`_, MongoDB
+write operations are "atomic on the level of a single document".
+
+Even when updating multiple documents with a single write operation,
+though the modification of each document is atomic,
+the operation as a whole is not, and other operations may interleave.
+
+Simply put, quoting the `FAQ <https://docs.mongodb.com/manual/faq/fundamentals/#does-mongodb-support-transactions>`_: "MongoDB does not support multi-document transactions".
+
+Neither does Doctrine MongoDB ODM.
+
+Limitation
+~~~~~~~~~~
+At the moment, Doctrine MongoDB ODM does not provide any native strategy to emulate multi-document transactions.
+
+Workaround
+~~~~~~~~~~
+To work around this limitation, one could perform `two phase commits <https://docs.mongodb.com/manual/tutorial/perform-two-phase-commits/>`_.
+
+Concurrency
+-----------
+
+Doctrine MongoDB ODM offers support for Pessimistic- and Optimistic-locking
+strategies natively. This allows to take very fine-grained control
+over what kind of locking is required for your Documents in your
+application.
+
+Optimistic Locking
+~~~~~~~~~~~~~~~~~~
+
+Approach
+^^^^^^^^
+
+Doctrine has integrated support for automatic optimistic locking
+via a ``version`` field. Any document that should be
+protected against concurrent modifications during long-running
+business transactions gets a ``version`` field that is either a simple
+number (mapping type: ``int``) or a date (mapping type:
+``date``).
+When changes are persisted at the end of a long-running conversation,
+the version of the document is added to the query. If no document has been updated by the operation,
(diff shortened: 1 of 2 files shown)


docs/en/reference/transactions-and-concurrency.rst
```diff
@@ -0,0 +1,305 @@
...
+        version:
+          type: date
+          version: true
+
+Choosing the field type
+"""""""""""""""""""""""
+
+When using ``date`` type in a highly concurrent environment, multiple documents could be created with the same version
+and create a conflict.
+This cause of conflict can be avoided by using ``int`` type.
```
jmikola: Rewritten:

> When using the `date` type in a high-concurrency environment, it is still possible to create multiple documents with the same version and cause a conflict. This can be avoided by using the `int` type.

@alcaeus: Since conflict is possible if versions are created within the same millisecond (i.e. date precision), I wonder why we even offer this option. Something to remove in 2.0 perhaps?

docs/en/reference/transactions-and-concurrency.rst
```diff
@@ -0,0 +1,305 @@
...
+``DocumentManager#flush()``, a ``LockException`` is thrown.
+This exception can be caught and handled. Potential responses to a
+``LockException`` are to present the conflict to the user or
+to refresh or reload objects and then retry the update.
+
+With PHP promoting a share-nothing architecture, the time between
+showing an update form and actually modifying the document can in the
+worst scenario be as long as your applications session timeout. If
+changes happen to the document in that time frame you want to know
+directly when retrieving the document that you will hit a locking exception.
```
jmikola: Rewritten:

> With PHP promoting a share-nothing architecture, the worst case scenario for a delay between rendering an update form (with existing document data) and modifying the document after a form submission may be your application's session timeout. If the document is changed within that time frame by some other request, it may be preferable to encounter a `LockException` when retrieving the document instead of executing the update.

docs/en/reference/transactions-and-concurrency.rst
```diff
@@ -0,0 +1,305 @@
...
+
+You can verify the version of a document during a request either when calling ``DocumentManager#find()``:
```
jmikola: > You can specify the expected version of a document during a query with `DocumentManager#find()`:

docs/en/reference/transactions-and-concurrency.rst
```diff
@@ -0,0 +1,305 @@
...
+        $document = $dm->find('User', $theDocumentId, LockMode::OPTIMISTIC, $expectedVersion);
+
+        // do the work
+
+        $dm->flush();
+    } catch(LockException $e) {
+        echo "Sorry, but someone else has already changed this document. Please apply the changes again!";
+    }
+
+Or you can use ``DocumentManager#lock()`` to find out:
```
jmikola: > Alternatively, an expected version may be specified for an existing document with `DocumentManager#lock()`:

docs/en/reference/transactions-and-concurrency.rst
```diff
@@ -0,0 +1,305 @@
...
+-  Alice reads the headline of the blog post being "Foo", at
+   optimistic lock version 1 (GET Request)
+-  Bob reads the headline of the blog post being "Foo", at
+   optimistic lock version 1 (GET Request)
+-  Bob updates the headline to "Bar", upgrading the optimistic lock
+   version to 2 (POST Request of a Form)
+-  Alice updates the headline to "Baz", ... (POST Request of a
+   Form)
+
+Now at the last stage of this scenario the blog post has to be read
```
jmikola: Remove "now".

docs/en/reference/transactions-and-concurrency.rst
```diff
@@ -0,0 +1,305 @@
...
+again from the database before Alice's headline can be applied. At
+this point you will want to check if the blog post is still at
+version 1 (which it is not in this scenario).
+
+Using optimistic locking correctly, you *have* to add the version
+as an additional hidden field (or into the SESSION for more
+safety). Otherwise you cannot verify the version is still the one
+being originally read from the database when Alice performed her
+GET request for the blog post. If this happens you might see lost
+updates you wanted to prevent with Optimistic Locking.
```
jmikola: > In order to correctly utilize optimistic locking, you *must* add the version as hidden form field or, for more security, session attribute. Otherwise, you cannot verify that the version at the time of update is the same as what was originally read from the database when Alice performed her original GET request for the blog post. Without correlating the version across form submissions, the application could lose updates.

docs/en/reference/transactions-and-concurrency.rst
```diff
@@ -0,0 +1,305 @@
...
+    $postVersion = (int)$_POST['version'];
+
+    $post = $dm->find('BlogPost', $postId, LockMode::OPTIMISTIC, $postVersion);
+
+Pessimistic Locking
+~~~~~~~~~~~~~~~~~~~
+
+Doctrine MongoDB ODM supports Pessimistic Locking.
+There is no native MongoDB support for pessimistic locking.
+The Doctrine implementation uses a ``lock`` field, that you have to configure if you wish to use pessimistic locking.
```
jmikola: > Doctrine MongoDB ODM also supports pessimistic locking via a configurable `lock` field. This functionality is implemented entirely by Doctrine; MongoDB has no native support for pessimistic locking.

docs/en/reference/transactions-and-concurrency.rst
```diff
@@ -0,0 +1,305 @@
...
+
+Document configuration
```
jmikola: "Configuration"

docs/en/reference/transactions-and-concurrency.rst
```diff
@@ -0,0 +1,305 @@
...
+^^^^^^^^^^^^^^^^^^^^^^
+
+For pessimistic locking to work, a lock field must be configured.
+The lock field must be of type ``int``.
+You designate a lock field in a document as follows.
```
jmikola: > Pessimistic locking requires a document to designate a lock field using the `int` type:

docs/en/reference/transactions-and-concurrency.rst
```diff
@@ -0,0 +1,305 @@
...
+
+        <field fieldName="lock" lock="true" type="int" />
+
+    .. code-block:: yaml
+
+        lock:
+          type: int
+          lock: true
+
+Lock modes
```
jmikola: "Modes"

docs/en/reference/transactions-and-concurrency.rst
```diff
@@ -0,0 +1,305 @@
...
+^^^^^^^^^^
+
+Doctrine MongoDB ODM currently supports two pessimistic lock modes:
+
+-  Pessimistic Write
+   (``\Doctrine\ODM\MongoDB\LockMode::PESSIMISTIC_WRITE``), locks the
+   underlying document for concurrent Read and Write operations.
```
jmikola: Using a colon after "Pessimistic Write" instead of a comma. Also remove capitalization in "read and write operations."

docs/en/reference/transactions-and-concurrency.rst
```diff
@@ -0,0 +1,305 @@
...
+-  Pessimistic Read (``\Doctrine\ODM\MongoDB\LockMode::PESSIMISTIC_READ``),
```
jmikola: Colon after "Pessimistic Read" instead of comma.

docs/en/reference/transactions-and-concurrency.rst
```diff
@@ -0,0 +1,305 @@
...
+   ``DocumentManager#lock($document, \Doctrine\ODM\MongoDB\LockMode::PESSIMISTIC_READ)``
+
+.. warning::
+
+    | A few things could go wrong:
+    |
+    | If your script fails to complete, for example because of an exception, you might end up with stale locks.
+    | Said locks would then have to be manually released.
+    | Or you would have to devise a strategy to automatically do so.
+    | For PHP exceptions, to prevent the problem, you might want to try / catch your code to handle exceptions gracefully, making sure you unlock your documents before your script ends.
```
jmikola: > If a request fails to complete (e.g. unhandled exception), you may end up with stale locks. Said locks would need to be manually released or you would need to device a strategy to automatically do so. One way to mitigate stale locks after an application error would be to gracefully catch the exception and ensure that relevant documents are unlocked before the request ends.

docs/en/reference/transactions-and-concurrency.rst
```diff
@@ -0,0 +1,305 @@
...
+    |
+    | You could also create `deadlocks <https://en.wikipedia.org/wiki/Deadlock>`_.
+    | Say that you have a P1 process that needs a R1 resource and has locked a R2 resource.
+    | You also have a P2 process that has locked the R1 resource but also needs the R2 resource.
+    | If both processes keep waiting for resources, processing is stuck.
+    | When loading a document, the ODM immediately throws an exception if it is already locked.
+    | If you try / catch the ``LockException``, you could create a deadlock by retrying to acquire the lock an infinite number of times in both processes.
+    | One way to avoid this situation would be to implement a maximum number of retries and automatically release possessed locks when script ends, which would allow one of the processes to end gracefully and the other to complete its task.
```
jmikola: > [Deadlock](https://en.wikipedia.org/wiki/Deadlock>) situations are also possible. Suppose process P1 needs resource R1 and has locked resource R2 and that another process P2 has locked resource R1 but also needs resource R2. If both processes continue waiting for the respective resources, the application will be stuck. When loading a document, Doctrine can immediately throw an exception if it is already locked. A deadlock could be created by endlessly retrying attempts to acquire the lock. One can avoid a possible deadlock by designating a maximum number of retry attempts and automatically releasing any active locks with the request ends, thereby allowing a process to end gracefully while another completes its task.

docs/en/reference/transactions-and-concurrency.rst
alcaeus: Yeah. I'll put it on the deprecations list. Thanks!

bmichalski:
@jmikola thank you for the review, I'm working on fixes.

docs/en/reference/transactions-and-concurrency.rst
```diff
@@ -0,0 +1,305 @@
...
+
+Doctrine has integrated support for automatic optimistic locking
+via a ``version`` field. Any document that should be
+protected against concurrent modifications during long-running
+business transactions gets a ``version`` field that is either a simple
+number (mapping type: ``int``) or a date (mapping type:
+``date``).
+When changes are persisted at the end of a long-running conversation,
+the version of the document is added to the query. If no document has been updated by the operation,
+a ``LockException`` is thrown, indicating that the document has already been modified by another query.
```
bmichalski: Single line break is not rendered as line break when building, hence there is no dedicated paragraph :)

bmichalski:
All requested changes have been integrated :)

malarzm:
Off we go - thanks @bmichalski!

PR #1588, opened 2017-05-03
Doctrine\ODM\MongoDB\LockException when using versioning, a custom name for the version field and embedMany

bmichalski:
I have a versioned document, with a version field, using a custom name.
"version" is the field name I use in the PHP object and "_version" is the name I use in the underlying document.
I have added an embedMany field and can no longer update my document.
A Doctrine\ODM\MongoDB\LockException is thrown.

It seems that the problem comes from the CollectionPersister, that adds the version field name to the query, when it should instead use the related field mapping name.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/CollectionPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/CollectionPersister.php
@@ -268,7 +268,7 @@ class CollectionPersister
         $id = $class->getDatabaseIdentifierValue($this->uow->getDocumentIdentifier($document));
         $query = array('_id' => $id);
         if ($class->isVersioned) {
-            $query[$class->versionField] = $class->reflFields[$class->versionField]->getValue($document);
+            $query[$class->fieldMappings[$class->versionField]['name']] = $class->reflFields[$class->versionField]->getValue($document);
         }
         $collection = $this->dm->getDocumentCollection($className);
         $result = $collection->update($query, $newObj, $options);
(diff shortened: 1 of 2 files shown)


malarzm:
Thanks @bmichalski! Hopefully it'll be one last place of infamous `name` vs `fieldName` issue, but frankly I hope so every time I see such PR :)

malarzm:
Ah, just noticed before pushing merge button, @bmichalski could you please retarget your PR for `1.x` branch? That may require an additional rebase for your commit to be the only one.

bmichalski:
I cherry-picked the commit and changed the base branch to doctrine:1.1.x, is this ok?

malarzm:
That'll do, thanks @bmichalski! Offhand there's neat `rebase --onto` if you'd ever need to do such thing again and had more than 1 commit to move :)

bmichalski:
@malarzm thank you for the advice :)
I'm not familiar with the lifecycle of this project: will there be a 1.1.5 tag soon now that the fix has been merged?

malarzm:
Usually we waited a bit if no more issues would arise, but I can tag new version somewhen this week if it's needed. Also you can always depend on `1.1.x` branch

alcaeus:
I'm a bit behind on my OSS work due to paid work issues, but I'll tag 1.1.5 and 1.2.0 within the next few days. It's long overdue anyways.

alcaeus:
Release tagged - we haven't tagged a release in a couple of months, so it's ok to tag for just one fix.

PR #1592, opened 2017-05-16
Lock exception leaves document with bad revision

mlodick:
A document's local version should only be incremented when the database version is, otherwise, if the lock exception is caught, a second persist & can succeed and overwrite data.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -391,6 +391,7 @@ class DocumentPersister
         // Include versioning logic to set the new version value in the database
         // and to ensure the version has not changed since this document object instance
         // was fetched from the database
+        $nextVersion = null;
         if ($this->class->isVersioned) {
             $versionMapping = $this->class->fieldMappings[$this->class->versionField];
             $currentVersion = $this->class->reflFields[$this->class->versionField]->getValue($document);
@@ -398,12 +399,10 @@ class DocumentPersister
                 $nextVersion = $currentVersion + 1;
                 $update['$inc'][$versionMapping['name']] = 1;
                 $query[$versionMapping['name']] = $currentVersion;
-                $this->class->reflFields[$this->class->versionField]->setValue($document, $nextVersion);
             } elseif ($versionMapping['type'] === 'date') {
                 $nextVersion = new \DateTime();
                 $update['$set'][$versionMapping['name']] = new \MongoDate($nextVersion->getTimestamp());
                 $query[$versionMapping['name']] = new \MongoDate($currentVersion->getTimestamp());
-                $this->class->reflFields[$this->class->versionField]->setValue($document, $nextVersion);
             }
         }
 
@@ -426,6 +425,8 @@ class DocumentPersister
 
             if (($this->class->isVersioned || $this->class->isLockable) && ! $result['n']) {
                 throw LockException::lockFailed($document);
+            } elseif ($this->class->isVersioned) {
+                $this->class->reflFields[$this->class->versionField]->setValue($document, $nextVersion);
             }
         }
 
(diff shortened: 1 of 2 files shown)


alcaeus:
Changes look good, can you create a test for this use-case? Thanks!

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -426,6 +425,8 @@ public function update($document, array $options = array())
 
             if (($this->class->isVersioned || $this->class->isLockable) && ! $result['n']) {
                 throw LockException::lockFailed($document);
+            } else if ($this->class->isVersioned && $nextVersion) {
```
malarzm: We tend to keep `elseif` written together. Also If the class is versioned the `$nextVersion` must be present so there's no need for checking it here (the latter may not be true by looking only at `if` statements you've modifed above but that's validated on [class metadata](https://github.com/doctrine/mongodb-odm/blob/68d4774c09de387a70dc492951f6e38df80ff092/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataInfo.php#L1971) level)

mlodick: fair enough.

malarzm:
@mlodick any chance you could write a test for the change you've implemented?

mlodick:
@malarzm sorry, I've been out of town. I'll get a test written today or tomorrow

alcaeus:
@mlodick could you add that test? I'd like to get this merged next week so we can release it in 1.1.6, otherwise I'd have to push it back to 1.1.7.

mlodick:
@alcaeus are these tests ok?

alcaeus:
@mlodick I'll take a look tonight and update them myself if necessary. Thank you very much! ❤️

tests/Doctrine/ODM/MongoDB/Tests/Functional/DocumentPersisterTest.php
```diff
@@ -466,6 +466,55 @@ public function testDefaultWriteConcernIsRespected()
         $this->dm->persist($testDocument);
         $this->dm->flush();
     }
+
+    public function testVersionIncrementOnUpdateSuccess()
+    {
+        $class = DocumentPersisterTestDocumentWithVersion::class;
+        $documentPersister = $this->uow->getDocumentPersister($class);
+
+        $collection = $this->createMock('\MongoCollection');
```
alcaeus: You can use `MongoCollection::class`

tests/Doctrine/ODM/MongoDB/Tests/Functional/DocumentPersisterTest.php
```diff
@@ -466,6 +466,55 @@ public function testDefaultWriteConcernIsRespected()
...
+            ->will($this->returnValue(array('n' => 0)));
+
+        $reflectionProperty = new \ReflectionProperty($documentPersister, 'collection');
+        $reflectionProperty->setAccessible(true);
+        $reflectionProperty->setValue($documentPersister, $collection);
+
+        $testDocument = new $class();
+        $testDocument->id = 12345;
+        $this->uow->registerManaged($testDocument, 12345, array('id' => 12345));
+        $this->setExpectedException('\Doctrine\ODM\MongoDB\LockException');
```
alcaeus: `::class`

tests/Doctrine/ODM/MongoDB/Tests/Functional/DocumentPersisterTest.php
```diff
@@ -466,6 +466,55 @@ public function testDefaultWriteConcernIsRespected()
...
+
+        $this->assertEquals(2, $testDocument->revision);
+    }
+
+    public function testNoVersionIncrementOnUpdateFailure()
+    {
+        $class = DocumentPersisterTestDocumentWithVersion::class;
+        $documentPersister = $this->uow->getDocumentPersister($class);
+
+        $collection = $this->createMock('\MongoCollection');
```
alcaeus: `::class`

mlodick:
@alcaeus I think this is all set then, though I'm not sure how to retrigger the test that failed - seems to have failed getting a pecl package.

malarzm:
@mlodick just restarted the job, let's see how it ends :)

malarzm:
Green it is, thanks @mlodick!

mlodick:
👍  thank you

PR #1593, opened 2017-05-16
Unset a document update for a document on mongo exception or lock exception

mlodick:
While testing https://github.com/doctrine/mongodb-odm/pull/1592, I found that a caught lock exception results in a document remaining in the unit of work and on the next `flush()` call of that type of document, a second attempt to save the document to the db is made. If that attempt fails, the unit of work is permanently stuck in a bad state, as that's the first document to be updated.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
+++ b/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
@@ -1171,7 +1171,15 @@ class UnitOfWork implements PropertyChangedListener
             $this->lifecycleEventManager->preUpdate($class, $document);
 
             if ( ! empty($this->documentChangeSets[$oid]) || $this->hasScheduledCollections($document)) {
-                $persister->update($document, $options);
+                try {
+                    $persister->update($document, $options);
+                } catch (\MongoException $e) {
+                    unset($this->documentUpdates[$oid]);
+                    throw $e;
+                } catch (LockException $e) {
+                    unset($this->documentUpdates[$oid]);
+                    throw $e;
+                }
             }
 
             unset($this->documentUpdates[$oid]);


malarzm:
I believe that removing document from `documentUpdates` is not the right course of action as quite few other things will not be undone at this point. Haven't used locking much but shouldn't the exception be caught in your application code, faulty document detached/refreshed and then operation retried?

mlodick:
removing from documentUpdates was based on the successful scenario a few lines below. Catching the exception is fine but leaving the unit of work in a bad state doesn't seem ideal to me. 

I'm not super invested personally as for my usecase I only persist and flush a single document at a time, so my solution is to clear out the entire unit of work when an exception is caught.

malarzm:
> removing from documentUpdates was based on the successful scenario a few lines below.

That's true, but there are more things happening [by the end of `UnitOfWork::commit()`](https://github.com/doctrine/mongodb-odm/blob/68d4774c09de387a70dc492951f6e38df80ff092/lib/Doctrine/ODM/MongoDB/UnitOfWork.php#L446-L457) after whole flush is uccessful and I'm worried that trying to fix UoW's state in the place you're suggesting will have some other side effects we won't foresee.

mlodick:
Are there additional changes you are suggesting, or do you just think this isn't a good solution? 

The most painful consequence of the current behavior is that if a lockException is caught and the document that failed is not detached/refreshed, and it lingers in the `documentUpdates`, then a subsequent flush *even when flushing a single unrelated document* will fail, with no indication that this is due to the previously failing update.

In my implementation I wrapped `UnitOfWork::commit()` in a try, and when catching the exception I do all of the unsets here: https://github.com/doctrine/mongodb-odm/blob/3cea3c9ddc8550733054bb8152b51c9fe95e21f1/lib/Doctrine/ODM/MongoDB/UnitOfWork.php#L446-L457, but that won't work unless you're only flushing a single document at a time.

malarzm:
> then a subsequent flush even when flushing a single unrelated document will fail, with no indication that this is due to the previously failing update.

On the same note calling flush again (without specifying any document) may update more documents if there was something left to be updated after the document for which the `LockException` was thrown. Won't bet my head on it but this should also happen when flushing single document as there will still be other documents in `documentUpdates` and that could be even more surprising.

Offhand, flushing single document can flush more data: basically all inserts and upserts will be executed too.

> Are there additional changes you are suggesting, or do you just think this isn't a good solution?

I'm afraid there is no good solution to the problem you're describing, mostly due to current UnitOfWork implementation and it's all-around statefulness and not cleaning things immediately (for instance whole "clean up" section I was linking to before). Also just took a look at ORM - in case of an error during commit entire EntityManager is shut down without any easy way of re-opening it (IIRC the only way was to instantiate new EntityManager) - we may consider such scenario but it's quite radical.

@alcaeus any thoughts on this?

alcaeus:
Currently, an exception in the write process will leave the unit of work in a state where you should just be able to `flush` a second time and commit the rest of the changes - given that you've removed the original cause for the exception.

> The most painful consequence of the current behavior is that if a lockException is caught and the document that failed is not detached/refreshed

Here's your problem: you're supposed to handle the exception. If UoW tells you that your document is out of date, you'll need to do something about it - either detach the document or refresh it from the database. Due to the statefulness of UnitOfWork we can't just throw away all pending document insertions, changesets, etc. since we don't know what has been executed and what hasn't.

> then a subsequent flush *even when flushing a single unrelated document* will fail

There is no "flushing a single document". When you pass one or more documents to `flush`, you're instructing UnitOfWork to only compute changesets for these documents, not only write them. Any previously computed changesets (e.g. document insertions, orphan removals, document deletions) are still executed - it's just that any managed documents without a changeset that are not in the given list are ignored. FWIW, this misconception is the exact reason why we're going to remove this in 2.0 - along with ORM most likely removing it in ORM 3.0.

At this time, I'm not sure what the best course of action would be. Shutting down UnitOfWork and DocumentManager is quite radical and I'm not sure we would want this. On the other hand, I don't have any better ideas except "make sure you handle every exception properly" which can quite hard to do, especially since you don't know what will be done when you call `flush`. This is especially true when using the default implicit change tracking policy, which will just compute changesets for every document known to the UnitOfWork.

I'll tag this for the 2.0 milestone for now, since that will most likely require large changes to the document persister logic, so we might be able to fit this in as well.

mlodick:
```

alcaeus:
Closing here: after giving this some more thought I've concluded that this is something the user needs to fix: either refresh the offending document or clear the UnitOfWork manually, but simply unsetting that one document update is not going to work. The `LockException` contains the offending document, so I'll leave it up to application code to catch this exception and handle it accordingly.

PR #1598, opened 2017-05-24
Update default repository factory

malarzm:
Lately I realized that our implementation is a bit off compared to ORM's one so here are 2 bug-fixes (since they're touching the same class and even in same place I packed both into same PR). Once we merge this I'd add `AbstractRepositoryFactory` as writing own one is quite tedious (copy default one and add your instantiation logic, I think we could add something like [AbstractPersistentCollectionFactory](https://github.com/doctrine/mongodb-odm/blob/b5d0c89cd63b3e05f58071c1025f7b776fcaa157/lib/Doctrine/ODM/MongoDB/PersistentCollection/AbstractPersistentCollectionFactory.php))


malarzm:
> Also, if you want this to be in 1.1.6 you need to target 1.1.x branch

Ah, branched off locally from `1.1.x` and then forgot to set the dropdown... I bet such things could be detected by GitGub

PR #1599, opened 2017-05-24
add info on hint() method

Xosofox:
hint was not mentioned before

### DIFF
--- a/docs/en/reference/query-builder-api.rst
+++ b/docs/en/reference/query-builder-api.rst
@@ -185,6 +185,25 @@ the ``select()`` method:
 In the results only the data from the username and password will be
 returned.
 
+Index hints
+~~~~~~~~~~~
+
+You can force MongoDB to use a specific index for a query with the ``hint()`` method (see `hint <https://docs.mongodb.com/manual/reference/operator/meta/hint/>`_)
+
+.. code-block:: php
+
+    <?php
+
+    $qb = $dm->createQueryBuilder('User')
+        ->hint('user_pass_idx');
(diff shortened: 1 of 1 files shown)


docs/en/reference/query-builder-api.rst
```diff
@@ -185,6 +185,25 @@ the ``select()`` method:
 In the results only the data from the username and password will be
 returned.
 
+Index hints
+~~~~~~~~~~~
+
+You can help mongo in finding the right index to use with the ``hint()`` method:
```
malarzm: `$hint` is rather forcing Mongo to use specified index, think it's worth reflecting and linking to https://docs.mongodb.com/manual/reference/operator/meta/hint/

docs/en/reference/query-builder-api.rst
```diff
@@ -185,6 +185,25 @@ the ``select()`` method:
 In the results only the data from the username and password will be
 returned.
 
+Index hints
+~~~~~~~~~~~
+
+You can force mongo to use a specific index for a query with the ``hint()`` method (see `hint <https://docs.mongodb.com/manual/reference/operator/meta/hint/>`_)
```
alcaeus: You should use "MongoDB", not "mongo"

Xosofox: You're right, that's not very polite

malarzm:
@Xosofox LGTM, would be great if you could squash your commits (or I can do that later today)

malarzm:
Thanks @Xosofox!

PR #1600, opened 2017-05-26
Add AbstractRepositoryFactory

malarzm:
As mentioned earlier, writing own Repository factory seemed tedious so I propose to introduce an abstract factory that will ease the process and allow developer to focus on instantiation itself.


lib/Doctrine/ODM/MongoDB/Repository/AbstractRepositoryFactory.php
```diff
@@ -0,0 +1,88 @@
...
+ * @since 1.2
+ */
+abstract class AbstractRepositoryFactory implements RepositoryFactory
+{
+    /**
+     * The list of DocumentRepository instances.
+     *
+     * @var array<\Doctrine\Common\Persistence\ObjectRepository>
+     */
+    private $repositoryList = array();
```
alcaeus: Short array syntax

lib/Doctrine/ODM/MongoDB/Repository/AbstractRepositoryFactory.php
```diff
@@ -0,0 +1,88 @@
...
+     *
+     * @return \Doctrine\Common\Persistence\ObjectRepository
+     */
+    protected function createRepository(DocumentManager $documentManager, $documentName)
+    {
+        $metadata            = $documentManager->getClassMetadata($documentName);
+        $repositoryClassName = $metadata->customRepositoryClassName;
+
+        if ($repositoryClassName === null) {
+            $configuration       = $documentManager->getConfiguration();
```
alcaeus: IMO you can skip the temporary variable if it's just used once, might even use the `?:` operator to skip the `if` completely.

lib/Doctrine/ODM/MongoDB/Repository/AbstractRepositoryFactory.php
```diff
@@ -0,0 +1,88 @@
...
+        return $repository;
+    }
+
+    /**
+     * Create a new repository instance for a document class.
+     *
+     * @param DocumentManager $documentManager The DocumentManager instance.
+     * @param string          $documentName    The name of the document.
+     *
+     * @return \Doctrine\Common\Persistence\ObjectRepository
```
alcaeus: `DocumentManager::getRepository` specifies the return type hint as `DocumentRepository`, so we should either do the same here or loosen the return type in DocumentManager. Personally I'd type all these variables to `DocumentRepository` and use an interface in 2.0.

malarzm:
@alcaeus agree with your comments, will apply them soon. Frankly I've just copy-pasted what was in original `DefaultRepositoryFactory` without thinking much :)

malarzm:
Offhand I've also started writing some docs about this, should be ready on Monday

malarzm:
@alcaeus actually I think that instead of narrowing `ObjectRepository` to `DocumentRepository` we should do the opposite, for instance I've found [such check](https://github.com/doctrine/mongodb-odm/blob/d27371789755bab701da32dc57af3e5c88346317/lib/Doctrine/ODM/MongoDB/Configuration.php#L513)

alcaeus:
Oh, good find. In that case, make it ObjectRepository and let's look at that again for 2.0. We might want to use our own interface depending on the functionality.

malarzm:
@alcaeus done

mdjaman:
@malarzm doc needed please!

malarzm:
@mdjaman the docs were updated long time ago: https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/1.2/reference/document-repositories.html#repositories-with-additional-dependencies

PR #1601, opened 2017-05-29
Add documentation for repositories

malarzm:


docs/en/reference/document-repositories.rst
```diff
@@ -0,0 +1,203 @@
...
+Following methods are available for default repository:
+
+- ``find()`` - finds one document by its identifier. This may skip database query
+if document is already known to ODM.
+- ``findAll()`` - gets all documents.
+- ``findBy()`` - gets all documents matching given criteria. Additionally documents
+can be sorted and a pagination applied.
+- ``findOneBy()`` - finds a document matching all given criteria.
+- ``matching()`` - finds all documents matching given criteria provided with Doctrine's
+Criteria API.
```
jmikola: > Finds all documents matching the given criteria, as expressed with Doctrine's Criteria API.

Can we link Criteria API to an existing documentation page?

docs/en/reference/document-repositories.rst
```diff
@@ -0,0 +1,203 @@
...
+
+Sometimes one may encounter method call that is not defined explicitly by
+``DocumentRepository`` but works thanks to convention:
```
jmikola: > Additional methods that are not defined explicitly in the repository class may also be used if they follow a specific naming convention:

docs/en/reference/document-repositories.rst
```diff
@@ -0,0 +1,203 @@
...
+    $group = $documentManager->find(Group::class, 123);
+    /* @var $repository \Doctrine\ODM\MongoDB\DocumentRepository */
+    $repository = $documentManager->getRepository(User::class);
+    $usersInGroup = $repository->findByGroup($group);
+    $randomUser = $repository->findOneByStatus('active');
+
+``findByGroup()`` and ``findOneByStatus()`` calls above are working thanks to magic
+``__call`` method which intercepts calls to unknown methods. Given invoked method's name
+starts with "findBy" or "findOneBy" ODM takes the rest ("Group" or "Status" as per example)
+and checks if such property was mapped. Above calls are equivalent to
```
jmikola: > In the above example, `findByGroup()` and `findOneByStatus()` will be handled by the `__call` method, which intercepts calls to undefined methods. If the invoked method's name starts with "findBy" or "findOneBy", ODM will attempt to infer mapped properties from the remainder of the method name ("Group" or "Status" as per example). The above calls are equivalent to:

docs/en/reference/document-repositories.rst
```diff
@@ -0,0 +1,203 @@
...
+    $group = $documentManager->find(Group::class, 123);
+    /* @var $repository \Doctrine\ODM\MongoDB\DocumentRepository */
+    $repository = $documentManager->getRepository(User::class);
+    $usersInGroup = $repository->findBy(['group' => $group]);
+    $randomUser = $repository->findOneBy(['status' => 'active']);
+
+.. note::
+
+    We do not endorse using magic ``findBy`` or ``findOneBy`` calls and strongly advocate
+    for using custom repositories with properly defined methods.
```
jmikola: > Users are strongly encouraged to use custom repository classes with explicitly defined methods instead of relying on this magic "findBy" and "findOneBy" behavior.

Is this something we expect will be removed in a future version? /cc @Ocramius

docs/en/reference/document-repositories.rst
```diff
@@ -0,0 +1,203 @@
...
+
+Custom Repositories
+-------------------
+
+Custom repository gives ability to concentrate all filtering logic in one class instead
+of having it spread all over the project. To indicate that particular document has
+a repository other than the default one you need to state that in the mapping:
```
jmikola: > A custom repository allows filtering logic to be consolidated into a single class instead of spreading it throughout a project. A custom repository class may be specified for a document class like so:

docs/en/reference/document-repositories.rst
```diff
@@ -0,0 +1,203 @@
...
+    .. code-block:: yaml
+
+        Documents\User:
+            repositoryClass: Repositories\\UserRepository
+            collection: user
+            # ...
+
+Next step is implementing your repository. Usually it is enough to extend ODM's default
+``DocumentRepository`` and only add methods you need. For more hand-crafted solutions
+that would require passing additional dependencies please see next chapter.
```
jmikola: > The next step is implementing your repository class. In most cases, ODM's default `DocumentRepository` class may be extended with additional methods that you need. More complex cases that require passing additional dependencies to a custom repository class will be discussed in the next chapter.

docs/en/reference/document-repositories.rst
```diff
@@ -0,0 +1,203 @@
...
+
+It is also possible to change ODM's default ``DocumentRepository`` to your own
+implementation for all documents (unless overridden by the mapping):
+
+.. code-block:: php
+
+    $documentManager->getConfiguration()
+        ->setDefaultRepositoryClassName(MyDefaultRepository::class);
+
+Repositories with additional dependencies
```
jmikola: Title capitalization: "Additional Dependencies"

docs/en/reference/document-repositories.rst
```diff
@@ -0,0 +1,203 @@
...
+-----------------------------------------
+
+.. note::
+    Option to implement own RepositoryFactory is available since version 1.0 however
+    ``AbstractRepositoryFactory`` used in the example is available only since 1.2.
```
jmikola: > Implementing your own RepositoryFactory is possible since version 1.0, but the `AbstractRepositoryFactory` class used in this example is only available since 1.2.

docs/en/reference/document-repositories.rst
```diff
@@ -0,0 +1,203 @@
...
+    {
+        public function __construct(DocumentManager $dm, UnitOfWork $uow, ClassMetadata $classMetadata)
+        {
+            /* constructor is inherited from DocumentRepository */
+            /* ... */
+        }
+    }
+
+To change the way Doctrine instantiates repositories you need to implement your own
+`RepositoryFactory <https://github.com/doctrine/mongodb-odm/blob/master/lib/Doctrine/ODM/MongoDB/Repository/RepositoryFactory.php>`_
```
jmikola: > In order to change the way Doctrine instantiates repositories, you will need to...

docs/en/reference/document-repositories.rst
```diff
@@ -0,0 +1,203 @@
...
+
+.. code-block:: php
+
+    <?php
+
+    use Doctrine\ODM\MongoDB\Repository\AbstractRepositoryFactory;
+    use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+
+    final class YourRepositoryFactory extends AbstractRepositoryFactory
```
jmikola: Random question: for pre-1.2, do users need to specify a lot of extra code?

docs/en/reference/document-repositories.rst
alcaeus: I'd say quite the opposite - well-defined magic can save lots of boilerplate code. If we say "we do not endorse" in the documentation we might as well go ahead and deprecate them altogether. In my opinion, there's no difference between `find(['status' => 'active'])` and `findByStatus('active')`, but declaring a repository method for those simple calls is overkill if you ask me.

docs/en/reference/document-repositories.rst
alcaeus: Yes, there was no abstract factory, so all logic had to be implemented by the user.

docs/en/reference/document-repositories.rst
malarzm: Personally I'm all for declaring own repositories for even simpliest `findBy` calls so I can mock them later or for the sake of logic change in future or to not scatter same calls all around the code :) I think we'll be better removing the note

docs/en/reference/document-repositories.rst
malarzm: @jmikola basically copy-paste the default factory and add instantiation logic

docs/en/reference/document-repositories.rst
Ocramius: `findBy(['foo' => 'bar'])` is much simpler than `findByFoo('bar')` handled via `__call`, as it is reflected in the signature of the class/interface. Having `__call` may appear interesting at first, but it's just a redundant, not documented, not type-checked magic method.

Unless there is a proven technical advantage, it's just an additional weakness added here.

docs/en/reference/document-repositories.rst
```diff
@@ -0,0 +1,203 @@
...
+.. code-block:: php
+
+    <?php
+
+    /* @var $repository \Doctrine\ODM\MongoDB\DocumentRepository */
+    $repository = $documentManager->getRepository(User::class);
+    $disabledUsers = $repository->findBy(['disabled' => true, 'activated' => true]);
+
+Array passed to ``findBy`` specifies criteria documents are matched against. ODM
+will help you with type conversions as much as it can:
```
malarzm: Not sure if we have that documented aywhere, will use your proposition

docs/en/reference/document-repositories.rst
```diff
@@ -0,0 +1,203 @@
...
+
+.. code-block:: php
+
+    <?php
+    $group = $documentManager->find(Group::class, 123);
+    /* @var $repository \Doctrine\ODM\MongoDB\DocumentRepository */
+    $repository = $documentManager->getRepository(User::class);
+    $usersInGroup = $repository->findBy(['group' => $group]);
```
malarzm: Yes

docs/en/reference/document-repositories.rst
```diff
@@ -0,0 +1,203 @@
...
+
+Following methods are available for default repository:
+
+- ``find()`` - finds one document by its identifier. This may skip database query
+if document is already known to ODM.
+- ``findAll()`` - gets all documents.
+- ``findBy()`` - gets all documents matching given criteria. Additionally documents
+can be sorted and a pagination applied.
+- ``findOneBy()`` - finds a document matching all given criteria.
```
malarzm: `findOneBy` accepts `$criteria` only so I'm omitting the second sentence here

docs/en/reference/document-repositories.rst
malarzm: `doctrine/collections` doesn't have its own documentation and in ORM I've found only http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/working-with-associations.html#filtering-collections which is a bit unfortunate as we don't have `matching` method implemented for collections

docs/en/reference/document-repositories.rst
malarzm: I'm removing this note altogether, will add a deprecation note later

docs/en/reference/document-repositories.rst
```diff
@@ -0,0 +1,203 @@
...
+    $group = $documentManager->find(Group::class, 123);
+    /* @var $repository \Doctrine\ODM\MongoDB\DocumentRepository */
+    $repository = $documentManager->getRepository(User::class);
+    $usersInGroup = $repository->findBy(['group' => $group]);
+
+Following methods are available for default repository:
+
+- ``find()`` - finds one document by its identifier. This may skip database query
+if document is already known to ODM.
+- ``findAll()`` - gets all documents.
```
malarzm: Added a note below

docs/en/reference/document-repositories.rst
```diff
@@ -0,0 +1,205 @@
...
+    <?php
+    $group = $documentManager->find(Group::class, 123);
+    /* @var $repository \Doctrine\ODM\MongoDB\DocumentRepository */
+    $repository = $documentManager->getRepository(User::class);
+    $usersInGroup = $repository->findBy(['group' => $group]);
+
+The default repository implementation provides the following methods:
+
+- ``find()`` - finds one document by its identifier. This may skip database query
+if document is already known to ODM.
```
malarzm: Previous comment so the change won't fell through the cracks:

> "This may skip database query if document is already known to ODM."
>
> This goes hand in hand with ODM trying to ensure only one document objects exists at a time for a given ID. In this case, it can avoid the query altogether. For the later methods, ODM might execute the query but still return existing documents if the _id fields in the results correspond to a managed document. Would it make sense to remind users of this behavior in a single note after this list? Bonus points if we can link to existing material on document management.

docs/en/reference/document-repositories.rst
```diff
@@ -0,0 +1,205 @@
...
+=====================
+
+.. note::
+
+    A repository mediates between the domain and data mapping layers using a
+    collection-like interface for accessing domain objects.
+
+In Doctrine, a repository is a class that concentrates code responsible for
+querying and filtering your documents. ODM provides you with a default
+``DocumentRepository`` for all your documents:
```
jmikola: "all of your"

docs/en/reference/document-repositories.rst
jmikola: Grammar: "This may skip a database query if the document is already known to ODM"

Also, would it make more sense to say "already managed by ODM" instead of "already known"?

docs/en/reference/document-repositories.rst
```diff
@@ -0,0 +1,205 @@
...
+
+        Documents\User:
+            repositoryClass: Repositories\\UserRepository
+            collection: user
+            # ...
+
+The next step is implementing your repository class. In most cases, ODM's default
+``DocumentRepository`` class may be extended with additional methods that you need.
+More complex cases that require passing additional dependencies to a custom repository
+class will be discussed in the next chapter.
```
jmikola: "next section". I think chapter would imply a different page entirely.

docs/en/reference/document-repositories.rst
malarzm: The more I look for related part of documentation on managed documents the more I want to burn it entirely and write from ground up... So I'll indeed change the wording to "already managed by ODM" and call it a day

docs/en/reference/document-repositories.rst
malarzm: Good point

PR #1604, opened 2017-06-13
Adjust "Working with Objects" chapter

malarzm:


docs/en/reference/working-with-objects.rst
```diff
@@ -105,10 +105,9 @@ Example:
 
 .. caution::
 
-    The \_id of a document is guaranteed to be available
-    after the next successful flush operation that involves the
-    document in question. You can not rely on a generated identifier to
-    be available directly after invoking ``persist``.
+    If not provided upfront the \_id of a document is generated
+    during ``persist``. You can not rely on a generated identifier to
+    be available during ``prePersist`` event.
```
jmikola: Suggested rephrasing:

> The document identifier is generated during `persist` if not previously specified. Users cannot rely on a document identifier being available during the `prePersist` event.

For the second sentence, I'm not sure if "generated identifier" is necessary. I think the takeaway is that users simply cannot on an identifier being available. Generated identifiers certainly won't be available, although user-created identifiers may be.

Also, I might suggest adding "(i.e. `_id` database field)" after "The document identifier" in the first sentence. This section is a high-level discussion. IIRC, a `prePersist` event would access properties on the document object (e.g. `$id` field) instead of raw database fields -- do correct me if I'm mistaken, though.

malarzm:
Oh my, I totally forgot about this PR :)

PR #1605, opened 2017-06-13
Deprecate Query Builder's group

malarzm:
The `group` command [was deprecated in MongoDB 3.4](https://docs.mongodb.com/v3.4/reference/method/db.collection.group/)

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Query/Builder.php
+++ b/lib/Doctrine/ODM/MongoDB/Query/Builder.php
@@ -344,6 +344,19 @@ class Builder extends \Doctrine\MongoDB\Query\Builder
     }
 
     /**
+     * @inheritdoc
+     * @deprecated Deprecated in version 1.2 - use Aggregation Builder's group stage instead.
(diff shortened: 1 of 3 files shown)


lib/Doctrine/ODM/MongoDB/Query/Builder.php
```diff
@@ -344,6 +344,15 @@ public function includesReferenceTo($document)
     }
 
     /**
+     * @inheritdoc
+     * @deprecated Deprecated in version 1.2 - use Aggregation Builder's group stage instead.
+     */
+    public function group($keys, array $initial, $reduce = null, array $options = [])
+    {
+        return parent::group($keys, $initial, $reduce, $options);
```
alcaeus: You could add a `@trigger_error(...)` call to help people spot usages of `group` in their code.

malarzm: Done, I'll also add triggers to existing deprecations later on

PR #1606, opened 2017-06-14
Trigger E_USER_DEPRECATED for deprecated features

malarzm:
I think I've covered all places that make sense. Not sure about all types annotations - should each deprecated one trigger error in its constructor?

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/EagerCursor.php
+++ b/lib/Doctrine/ODM/MongoDB/EagerCursor.php
@@ -19,6 +19,9 @@
 
 namespace Doctrine\ODM\MongoDB;
 
+use Doctrine\MongoDB\CursorInterface;
+use Doctrine\ODM\MongoDB\Mapping\ClassMetadata;
+
 /**
  * EagerCursor wraps a Cursor instance and fetches all of its results upon
  * initialization.
@@ -28,4 +31,12 @@ namespace Doctrine\ODM\MongoDB;
  */
 class EagerCursor extends Cursor
 {
+    public function __construct(CursorInterface $baseCursor, UnitOfWork $unitOfWork, ClassMetadata $class)
+    {
+        @trigger_error(
+            sprintf('%s is deprecated - use %s instead.', __CLASS__, Cursor::class),
+            E_USER_DEPRECATED
+        );
+        parent::__construct($baseCursor, $unitOfWork, $class);
+    }
 }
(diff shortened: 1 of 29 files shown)


lib/Doctrine/ODM/MongoDB/EagerCursor.php
```diff
@@ -18,6 +18,8 @@
  */
 
 namespace Doctrine\ODM\MongoDB;
+use Doctrine\MongoDB\CursorInterface;
```
alcaeus: I believe we've previously used a blank line before `use`.

malarzm: Yes we did, must have overlooked Storm adding `use` in the first place :+1:

malarzm:
> should each deprecated one trigger error in its constructor?

Unfortunately it's not possible as [`Annotation::__construct` is final`](https://github.com/doctrine/annotations/blob/7c830a576f9471f71096d200d633b6400513ccfb/lib/Doctrine/Common/Annotations/Annotation.php#L44) :( Next idea is hooking into `AnnotationDriver` and detect when there's deprecated annotation used, thoughts @alcaeus?

alcaeus:
Can we do away with [DoctrineAnnotations.php](https://github.com/doctrine/mongodb-odm/blob/master/lib/Doctrine/ODM/MongoDB/Mapping/Annotations/DoctrineAnnotations.php) and have the annotations autoloaded? In that case you could put the deprecation notice in the file itself. If not, we'll have to handle it in the driver. In that case, we could add a deprecation message to the annotation class itself by using a `DEPRECATED` constant (along with the usual `@deprecated` doc) and check against that in the driver. Hacky, but it avoids keeping a map of deprecated annotations and their messages in the driver.

malarzm:
Not really as that would be a BC break, there's [AnnotationDriver::registerAnnotationClasses](https://github.com/doctrine/mongodb-odm/blob/80e8df78856372136734ec6920582d855703796e/lib/Doctrine/ODM/MongoDB/Mapping/Driver/AnnotationDriver.php#L52) which was recommended to use. I guess driver it is

alcaeus:
> I guess driver it is

I guess driver it is then.

malarzm:
@alcaeus annotations done, also added 2 more notices I forgot about before.

malarzm:
> Note: apparently the email address associated with your first commit isn't associated with your GitHub account, you might want to take a look at your configs ;)

Woah nice catch! Changed that one to be public me and claimed that email address, must have came from my workstation :)

alcaeus:
> Woah nice catch!

Saw your picture twice in the commit list. Once is plenty enough 😜

PR #1608, opened 2017-06-16
Deprecate AnnotationDriver::registerAnnotationClasses

malarzm:
I wonder if I should add trigger in `DoctrineAnnotations.php` in case somebody required that file on one's own...


malarzm:
@alcaeus added

alcaeus:
Thanks!

PR #1609, opened 2017-06-16
[RFC] Deprecate repositories' magic findBy and findOneBy

malarzm:
See https://github.com/doctrine/mongodb-odm/pull/1601#discussion_r119143856 for the original discussion.

Please vote with 👍 and 👎


malarzm:
@alcaeus it's 12:1 after a week, I'm afraid your side is losing :)

alcaeus:
I'll survive ;)

PR #1610, opened 2017-06-16
Deprecate slaveOkay

malarzm:
ATM there is no replacement for `slaveOkay` specified at the document level - do you guys think it's worth adding?


lib/Doctrine/ODM/MongoDB/Cursor.php
```diff
@@ -549,9 +549,15 @@ public function skip($num)
      * @see http://php.net/manual/en/mongocursor.slaveokay.php
      * @param boolean $ok
      * @return $this
+     *
+     * @deprecated in version 1.2 - use setReadPreference instead.
```
alcaeus: The new cursor classes don't have an equivalent to `slaveOkay` since they don't share the pre/post execution state of the legacy driver cursors. We might want to specify: "use setReadPreference on the query instead"

malarzm:
`writeConcern` is already supported :)

lib/Doctrine/ODM/MongoDB/Cursor.php
malarzm: :+1:

jmikola:
> Regarding your question, we might consider adding readPreference, writeConcern and readConcern to the document level, but I haven't invested too much thought into it so far.

Encouraging read preference per document class may be problematic if documents use inconsistent read preferences. If used without a read concern, it creates a situation where some documents could be fetched from a primary and others from a secondary, where they may have stale data or not exist at all.

I think users would be better served specifying read preferences in custom repository methods if they actually need to query secondaries for some reason.

alcaeus:
> it creates a situation where some documents could be fetched from a primary and others from a secondary, where they may have stale data or not exist at all.

That is true regardless of where it's defined. Even when explicitly fetching a document from a secondary using queries, the references contained therein would be fetched from the master, unless they were also loaded with a readPreference that allows reading from a secondary.

> I think users would be better served specifying read preferences in custom repository methods if they actually need to query secondaries for some reason.

That would create an overhead for references, where even simple references (`mappedBy="user"`) would have to be rewritten to use a repository method just to specify a read preference or read concern.

IMO, we should allow specifying both values at the document level as well as for references. At the same time, we should properly document the meaning of both and the problems that can arise from using the wrong configuration.

PR #1612, opened 2017-06-23
Add read only documents

### LINKED ISSUE #1602: Add read-only option for document mappings
Doctrine ORM allows to map a class as read-only. This ODM implementation does not support something like this.
It would be a great feature to mark document classes read-only to make sure their data is immutable.

malarzm:
Closes #1602 (cc @tux-rampage)

Allowing document to be upserted opens a way to overwriting its value but otherwise it would be impossible to have read only document with user-generated identifier. We may want to revisit this once we have official `upsert` API :)

### DIFF
--- a/docs/en/reference/annotations-reference.rst
+++ b/docs/en/reference/annotations-reference.rst
@@ -396,6 +396,9 @@ Optional attributes:
 -
    indexes - Specifies an array of indexes for this document.
 -
+   readOnly - Prevents document from being updated: it can only be inserted,
+   upserted or removed.
+-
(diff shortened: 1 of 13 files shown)


tux-rampage:
:+1: nice

malarzm:
Just rebased atop current `master` branch, will merge once tests are green

alcaeus:
@malarzm could you give this another rebase? Tests should be stable now that build failures due to version issues are over.

malarzm:
Here it goes, let's see what happens :)

PR #1613, opened 2017-06-23
[2.0][RFC] Add get to default repository

malarzm:
Today I have realized that most of my custom repositories are beginning with `get` method which looks just like the one I have here. So how about adding it to core? If we say "yes" then here's still few things to do like:

- [ ] change `ObjectRepository` occurrences to new `Repository`
- [ ] especially after rebasing `odm-ng` atop current `master`
- [ ] move PHPDocs to interface

Also since it's for 2.0 we may consider moving Repository-related classes around as we got ourselves "Repository" namespace just before 1.0, but that will mean more things to fix for users when upgrading.


alcaeus:
As mentioned in IRC, I see a few issues with the method:
* We have `find`, `findBy`, `findOneBy`. Especially with `find`, it's not easy to understand the difference between `find` and `get` since they are not self-explanatory.
* One could argue that the current API covers the use cases. For those cases where the code can't continue if no object is returned, the code may as well throw the exception themselves and throw a specific exception instead of a generic `DocumentNotFoundException`.
* Last but not least, there's no comparable method in ORM and I don't know of any plans to add one. I'd always try to keep the APIs as comparable as possible, especially in those classes that don't deal with low-level logic that might depend on the database system in use.

Considering this, I'm 👎  on adding this.

PR #1620, opened 2017-07-18
Add ReadPreference on a document level

### LINKED ISSUE #1611: Allow readPreference and readConcern on document-level
This is a follow-up to #1610, as per @alcaeus' suggestion it might be feasible to add it to references as well.

malarzm:
Closes #1611

- [x] Allow specifying read preference for documents (mind inheritance)
- [x] ~Allow specifying read preference for references~ - will be done separately
- [x] Documentation


lib/Doctrine/ODM/MongoDB/DocumentManager.php
```diff
@@ -352,6 +352,10 @@ public function getDocumentCollection($className)
             $collection->setSlaveOkay($metadata->slaveOkay);
         }
 
+        if ($metadata->readPreference !== null) {
```
alcaeus: We might give `readPreference` precedence over slaveOkay here - that would also solve conflicts where a readPreference of `primary` is given but `slaveOkay` is also true.

lib/Doctrine/ODM/MongoDB/Query/Builder.php
```diff
@@ -429,6 +429,11 @@ public function getQuery(array $options = array())
             $query['slaveOkay'] = $this->class->slaveOkay;
         }
 
+        if ($this->class->readPreference && ! array_key_exists('readPreference', $query)) {
```
alcaeus: Same here - `readPreference` should have precedence over `slaveOkay`

lib/Doctrine/ODM/MongoDB/DocumentManager.php
malarzm: By precedence you mean putting `slaveOkay` in `else` block if `readPreference` is not set?

malarzm: Also with exception in place there'd be no need for that :)

alcaeus: Correct, `slaveOkay` is only used if `readPreference` is not set. Also, yes, not necessary if we throw a mapping exception on conflict. Might be the better solution anyways because it exposes conflicts instead of silently solving them somehow.

malarzm: Exception it is

lib/Doctrine/ODM/MongoDB/Query/Builder.php
malarzm: As this reads from class metadata the exception thrown there should be sufficient

malarzm:
> One thing that just came to me, do we want to create constants for read preference strings to avoid spreading magic strings through people's code, or even worse: get them to use the \MongoDB\Driver\ReadPreference::RP_* constants that have an entirely different meaning.

Reading [this part of ctor](https://github.com/mongodb/mongo-php-driver/blob/8742e1bc6fd5261049b8e0bf4e2a3bfc370ab484/src/MongoDB/ReadPreference.c#L54-L66) I'd assume that people using `\MongoDB\Driver\ReadPreference::RP_*` will be OK, no?

alcaeus:
Yes, that's correct. I thought I wouldn't have to read driver code now that I'm done with ext-mongo, turns out I was mistaken :)

docs/en/reference/annotations-reference.rst
```diff
@@ -1207,6 +1207,63 @@ the method to be registered.
 
 See :ref:`lifecycle_events` for more information.
 
+@ReadPreference
+---------------
+
+Specifies `Read Preference <https://docs.mongodb.com/manual/core/read-preference/>_`
```
alcaeus: We could document the conflict (and resulting exception) here and in the corresponding documentation for `slaveOkay`.

docs/en/reference/annotations-reference.rst
```diff
@@ -1207,6 +1207,63 @@ the method to be registered.
...
+         *   { "dc"="east" },
+         *   { "dc"="west" },
+         *   {  }
+         * })
+         */
+        class User
+        {
+        }
+
+    .. code-block:: xml
```
alcaeus: Since this is the annotations reference, we don't normally document any other mappings than annotations here. Given the fact that we're kicking yaml and XML is pretty self-explanatory thanks to the schema, I'd vote for omitting these here. We can always add an XML mapping reference if we want or create a more comprehensive mapping reference than we have now. Just documenting XML for this one annotation doesn't make a lot of sense here IMO.

lib/Doctrine/ODM/MongoDB/Mapping/MappingException.php
```diff
@@ -396,4 +396,9 @@ public static function referencePrimersOnlySupportedForInverseReferenceMany($cla
     {
         return new self("Cannot use reference priming on '$fieldName' in class '$className'. Reference priming is only supported for inverse references");
     }
+
+    public static function canNotCombineReadPreferenceAndSlaveOkay($className)
+    {
+        return new self("Cannot use ReadPreference and SlaveOkay at a same time in class '$className'.");
```
alcaeus: > at the same time

malarzm:
> Yes, that's correct. I thought I wouldn't have to read driver code now that I'm done with ext-mongo, turns out I was mistaken :)



cc @jmikola ;)

docs/en/reference/annotations-reference.rst
malarzm: Documented only here as apparently `slaveOkay` wasn't even documented in the reference :baby:

alcaeus:
Note: might want to look at `Doctrine\ODM\MongoDB\Tests\Functional\ReadPreferenceTest::testDocumentLevelReadPreferenceCanBeOverriddenInQueryBuilder` ;)

malarzm:
@alcaeus do you have any idea what's wrong with https://travis-ci.org/doctrine/mongodb-odm/jobs/290870337#L649 ? I tried bumping `doctrine/annotations` to no avail :/

alcaeus:
@malarzm you need to improve your dependency-bumping skills 😛  #1654 bumps the dependency to `~1.2` as I want to be able to use `::class` in tests. That incidentally also fixes your problem.

I've taken the liberty to amend your commit and have also moved a class from the Yaml mapping test to the abstract mapping test as it failed when running only the annotation driver test.

malarzm:
Oh my, for some reasons I was sure it was 1.1 that introduced support for `::class` and is latest one... Anyways rerunning test that failed :)

lib/Doctrine/ODM/MongoDB/Mapping/MappingException.php
```diff
@@ -396,4 +396,9 @@ public static function referencePrimersOnlySupportedForInverseReferenceMany($cla
     {
         return new self("Cannot use reference priming on '$fieldName' in class '$className'. Reference priming is only supported for inverse references");
     }
+
+    public static function canNotCombineReadPreferenceAndSlaveOkay($className)
```
alcaeus: We could deprecate this method as we’re removing it in 2.0.

alcaeus:
Thanks @malarzm!

PR #1623, opened 2017-07-26
Add generic reference as replacement for DBRef

alcaeus:
Supersedes #1566.

As discussed in our last hangout, this is what remains of the DBRef replacement. DBRef is not deprecated (yet), athough we might decide to deprecate it at this time. At this time the PR is feature complete, I'd like to write a whole bunch of tests to ensure the whole discriminator handling/omitting target collection name works as I'd expect it to.

The new reference object only stores an ID and optionally a discriminator field. The main difference to DBRef is that it can be used in aggregation pipeline stages that don't handle DBRef objects because of the `$` in field names (e.g. `$lookup` and `$graphLookup`).

I'll also revisit #1527 to see if we can store additional fields in the reference (which would also apply to DBRef objects).

### DIFF
--- a/docs/en/reference/annotations-reference.rst
+++ b/docs/en/reference/annotations-reference.rst
@@ -1222,10 +1222,11 @@ Optional attributes:
 -
     simple - deprecated (use ``storeAs: id``)
 -
-    storeAs - Indicates how to store the reference. ``id`` uses ``MongoId``,
-    ``dbRef`` uses a `DBRef`_ without ``$db`` value and ``dbRefWithDb`` stores
-    a full `DBRef`_ (``$ref``, ``$id``, and ``$db``). Note that ``id``
-    references are not compatible with the discriminators.
+    storeAs - Indicates how to store the reference. ``id`` stores the identifier,
+    ``ref`` an embedded object containing the ``id`` field and (optionally) a
+    discriminator. ``dbRef`` and ``dbRefWithDb`` store a ``DBRef`` object. They
+    are deprecated in favor of ``ref``. Note that ``id`` references are not
+    compatible with the discriminators.
 -
     cascade - Cascade Option
 -
@@ -1297,8 +1298,8 @@ Optional attributes:
     simple - deprecated (use ``storeAs: id``)
 -
     storeAs - Indicates how to store the reference. ``id`` uses ``MongoId``,
-    ``dbRef`` uses a `DBRef`_ without ``$db`` value and ``dbRefWithDb`` stores
-    a full `DBRef`_ (``$ref``, ``$id``, and ``$db``). Note that ``id``
+    ``ref`` stores fields ``id`` and ``ref`` (similar to DBRef without $ prefix) value and ``refWithDb`` stores
+    a additionally db as parameter. ``dbRef`` and ``dbRefWithDb`` use ``DBRef``, they are deprecated in favor of ref and refWithDb. Note that ``id``
     references are not compatible with the discriminators.
 -
     cascade - Cascade Option
--- a/docs/en/reference/annotations-reference.rst
+++ b/docs/en/reference/annotations-reference.rst
@@ -1218,25 +1218,26 @@ documents.
 Optional attributes:
 
 -
-    targetDocument - A |FQCN| of the target document.
+    targetDocument - A |FQCN| of the target document. A ``targetDocument`` is
+    required when using ``storeAs: id``.
 -
     simple - deprecated (use ``storeAs: id``)
 -
     storeAs - Indicates how to store the reference. ``id`` stores the identifier,
     ``ref`` an embedded object containing the ``id`` field and (optionally) a
-    discriminator. ``dbRef`` and ``dbRefWithDb`` store a ``DBRef`` object. They
+    discriminator. ``dbRef`` and ``dbRefWithDb`` store a `DBRef`_ object. They
     are deprecated in favor of ``ref``. Note that ``id`` references are not
     compatible with the discriminators.
 -
     cascade - Cascade Option
 -
     discriminatorField - The field name to store the discriminator value within
-    the `DBRef`_ object.
+    the reference object.
 -
     discriminatorMap - Map of discriminator values to class names.
 -
     defaultDiscriminatorValue - A default value for discriminatorField if no value
-    has been set in the embedded document.
+    has been set in the referenced document.
 -
     inversedBy - The field name of the inverse side. Only allowed on owning side.
 -
@@ -1293,24 +1294,26 @@ Defines an instance variable holds a related document instance.
 Optional attributes:
 
 -
-    targetDocument - A |FQCN| of the target document.
+    targetDocument - A |FQCN| of the target document. A ``targetDocument`` is
+    required when using ``storeAs: id``.
 -
     simple - deprecated (use ``storeAs: id``)
 -
-    storeAs - Indicates how to store the reference. ``id`` uses ``MongoId``,
-    ``ref`` stores fields ``id`` and ``ref`` (similar to DBRef without $ prefix) value and ``refWithDb`` stores
-    a additionally db as parameter. ``dbRef`` and ``dbRefWithDb`` use ``DBRef``, they are deprecated in favor of ref and refWithDb. Note that ``id``
-    references are not compatible with the discriminators.
+    storeAs - Indicates how to store the reference. ``id`` stores the identifier,
+    ``ref`` an embedded object containing the ``id`` field and (optionally) a
+    discriminator. ``dbRef`` and ``dbRefWithDb`` store a `DBRef`_ object. They
+    are deprecated in favor of ``ref``. Note that ``id`` references are not
+    compatible with the discriminators.
 -
     cascade - Cascade Option
 -
     discriminatorField - The field name to store the discriminator value within
-    the `DBRef`_ object.
+    the reference object.
 -
     discriminatorMap - Map of discriminator values to class names.
 -
     defaultDiscriminatorValue - A default value for discriminatorField if no value
-    has been set in the embedded document.
+    has been set in the referenced document.
 -
     inversedBy - The field name of the inverse side. Only allowed on owning side.
 -
--- a/docs/en/reference/annotations-reference.rst
+++ b/docs/en/reference/annotations-reference.rst
@@ -1225,7 +1225,7 @@ Optional attributes:
 -
     storeAs - Indicates how to store the reference. ``id`` stores the identifier,
     ``ref`` an embedded object containing the ``id`` field and (optionally) a
-    discriminator. ``dbRef`` and ``dbRefWithDb`` store a `DBRef`_ object. They
+    discriminator. ``dbRef`` and ``dbRefWithDb`` store a `DBRef`_ object and
     are deprecated in favor of ``ref``. Note that ``id`` references are not
     compatible with the discriminators.
 -
@@ -1301,7 +1301,7 @@ Optional attributes:
 -
     storeAs - Indicates how to store the reference. ``id`` stores the identifier,
     ``ref`` an embedded object containing the ``id`` field and (optionally) a
-    discriminator. ``dbRef`` and ``dbRefWithDb`` store a `DBRef`_ object. They
+    discriminator. ``dbRef`` and ``dbRefWithDb`` store a `DBRef`_ object and
     are deprecated in favor of ``ref``. Note that ``id`` references are not
     compatible with the discriminators.
 -
--- a/lib/Doctrine/ODM/MongoDB/DocumentManager.php
+++ b/lib/Doctrine/ODM/MongoDB/DocumentManager.php
@@ -668,15 +668,15 @@ class DocumentManager implements ObjectManager
     }
 
     /**
-     * Returns a DBRef array for the supplied document.
+     * Returns a reference to the supplied document.
      *
-     * @param mixed $document A document object
+     * @param object $document A document object
      * @param array $referenceMapping Mapping for the field that references the document
      *
      * @throws \InvalidArgumentException
-     * @return array A DBRef array
+     * @return mixed The reference for the document in question, according to the desired mapping
      */
-    public function createDBRef($document, array $referenceMapping = null)
+    public function createReference($document, array $referenceMapping)
     {
         if ( ! is_object($document)) {
             throw new \InvalidArgumentException('Cannot create a DBRef, the document is not an object');
@@ -691,7 +691,8 @@ class DocumentManager implements ObjectManager
             );
         }
 
-        switch ($referenceMapping['storeAs']) {
+        $storeAs = isset($referenceMapping['storeAs']) ? $referenceMapping['storeAs'] : null;
+        switch ($storeAs) {
             case ClassMetadataInfo::REFERENCE_STORE_AS_ID:
                 if ($class->inheritanceType === ClassMetadataInfo::INHERITANCE_TYPE_SINGLE_COLLECTION) {
                     throw MappingException::simpleReferenceMustNotTargetDiscriminatedDocument($referenceMapping['targetDocument']);
@@ -710,7 +711,7 @@ class DocumentManager implements ObjectManager
                     '$ref' => $class->getCollection(),
                     '$id'  => $class->getDatabaseIdentifierValue($id),
                 ];
-                if ($referenceMapping['storeAs'] === ClassMetadataInfo::REFERENCE_STORE_AS_DB_REF_WITH_DB) {
+                if ($storeAs === ClassMetadataInfo::REFERENCE_STORE_AS_DB_REF_WITH_DB) {
                     $dbRef['$db'] = $this->getDocumentDatabase($class->name)->getName();
                 }
         }
@@ -728,7 +729,7 @@ class DocumentManager implements ObjectManager
         /* Add a discriminator value if the referenced document is not mapped
          * explicitly to a targetDocument class.
          */
-        if ($referenceMapping !== null && ! isset($referenceMapping['targetDocument'])) {
+        if (! isset($referenceMapping['targetDocument'])) {
             $discriminatorField = $referenceMapping['discriminatorField'];
             $discriminatorValue = isset($referenceMapping['discriminatorMap'])
                 ? array_search($class->name, $referenceMapping['discriminatorMap'])
@@ -748,6 +749,27 @@ class DocumentManager implements ObjectManager
         }
 
         return $dbRef;
+    }
+
+    /**
+     * Returns a DBRef array for the supplied document.
+     *
+     * @param mixed $document A document object
+     * @param array $referenceMapping Mapping for the field that references the document
+     *
+     * @throws \InvalidArgumentException
+     * @return array A DBRef array
+     * @deprecated Deprecated in favor of createReference; will be removed in 2.0
+     */
+    public function createDBRef($document, array $referenceMapping = null)
+    {
+        @trigger_error('The ' . __METHOD__ . ' method has been deprecated and will be removed in ODM 2.0. Use createReference() instead.', E_USER_DEPRECATED);
+
+        if (!isset($referenceMapping['storeAs'])) {
+            $referenceMapping['storeAs'] = ClassMetadataInfo::REFERENCE_STORE_AS_DB_REF;
+        }
+
+        return $this->createReference($document, $referenceMapping);
     }
 
     /**
--- a/lib/Doctrine/ODM/MongoDB/DocumentManager.php
+++ b/lib/Doctrine/ODM/MongoDB/DocumentManager.php
@@ -703,17 +703,26 @@ class DocumentManager implements ObjectManager
 
 
             case ClassMetadataInfo::REFERENCE_STORE_AS_REF:
-                $dbRef = ['id' => $class->getDatabaseIdentifierValue($id)];
+                $reference = ['id' => $class->getDatabaseIdentifierValue($id)];
                 break;
 
-            default:
-                $dbRef = [
+            case ClassMetadataInfo::REFERENCE_STORE_AS_DB_REF:
+                $reference = [
                     '$ref' => $class->getCollection(),
                     '$id'  => $class->getDatabaseIdentifierValue($id),
                 ];
-                if ($storeAs === ClassMetadataInfo::REFERENCE_STORE_AS_DB_REF_WITH_DB) {
-                    $dbRef['$db'] = $this->getDocumentDatabase($class->name)->getName();
-                }
+                break;
+
+            case ClassMetadataInfo::REFERENCE_STORE_AS_DB_REF_WITH_DB:
+                $reference = [
+                    '$ref' => $class->getCollection(),
+                    '$id'  => $class->getDatabaseIdentifierValue($id),
+                    '$db'  => $this->getDocumentDatabase($class->name)->getName(),
+                ];
+                break;
+
+            default:
+                throw new \InvalidArgumentException("Reference type {$storeAs} is invalid.");
         }
 
         /* If the class has a discriminator (field and value), use it. A child
@@ -721,7 +730,7 @@ class DocumentManager implements ObjectManager
          * discriminator field and no value, so default to the full class name.
          */
         if (isset($class->discriminatorField)) {
-            $dbRef[$class->discriminatorField] = isset($class->discriminatorValue)
+            $reference[$class->discriminatorField] = isset($class->discriminatorValue)
                 ? $class->discriminatorValue
                 : $class->name;
         }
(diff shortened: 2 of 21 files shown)


lib/Doctrine/ODM/MongoDB/DocumentManager.php
```diff
@@ -691,20 +691,28 @@ public function createDBRef($document, array $referenceMapping = null)
...
-            '$ref' => $class->getCollection(),
-            '$id'  => $class->getDatabaseIdentifierValue($id),
-        );
-
-        if ($referenceMapping['storeAs'] === ClassMetadataInfo::REFERENCE_STORE_AS_DB_REF_WITH_DB) {
-            $dbRef['$db'] = $this->getDocumentDatabase($class->name)->getName();
+        switch ($referenceMapping['storeAs']) {
+            case ClassMetadataInfo::REFERENCE_STORE_AS_ID:
+                if ($class->inheritanceType === ClassMetadataInfo::INHERITANCE_TYPE_SINGLE_COLLECTION) {
+                    throw MappingException::simpleReferenceMustNotTargetDiscriminatedDocument($referenceMapping['targetDocument']);
```
jmikola: In this case, shouldn't we throw if `$class->inheritanceType` is not `ClassMetadataInfo::INHERITANCE_TYPE_NONE`?

Wouldn't `simpleReferenceMustNotTargetDiscriminatedDocument` apply to both single- and multi-collection inheritance, since the destination collection is relevant in both cases? A discriminator is only necessary for single-collection inheritance, since the collection name fulfills that role for multi-collection inheritance.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataInfo.php
```diff
@@ -468,6 +469,29 @@ public function __construct($documentName)
     }
 
     /**
+     * Helper method to get reference id of ref* type references
+     * @param array  $reference
+     * @param string $storeAs
+     * @return mixed
+     * @internal
+     */
+    public static function getReferenceId($reference, $storeAs)
```
jmikola: Type hint `$reference` with `array`.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataInfo.php
```diff
@@ -468,6 +469,29 @@ public function __construct($documentName)
...
+    {
+        return $reference[ClassMetadataInfo::getReferencePrefix($storeAs) . 'id'];
+    }
+
+    /**
+     * Returns the reference prefix used for a referene
```
jmikola: Typo: "for a reference"

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataInfo.php
```diff
@@ -468,6 +469,29 @@ public function __construct($documentName)
...
+     * @param string $storeAs
+     * @return string
+     * @internal
+     */
+    public static function getReferencePrefix($storeAs)
+    {
+        return $storeAs === ClassMetadataInfo::REFERENCE_STORE_AS_REF ? '' : '$';
```
jmikola: I assume "$" only makes sense for one of the two DBRef reference types. Would you want that returned if this is somehow called for `REFERENCE_STORE_AS_ID`? Alternatively, if this method should never be called in that case perhaps we should throw a logic exception.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -722,7 +722,7 @@ private function loadReferenceManyCollectionOwningSide(PersistentCollectionInter
                 $mongoId = $reference;
             } else {
                 $className = $this->uow->getClassNameForAssociation($mapping, $reference);
-                $mongoId = $reference['$id'];
+                $mongoId = ClassMetadataInfo::getReferenceId($reference, $mapping['storeAs']);
```
jmikola: Earlier, I noted that you may want to add the `array` type hint to `getReferenceId()`'s first parameter. Alternatively, it may be worth changing `getReferenceId()` and `UnitOfWork::getClassNameForAssociation()` to work on simple references, which would let you remove this if/else entirely and simply assign `$className` and `$mongoId` the same for all reference types.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -805,7 +805,12 @@ public function createReferenceManyInverseSideQuery(PersistentCollectionInterfac
         $ownerClass = $this->dm->getClassMetadata(get_class($owner));
         $targetClass = $this->dm->getClassMetadata($mapping['targetDocument']);
         $mappedByMapping = isset($targetClass->fieldMappings[$mapping['mappedBy']]) ? $targetClass->fieldMappings[$mapping['mappedBy']] : array();
-        $mappedByFieldName = isset($mappedByMapping['storeAs']) && $mappedByMapping['storeAs'] === ClassMetadataInfo::REFERENCE_STORE_AS_ID ? $mapping['mappedBy'] : $mapping['mappedBy'] . '.$id';
+        if (isset($mappedByMapping['storeAs']) && $mappedByMapping['storeAs'] === ClassMetadataInfo::REFERENCE_STORE_AS_ID) {
+            $mappedByFieldName = $mapping['mappedBy'];
+        } else {
+            $mappedByFieldName = $mapping['mappedBy'] . '.' . ClassMetadataInfo::getReferencePrefix(isset($mappedByMapping['storeAs']) ? $mappedByMapping['storeAs'] : null) . 'id';
```
jmikola: Similarly, perhaps we can have a single method here that returns the `mappedByFieldName` for any reference type. That may allow us to keep `getReferencePrefix()` private.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -1424,20 +1429,25 @@ private function getWriteOptions(array $options = array())
     private function prepareDbRefElement($fieldName, $value, array $mapping, $inNewObj)
     {
         $dbRef = $this->dm->createDBRef($value, $mapping);
-        if ($inNewObj) {
+        if ($inNewObj || $mapping['storeAs'] === ClassMetadataInfo::REFERENCE_STORE_AS_ID) {
```
jmikola: I don't quite recall what `prepareDbRefElement()` is used for, but these changes appear to be correct. I assume discriminator values (if applicable) are added later, since this method only deals with basic keys for database references?

lib/Doctrine/ODM/MongoDB/Query/ReferencePrimer.php
```diff
@@ -255,6 +255,7 @@ private function addManyReferences(PersistentCollectionInterface $persistentColl
     {
         $mapping = $persistentCollection->getMapping();
 
+
```
jmikola: Unnecessary line break?

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/LookupTest.php
```diff
@@ -133,6 +133,42 @@ public function testLookupStageReferenceMany()
         $this->assertSame('malarzm', $result[1]['users'][0]['username']);
     }
 
+    public function testLookupStageReferenceManyStoreAsRef()
+    {
+        $this->requireMongoDB32('$lookup tests require at least MongoDB 3.2.0');
+        $this->insertTestData();
```
jmikola: Would it make sense for this to go in the `setUp()` function for the test class, or is it not needed by each test case?

docs/en/reference/annotations-reference.rst
```diff
@@ -1218,10 +1218,11 @@ Optional attributes:
 -
     simple - deprecated (use ``storeAs: id``)
 -
-    storeAs - Indicates how to store the reference. ``id`` uses ``MongoId``,
-    ``dbRef`` uses a `DBRef`_ without ``$db`` value and ``dbRefWithDb`` stores
-    a full `DBRef`_ (``$ref``, ``$id``, and ``$db``). Note that ``id``
-    references are not compatible with the discriminators.
+    storeAs - Indicates how to store the reference. ``id`` stores the identifier,
+    ``ref`` an embedded object containing the ``id`` field and (optionally) a
+    discriminator. ``dbRef`` and ``dbRefWithDb`` store a ``DBRef`` object. They
```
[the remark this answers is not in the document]
alcaeus: Yes, we should. Will update that.

lib/Doctrine/ODM/MongoDB/DocumentManager.php
```diff
@@ -691,20 +691,28 @@ public function createDBRef($document, array $referenceMapping = null)
...
-            if ($class->inheritanceType === ClassMetadataInfo::INHERITANCE_TYPE_SINGLE_COLLECTION) {
-                throw MappingException::simpleReferenceMustNotTargetDiscriminatedDocument($referenceMapping['targetDocument']);
-            }
-            return $class->getDatabaseIdentifierValue($id);
-        }
-
-        $dbRef = array(
-            '$ref' => $class->getCollection(),
-            '$id'  => $class->getDatabaseIdentifierValue($id),
-        );
```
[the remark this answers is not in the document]
alcaeus: The name would suggest it creates a `DBRef` object, but it was used to create whatever reference was required according to the mapping. I'd suggest deprecating this and introducing a new `createReference` method to take its place.

lib/Doctrine/ODM/MongoDB/DocumentManager.php
alcaeus: Reading through the code again, I agree. Throwing on single collection inheritance doesn't make sense. I'll have to investigate this further since the code was already in place.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataInfo.php
alcaeus: Agree on the exception - neither prefix is valid for `REFERENCE_STORE_AS_ID`.

lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
```diff
@@ -260,7 +260,7 @@ private function generateHydratorClass(ClassMetadata $class, $hydratorClassName,
                 \$mongoId = \$reference;
             } else {
                 \$className = \$this->unitOfWork->getClassNameForAssociation(\$this->class->fieldMappings['%2\$s'], \$reference);
-                \$mongoId = \$reference['\$id'];
+                \$mongoId = \$reference[ClassMetadataInfo::getReferencePrefix(\$this->class->fieldMappings['%2\$s']['storeAs']) . 'id'];
```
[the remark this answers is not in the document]
alcaeus: Not sure what you mean. While `%2\$s` may look weird, it's necessary to escape the `$` in HEREDOC, so `sprintf` will only receive `%2$s`. That again is standard syntax to reference the second argument.

lib/Doctrine/ODM/MongoDB/DocumentManager.php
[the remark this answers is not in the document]
alcaeus: Introduced `createReference` which also makes the `$referenceMapping` argument mandatory. I've also replaced all occurences of `createDBRef` with `createReference`, hoping I didn't create a BC break with that because people expect `createDBRef` to be called.

alcaeus:
@jmikola took some time to work on this and applied most of your feedback. I still need to check for discriminator handling in `prepareDBRef` (which was renamed to `prepareReference` in DocumentPersister. I'll also have to revisit the exception mentioned in https://github.com/doctrine/mongodb-odm/pull/1623#discussion_r130397450 and add some tests. I'll hopefully get to that in a few days, but I believe the pull request is worth another look in the meantime.

alcaeus:
Note: build fails due to new driver version. I'll take a look at that in master branch.

lib/Doctrine/ODM/MongoDB/DocumentManager.php
```diff
@@ -691,20 +691,29 @@ public function createDBRef($document, array $referenceMapping = null)
...
-
-        if ($referenceMapping['storeAs'] === ClassMetadataInfo::REFERENCE_STORE_AS_DB_REF_WITH_DB) {
-            $dbRef['$db'] = $this->getDocumentDatabase($class->name)->getName();
+        $storeAs = isset($referenceMapping['storeAs']) ? $referenceMapping['storeAs'] : null;
+        switch ($storeAs) {
+            case ClassMetadataInfo::REFERENCE_STORE_AS_ID:
+                if ($class->inheritanceType === ClassMetadataInfo::INHERITANCE_TYPE_SINGLE_COLLECTION) {
+                    throw MappingException::simpleReferenceMustNotTargetDiscriminatedDocument($referenceMapping['targetDocument']);
```
jmikola: This appears to be the exception I was expecting for the `id` strategy and discriminated reference conflict.

lib/Doctrine/ODM/MongoDB/DocumentManager.php
```diff
@@ -691,20 +691,29 @@ public function createDBRef($document, array $referenceMapping = null)
...
+
+                return $class->getDatabaseIdentifierValue($id);
+                break;
+
+
+            case ClassMetadataInfo::REFERENCE_STORE_AS_REF:
+                $dbRef = ['id' => $class->getDatabaseIdentifierValue($id)];
+                break;
+
+            default:
```
jmikola: Might be safer to explicitly handle both cases here and leave `default:` in place to throw an exception for an unsupported strategy. That keeps the code defensive in case we ever add more strategies down the line.

lib/Doctrine/ODM/MongoDB/DocumentManager.php
```diff
@@ -691,20 +691,29 @@ public function createDBRef($document, array $referenceMapping = null)
...
+                if ($class->inheritanceType === ClassMetadataInfo::INHERITANCE_TYPE_SINGLE_COLLECTION) {
+                    throw MappingException::simpleReferenceMustNotTargetDiscriminatedDocument($referenceMapping['targetDocument']);
+                }
+
+                return $class->getDatabaseIdentifierValue($id);
+                break;
+
+
+            case ClassMetadataInfo::REFERENCE_STORE_AS_REF:
+                $dbRef = ['id' => $class->getDatabaseIdentifierValue($id)];
```
jmikola: Would it make sense to rename this local variable to `$ref`, since this was the original name used for two DBRef strategies handled by this function?

lib/Doctrine/ODM/MongoDB/DocumentManager.php
```diff
@@ -743,6 +752,27 @@ public function createDBRef($document, array $referenceMapping = null)
...
+     * @param mixed $document A document object
+     * @param array $referenceMapping Mapping for the field that references the document
+     *
+     * @throws \InvalidArgumentException
+     * @return array A DBRef array
+     * @deprecated Deprecated in favor of createReference; will be removed in 2.0
+     */
+    public function createDBRef($document, array $referenceMapping = null)
+    {
+        @trigger_error('The ' . __METHOD__ . ' method has been deprecated and will be removed in ODM 2.0. Use createReference() instead.', E_USER_DEPRECATED);
```
jmikola: What role does `@` server here? I'd think we'd want to avoid suppression so that a framework's error handler can step in and convert this to an exception if it likes.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataInfo.php
jmikola: Should we still type hint here? Alternatively, should the `@param` doc above be relaxed to `mixed`, since we may return `$reference` directly and we have `@return mixed`?

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -1422,23 +1418,28 @@ private function getWriteOptions(array $options = array())
      * @param bool $inNewObj
      * @return array
      */
-    private function prepareDbRefElement($fieldName, $value, array $mapping, $inNewObj)
+    private function prepareReference($fieldName, $value, array $mapping, $inNewObj)
     {
-        $dbRef = $this->dm->createDBRef($value, $mapping);
-        if ($inNewObj) {
+        $dbRef = $this->dm->createReference($value, $mapping);
```
jmikola: May want to rename this to `$ref` as well.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -1422,23 +1418,28 @@ private function getWriteOptions(array $options = array())
...
-        $keys = ['$ref' => true, '$id' => true, '$db' => true];
-        if ($mapping['storeAs'] === ClassMetadataInfo::REFERENCE_STORE_AS_ID) {
-            unset($keys['$db']);
-        }
-        if (isset($mapping['targetDocument'])) {
-            unset($keys['$ref'], $keys['$db']);
+
+        if ($mapping['storeAs'] === ClassMetadataInfo::REFERENCE_STORE_AS_REF) {
+            $keys = ['id' => true];
+        } else {
```
jmikola: As with `createReference()`, should we make this more defensive by expecting one of the two DBRef-related strategies?

lib/Doctrine/ODM/MongoDB/Query/Expr.php
```diff
@@ -73,28 +73,34 @@ public function references($document)
     {
         if ($this->currentField) {
             $mapping = $this->getReferenceMapping();
-            $dbRef = $this->dm->createDBRef($document, $mapping);
+            $dbRef = $this->dm->createReference($document, $mapping);
```
jmikola: Rename to `$ref`?

lib/Doctrine/ODM/MongoDB/Query/Expr.php
```diff
@@ -73,28 +73,34 @@ public function references($document)
...
             $storeAs = array_key_exists('storeAs', $mapping) ? $mapping['storeAs'] : null;
 
             if ($storeAs === ClassMetadataInfo::REFERENCE_STORE_AS_ID) {
                 $this->query[$mapping['name']] = $dbRef;
             } else {
-                $keys = array('ref' => true, 'id' => true, 'db' => true);
+                if ($storeAs === ClassMetadataInfo::REFERENCE_STORE_AS_REF) {
+                    $keys = ['id' => true];
+                } else {
```
jmikola: Another place we might want to be defensive about `$storeAs` values.

lib/Doctrine/ODM/MongoDB/Query/Expr.php
```diff
@@ -73,28 +73,34 @@ public function references($document)
...
                 }
 
                 foreach ($keys as $key => $value) {
-                    $this->query[$mapping['name'] . '.$' . $key] = $dbRef['$' . $key];
+                    $this->query[$mapping['name'] . '.' . $key] = $dbRef[$key];
                 }
             }
         } else {
-            $dbRef = $this->dm->createDBRef($document);
+            @trigger_error('Calling ' . __METHOD__ . ' without a current field set will no longer be possible in ODM 2.0.', E_USER_DEPRECATED);
```
jmikola: Same question as before about role of `@`.

lib/Doctrine/ODM/MongoDB/Query/Expr.php
```diff
@@ -111,28 +117,34 @@ public function includesReferenceTo($document)
...
+            $dbRef = $this->dm->createReference($document, $mapping);
             $storeAs = array_key_exists('storeAs', $mapping) ? $mapping['storeAs'] : null;
 
             if ($storeAs === ClassMetadataInfo::REFERENCE_STORE_AS_ID) {
                 $this->query[$mapping['name']] = $dbRef;
             } else {
-                $keys = array('ref' => true, 'id' => true, 'db' => true);
+                if ($storeAs === ClassMetadataInfo::REFERENCE_STORE_AS_REF) {
+                    $keys = ['id' => true];
+                } else {
```
jmikola: Defensive opportunity.

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/LookupTest.php
```diff
@@ -4,11 +4,17 @@
 
 class LookupTest extends \Doctrine\ODM\MongoDB\Tests\BaseTest
 {
-    public function testLookupStage()
+    /**
+     * @before
+     */
+    public function prepareTest()
```
jmikola: Is this an alternative to `setUp()` and calling `parent::setUp()`?

lib/Doctrine/ODM/MongoDB/DocumentManager.php
alcaeus: I'm not sure why, but I remember Symfony handling their deprecations like that and we've also done it this way previously.

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/LookupTest.php
alcaeus: Exactly - not sure where the advantages of each are, but I've started using `@before` and `@after` instead of `setUp` and `tearDown` in test classes and only use the latter in abstract TestCase classes.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataInfo.php
alcaeus: I'll relax the doc typehint, we can't typehint in the method since it can either be a reference array (`ref` or `dbRef` or an identifier, which again could be a `string` or `int`.

docs/en/reference/annotations-reference.rst
```diff
@@ -1214,24 +1214,26 @@ documents.
...
+    ``ref`` an embedded object containing the ``id`` field and (optionally) a
+    discriminator. ``dbRef`` and ``dbRefWithDb`` store a `DBRef`_ object. They
+    are deprecated in favor of ``ref``. Note that ``id`` references are not
+    compatible with the discriminators.
 -
     cascade - Cascade Option
 -
     discriminatorField - The field name to store the discriminator value within
-    the `DBRef`_ object.
+    the reference object.
```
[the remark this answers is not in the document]
alcaeus: It's specified where we explain the different `storeAs` options:
> Note that ``id`` references are not compatible with the discriminators.

I can add another note to each discriminator option for clarity.

lib/Doctrine/ODM/MongoDB/DocumentManager.php
```diff
@@ -736,10 +754,31 @@ public function createDBRef($document, array $referenceMapping = null)
...
+
+        if (!isset($referenceMapping['storeAs'])) {
```
malarzm: Shouldn't this be fixed on class metadata level?

lib/Doctrine/ODM/MongoDB/Query/Expr.php
```diff
@@ -111,29 +125,43 @@ public function includesReferenceTo($document)
...
             $mapping = $this->getReferenceMapping();
-            $dbRef = $this->dm->createDBRef($document, $mapping);
+            $reference = $this->dm->createReference($document, $mapping);
             $storeAs = array_key_exists('storeAs', $mapping) ? $mapping['storeAs'] : null;
 
-            if ($storeAs === ClassMetadataInfo::REFERENCE_STORE_AS_ID) {
-                $this->query[$mapping['name']] = $dbRef;
-            } else {
-                $keys = array('ref' => true, 'id' => true, 'db' => true);
+            switch ($storeAs) {
```
malarzm: Same here

lib/Doctrine/ODM/MongoDB/DocumentManager.php
alcaeus: Since it's a public method I decided to be defensive regarding what we receive. There's no guarantee someone will pass an actual mapping, they could call it with a partial mapping array (e.g. our tests). One more reason to look forward to metadata objects.

lib/Doctrine/ODM/MongoDB/Query/Expr.php
```diff
@@ -73,29 +73,43 @@ public function references($document)
...
```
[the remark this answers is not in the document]
alcaeus: I can try and extract the logic to a common method - maybe I can use `DocumentManager::createReference` for it.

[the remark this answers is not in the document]
malarzm: No big deal if it's complicated or anything, just occurred to me while reading :)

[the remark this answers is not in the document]
alcaeus: Nah, it's worth a shot :)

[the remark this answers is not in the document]
alcaeus: Turns out the methods are similar to suggest extracting the logic to a separate method, but different enough to not really change anything. Decided to skip it after all.

Issue #1195, opened 2015-08-09
EmbedMany persistence doesn't create empty array


ofetisov:
Hi there! I've encountered a tricky problem and couldn't find a appropriate solution. I have document with EmbedMany field like

```
    // It is configured to use discriminator map I think it doesn't has 
    // any influence on my question so i omitted the details.

    /**
     * @ODM\EmbedMany(
     *     discriminatorField="t",
     *     discriminatorMap={"a"="ClassA"}
     * )
     */
    protected $comments = array();
```

So the problem is: when i'm trying to save (persist & flush) new document with empty comments field - ODM doesn't creates comment field at all. And i need to have empty array there to be able to run [$size](http://docs.mongodb.org/manual/reference/operator/aggregation/size/) queries on this field.

I've debugged a little bit and found the root of the problem. [PersistenceBuilder (line 77-80)](https://github.com/doctrine/mongodb-odm/blob/1.0.0-BETA13/lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php#L77-L80)

```
    // @ReferenceMany and @EmbedMany are inserted later
    if ($mapping['type'] === ClassMetadata::MANY) {
        continue;
    }
```

It skips this field here, and never "inserts it later" due to the fact that no EmbedDocument is scheduled to persist/insert.

So could somebody give me an advice how to deal with this?

malarzm:
@ofetisov you can use [$ifNull](http://docs.mongodb.org/manual/reference/operator/aggregation/ifNull/#exp._S_ifNull):

```
db.col.aggregate([ 
    { 
        "$project": { 
            someValue: 1, 
            nb: { "$size": { "$ifNull": [ "$array", [  ] ] } } } 
    } 
]);
```

ofetisov:
@malarzm Thank you for quick response! Yes, I know about $ifNull but i would prefer to have empty array in my DB. Maybe there is some way to configure embedMany?

malarzm:
> Maybe there is some way to configure embedMany?

Sorry, there is not or at least I can't recall it :)

ofetisov:
@malarzm bad news =( But in any case thank you for your time.

josefsabl:
@malarzm Very bad news indeed. We ran into serious problems due to this too. Why I think this is a bug and not a feature?

1) Mongo itself supports empty array with embedded documents so no reason to say that it is a bad practice.
2) It is inconsistent with other field types that respect default php value including empty arrays.

[...]

PR #1624, opened 2017-07-28
ReferenceMany: insert an empty array

ajax13:
Related to #1195

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
@@ -104,7 +104,7 @@ class PersistenceBuilder
             // We're excluding collections using addToSet since there is a risk
             // of duplicated entries stored in the collection
             } elseif ($mapping['type'] === ClassMetadata::MANY && ! $mapping['isInverseSide']
-                    && $mapping['strategy'] !== ClassMetadataInfo::STORAGE_STRATEGY_ADD_TO_SET && ! $new->isEmpty()) {
+                    && $mapping['strategy'] !== ClassMetadataInfo::STORAGE_STRATEGY_ADD_TO_SET) {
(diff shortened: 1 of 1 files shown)


malarzm:
@ajax13 thanks for the PR! We require though that each change has a corresponding test so there will be no regressions in the future. Also I'm afraid that simply putting empty arrays after years of not doing anything will be a BC break for people looking for no existing field through query builder... thoughts on this @alcaeus @jmikola?

jmikola:
> Also I'm afraid that simply putting empty arrays after years of not doing anything will be a BC break for people looking for not existing field through query builder...

I think a BC-friendly approach would be having an option on the EmbedMany or ReferenceMany option that allows an empty array to be created if the collection is empty (instead of leaving the field omitted).

If I had to guess, the current behavior of omitting empty collections is related to [`@Field`](http://docs.doctrine-project.org/projects/doctrine-mongodb-odm/en/latest/reference/annotations-reference.html#field) and `nullable` (by default, fields with `null` values are unset in the document).

Otherwise, this seems like something you might simply change in 2.0 and document in the release notes.

alcaeus:
> I think a BC-friendly approach would be having an option on the EmbedMany or ReferenceMany option that allows an empty array to be created if the collection is empty

I agree on this and I also agree on using `nullable`, with the only difference that we wouldn't store `null` but rather `[]` (or `{}` if using `set` or `atomicSet` strategies).

malarzm:
Putting into 2.0 bucket

spiider8:
Hi, will be this funkcionality merged sometime pls?

malarzm:
We can't merge the PR as-is, it requires more work. Somebody interested needs to pick it up

khaperets:
> Hi, will be this funkcionality merged sometime pls?

I hope someone fixes it

malarzm:
Superseded by #2547

PR #1628, opened 2017-08-11
Don't generate proxy classes for embedded documents

### LINKED ISSUE #1627: Don't generate proxy classes for embedded documents
Originally reported as https://github.com/doctrine/common/issues/813.

With doctrine/common 2.8.0, generating proxy objects for final classes throws an error. Since documents should not be final, this does not cause any issues. However, ODM also generates proxy classes for embedded documents, which shouldn't happen because they won't be used.

alcaeus:
Fixes #1627.

Not sure about tests here, but it fixes the issue.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Tools/Console/Command/GenerateProxiesCommand.php
+++ b/lib/Doctrine/ODM/MongoDB/Tools/Console/Command/GenerateProxiesCommand.php
@@ -19,6 +19,7 @@
 
 namespace Doctrine\ODM\MongoDB\Tools\Console\Command;
 
+use Doctrine\ODM\MongoDB\Mapping\ClassMetadata;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console;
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ODM/MongoDB/Tools/Console/Command/GenerateProxiesCommand.php
```diff
@@ -62,7 +63,9 @@ protected function execute(Console\Input\InputInterface $input, Console\Output\O
     {
         $dm = $this->getHelper('documentManager')->getDocumentManager();
         
-        $metadatas = $dm->getMetadataFactory()->getAllMetadata();
+        $metadatas = array_filter($dm->getMetadataFactory()->getAllMetadata(), function (ClassMetadata $classMetadata) {
+            return !$classMetadata->isEmbeddedDocument;
```
malarzm: I think we should also exclude `$classMetadata->isQueryResultDocument` while we're here (and maybe `$classMetadata->isMappedSuperclass` too)?

malarzm:
> Not sure about tests here, but it fixes the issue.

I wouldn't worry too much here, it's only commands ;)

lib/Doctrine/ODM/MongoDB/Tools/Console/Command/GenerateProxiesCommand.php
alcaeus: I'll just change the check to `isDocument`, that's simpler. I'll have to think about `isQueryResultDocument` - there may be the need for proxies there.

alcaeus: Updated. Turns out we don't have `isDocument`, so I can't test against that. We also don't need to generate proxies for query result documents, so I'll add that when merging down to master. 👍

PR #1631, opened 2017-08-17
Don't generate proxies for query result documents

alcaeus:
Follows up on #1628 and adds query result documents to the proxy generation exclusion list.


malarzm:
:+1:

PR #1634, opened 2017-08-22
Enable a custom starting id

stampycode:
Need to be able to set a custom starting point for an auto-increment field - to prevent collisions when combining datasets, for example.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Id/IncrementGenerator.php
+++ b/lib/Doctrine/ODM/MongoDB/Id/IncrementGenerator.php
@@ -38,6 +38,7 @@ class IncrementGenerator extends AbstractIdGenerator
 {
     protected $collection = null;
     protected $key = null;
+    protected $startingId = 1;
 
     public function setCollection($collection)
     {
@@ -47,6 +48,11 @@ class IncrementGenerator extends AbstractIdGenerator
     public function setKey($key)
     {
         $this->key = $key;
+    }
+    
+    public function setStartingId($startingId)
+    {
+        $this->startingId = $startingId;
     }
 
     /** @inheritDoc */
@@ -59,7 +65,7 @@ class IncrementGenerator extends AbstractIdGenerator
         $key = $this->key ?: $dm->getDocumentCollection($className)->getName();
 
         $query = array('_id' => $key);
-        $newObj = array('$inc' => array('current_id' => 1));
+        $newObj = array('$inc' => array('current_id' => $this->startingId));
 
         $command = array();
         $command['findandmodify'] = $coll;
--- a/lib/Doctrine/ODM/MongoDB/Id/IncrementGenerator.php
+++ b/lib/Doctrine/ODM/MongoDB/Id/IncrementGenerator.php
@@ -70,7 +70,7 @@ class IncrementGenerator extends AbstractIdGenerator
          * Results in error: Cannot update 'current_id' and 'current_id' at the same time
          */
         $command = array(
-            'findandmodify' => $coll,
+            'findAndModify' => $coll,
             'query' => array('_id' => $key),
             'update' => array('$inc' => array('current_id' => 1)),
             'new' => true,
--- a/lib/Doctrine/ODM/MongoDB/Id/IncrementGenerator.php
+++ b/lib/Doctrine/ODM/MongoDB/Id/IncrementGenerator.php
@@ -71,19 +71,25 @@ class IncrementGenerator extends AbstractIdGenerator
          */
         $command = array(
             'findAndModify' => $coll,
-            'query' => array('_id' => $key),
-            'update' => array('$inc' => array('current_id' => 1)),
+            'query' => array('_id' => $key, 'current_id' => array('$exists' => true)),
+            'update' => array('$inc' => array('current_id' => $this->startingId)),
+            'upsert' => false,
             'new' => true,
         );
         $result = $db->command($command);
 
         /*
          * Updated nothing - counter doesn't exist, creating new counter.
+         * Not bothering with {$exists: false} in the criteria as that won't avoid
+         * an exception during a possible race condition.
          */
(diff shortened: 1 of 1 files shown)


lib/Doctrine/ODM/MongoDB/Id/IncrementGenerator.php
```diff
@@ -59,7 +65,7 @@ public function generate(DocumentManager $dm, $document)
         $key = $this->key ?: $dm->getDocumentCollection($className)->getName();
 
         $query = array('_id' => $key);
-        $newObj = array('$inc' => array('current_id' => 1));
+        $newObj = array('$inc' => array('current_id' => $this->startingId));
```
alcaeus: This doesn't change the starting value, but changes the increment. Setting `startingId` to 5 would create a sequence of `[5, 10, 15, 20, ...]`, not `[5, 6, 7, 8, 9, ...]` as one would expect.

stampycode: oops.

alcaeus: Had to look it up in the docs, but you can use [`$setOnInsert`](https://docs.mongodb.com/manual/reference/operator/update/setOnInsert/) to set the initial value if the document was inserted:
```
$newObj = [
    '$inc' => ['current_id' => 1],
    '$setOnInsert' => ['current_id' => $this->startingId + 1],
];
```

stampycode: @alcaeus you legend. Will commit shortly. With tests.

lib/Doctrine/ODM/MongoDB/Id/IncrementGenerator.php
```diff
@@ -58,16 +64,31 @@ public function generate(DocumentManager $dm, $document)
...
-
-        $command = array();
-        $command['findandmodify'] = $coll;
-        $command['query'] = $query;
-        $command['update'] = $newObj;
-        $command['upsert'] = true;
-        $command['new'] = true;
+        /*
+         * Unable to use '$inc' and '$setOnInsert' together due to known bug.
+         * @see https://jira.mongodb.org/browse/SERVER-10711
```
alcaeus: Almost 4 years and counting. Yay! Workaround looks good though 👍

stampycode: Only thing that isn't 100% is the atomicity - I assume the upsert is atomic, but this workaround certainly isn't. Not sure whether adding a collection lock would suck performance...

alcaeus: The upsert would be atomic, yes. With the workaround, I'm hoping that the combination of "how often will this feature be used" and "how often would an upsert have to happen" narrows it down enough for it not to be a problem.

jmikola: I think we can preserve atomicity with the following two-step process:

```php
// 1. Optimize for the common case: increment existing sequence
$command = [
    'findAndModify' => $coll,
    'query' => ['_id' => $key, 'current_id' => ['$exists' => true]],
    'update' => ['$inc' => ['current_id' => $this->startingId]],
    'upsert' => false,
    'new' => true,
];

/* 2. If no result was returned, we upsert a new sequence. Don't bother to
 * incorporate {$exists: false} into the criteria as that doesn't won't avoid
 * an exception during a possible race condition. */
$command = [
    'findAndModify' => $coll,
    'query' => ['_id' => $key],
    'update' => ['$inc' => ['current_id' => $this->startingId]],
    'upsert' => true,
    'new' => true,
];

/* 3. If a duplicate key exception was thrown, we encountered a race condition
 * where another process created the sequence between our previous findAndModify
 * commands. In that case, we can ignore the exception and attempt the first,
 * increment approach one more time (perhaps throwing our exception if that then
 * fails -- indicating that someone is rapidly inserting and deleting documents
 * in the sequence table. */
```

alcaeus:
@stampycode could you take a look at the issues @jmikola and I mentioned? I'd like to wrap up ODM 1.2 soon and this is currently on the milestone.

lib/Doctrine/ODM/MongoDB/Id/IncrementGenerator.php
stampycode: Ok I've made some of the recommended changes, just need to find and catch the duplicate key exception...
... and write tests...

stampycode:
@jmikola @alcaeus 
So I've run into a stumbling block in my tests.
Using the [code](https://github.com/doctrine/mongodb-odm/pull/1634/commits/c3ec6a5f28e65f06e98b7b0c8f1e3f2313148e28) I originally suggested works ok:
+ set `$startingId` to 1000001
+ first element fails to find non-existent counter
+ creates counter with number 1000001
+ returns 1000001
+ second request finds counter, increments it
+ returns 1000002

Unfortunately, and for reasons I have not yet fathomed, the [suggested changes](https://github.com/doctrine/mongodb-odm/pull/1634/commits/946b1ef492be119a600b1547152deec2060ea4de) do not have the desired effect:
+ set `$startingId` to 1000001
+ first element fails to find non-existent counter
+ creates counter with number 2000001
+ returns 2000001
+ second request finds counter, increments it
+ returns 3000001

I have no idea what is going on here.

alcaeus:
@stampycode I took a look at the changes, looks like there was an error in the code @jmikola suggested. The first command of course has to increment the `current_id` field by one, not by the starting value. Only the second command that is supposed to upsert a record may increment it with the starting value. I've pushed a fix to #1661 and merge it once the build has finished. Thanks!

jmikola:
@alcaeus: Indeed, my example was incrementing by `startingId` instead of `1`. @stampycode: Sorry for the misleading suggestion!

PR #1639, opened 2017-09-02
Fix Lookup::from() return type

olvlvl:
Hi,

This PR fixes `Lookup::from()` return type, which is either `Lookup` or `BaseStage\Lookup`. `$this` should be used when the same instance is returned, which is not the case here.


alcaeus:
> which is not the case here.

Why? The parent class returns $this as well, so the return type is correct: the same instance is returned.

olvlvl:
You're right, it's the same instance.

PR #1640, opened 2017-09-02
Fix Match::$query type

olvlvl:
Hi,

`Doctrine\MongoDB\Query\Expr` doesn't have the methods `references()` or `includesReferenceTo()`, but `Doctrine\ODM\MongoDB\Query\Expr` does.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Match.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Match.php
@@ -20,12 +20,18 @@
 namespace Doctrine\ODM\MongoDB\Aggregation\Stage;
 
 use Doctrine\MongoDB\Aggregation\Stage as BaseStage;
+use Doctrine\ODM\MongoDB\Query\Expr;
 
 /**
  * Fluent interface for building aggregation pipelines.
  */
 class Match extends BaseStage\Match
 {
+    /**
+     * @var Expr
+     */
+    protected $query;
+
     /**
      * @param object $document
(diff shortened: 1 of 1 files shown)


lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Match.php
```diff
@@ -20,13 +20,19 @@
...
 
 /**
  * Fluent interface for building aggregation pipelines.
  */
 class Match extends BaseStage\Match
 {
     /**
+     * @var Expr
+     */
+    protected $query;
```
alcaeus: This does not need to be redeclared just to change a typehint. Use the property annotation on the class if you want to fix that.

olvlvl: The `@property` annotation is for public properties, this one is protected.

alcaeus: Oh. I forgot, sorry about that. In that case, I’d much rather just leave this out. It’s a non-issue at the docs level, and we’re changing the API for 2.0 anyways: the base classes of the query and aggregation builders will move into this project. doctrine/mongodb won’t be adapted for ext-mongodb (its job is done by mongodb/mongodb now) and there’s no point bothering with that two-level API. It’s caused enough problems as it is.

olvlvl: Thanks for the detailed explanation.

PR #1642, opened 2017-09-04
Fix DocumentManager::getHydratorFactory() return type

olvlvl:
Hi,

`DocumentManager::getHydratorFactory()` return type is wrong, see https://github.com/doctrine/mongodb-odm/blob/f7379e4eeaffcd33d83fb4b0fcc4ccce1acb8a4e/lib/Doctrine/ODM/MongoDB/DocumentManager.php#L84


malarzm:
Thank you very much!

PR #1646, opened 2017-09-25
Fix invalid test data for UUID bin data

alcaeus:
This fixes failures when running the tests against the new 1.3.0 mongodb driver version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ env:
 matrix:
   include:
     - php: 5.6
-      env: DRIVER_VERSION="1.5.8" SERVER_VERSION="3.0" COMPOSER_FLAGS="--prefer-lowest" KEY_ID="7F0CEB10"
+      env: DRIVER_VERSION="1.6.3" SERVER_VERSION="3.0" COMPOSER_FLAGS="--prefer-lowest" KEY_ID="7F0CEB10"
     - php: 5.6
       env: DRIVER_VERSION="stable" SERVER_VERSION="3.2" KEY_ID="EA312927"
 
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -22,7 +22,7 @@ env:
 matrix:
   include:
     - php: 5.6
-      env: DRIVER_VERSION="1.5.8" SERVER_VERSION="3.0" COMPOSER_FLAGS="--prefer-lowest" KEY_ID="7F0CEB10"
+      env: DRIVER_VERSION="1.6.3" SERVER_VERSION="3.0" COMPOSER_FLAGS="--prefer-lowest" KEY_ID="7F0CEB10"
```
alcaeus: No idea why tests failed on 1.5.8, but using 1.6.3 fixes build failures. I couldn't install old driver versions on my machine to investigate this any further, so this is what I came up with 🤷‍♂️

alcaeus: After further investigation, it turns out that travis-ci now has MongoDB 3.2 installed by default, which prevents us from installing 3.0 via apt. Since the 1.5 driver doesn't like MongoDB 3.2, I'd just bump the version we test; ODM 1.2 will bump this anyways due to a requirement bump to `doctrine/mongodb`, which requires `ext-mongo` 1.6.7 or newer.

alcaeus: @jmikola correct - I jumped through versions and figured out that 1.6.3 is the lowest that will make the build pass. Since I'm trying to release 1.2, which would end support for 1.1 in the process, I don't think it's worth investigating this more. I'll also have to spend some more time getting 3.0 running on travis-ci again, which might fix this problem altogether.

tests/Doctrine/ODM/MongoDB/Tests/Functional/IdTest.php
```diff
@@ -276,12 +276,12 @@ public function getTestIdTypesAndStrategiesData()
...
      * @dataProvider getTestBinIdsData
      */
-    public function testBinIds($type, $expectedMongoBinDataType)
+    public function testBinIds($type, $expectedMongoBinDataType, $id)
     {
         $className = $this->createIdTestClass($type, 'none');
 
         $object = new $className();
-        $object->id = 'ABRWTIFGPEeSFf69fISAOA==';
+        $object->id = $id;
```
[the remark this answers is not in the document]
alcaeus: Fixed in #1650.

Issue #1051, opened 2015-03-19
[RFC] "Nested" calls to DocumentManager::flush()


alcaeus:
Hello all,

currently it is possible to `flush()` the `DocumentManager` multiple times while it is already flushing. One example is a lifecycle event subscriber that calls some method that inserts another document and flushes the `DocumentManager`.
Currently, this case is not handled, the `DocumentManager` will happily run the nested `flush()` operation which can cause a bunch of issues. My suggestion is that the `DocumentManager` prevents such calls, unless such functionality breaks existing behavior. Comments?

gruberro:
I'd like to see this change, as this would prevent me from a lot of trouble :+1:

malarzm:
I remember using this to flush single document during bigger flush and nothing bad happened to me in the meantime

alcaeus:
Let's say there is an event subscriber that listens to the ```onFlush``` event. That subscriber loops all changesets and does something if a specific field has changed. You've also got a second subscriber that listens to the ```postPersist``` event. That one creates a new document to be inserted and calls ```flush($document)```. This will cause the ```onFlush``` handler to be called a second time, and it will actually re-do whatever it has done for the first document since the document is still in the changesets (it will only be removed from there after the ```postFlush``` event has been triggered.

malarzm:
I'm not saying it can't break, just throwing my 2 cents saying that I'm currently using flush within a flush :)

alcaeus:
Ah, sorry :)
Yeah, I was thinking about how it could be made optional or whether the flushes could be correctly isolated. Technically, if the second flush tracked its own set of changes, it would be no problem.

alcaeus:
Long time no see - just referencing the comments over in the ORM repository:

> Calling `flush()` inside a lifecycle listener is indeed not allowed.

([Source](https://github.com/doctrine/doctrine2/issues/6659#issuecomment-329980340)

Given that, I'd try to come up with a solution for 2.0 along with a deprecation notice for 1.2.

malarzm:
#1722 prevents user from flushing more than once at a time

Issue #1584, opened 2017-04-27
Throw exceptions for disallowed operations during flush


araines:
I've been playing around with the events system a bit more recently and found that the combination of the documentation being a little light and the UnitOfWork silently failing makes for some fairly challenging debug sessions.

I'd like to propose that the UnitOfWork throws exceptions when disallowed operations take place.  For example, you cannot persist a new document during a postUpdate as it will be simply lost - so rather than just losing it, it would be nice to get an exception telling the developer what they are doing wrong.

alcaeus:
@araines thanks for bringing this up here after our conversion in IRC earlier today. I agree with you, the events are a bit complicated in this regard, which is why I'd go as far as saying that you shouldn't use them unless you know what you are doing. That being said, you can't know what you're doing unless we provide some good documentation.

I've tagged this issue as a documentation issue and will try to come up with some documentation about the events, especially:
* What does the flush process look like: changeset computation, order of operation
* At what stage are events triggered and what actions are permissible
* Ideally, add some common use cases of what events are used for and indicate what should and should not be done.

Now, these are just some ideas, obviously we can use all the input we can get here. All of those caveats of the event system make a lot of sense when you've dealt with them for years, but for the average developer it may be quite confusing and overwhelming.

> I'd like to propose that the UnitOfWork throws exceptions when disallowed operations take place.

Having previously experimented with functionality like that to prevent nested flushes (which caused issues due to excessive event handling involved) I'd caution against that, at least for now. There are just too many variables involved, and generally speaking, a single exception within the flush process or event handling process will cause all kinds of issues.

araines:
No problem - more than happy with documentation as the solution to this problem.  Whilst I don't mind diving through code of libraries to learn how they work, it's much nicer to be able to get going quickly based on the docs!

Also happy to contribute if I can be of service.

alcaeus:
#1647 starts this process by throwing deprecation notices when calling `flush` from an event subscriber/event listener/lifecycle callback. These calls will definitely be removed in 2.0. I'll have to think whether it's feasible to apply the very same logic to `persist` and other calls 🤔

malarzm:
> I'll have to think whether it's feasible to apply the very same logic to persist and other calls 🤔

I'm not sure whether it's needed, other events are quite safe to be called from one another and removing other documents while removing one is quite handy (or am I misunderstanding the direction?)

malarzm:
[...]

PR #1647, opened 2017-09-27
Throw deprecation warning on nested flushes

alcaeus:
As indicated in #1051, calling `flush` in an event subscriber can have unwanted side effects and shouldn't be done. To prepare people for this change, ODM 1.2 will throw deprecation warnings to give users a chance to fix offending event subscribers before 2.0 starts throwing exceptions.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
+++ b/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
@@ -261,6 +261,11 @@ class UnitOfWork implements PropertyChangedListener
     private $embeddedDocumentsRegistry = array();
 
     /**
+     * @var int
+     */
+    private $commitsInProgress = 0;
+
+    /**
      * Initializes a new UnitOfWork instance, bound to the given DocumentManager.
      *
      * @param DocumentManager $dm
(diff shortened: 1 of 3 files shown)


lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -407,54 +412,62 @@ public function commit($document = null, array $options = array())
...
+            $this->documentDeletions =
+            $this->documentChangeSets =
+            $this->collectionUpdates =
+            $this->collectionDeletions =
+            $this->visitedCollections =
+            $this->scheduledForDirtyCheck =
+            $this->orphanRemovals =
+            $this->hasScheduledCollections = array();
+        } finally {
+            $this->commitsInProgress--;
```
malarzm: We should have a test checking that `commitsInProgress` is decreased in case of exception or any error.

alcaeus: Test added.

alcaeus:
Looking at #1584, the suggestion is to also deprecate calls to other persistence methods (`persist`, `remove`, etc. during event subscriber calls. These calls could be allowed in `preFlush` and `onFlush` event listeners, but considered unsafe afterwards because any new operations scheduled might not be executed. What do you think @malarzm?

malarzm:
I think it will require a lot of ifology as for instance it is safe to call `remove` on `postPersist` as removals are last. But lemme think about it a bit more :)

alcaeus:
> I think it will require a lot of ifology as for instance it is safe to call `remove` on `postPersist` as removals are last. But lemme think about it a bit more :)

Which is exactly why I wouldn't allow any events at all after `onFlush`. There are just too many variables for us to figure out whether the event call is allowed, and it gets even worse for users that don't know the inner workings of UnitOfWork. I see it as an opportunity to prevent users from shooting themselves in the foot 😉

malarzm:
> I see it as an opportunity to prevent users from shooting themselves in the foot :wink:

While this is a noble thing to do, this also takes away feature that is quite powerful and working without side effects (to some degree of course) :P I mean I can see people wanting to remove a document after successful update of another one, and if somebody is caring enough to go look into UoW's internals to figure out it's OK to do so, I'm happy for them.

alcaeus:
> I can see people wanting to remove a document after successful update of another one

That's a valid use case, although I think people should be handling this in `postFlush`. However, when `postFlush` has been called, changeSets have not been cleared, so it isn't save to make another call to `flush` just yet.

Furthermore, even by looking into UnitOfWork, scheduling a new operation while the flush is in progress is a big no-go as you can't be sure it will be executed (for one, order of operations may forbid it or we may ignore it because we already handled insertions for that type, etc.).

alcaeus:
Merging without additional deprecations - we can always re-visit this. Ideally we'll find a different solution to the problem at hand.

PR #1658, opened 2017-10-16
Allow unauthenticated install of mongodb-org

alcaeus:
Fixes https://travis-ci.org/doctrine/mongodb-odm/jobs/287745628#L501

Also removes the tests against MongoDB 3.0 since travis-ci comes with 3.2 installed and we can't downgrade. It's not optimal, but I don't have a better solution at this time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ before_install:
   - sudo apt-get update -qq
 
 install:
-  - sudo apt-get install mongodb-org
+  - sudo apt-get --allow-unauthenticated install mongodb-org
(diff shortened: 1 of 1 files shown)


jmikola:
If you're resigned to testing on 3.2+, would it be worth switching over to container builds (https://github.com/mongodb/mongo-php-library/pull/411)?

alcaeus:
@jmikola that's a good idea, sadly I remember needing `sudo` to install mongo-orchestration. It's a long time ago, so it might be worth another shot.

PR #1660, opened 2017-10-20
Add upgrade document for ODM 1.2

alcaeus:
This pull request adds an UPGRADE document for ODM 1.2 and adds a deprecation we missed. This document will evolve as remaining pull requests are merged.

### DIFF
--- a/UPGRADE-1.2.md
+++ b/UPGRADE-1.2.md
@@ -1,0 +1,431 @@
+# UPGRADE FROM 1.x TO 1.2
+
+## Mapping changes
+
+### Immutable documents
+
+Documents can be marked as immutable by setting the `readOnly` option in the
+document mapping:
+
+```php
+/**
+ * @ODM\Document(readOnly=true)
+ */
+class ReadOnlyDocument
+{
+    // ...
+}
+```
+```xml
+<?xml version="1.0" encoding="UTF-8"?>
+
+<doctrine-mongo-mapping xmlns="http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping"
+                  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                  xsi:schemaLocation="http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping
+                  http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping.xsd">
+
+    <document name="ReadOnlyDocument" read-only="true">
+        <!-- ... -->
+    </document>
+</doctrine-mongo-mapping>
+```
+```yml
+ReadOnlyDocument:
+  type: document
+  readOnly: true
+```
+
+Immutable documents will not be updated in the database after they've been
+persisted. You can still create a query builder and manually update data at your
+own risk.
+
+### Read preference
+
+In addition to specifying a read preference for a query, you can now specify a
+default read preference for a document using the `readPreference` option:
+
+```php
+/**
+ * @Document
+ * @ODM\ReadPreference("primaryPreferred", tags={
+ *   { "dc"="east" },
+ *   { "dc"="west" },
+ *   {  },
+ * })
+ */
+class ReadPreferenceDocument
+{
+    // ...
+}
+```
+```xml
+<?xml version="1.0" encoding="UTF-8"?>
+
+<doctrine-mongo-mapping xmlns="http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping"
+                  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                  xsi:schemaLocation="http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping
+                  http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping.xsd">
+
+    <document name="ReadPreferenceDocument">
+        <read-preference mode="primaryPreferred">
+            <tag-set>
+                <tag name="dc" value="east"/>
+            </tag-set>
+            <tag-set>
+                <tag name="dc" value="west"/>
+            </tag-set>
+            <tag-set />
+        </read-preference>
+
+        <!-- ... -->
+    </document>
+</doctrine-mongo-mapping>
+```
+```yml
+ReadPreferenceDocument:
+  type: document
+  readPreference:
+    mode: primaryPreferred
+    tagSets:
+      - { dc: east }
+      - { dc: west }
+      - {  }
+```
+
+Note that specifying a `readPreference` and `slaveOkay` at the same time will
+cause an exception.
+
+To replace an existing `slaveOkay` mapping, you can apply a `readPreference` of
+`secondaryPreferred`. 
+
+### Reference mapping
+
+When mapping references, ODM previously previously offered three options to 
+store a reference via the `storeAs` option:
+
+ * `storeAs=id` (previously known as `simple` reference): this reference type
+   simply stores the identifier of the referenced document in the owning
+   document.
+ * `storeAs=dbRef`: this reference type creates a DBRef object, but doesn't
+   include a value for the `$db` field
+ * `storeAs=dbRefWithDb`: similar to `dbRef`, but also writes the `$db` field
+
+Since `id` can't be used with discriminated references and the two `dbRef` types
+may be incompatible with aggregation pipelines, there is a new reference type
+`ref`. This reference type is similar to `dbRef` as it stores an object, but it 
+only stores the identifier of the reference as `id` (without a leading dollar 
+sign). This makes it usable with discriminated references (which are stored 
+inside the reference object) as well as aggregation pipelines. While not 
+officially deprecated, using the `dbRef` reference types is strongly limited 
+because of its [limited support in aggregation pipeline queries](https://jira.mongodb.org/browse/SERVER-14466).
+
+### Priming inverse references
+
+When specifying an inverse reference via `mappedBy` or `repositoryMethod`, you
+can now specify primers that will be run on mapped documents automatically:
+
+```php
+/** @Document */
+class Blog
+{
+    /** @ReferenceMany(targetDocument=Post::class, prime={"author"}) */
+    private $posts;
+}
+```
+```xml
+<?xml version="1.0" encoding="UTF-8"?>
+
+<doctrine-mongo-mapping xmlns="http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping"
+                  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                  xsi:schemaLocation="http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping
+                  http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping.xsd">
+
+    <document name="Blog">
+        <reference-many target-document="Post" field="posts">
+            <prime>
+                <field name="author" />
+            </prime>
+        </reference-many>
+    </document>
+</doctrine-mongo-mapping>
+```
+```yml
+Blog:
+  type: document
+  referenceMany:
+    posts:
+      targetDocument: Post
+      prime:
+        - author
+```
+
+### Query result documents
+
+Query result documents are similar to regular documents, except that they won't
+be written to the database. Query result documents are used to return objects
+from aggregation pipeline queries. You can map a query result document similar
+to a regular document:
+
+```php
+/** @QueryResultDocument */
+class UserPurchases
+{
+    /** @ReferenceOne(targetDocument="User", name="_id") */
+    private $user;
+
+    /** @Field(type="int") */
+    private $numPurchases;
+
+    /** @Field(type="float") */
+    private $amount;
+}
+```
+```xml
+<?xml version="1.0" encoding="UTF-8"?>
+<doctrine-mongo-mapping xmlns="http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping"
+                  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                  xsi:schemaLocation="http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping
+                  http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping.xsd">
+    <query-result-document name="UserPurchases">
+        <field fieldName="numPurchases" type="int" />
+        <field fieldName="amount" type="float" />
+        <reference-one field="user" target-document="Documents\User" name="_id" />
+    </query-result-document>
+</doctrine-mongo-mapping>
+```
+```yml
+UserPurchases:
+  type: queryResultDocument
+  fields:
+    user:
+      name: _id
+      targetDocument: Documents\User
+    numPurchases:
+      type: int
+    amount:
+      type: float
+```
+
+Attempting to persist a query result document will result in an exception.
+
+Query result documents can be used to hydrate results from an aggregation
+pipeline query:
+```php
+$purchaes = $documentManager->getRepository(Purchases::class)->createAggregationBuilder()
+    ->hydrate(UserPurchases::class)
+    ->group()
+        ->field('id')
+        ->expression('$user')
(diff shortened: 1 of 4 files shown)


UPGRADE-1.2.md
```diff
@@ -0,0 +1,415 @@
...
+```php
+/** 
+ * @ODM\Document(readOnly=true)
+ */
+class ReadOnlyDocument
+{
+    // ...
+}
+```
+```xml
```
jmikola: Does RST need line breaks between code blocks? I noticed the YAML example is also right up against this XML block.

UPGRADE-1.2.md
```diff
@@ -0,0 +1,415 @@
...
+<?xml version="1.0" encoding="UTF-8"?>
+
+<doctrine-mongo-mapping xmlns="http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping"
+                  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                  xsi:schemaLocation="http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping
+                  http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping.xsd">
```
jmikola: This indentation level seems arbitrary. Was it pulled from some other docs example? So long as it's consistent, I have no objection.

UPGRADE-1.2.md
```diff
@@ -0,0 +1,415 @@
...
+
+    <document name="ReadOnlyDocument" read-only="true">
+        <!-- ... -->
+    </document>
+</doctrine-mongo-mapping>
+```
+```yml
+ReadOnlyDocument:
+  type: document
+  readOnly: 1
```
jmikola: Why `1` instead of `true`?

UPGRADE-1.2.md
```diff
@@ -0,0 +1,415 @@
...
+```
+
+Immutable documents will not be updated in the database after they've been
+persisted. You can still create a query builder and manually update data at your
+own risk.
+
+### Read preference
+
+In addition to specifying read preference for a query, you can now specify a
```
jmikola: "a read preference"

UPGRADE-1.2.md
```diff
@@ -0,0 +1,415 @@
...
+default read preference for a document using the `readPreference` option:
+
+```php
+/**
+ * @Document
+ * @ODM\ReadPreference("primaryPreferred", tags={
+ *   { "dc"="east" },
+ *   { "dc"="west" },
+ *   {  },
+ * })
```
jmikola: Do annotations use `{ }` to denote arrays? I would have expected `[ ]`.

UPGRADE-1.2.md
```diff
@@ -0,0 +1,415 @@
...
+  readPreference:
+    mode: primaryPreferred
+    tagSets:
+      - { dc: east }
+      - { dc: west }
+      - {  }
+```
+
+Note that specifying a `readPreference` and `slaveOkay` at the same time will
+cause an exception.
```
jmikola: I realize it's below, but would it be good to repeat that `slaveOkay` is synonymous with a "secondaryPreferred" read preference and is therefore deprecated?

UPGRADE-1.2.md
```diff
@@ -0,0 +1,415 @@
...
+
+### Reference mapping
+
+When mapping references, you previously had three options to store a reference
+via the `storeAs` option:
```
jmikola: I think "ODM previously offered three options" reads better.

Also, should we have a line break between the preceding paragraph and list?

UPGRADE-1.2.md
jmikola: I just realized this is a Markdown file, not RST. Is this the first time we're presenting PHP, XML, and YAML examples in Markdown? I'm just curious if we have an existing format for that, or if the intention is just to present them in sequence. In RST, I assume these would be tabbed language examples.

UPGRADE-1.2.md
```diff
@@ -0,0 +1,415 @@
...
+ * `storeAs=id` (previously known as `simple` reference): this reference type
+   simply stores the identifier of the referenced document in the owning 
+   document.
+ * `storeAs=dbRef`: this reference type creates a DBRef object, but doesn't
```
jmikola: Do we want to link DBRef to something?

UPGRADE-1.2.md
```diff
@@ -0,0 +1,415 @@
...
+   include a value for the `$db` field
+ * `storeAs=dbRefWithDb`: similar to `dbRef`, but also writes the `$db` field
+ 
+Since `id` can't be used with discriminated references and the two `dbRef` types
+can't be used with aggregation builder, there is a new reference type `ref`.
```
jmikola: Thoughts on "DBRef objects may be incompatible with aggregation pipelines"? That suggests that this is rightfully a server limitation.

UPGRADE-1.2.md
```diff
@@ -0,0 +1,415 @@
...
+This reference type is similar to `dbRef` as it stores an object, but it only
+stores the identifier of the reference as `id` (without a leading dollar sign).
+This makes it usable with discriminated references (which are stored inside the
+rererence object) as well as aggregation pipelines. While not officially
```
jmikola: Typo: "reference"

UPGRADE-1.2.md
```diff
@@ -0,0 +1,415 @@
...
+deprecated, using the `dbRef` reference types is strongly limited because its
```
jmikola: "because of its"

UPGRADE-1.2.md
```diff
@@ -0,0 +1,415 @@
...
+limited use in aggregation pipeline queries.
```
jmikola: "limited support in aggregation pipelines".

Consider linking to https://jira.mongodb.org/browse/SERVER-22005 or https://jira.mongodb.org/browse/SERVER-14466.

UPGRADE-1.2.md
```diff
@@ -0,0 +1,415 @@
...
+
+### Priming inverse references
+
+When specifying an inverse reference via `mappedBy` or `repositoryMethod`, you
+can now specify primers which will be run on mapped documents automatically:
```
jmikola: "primers that will". "Which" typically follows a comma.

UPGRADE-1.2.md
```diff
@@ -0,0 +1,415 @@
...
+
+```php
+/** @Document */
+class Blog
+{
+    /** @ReferenceMany(targetDocument=Post::class, prime={"author"}) */
```
jmikola: Should "Post::class" be quoted, or does the annotation driver resolve this as PHP syntax?

UPGRADE-1.2.md
```diff
@@ -0,0 +1,415 @@
...
+      prime:
+        - author
+```
+
+### Query result documents
+
+Query result documents are similar to regular documents, except that they won't
+be written to the database. Query result documents are used to return objects
+from aggregation pipeline queries. You can map a query result document similar
+to a regular document:
```
jmikola: Is the intention to also use these for results from [view](https://docs.mongodb.com/manual/core/views/) queries?

This section also doesn't have an example of showing how to run a query that returns an instance of UserPurchases. Should we add that?

UPGRADE-1.2.md
```diff
@@ -0,0 +1,415 @@
...
+      type: float
+```
+
+Attempting to persist a query result document will result in an exception.
+
+### Validate mappings
+
+The new `odm:schema:validate` command allows you to validate your schema to
+ensure it is cached and read properly. Errors in mapping can sometimes only
+manifest themselves after the mapping has been loaded from cache (as disabled
```
jmikola: "from the cache"

UPGRADE-1.2.md
```diff
@@ -0,0 +1,415 @@
...
+features are not written to the metadata cache). This command gives you a chance
```
jmikola: What are "disabled features"?

UPGRADE-1.2.md
```diff
@@ -0,0 +1,415 @@
...
+to catch these errors.
+
+### Custom starting value for increment generators
+
+When using an increment ID generator for a document, you can now specify a
+custom starting ID that will be used if no record is present:
```
jmikola: "no record" might confuse users into thinking it refers to new documents.

Should we elaborate to say that "no record of the sequence"?

UPGRADE-1.2.md
```diff
@@ -0,0 +1,415 @@
...
+   take a look at the [`notablescan`](https://docs.mongodb.com/manual/reference/parameters/#param.notablescan)
+   option in the MongoDB server.
+ * The `name` and `fieldName` properties in the `@DiscriminatorField` annotation
+   have been deprecated and will be removed in ODM 2.0. Use the annotation
+   without specifying a property instead: `@DiscriminatorField("field")`
+ * Loading annotations through the `DoctrineAnnotations.php` file has been
+   deprecated. The file will be removed in ODM 2.0.
+ * The `Doctrine\ODM\MongoDB\Mapping\Driver\AnnotationDriver::registerAnnotationClasses`
+   method has been deprecated and will be removed in ODM 2.0. Use the 
+   `AnnotationRegistry::registerLoader` method to register an autoloader.
```
jmikola: Should the method names have `()` after them?

UPGRADE-1.2.md
```diff
@@ -0,0 +1,415 @@
...
+```
+
+As an alternative:
+```php
+class MyRepository extends DocumentRepository
+{
+    /**
+     * @return Order[]
+     */
+    public function findByStatus(string $status): array
```
jmikola: Does ODM always return multiple results as an array? I seem to recall these returning a Cursor.

UPGRADE-1.2.md
```diff
@@ -0,0 +1,415 @@
...
+        return $this->findOneBy(['email' => $email);
+    }
+}
+```
+
+### Abstract repository factory
+
+ODM 1.2 introduces an `AbstractRepositoryFactory` that you can extend if you're
+building your own repository factory. This class takes care of tracking
+repository instances and only requires you to implement the
```
jmikola: "requires that you implement"

UPGRADE-1.2.md
```diff
@@ -0,0 +1,415 @@
...
+    ->updateMany()
+    ->field('enabled')
+    ->set(true);
+```
+
+## Group helper
+
+The `group` method in the query builder has been deprecated and will be removed
+in ODM 2.0. The `group` command has been deprecated in MongoDB and is replaced
+by the `$group` aggregation pipeline stage.
```
jmikola: Although not technically replaced, I would suggest saying that `group` has been superseded by the `mapReduce` command and `$group` aggregation pipeline station, or generally suggest that users migrate old code to use aggregation instead.

UPGRADE-1.2.md
```diff
@@ -0,0 +1,415 @@
...
+## Deprecations
+
+ * The `Doctrine\ODM\MongoDB\EagerCursor` class has been deprecated and will be
+   removed in ODM 2.0. Use the `Doctrine\ODM\MongoDB\Cursor` class instead.
+ * The `Doctrine\ODM\MongoDB\MongoDBException::queryNotIndexed` method has been
+   deprecated and will be removed in ODM 2.0.
+ * The `Doctrine\ODM\MongoDB\Query\FieldExtractor` class has been deprecated and
+   will be removed in ODM 2.0.
+ * The `$requireIndexes` argument to `Doctrine\ODM\MongoDB\Query\Query::__construct`
+   has been deprecated and will be removed in ODM 2.0.
```
jmikola: `$requireIndexes` is currently the second-to-last argument in the Query constructor. Should we simply say that the constructor's signature will change in ODM 2.0? I expect we might want to do quite a bit more refactoring on it to avoid such a long signature. Since keywork args aren't an issue, we might move towards an options array for all non-required params.

UPGRADE-1.2.md
alcaeus: Yeah, I pulled it from one of our test documents. I'm not sure what the correct indentation should be, so I'm open to suggestions.

UPGRADE-1.2.md
alcaeus: Good point, will change.

UPGRADE-1.2.md
alcaeus: I've hardly ever used `{ }`. I've copied the annotation from our tests, so I can give `[ ]` a quick test.

UPGRADE-1.2.md
alcaeus: Agree, might let people know where the limitation lies in this case.

UPGRADE-1.2.md
alcaeus: Starting with version 1.2 which we now require, you can use `::class` syntax in annotations. Since you're not presenting a string literal but rather a class constant, it goes unquoted.

UPGRADE-1.2.md
alcaeus: I was torn between adding an example and linking to the official aggregation builder docs. I can add an example.

As for views, I haven't given them much thought yet. One option would be to provide an aggregation pipeline in the document mapping, which then automatically creates the view when creating the schema. The only other use-case I've had for query result documents where `group` queries, but there's no point adding those there now that we've deprecated that. Not sure how useful hydration is for `mapReduce`.

UPGRADE-1.2.md
jmikola: No worries.

UPGRADE-1.2.md
jmikola: Just in case, note that my  comment was about the list of tag sets, not the empty tag set `{  }` (which is rightfully an empty object) a line above this comment.

UPGRADE-1.2.md
alcaeus: Bad choice of wording on my part. Basically, we selectively serialize metadata information depending on some flags, for example https://github.com/doctrine/mongodb-odm/blob/master/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php#L128..L136. This can lead to problems because the mapping is fine when not cached, but completely different when loading from cache.

UPGRADE-1.2.md
alcaeus: So did I until I checked the repository: https://github.com/doctrine/mongodb-odm/blob/master/lib/Doctrine/ODM/MongoDB/DocumentRepository.php#L173..L186

That's something we'll need to address in 2.0.

UPGRADE-1.2.md
alcaeus: Looking at the result, Markdown doesn't need this. If we were to move this to RST, this would be tabbed language examples.
I don't remember doing this before, which is why I shamelessly copied the syntax Symfony uses in their documents.

UPGRADE-1.2.md
alcaeus: I might have copied that from one of our tests - making sure we use `true`.

UPGRADE-1.2.md
alcaeus: `{ }` it is:
> [Syntax Error] Expected PlainValue, got '[' at position 55 in class Doctrine\ODM\MongoDB\Tests\Functional\DocumentWithReadPreference.

UPGRADE-1.2.md
alcaeus: Decided to remove the disabled feature part.

UPGRADE-1.2.md
alcaeus: I removed this since we'll most likely completely refactor the constructor anyways. We also only added `deprecated` as description, with no deprecation warning if it's used in the constructor (since the query builder already throws a deprecation notice).

alcaeus:
@jmikola thanks for the comprehensive review! I've added documentation for how to use query result documents and applied most of the wording changes you suggested.

UPGRADE-1.2.md
jmikola: Perhaps I mistook with this the query builder, which I believe does return Cursors by default.

PR #1666, opened 2017-10-25
Recommend storeAs=ref in docs

amcsi:
To  be able to take use of both polymorphism and the ability to use $lookup in aggregations.


alcaeus:
Thanks!

PR #1667, opened 2017-10-25
add double-quotes to documentation

jdeniau:


docs/en/reference/reference-mapping.rst
```diff
@@ -371,7 +371,7 @@ The ``storeAs`` option has the following possible values:
 .. note::
 
     For backwards compatibility ``storeAs=dbRefWithDb`` is the default, but
-    ``storeAs=dbRef`` is the recommended setting.
+    ``storeAs=ref`` is the recommended setting.
```
jmikola: Should both of these lines also add quotes around the `storeAs` value?

jdeniau: Well I don't know why these commits appear in my PR but I think they should yes.

jmikola: Ah, I didn't realize this had pulled in commits from https://github.com/doctrine/mongodb-odm/pull/1666.

If you like, I suppose you can rebase on `1.2.x` and amend your commit to fix those instances. Otherwise, we can probably just merge your last commit as-is.

alcaeus: The reason for those commits appearing is me changing the base to 1.2.x without having an environment to rebase them on. I’ll fix this shortly.

alcaeus: I've changed the base back - I didn't realize I merged down #1666 to master, but then it doesn't matter since documentation is generated for master only.

alcaeus:
Thanks @jdeniau!

PR #1676, opened 2017-11-27
Refactoring tests

gabriel-caruso:
I've refactored some tests with [`assertFileExists`](https://phpunit.de/manual/current/pt_br/appendixes.assertions.html#appendixes.assertions.assertFileExists), [`assertNull`](https://phpunit.de/manual/current/pt_br/appendixes.assertions.html#appendixes.assertions.assertNull), [`assertFalse`](https://phpunit.de/manual/current/pt_br/appendixes.assertions.html#appendixes.assertions.assertFalse), [`assertTrue`](https://phpunit.de/manual/current/pt_br/appendixes.assertions.html#appendixes.assertions.assertTrue) and [`assertEmpty`](https://phpunit.de/manual/current/pt_br/appendixes.assertions.html#appendixes.assertions.assertEmpty) methods.

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH816Test.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH816Test.php
@@ -10,22 +10,22 @@ class GH816Test extends \Doctrine\ODM\MongoDB\Tests\BaseTest
     {
         $d=new GH816Document();
         $d->_id=new \MongoId();
-        $this->assertSame(array(), $this->dm->getRepository('Doctrine\ODM\MongoDB\Tests\GH816Document')->findAll());
+        $this->assertEmpty($this->dm->getRepository('Doctrine\ODM\MongoDB\Tests\GH816Document')->findAll());
         $this->dm->persist($d);
         $this->dm->detach($d);
         $this->dm->flush();
-        $this->assertSame(array(), $this->dm->getRepository('Doctrine\ODM\MongoDB\Tests\GH816Document')->findAll());
+        $this->assertEmpty($this->dm->getRepository('Doctrine\ODM\MongoDB\Tests\GH816Document')->findAll());
     }
-    
+
     public function testPersistAfterDetachWithTitleSet()
     {
         $d=new GH816Document();
         $d->title="Test";
-        $this->assertSame(array(), $this->dm->getRepository('Doctrine\ODM\MongoDB\Tests\GH816Document')->findAll());
+        $this->assertEmpty($this->dm->getRepository('Doctrine\ODM\MongoDB\Tests\GH816Document')->findAll());
         $this->dm->persist($d);
         $this->dm->detach($d);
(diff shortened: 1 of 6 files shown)


tests/Doctrine/ODM/MongoDB/Tests/Mapping/Driver/XmlDriverTest.php
```diff
@@ -43,16 +43,16 @@ public function testDriverShouldParseNonStringAttributes()
         $classMetadata = new ClassMetadata('TestDocuments\UserNonStringOptions');
         $this->driver->loadMetadataForClass('TestDocuments\UserNonStringOptions', $classMetadata);
 
-        $this->assertSame(true, $classMetadata->requireIndexes);
-        $this->assertSame(false, $classMetadata->slaveOkay);
+        $this->assertTrue($classMetadata->requireIndexes);
```
malarzm: Does `assertTrue` check that result is in fact `bool`?

gabriel-caruso: Yeap, check [this PHPUnit method](https://github.com/sebastianbergmann/phpunit/blob/master/src/Framework/Constraint/IsTrue.php).

malarzm: Cool, thanks for the confirmation!

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH816Test.php
```diff
@@ -10,22 +10,22 @@ public function testPersistAfterDetachWithIdSet()
     {
         $d=new GH816Document();
         $d->_id=new \MongoId();
-        $this->assertSame(array(), $this->dm->getRepository('Doctrine\ODM\MongoDB\Tests\GH816Document')->findAll());
+        $this->assertEmpty($this->dm->getRepository('Doctrine\ODM\MongoDB\Tests\GH816Document')->findAll());
```
alcaeus: In this case we should be ok, just mentioning that `assertSame([], ...)` does not necessarily `assertEmpty(...)`: the latter doesn't include the implicit type check for an array.

gabriel-caruso: @alcaeus The [method from PHPUnit](https://github.com/sebastianbergmann/phpunit/blob/master/src/Framework/Constraint/IsEmpty.php) by checking if is an instance of `Countable` isn't check for an array?

alcaeus: No, it checks for anything that implements `\Countable`, including the ODM cursor object. If you want to specifically check if an array is returned, either use `assertSame([], ...)` or `assertEmpty(...)` in combination with `assertInternalType('array', ...)`.

PR #1679, opened 2017-12-03
Fixed typo in map-reduce doc

pensiero:
There is a typo, if we want to map-reduce the event document based, we need to `createQueryBuilder` on top of the document `Documents\Event`


alcaeus:
Thanks @pensiero!

pensiero:
More then welcome ;)

PR #1681, opened 2017-12-04
Fix/map many embedded documents and storage strategy

healerz:


malarzm:
@healerz By all means you are right that such usage will produce a notice but most probably absence of `strategy` will cause a lot of bugs later down the line that can be harder to catch. I'm inclined to say, that no `strategy` specified for anything that has `type: many` should warrant an exception.

Out of curiosity, what led you to finding this? Frankly I'd myself be horrified if I'd need to work directly with configuring field mappings :)

healerz:
@malarzm If it should result in a exception or not is more of a philosophical question ;) But i do not agree, as far as i understand the code, it will not result in bugs later down. Due to the method applyStorageStrategy (called here: https://github.com/doctrine/mongodb-odm/blob/957f4199c417aec34e489d297563afa34bd5b174/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataInfo.php#L1491) which will apply the default strategy for embedded many documents (here: https://github.com/doctrine/mongodb-odm/blob/957f4199c417aec34e489d297563afa34bd5b174/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataInfo.php#L1521 / https://github.com/doctrine/mongodb-odm/blob/957f4199c417aec34e489d297563afa34bd5b174/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataInfo.php#L1538)
I suggest either implementing an exception for this case (forcing users to define the strategy for embedded many documents) or applying the suggested fix.

And for your curiosity (long story short): me and my team members dislike yaml/xml/annotation configurations due to the fact, that non php configs are not really unit testable. We are using https://github.com/saxulum/saxulum-doctrine-mongodb-odm-provider as a service provider for our slim (pimple) application, which provides the possibility to define the mappings in a php mapping file. At the end of the day the usage is similar to yaml/xml/... .

malarzm:
All right, I forgot we were setting a default strategy in case it was absent, I was pretty sure it was coming from the drivers only.

healerz:
👍

malarzm:
Thanks @healerz!

PR #1682, opened 2017-12-04
Don't use ignore-platform-reqs when installing mongo-php-adapter

alcaeus:
Using `ignore-platform-reqs` can cause a bunch of errors down the line (e.g. by installing incompatible package versions not suited for the current PHP version). Thus, `ext-mongodb` is provided via `config.platform`.


alcaeus:
> Just confirming that master branch is still relying on the adapter library.

Builds would start failing if the adapter was no longer installed, as all `\Mongo*` classes would not be found on PHP 7.

PR #1684, opened 2017-12-05
Refactoring tests

carusogabriel:
Continuing to refactory tests, like #1676, I used:
- `assertCount` instead of `count` function;
- `assertInstanceOf` instead of `instanceof` operator;
- `assertArrayHasKey` instead of `array_key_exists` function;
- `assertInternalType` instead of `is_*` functions;
- `assertFalse` instead of strict comparisons with `false` keyword;
- `assertGreaterThan` for mathematical comparisons;
- `assertSame` and `assertNotSame` instead of strict comparisons;
- `assertContains` instead of `in_array` function;
- `assertEmpty` and `assertNotEmpty` instead of `empty` function.

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/CollectionPersisterTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/CollectionPersisterTest.php
@@ -16,7 +16,7 @@ class CollectionPersisterTest extends BaseTest
         $persister->delete($user->phonenumbers, array());
 
         $user = $this->dm->getDocumentCollection(__NAMESPACE__ . '\CollectionPersisterUser')->findOne(array('username' => 'jwage'));
-        $this->assertFalse(isset($user['phonenumbers']), 'Test that the phonenumbers field was deleted');
+        $this->assertArrayNotHasKey('phonenumbers', $user, 'Test that the phonenumbers field was deleted');
     }
 
     public function testDeleteEmbedMany()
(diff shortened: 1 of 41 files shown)


tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/GraphLookupTest.php
```diff
@@ -90,7 +90,7 @@ public function testGraphLookupWithEmployees(\Closure $addGraphLookupStage, arra
                 continue;
             }
 
-            $this->assertGreaterThan(0, count($reportingHierarchy->reportingHierarchy));
+            $this->assertGreaterThan(0, $reportingHierarchy->reportingHierarchy);
```
alcaeus: I believe this change is incorrect: greaterThan compares expected and actual directly and doesn't account for countables.

carusogabriel: You are correct, assertGreaterThan just [check the mathematical condition](https://github.com/sebastianbergmann/phpunit/blob/master/src/Framework/Constraint/GreaterThan.php#L43), it doesn't count.

carusogabriel:
I've made the requested changes, and, use `assertStringEqualsFile` when comparing a `string` and a file.

carusogabriel:
I've refactored more tests using `assertArrayHasKey` and `assertArrayNotHasKey` instead of checking the key with isset function.

alcaeus:
Thanks @carusogabriel!

PR #1685, opened 2017-12-05
Remove default value from type attribute

MidnightDesign:
The XSD defines a default value (`string`) for the `type` attribute of the `field` tag. PhpStorm (or an extension?) even issues a warning ("Redundant default attribute value assignment") when I set the attribute to `string`.

So either the code has to be changed to reflect the XML definition, or the definition needs to be changed - which this PR does.

### DIFF
--- a/doctrine-mongo-mapping.xsd
+++ b/doctrine-mongo-mapping.xsd
@@ -79,7 +79,7 @@
     </xs:sequence>
     <xs:attribute name="id" type="xs:boolean" default="false" />
     <xs:attribute name="name" type="xs:NMTOKEN" />
-    <xs:attribute name="type" type="xs:NMTOKEN" default="string" />
+    <xs:attribute name="type" type="xs:NMTOKEN" />
     <xs:attribute name="strategy" type="xs:NMTOKEN" default="set" />
     <xs:attribute name="fieldName" type="xs:NMTOKEN" />
     <xs:attribute name="embed" type="xs:boolean" />


alcaeus:
I'll have to check whether this makes sense: if there is a default type in the XmlDriver or the mapping itself, it should be reflected in the schema (as it is now).

You can help this by writing a test case that loads an XML mapping containing a field without a type specified: the field shouldn't have a type of `string`. If it's handled as a string, the default needs to remain in the schema.

MidnightDesign:
I just realized that I left out the most important part from my original post: If I do omit the `type` attribute, I get an ["Undefined index: type" error](https://github.com/doctrine/mongodb-odm/blob/845f07a3b69215eca5d267719c9bc81c4913857d/lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php#L252). As I said: maybe the code should be fixed instead of the XML definition.

Also, IIRC the ORM has the same problem.

malarzm:
FWIW we're having `string` as the default in [annotations](https://github.com/doctrine/mongodb-odm/blob/1cd7c2434384abaacf38e1ca94d4b1ba13a2547d/lib/Doctrine/ODM/MongoDB/Mapping/Annotations/AbstractField.php#L27)

alcaeus:
@MidnightDesign in that case, you already answered what I didn't have time to investigate yet. If the code indeed requires you to provide a value for the `type` attribute, it shouldn't have a default value in the schema. If the same applies to ORM, go ahead and create a pull request there as well.

Thanks!

PR #1687, opened 2017-12-07
Enable shard key to be a referenced object id

notrix:
From mongodb perspective it's all good to shard on refence id. So doctrine should know how to generate query for such sharded collections.


darceee2:
:+1:

alcaeus:
Thanks @notrix!

PR #1688, opened 2017-12-07
Remove usage of pushAll operator in CollectionPersister

alcaeus:
MongoDB 3.6 removes the `$pushAll` operator, which has been deprecated since MongoDB 2.4. This pull request replaces `$pushAll` for `$push` with `$each`, as [recommended by MongoDB](https://docs.mongodb.com/manual/release-notes/3.6-compatibility/#pushall-compatibility).

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/CollectionPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/CollectionPersister.php
@@ -200,21 +200,29 @@ class CollectionPersister
         }
 
         $mapping = $coll->getMapping();
+
+        switch ($mapping['strategy']) {
+            case ClassMetadataInfo::STORAGE_STRATEGY_PUSH_ALL:
+                $operator = 'push';
+                break;
+
+            case ClassMetadataInfo::STORAGE_STRATEGY_ADD_TO_SET:
+                $operator = 'addToSet';
+                break;
+
+            default:
+                throw new \LogicException("Invalid strategy {$mapping['strategy']} given for insertElements");
+        }
+
         list($propertyPath, $parent) = $this->getPathAndParent($coll);
 
-        $pb = $this->pb;
-
         $callback = isset($mapping['embedded'])
-            ? function($v) use ($pb, $mapping) { return $pb->prepareEmbeddedDocumentValue($mapping, $v); }
-            : function($v) use ($pb, $mapping) { return $pb->prepareReferencedDocumentValue($mapping, $v); };
+            ? function($v) use ($mapping) { return $this->pb->prepareEmbeddedDocumentValue($mapping, $v); }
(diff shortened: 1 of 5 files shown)


lib/Doctrine/ODM/MongoDB/Persisters/CollectionPersister.php
```diff
@@ -209,12 +209,9 @@ private function insertElements(PersistentCollectionInterface $coll, array $opti
...
 
         $value = array_values(array_map($callback, $insertDiff));
+        $operator = $mapping['strategy'] === ClassMetadataInfo::STORAGE_STRATEGY_PUSH_ALL ? 'push' : $mapping['strategy'];
 
-        if ($mapping['strategy'] === ClassMetadataInfo::STORAGE_STRATEGY_ADD_TO_SET) {
-            $value = array('$each' => $value);
-        }
-
-        $query = array('$' . $mapping['strategy'] => array($propertyPath => $value));
+        $query = array('$' . $operator => array($propertyPath => array('$each' => $value)));
```
jmikola: Previously, `$each` wrapping was only done for `addToSet`. With this change, we're doing it for all strategies. Are you sure this will work for all [Storage Strategies](http://docs.doctrine-project.org/projects/doctrine-mongodb-odm/en/latest/reference/storage-strategies.html) that reach this code path? Given that `set` and its variants are not two-step operations and presumably handled elsewhere, is this code path only reached for `addToSet` and `pushAll`? If so, I think we're OK, although it wouldn't hurt to have some LogicException thrown for unexpected strategies here (to preempt a server error).

I might also suggest update the docs in this PR to say that `pushAll` uses `$push` and `$each`.

I assume this is going into a patch release, but if we have another 1.x minor release you may want to deprecate `pushAll` (just emit E_DEPRECATED) and advertise a new `pushEach` strategy. In 2.0, I think we'd be better off documenting a `pushEach` strategy instead of `pushAll` (maybe silently parse `pushAll` to ease BC woes). If there are no more minor releases in 1.x, we can also use 2.0 to introduce E_DEPRECATED and then drop `pushAll` entirely in 3.0.

Side note: Can we use short array syntax here?

alcaeus: The method in question is called for the `pushAll` and `addToSet` strategies only, but you're right: throwing an exception if it's called for another strategy should be done.

> I might also suggest update the docs in this PR

Good point, thanks!

Regarding the deprecation, I'd suggest the following:
1. Release this fix as 1.2.1, no deprecations
2. For 2.0, rename `pushAll` to `push`
3. As with other features that are being removed/renamed in 2.0, add it to the list of deprecations in 1.3 that will be released around the same time as the first 2.0 beta, thus locking in the API.

> Can we use short array syntax here?

Absolutely. Can't wait to do that in odm-ng and remove every single occurence of `array(` (which doesn't make sense until shortly before a BC-breaking merge due to open pull requests getting code conflicts).

PR #1693, opened 2017-12-19
Schema manager should handle shard key on reference

notrix:
Schema manager should also know how to handle shard key on reference
This PR is part of sharding key on reference fix. Referencing another PR #1687

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/SchemaManager.php
+++ b/lib/Doctrine/ODM/MongoDB/SchemaManager.php
@@ -593,11 +593,27 @@ class SchemaManager
         $shardKey = $class->getShardKey();
         $adminDb  = $this->dm->getClient()->selectDatabase('admin');
 
+        $shardKeyPart = [];
+        foreach ($shardKey['keys'] as $key => $order) {
+            if ($class->hasField($key)) {
+                $mapping   = $class->getFieldMapping($key);
+                $fieldName = $mapping['name'];
+
+                if ($class->isSingleValuedReference($key)) {
+                    $fieldName = ClassMetadata::getReferenceFieldName($mapping['storeAs'], $fieldName);
+                }
+            } else {
+                $fieldName = $key;
+            }
+
+            $shardKeyPart[$fieldName] = $order;
+        }
+
         return $adminDb->command(
             array_merge(
                 [
                     'shardCollection' => $dbName . '.' . $class->getCollection(),
-                    'key'             => $shardKey['keys'],
+                    'key'             => $shardKeyPart,
                 ],
                 $this->getWriteOptions(null, $writeConcern)
             )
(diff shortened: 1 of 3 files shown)


darceee2:
:+1:

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -609,10 +610,32 @@ private function runShardCollectionCommand($documentName)
         $shardKey = $class->getShardKey();
         $adminDb = $this->dm->getConnection()->selectDatabase('admin');
 
+        $shardKeyPart = array();
+        foreach ($shardKey['keys'] as $key => $order) {
+            try {
+                $mapping = $class->getFieldMappingByDbFieldName($key);
```
alcaeus: I'll have to investigate field name translation here - there may be a couple of things wrong with field name translations in this method, unrelated to your patch.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -609,10 +610,32 @@ private function runShardCollectionCommand($documentName)
...
+                if (isset($mapping['association']) && $mapping['association'] === ClassMetadata::REFERENCE_ONE) {
+                    switch ($mapping['storeAs']) {
```
alcaeus: Use `ClassMetadataInfo::getReferenceId` instead of duplicating logic.

notrix: I could not find how to use ClassMetadataInfo::getReferenceId() as I don't have reference mongo data. But I have found ClassMetadataInfo::getReferenceFieldName() useful in this case.

alcaeus: Yes, my apologies - got confused. `getReferenceFieldName` is correct.

alcaeus:
@notrix after an eternity without updates, I finally managed to revisit this. I've created #1748 to fix this issue in the 1.2 branch, to be released in 1.2.2 soon.

I've rebased this branch on top of the current master to spare you the trouble - with the `ext-mongodb` and CS merges lots of things have changed. Due to the `ext-mongodb` change, sharding support in master is currently incomplete, so I've skipped the test for now. I'll revisit this in #1698 to work around the issues preventing finalizing this. Once #1748 has been merged, I'll also merge this branch down.

Thanks for your contribution! 🎉

tests/Documents/Sharded/ShardedByUser.php
```diff
@@ -0,0 +1,20 @@
...
+
+declare(strict_types=1);
+
+namespace Documents\Sharded;
+
+use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
+
+/**
+ * @ODM\Document(collection="sharded.users")
+ * @ODM\ShardKey(keys={"user"="asc"})
```
alcaeus: FYI @notrix, I had to change this from `desc` to `asc` as the test failed with the error that the shard key would could only be ascending or hashed.

PR #1694, opened 2017-12-20
Add a link to chat room on gitter

malarzm:
Also I've removed IRC as we're abandoning the ship and moved SO up so it looks more "preferred"

PR #1695, opened 2017-12-20
Fix link name for Stack Overflow

jmikola:
See: https://github.com/doctrine/mongodb-odm/pull/1694#discussion_r158069270


malarzm:
Sure thing is!

PR #1709, opened 2017-12-25
Use build stages for travis-ci

alcaeus:
This makes it easier to add code quality checks later on. Additionaly, xdebug is disabled for all builds but one that runs with coverage; that should hopefully make the test suite faster again.

Also, I've noticed that we still support PHP 7.0, so that's definitely going out the door. Might even drop 7.1 later on.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ language: php
 
 # Note: latest PHP version is tested with coverage; lowest is tested with --prefer-lowest
 php:
-  - 7.0
   - 7.1
   - 7.2
 
@@ -38,7 +37,7 @@ jobs:
 
     # Test against lowest dependencies, including driver and server versions
     - stage: Test
-      php: 7.0
+      php: 7.1
       env: SERVER_VERSION="3.2" KEY_ID="EA312927" DRIVER_VERSION="1.3.0"
       install:
         - composer update --prefer-lowest
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ dist: trusty
 sudo: required
 language: php
 
+# Note: latest PHP version is tested with coverage; lowest is tested with --prefer-lowest
 php:
   - 7.0
   - 7.1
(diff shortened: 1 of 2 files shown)


alcaeus:
FWIW, here's the test suite on PHP 7.2 with coverage enabled:
> Time: 5.41 minutes, Memory: 308.00MB

Here with coverage disabled:
> Time: 25.98 seconds, Memory: 244.00MB

With that in mind, I'd go ahead and expand the build matrix to test all versions against MongoDB 3.2, 3.4 and 3.6.

.travis.yml
```diff
@@ -15,17 +15,36 @@ env:
...
+
+    # Run test against MongoDB 3.4
+    - stage: Test
+      php: 7.2
+      env: SERVER_VERSION="3.4" KEY_ID="0C49F3730359A14518585931BC711F9BA15703C6"
+
+    # Test against lowest dependencies, including driver and server versions
+    - stage: Test
+      php: 7.0
+      env: COMPOSER_FLAGS="--prefer-lowest" SERVER_VERSION="3.2" KEY_ID="EA312927" DRIVER_VERSION="1.3.0"
```
Majkl578: These flags look a bit hacky here now, why not override specific target (preferably `install:`) itself? See DBAL's config for inspiration.

.travis.yml
```diff
@@ -2,10 +2,10 @@ dist: trusty
 sudo: required
 language: php
 
+# Note: latest PHP version is tested with coverage; lowest is tested with --prefer-lowest
 php:
   - 7.0
   - 7.1
-  - 7.2
```
Majkl578: It's somewhat confusing to not include all PHP versions here. Also it won't test full matrix of combinations (which is probably desired?), thus may theoretically not catch all errors on different versions.

.travis.yml
alcaeus: For one, I was hoping to avoid duplicating the long list of `install` steps. Apart from that, I'm not sure how I could simplify it without doing things twice (e.g. installing all dependencies in `install` and updating to lowest dependencies in `before_script`)

Majkl578: `before_install` and `install` are not overriden so you can probably merge those two together, also with `composer self-update`. That would free-up `install` target for `composer install/update` and its overriding.

alcaeus: I'll take a look, thanks for the suggestion 👍

PR #1713, opened 2018-01-03
Use array_filter and array_map

carusogabriel:
They can help us save some lines of code :put_litter_in_its_place:


alcaeus:
Thanks @carusogabriel!

PR #1718, opened 2018-01-07
[2.0] Remove database creation from SchemaManager

malarzm:


lib/Doctrine/ODM/MongoDB/Tools/Console/Command/Schema/CreateCommand.php
```diff
@@ -20,7 +20,6 @@ protected function configure()
             ->setName('odm:schema:create')
             ->addOption('class', 'c', InputOption::VALUE_REQUIRED, 'Document class to process (default: all classes)')
             ->addOption('timeout', 't', InputOption::VALUE_OPTIONAL, 'Timeout (ms) for acknowledged index creation')
-            ->addOption(self::DB, null, InputOption::VALUE_NONE, 'Create databases')
```
alcaeus: Can't comment on the actual line, but we can change `$this->createOrder` in L13 and drop `self::DB` there as well.

malarzm: Sure :+1:

Issue #1702, opened 2017-12-21
[2.0] Code cleanup


alcaeus:
Waiting on https://github.com/doctrine/coding-standard/pull/9.

* Check phpcs against doctrine/coding-standard
* Clean up usage of class name strings, replacing them with `::class`
* Use short array syntax consistently (and add it to doctrine/coding-standard if not already there)

Majkl578:
> and add it to doctrine/coding-standard if not already there)

Good point, it's not. I didn't include it (and similar) in my PR to not make it too heavy to review.

Check out [our company's ruleset.xml](https://github.com/cdn77/coding-standard/blob/master/Cdn77CodingStandard/ruleset.xml) for additional useful rules, we should compile another PR to CS repo with them? 👍

alcaeus:
Yep, would make sense to add it in a second step. I can do so when we've taken care of the current PR.

carusogabriel:
@alcaeus Let's start this with `tests`?

alcaeus:
@carusogabriel I've got a WIP branch with the automatic phpcs fixes in `src/` and `tests/` on my disk, I'll prepare it for a pull request tomorrow. I've decided to temporarily disable sniffs that require manual fixing so we can then fix those in pull requests that are smaller to review.

alcaeus:
@carusogabriel  The first batch of changes has been merged in #1743. It incorporates all automatic fixes from `doctrine/coding-standard` v4 and adds an exclude list of sniffs that need to be fixed manually. The idea is to keep the diffs small and reviewable by fixing them one by one.

The list of sniffs can be found in [phpcs.xml.dist](https://github.com/alcaeus/mongodb-odm/blob/fed7512c163f65be35b08e00950849fad6842ff8/phpcs.xml.dist#L40..L63). Some of these sniffs cause lots of errors that need to be manually fixed.

carusogabriel:
@alcaeus Sure. I'll be starting to fix them separately. But, of course, with we can have a diff with 500 maximum, including 3 rules, is also valid. I'll focus on those simpler, and let the more complicated, and with more occurrences to the end.

alcaeus:
I agree, for those sniffs that only produce a couple of violations, feel free to group them together. We just don’t want 10k+ line diffs ;)

malarzm:
I think we're good for now, I'd rather not rename too much stuff for 2.0 ("SuperfluousAbstractClassNaming" et al)

[...]

PR #1719, opened 2018-01-07
[2.0] Remove repopsitories' magic findBy and findOneBy

malarzm:


lib/Doctrine/ODM/MongoDB/DocumentRepository.php
```diff
@@ -179,50 +179,6 @@ public function findOneBy(array $criteria)
...
-            throw new \BadMethodCallException(
-                "Undefined method: '$method'. The method name must start with 'findBy' or 'findOneBy'!"
-            );
-        }
-
-        if (!isset($arguments[0])) {
-            throw MongoDBException::findByRequiresParameter($method . $by);
-        }
-
-        $fieldName = Inflector::camelize($by);
```
alcaeus: This also drops our dependency on `doctrine/inflector`, so we can remove it from composer.json.

malarzm:
> I'd suggest using short array syntax consistently here, but I can also take care of it when doing #1702.

I was not changing them on purpose as we'll need to do it project-wide anyway :)

lib/Doctrine/ODM/MongoDB/DocumentRepository.php
malarzm: Good catch, dropped :+1:

PR #1721, opened 2018-01-07
[2.0] Remove DocumentManager::createDbRef

malarzm:
FWIW there are 2 usages of the removed method but they're both in code paths removed by #1720


malarzm:
Stripping `needs-1.3-compat-layer` as method was already deprecated and PR was a clean up - nothing more we can do.

PR #1722, opened 2018-01-07
[2.0] Disallow nested commits

malarzm:


lib/Doctrine/ODM/MongoDB/MongoDBException.php
```diff
@@ -159,4 +159,9 @@ public static function failedToEnsureDocumentSharding($className, $errorMessage)
             $errorMessage
         ));
     }
+
+    public static function commitInProgress()
+    {
+        return new self('There is already a commit operation in progress');
```
alcaeus: Maybe add an explanation, like `did you call flush` from an event listener?

malarzm: Asking about event listener is indeed better than just saying that flush from there is forbidden, adding it :+1:

PR #1725, opened 2018-01-09
Use $strict param in functions that have it

carusogabriel:
Some functions, e.g. `in_array`, have a third param to enable strict comparison :shield:

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Sort.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Sort.php
@@ -33,7 +33,7 @@ class Sort extends Stage
 
         foreach ($fields as $fieldName => $order) {
             if (is_string($order)) {
-                if (in_array($order, $allowedMetaSort)) {
+                if (in_array($order, $allowedMetaSort, true)) {
                     $order = ['$meta' => $order];
                 } else {
                     $order = strtolower($order) === 'asc' ? 1 : -1;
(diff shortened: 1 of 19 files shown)


malarzm:
Does it change anything (to be better) if we're not checking objects?

alcaeus:
> Does it change anything (to be better) if we're not checking objects?

It forces type-safety:
```
php > var_dump(in_array(0, ['0']));
php shell code:1:
bool(true)
php > var_dump(in_array(0, ['0'], true));
php shell code:1:
bool(false)
```

Useful indeed 👍

carusogabriel:
Eh, I just made this changes to be more strict, like `===` on top of `==` :grimacing:

malarzm:
I'm wondering whether it won't bite someone (read whether it's not a slight BC break), for instance now check whether index `['foo' => '1']` will return `false` as the existing one is `['foo' => 1]`.

carusogabriel:
@malarzm Oh, now I get your point. As you know the library better them me, there are some changes that should be rollback?

alcaeus:
@malarzm yes, BC breaks may be bad, but then this is for 2.0 so we can just go ahead and do it :)

malarzm:
Ok, went through the changes and nothing seems to be user facing. But on the other hand there seem to be no real benefit in adding the third parameter, we're mostly checking whether some needle is present in a defined list of strings.

jmikola:
I'd be cautious about adding this for any inspections of BSON data, as values with different types can be equivalent in some contexts (e.g. index specifications with `{ x: 1 }` and `{ x: true }`). No objections if it's just being used to add some strict checking for internal data, though.

PR #1714, opened 2018-01-06
Add PHPStan

carusogabriel:
This starts #1701 by adding `PHPStan` :man_technologist:

PR #1726, opened 2018-01-09
Use proper variadic arguments

alcaeus:
ping @carusogabriel

Helps with #1702 and #1714.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
@@ -75,10 +75,10 @@ class Expr
      * @see http://docs.mongodb.org/manual/reference/operator/aggregation/add/
      * @param mixed|self $expression1
      * @param mixed|self $expression2
-     * @param mixed|self $expression3,... Additional expressions
+     * @param mixed|self ...$expressions Additional expressions
      * @return $this
      */
-    public function add($expression1, $expression2 /* , $expression3, ... */)
+    public function add($expression1, $expression2, ...$expressions)
     {
         return $this->operator('$add', func_get_args());
     }
@@ -88,9 +88,10 @@ class Expr
      *
      * @see http://docs.mongodb.org/manual/reference/operator/aggregation/and/
      * @param array|self $expression
+     * @param array|self ...$expressions
      * @return $this
      */
-    public function addAnd($expression /*, $expression2, ... */)
+    public function addAnd($expression, ...$expressions)
     {
         if (! isset($this->expr['$and'])) {
             $this->expr['$and'] = [];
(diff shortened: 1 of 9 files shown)


lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
```diff
@@ -88,9 +88,10 @@ public function add($expression1, $expression2 /* , $expression3, ... */)
      *
      * @see http://docs.mongodb.org/manual/reference/operator/aggregation/and/
      * @param array|self $expression
+     * @param array|self ...$expressions
      * @return $this
      */
-    public function addAnd($expression /*, $expression2, ... */)
+    public function addAnd($expression, ...$expressions)
```
Majkl578: Shouldn't this simply become `addAnd(...$expressions)`?

alcaeus: That would allow a call like `$expr->addAnd()`. This way requires at least one argument to be present.

Majkl578: Good point, but would eliminate the need for `func_get_args()`. `assert()` could help, but it's a hack. Hmmm... :/

alcaeus: Yeah, it's difficult. TBH, I prefer the method signature to be clear and concise about the number of arguments. Using `addAnd(...$expressions)` needs documentation about the required parameter and is also inconsistent with other method calls that have too few parameters.

malarzm: The method could be NOP for `$expr->addAnd()`. Or throw `\InvalidArgumentException` :)

malarzm: But yeah, then there are functions requiring three arguments so I think indicating minimum is not bad idea after all.

carusogabriel:
This is help us with PHPStan 🙏🏻

lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
alcaeus: > Or throw `\InvalidArgumentException`

That's exactly what I want to avoid - I want the method signature to signify that you have to add at least one condition - the only way to achieve this ATM is by doing `$param, ...$params` and use `func_get_args` (or use `array_merge([$param], $params)` which is even worse).

For methods like `geoWithinPolygon` that requires 3 or more arguments it makes a lot of sense to make this requirement explicit instead of hiding it in documentation or behind an exception.

alcaeus:
Just a quick note that these are just the methods that were painfully obvious. There are still a few methods with mixed signatures (e.g. `sort('field', 'asc')` vs. `sort(['field' => 'asc'])` that may be cleaned up later).

PR #1730, opened 2018-01-11
Converter to XML for mapping

malarzm:
As we're going to kill YAML driver for 2.0 let's provide users with a command to convert their mappings to XML. I suggest to make this feature exclusive for 1.3 as I'm not really willing to maintain this :) Left to do:

- [x] finish index's partial filter expressions - needs to be migrated manually
- [x] go through XSD AND XmlDriver once again to see if everything is covered - I believe it is
- [x] actually save XMLs somewhere
- [x] so far this is a brain dump, need to test the converter
- [x] think about some unit tests - thought about it: naaah
- [x] ask community to test the tool - https://twitter.com/malarzm/status/1057233251067150336
- [x] add deprecation notice
- [x] target `1.3.x` branch once created

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Tools/Console/Command/ConvertMappingToXmlCommand.php
+++ b/lib/Doctrine/ODM/MongoDB/Tools/Console/Command/ConvertMappingToXmlCommand.php
@@ -1,0 +1,388 @@
+<?php
+
+namespace Doctrine\ODM\MongoDB\Tools\Console\Command;
+
+use Doctrine\Common\Util\Inflector;
+use Doctrine\ODM\MongoDB\DocumentManager;
+use Doctrine\ODM\MongoDB\Mapping\ClassMetadataInfo;
+use Doctrine\ODM\MongoDB\Tools\Console\MetadataFilter;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * @deprecated class was deprecated in 1.3 and will be removed in 2.0.
+ */
+class ConvertMappingToXmlCommand extends Command
+{
+    protected function configure()
+    {
+        $this
+            ->setName('odm:mapping:dump-xml')
+            ->setDescription('Converts mappings to the XML format. Available in 1.3 only.')
+            ->addArgument('dest-path', InputArgument::REQUIRED, 'The path to generate your XML mappings.')
+            ->addOption('filter', null, InputOption::VALUE_REQUIRED | InputOption::VALUE_IS_ARRAY, 'A string pattern used to match documents that should be processed.')
+            ->addOption('ext', null, InputOption::VALUE_OPTIONAL, 'Extension for created files.', 'dcm.xml')
+        ;
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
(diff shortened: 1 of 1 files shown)


malarzm:
After some polishing I managed to convert [almost full-blown YAML file](https://github.com/doctrine/mongodb-odm/blob/master/tests/Doctrine/ODM/MongoDB/Tests/Mapping/yaml/Doctrine.ODM.MongoDB.Tests.Mapping.AbstractMappingDriverUser.dcm.yml). First organoleptic tests are promising :) On Monday I'll try to setup something to compare actual metadata read from YAML and XML formats and see if there are any differences.

alcaeus:
@malarzm what's the status here? Should we invite some of our users to test their tool and report their difference? I know there are a few cases that aren't covered (e.g. `partialFilterExpressions` in indexes) but I'm willing to overlook them for now.

We could maybe add some documentation on how to migrate from YAML to XML and ask people to test those steps and report back, what do you think?

malarzm:
I'll see if I'll manage to somehow compare the results today, but you're right. I'll find how I ran this command on my workstation, post instructions here and on Twitter

alcaeus:
Instructions can be added in the docs and reference in the `UPGRADE-2.0` document, WDYT?

malarzm:
Sorry for the lag, I was pulled into some old project at work :/

The command to run this is actually simple and would land in the bundle, but for the time being:

```
<?php

namespace App\Command;

use Doctrine\Bundle\MongoDBBundle\Command\DoctrineODMCommand;
use Doctrine\ODM\MongoDB\Tools\Console\Command\ConvertMappingToXmlCommand;
use Symfony\Component\Console\Input\InputInterface;
use Symfony\Component\Console\Input\InputOption;
use Symfony\Component\Console\Output\OutputInterface;

class DumpXmlCommand extends ConvertMappingToXmlCommand
{
    protected function configure()
    {
        parent::configure();
        $this
            ->addOption('dm', null, InputOption::VALUE_OPTIONAL, 'The document manager to use for this command.')
        ;
    }
    protected function execute(InputInterface $input, OutputInterface $output)
    {
        DoctrineODMCommand::setApplicationDocumentManager($this->getApplication(), $input->getOption('dm'));
        return parent::execute($input, $output);
    }
}
```

malarzm:
@redthor would you mind giving this a spin?

redthor:
It processed 133 of our YAML files without a failure!

Getting the XML files to run in our app will take a bit more work because we have our system broken into a number of packages...

I have received a `Property id does not exist` on a mapping file that uses the mapped-superclass facility. I'll have a look more into what is going on. But it might be worthwhile to continue with the PR in any case because there might be something wrong with my test. And it'll be easy to submit a fix if we find an issue when we actually switch over.

Thanks

alcaeus:
That sounds really good, thanks for your help @redthor!

malarzm:
@alcaeus all is done here

alcaeus:
I'm just now realising this will produce mapping files that cause deprecation notices (due to new `id` and `field-name` mappings). @malarzm do you want to revisit this?

malarzm:
@alcaeus can't really say when I'll have a moment to do so, if you'd have a minute please go ahead and take over here. If you don't I'll fix the issues as soon as I have some time on my hands to do so :)

alcaeus:
Updated `field-name` and added support for `id` mappings. Also fixed a tiny inconsistency regarding `name` vs. `field-name`. Feel free to re-check this against your original mappings, I dumped the ones from the sandbox and they worked just fine.

malarzm:
LGTM, thank you!

PR #1732, opened 2018-01-12
Add missing cascade detach to XSD

malarzm:
Yet another thing missing in XSD :)


malarzm:
Backported in https://github.com/doctrine/mongodb-odm/commit/642646d6e8d7a8a511cb4a632fbc77e21fe9e723

PR #1734, opened 2018-01-13
[2.0] Remove YAML mapping support

### LINKED ISSUE #1706: [RFC] [2.0] Drop YAML mapping driver
Similar to ORM, I'm very inclined to drop the YAML mapping driver for 2.0.

In applications, I'd recommend using annotations to map documents: the mapping is where the code lives, and since the code isn't designed to be extended there's no reason to be able to override the mapping.
For libraries that provide mappings to ODM, it has always been recommended to rely on XML mappings as it allows overriding and is also validated against the schema, which isn't the case for YAML.

Currently, the YAML driver only creates maintenance overhead, which I'd like to avoid: there are several changes coming up that will require work on the drivers (changing metadata objects, views in MongoDB, document validation, etc.) and I'd like to avoid as much of that as possible.

malarzm:
I felt the urge to destroy something :) Closes #1706


malarzm:
Correct, YAML removed. I've also removed entire `suggest` section as your adapter is no longer needed for 2.0 :)

BTW I'd merge this once I sort out #1735

malarzm:
:fire: :gun:

PR #1740, opened 2018-01-18
Bump latest docs to 2.0

malarzm:


docs/en/conf.py
```diff
@@ -45,9 +45,9 @@
 # built documents.
 #
 # The short X.Y version.
-version = '1.2'
+version = 2.0
```
malarzm: Shoot, I've accidentally removed `'`... I'll fix this directly

PR #1742, opened 2018-02-02
Require minimum PHPUnit 7 as we require PHP 7.1

carusogabriel:


malarzm:
@carusogabriel seems like more changes are required

carusogabriel:
@malarzm Yes, most of them are PHP 7 return and scalar types, working on that :construction:

carusogabriel:
Ready to go :+1:

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/AggregationTestTrait.php
```diff
@@ -6,14 +6,15 @@
...
 use Doctrine\ODM\MongoDB\Query\Expr as QueryExpr;
 use Documents\User;
+use PHPUnit\Framework\MockObject\MockBuilder;
 
 trait AggregationTestTrait
 {
     /**
      * @param string $className
-     * @return \PHPUnit_Framework_MockObject_MockBuilder
+     * @return MockBuilder
```
malarzm: This is redundant now I think :)

carusogabriel: Should we drop?

alcaeus: Yep you can drop it - phpcs will catch it in the next PR ;)

PR #1743, opened 2018-02-09
Add phpcs to build and apply automatic fixes

alcaeus:
Part of #1702. To help with reviews, this PR only applies automatic fixes and disables all other sniffs for later fixing.

Note: some tests fail, these will have to be fixed. Most of them are due to strict typing.

Note: if you want to expand all diffs by default (at your own risk), you can copy the following snippet into the dev console while on the "Files changed" tab:
```
document.querySelectorAll('.load-diff-button').forEach(item => item.click())
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,10 +43,16 @@ jobs:
         - composer update --prefer-lowest
 
     - stage: Code Quality
-      php: 7.2
-      env: SERVER_VERSION="3.6" KEY_ID="2930ADAE8CAF5059EE73BB4B58712A2291FA4AD5"
+      php: 7.1
+      env: STATIC_ANALYSIS SERVER_VERSION="3.6" KEY_ID="2930ADAE8CAF5059EE73BB4B58712A2291FA4AD5"
       script:
         - vendor/bin/phpstan analyse -l 1 lib
+
+    - stage: Code Quality
+      php: 7.1
+      env: CODING_STANDARDS SERVER_VERSION="3.6" KEY_ID="2930ADAE8CAF5059EE73BB4B58712A2291FA4AD5"
+      script:
+        - vendor/bin/phpcs
 
 before_install:
   - mv ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/xdebug.ini{,.disabled} || echo "xdebug not available"
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
@@ -1,15 +1,23 @@
 <?php
 
+declare(strict_types=1);
+
 namespace Doctrine\ODM\MongoDB\Aggregation;
 
 use Doctrine\ODM\MongoDB\DocumentManager;
 use Doctrine\ODM\MongoDB\Iterator\CachingIterator;
 use Doctrine\ODM\MongoDB\Iterator\HydratingIterator;
 use Doctrine\ODM\MongoDB\Iterator\Iterator;
+use Doctrine\ODM\MongoDB\Mapping\ClassMetadata;
+use Doctrine\ODM\MongoDB\Persisters\DocumentPersister;
 use Doctrine\ODM\MongoDB\Query\Expr as QueryExpr;
 use GeoJson\Geometry\Point;
 use MongoDB\Collection;
 use MongoDB\Driver\Cursor;
+use function array_map;
+use function array_merge;
+use function array_unshift;
+use function is_array;
 
 /**
  * Fluent interface for building aggregation pipelines.
@@ -26,7 +34,7 @@ class Builder
     /**
      * The ClassMetadata instance.
      *
-     * @var \Doctrine\ODM\MongoDB\Mapping\ClassMetadata
+     * @var ClassMetadata
      */
     private $class;
 
@@ -50,7 +58,6 @@ class Builder
     /**
      * Create a new aggregation builder.
      *
-     * @param DocumentManager $dm
      * @param string $documentName
      */
     public function __construct(DocumentManager $dm, $documentName)
@@ -127,8 +134,6 @@ class Builder
      * the pipeline returns an error.
      *
      * @see http://docs.mongodb.org/manual/reference/operator/aggregation/collStats/
-     * @since 1.5
-     *
      * @return Stage\CollStats
      */
     public function collStats()
@@ -200,7 +205,7 @@ class Builder
      * @see http://docs.mongodb.org/manual/reference/operator/aggregation/geoNear/
      *
      * @param float|array|Point $x
-     * @param float $y
+     * @param float             $y
      * @return Stage\GeoNear
      */
     public function geoNear($x, $y = null)
@@ -221,7 +226,9 @@ class Builder
     public function getPipeline()
     {
         $pipeline = array_map(
-            function (Stage $stage) { return $stage->getExpression(); },
+            function (Stage $stage) {
+                return $stage->getExpression();
+            },
             $this->stages
         );
 
@@ -240,12 +247,12 @@ class Builder
     /**
      * Returns a certain stage from the pipeline
      *
-     * @param integer $index
+     * @param int $index
      * @return Stage
      */
     public function getStage($index)
     {
-        if ( ! isset($this->stages[$index])) {
+        if (! isset($this->stages[$index])) {
             throw new \OutOfRangeException("Could not find stage with index {$index}.");
         }
 
@@ -311,7 +318,7 @@ class Builder
      *
      * @see http://docs.mongodb.org/manual/reference/operator/aggregation/limit/
      *
-     * @param integer $limit
+     * @param int $limit
      * @return Stage\Limit
      */
     public function limit($limit)
@@ -424,7 +431,7 @@ class Builder
      *
      * @see https://docs.mongodb.org/manual/reference/operator/aggregation/sample/
      *
-     * @param integer $size
+     * @param int $size
      * @return Stage\Sample
      */
     public function sample($size)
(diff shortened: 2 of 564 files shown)


lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
```diff
@@ -200,7 +205,7 @@ public function facet()
      * @see http://docs.mongodb.org/manual/reference/operator/aggregation/geoNear/
      *
      * @param float|array|Point $x
-     * @param float $y
+     * @param float             $y
```
malarzm: We have a rule to align params?

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/AddFields.php
```diff
@@ -1,11 +1,12 @@
 <?php
 
+declare(strict_types=1);
+
 namespace Doctrine\ODM\MongoDB\Aggregation\Stage;
 
 /**
  * Fluent interface for adding a $addFields stage to an aggregation pipeline.
  *
- * @author Boris Guéry <guery.b@gmail.com>
```
malarzm: Nitpicking, but empty line above should be removed

lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
alcaeus: Apparently. Personally, I dislike it but I tried to not deviate too much from our official coding standard (except for space before colon because what kind of monster would want that?!?)

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/AddFields.php
alcaeus: @kukulich might know if that's easily added to the sniff in question - if not I'd leave it until the sniff can deal with it.

malarzm: It's not the only place, is there any way to configure CS fixer to disallow empty lines in comments when it's last line?

lib/Doctrine/ODM/MongoDB/Event/PreUpdateEventArgs.php
```diff
@@ -17,11 +20,10 @@ class PreUpdateEventArgs extends LifecycleEventArgs
...
 
     /**
-     * Constructor.
      *
-     * @param object          $document
-     * @param DocumentManager $dm
-     * @param array           $changeSet
+     *
+     * @param object $document
+     * @param array  $changeSet
```
malarzm: This is not treated as redundant with method definition?

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataFactory.php
```diff
@@ -290,10 +294,7 @@ private function addInheritedFields(ClassMetadata $subClass, ClassMetadata $pare
     /**
      * Adds inherited association mappings to the subclass mapping.
      *
-     * @param \Doctrine\ODM\MongoDB\Mapping\ClassMetadata $subClass
-     * @param \Doctrine\ODM\MongoDB\Mapping\ClassMetadata $parentClass
      *
-     * @return void
```
malarzm: This got removed even though there's no `: void` in the definition, is it desirable? I mean will some tool later infer that there's no `return` in the function so `: void` should be added?

malarzm: On the other hand next methods didn't have `@return void` at all so I guess it's not a big deal

lib/Doctrine/ODM/MongoDB/MongoDBException.php
```diff
@@ -67,19 +76,20 @@ public static function cannotPersistMappedSuperclass($className)
...
+     * @param mixed        $got
      * @return MongoDBException
      */
     public static function invalidValueForType($type, $expected, $got)
     {
         if (is_array($expected)) {
-            $expected = sprintf('%s or %s',
+            $expected = sprintf(
+                '%s or %s',
                 join(', ', array_slice($expected, 0, -1)),
```
malarzm: Forbidden functions are not changed automatically or it's not that version of our CS yet?

lib/Doctrine/ODM/MongoDB/Utility/LifecycleEventManager.php
```diff
@@ -73,78 +69,70 @@ public function postCollectionLoad(PersistentCollectionInterface $coll)
...
      * @param object $document
      */
     public function preUpdate(ClassMetadata $class, $document)
     {
-        if ( ! empty($class->lifecycleCallbacks[Events::preUpdate])) {
-            $class->invokeLifecycleCallbacks(Events::preUpdate, $document, array(
-                new PreUpdateEventArgs($document, $this->dm, $this->uow->getDocumentChangeSet($document))
-            ));
+        if (! empty($class->lifecycleCallbacks[Events::preUpdate])) {
+            $class->invokeLifecycleCallbacks(Events::preUpdate, $document, [new PreUpdateEventArgs($document, $this->dm, $this->uow->getDocumentChangeSet($document))]);
```
malarzm: Are we ignoring line's length from now on?

lib/Doctrine/ODM/MongoDB/Event/PreUpdateEventArgs.php
alcaeus: Interesting, seems to be a bug. I'll revisit those anyways when requiring 7.2 and introducing `object` typehint along with scalar typehints.

lib/Doctrine/ODM/MongoDB/MongoDBException.php
alcaeus: It doesn't fix them automatically (it allows free configuration, so function signatures may vary), which is why it's not fixed in this PR.

lib/Doctrine/ODM/MongoDB/Utility/LifecycleEventManager.php
alcaeus: PSR-2 does not include a hard limit. That said, I'm fine changing any lines you mark as "too long, could be split into multiples" 😛

malarzm: No need, it just stood out, I don't mind scrolling sideways that much ;)

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/AddFields.php
kukulich: I'll improve/fix the fixer. I hope it will be done today in next bugfix version.

alcaeus: No rush - we'll have plenty more pull requests until we're done fixing phpcs in this project. Thank you for your help!

carusogabriel: Have experienced that one in [doctrine/doctrine2#7046](https://github.com/doctrine/doctrine2/pull/7046/files#diff-408359faa7db56dc64cf1f0b2006625dL43) as well

lib/Doctrine/ODM/MongoDB/Event/PreUpdateEventArgs.php
kukulich: `object` is not required because of `enableObjectTypeHint` option that is enabled by default only if you run phpcs on PHP 7.2+.

`array` is ok because there is `<exclude name="SlevomatCodingStandard.TypeHints.TypeHintDeclaration.MissingTraversableParameterTypeHintSpecification" />` in phpcs.xml.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/AddFields.php
kukulich: You can try `dev-master` of Slevomat CS.

lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
Majkl578: > except for space before colon because what kind of monster would want that?!?

Me for example. 🐲

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/AddFields.php
Majkl578: Hmm, been fixing these empty _lines_ manually... :| You have it too easy @alcaeus. :P

Majkl578: @kukulich Btw same applies to TypeHintDeclaration sniff when `@param`/`@return` is removed. That really sounds like a candidate for `DocCommentSpacingSniff`? 😊

kukulich: Yes, new sniff is the best solution. It should be easy so maybe in the weekend or next week :)

.travis.yml
```diff
@@ -44,10 +44,16 @@ jobs:
 
     - stage: Code Quality
       php: 7.2
-      env: SERVER_VERSION="3.6" KEY_ID="2930ADAE8CAF5059EE73BB4B58712A2291FA4AD5"
+      env: STATIC_ANALYSIS SERVER_VERSION="3.6" KEY_ID="2930ADAE8CAF5059EE73BB4B58712A2291FA4AD5"
       script:
         - vendor/bin/phpstan analyse -l 1 lib
 
+    - stage: Code Quality
+      php: 7.2
```
Majkl578: This should be 7.1 since 7.1 is required in composer, otherwise you may end up with `object` typehint,

carusogabriel:
@alcaeus I believe 60e293e can be surpassed by Doctrine CS v4 now :)

alcaeus:
Yeah, have to catch up with the times...we're running an antique version of our coding-standard here 😉

carusogabriel:
@alcaeus @malarzm May I update to CS v4?

alcaeus:
Damn, I thought I had that in there as well. Yes, feel free to do so.

PR #1746, opened 2018-02-27
Fix normal/capped collection tests

jmikola:
Previously, the capped collection document was never used. This also revises both tests to assign mapped fields on the documents for consistency.

Just fixing something I noticed while investigating https://github.com/doctrine/mongodb-odm/issues/1745.


alcaeus:
Thanks @jmikola!

PR #1748, opened 2018-03-09
Schema manager should also know how to handle shard key on reference

alcaeus:
This backports #1693 to the 1.2 branch. My initial hunch regarding field name translation was correct - this was an additional issue affecting all shard keys that have a different database name. This is also fixed in this PR.

// cc @notrix

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/SchemaManagerTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/SchemaManagerTest.php
@@ -543,6 +543,33 @@ class SchemaManagerTest extends TestCase
         $this->schemaManager->ensureDocumentSharding($classMetadata->getName());
     }
 
+    public function testEnsureDocumentShardingWithShardByReference()
+    {
+        $dbName = DOCTRINE_MONGODB_DATABASE;
+        $classMetadata = $this->dm->getClassMetadata(\Documents\Sharded\ShardedByUser::class);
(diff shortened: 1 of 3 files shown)


tests/Doctrine/ODM/MongoDB/Tests/SchemaManagerTest.php
```diff
@@ -543,6 +543,33 @@ public function testEnsureDocumentShardingIgnoresAlreadyShardedError()
...
+        $connMock = $this->getMockConnection();
+        $connMock->method('selectDatabase')->with('admin')->willReturn($adminDBMock);
+        $this->dm->connection = $connMock;
+        $this->dm->documentDatabases = array($classMetadata->getName() => $dbMock);
+
+        $adminDBMock
+            ->expects($this->at(0))
+            ->method('command')
+            ->with(array('enableSharding' => $dbName))
+            ->willReturn(array('ok' => 1));
```
malarzm: Why is this mocked instead of sending actual command and inspecting schema later?

alcaeus: Most tests in schema manager build on mocks. These are scheduled for a refactoring in `master` branch since the mocks are no longer valid.

malarzm: All right :+1:

PR #1749, opened 2018-03-09
[CS] Start to fix excluded ones

carusogabriel:
On this one:
- Remove alias functions, as well forbidden ones
- Remove assignments in `if` condition

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
@@ -13,7 +13,6 @@ use Doctrine\ODM\MongoDB\Utility\CollectionHelper;
 use function array_search;
 use function array_values;
 use function get_class;
-use function is_null;
 
 /**
  * PersistenceBuilder builds the queries used by the persisters to update and insert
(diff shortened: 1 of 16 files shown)


lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
```diff
@@ -186,7 +186,7 @@ public function prepareUpdateData($document)
             // @ReferenceOne
             } elseif (isset($mapping['association']) && $mapping['association'] === ClassMetadata::REFERENCE_ONE) {
                 if (isset($new) || $mapping['nullable'] === true) {
-                    $updateData['$set'][$mapping['name']] = (is_null($new) ? null : $this->prepareReferencedDocumentValue($mapping, $new));
+                    $updateData['$set'][$mapping['name']] = ($new === null ? null : $this->prepareReferencedDocumentValue($mapping, $new));
```
carusogabriel: Can this be refactory to `$new ?: $this->prepareReferencedDocumentValue($mapping, $new)`?

malarzm: No, as it would change the logic: `null`s are to be left intact and everything else must go through `prepare...`

lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
```diff
@@ -257,7 +256,7 @@ public function prepareUpsertData($document)
...
```
alcaeus: Parentheses are unnecessary

lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
```diff
@@ -186,7 +185,7 @@ public function prepareUpdateData($document)
...
```
alcaeus: Parentheses are unnecessary.

carusogabriel:
Conflicts solved, should be RTM :+1:

PR #1753, opened 2018-03-14
[CS] Captalize @throws annotation comment and reduce excluded rules

carusogabriel:


Majkl578:
Should also end with a dot, afaik.

carusogabriel:
@Majkl578 Indeed, thanks.

PR #1757, opened 2018-03-15
Stop ignoring ControlStructures.EarlyExit

malarzm:
Automatic fixes

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
@@ -857,7 +857,9 @@ class ClassMetadata implements BaseClassMetadata
                     $lower = strtolower($value);
                     if ($lower === 'asc') {
                         return 1;
-                    } elseif ($lower === 'desc') {
+                    }
+
+                    if ($lower === 'desc') {
                         return -1;
                     }
                 }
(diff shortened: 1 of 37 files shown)


lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -845,7 +845,9 @@ public function addIndex($keys, array $options = [])
                     $lower = strtolower($value);
                     if ($lower === 'asc') {
                         return 1;
-                    } elseif ($lower === 'desc') {
+                    }
+
+                    if ($lower === 'desc') {
```
malarzm: I'm not sure why it was caught (unless `elseif` is a bad practice), there's also one more such change. Should I revert?

alcaeus: That looks like a bug. We should change them back and report it upstream.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -1703,9 +1709,11 @@ public function setParentClasses(array $classNames)
     {
         $this->parentClasses = $classNames;
 
-        if (count($classNames) > 0) {
-            $this->rootDocumentName = array_pop($classNames);
+        if (count($classNames) <= 0) {
```
Majkl578: This can never be <0, but I understand how this happened. :D

malarzm: Better safe than sorry... or something :P

PR #1759, opened 2018-03-15
Stop ignoring TypeHintDeclaration.MissingPropertyTypeHint

malarzm:


lib/Doctrine/ODM/MongoDB/Id/IncrementGenerator.php
```diff
@@ -22,20 +22,34 @@
...
+    /** @var string|null */
     protected $collection = null;
+
+    /** @var string|null */
     protected $key = null;
+
+    /** @var int */
     protected $startingId = 1;
 
+    /**
```
alcaeus: Shouldn't these docblocks be single-lined?

malarzm: This is method's docblock so I believe not

alcaeus: Ah, could be. Well, phpcs will complain eventually if it should be single-lined, so I'm good with it.

PR #1760, opened 2018-03-16
Move coverage to Code Quality stage

malarzm:
This one runs for the longest time and blocks Code Quality while they could go in parallel.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,8 +22,15 @@ env:
 
 jobs:
   include:
+    # Test against lowest dependencies, including driver and server versions
+    - stage: Test
+      php: 7.1
+      env: SERVER_VERSION="3.2" KEY_ID="EA312927" DRIVER_VERSION="1.3.0"
+      install:
+        - composer update --prefer-lowest
+        
     # Run tests with coverage
-    - stage: test
+    - stage: Code Quality
       php: 7.2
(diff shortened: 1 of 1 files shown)


Majkl578:
We've moved coverage to tests because it runs the longest so that it doesn't block whole Travis setup. :thinking:
cc @lcobucci

alcaeus:
@Majkl578 stages are executed sequentially, with one stage only starting once the previous stage has been completed. Since build workers are limited to 5, it makes sense to have long-running jobs at the beginning of a stage so they receive the first build executors available. Apart from that, this yields faster results from the `phpcs` and `phpstan` builds as they will be run in parallel to the coverage build.

@malarzm I'm not sure if it's a good idea, but it might be possible to allow failures for the coverage build (since the test itself has already completed successfully) and turn on `fast_finish`. That would report success and failure states to GitHub way before the coverage build has completed.
In addition to that, we should evaluate whether we even need/use coverage data. As long as we don't use the metric, we don't need to occupy a build executor for 11+ minutes.

malarzm:
@Majkl578 @lcobucci reason I started thinking about this was coverage running last in our setup: https://travis-ci.org/doctrine/mongodb-odm/builds/353865947?utm_source=github_status&utm_medium=notification

As for needing test coverage: I must confess that I'm ignoring emails from Scrutinizer for some time but anyway, it's not bad as a metric (in case somebody would not know how pedantic we can be as an organization :P )

PR #1761, opened 2018-03-19
Fix behavior of null values in upsert queries

### LINKED ISSUE #1671: Upserting a field of an existing Document with null does not update that field
Given a Document with a certain `_id` exists in the database and has `someField` set to a non-null value.

If I create a new Object of that Document class, set its `_id` to that of the existing Document but set `someField` to  `null`, persisting that object does not set `someField` to `null` but leaves it as it was.

Setting `someField` of the new Object to a non-null value will update that Document.

Thus, the update query excludes fields that are set to `null`. I would expect it to be possible to update an existing Document's field with `null` during an upsert operation.

alcaeus:
Fixes #1671.

In upsert queries, `null` values were only properly handled if the field in question was marked as `nullable`. This is incorrect, as we must assume that the document state should be properly upserted to the database, which also means clearing any `null` fields from the document. To achieve this, we also use `$unset` for `null` values, as is done in insert queries.

This commit also hardens the handling of the `$inc` and `$unset` operators in upsert queries to ensure no empty operators are sent to the database. Also, any fields from the criteria of an upsert (normally only the identifier, except for sharded collections) are also excluded from any `$inc` and `$unset` operations, not only from `$set`.

Note: PR will need cherry-picking to master branch. I'll take care of that once this has been merged.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -329,11 +329,15 @@ class DocumentPersister
 
         foreach (array_keys($criteria) as $field) {
             unset($data['$set'][$field]);
+            unset($data['$inc'][$field]);
+            unset($data['$unset'][$field]);
         }
 
-        // Do not send an empty $set modifier
-        if (empty($data['$set'])) {
-            unset($data['$set']);
+        // Do not send an empty $set, $inc or $unset modifiers
+        foreach (['$set', '$unset', '$unset'] as $operator) {
+            if (isset($data[$operator]) && empty($data[$operator])) {
+                unset($data[$operator]);
+            }
         }
 
         /* If there are no modifiers remaining, we're upserting a document with
--- a/lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
@@ -251,9 +251,12 @@ class PersistenceBuilder
                         $operator = '$set';
                         $value = $new === null ? null : Type::getType($mapping['type'])->convertToDatabaseValue($new);
                     }
-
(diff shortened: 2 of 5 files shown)


lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -329,11 +329,15 @@ private function executeUpsert($document, array $options)
...
+            unset($data['$inc'][$field]);
+            unset($data['$unset'][$field]);
         }
 
-        // Do not send an empty $set modifier
-        if (empty($data['$set'])) {
-            unset($data['$set']);
+        // Do not send an empty $set, $inc or $unset modifiers
+        foreach (['$set', '$unset', '$unset'] as $operator) {
+            if (isset($data[$operator]) && empty($data[$operator])) {
```
malarzm: `isset` is superfluous, if `$data[$operator]` is in fact not set, `unset` will do no harm anyway

malarzm:
I'm still not convinced whether this should be fixed in a bugfix release, failing tests are failing exactly due to new `$unset` in queries.

alcaeus:
> failing tests are failing exactly due to new `$unset` in queries.

Yes, because for some reason I forgot to update those. TBH, I consider this a bug - we must assume that the document passed to `persist` is the state they want to be written to the database. If we don't, then we'd also have to change the handling for `nullable` fields, where a `null` value overwrites the current value in the database. Why should it be different for non-nullable fields?

malarzm:
I agree with this being a bug, but IMO fixing it in 1.2.x is not worth breaking apps relying on this behaviour. I do believe it's worth fixing, but in 2.0 and listed as a BC break.

alcaeus:
Closing as per https://github.com/doctrine/mongodb-odm/issues/1671#issuecomment-375554586 - this is intended behavior. I'll open a separate pull request to update the handling of `null` values for `nullable` fields in upsert queries.

PR #1763, opened 2018-03-25
Add LazyCriteriaCollection to do an efficient count on large collections

lhpalacio:
This PR is a suggest to solve the problem with count on large collections when using Criteria.

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/LazyCriteriaCollectionTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/LazyCriteriaCollectionTest.php
@@ -1,0 +1,25 @@
+<?php
+
+namespace Doctrine\ODM\MongoDB\Tests;
+
+use Doctrine\Common\Collections\Criteria;
+use Doctrine\ODM\MongoDB\LazyCriteriaCollection;
+use Documents\User;
+
+class LazyCriteriaCollectionTest extends BaseTest
+{
+    public function testCountIsCached()
+    {
+        $user = new User();
+        $user->setUsername('lhpalacio');
+        $this->dm->persist($user);
+        $this->dm->flush();
+
+        $queryBuilder = $this->dm->createQueryBuilder(User::class);
+        $lazyCriteriaCollection = new LazyCriteriaCollection($queryBuilder, new Criteria());
+
+        self::assertSame(1, $lazyCriteriaCollection->count());
+        self::assertSame(1, $lazyCriteriaCollection->count());
+        self::assertSame(1, $lazyCriteriaCollection->count());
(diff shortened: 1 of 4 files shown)


tests/Doctrine/ODM/MongoDB/Tests/LazyCriteriaCollectionTest.php
```diff
@@ -0,0 +1,25 @@
...
+
+namespace Doctrine\ODM\MongoDB\Tests;
+
+use Doctrine\Common\Collections\Criteria;
+use Doctrine\ODM\MongoDB\LazyCriteriaCollection;
+use Documents\User;
+
+class LazyCriteriaCollectionTest extends BaseTest
+{
+    public function testCountIsCached()
```
jwage: Thanks for the contribution AND the test! What do you think about enhancing it to insert multiple documents and then apply a limit and offset to make sure we get the expected results in that scenario?

lhpalacio: Done! I also added some new tests.

alcaeus:
Thanks for your contribution!

> This PR is a suggest to solve the problem with count on large collections when using Criteria.

In the future, please explain why the problem happens - this took me a while because our `PersistentCollection` implementation already does this.

The problem is not `count` on a large collection, the problem is that the `matching` method creates an `ArrayCollection`, which requires a hydration cycle of all results returned. There are several workarounds for this issue:
- one would be to create a dedicated repository method instead of relying on `matching`: Repository methods normally return cursors, which (at least in 1.2.x) are countable and hydrate results on demand instead of upfront
- unless you're likely to immediately iterate over all items in the collections, it's better to hydrate the results and count the resulting array (which is the same behavior as with `matching`). When you call `count` on a cursor, the driver submits a separate `count` command to the server, which may yield different results from the original `find` command. This is one of the reasons cursors are no longer countable in the new driver
- if you are only interested in the number of items, don't run `find` but rather run `count` - this saves the server the trouble of creating and holding a cursor when all you need is the (separate) `count` command sent when calling `MongoCursor::count`.

Last but not least, this PR targets the `1.2.x` branch, which is bugfix only. Since you're introducing a new class, semantic versioning dictates that it goes into a 1.3 release. For now, I suggest opening this pull request against the `master` branch, and we will  backport it into 1.3 if necessary. 1.3 will be released around the same time as 2.0 and was meant to contain deprecations only. When creating a new pull request, keep in mind that the `master` branch uses the new MongoDB driver and thus doesn't have a `count` method on the cursor: instead, the `count` command has to be issued with a separate query builder.

PR #1764, opened 2018-03-27
Fix upsert handling of nullable fields

alcaeus:
Based on the feedback on #1761, the handling for upserting fields with a null value is correct: `null` values should never overwrite pre-existing data in upsert queries, since the document being upserted is in an unknown state and we can't know for certain that the value should be unset.

This PR brings this same behavior to fields with the `nullable` attribute set, where an upsert would always overwrite database values with a `null` value: this is now only done if the upsert results in an insertion, thus preserving pre-existing data but ensuring the normal behavior of the `nullable` flag is preserved.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -329,11 +329,15 @@ class DocumentPersister
 
         foreach (array_keys($criteria) as $field) {
             unset($data['$set'][$field]);
+            unset($data['$inc'][$field]);
+            unset($data['$setOnInsert'][$field]);
         }
 
-        // Do not send an empty $set modifier
-        if (empty($data['$set'])) {
-            unset($data['$set']);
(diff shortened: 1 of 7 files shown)


lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -329,11 +329,15 @@ private function executeUpsert($document, array $options)
 
         foreach (array_keys($criteria) as $field) {
             unset($data['$set'][$field]);
+            unset($data['$inc'][$field]);
+            unset($data['$setOnInsert'][$field]);
         }
 
         // Do not send an empty $set modifier
```
jmikola: Perhaps this should be updated to "Do not send empty update operators"?

alcaeus: Good point, changed it 👍

malarzm:
Merge time! Thanks @alcaeus

PR #1770, opened 2018-04-03
Use ::class where possible

malarzm:
This isn't even my final form

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH936Test.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH936Test.php
@@ -7,6 +7,7 @@ namespace Doctrine\ODM\MongoDB\Tests\Functional\Ticket;
 use Doctrine\ODM\MongoDB\Event\LifecycleEventArgs;
 use Doctrine\ODM\MongoDB\Events;
 use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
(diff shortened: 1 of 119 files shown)


tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH936Test.php
```diff
@@ -24,25 +24,23 @@ public function testRemoveCascadesThroughProxyDocuments()
         $this->dm->flush();
         $this->dm->clear();
 
-        $foo = $this->dm->find(GH936Document::CLASSNAME, $foo->id);
+        $foo = $this->dm->find(GH936Document::class, $foo->id);
 
-        $this->assertInstanceOf('Doctrine\ODM\MongoDB\Proxy\Proxy', $foo->ref);
+        $this->assertInstanceOf(\Doctrine\ODM\MongoDB\Proxy\Proxy::class, $foo->ref);
```
carusogabriel: CS is breaking here. This should be referenced via `use` :+1:

malarzm: Thanks for catching :+1:

malarzm:
> What did you use to do it? Something we can apply to Doctrine's CS?

@carusogabriel sadly it was manual work

malarzm:
Chances are this is everything: some occurrences were left on purpose, some might have been missed. @carusogabriel I did separate commit on purpose in case you were reading the changes, after accepting I'll squash commits

PR #1773, opened 2018-04-07
Cleanup docs and remove old sphinx related files.

jwage:


jwage:
@malarzm what should we do to get this merged to the older maintained versions?

malarzm:
@jwage I can cherry-pick to `1.2.x` as it's maintained, but not sure if it's really important to backport :)

jwage:
@malarzm makes sense. If we make other changes that modify content/structure we can backport.

PR #1781, opened 2018-04-10
Fix for #1775

### LINKED ISSUE #1775: Saving document with bi-directional reference leads to deletion of references of other documents

pascal-hofmann:
This fixes #1775 (Saving document with bi-directional reference leads to deletion of references of other documents) by explicitly loading proxied documents prior to hydration.

See #1776 for a PR that includes just the test case for #1775.


alcaeus:
Closing this as per https://github.com/doctrine/mongodb-odm/issues/1775#issuecomment-380563033 - this fix causes an additional database when the data is already given.

Issue #1697, opened 2017-12-21
[2.0] Re-implement GridFS


alcaeus:
GridFS has been removed from 2.0.x-dev to limit the complexity of the switch to `ext-mongodb`. GridFS support needs to be readded for 2.0.

Here are the notes from previous hangouts collecting ideas for a new implementation of GridFS support:

## Mapping:

```
/**
 * @ODM\File(bucketName=””, chunkSizeBytes=””, readConcern=””)
 * @ODM\WriteConcern() - optional
 * @ODM\ReadPreference() - optional
 */
private $file;
```

ODM will no longer inject additional properties like md5 or uploadDate to an owning document  - thanks to this we’ll enable having more than one @ODM\File annotated field in a document.

A field mapped as File can accept several types. From user perspective, they can assign a string or stream to the field for ODM to transform it during persist (similar to collections that are array  - they are transformed to a persistent collection). When the data is hydrated (or said transformation happens) ODM will use specialized class that wraps MongoDB\GridFS\Bucket, limiting its original API to what is needed to interact with particular file and exposing access to properties like MD5 hash, chunkFileSize et al. Effectively the class will look similar to:

```
class GridFSFile
{
    public function __construct(Bucket $bucket, ObjectId $id, array $options);

    public function downloadToStream();

    public function openDownloadStream();

    public function getMD5Hash();

    // other getters for file’s metadata
}
```

Whether we want to expose methods like delete or rename is subject to a discussion, best would be to base it on real life usages which, unfortunately, are not known to me. Removing a file could also happen any time user changes value of a field mapped as File, controlled by a mapping property.

**UPDATE**: Actually we might provide another implementation of said GridFSFile that would accept a string or a stream. This way end-user could rely on an (internal) interface and storage would be transparent to him:

```
class NotPersistedFile implements File
{
    private function __construct();

    public static function fromPath($path);

    public static function fromStream();

    // other methods from the interface (thus GridFSFile)
}
```

## Issues with current GridFS implementation
* Files are mutable - we should no longer allow changing the contents of a file document and instead treat them as write-once.
* It’s possible to store data in the root document in fs.files - this is not supported by the new driver and needs to go
* Mapping a single property as file feels wrong, the whole document is the “file”. What is mapped with `@File` is merely a placeholder for whatever currently holds the data (e.g. a string containing a path)
[...]

PR #1790, opened 2018-05-13
[2.0] Add GridFS implementation on top of mongodb/mongodb

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | yes
| Fixed issues | #1697 

### Summary

This PR brings back support for GridFS. The idea is to provide mapped documents for GridFS files, while not re-implementing upload/download functionality currently in mongodb/mongodb. The general idea is:
* Classes used for GridFS files are no longer mapped as documents, but rather as files
* For classes mapped as files, only the fields specified in the GridFS specification (and not marked as deprecated) are allowed: `_id`, `length`, `chunkSize`, `filename`, `uploadDate` and `metadata`
* Of the above fields, only the `metadata` field is checked for changes (if mapped)
* Calling `persist` on a new file instance is not supported - the file has to be uploaded using the methods in the new `GridFSRepository`. This is because the specification suggests using streams, which isn't entirely feasible at this time

### Mapping

Annotation mapping is done using special annotations (e.g. `@File`, `@File\Filename`, `@File\UploadDate`) that have predefined types and database field names and are marked as `notSaved` as a precaution.

XML mapping is done using special field mappings:
```
<gridfs-file name="Doctrine\ODM\MongoDB\Tests\Mapping\AbstractMappingDriverFile">
    <field fieldName="id" id="true" />
    <length fieldName="size" />
    <chunk-size />
    <upload-date />
    <filename fieldName="name" />

    <metadata target-document="Doctrine\ODM\MongoDB\Tests\Mapping\AbstractMappingDriverFileMetadata" />
</gridfs-file>
```

Note: the metadata field is treated like any other `embed-one` relationship; the repository does not check the given metadata object for its type when uploading a new file (see below).

### Files and objects

Files can be read and hydrated into objects like all other documents. File contents are no longer available in the object itself but has to be read using the repository methods (`downloadToStream` at this time).

### GridFS repository

This PR introduces a new interface for GridFS repositories along with a default implementation. The repository provides additional methods to upload and download files from and to streams. The repository is the only way to upload files, as mentioned above persisting a new file is not allowed and will cause an exception in `UnitOfWork` when calling `persist($newFile)`. Once a file has been read from the database, it can be updated like all other documents.

### BC breaks

* `Doctrine\ODM\MongoDB\Configuration`: `setDefaultRepositoryClassName` and `getDefaultRepositoryClassName` have been replaced by `setDefaultDocumentRepositoryClassName` and `getDefaultDocumentRepositoryClassName`, respectively to accomodate the new GridFS repositories.
* The `@File` annotation has been changed and is no longer a property-level annotation but a class-level annotation
* The GridFS implementation no longer allows mapping arbitrary fields on the file level. Any fields not indicated in the GridFS specification need to be saved in the `metadata` document
* The `Doctrine\ODM\MongoDB\DocumentRepository` class was moved to `Doctrine\ODM\MongoDB\Repository\DocumentRepository`.

### Todo

- [x] Finalize tests
- [x] ~Document BC breaks in `UPGRADE` document~
- [x] Write docs for GridFS feature, including migration


lib/Doctrine/ODM/MongoDB/Repository/DocumentRepository.php
```diff
@@ -2,15 +2,22 @@
 
 declare(strict_types=1);
 
-namespace Doctrine\ODM\MongoDB;
+namespace Doctrine\ODM\MongoDB\Repository;
```
malarzm: Can you add this to the list of BC breaks? It's also a candidate for forward compat for 1.3

alcaeus: Oh, good point 👍

lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -471,32 +474,59 @@ public function getFilterParameters($name)
...
+    /**
+     * Sets default repository class for GridFS files.
+     *
+     * @param string $className
+     *
+     * @throws MongoDBException If the class does not implement the GridFSRepository interface.
+     */
+    public function setDefaultGridFSRepositoryClassName($className)
+    {
+        $reflectionClass = new \ReflectionClass($className);
```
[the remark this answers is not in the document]
alcaeus: Apparently not - but we may also just be a couple of versions behind. I'll take a look at that in a separate PR.

lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -471,32 +474,59 @@ public function getFilterParameters($name)
...
+     * Sets default repository class for documents.
      *
      * @param string $className
      *
-     *
-     * @throws MongoDBException If not is a ObjectRepository.
+     * @throws MongoDBException If the class does not implement the ObjectRepository interface.
      */
-    public function setDefaultRepositoryClassName($className)
+    public function setDefaultDocumentRepositoryClassName($className)
```
jmikola: I imagine this was changed to distinguish between regular documents and GridFS files. Is this BC break preferable to just having `setDefaultRepositoryClassName` no longer apply to GridFS classes?

lib/Doctrine/ODM/MongoDB/DocumentManager.php
```diff
@@ -322,7 +330,12 @@ public function getDocumentCollection($className)
     {
         $className = ltrim($className, '\\');
 
+        /** @var ClassMetadata $metadata */
         $metadata = $this->metadataFactory->getMetadataFor($className);
+        if ($metadata->isFile) {
+            return $this->getDocumentBucket($className)->getFilesCollection();
```
jmikola: `getDocumentCollection()` doesn't have a return type hint, but the doc block indicates `Collection`. Does that need to be changed?

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -758,6 +765,10 @@ public function setDiscriminatorField($discriminatorField)
             return;
         }
 
+        if ($this->isFile) {
+            throw MappingException::discriminatorNotAllowedForGridFS($this->name);
```
jmikola: Would it make sense for this to go above the previous conditional that allows `null` assignment?

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -820,6 +835,10 @@ public function setDefaultDiscriminatorValue($defaultDiscriminatorValue)
...
```
jmikola: Likewise a question if it makes sense to do this before the `null` assignment condition above.

lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
```diff
@@ -81,6 +85,9 @@ public function loadMetadataForClass($className, \Doctrine\Common\Persistence\Ma
                 $class->setCollection((string) $xmlRoot['collection']);
             }
         }
+        if (isset($xmlRoot['bucket'])) {
+            $class->setCollection((string) $xmlRoot['bucket']);
```
jmikola: In this case, are you setting the GridFS bucket prefix (e.g. "fs") as the collection name? If so, does that mean that interpretation of the collection property depends on the value of `$class->isFile`?

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -148,6 +152,12 @@ public function __construct(
         $this->class = $class;
         $this->collection = $dm->getDocumentCollection($class->name);
         $this->cp = $this->uow->getCollectionPersister();
+
+        if (! $class->isFile) {
+            return;
```
jmikola: Since this is a constructor and not a specific GridFS method, I wonder if it makes more sense to conditionally initialize `$this->bucket` instead of having a guarded return here.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -451,6 +461,15 @@ public function update($document, array $options = [])
      */
     public function delete($document, array $options = [])
     {
+        if ($this->bucket) {
```
jmikola: Is this consistent with other parts of ODM, or should this use `isset($this->bucket)` or `$this->bucket !== null`?

lib/Doctrine/ODM/MongoDB/Repository/AbstractRepositoryFactory.php
```diff
@@ -48,12 +48,19 @@ public function getRepository(DocumentManager $documentManager, $documentName)
      * @param DocumentManager $documentManager The DocumentManager instance.
      * @param string          $documentName    The name of the document.
      *
-     * @return ObjectRepository
+     * @return ObjectRepository|GridFSRepository
```
jmikola: GridFSRepository extends DocumentRepository, which itself is an ObjectRepository.

Would it make more sense to change this to `DocumentRepository|GridFSRepository`, or just leave it as ObjectRepository?

lib/Doctrine/ODM/MongoDB/Repository/DefaultGridFSRepository.php
```diff
@@ -0,0 +1,79 @@
...
+use MongoDB\GridFS\Bucket;
+use MongoDB\GridFS\Exception\FileNotFoundException;
+use const PATHINFO_BASENAME;
+use function fclose;
+use function fopen;
+use function pathinfo;
+
+class DefaultGridFSRepository extends DocumentRepository implements GridFSRepository
+{
+    public function downloadToStream($id, $destination): void
```
jmikola: Would it make sense to have placeholder doc blocks here, even if they only report `@see MongoDB\GridFS\Bucket::method()`?

lib/Doctrine/ODM/MongoDB/Repository/DefaultGridFSRepository.php
```diff
@@ -0,0 +1,79 @@
...
+        return $this->getDocumentBucket()->openUploadStream($filename, $options);
+    }
+
+    public function uploadFromStream($filename, $source, $metadata = null)
+    {
+        $options = $this->prepareMetadataOptions($metadata);
+
+        $id = $this->getDocumentBucket()->uploadFromStream($filename, $source, $options);
+
+        // TODO: apply primary read preference
```
jmikola: Is this to ensure that the subsequent `find()` sees the same document that was just uploaded? Is there a particular reason you don't want to just return the document's ID as the library does?

lib/Doctrine/ODM/MongoDB/Repository/DefaultGridFSRepository.php
```diff
@@ -0,0 +1,79 @@
...
+
+        return $this->find($id);
+    }
+
+    public function uploadFromFile(string $source, ?string $filename = null, $metadata = null)
+    {
+        $resource = fopen($source, 'r');
+        if (! $resource) {
```
jmikola: Should this be a strict check for `flase`? I'm not sure if a valid file pointer in PHP can possibly evaluate to zero.

lib/Doctrine/ODM/MongoDB/Repository/DefaultGridFSRepository.php
```diff
@@ -0,0 +1,79 @@
...
+            throw MongoDBException::cannotReadGridFSSourceFile($source);
+        }
+
+        if (! $filename) {
```
jmikola: Would it make sense to explicitly check for `null` here, in case someone wanted to use an empty string or `"0"` as a filename?

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -225,6 +234,10 @@ public function ensureDocumentIndexes($documentName, $timeout = null)
             throw new \InvalidArgumentException('Cannot create document indexes for mapped super classes, embedded documents or query result documents.');
         }
 
+        if ($class->isFile) {
+            $this->ensureGridFSIndexes($class);
```
jmikola: Is this necessary? The library should already do this when inserting a file into a collection: https://github.com/mongodb/mongo-php-library/blob/1.3.2/src/GridFS/CollectionWrapper.php#L309

lib/Doctrine/ODM/MongoDB/Repository/DefaultGridFSRepository.php
SenseException: Isn't the style for return types with a space before `:` also used in ODM?
`public function downloadToStream($id, $destination) : void`

lib/Doctrine/ODM/MongoDB/Repository/DefaultGridFSRepository.php
SenseException: As far as I know a successful created resource is never true with a check like this `$resource == false`.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -542,4 +574,34 @@ private function runShardCollectionCommand($documentName)
...
+    {
+        $this->ensureChunksIndex($class);
+        $this->ensureFilesIndex($class);
+    }
+
+    private function ensureChunksIndex(ClassMetadata $class): void
+    {
+        $chunksCollection = $this->dm->getDocumentBucket($class->getName())->getChunksCollection();
+        foreach ($chunksCollection->listIndexes() as $index) {
+            if ($index->isUnique() && $index->getKey() === ['files_id' => 1, 'n' => 1]) {
```
SenseException: Could this duplicated array be in a constant or variable? It's also used in `createIndex()` a few lines later.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -542,4 +574,34 @@ private function runShardCollectionCommand($documentName)
...
+        }
+
+        $chunksCollection->createIndex(['files_id' => 1, 'n' => 1], ['unique' => true]);
+    }
+
+    private function ensureFilesIndex(ClassMetadata $class): void
+    {
+        $filesCollection = $this->dm->getDocumentCollection($class->getName());
+        foreach ($filesCollection->listIndexes() as $index) {
+            if ($index->getKey() === ['filename' => 1, 'uploadDate' => 1]) {
```
SenseException: Same as with the array above.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -664,9 +665,20 @@ private function computeOrRecomputeChangeSet(ClassMetadata $class, $document, $r
                 $changeSet = [];
             }
 
+            $gridFSMetadataProperty = null;
+
+            if ($class->isFile) {
+                try {
+                    $gridFSMetadata = $class->getFieldMappingByDbFieldName('metadata');
+                    $gridFSMetadataProperty = $gridFSMetadata['fieldName'];
+                } catch (MappingException $e) {
```
SenseException: Why is there no exception handling here? If `$class->getFieldMappingByDbFieldName('metadata')` throws a MappingException, it would also affect `$gridFSMetadataProperty`, or is this supposed to behave like that?

lib/Doctrine/ODM/MongoDB/Repository/DefaultGridFSRepository.php
alcaeus: We currently exclude the rule - this will be adjusted in a separate PR.

lib/Doctrine/ODM/MongoDB/Repository/DefaultGridFSRepository.php
alcaeus: Using a type-safe check never hurts - will update this.

lib/Doctrine/ODM/MongoDB/Repository/DefaultGridFSRepository.php
alcaeus: Correct - it's to avoid issues like https://github.com/doctrine/mongodb/pull/324 from the start. I haven't considered only returning the ID, it would've been nice to get a file you can use directly. The other option would be to return a proxy to avoid a read cycle that may be useless if the user doesn't want to use the newly persisted file.

lib/Doctrine/ODM/MongoDB/Repository/DefaultGridFSRepository.php
alcaeus: Yes, the check should be stricter.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
alcaeus: The idea was that the SchemaManager creates all collections and indexes, so no reindex is necessary later. Keeping with the thought, I decided it'd be worth creating the GridFS indexes as well. Since the GridFS classes in mongo-php-library don't expose methods for doing so, I duplicated the logic.

lib/Doctrine/ODM/MongoDB/Configuration.php
alcaeus: Yes, that would be an option if we wanted to avoid the BC break. Given how this method is mainly called by the bundle and people building the Configuration object themselves, I believe the BC break would be worth it to clarify what the repository is used for and to better distinguish it from the GridFSRepository.

lib/Doctrine/ODM/MongoDB/DocumentManager.php
alcaeus: Yes, we haven't yet consistently pulled in return types - this might or might not happen depending on how BC-breaky we want 2.0 to be.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
alcaeus: The initial thought was to allow an extending class to remove the discriminator. On second thought, that kind of object hierarchy doesn't really make sense - I'll move the check.

lib/Doctrine/ODM/MongoDB/Repository/DefaultGridFSRepository.php
alcaeus: Using a type-safe check never hurts - will update this.

lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
jmikola: I think that may be clearer, but I'll defer to @jwage and @malarzm.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
jmikola: Is that sniff designed to simply detect an `if` at the end of a function? Seems a bit odd in this case but if this needs to stay as-is for that reason I have no objection.

lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
jwage: I think we should enhance this to make it more clear and remove the dual-use of the collection property.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
alcaeus: The sniff generally enforces that you exit/continue as early as possible in control structures. Examples:
```
foreach ($array as $item) {
    if (!$item->isSpecial()) {
        $item->markSpecial();
    }
}
```

In this case, the sniff would require changing the code to
```
foreach ($array as $item) {
    if ($item->isSpecial()) {
        continue;
    }

    $item->markSpecial();
}
```

The idea is to reduce nesting. In this case, it has an unfortunate side effect.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
alcaeus: There is no exception handling: the piece of code checks to see if the mapped file has a property mapped to the `metadata` field and stores it so it can compute the changeset for it later. If there is a mapping exception anywhere here, we assume there is no `metadata` field and thus no changeset needs to be computed.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
alcaeus: As per the GridFS spec, the implementation creates the files and chunks indexes on the first write if the files collection is empty. I'm not sure what's preferrable in this case: treating GridFS as a "black box" as the spec suggests and waiting for the implementation to create the indexes at the appropriate time or whether to ensure the files and chunks collections are correctly "primed" like all other collections are.

lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -471,32 +474,59 @@ public function getFilterParameters($name)
...
         if (! $reflectionClass->implementsInterface(ObjectRepository::class)) {
             throw MongoDBException::invalidDocumentRepository($className);
         }
 
-        $this->attributes['defaultRepositoryClassName'] = $className;
+        $this->attributes['defaultDocumentRepositoryClassName'] = $className;
+    }
+
+    /**
+     * Get default repository class for documents.
```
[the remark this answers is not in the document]
alcaeus: A quick look revealed that some methods use third person, some others don't. Given the general "usefulness" of these comments (I'd say it's obvious that `getDefaultDocumentRepositoryClass` returns the default repository class for documents) I'll remove the comments in a separate pull request. I've created #1796 for this purpose.

alcaeus:
Thanks everyone for the feedback. I've adressed the feedback and squashed it into original commits where appropriate. I have a few edge cases that I still want to test, but I'd like to talk about the XML mapping first.

At the moment, GridFS files are mapped almost like regular documents, with the exception that they use a `gridfs-file` root tag instead of `document`: https://github.com/alcaeus/mongodb-odm/blob/6b270f432040d55278795c46e26147727b152b34/tests/Doctrine/ODM/MongoDB/Tests/Mapping/xml/Doctrine.ODM.MongoDB.Tests.Mapping.AbstractMappingDriverFile.dcm.xml

If we wanted to be more strict about what's allowed, we could force users to not use a generic `field` tag to map fields in a GridFS document but rather specific tags. The file linked above would thus look like this:
```
<gridfs-file name="Doctrine\ODM\MongoDB\Tests\Mapping\AbstractMappingDriverFile">
    <id fieldName="id" />
    <length fieldName="size" />
    <chunk-size fieldName="chunkSize" /> <!-- fieldName property could be omitted since it's the same as in the document -->
    <upload-date fieldName="uploadDate" />
    <name fieldName="name" />

    <metadata fieldName="metadata" className="Doctrine\ODM\MongoDB\Tests\Mapping\AbstractMappingDriverFileMetadata" />
</gridfs-file>
```

This more restrictive format could certainly help users with what they can do in GridFS mappings but it's very different. Given how little mapping there is to do for most of the default fields, these could even be attributes for the `gridfs-file` tag, leaving only the ID field to be mapped separately (due to additional properties such as type, etc.:
```
<gridfs-file
    name="Doctrine\ODM\MongoDB\Tests\Mapping\AbstractMappingDriverFile"
    lengthField="size"
    chunkSizeField="chunkSize"
    uploadDateField="uploadDate"
    nameField="name"
    metadataField="metadata"
    metadataClassName="Doctrine\ODM\MongoDB\Tests\Mapping\AbstractMappingDriverFileMetadata"
>
    <field fieldName="id" id="true" />
</gridfs-file>
```

I would appreciate some feedback on this. Thanks!

malarzm:
I think the attributes are not the best since metadata can have more settings like discriminator map which will be a nightmare to have in an attribute.

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/File/Metadata.php
```diff
@@ -0,0 +1,34 @@
...
+declare(strict_types=1);
+
+namespace Doctrine\ODM\MongoDB\Mapping\Annotations\File;
+
+use Doctrine\ODM\MongoDB\Mapping\Annotations\AbstractField;
+
+/**
+ * @Annotation
+ */
+final class Metadata extends AbstractField
```
malarzm: Out of curiosity, why doesn't this extend `EmbedOne` since technically it's one?

jmikola:
> I think the attributes are not the best since metadata can have more settings like discriminator map which will be a nightmare to have in an attribute.

I'm not sure what the use case is for discriminated GridFS files, but based on examples in https://github.com/doctrine/mongodb-odm/pull/1790#issuecomment-390104016 I think a discriminator could still be added as a child of `<gridfs-file>` since `<field>` is still permitted for the identifier. This is only moving *some* child tags into `<gridfs-file>` attributes.

In either example, does the absence of an attribute or child tag mean that the field will be unmapped?

While the attribute-based approach is most concise, I think the tag approach is consistent with existing syntax for mapped fields and even discriminator fields. The latter could easily have been an attribute, since it has one argument (field name), but it is implemented using a `<discriminator-field name="foo">` child tag.

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/File/Metadata.php
alcaeus: Good point, will update that.

alcaeus:
> I think the attributes are not the best since metadata can have more settings like discriminator map which will be a nightmare to have in an attribute.

Good point, that would be impossible when using attributes.

> I'm not sure what the use case is for discriminated GridFS files, but based on examples in #1790 (comment) I think a discriminator could still be added as a child of `<gridfs-file>` since `<field>` is still permitted for the identifier. This is only moving some child tags into `<gridfs-file>` attributes.

With the GridFS spec forbidding additional fields at the root level, we can't use the regular discriminator logic for the file itself, since there's nowhere to store the discriminator value. The `metadata` document could easily be discriminated since its discriminator value would be stored in the embedded document.

alcaeus:
XML mappings have been finalized, see original comment for example. I've also created #1806 to track changing the ID mapping to the same kind of mapping as with GridFS. This is mainly done to avoid allowing generic `field` mappings in GridFS files, with the added benefit of clearer, more concise ID mappings for all documents.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
SenseException: I'd like to see a test where a `MappingException` is thrown. So we can be sure that the method is working as expected.

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/File/Metadata.php
alcaeus: @malarzm so, the reason why I rewrote this rather than extending from it was to not have to make `EmbedOne` non-final or extracting another abstract class out of it. Not sure what's preferrable in this case.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
alcaeus: Good point - will do 👍

alcaeus: Test added, see https://github.com/doctrine/mongodb-odm/pull/1790/files#diff-bdb27b9b69468508557b928309ae6e2dR379.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
alcaeus: After consideration, I've decided to keep this as is. We're working very close to the spec anyways, so we might as well create those index ourselves if the user wants to create the complete schema.

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/File/Metadata.php
malarzm: I'd say it depends on amount of code needed to handle `Metadata` that could be handled by code dealing with `EmbedOne` already. If there's little to none, let's leave this as is :)

alcaeus: There's little to no advantage: `ClassMetadata` doesn't care about annotations, it gets field mappings. Since the mapping itself is extracted by the AnnotationDriver, there's almost no overhead there. 👍

docs/en/reference/annotations-reference.rst
```diff
@@ -373,6 +374,87 @@ Examples:
...
+--------------
+
+This maps the ``metadata`` property of a GridFS file to a property. Metadata can
+be used to store additional properties in a file. The metadata document must be
+an embedded document mapped using `@EmbeddedDocument`_.
+
+Optional attributes:
+
+-
+    targetDocument - A |FQCN| of the target document.
```
jwage: Wrap targetDocument, discriminatorField, etc. in ``? We do it in other parts of the docs and it is done in storing-files-with-mongogridfs.rst in this PR.

alcaeus: Copy-pasted it from the original documentation, I'll fix it throughout the file for consistency 👍

alcaeus:
Feature done from my point of view. I'll have to run some more tests with it, but I believe this is set for a last round of reviews  🎉

docs/en/reference/storing-files-with-mongogridfs.rst
```diff
@@ -1,190 +1,218 @@
...
 
-        public function getId(): ?string
+        public function __construct(string $contentType)
         {
-          return $this->id;
+            $this->contentType = $contentType;
         }
 
-        public function getName(): ?string
+        public function getMetadata(): string
```
jmikola: Should this also return an optional string (`?string`) in the event the `metadata.contentType` field is missing?

alcaeus: Yes, and it should also be named `getContentType` - didn't spot that until now. Will update 👍

docs/en/reference/storing-files-with-mongogridfs.rst
```diff
@@ -1,190 +1,218 @@
...
+    finally {
+        fclose($stream);
+    }
 
-    header('Content-type: image/png;');
-    echo $image->getFile()->getBytes();
+The ``downloadToStream`` method takes the identifier of a file as first argument
+and a writable stream as the second arguments. If you need to manipulate the
+file contents before writing it to disk or sending it to the client, consider
+using a memory stream using the ``php://memory`` stream wrapper.
```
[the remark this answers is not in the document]
alcaeus: We can certainly expose that wrapper as well, I haven't considered it until now

lib/Doctrine/ODM/MongoDB/Repository/DefaultGridFSRepository.php
```diff
@@ -0,0 +1,87 @@
...
+            throw DocumentNotFoundException::documentNotFound($this->getClassName(), $id);
+        }
+    }
+
+    /**
+     * @see Bucket::openUploadStream
+     */
+    public function openUploadStream(string $filename, $metadata = null)
+    {
+        $options = $this->prepareMetadataOptions($metadata);
```
jmikola: Just noticed that other upload options are not supported here. I don't recall if ODM 1.2 and earlier supported custom chunk sizes, but would it make sense to allow a configurable `chunkSizeBytes` (either on a class-basis defined in the mapping or a per-object basis)? I'm not too concerned about `_id` and `disableMD5` missing here, since those don't seem relevant.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -334,7 +361,14 @@ public function dropDocumentCollection($documentName)
         if ($class->isMappedSuperclass || $class->isEmbeddedDocument || $class->isQueryResultDocument) {
             throw new \InvalidArgumentException('Cannot delete document indexes for mapped super classes, embedded documents or query result documents.');
         }
+
         $this->dm->getDocumentCollection($documentName)->drop();
+
+        if (! $class->isFile) {
```
jmikola: Formatting, do we want a space before `!` as well?

doctrine-mongo-mapping.xsd
```diff
@@ -35,6 +36,32 @@
...
+      <xs:element name="indexes" type="odm:indexes" minOccurs="0"/>
+      <xs:element name="shard-key" type="odm:shard-key" minOccurs="0"/>
+      <xs:element name="read-preference" type="odm:read-preference" minOccurs="0" maxOccurs="1"/>
+    </xs:sequence>
+
+    <xs:attribute name="name" type="xs:string" />
+    <xs:attribute name="db" type="xs:NMTOKEN" />
+    <xs:attribute name="bucket-name" type="xs:NMTOKEN" default="fs" />
+    <xs:attribute name="repository-class" type="xs:string"/>
+    <xs:attribute name="writeConcern" type="xs:string" />
```
alcaeus: Reminder to myself: don't use camelcase here, use dashes for consistency instead.

lib/Doctrine/ODM/MongoDB/Repository/DefaultGridFSRepository.php
jmikola: Let's make a `chunkSizeBytes` an option at in the GridFS File mapping (e.g. XML attribute, annotation option), alongside `bucketName`. Additionally, we can support `chunkSizeBytes` as an option in `openUploadStream()` and `uploadFromStream()`, which are the only two ways to create new files AFAIK.

doctrine-mongo-mapping.xsd
alcaeus: Note: I'll do this in a separate PR as it needs some more changes and deprecations for 1.3.

lib/Doctrine/ODM/MongoDB/Repository/DefaultGridFSRepository.php
alcaeus: I've added the `chunkSizeBytes` option to the file mappings - I'm just not entirely sure about the name. In the GridFS file, the field is called `chunkSize`, which is why I named the mapping annotation and Xml field mapping `chunkSize` as well. I know we picked `chunkSizeBytes` to be consistent with the driver option, but we end up with two different naming within the library which is unfortunate. The question is, which one shall we use consistently?

doctrine-mongo-mapping.xsd
alcaeus: Created #1834 to fix existing occurrences and will update new names after rebasing. We use kebab-case consistently except for `fieldName` and `writeConcern` which will be fixed now.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
alcaeus: Doctrine CS doesn't take a space before `!`, especially since it's forbidden after the opening parentheses of an `if` statement.

lib/Doctrine/ODM/MongoDB/Repository/DefaultGridFSRepository.php
jmikola: `chunkSize` is the field name within documents in `fs.files`, which has not been changed by the GridFS spec. `chunkSizeBytes` is the correct name of the option in the GridFS spec. Per spec, we aim to use suffixes denoting the unit where possible.

If it's possible to use different names for each, I think that's best.

docs/en/reference/storing-files-with-mongogridfs.rst
```diff
@@ -1,190 +1,218 @@
...
-        ->field('name')->equals('Jonathan H. Wage')
-        ->getQuery()
-        ->getSingleResult();
+    $repository = $documentManager->getRepository(Documents\Image::class);
+    $file = $repository->uploadFromFile('image.jpg', '/tmp/path/to/image', new Documents\ImageMetadata('image/jpeg'));
 
-    $image = $profile->getImage();
+    $stream = fopen('tmp/path/to/copy', 'w+');
+    try {
+        $repository->downloadToStream($file->getid(), $stream);
```
malarzm: Nitpicking, but you have lowercase "i" in `getId()` ;)

docs/en/reference/annotations-reference.rst
```diff
@@ -373,6 +376,88 @@ Examples:
...
+-
+   ``readOnly`` - Prevents the file from being updated: it can only be inserted,
+   upserted or removed.
+-
+   ``writeConcern`` - Specifies the write concern for this file that overwrites
+   the default write concern specified in the configuration.
+
+.. _file_chunksize:
+
+@File\ChunkSize
```
jmikola: Does RST require escaping the backslash here, or is that only needed when referencing these titles (e.g. `` `@File\\Metadata`_ ``)?

docs/en/reference/storing-files-with-mongogridfs.rst
```diff
@@ -1,190 +1,218 @@
...
+metadata. More information on GridFS can be found in the
+`MongoDB GridFS documentation <https://docs.mongodb.com/manual/core/gridfs/>`_.
+
+GridFS files provide the following properties
+-
+    ``_id`` stores the identifier of the file. By default, it uses a BSON
+    ObjectId, although you can override this in the mapping.
+-
+    ``chunkSize`` stores the size of a single chunk in bytes. By default, chunks
+    are 255 kb in size.
```
jmikola: "kb" may be ambiguous. I'd suggest quoting the PHPLIB docs here: "261120 (i.e. 255 KiB)"

https://github.com/mongodb/mongo-php-library/blob/bd148eab0493e38354e45e2cd7db59b90fdcad79/src/GridFS/Bucket.php#L69-L70

lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -342,20 +345,39 @@ public function getFilterParameters(string $name): ?array
     /**
      * @throws MongoDBException If not is a ObjectRepository.
      */
-    public function setDefaultRepositoryClassName(string $className): void
+    public function setDefaultDocumentRepositoryClassName(string $className): void
```
malarzm: I think compat layer for this could actually should land in 1.3

docs/en/reference/storing-files-with-mongogridfs.rst
```diff
@@ -1,190 +1,218 @@
...
-        ->getSingleResult();
+            <metadata target-document="Documents\ImageMetadata" />
+        </gridfs-file>
+    </doctrine-mongo-mapping>
 
-    header('Content-type: image/png;');
-    echo $image->getFile()->getBytes();
+With XML mappings, the fields are automatically mapped to camel-cased properties.
+To change property names, simply override the ``fieldName`` attribute for each
+field. You can not override any other options for GridFS fields.
```
jmikola: Should this be "cannot"? I'm not sure what is most commonly used in the docs.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -183,6 +185,12 @@ class ClassMetadata implements BaseClassMetadata
      */
     public $collection;
 
+    /**
+     * READ-ONLY: The name of the GridFS bucket the document is mapped to.
+     * @var string
```
malarzm: This should be `string|null` I think

docs/en/reference/storing-files-with-mongogridfs.rst
```diff
@@ -1,190 +1,218 @@
...
 
-You can of course make references to this Image document from
-another document. Imagine you had a Profile document and you wanted
-every Profile to have a profile image:
+The ``ImageMetadata`` class can be any embedded document:
```
jmikola: It makes sense to say that the metadata document can be any embedded document class, but it seems odd to say that about `ImageMetadata`. `ImageMetadata` either is an embedded document class (can be used) or it isn't (cannot be used).

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -408,6 +416,20 @@ class ClassMetadata implements BaseClassMetadata
...
+     * @var bool
+     */
+    public $isFile = false;
+
+    /**
+     * READ-ONLY: The default chunk size in bytes for the file
+     *
+     * @var int|null
+     */
+    public $chunkSizeBytes = null;
```
malarzm: `= null` is redundant

docs/en/reference/storing-files-with-mongogridfs.rst
```diff
@@ -1,190 +1,218 @@
...
 
 .. code-block:: php
 
     <?php
 
-    $image = new Image();
-    $image->setName('Test image');
-    $image->setFile('/path/to/image.png');
+    $repository = $documentManager->getRepository(Documents\Image::class);
+    $file = $repository->uploadFromFile('image.jpg', '/tmp/path/to/image', new Documents\ImageMetadata('image/jpeg'));
```
jmikola: I forget if we discussed this earlier, but does it make sense to solicit optional arguments as positional arguments instead of an options array? This could prove limiting if you decide to allow future options, such as a custom identifier or `chunkSizeBytes` (assuming you allow it to be overridden on a per-file basis like PHPLIB).

docs/en/reference/storing-files-with-mongogridfs.rst
```diff
@@ -1,190 +1,218 @@
...
-    $profile->setName('Jonathan H. Wage');
-    $profile->setImage($image);
+Reading files from GridFS buckets
+---------------------------------
 
-    $dm->persist($profile);
-    $dm->flush();
+When reading GridFS files, they behave like all other documents. You can query
+for them using the ``find*`` methods in the repository, create query or
+aggregation pipeline builders and also use them as ``targetDocument`` in
```
jmikola: Oxford comma: "builders, and"

lib/Doctrine/ODM/MongoDB/MongoDBException.php
```diff
@@ -160,4 +170,19 @@ public static function commitInProgress()
...
     }
+
+    public static function documentBucketOnlyAvailableForGridFSFiles(string $className): self
+    {
+        return new self(sprintf('Cannot fetch document bucket for document "%s".', $className));
+    }
+
+    public static function cannotPersistGridFSFile(string $className): self
+    {
+        return new self(sprintf('Cannot persist GridFS file for class "%s" through UnitOfWork', $className));
```
malarzm: Missing "." by the end

lib/Doctrine/ODM/MongoDB/MongoDBException.php
```diff
@@ -160,4 +170,19 @@ public static function commitInProgress()
...
+    }
+
+    public static function cannotReadGridFSSourceFile(string $filename): self
+    {
+        return new self(sprintf('Cannot open file "%s" for uploading to GridFS', $filename));
```
malarzm: Missing "." by the end

lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -342,20 +345,39 @@ public function getFilterParameters(string $name): ?array
...
-        return $this->attributes['defaultRepositoryClassName'] ?? DocumentRepository::class;
+        return $this->attributes['defaultDocumentRepositoryClassName'] ?? DocumentRepository::class;
+    }
+
+    /**
+     * @throws MongoDBException If the class does not implement the GridFSRepository interface.
+     */
+    public function setDefaultGridFSRepositoryClassName(string $className): void
+    {
+        $reflectionClass = new \ReflectionClass($className);
```
jmikola: Should we have a `use ReflectionClass` for this instead of inline `\ReflectionClass`? Not sure what convention dictates.

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/File.php
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * @Annotation
+ */
+final class File extends AbstractDocument
+{
+    /** @var string|null */
+    public $db;
+
+    /** @var string */
+    public $bucketName = 'fs';
```
jmikola: Isn't this going to be overridden by the document name? IIRC, a `Photo` class mapped as a file in ODM 1.0 would use `photo.files` and photo.chunks` (or perhaps the plural: `photos`).

If that's the case, does it make sense to provide a default here if it will never be used?

lib/Doctrine/ODM/MongoDB/Repository/DefaultGridFSRepository.php
```diff
@@ -0,0 +1,91 @@
...
+use function fclose;
+use function fopen;
+use function pathinfo;
+
+class DefaultGridFSRepository extends DocumentRepository implements GridFSRepository
+{
+    /**
+     * @see Bucket::downloadToStream
+     */
+    public function downloadToStream($id, $destination): void
```
jmikola: Did you still intend to create an `openDownloadStream()` method? If not in this PR, should there be a separate ticket to do so before 2.0?

See: https://github.com/doctrine/mongodb-odm/pull/1790/files#r203132136

docs/en/reference/storing-files-with-mongogridfs.rst
alcaeus: Haha, will fix before merging.

lib/Doctrine/ODM/MongoDB/Configuration.php
alcaeus: Yes, definitely.

docs/en/reference/annotations-reference.rst
alcaeus: I believe we're safe when they're written as titles. I'll confirm once this is merged and fix just in case.

lib/Doctrine/ODM/MongoDB/Configuration.php
alcaeus: I believe we're a bit behind on our CS rules - I'll update those in a separate PR.

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/File.php
alcaeus: The annotation driver explicitly reads the `bucketName` instead of the document name for GridFS files: https://github.com/doctrine/mongodb-odm/pull/1790/files#diff-30e8df14148d3d67d056bb220a5cb51fR114

docs/en/reference/storing-files-with-mongogridfs.rst
alcaeus: Reworded to "must be an embedded document".

docs/en/reference/storing-files-with-mongogridfs.rst
alcaeus: A quick search favours "cannot" in our docs - updated.

docs/en/reference/storing-files-with-mongogridfs.rst
alcaeus: Updated, thanks for the suggestion.

lib/Doctrine/ODM/MongoDB/Repository/DefaultGridFSRepository.php
alcaeus: That must have slipped through, I added it now.

docs/en/reference/storing-files-with-mongogridfs.rst
alcaeus: You're correct - having optional arguments can get difficult should we add a few more options, especially since we can't add them to the interface anyways. I'm not a big fan of structs since they make static type checking difficult, so a value object may be the way to go here. What do you think of this:
```
$uploadOptions = new UploadOptions();
$uploadOptions->chunkSize = 1024 * 1024;
$uploadOptions->metadata = new Documents\ImageMetadata('image/jpeg');

$repository = $documentManager->getRepository(Documents\Image::class);
$file = $repository->uploadFromFile('image.jpg', '/tmp/path/to/image', $uploadOptions);
```

jmikola: Is there prior art for having an options object with public properties? Doesn't that also lead to type-checking issues, since properties don't yet have type hints?

All in all, I'm OK with any solution you take here provided we don't lock ourselves into taking individual, optional parameters in a method signature. That would bring us back to early ODM and `doctrine/mongodb` mistakes with super-long constructor signatures.

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/File.php
jmikola: Ah, but does the bucket name default to the document name? I'm curious if this is a behavior departure from ODM 1.0 and worry that all file documents (i.e. various mapped classes) will end up in the same `fs.files` collection if so.

docs/en/reference/storing-files-with-mongogridfs.rst
```diff
@@ -1,190 +1,248 @@
...
-        private $id;
+When using the default GridFS repository implementation, the ``uploadFromFile``
+and ``uploadFromStream`` methods return a proxy object of the file you just
+uploaded.
 
-        /** @Field */
-        private $name;
+If you want to pass options, such as a metadata object to the uploaded file, you
+If you want to add metadata to the uploaded file, you can pass it as the last
+can pass an ``UploadOptions`` object as the last argument to the
```
jmikola: "can pass it as the last" on the line above should be deleted.

docs/en/reference/storing-files-with-mongogridfs.rst
```diff
@@ -1,190 +1,248 @@
...
-    {
-        /** @Id */
-        private $id;
+When using the default GridFS repository implementation, the ``uploadFromFile``
+and ``uploadFromStream`` methods return a proxy object of the file you just
+uploaded.
 
-        /** @Field */
-        private $name;
+If you want to pass options, such as a metadata object to the uploaded file, you
```
jmikola: I think this line should be deleted entirely.

docs/en/reference/storing-files-with-mongogridfs.rst
```diff
@@ -1,190 +1,248 @@
...
+If you want to add metadata to the uploaded file, you can pass it as the last
+can pass an ``UploadOptions`` object as the last argument to the
+``uploadFromFile``, ``uploadFromStream`` or ``openUploadStream`` method call:
```
jmikola: Oxford comma after ``uploadFromStream``.

Issue #1723, opened 2018-01-09
Impossible to query nested documents with STORE_AS_REF


TiMESPLiNTER:
**doctrine/mongodb-odm 1.2.1**

I like to find all persons having an entry for company with id `foo-123`. I use the following query for it:

```php
$persons = $this->personRepository->createQueryBuilder()
            ->field('companies.company.id')->equals('foo-123')
            ->getQuery()->getQuery();
```

This leads to the following query:

```
Array
(
    [type] => 1
    [query] => Array
        (
            [companies.company._id] => "foo-123"
        )

    [newObj] => Array
        (
        )

)
```

The problem is that it somehow converts the `.id` part to `._id` although the ODM STORE_AS_REF object has only a field named `id` (without underscore).

This seems to me like a bug.

This is an example of a person document:

```json
{
    "_id" : "bar-456",
    "gender" : "MALE",
    "firstName" : "John",
    "lastName" : "Doe",
    "companies" : [ 
        {
            "company" : {
                "id" : "foo-123"
            },
            "role" : "CEO"
        }
    ]
}
```

`companies` holds `mapManyEmbedded`s of `ReferenceCompany` and this `mapOneReference`s a `STORE_AS_REF` on the `Company` document itself.

**Person ODM mapping**
```php
$metadata->mapManyEmbedded([
    'fieldName' => 'companies',
    'targetDocument' => ReferenceCompany::class,
]);
```

**ReferenceCompany ODM mapping**
```php
$metadata->mapOneReference([
    'fieldName' => 'company',
    'targetDocument' => Company::class,
    'storeAs' => ClassMetadataInfo::REFERENCE_STORE_AS_REF
]);
```

malarzm:
If I'm seeing correctly this is happening because preparing for queries does not check whether a field is a reference, just whether it has a `targetDocument` [here](https://github.com/doctrine/mongodb-odm/blob/adf0b7ed469113e93829fb777619754e7d709a7d/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php#L1169) - probably it wasn't expected that a ref can get there. @TiMESPLiNTER are you also willing to work on a fix?

Offhand you should be ok using `references` (given you have an object that is referenced).

TiMESPLiNTER:
@malarzm I'm willing to contribute yes.

**I added a test case to reproduce the behavior: #1728** 

When I try to use `references` I end in a PHP error:

```php
[...]

PR #1793, opened 2018-05-17
Fix querying fields in reference structures

malarzm:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #1723

#### Summary

While fixing #1723 it turned out that querying any reference-structure is producing wrong query (like `_id` instead of `$id` in a `DBRef` object). This should be fixed now, at least when references are in embedded documents.

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/QueryTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/QueryTest.php
@@ -6,7 +6,10 @@ use Documents\Article;
 use Documents\Account;
 use Documents\Address;
 use Documents\CmsComment;
+use Documents\Group;
+use Documents\IndirectlyReferencedUser;
 use Documents\Phonenumber;
+use Documents\ReferenceUser;
 use Documents\User;
 
 class QueryTest extends \Doctrine\ODM\MongoDB\Tests\BaseTest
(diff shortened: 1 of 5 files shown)


tests/Doctrine/ODM/MongoDB/Tests/Functional/QueryTest.php
```diff
@@ -370,6 +373,35 @@ public function testQueryReferences()
...
+        $user->indirectlyReferencedUsers[] = $indirectlyReferencedUser;
+
+        $this->dm->persist($referencedUser);
+        $this->dm->persist($user);
+        $this->dm->flush();
+
+        $qb = $this->dm->createQueryBuilder('Documents\ReferenceUser');
+
+        $referencedUsersQuery = $qb
+            ->field('indirectlyReferencedUsers.user.id')->equals(new \MongoId($referencedUser->getId()))
```
alcaeus: Would this also work with `field('indirectlyReferencedUsers.user')->references($referencedUser)`?

malarzm: No as `references` don't know how to handle dot notation (https://github.com/doctrine/mongodb-odm/issues/1723#issuecomment-356531034)

PR #1803, opened 2018-05-19
Increase PHPStan's level

malarzm:
Leftovers for level 2:

```
 ------ -----------------------------------------------------------------------------
  Line   lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Match.php
 ------ -----------------------------------------------------------------------------
  452    Call to an undefined method Doctrine\ODM\MongoDB\Query\Expr::maxDistance().
  474    Call to an undefined method Doctrine\ODM\MongoDB\Query\Expr::minDistance().
 ------ -----------------------------------------------------------------------------
```

@alcaeus I'm pretty sure I was asking you about it some time ago but I can't really remember the answer nor find anything useful on gitter...

```
 ------ ----------------------------------------------------------------------
  Line   lib/Doctrine/ODM/MongoDB/Proxy/ProxyFactory.php
 ------ ----------------------------------------------------------------------
  126    Call to an undefined method Doctrine\Common\Proxy\Proxy::__wakeup().
 ------ ----------------------------------------------------------------------
```

Technically the method is there as it's inside of a path that checked for `__wakeup` existance. I was trying getting rid of it with `assert(method_exists($proxy, '__wakeup')` to no avail. Ideas?

To get to level 4 we need to fix Annotation's `$value` typehint as it hints against `string` while it can be anything:

```
 ------ --------------------------------------------------------------
  Line   lib/Doctrine/ODM/MongoDB/Mapping/Driver/AnnotationDriver.php
 ------ --------------------------------------------------------------
  65     Call to function is_array() will always evaluate to false.
  143    Call to function is_array() will always evaluate to false.
 ------ --------------------------------------------------------------
```

To deal with level 5 we again need to tweak Common stuff:

```
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------
  Line   lib/Doctrine/ODM/MongoDB/Mapping/Driver/AnnotationDriver.php
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------
  73     Parameter #1 $map of method Doctrine\ODM\MongoDB\Mapping\ClassMetadata::setDiscriminatorMap() expects array, string given.
  266    Parameter #1 $reader of class Doctrine\ODM\MongoDB\Mapping\Driver\AnnotationDriver constructor expects Doctrine\Common\Annotations\AnnotationReader,
         Doctrine\Common\Annotations\Reader given.
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------
```

Also I have no clue so far why PHPStan is complaining about:

```
 ------ ----------------------------------------------------------------------------------------------------------------------------------------
  Line   lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
 ------ ----------------------------------------------------------------------------------------------------------------------------------------
  170    Parameter #2 $expression of method Doctrine\ODM\MongoDB\Aggregation\Expr::operator() expects
         array<Doctrine\ODM\MongoDB\Aggregation\Expr>|Doctrine\ODM\MongoDB\Aggregation\Expr, array|Doctrine\ODM\MongoDB\Aggregation\Expr given.
 ------ ----------------------------------------------------------------------------------------------------------------------------------------
```


alcaeus:
> Call to an undefined method Doctrine\ODM\MongoDB\Query\Expr::maxDistance().

You can remove the `maxDistance` and `minDistance` helpers. The methods in `Expr` were removed in 0323fb187447c6250062477c16414c195424a0d7 when we dropped `geoNear` commands completely in favor of the respective pipeline stage. I apparently forgot to throw them out in the match aggregation pipeline stage.

As for the undefined `__wakeup` method, did you try calling it conditionally, e.g. `if method_exists(...) { $proxy->__wakeup();`? If it still reports an error, this is something that may be an issue in PHPStan.

Last but not least, the argument type mismatch doesn't make any sense. The typehint for `operator` is `array|self[]|self` while the typehint for `anyElementTrue` is `array|self`, which should be perfectly safe for passing to `operator`. Again, this might be an issue with PHPStan.

carusogabriel:
PHPStan 0.10 was released. Maybe we should use it here?

malarzm:
Heh I still need to fix stuff in other Doctrine packages :/ Maybe I can schedule some time tomorrow to get back to this, I'll see how much stuff needs to change with 0.10

alcaeus:
@malarzm can you rebase and continue this or would you prefer me to take over?

malarzm:
FWIW I'm trying to rebase this

malarzm:
So this PR is again merge'able but I need to revisit levels 3 and higher as I stopped fixing errors by then just to get through all the conflicts today. BTW @jmikola I've found interesting error that might be worth fixing in the driver (but not sure if it should so I'm not opening an issue there yet):

> PHPDoc tag @throws with type MongoDB\Driver\Exception\Exception is not subtype of Throwable

alcaeus:
Reported the exception issue to the driver: https://github.com/mongodb/mongo-php-driver/issues/940. In the meantime, we can just ignore the error entirely.

lib/Doctrine/ODM/MongoDB/Types/DateType.php
```diff
@@ -69,13 +66,15 @@ public static function getDateTime($value) : DateTimeInterface
         return $datetime;
     }
 
-    private static function craftDateTime(int $seconds, $microseconds = 0) : DateTime
+    /**
+     * @return DateTime|false
```
malarzm: I think we should throw an exception on `false` and get back to returning `\DateTime`

lib/Doctrine/ODM/MongoDB/Tools/Console/Command/QueryCommand.php
```diff
@@ -67,9 +69,12 @@ protected function configure()
      */
     protected function execute(Console\Input\InputInterface $input, Console\Output\OutputInterface $output)
     {
+        $query = $input->getArgument('query');
+        assert(is_string($query));
+
         $dm = $this->getHelper('documentManager')->getDocumentManager();
         $qb = $dm->getRepository($input->getArgument('class'))->createQueryBuilder();
-        $qb->setQueryArray((array) json_decode($input->getArgument('query')));
```
malarzm: I wonder if we'd get away with `(string) $input->getArgument('query')`

alcaeus: Feel free to give this a shot 👍

alcaeus:
Regarding assertions, we should make sure assertion checks are enabled on travis-ci to make sure we're triggering an error instead of using `assert` only for documentation purposes in PHPStan.

lib/Doctrine/ODM/MongoDB/Types/DateType.php
alcaeus: We do this already a few lines later. The reason we don't do this from within `craftDateTime` is that we want to display the raw value that couldn't be converted, which isn't available inside.

lib/Doctrine/ODM/MongoDB/Tools/Console/Command/QueryCommand.php
alcaeus: Nope:

> Cannot cast array<string>|string|null to string.

alcaeus:
I finished up fixing stuff including level 7. @malarzm I think it'd be good if you could take a look at fixes from later commits.

lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
```diff
@@ -78,7 +79,9 @@ public function __construct(DocumentManager $dm, string $documentName)
      */
     public function addFields() : Stage\AddFields
     {
-        return $this->addStage(new Stage\AddFields($this));
+        $stage = $this->addStage(new Stage\AddFields($this));
+        assert($stage instanceof Stage\AddFields);
+        return $stage;
```
greg0ire: It think this can be fixed with a simple `@return static` on `addStage`.

alcaeus: No it can't: `addStage` returns the first argument and isn't extended by anything. However, I was able to work around this without an `assert` by storing the original object in a variable and returning that instead of the return value of `addStage`. Thanks for bringing it up 👍

greg0ire: Oh right I completely misread that method :sweat_smile: Good job!

malarzm:
LGTM 👍

Issue #1810, opened 2018-06-11
Add arrayToObject expression


markitosgv:
### Feature Request

Add arrayToObject expression

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

Add arrayToObject (new in Mongo 3.6) expression

malarzm:
Closing in favour of the PR :)

PR #1811, opened 2018-06-11
Add arrayToObject expression

markitosgv:
Add arrayToObject expression

|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | #1810

#### Summary

Add arrayToObject expression

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
@@ -189,7 +189,7 @@ class Expr
     /**
      * Converts an array into a single document.
      *
-     * The <expression> can be any valid expression that resolves to an array of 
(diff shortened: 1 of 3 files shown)


markitosgv:
I don't know how to fix CS style for php 7.1...

malarzm:
@markitosgv you can run `vendor/bin/phpcbf` to fix violations automatically. Apart from that, could you please add a test for the new operator? I guess it could be added [here](https://github.com/doctrine/mongodb-odm/blob/fd7345435e2618199edb630776494c38b1c317fb/tests/Doctrine/ODM/MongoDB/Tests/Aggregation/AggregationOperatorsProviderTrait.php) but I'm sure @alcaeus will know better :)

alcaeus:
Correct - the method needs to be tested as expression operator in `AggregationOperatorsProviderTrait` - you can use `$abs` as an example.

PR #1814, opened 2018-06-15
[2.0] Validate mapping files against schema

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

#### Summary

With this change, invalid XML mapping files are no longer accepted by the XML driver. It also replaces the pesky `sequence` element in the schema with `choice`, allowing fields in any order the user wants to map them. This mirrors a change done in ORM a while ago.

Last but not least, this removes leftover handling for the `id` attribute in the `field` mapping that should have been removed in #1807.

When backporting this feature to 1.3, we have to introduce a strict mode for mapping files that is disabled by default. When disabled, it will trigger a deprecation warning to inform the user that it will  throw exceptions in 2.0; when enabled, it will behave like the driver in 2.0 and throw exceptions on invalid mapping files.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
@@ -6,18 +6,27 @@ namespace Doctrine\ODM\MongoDB\Mapping\Driver;
 
 use Doctrine\Common\Persistence\Mapping\Driver\FileDriver;
 use Doctrine\ODM\MongoDB\Mapping\ClassMetadata;
+use Doctrine\ODM\MongoDB\Mapping\MappingException;
 use Doctrine\ODM\MongoDB\Utility\CollectionHelper;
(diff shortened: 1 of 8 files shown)


lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
```diff
@@ -548,4 +560,31 @@ protected function loadMappingFile($file)
...
         return $result;
     }
+
+    private function validateSchema(string $filename): void
+    {
+        $document = new DOMDocument();
+        $document->load($filename);
+
+        try {
+            $previousUseErrors = libxml_use_internal_errors(true);
```
malarzm: Moving this outside of the `try` block should be safe and satisfy PHPStan

alcaeus: Ah yes, good point! 👍

PR #1815, opened 2018-06-19
Allow increment strategy for not defined types

uLow:
|      Q       |   A
|------------- | -----------
| BC Break     | no

#### Summary

As decimal128 is not provided by doctrine types, it may be created as custom type. But even if mongo allows to use increment strategy with decimal128, doctrine throws an exception, that increment storage strategy is not allowed. Only doctrine restriction is to deal with. No other implementation is needed.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
@@ -1234,7 +1234,7 @@ class ClassMetadata implements BaseClassMetadata
 
             default:
                 $defaultStrategy = self::STORAGE_STRATEGY_SET;
-                $allowedStrategies = [self::STORAGE_STRATEGY_SET];
+                $allowedStrategies = [self::STORAGE_STRATEGY_SET, self::STORAGE_STRATEGY_INCREMENT];
         }
 
         if (! isset($mapping['strategy'])) {


alcaeus:
> As decimal128 is not provided by doctrine types, it may be created as custom type.

Yes and no - we should rather ship the new type by default, allowing users to use it without implementing custom types. This would also allow us to not allow the increment strategy for all types by default but rather allow it for the new decimal128 type.

Would you be up for adding the decimal128 type to ODM 2.0?

jmikola:
Note that in order to increment a [Decimal128](http://php.net/manual/en/class.mongodb-bson-decimal128.php) object one will need to convert it to a string, calculate the increment (likely using [`bcadd()`](http://php.net/manual/en/function.bcadd.php) from [BCMath](http://php.net/bcmath)), and convert the result back to a Decimal128 object<sup>1</sup>.

The driver's handling of Decimal128 is currently limited to conversion to/from a string. It does not implement the necessary `do_operation` object handler to allow math operations on these objects. [PHPC-847](https://jira.mongodb.org/browse/PHPC-847) is tracking implementation of the `compare_objects` handler to allow two Decimal128 objects to be compared with one another. While that's likely simpler than implementing `do_operation`, it is still not a trivial task.

On a related note, [Internal operator overloading and GMP improvements](https://wiki.php.net/rfc/operator_overloading_gmp) was implemented in PHP 5.6 (https://github.com/php/php-src/pull/342). GMP only supports integer math. The RFC did allude to supporting BCMath but I don't think that was ever implemented. That is likely because [the `bcmath` extension](http://php.net/bcmath) does not provide any objects with which to implement the handlers -- unlike `gmp`, which provides both a procedural API and an object-oriented API using the [GMP](http://php.net/manual/en/class.gmp.php) class.

The relevant code for actually handling the "increment" storage strategy is in PersistenceBuilder:

https://github.com/doctrine/mongodb-odm/blob/e7c926b8a0cae830c0b7f48005afde83fd3921a2/lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php#L149-L151

@uLow: That will require some modifications to actually support the feature you're asking for. Simply adding `STORAGE_STRATEGY_INCREMENT` to the list of allowed strategies isn't going to cut it.

Thinking more generally, the best way to implement this _may_ be to allow types (either custom or ODM's built-in types) to implement some interface that provides a method to calculate an increment.This would basically be a limited form of operator overloading used expressly for ODM's increment strategy. If implemented, a future Decimal128 type in ODM could then implement this interface and utilize BCMath internally as it wishes.

---

1. Alternatively, a future ODM type for Decimal128 might convert it to a string for its PHP representation to save users the trouble of doing so themselves. This would be similar to how BSON dates are converted to a PHP DateTime when set on entities.

alcaeus:
Given the excellent explanation by @jmikola, I'm closing this one here. Just allowing the `increment` strategy will do you no good since we'd also have to use `bcmath` to calculate the difference to send to the database server. Without that, you might as well not use the `Decimal128` type and stick with 64-bit numbers.

I've created #1835 to track the feature request to add a new `Decimal128` type which would also incorporate the logic to properly calculate the difference when using an `increment` strategy. Given our current progress on 2.0 and the timeline, I've scheduled it for inclusion in a 2.1 release.

PR #1819, opened 2018-07-06
Improvements

carusogabriel:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | -

#### Summary

Together with PHPStorm and https://github.com/kalessil/phpinspectionsea, here are some improvements that we can make to MongoDB ODM, such as unnecessaries parentheses, casts; null coalesce operator usage; in_array instead of array_search + comparison, etc

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -1365,7 +1365,7 @@ class DocumentPersister
         }
 
         // If a defaultDiscriminatorValue is set and it is among the discriminators being queries, add NULL to the list
-        if ($metadata->defaultDiscriminatorValue && array_search($metadata->defaultDiscriminatorValue, $discriminatorValues) !== false) {
+        if ($metadata->defaultDiscriminatorValue && in_array($metadata->defaultDiscriminatorValue, $discriminatorValues)) {
             $discriminatorValues[] = null;
         }
 
--- a/lib/Doctrine/ODM/MongoDB/Types/CustomIdType.php
+++ b/lib/Doctrine/ODM/MongoDB/Types/CustomIdType.php
@@ -6,18 +6,17 @@ namespace Doctrine\ODM\MongoDB\Types;
 
 /**
  * The Id type.
- *
  */
 class CustomIdType extends Type
 {
     public function convertToDatabaseValue($value)
     {
-        return $value !== null ? $value : null;
+        return $value;
     }
 
     public function convertToPHPValue($value)
     {
-        return $value !== null ? $value : null;
+        return $value;
     }
 
     public function closureToMongo()
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
@@ -1311,7 +1311,7 @@ class Expr
      */
     public function stdDevPop($expression1, ...$expressions)
     {
-        $expression = (empty($expressions)) ? $expression1 : func_get_args();
+        $expression = empty($expressions) ? $expression1 : func_get_args();
 
         return $this->operator('$stdDevPop', $expression);
     }
@@ -1329,7 +1329,7 @@ class Expr
      */
     public function stdDevSamp($expression1, ...$expressions)
     {
-        $expression = (empty($expressions)) ? $expression1 : func_get_args();
+        $expression = empty($expressions) ? $expression1 : func_get_args();
 
         return $this->operator('$stdDevSamp', $expression);
     }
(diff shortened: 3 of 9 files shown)


lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -750,7 +750,7 @@ private function loadReferenceManyCollectionOwningSide(PersistentCollectionInter
             }
 
             // only query for the referenced object if it is not already initialized or the collection is sorted
-            if (! (($reference instanceof Proxy && ! $reference->__isInitialized__)) && ! $sorted) {
+            if (! $reference instanceof Proxy && ! $reference->__isInitialized__ && ! $sorted) {
```
SenseException: This if-block has a different behaviour than before the changes.

lib/Doctrine/ODM/MongoDB/Query/ReferencePrimer.php
```diff
@@ -261,7 +261,7 @@ private function addManyReferences(PersistentCollectionInterface $persistentColl
 
             $document = $this->uow->tryGetById($id, $class);
 
-            if ($document && ! (($document instanceof Proxy && ! $document->__isInitialized()))) {
+            if ($document && ! $document instanceof Proxy && ! $document->__isInitialized()) {
```
SenseException: This if-block has a different behaviour than before the changes.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -1365,7 +1365,7 @@ private function getClassDiscriminatorValues(ClassMetadata $metadata)
         }
 
         // If a defaultDiscriminatorValue is set and it is among the discriminators being queries, add NULL to the list
-        if ($metadata->defaultDiscriminatorValue && array_search($metadata->defaultDiscriminatorValue, $discriminatorValues) !== false) {
+        if ($metadata->defaultDiscriminatorValue && in_array($metadata->defaultDiscriminatorValue, $discriminatorValues)) {
```
SenseException: I'm not sure with this one. The function `array_search()` returns null if invalid parameters are passed to it, so was the null-value in the old code a forgotten source for a bug or was null an expected result? In case this wasn't covered by tests before, can we have some to be sure that `in_array()` has the same behaviour?

alcaeus: Looks like that was a potential bug - we're only doing this when the configured `defaultDiscriminatorValue` was passed in the list of discriminator values to query for - if there's something fancy going on with those parameters, we don't need to query for `null`.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
alcaeus: Nice catch!

lib/Doctrine/ODM/MongoDB/Types/CustomIdType.php
```diff
@@ -12,12 +12,12 @@ class CustomIdType extends Type
 {
     public function convertToDatabaseValue($value)
     {
-        return $value !== null ? $value : null;
+        return $value ?? null;
```
alcaeus: What's the actual point of this line anyways? Couldn't we just make it `return $value`? No point returning `null` if the value is already `null`.

lib/Doctrine/ODM/MongoDB/Types/CustomIdType.php
```diff
@@ -12,12 +12,12 @@ class CustomIdType extends Type
...
     }
 
     public function convertToPHPValue($value)
     {
-        return $value !== null ? $value : null;
+        return $value ?? null;
```
alcaeus: Same as above.

lib/Doctrine/ODM/MongoDB/Types/CustomIdType.php
jmikola: Non-null values were originally cast to a string in https://github.com/doctrine/mongodb-odm/commit/ac4ea202f923768e716052a18dd38c919c21c1ea#diff-b229b91262ab7a824d0ea63ff9b11c3d; however, that cast was removed in https://github.com/doctrine/mongodb-odm/commit/4645babd9ab1cb4ee2473f662744dc60a3582243#diff-b229b91262ab7a824d0ea63ff9b11c3d.

There's no reason this cannot be changed to simply return the value as-is.

lib/Doctrine/ODM/MongoDB/Types/CustomIdType.php
jmikola: Likewise, this can be reduced to `return $value;`.

alcaeus:
Thanks @carusogabriel!

PR #1825, opened 2018-07-13
Use PSR-4

caciobanu:
|      Q       |   A
|------------- | -----------
| BC Break     | yes/no
| Fixed issues | none

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,7 @@
         "doctrine/coding-standard": "^4.0"
     },
     "autoload": {
-        "psr-0": { "Doctrine\\ODM\\MongoDB": "lib/" }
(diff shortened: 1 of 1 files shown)


jmikola:
Cross-referencing with https://github.com/doctrine/mongodb-odm/pull/1546.

caciobanu:
Totally forgot about that...

alcaeus:
Yeah, opinions change. Let's do this one thing at a time - changing to a PSR-4 autoloader doesn't change a lot of things, but then it doesn't hurt either. I'll refactor the tests bootstrap file (the other change in #1546) for 2.0 - alternatively, feel free to create another PR for that @caciobanu.

Whether we'll simplify library paths now that we're using PSR-4 is a different story, I'm not entirely convinced it's worth it but I appreciate any and all opinions for/against it to make up my mind 😉

PR #1832, opened 2018-07-23
Fix schema create command namespace

l-vo:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | 

#### Summary

In the "Indexes" page, the namespace of the `schema:create` command seems to come from an old version of mongodb-odm. This PR fixes this mistake.


l-vo:
Pushed again for making tests successful

PR #1836, opened 2018-07-27
Added dev autoload for composer & removed it from tests boostrap.

caciobanu:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | N/A

Follow up https://github.com/doctrine/mongodb-odm/pull/1825#issuecomment-408313167


malarzm:
Thanks @caciobanu!

PR #1840, opened 2018-08-04
[1.3] Deprecate setDefaultRepositoryClassName and getDefaultRepositoryClassName

malarzm:


lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -505,27 +505,55 @@ public function getFilterParameters($name)
...
      * Get default repository class.
      *
      * @return string
      */
-    public function getDefaultRepositoryClassName()
+    public function getDefaultDocumentRepositoryClassName()
     {
-        return isset($this->attributes['defaultRepositoryClassName'])
-            ? $this->attributes['defaultRepositoryClassName']
+        return isset($this->attributes['defaultDocumentRepositoryClassName'])
```
alcaeus: This might be a tiny BC break: people May have extended this class and expect to find the attribute with the old name. Not sure if we want to deal with that, just bringing it up.

Issue #1698, opened 2017-12-21
[2.0] Finish sharding support


alcaeus:
Sharding support is currently not completely implemented in 2.0.x-dev. The `enableSharding` action in SchemaManager relies on the driver exception to return the index suggested by the MongoDB server. Currently, `ext-mongodb` does not expose this information (see [PHPC-578](https://jira.mongodb.org/browse/PHPC-578)). Depending on the outcome and timeline, we may want to work around this limitation by ensuring the shard key is covered by a suitable index ourselves.

alcaeus:
Note: `ext-mongodb` 1.5.0 has been released, fixing PHPC-578 and exposing result documents for failed commands: sharing support can now be finished.

PR #1848, opened 2018-08-10
Finalize sharding support for 2.0

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | #1698 

#### Summary

This PR re-introduces full sharding support to ODM 2.0. Due to changes in the server, the implementation now requires the user to map an index covering the shard key. Previously, this index was created automatically by ODM based on an index suggestion returned by the server as a response to the `shardCollection` command. Since this is no longer returned, we can't implicitly create an index without duplicating the index check logic in the server. Thus, we require the user to create the index themselves.

A compat layer for 1.3 should ideally warn the user about the change when an index is implicitly created when enabling sharding for a collection.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/SchemaManager.php
+++ b/lib/Doctrine/ODM/MongoDB/SchemaManager.php
@@ -7,18 +7,21 @@ namespace Doctrine\ODM\MongoDB;
 use Doctrine\ODM\MongoDB\Mapping\ClassMetadata;
 use Doctrine\ODM\MongoDB\Mapping\ClassMetadataFactory;
 use MongoDB\Driver\Exception\RuntimeException;
+use MongoDB\Driver\Exception\ServerException;
 use MongoDB\Model\IndexInfo;
 use function array_filter;
 use function array_unique;
+use function iterator_count;
 use function iterator_to_array;
 use function ksort;
-use function strpos;
 
 class SchemaManager
 {
     private const GRIDFS_FILE_COLLECTION_INDEX = ['files_id' => 1, 'n' => 1];
 
     private const GRIDFS_CHUNKS_COLLECTION_INDEX = ['filename' => 1, 'uploadDate' => 1];
+
+    private const CODE_SHARDING_ALREADY_INITIALIZED = 23;
 
     /** @var DocumentManager */
     protected $dm;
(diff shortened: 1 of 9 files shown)


lib/Doctrine/ODM/MongoDB/MongoDBException.php
```diff
@@ -103,13 +103,13 @@ public static function failedToEnableSharding(string $dbName, string $errorMessa
         ));
     }
 
-    public static function failedToEnsureDocumentSharding(string $className, string $errorMessage): self
+    public static function failedToEnsureDocumentSharding(string $className, string $errorMessage, ?\Throwable $previous = null): self
```
jmikola: Is the previous exception ever used? I only see this called from `SchemaManager::ensureDocumentSharding()` later in this PR:

```php
throw MongoDBException::failedToEnsureDocumentSharding($documentName, $e->getMessage());
```

Does it make sense to type hint against the driver's RuntimeException here (or ServerException, in light of my comments to follow), or is that undesirable since this is technically public API? Alternatively, if this method isn't intended for public use you can simply annotate with `@internal` and do what you like.

lib/Doctrine/ODM/MongoDB/MongoDBException.php
```diff
@@ -103,13 +103,13 @@ public static function failedToEnableSharding(string $dbName, string $errorMessa
...
     {
         return new self(sprintf(
             'Failed to ensure sharding for document "%s". Error from MongoDB: %s',
             $className,
             $errorMessage
-        ));
+        ), 0, $previous);
```
jmikola: I was going to ask if we should re-use the exception code from the driver's RuntimeException (in the event that it's significant), but I suppose users can always access that through the previous exception. I think this is fine as-is.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -599,7 +574,7 @@ public function enableShardingForDbByDocumentName(string $documentName): void
         try {
             $adminDb->command(['enableSharding' => $dbName]);
         } catch (RuntimeException $e) {
-            if ($e->getCode() !== 23 || $e->getMessage() === 'already enabled') {
+            if ($e->getCode() !== 23 || strpos($e->getMessage(), 'already enabled') === false) {
```
jmikola: Looking at the [server error codes](https://github.com/mongodb/mongo/blob/master/src/mongo/base/error_codes.err), 23 corresponds to "AlreadyInitialized". Two things:

 * Consider making this an internal class constant (or private static) to avoid a magic literal here and provide some context.
 * Technically, the driver only guarantees that an exception code originates from the server for instances of [ServerException](http://php.net/manual/en/class.mongodb-driver-exception-serverexception.php) (introduced in 1.5). You likely still want to catch RuntimeException, since that considers other possible failures (e.g. socket errors), but error code checking should likely be restricted to ServerException.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -651,4 +626,18 @@ private function ensureFilesIndex(ClassMetadata $class): void
 
         $filesCollection->createIndex(self::GRIDFS_CHUNKS_COLLECTION_INDEX);
     }
+
+    private function collectionIsSharded(string $documentName): bool
+    {
+        $class = $this->dm->getClassMetadata($documentName);
+
+        $database = $this->dm->getDocumentDatabase($documentName);
+        $collections = iterator_to_array($database->listCollections(['filter' => ['name' => $class->getCollection()]]));
```
jmikola: Since `Database::listCollections()` returns a [CollectionInfoIterator](https://github.com/mongodb/mongo-php-library/blob/master/src/Model/CollectionInfoIterator.php) and you don't actually access the collection info here, would it make sense to use [`iterator_count()`](http://php.net/iterator_count) instead and bypass the array conversion?

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -651,4 +626,18 @@ private function ensureFilesIndex(ClassMetadata $class): void
...
+        if (! count($collections)) {
+            return false;
+        }
+
+        $stats = $database->command(['collstats' => $class->getCollection()])->toArray()[0];
+        return $stats['sharded'] ?? false;
```
jmikola: Forgive my ignorance, but will the `: bool` return type hint coerce `$stats['sharded']` to a boolean if it happens to be a different type? If not, would an explicit `(boolean)` cast be warranted? I'd rather be defensive than expect the MongoDB server to be consistent about its return types :)

tests/Doctrine/ODM/MongoDB/Tests/Functional/EnsureShardingTest.php
```diff
@@ -32,16 +34,12 @@ public function testEnsureShardingForNewCollection()
...
     {
-        $this->markTestSkipped('Test does not pass due to https://github.com/mongodb/mongo-php-driver/issues/296');
         $class = ShardedOne::class;
-        $collection = $this->dm->getDocumentCollection($class);
-        $doc = ['title' => 'hey', 'k' => 'hi'];
-        $collection->insertOne($doc);
-
         $this->dm->getSchemaManager()->ensureDocumentSharding($class);
 
+        $collection = $this->dm->getDocumentCollection($class);
```
jmikola: What's the point of `$indexes = iterator_to_array($collection->listIndexes());` on the next line? `$indexes` is never used.

I think that assignment of `$collection` should immediately precede the `$stats` assignment.

If the original intention was to test that no additional indexes are created, you probably want to keep the `$indexes` assignment but also assert that there's only one index (for the identifier) -- assuming the collection doesn't exist before this test.

tests/Doctrine/ODM/MongoDB/Tests/Functional/EnsureShardingTest.php
```diff
@@ -50,38 +48,53 @@ public function testEnsureShardingForCollectionWithDocuments()
...
-        $this->markTestSkipped('Test does not pass due to https://github.com/mongodb/mongo-php-driver/issues/296');
+        $class = ShardedOne::class;
+
         $document = new ShardedOne();
         $this->dm->persist($document);
         $this->dm->flush();
 
-        $class = ShardedOne::class;
+        $this->expectException(MongoDBException::class);
+        $this->expectExceptionMessageRegExp('#please create an index that starts with the (proposed )?shard key before sharding#i');
```
jmikola: Are we asserting a server error message here? If possible, I'd strongly suggest asserting that you receive a `MongoDB\Driver\Exception\ServerException` (1.5+) and use an [error code](https://github.com/mongodb/mongo/blob/master/src/mongo/base/error_codes.err) instead. Newer versions of the server should be more consistent about error codes, while message strings can change at any time.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
alcaeus: Correct, the result of the expression still needs to be cast to `bool` to ensure type safety.

tests/Doctrine/ODM/MongoDB/Tests/Functional/EnsureShardingTest.php
alcaeus: `$indexes` is used in an assertion a few lines below to make sure an index has been created.

tests/Doctrine/ODM/MongoDB/Tests/Functional/EnsureShardingTest.php
alcaeus: I dropped checking against the MongoDB portion of the exception and only ensured that this is the exception we're throwing. I'd like to avoid exposing MongoDB error codes as our own and we're currently not exposing MongoDB exceptions in instances where we catch and handle them.

tests/Doctrine/ODM/MongoDB/Tests/Functional/EnsureShardingTest.php
jmikola: Ah, I missed those assertions.

tests/Doctrine/ODM/MongoDB/Tests/Functional/EnsureShardingTest.php
jmikola: Sounds good to me.

Issue #1791, opened 2018-05-13
[2.0] Fix skipped tests


alcaeus:
As evidenced in the master build, there are a bunch of skipped tests: https://travis-ci.org/doctrine/mongodb-odm/jobs/376653436#L1607. Some of these relate to sharding support, others (e.g. SchemaManagerTest) needs to be refactored.

caciobanu:
Wanted to give **$this->markTestSkipped('mongodb-driver: query logging does not exist');**  a try but not sure how to fix it. We should delete the tests or the feature has to be implemented ?

alcaeus:
Update, there are now 3 tests that need fixing:
*  [`Doctrine\ODM\MongoDB\Tests\Functional\DocumentPersisterTest::testVersionIncrementOnUpdateSuccess`](https://github.com/doctrine/mongodb-odm/blob/4b59521bb4a16031b8b024008d88e118fc5cc5ec/tests/Doctrine/ODM/MongoDB/Tests/Functional/DocumentPersisterTest.php#L543)
* [`Doctrine\ODM\MongoDB\Tests\Functional\DocumentPersisterTest::testNoVersionIncrementOnUpdateFailure`](https://github.com/doctrine/mongodb-odm/blob/4b59521bb4a16031b8b024008d88e118fc5cc5ec/tests/Doctrine/ODM/MongoDB/Tests/Functional/DocumentPersisterTest.php#L569)
* [`Doctrine\ODM\MongoDB\Tests\Functional\ReferencePrimerTest::testPrimeReferencesInReferenceManyWithRepositoryMethodThrowsException`](https://github.com/doctrine/mongodb-odm/blob/4b59521bb4a16031b8b024008d88e118fc5cc5ec/tests/Doctrine/ODM/MongoDB/Tests/Functional/ReferencePrimerTest.php#L585)

PR #1849, opened 2018-08-10
[2.0] Removed QueryLogger from tests

caciobanu:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #1791 

#### Summary

There is still one test file left: tests/Doctrine/ODM/MongoDB/Tests/Functional/ShardKeyTest.php. I'm not sure how to update the tests in that file. Some hints would be appreciated.


malarzm:
I'm 👎 for this, without query logging tests lose most of its value - they were written specifically to ensure that `atomic` operations are firing exactly 1 query

caciobanu:
It'll take an in depth look at how that can be achieved. If you any hints would be great.

caciobanu:
I'm closing this PR as @malarzm is right.

PR #1856, opened 2018-08-24
Add UPGRADE document for ODM 2.0

alcaeus:
In preparation for ODM 2.0.0-alpha1 this adds the UPGRADE document helping users keep track of BC breaks and update tasks. From now on, pull requests have to update this document accordingly in order to be mergeable.


UPGRADE-2.0.md
```diff
@@ -0,0 +1,217 @@
...
+  check your extension points to make sure the function signatures are correct.
+* All files in ODM 2.0 use strict typing. Please make sure to not rely on type
+  coercion.
+
+## MongoDB driver change
+
+* MongoDB ODM 2.0 now requires the new MongoDB extension, `ext-mongodb`. If 
+  you've been using `alcaeus/mongo-php-adapter` you can remove it completely and
+  use ODM 2.0 directly.
+* `doctrine/mongodb` is no longer used by ODM. If you'e been relying on its
```
malarzm: you've

UPGRADE-2.0.md
```diff
@@ -0,0 +1,217 @@
...
+  functionality, plesae update accordingly. Most classes from `doctrine/mongodb`
+  have been incorporated in ODM, with the rest being replaced by the MongoDB 
+  library (`mongodb/mongodb`).
+* The constructor signature of `Doctrine\ODM\MongoDB\DocumentManager` as well as
+  the `create`, `getClient`, `getDocumentDatabase`, `getDocumentCollection`,
+  `getDocumentCollections`, have been updated to handle classes from
```
malarzm: with "methods" after `getDocumentCollections` this would sound better IMO

UPGRADE-2.0.md
```diff
@@ -0,0 +1,217 @@
...
+* Methods that previously returned a `MongoCursor` instance no longer return a
+  cursor directly but rather compose the cursor in iterators implementing the
+  `Doctrine\ODM\MongoDB\Iterator\Iterator` interface. This class provides a
+  `toArray` method in addition to the methods provided in the default `Iterator`
+  interface PHP provides.
+* The `Doctrine\ODM\MongoDB\Cursor`, `Doctrine\ODM\MongoDB\EagerCursor`, and
+  `Doctrine\ODM\MongoDB\CommandCursor` classes has been dropped without
+  replacement. You should always use the iterator interface mentioned above in
+  type hints.
+*
```
malarzm: Empty point

UPGRADE-2.0.md
```diff
@@ -0,0 +1,217 @@
...
+  this is default behavior by MongoDB that can't be changed.
+* Combining `repositoryMethod` with `criteria`, `sort`, or `limit` options on
+  inverse references is no longer supported.
+* Repository methods used with `repositoryMethod` in inverse references now have
+  to return an iterator of type `Doctrine\ODM\MongoDB\Iterator\Iterator` instead
+  of a cursor.
+* References use `dbref` as default type when storing references instead of
+  `dbRefWithDB`. This omits the `$db` key in the `DBRef` object when possible.
+  This only has implications if you consume documents outside of ODM and require
+  the `$db` key to be set.
```
malarzm: or use multiple databases

UPGRADE-2.0.md
```diff
@@ -0,0 +1,217 @@
...
+
+### YAML mapping
+
+The YAML mapping drivers have been removed. Please map your documents using XML
+or annotations.
```
malarzm: Is it worth mentioning https://github.com/doctrine/mongodb-odm/pull/1730 just yet?

UPGRADE-2.0.md
alcaeus: If users are using multiple databases we'll still insert the `$db` key if the referenced document is stored in a different database than the one owning the reference, correct?

UPGRADE-2.0.md
alcaeus: I would suggest adding that to the changelog in the PR itself - that way it's added once we merge it.

malarzm: We won't be merging #1730 to `master`, it's 1.3 only (I think :) ). But anyways, it can be added later.

UPGRADE-2.0.md
malarzm: Nope, `$db` gets inserted only if `dbRefWithDb` is used: https://github.com/doctrine/mongodb-odm/blob/a4b3c826bec0b5acf8ae00da63934f021a2903cf/lib/Doctrine/ODM/MongoDB/DocumentManager.php#L696

UPGRADE-2.0.md
alcaeus: Ah, yes. We should definitely mention both. I'll work something out to tell people to migrate to 1.3, migrate away from YAML and then go to 2.0.

UPGRADE-2.0.md
alcaeus: Whoops. Will update.

UPGRADE-2.0.md
```diff
@@ -0,0 +1,217 @@
...
+  explicit methods or use the `findBy` and `findOneBy` methods instead.
+* The `Doctrine\ODM\MongoDB\Repository\DefaultRepositoryFactory` class has been
+  made `final`.
+
+## Query
+
+* The `requireIndexes` method in `Doctrine\ODM\MongoDB\Query\Builder` has been
+  dropped without replacement. If you want to require indexes for queries, use
+  the `notablescan` option in the MongoDB server.
+* Running `geoNear` queries is no longer supported. Please use the `$geoNear`
```
jmikola: Thinking about this some more, was "`geoNear` queries" written as such because this is initialized through the query builder? To that end, I suppose you might also say "`mapReduce` queries" if needed.

If that was the intention, you can ignore my other comments about rephrasing "`geoNear` queries" to specifically refer to the `geoNear` command helper.

UPGRADE-2.0.md
```diff
@@ -0,0 +1,217 @@
...
+
+## Schema manager
+
+* The schema manager no longer implicitly creates indexes covering the shard key
+  when sharding a collection. While MongoDB creates this index automatically
+  when sharding an empty collection, it is recommended users explicitly map an
+  index covering the shard key.
+* The `createDatabases` and `createDocumentDatabases` methods have been removed
+  from `Doctrine\ODM\MongoDB\SchemaManager`. Databases are created implicitly in
+  MongoDB 3.0 with the corresponding command being deprecated.
```
jmikola: What "corresponding command" was deprecated? IIRC, there was never a command to create a database in MongoDB (at least I can't find it in the list of [2.6 administration commands](https://docs.mongodb.com/v2.6/reference/command/nav-administration/)).

UPGRADE-2.0.md
```diff
@@ -0,0 +1,217 @@
...
+
+* The `Doctrine\ODM\MongoDB\Types\FileType` class was removed completely.
+* The `Doctrine\ODM\MongoDB\Types\IncrementType` class was removed completely.
+  Use an `increment` strategy on the field mapping for `int` and `float` fields
+  instead.
+
+## Unit of work
+
+* The `commit` method in `Doctrine\ODM\MongoDB\UnitOfWork` no longer takes a
+  `$document` as first argument. Flushing single documents has been removed. If
```
jmikola: "as its first argument"

UPGRADE-2.0.md
```diff
@@ -0,0 +1,217 @@
...
+  you don't want to implicitly flush changes in all documents to the database,
+  switch to a different changeset computation strategy (e.g. explicit).
```
jmikola: You may prefer "consider using a" instead of "switch to a". (same as my earlier suggestion)

UPGRADE-2.0.md
```diff
@@ -0,0 +1,217 @@
...
+  replacement. You should always use the iterator interface mentioned above in
+  type hints.
+
+## Document manager
+
+* The `flush` method in `Doctrine\ODM\MongoDB\DocumentManager` no longer takes a
+  `$document` as first argument. Flushing single documents has been removed. If
+  you don't want to implicitly flush changes in all documents to the database,
+  switch to a different changeset computation strategy (e.g. explicit).
+* The `createDBRef` method has been dropped in favor of `createReference`.
```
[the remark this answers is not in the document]
alcaeus: Added "This new method handles the creation of different kinds of references, not only `DBRef`."

I didn't add anything about `DBRef` being deprecated, because I don't think we should do that while it's a part of MongoDB. People may by forced to use `DBRef` because of compatibility to other mappers.

UPGRADE-2.0.md
```diff
@@ -0,0 +1,217 @@
...
+
+## MongoDB driver change
+
+* MongoDB ODM 2.0 now requires the new MongoDB extension, `ext-mongodb`. If
+  you've been using `alcaeus/mongo-php-adapter` you can remove it completely and
+  use ODM 2.0 directly.
+* `doctrine/mongodb` is no longer used by ODM. If you've been relying on its
+  functionality, plesae update accordingly. Most classes from `doctrine/mongodb`
+  have been incorporated in ODM, with the rest being replaced by the MongoDB
+  library (`mongodb/mongodb`).
```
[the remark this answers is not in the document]
alcaeus: Rewording to "Most utility classes from `doctrine/mongodb` have been merged into their ODM counterparts. Classes handling connections to MongoDB servers are being replaced by the MongoDB library (`mongodb/mongodb`)."

UPGRADE-2.0.md
```diff
@@ -0,0 +1,217 @@
...
+GridFS support has been adapted to the new GridFS specification. The following
+are no longer possible:
+* Metadata has to be stored in a `metadata` embedded document. Storing
+  additional metadata in the root document is no longer supported. If you have
+  documents that store metadata in the root document, migrate those documents to
+  the new format.
+* New files are no longer persisted by flushing the DocumentManager. Instead,
+  they are uploaded using special methods in the new
+  `Doctrine\ODM\MongoDB\Repository\GridFSRepository` class.
+* GridFS files are no longer mapped as documents but as files.
```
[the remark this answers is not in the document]
alcaeus: Reworded as "The mapping of GridFS files has changed significantly: GridFS files are no longer mapped as documents but as files and there are specific annotations for the GridFS metadata fields."

UPGRADE-2.0.md
```diff
@@ -0,0 +1,217 @@
...
+  directly accessible from the document.
+
+## Mapping
+
+### General mapping changes
+
+* The `Doctrine\ODM\MongoDB\Mapping\ClassMetadataInfo` class was dropped and its
+  functionality integrated in `Doctrine\ODM\MongoDB\Mapping\ClassMetadata`.
+* The `safe` option for index creation has been dropped without replacement as
+  this is default behavior by MongoDB that can't be changed.
```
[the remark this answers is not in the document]
alcaeus: `writeConcern` for index creation is tracked in https://github.com/doctrine/mongodb-odm/issues/1861 without a target version.

UPGRADE-2.0.md
```diff
@@ -0,0 +1,217 @@
...
+  to return an iterator of type `Doctrine\ODM\MongoDB\Iterator\Iterator` instead
+  of a cursor.
+* References use `dbref` as default type when storing references instead of
+  `dbRefWithDB`. This omits the `$db` key in the `DBRef` object when possible.
+  This only has implications if you consume documents outside of ODM and require
+  the `$db` key to be set or if you are using multiple databases to store data.
+
+### YAML mapping
+
+The YAML mapping drivers have been removed. Please map your documents using XML
```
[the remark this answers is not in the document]
alcaeus: Plural because there were two, `YamlDriver` and `SimpleYamlDriver`.

UPGRADE-2.0.md
```diff
@@ -0,0 +1,217 @@
...
+or annotations. To migrate away from YAML mappings, first update to MongoDB ODM
+1.3 and convert your mappings using the provided command.
```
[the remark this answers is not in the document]
alcaeus: Decided to add this after a suggestion by @malarzm. We will elaborate on this tool in the 1.3 changelog once we create it.

UPGRADE-2.0.md
```diff
@@ -0,0 +1,217 @@
...
+  options.
+* The `strategy` attribute for `embed-one` and `reference-one` has been dropped.
+  It was never used and does not apply to single-document relationships.
+* The `cascade` element for documents now supports cascading `detach` operations
+  to related documents.
+* The `safe` attribute of `index` has been dropped without replacement.
+
+### Annotation mapping
+
+* The `$safe` property in the `@Index` annotation has been dropped without
```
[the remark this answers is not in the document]
alcaeus: This would also affect the `@UniqueIndex` class as well - I added that annotation to the sentence.

UPGRADE-2.0.md
```diff
@@ -0,0 +1,217 @@
...
+  `Date`, `Float`, `Hash`, `Increment`, `Int`, `Integer`, `Key`, `ObjectId`,
+  `Raw`, `String`, `Timestamp`.
+* The `$name` and `$fieldName` properties in the `DiscriminatorField` annotation
+  class have been dropped. The field name is now passed via the default `$value`
+  property.
+* The `Distance` annotation class has been dropped without replacement. If
+  you've been using `geoNear` queries, you have to change your queries to
+  aggregation pipeline queries.
+* The `DoctrineAnnotations.php` loader has been removed. You should register a
+  class loader in the `AnnotationRegistry` instead.
```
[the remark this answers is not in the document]
alcaeus: Correct - we should highlight that this is necessary if not using a framework integration like the Symfony Bundle or Zend Framework Module.

UPGRADE-2.0.md
```diff
@@ -0,0 +1,217 @@
...
+* The `requireIndexes` option in the `Document` annotation has been dropped
+  without replacement.
+* The `slaveOkay` option in the `Document` annotation has been dropped. Use a
+  `ReadPreference` annotation for more fine-grained read control.
+* The `File` annotation class is no longer a field-level annotation but now used
+  on a class. More information on this is available in the GridFS update
+  chapter.
+* The `simple` option for `ReferenceOne` and `ReferenceMany` annotations has
+  been dropped. Use the `storeAs` option with an appropriate value instead.
```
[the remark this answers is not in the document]
alcaeus: These are documented in the annotation reference and have been part of ODM since 1.1 with the latest option being added in 1.2.

UPGRADE-2.0.md
alcaeus: I remembered wrong then - we used `eval` to run some code which would create the database. I removed the "corresponding command" part.

PR #1860, opened 2018-09-02
Update to Doctrine CS 5.0

alcaeus:
* Aligns return type colon spacing for consistency with Doctrine CS
* Fix new issues found by Doctrine CS 5.0

// cc @Majkl578


carusogabriel:
@alcaeus Do you want some help here? :)

alcaeus:
I’m good, I’ll take care of this today.

phpcs.xml.dist
```diff
@@ -38,21 +38,14 @@
         <exclude name="SlevomatCodingStandard.Classes.SuperfluousInterfaceNaming" />
     </rule>
 
-    <!-- Require no space around colon in return types -->
-    <rule ref="SlevomatCodingStandard.TypeHints.ReturnTypeHintSpacing">
```
Majkl578: :heart:

tests/Doctrine/ODM/MongoDB/Tests/Types/DateTypeTest.php
```diff
@@ -112,10 +116,12 @@ public function testConvertToPHPValueDoesNotConvertNull()
     public function testClosureToPHP($input, $output)
     {
         $type = Type::getType(Type::DATE);
-        $return = null;
 
-        call_user_func(function ($value) use ($type, &$return) {
+        $return = call_user_func(static function ($value) use ($type) {
```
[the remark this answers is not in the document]
alcaeus: That one had to be done manually.

tests/Documents/Tournament/Participant.php
```diff
@@ -9,7 +9,7 @@
 /**
  * @ODM\Document
  * @ODM\InheritanceType("SINGLE_COLLECTION")
- * @ODm\DiscriminatorMap({"solo"=ParticipantSolo::class, "team"=ParticipantTeam::class})
+ * @ODM\DiscriminatorMap({"solo"=ParticipantSolo::class, "team"=ParticipantTeam::class})
```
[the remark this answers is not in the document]
alcaeus: This is handled by `SlevomatCodingStandard.Namespaces.UnusedUses.MismatchingCaseSensitivity`. We had that disable since having a class named `Group` in our documents causes false positives with the `@group` annotation for PHPUnit. I've dropped the offending `@group` annotations since they weren't used at all anyways.

alcaeus:
Updated to the stable version of doctrine/coding-standard. Also enabled the `Generic.Formatting.MultipleStatementAlignment` rule that was left off from our initial migration to Doctrine CS.

tests/Doctrine/ODM/MongoDB/Tests/Functional/ReadPreferenceTest.php
```diff
@@ -40,7 +40,6 @@ public function testHintIsNotSetByDefault()
     }
 
     /**
-     * @group replication_lag
```
Majkl578: Imho this is something that should be reported to Slevomat CS, likely not as a bug report, rather feature request, something like `standaloneAnnotations` (those that don't have to match to `use`s).

alcaeus: See https://github.com/slevomat/coding-standard/issues/469. Since we don’t need the groups here I took the easy way out.

PR #1866, opened 2018-09-27
Replace QueryLogger in tests with CommandLogger

alcaeus:
This PR supersedes #1850. I took the original idea and tracked the `CommandStartedEvent` as well. The reason behind this is that we need to store the original command to use it in the ShardKeyTest. Apart from that, the DataCollector used in the Symfony Bundle also needs to know all commands that were run, which is why I've moved the class to its own Sub-Namespace. With it being part in the API, we can use it in the upcoming bundle version to track queries for the web developer toolbar.

Thanks @caciobanu for getting this started - it was your implementation that gave me the idea how to handle this in the bundle 👍


lib/Doctrine/ODM/MongoDB/APM/CommandLogger.php
```diff
@@ -0,0 +1,102 @@
...
+    {
+        $commandStartedEvent = $this->getCommandStartedEvent($event->getRequestId());
+        if (! $commandStartedEvent) {
+            return;
+        }
+
+        $this->logCommand(Command::createForSucceededCommand($commandStartedEvent, $event));
+    }
+
+    public function commandFailed(CommandFailedEvent $event)
```
malarzm: What is the use case for having 2 methods for finishing a command? I think it'd be easier for a consumer to have one `commandFinished(CommandSucceededEvent|CommandFailedEvent $event)

lib/Doctrine/ODM/MongoDB/APM/CommandLogger.php
```diff
@@ -0,0 +1,102 @@
...
+     */
+    public function getAll(): array
+    {
+        return $this->commands;
+    }
+
+    private function getCommandStartedEvent(string $requestId): ?CommandStartedEvent
+    {
+        $startedEvent = $this->startedCommands[$requestId] ?? null;
+        unset($this->startedCommands[$requestId]);
```
malarzm: I think method name should reflect that state is changing, right now calling this twice will get different results

alcaeus: What do you think about `findCommandStartedEvent`?

lib/Doctrine/ODM/MongoDB/APM/CommandLogger.php
alcaeus: You'll have to ask @jmikola or @derickr as this interface is provided by the driver: http://php.net/manual/en/class.mongodb-driver-monitoring-commandsubscriber.php.

malarzm: Ah, that makes more sense now, I somehow thought that users will be calling these methods. Need a coffee ;)

lib/Doctrine/ODM/MongoDB/APM/CommandLogger.php
malarzm: `findAndRemoveCommandStartedEvent`? Lengthy but self explaining

alcaeus:
Note: corresponding functionality (still WIP) can be seen in https://github.com/doctrine/DoctrineMongoDBBundle/pull/475/commits/32f92ec481d2b73ace360ee4fd982e380d43500a.

lib/Doctrine/ODM/MongoDB/APM/CommandLogger.php
jmikola: Don't blame us, this is a cross-driver spec: https://github.com/mongodb/specifications/blob/master/source/command-monitoring/command-monitoring.rst

:smile:

PR #1867, opened 2018-09-27
Update PHPStan

alcaeus:
Updates PHPStan to the latest version and creates a config file with default settings. This allows people to run PHPStan without having to know what paths we check and what level is checked when they call it.


lib/Doctrine/ODM/MongoDB/Query/QueryExpressionVisitor.php
```diff
@@ -27,7 +27,6 @@ class QueryExpressionVisitor extends ExpressionVisitor
         Comparison::GT => 'gt',
         Comparison::GTE => 'gte',
         Comparison::IN => 'in',
-        Comparison::IS => 'equals',
```
alcaeus: > Array has 2 duplicate keys with value `'='` (`\Doctrine\Common\Collections\Expr\Comparison::EQ`, `\Doctrine\Common\Collections\Expr\Comparison::IS`).

PR #1868, opened 2018-09-28
Add docker-compose for controlled local development

olvlvl:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no

Hi,

This PR adds Docker support to help with local development.

Run `make test-container` to create the container and its MongoDB dependency. A terminal is opened once the container is ready, run `make test` to run the test suite. When you're done with the tests just enter `exit` and the container will be destroyed and it's dependency shutdown.

The container image is built from the [official PHP 7.2 image](https://hub.docker.com/_/php/), and includes the MongoDB driver, the `bcmath` extension, and Composer (to install packages from within the container).

I also added `ext-bcmath` to `composer.json` since the extension is required by the `AlnumGenerator` class, but I didn't indicate a version because I didn't know which one to pick.

I hope this helps.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,19 @@
+FROM php:7.2-alpine3.8
+
+RUN apk add --no-cache make && \
+    apk add --no-cache --virtual .php-ext-deps autoconf g++ pcre-dev && \
+    pecl channel-update pecl.php.net && \
+    pecl install mongodb && \
+    docker-php-ext-install bcmath && \
+    docker-php-ext-enable mongodb && \
+    docker-php-source delete && \
+    apk del .php-ext-deps
+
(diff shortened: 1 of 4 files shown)


alcaeus:
After internal discussion, we've decided not to merge this. None of us use docker for our development environments, so maintaining the infrastructure would be an additional chore with no extra gain.

As for the `bcmath` extension, while technically required for `AlnumGenerator`, we've intentionally left it out since it's only required if you actively specify the generator in the mapping. Instead of a hard requirement, this should be checked by the class itself, with the extension put into a `suggest` section in composer.json to explain where it's used. Please create a separate pull request for that if you're interested in making the change.

Issue #1869, opened 2018-09-28
[2.0] must be of the type array, object given


olvlvl:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | dk
| Version     | 2.0.0-alpha1

Hi,

I'm trying out `2.0.0-alpha1` and I get a lot of these: `must be of the type array, object given`. Please consider the following excerpt from `Doctrine\ODM\MongoDB\Persisters\DocumentPersister`:

```php
    /**
     * Creates or fills a single document object from an query result.
     *
     * @param object $result   The query result.
     * @param object $document The document object to fill, if any.
     * @param array  $hints    Hints for document creation.
     *
     * @return object The filled and managed document object or NULL, if the query result is empty.
     */
    private function createDocument($result, ?object $document = null, array $hints = []) : ?object
    {
        if ($result === null) {
            return null;
        }

        if ($document !== null) {
            $hints[Query::HINT_REFRESH] = true;
            $id                         = $this->class->getPHPIdentifierValue($result['_id']);
            $this->uow->registerManaged($document, $id, $result);
        }

        return $this->uow->getOrCreateDocument($this->class->name, $result, $hints, $document);
    }
```

`$result` is said to be an object (the `|null` is missing). In the test suite it's always an array, but in my application it's often a `MongoDB\Model\BSONDocument`, which can be used as an array because it extends `ArrayObject`. The problem is that both `registerManaged()` and `getOrCreateDocument()` expect `$result` to be an `array`.

From what I gathered from my tests, this happens when I'm searching for a document, using `findOneBy()` for instance.

```
        $family = $this->repository->findOneBy([

            Family::PROPERTY_COUNTRY => (string) $country,
            Family::PROPERTY_HANDLE => $handle,

        ]);
```

Raises the following exception:

```
1) HelloFresh\MenuService\Application\Command\Menu\IntegrationTest::shouldFailToActivateAnInvalidMenu
Failed asserting that exception of type "TypeError" matches expected exception "\HelloFresh\MenuService\Application\Exception\ValidationException". Message was: "Argument 2 passed to Doctrine\ODM\MongoDB\UnitOfWork::getOrCreateDocument() must be of the type array, object given, called in /app/vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php on line 601" at
[...]

PR #1870, opened 2018-09-29
Fix type hint: createDocument(array $result)

olvlvl:
|      Q       |   A
|------------- | -----------
| BC Break     | no
| Related issues | https://github.com/doctrine/mongodb-odm/issues/1869

#### Summary

I found out the hard way that the implementation only works if the MongoDB client is configured to convert `root` and `document` types to array. Therefore, I believe the type hint of the `$result` parameter of the `createDocument()` method should be `?array` instead of `object`.

I also added a commit that moves the type check to the caller method as to avoid an necessary method call. This change also allows narrowing the type hint of the `$result` parameter to `array`.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -470,6 +470,10 @@ class DocumentPersister
             }
         }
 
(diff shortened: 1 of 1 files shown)


alcaeus:
Cherry-picked into #1871. Thanks.

Issue #1813, opened 2018-06-15
Replace Doctrine\Common\Proxy with ocramius/proxy-manager


alcaeus:
As `Doctrine\Common\Proxy` is being deprecated for doctrine/common 2.9, we should use ocramius/proxy-manager for proxies in 2.0.

See https://github.com/doctrine/doctrine2/pull/6719 for the change in ORM.

jmikola:
![](https://vignette.wikia.nocookie.net/mario-fanon/images/4/40/Mario_cape_running.gif/revision/latest?cb=20160305155015&path-prefix=es)

PR #1875, opened 2018-10-09
Drop doctrine/common proxies in favor of ProxyManager

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | yes
| Fixed issues | #1813 

#### Summary

This PR replaces the proxy implementation from doctrine/common with ocramius/proxy-manager. Tests are stable, with the exception of two tests that were no longer relevant. BC breaks still need to be documented.

As for a possible BC layer, I'm not sure what this would look like or if we want to spend the time creating one.

Todos:
- [x] Prepare BC layer for 1.3? (separate PR)
- [x] Document BC breaks in `UPGRADE` document
- [x] Update documentation
- [x] ~Replace calls to `get_class` with calls to `ClassNameResolver`, in which case it should be moved out of the `Proxy` namespace?~


lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -154,15 +165,16 @@ public function setMetadataCacheImpl(Cache $cacheImpl) : void
      */
     public function setProxyDir(string $dir) : void
     {
-        $this->attributes['proxyDir'] = $dir;
+        $this->getProxyManagerConfiguration()->setProxiesTargetDir($dir);
+        $this->setAutoGenerateProxyClasses(self::AUTOGENERATE_FILE_NOT_EXISTS);
```
malarzm: I don't think that setting proxy directory should change generation strategy

lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -180,17 +192,34 @@ public function getAutoGenerateProxyClasses() : int
...
+
+        switch ($mode) {
+            case self::AUTOGENERATE_ALWAYS:
+            case self::AUTOGENERATE_FILE_NOT_EXISTS:
+                $proxyManagerConfig->setGeneratorStrategy(new FileWriterGeneratorStrategy(
+                    new FileLocator($proxyManagerConfig->getProxiesTargetDir())
+                ));
+
+                break;
+            case self::AUTOGENERATE_NEVER:
```
malarzm: This setting eval strategy looks weird

lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -180,17 +192,34 @@ public function getAutoGenerateProxyClasses() : int
...
+            case self::AUTOGENERATE_EVAL:
+            default:
```
malarzm: IMO `default` should throw `\InvalidArgumentException`

lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -410,4 +439,15 @@ public function getPersistentCollectionGenerator() : PersistentCollectionGenerat
...
+
+    public function buildGhostObjectFactory() : LazyLoadingGhostFactory
+    {
+        return new LazyLoadingGhostFactory(clone $this->getProxyManagerConfiguration());
+    }
+
+    public function getProxyManagerConfiguration() : ProxyManagerConfiguration
+    {
+        return $this->proxyManagerConfiguration
+            ?? $this->proxyManagerConfiguration = new ProxyManagerConfiguration();
```
malarzm: `proxyManagerConfiguration` could be initialized in constructor

lib/Doctrine/ODM/MongoDB/Proxy/ClassNameResolver.php
```diff
@@ -0,0 +1,41 @@
...
+     */
+    public static function getClass(object $object) : string
+    {
+        return self::getRealClass(get_class($object));
+    }
+
+    private static function getClassNameInflector() : ClassNameInflectorInterface
+    {
+        return self::$classNameInflector
+            ?? self::$classNameInflector = (new Configuration())->getClassNameInflector();
```
malarzm: Can we instantiate class name inflector on our own?

lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -410,4 +439,15 @@ public function getPersistentCollectionGenerator() : PersistentCollectionGenerat
         }
         return $this->attributes['persistentCollectionGenerator'];
     }
+
+    public function buildGhostObjectFactory() : LazyLoadingGhostFactory
+    {
+        return new LazyLoadingGhostFactory(clone $this->getProxyManagerConfiguration());
```
malarzm: What's the purpose of `clone` here?

lib/Doctrine/ODM/MongoDB/Configuration.php
alcaeus: Copy-pasted this from ORM, can drop it if necessary.

lib/Doctrine/ODM/MongoDB/Configuration.php
alcaeus: Again, this was copied from ORM: there is no `never` strategy anymore.

lib/Doctrine/ODM/MongoDB/Configuration.php
alcaeus: Not reusing the configuration object, since any later changes to the proxy configuration can cause issues. Again, only replicated ORM behaviour.

lib/Doctrine/ODM/MongoDB/Proxy/ClassNameResolver.php
alcaeus: Sure, why would we want to? This class is intended to be used in place of `get_class($document)`, so always carrying an instance around doesn't make a lot of sense, it's better to handle this statically.

lib/Doctrine/ODM/MongoDB/Configuration.php
alcaeus: Yeah, makes sense! 👍

lib/Doctrine/ODM/MongoDB/Configuration.php
malarzm: I'd drop it, consider this:
```
$config->setAutoGenerateProxyClasses(Configuration::AUTOGENERATE_ALWAYS);
$config->setProxyDir('/tmp');
```
you'd expect strategy to be AUTOGENERATE_ALWAYS, but it isn't :)

lib/Doctrine/ODM/MongoDB/Configuration.php
malarzm: Drop it (like it's hot)?

lib/Doctrine/ODM/MongoDB/Proxy/ClassNameResolver.php
malarzm: I meant not doing `new Configuration`. Also I'm not sure about implications, but [`getClassNameInflector()`](https://github.com/Ocramius/ProxyManager/blob/f7e83a6a87b8973531446394e3038849f2230212/src/ProxyManager/Configuration.php#L104) internally uses configured proxies namespaces, which can be different if somebody configured ODM to use non-standard namespace

stof: well, this is using a different configuration than the config of the DocumentManager, so this is indeed wrong (it ignores any change made by the user compared to the default config)

alcaeus: That's correct, I believe changing proxy namespace would essentially break this. I'll confirm and also report upstream in ORM where this is from - if it's broken here, it will be broken there as well.

alcaeus:
Regarding generation strategies, I'd like to give those some though. Here's what we used to have:
* `AUTOGENERATE_ALWAYS`: Regardless of whether cached proxy file exists, regenerate and write to disk
* `AUTOGENERATE_FILE_NOT_EXISTS`: If cached proxy file does not exists, generate and write to disk for future use
* `AUTOGENERATE_EVAL`: Always use `eval` to create proxies
* `AUTOGENERATE_NEVER`: Throw an error if requested proxy was not found in cache

Proxy Manager now only has two strategies that we can currently use:
* `FileWriterGeneratorStrategy`: writes cached files to disk, essentially replicating `AUTOGENERATE_FILE_NOT_EXISTS`
* `EvaluatingGeneratorStrategy`: uses `eval` to generate proxy classes

We could write our own strategies to replicate the `always` and `never` options. `always` is quite useful for development because you don't need to worry about clearing the cache while `never` is nice for production to prevent performance regressions. However, I'm not sure if we want to keep them around, WDYT?

stof:
`AUTOGENERATE_NEVER` was added to match the old behavior of doctrine/common when adding more strategies, to allow preserving BC. But `FILE_NOT_EXISTS` is a better solution anyway.

In case of read-only filesystem, both strategies have the same shortcoming: they both require you to generate all your proxies upfront:
- `never` always have this requirement (it will require the file without checking for existence, so a non-existent file is a Fatal Error)
- `file_not_exists` will try to create the missing file, and will fail on read-only filesystems.

So I don't thing the `never` strategy is actually necessary to support read-only filesystems. If you can generate all proxies upfront, `file_not_exists` will work just fine for you. and if you cannot, your only choice will be to use `eval` anyway.

alcaeus:
Ok, so we're deprecating `AUTOGENERATE_NEVER` in favour of `AUTOGENERATE_FILE_EXISTS` and `AUTOGENERATE_ALWAYS` in favour of `AUTOGENERATE_EVAL` in 1.3 and dropping them in 2.0. I'll keep the constants around for future compatibility 👍

lib/Doctrine/ODM/MongoDB/Configuration.php
alcaeus: Updated - default strategy is now set in constructor to avoid overriding later on.

lib/Doctrine/ODM/MongoDB/Proxy/ClassNameResolver.php
alcaeus: The class name resolver can now be fetched from the document manager. I was torn between exposing it in DM vs. our Configuration class but decided on the document manager since that is more readily available. I could also move it to the `ProxyFactory` but it seems counter-intuitive to fetch the proxy factory just to resolve a class name.

UPGRADE-2.0.md
```diff
@@ -168,6 +172,24 @@ or annotations. To migrate away from YAML mappings, first update to MongoDB ODM
   `Doctrine\ODM\MongoDB\Mapping\Driver\AnnotationDriver` has been dropped
   without replacement. You should register a class loader in the
   `AnnotationRegistry` instead.
+  
+## Proxy objects
+
+The proxy implementation no longer relies on Doctrine proxies but rather
+the Proxy Manager library by ocramius. If you are checking for proxies, the
```
malarzm: I think it's worth linking

UPGRADE-2.0.md
```diff
@@ -37,6 +37,10 @@
 * The `setRetryConnect` and `setRetryQuery` methods have been dropped without
   replacement. You should implement proper error handling instead of simply
   re-running queries or connection attempts.
+* The `AUTOGENERATE_ALWAYS` and `AUTOGENERATE_NEVER` generation strategies for
+  proxy objects have been deprecated. Use `AUTOGENERATE_EVAL` and
```
malarzm: I think you meant that those are removed in 2.0 as they'll throw an exception now?

malarzm:
About that "big :+1:" - tests are quite failing :see_no_evil:

alcaeus:
Weird, they were quite stable on my machine. Will have to check that out later.

Majkl578:
🎉

Issue #867, opened 2014-05-01
Exception when persisting class unlisted in disciminator map


jmikola:
For context: https://github.com/doctrine/mongodb-odm/issues/788#issuecomment-41859060

Behavior for 1.0 is to write the FQCN in the discriminator field if the particular class is not listed in the map (#865). This could lead to situations where the FQCN in the database is no longer relevant (e.g. classes were moved or renamed in code). That situation will likely always be possible when target document mappings are omitted, but if the user is opting into use a discriminator map, we can assume that they want all classes to be in that map. Throwing an exception would be an extra precaution.

PR #1812, opened 2018-06-11
 [2.0] Drop namespace property from ClassMetadata

caciobanu:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | #1795 

#### Summary

Drop namespace property from ClassMetadata and same-namespace resolution magic.

Issue #1795, opened 2018-05-18
[2.0] Drop namespace property from ClassMetadata


alcaeus:
https://github.com/doctrine/mongodb-odm/blob/8a98856ba14f5740054f49a83aedd6754a649d26/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php#L253

This involves all same-namespace name resolution magic, as used for `repositoryClassName` (and possibly other instances): https://github.com/doctrine/mongodb-odm/blob/8a98856ba14f5740054f49a83aedd6754a649d26/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php#L628

malarzm:
Closing as #1812 was merged

PR #1887, opened 2018-10-30
Exception when persisting class unlisted in disciminator map (Issue 867)

watari:
|      Q       |   A
|------------- | -----------
| BC Break     | yes
| Fixed issues | https://github.com/doctrine/mongodb-odm/issues/867

#### Summary
Updated persisting logic to throw exception of class is not listed in discriminator map.

There is no new tests is added since library have a set of old tests that is cover most of the cases.

Right now exception message contain information only about unlisted class without parent class since in the places of the checks I don't find a way of extracting parent class info.

### DIFF
--- a/docs/en/reference/annotations-reference.rst
+++ b/docs/en/reference/annotations-reference.rst
@@ -115,11 +115,11 @@ during hydration to select the instantiation class.
 This annotation is required for the top-most class in a
 :ref:`single collection inheritance <single_collection_inheritance>` hierarchy.
 It takes an array as its only argument, which maps keys to class names. The
-class constant `::class` must be used as source of class name. When
+class names must be fully qualified. Using the ``::class constant`` is supported. When
 a document is persisted to the database, its class name key will be stored in
-the discriminator field instead of the |FQCN|. If class of persisted to the
-database document is not registered in discriminator map and discriminator map
-is not empty - ``\Doctrine\ODM\MongoDB\Mapping\MappingException`` will be thrown.
+the discriminator field instead of the |FQCN|. If the discriminator map is non-empty
+and it does not contain the class name of the persisted document, a
+``\Doctrine\ODM\MongoDB\Mapping\MappingException`` will be thrown.
 
 .. code-block:: php
 


lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -1898,10 +1899,25 @@ public function mapField(array $mapping) : array
...
+         * detect discriminated entity even if it have annotations. By this reason default discriminator field name is
+         * assigned only when discriminator map is provided.
+         */
+        if (isset($mapping['association']) && ! isset($mapping['targetDocument']) && ! isset($mapping['discriminatorField']) && isset($mapping['discriminatorMap'])) {
             $mapping['discriminatorField'] = self::DEFAULT_DISCRIMINATOR_FIELD;
         }
 
+        if (isset($mapping['targetDocument']) && ! class_exists($mapping['targetDocument']) && ! interface_exists($mapping['targetDocument'])) {
+            $fullClassName = $this->reflClass->getNamespaceName() . '\\' . $mapping['targetDocument'];
+            if (! class_exists($fullClassName) && ! interface_exists($fullClassName)) {
```
watari: In some tests interfaces was used instead of classes, so I added here check for interface existence.

tests/Doctrine/ODM/MongoDB/Tests/Functional/FunctionalTest.php
```diff
@@ -599,8 +599,8 @@ public function testFavoritesReference()
         $this->assertTrue(isset($test['favorites'][0]['type']));
         $this->assertEquals('project', $test['favorites'][0]['type']);
         $this->assertEquals('group', $test['favorites'][1]['type']);
-        $this->assertTrue(isset($test['favorite']['_doctrine_class_name']));
-        $this->assertEquals(Project::class, $test['favorite']['_doctrine_class_name']);
```
watari: Here default discriminator field was on the nested document level and 'type' - on the field mapping level. I decided that in such cases field level discriminator annotations must rewrite annotations from nested document.

tests/Doctrine/ODM/MongoDB/Tests/Functional/ReferencePrimerTest.php
```diff
@@ -118,6 +118,11 @@ public function testLazyCursorForbidsPrime()
 
     public function testPrimeReferencesWithDBRefObjects()
     {
+        $this->markTestSkipped(
+            "This test is store user class name as discriminator value in default discriminator field. 
+            Since logic was updated to not use default discriminator field when dicriminator annotations is not present
+            I can't figure out how to make this test green without adding 'targetDocument' option. So skipped to review"
+        );
         $user = new User();
```
watari: Please check this out. Should I add `targetDocument` option for making test green?

lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
```diff
@@ -246,6 +246,12 @@ private function generateHydratorClass(ClassMetadata $class, string $hydratorCla
         if (isset(\$data['%1\$s'])) {
             \$reference = \$data['%1\$s'];
             \$className = \$this->unitOfWork->getClassNameForAssociation(\$this->class->fieldMappings['%2\$s'], \$reference);
+             if(!class_exists(\$className)) {
```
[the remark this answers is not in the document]
malarzm: I think that instead of adding checks we should ensure that we return FQCN from `getClassNameForAssociation`

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
malarzm: Can you please link to those tests? This sounds weird

lib/Doctrine/ODM/MongoDB/DocumentManager.php
```diff
@@ -734,27 +734,62 @@ public function createReference(object $document, array $referenceMapping)
...
-
-            /* If the discriminator value was not found in the map, use the full
-             * class name. In the future, it may be preferable to throw an
-             * exception here (perhaps based on some strictness option).
-             *
-             * @see PersistenceBuilder::prepareEmbeddedDocumentValue()
+
+            $discriminatorMap = null;
+            if (isset($referenceMapping['discriminatorMap'])) {
+                $discriminatorMap = isset($referenceMapping['discriminatorMap']);
```
malarzm: This effectively makes `$dicriminatorMap = true`

lib/Doctrine/ODM/MongoDB/DocumentManager.php
```diff
@@ -734,27 +734,62 @@ public function createReference(object $document, array $referenceMapping)
...
+            } elseif (isset($referenceMapping['discriminatorMap'])) {
```
malarzm: This is same condition as above, no?

lib/Doctrine/ODM/MongoDB/DocumentManager.php
```diff
@@ -734,27 +734,62 @@ public function createReference(object $document, array $referenceMapping)
...
+                $discriminatorMap = $referenceMapping['discriminatorMap'];
+            }
+            if (! isset($discriminatorMap)) {
```
malarzm: If this condition aims to disallow not specifying a discriminator map then I think we're taking strictness too far.

malarzm:
> This logic forbid use of empty discriminator map. Now all embedded and referenced documents must have target document or discriminator map.

This answers one of my comments :) I don't think we should go that far, not specifying `discriminatorMap` is IMO very useful and the only strictness we should introduce is, as per #867, throwing an exception if user has opted in for using a map AND document being persisted is not listed there.

lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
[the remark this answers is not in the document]
watari: In UoW in current conditions we cannot determine namespace of parent class (for example `GH426Test::testTest` test). 
Maybe we can force users use FQCNs in annotations? It will leave only one way of needed class indication and allow us remove all such checks.

lib/Doctrine/ODM/MongoDB/DocumentManager.php
watari: My mistake. Quick fix of Phpstan warning.

watari:
> > This logic forbid use of empty discriminator map. Now all embedded and referenced documents must have target document or discriminator map.
> 
> This answers one of my comments :) I don't think we should go that far, not specifying `discriminatorMap` is IMO very useful and the only strictness we should introduce is, as per #867, throwing an exception if user has opted in for using a map AND document being persisted is not listed there.

@malarzm , I really gone to far with strictness). I will decrease checks for making `discriminatorMap` optional. Can you, please, check my answers to your comments?

lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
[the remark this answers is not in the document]
malarzm: I think it should be possible to "fix" this while building class metadata: if we're mapping reference and `targetDocument` has no `\` in it, ODM should prepend it with document's namespace

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
malarzm: Eventually the interface will be "resolved" to a concrete class. Also I'm not sure if this check is needed, I think previously there was no check?

tests/Doctrine/ODM/MongoDB/Tests/Functional/FunctionalTest.php
watari: Ok. I revert my changes here and mark it as skipped with appropriate description of problem.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
watari: Previously there was no check. It was added by me and actually it is really redundant here. It will be removed

watari: After checking my changes by myself I decided to keep this check and add another similar for `discriminatorMap`. This checks is required since they is transforming short class names in FQCN. 
BTW, I think you mean something like this in https://github.com/doctrine/mongodb-odm/pull/1887#discussion_r230853646

lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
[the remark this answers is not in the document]
watari: No all is ok. I implemented it in `ClassMetadata`

watari:
@malarzm , it was not so hard revert partially changes and introduce only exception in this PR). Also I updated PR description. Can you, please, check all again?

lib/Doctrine/ODM/MongoDB/DocumentManager.php
```diff
@@ -734,27 +734,47 @@ public function createReference(object $document, array $referenceMapping)
          * class that is not defined in the discriminator map may only have a
          * discriminator field and no value, so default to the full class name.
          */
-        if (isset($class->discriminatorField)) {
-            $reference[$class->discriminatorField] = $class->discriminatorValue ?? $class->name;
+        $discriminatorField = null;
```
alcaeus: I think this entire logic needs to be extracted to a private method that returns the discriminator portion of the reference as an array (e.g. `return [$discriminatorField => $discriminatorValue];`) or an empty array if no discriminator is to be used. That would help reduce nesting and `else` code paths.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -1902,6 +1903,28 @@ public function mapField(array $mapping) : array
             $mapping['discriminatorField'] = self::DEFAULT_DISCRIMINATOR_FIELD;
         }
 
+        if (isset($mapping['targetDocument']) && ! class_exists($mapping['targetDocument']) && ! interface_exists($mapping['targetDocument'])) {
```
alcaeus: This entire block needs to go. We've explicitly removed this behaviour in #1812 as decided in #1795. Users should not rely on magic strings to be in the same namespace, instead relying on `::class` constants. If any tests still use that, it's time to update them now.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -1902,6 +1903,28 @@ public function mapField(array $mapping) : array
...
+
+            $mapping['targetDocument'] = $fullClassName;
+        }
+
+        if (isset($mapping['discriminatorMap'])) {
+            foreach ($mapping['discriminatorMap'] as $value => $targetDocument) {
+                if (class_exists($targetDocument) || interface_exists($targetDocument)) {
+                    continue;
+                }
+                $fullClassName = $this->reflClass->getNamespaceName() . '\\' . $targetDocument;
```
alcaeus: Again, no same-namespace magic here.

lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
```diff
@@ -96,7 +97,13 @@ public function prepareInsertData($document)
 
         // add discriminator if the class has one
         if (isset($class->discriminatorField)) {
-            $insertData[$class->discriminatorField] = $class->discriminatorValue ?? $class->name;
+            if ($class->discriminatorValue === null) {
```
alcaeus: How can this case come up and not be caught during the mapping process?

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH267Test.php
```diff
@@ -67,7 +67,7 @@ class GH267User
     /** @ODM\Field(type="string") */
     protected $name;
 
-    /** @ODM\ReferenceOne(name="company", targetDocument=GH267Company::class, discriminatorMap={"seller"="SellerCompany", "buyer"="BuyerCompany"}, inversedBy="users") */
+    /** @ODM\ReferenceOne(name="company", targetDocument=GH267Company::class, discriminatorMap={"seller"="GH267SellerCompany", "buyer"="GH267BuyerCompany"}, inversedBy="users") */
```
alcaeus: `::class` constants please.

lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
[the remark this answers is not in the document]
alcaeus: > I think it should be possible to "fix" this while building class metadata: if we're mapping reference and `targetDocument` has no `\` in it, ODM should prepend it with document's namespace

Nope - we decided to remove that logic for repository classes in #1795 and done so in #1812, so we shouldn't add this here.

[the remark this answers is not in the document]
malarzm: Oh nice, totally forgot about that 👍

lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
watari: Before my changes there was next line:
`$insertData[$class->discriminatorField] = $class->discriminatorValue ?? $class->name;`
So here was cases when discriminator value can become empty and instead of it name of the class was used. I only added checks here.
If I comment this out I get next tests failing:

1. [`Doctrine\ODM\MongoDB\Tests\GH788Test::testDocumentWithDiscriminatorMap`](https://github.com/doctrine/mongodb-odm/blob/master/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH788Test.php#L13) (one of the documents is unregistered in discriminator map);
2. [`Doctrine\ODM\MongoDB\Tests\GH788Test::testRefManyWithNoTargetAndExternalDiscriminatorMap`](https://github.com/doctrine/mongodb-odm/blob/master/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH788Test.php#L231) (same case);
3. [`Doctrine\ODM\MongoDB\Tests\GH788Test::testRefOneWithNoTargetAndExternalDiscriminatorMap`](https://github.com/doctrine/mongodb-odm/blob/master/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH788Test.php#L297) (same case);
4. [`Doctrine\ODM\MongoDB\Tests\Functional\Ticket\GH942Test::testDiscriminatorValueUsesClassNameIfNotInMap`](https://github.com/doctrine/mongodb-odm/blob/master/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH942Test.php#L29) (again same case).
So exception in this block is triggered when class is not registered is discriminator map that is assigned to class itself. Classname as discriminator value assignment (line 104) is used by [`\Doctrine\ODM\MongoDB\Tests\Functional\Ticket\GH942Test::testDiscriminatorValueUsesClassNameIfMapIsNotDefined`](https://github.com/doctrine/mongodb-odm/blob/master/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH942Test.php#L13)

alcaeus: Gotcha, thanks for the explanation!

watari:
@alcaeus , @malarzm , is there will be any other remarks about PR? All previous remarks is already implemented.

docs/en/reference/annotations-reference.rst
```diff
@@ -115,9 +115,11 @@ during hydration to select the instantiation class.
 This annotation is required for the top-most class in a
 :ref:`single collection inheritance <single_collection_inheritance>` hierarchy.
 It takes an array as its only argument, which maps keys to class names. The
-class names may be fully qualified or relative to the current namespace. When
+class constant `::class` must be used as source of class name. When
 a document is persisted to the database, its class name key will be stored in
-the discriminator field instead of the |FQCN|.
+the discriminator field instead of the |FQCN|. If class of persisted to the
+database document is not registered in discriminator map and discriminator map
+is not empty - ``\Doctrine\ODM\MongoDB\Mapping\MappingException`` will be thrown.
```
jmikola: I would suggest rewriting this sentence as:

> If the discriminator map is non-empty and it does not contain the class name of the persisted document, a ``\Doctrine\ODM\MongoDB\Mapping\MappingException`` will be thrown.

Alternatively, "is not empty". I'll defer to @alcaeus' preference there.

alcaeus: Either works for me, the rewording sounds good 👍

watari: Done

docs/en/reference/annotations-reference.rst
```diff
@@ -115,9 +115,11 @@ during hydration to select the instantiation class.
 This annotation is required for the top-most class in a
 :ref:`single collection inheritance <single_collection_inheritance>` hierarchy.
 It takes an array as its only argument, which maps keys to class names. The
-class names may be fully qualified or relative to the current namespace. When
+class constant `::class` must be used as source of class name. When
```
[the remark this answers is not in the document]
watari: Done

watari:
Updated docs according to suggestions

PR #1894, opened 2018-11-16
Fix inheritance of GridFS mapping properties

### LINKED ISSUE #1855: [2.0] Correctly apply GridFS file settings when inheriting metadata
When extending a class mapped as GridFS file, the document level mapping options are not correctly applied to the child class.

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | Fixes #1855

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Mapping/Annotations/File.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/Annotations/File.php
@@ -14,8 +14,8 @@ final class File extends AbstractDocument
     /** @var string|null */
     public $db;
 
-    /** @var string */
-    public $bucketName = 'fs';
+    /** @var string|null */
+    public $bucketName;
 
     /** @var string|null */
     public $repositoryClass;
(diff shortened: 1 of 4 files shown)


lib/Doctrine/ODM/MongoDB/Mapping/Annotations/File.php
```diff
@@ -14,8 +14,8 @@
     /** @var string|null */
     public $db;
 
-    /** @var string */
-    public $bucketName = 'fs';
+    /** @var string|null */
+    public $bucketName;
```
malarzm: I'd expect to find default value here from a user's perspective. Are you changing this for the sake of XML driver and default value?

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -194,7 +194,7 @@ class ClassMetadata implements BaseClassMetadata
      *
      * @var string|null
      */
-    public $bucketName;
+    public $bucketName = 'fs';
```
malarzm: This can now be typed to `string`

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/File.php
alcaeus: Without this change, just annotating an extending class as file without specifying a bucket name (as is done in the test I added) would reset the bucket name in the extending class to “fs”.

malarzm: Good point 👍

alcaeus:
@malarzm type hint is updated.

malarzm:
🎉

Issue #1854, opened 2018-08-24
[2.0] Drop simple option from reference-one and reference-many XML mapping


alcaeus:
This has been deprecated and removed from annotation mappings already.

PR #1895, opened 2018-11-16
Drop "simple" attribute from references in XML schema

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | #1854 

#### Summary

The `simple` option for references has been dropped in favour of the `storeAs` directive, so it makes sense to drop this from the schema. This PR also updates the wording in some exception messages to no longer refer to these references as "simple" but rather "identifier references".

### DIFF
--- a/doctrine-mongo-mapping.xsd
+++ b/doctrine-mongo-mapping.xsd
@@ -213,7 +213,6 @@
     <xs:attribute name="target-document" type="xs:string" />
     <xs:attribute name="field" type="xs:NMTOKEN" use="required" />
     <xs:attribute name="field-name" type="xs:NMTOKEN" />
-    <xs:attribute name="simple" type="xs:boolean" /><!-- deprecated -->
     <xs:attribute name="store-as" type="odm:reference-store-as" default="dbRefWithDb" />
     <xs:attribute name="inversed-by" type="xs:NMTOKEN" />
     <xs:attribute name="mapped-by" type="xs:NMTOKEN" />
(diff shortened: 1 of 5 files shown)


malarzm:
Tests are unhappy with your change ;)

alcaeus:
Meh, took a gamble on those exception messages. Who’d think we actually check them in tests 😛

malarzm:
You'd be surprised :D https://github.com/doctrine/mongodb-odm/search?q=expectedExceptionMessage&type=Code

alcaeus:
Tests fixed 👍

malarzm:
Awesome! 👍

etshy:
Just to let you know about this change.
Seems  IDEs (PhpStorm at least) don't like the `<!-- deprecated -->` comment after the element.
When I map Documents with XML PhpStorm warn me the `store-as` attribute is deprecated.

I don't know if it comes only from how Phpstorm is reading the xsd file or if the `<!-- deprecated -->` comment should be before the element that's deprecated.

jmikola:
@etshy: Would you mind submitting a quick PR to move the comment on its own line above the "simple" attributes? Feel free to link back to this PR in your description for added context.

According to https://stackoverflow.com/q/1950075/162228, there isn't a standardized way to mark elements in an XSD as deprecated. It does seem like PhpStorm is picking up on the comment and applying it to the next element it sees; however, I wasn't able to find any reference to that behavior online. The SO thread above also suggests using an `xs:documentation` element in lieu of a `<!--`, which might suppress PhpStorm's warning altogether, although it sounds like it might actually be helpful if we can get it to trigger on the "simple" attribute instead of "store-as".

etshy:
I just realized my problem is not about this PR but surely a much older one... as this one remove completely the `simple` attribute support.

Do I still make a PR to 1.3 branch, or there is no need ?

PR #1896, opened 2018-11-16
Drop dependency on doctrine/common

### LINKED ISSUE #1823: [2.0] Drop doctrine/common
With doctrine/common [being phased out](https://www.doctrine-project.org/2018/07/12/common-2-9-and-dbal-2-8-and-orm-2-6-2.html), we should remove our dependency on `doctrine/common` in favor of separate packages.

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | Fixes #1823 

#### Summary

This PR removes the dependency on doctrine/common and replaces it with more explicit requirements on the other packages. This PR needs to wait for the release of `doctrine/persistence` 1.1 which will include the `NotifyPropertyChanged` interfaces.


composer.json
```diff
@@ -27,11 +27,13 @@
         "doctrine/annotations": "^1.6",
         "doctrine/cache": "^1.7",
         "doctrine/collections": "^1.5",
-        "doctrine/common": "^2.9",
+        "doctrine/event-manager": "^1.0",
         "doctrine/instantiator": "^1.1",
+        "doctrine/persistence": "dev-master",
```
malarzm: FTR, this needs to change before merging

alcaeus: 1.1.0 has been released and PR updated 👍

PR #1897, opened 2018-11-16
add mergeObjects() function

Romaxx:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
@@ -907,7 +907,23 @@ class Expr
     {
         return $this->operator('$max', $expression);
     }
-
+    
+    /**
+     * Combines multiple documents into a single document.
+     *
+     * @see http://docs.mongodb.com/manual/reference/operator/aggregation/mergeObjects/
+     *
+     * @param Object1,Object2,...
+     */
+    public function mergeObjects()
+    {
+        $arrayarg = array();
+        foreach (func_get_args() as $param) {
+            $arrayarg[] =$param;
+        }
+        return $this->operator('$mergeObjects', $arrayarg);
+    }
+    
     /**
      * Returns the metadata associated with a document in a pipeline operations.
      *


lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
```diff
@@ -907,7 +907,23 @@ public function max($expression) : self
...
         return $this->operator('$max', $expression);
     }
-
+    
+    /**
+     * Combines multiple documents into a single document.
+     *
+     * @see http://docs.mongodb.com/manual/reference/operator/aggregation/mergeObjects/
+     *
+     * @param Object1,Object2,...
```
alcaeus: This typehint is wrong, please refer to other methods for guidance.

lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
```diff
@@ -907,7 +907,23 @@ public function max($expression) : self
...
+     */
+    public function mergeObjects()
```
alcaeus: Please look at other methods as to how to handle these arguments. Looking at the documentation for this operator, you can use `$expression1, $expression2, ...$expressions` and appropriate `@param` blocks. The `add` helper has a similar signature.

lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
```diff
@@ -907,7 +907,23 @@ public function max($expression) : self
...
+    {
+        $arrayarg = array();
```
alcaeus: There's no need to copy the contents of `func_get_args` in a separate array, just pass them to the `operator` helper directly (again, see the `add` helper for example).

lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
```diff
@@ -907,7 +907,23 @@ public function max($expression) : self
     {
         return $this->operator('$max', $expression);
     }
-
+
```
alcaeus: There are trailing spaces in this line which must not be there.

alcaeus (changes requested):
Thank you for your contribution. Apart from the comments below, please note the following:
- Please add tests as they are there for other aggregation pipeline operators. Take a look at `AggregationOperatorsProviderTrait` for guidance.
- Just adding the operator to `Expr` doesn't help, you also need to expose it in the `Operator` class.

We'll have to see how to deal with the two different syntaxes for `$mergeObjects`: it can take a single argument (e.g. when used as accumulator in a `$group` stage) or 2 or more arguments (when used in other stages or operators). This also affects some other operators (e.g. `$min` and `$max` and currently isn't properly handled).

Please also note that I'm scheduling this for 2.1 as we've got a whole load of aggregation operators that need adding. You can always create the operator manually through the `expression` helper in the meantime.

malarzm:
`mergeObjects` was introduced with https://github.com/doctrine/mongodb-odm/pull/2514

PR #1899, opened 2018-11-20
Error in group aggregation builder

nassuf:
After spending some hours trying to figure out why I had the error "can't convert from BSON type string to Date" I founded that there is an error in the doc
"->month('purchaseDate')" and "->year('purchaseDate')" instead of "->month('$purchaseDate')" and "->year('$purchaseDate')"

|      Q       |   A
|------------- | -----------
| BC Break     | yes/no
| Fixed issues |


alcaeus:
Thanks @nassuf!

PR #1901, opened 2018-11-26
Fix skipped tests

### LINKED ISSUE #1791: [2.0] Fix skipped tests

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | Fixes #1791

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/DocumentPersisterTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/DocumentPersisterTest.php
@@ -540,23 +540,10 @@ class DocumentPersisterTest extends BaseTest
 
     public function testVersionIncrementOnUpdateSuccess()
     {
-        $this->markTestSkipped('Mocking results to update calls is no longer possible. Rewrite test to not rely on mocking');
+        $testDocument = new DocumentPersisterTestDocumentWithVersion();
+        $this->dm->persist($testDocument);
+        $this->dm->flush();
 
-        $class             = DocumentPersisterTestDocumentWithVersion::class;
-        $documentPersister = $this->uow->getDocumentPersister($class);
-
-        $collection = $this->createMock(Collection::class);
-        $collection->expects($this->any())
-            ->method('updateOne')
-            ->will($this->returnValue(['n' => 1]));
-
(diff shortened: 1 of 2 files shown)


tests/Doctrine/ODM/MongoDB/Tests/Functional/ReferencePrimerTest.php
```diff
@@ -580,35 +580,6 @@ public function testPrimeReferencesInReferenceMany()
         $this->assertTrue($comment->author->isProxyInitialized());
     }
 
-    public function testPrimeReferencesInReferenceManyWithRepositoryMethodThrowsException()
```
alcaeus: This test was added since we were requiring eager cursors in order to prime references; this is no longer a requirement so I decided to drop the test.

tests/Doctrine/ODM/MongoDB/Tests/Functional/DocumentPersisterTest.php
```diff
@@ -566,26 +553,17 @@ public function testVersionIncrementOnUpdateSuccess()
...
-        $reflectionProperty->setValue($documentPersister, $collection);
+        $this->dm->getDocumentCollection($class)->updateOne([], ['$inc' => ['revision' => 1]]);
 
-        $testDocument     = new $class();
-        $testDocument->id = 12345;
-        $this->uow->registerManaged($testDocument, 12345, ['id' => 12345]);
-        $this->expectException(LockException::class);
         $testDocument->name = 'test';
         $this->dm->persist($testDocument);
+        $this->expectException(LockException::class);
```
malarzm: I think this was broken before, but `$this->assertEquals(1, $testDocument->revision);` won't be executed with exception thrown from `flush()`?

alcaeus: Yep. I'll drop it.

PR #1902, opened 2018-11-26
Remove eager cursor functionality without replacement

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues |

#### Summary

After the cursor refactoring which was done along with the driver switch, we no longer respected the `eagerCursor` flag in the query options as it wasn't really necessary. Eager cursor were introduced for two reasons:
* The legacy driver allowed iterating a cursor multiple times, but would re-execute the query multiple times. An eager cursor prevented this. Since the new driver forbids rewinding a cursor, we use a CachingIterator to cache results on the first iteration and use that result cache on subsequent operations
* Reference priming required the use of an eager cursor to prevent different results on the first iteration by the user (which was the second on a cursor due to priming). Again, with the CachingIterator this is no longer necessary
* The eager cursor fetched all data at the beginning of the iteration without talking to the MongoDB server afterwards. I don't think this is a use case anymore since the connection to MongoDB has gotten way more stable than back in the days. Users that want to avoid issues during iteration due to connection problems can force an initialisation of the returned iterator at the beginning of their use.


malarzm:
🎉

PR #1903, opened 2018-11-26
Throw exception on duplicate database names within a document

### LINKED ISSUE #924: Raise an exception if a same name is defined for multiple fields
Consider following document

``` php
/**
 * @MongoDB\Document
 */
class Person
{

    /**
     * @MongoDB\String(name="a")
     */
    public $address;

    /**
     * @MongoDB\Boolean(name="a")
     */
    public $active;
}
```

Even `a` is set as a name for two fields, ODM silently ignores this and takes the last field in use. It would be useful if a `MappingException` or some other suitable exception was thrown.

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | yes
| Fixed issues | Fixes #924 

#### Summary

As described in #924, this adds check to the metadata drivers to ensure that database names are not reused within a document. This check is done in the metadata drivers to ensure that it's still possible to re-declare a field from a parent class to change its mapping.


tests/Documents/User.php
```diff
@@ -50,10 +50,10 @@ class User extends BaseDocument
     /** @ODM\ReferenceMany(targetDocument=Group::class, cascade={"all"}, strategy="addToSet") */
     protected $uniqueGroups;
 
-    /** @ODM\ReferenceMany(targetDocument=Group::class, name="groups", sort={"name"="asc"}, strategy="setArray") */
+    /** @ODM\ReferenceMany(targetDocument=Group::class, name="groupsAsc", sort={"name"="asc"}, strategy="setArray") */
```
alcaeus: This was the only semi-legitimate use of the old "feature" I found. These sorted collections are supposed to behave like inverse collections and they do so at hydration time. However, any change to these collections would also be written back to the original property, potentially leading to duplicates or conflicts.

The functionality tested by these properties could be re-introduced with read-only properties that are ignored on changes computation. Read-only properties could be excluded by the duplicate check in the drivers as they would not be written to the database at all.

alcaeus: And I just now realised that we have the "NotSaved" functionality for exactly that purpose. Will update the code accordingly.

alcaeus:
I was able to move the check to the metadata class, so we don't need to implement it in every driver 🎉 

With this change, we're now throwing an exception if there are two properties with different names mapped to the same field in the database and both are not marked as `notSaved`. If they share the same property name we assume the original mapping is overwritten, while if either of them is marked as `notSaved` we also don't throw a mapping error since we won't have any conflicts due to the mapping. The sorted group relationships in the user document in our tests shows where this can be useful.

To allow a mapping references and embedded relationships as `notSaved`, I had to drop the `NotSaved` annotation and move it to the `AbstractField` annotation to allow references and embedded documents being marked as readonly. I'll need to come up with some more tests to ensure this works consistently, so I might extract that change to a separate pull request.

UPGRADE-2.0.md
```diff
@@ -150,6 +150,9 @@ or annotations. To migrate away from YAML mappings, first update to MongoDB ODM
   `BinMD5`, `BinUUID`, `BinUUIDRFC4122`, `Bool`, `Boolean`, `Collection`,
   `Date`, `Float`, `Hash`, `Increment`, `Int`, `Integer`, `Key`, `ObjectId`,
   `Raw`, `String`, `Timestamp`.
+* The `NotSaved` annotation has been dropped in favor of the `notSaved`
+  attribute on the `Field` annotation. The `notSaved` attribute can also be
+  applied to reference and embed mappings.
```
jmikola: Will users encounter an error if they forget to remove the NotSaved annotation? I assume so, as the class itself will not be found.

Just want to make sure folks won't accidentally shoot themselves in the foot if they forget to migrate over to the `notSaved` attribute. An explicit failure is preferable to silently writing to the field, of course.

docs/en/reference/annotations-reference.rst
```diff
@@ -354,6 +360,9 @@ Optional attributes:
    ``nullable`` - By default, ODM will ``$unset`` fields in MongoDB if the PHP
    value is null. Specify true for this option to force ODM to store a null
    value in the database instead of unsetting the field.
+-
+   ``notSaved`` - The property is loaded if it exists in the database; however,
+   ODM will not save the property value back to the database.
```
jmikola: Should the documentation for all `notSaved` attributes indicate that it defaults to `false`?

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -1940,6 +1940,29 @@ public function mapField(array $mapping) : array
 
         $this->applyStorageStrategy($mapping);
 
+        $isNotSaved = $mapping['notSaved'] ?? false;
```
jmikola: Small nit, but would `$isSaved = ! ($mapping['notSaved'] ?? false);` be clearer? That would avoid the double negation for the `if` condition below.

tests/Doctrine/ODM/MongoDB/Tests/Mapping/xml/Doctrine.ODM.MongoDB.Tests.Mapping.AbstractMappingDriverDuplicateDatabaseNameNotSaved.dcm.xml
```diff
@@ -0,0 +1,14 @@
...
+<doctrine-mongo-mapping xmlns="http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping"
+                  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                  xsi:schemaLocation="http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping
+                  http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping.xsd">
+
+    <document name="Doctrine\ODM\MongoDB\Tests\Mapping\AbstractMappingDriverDuplicateDatabaseNameNotSaved">
+        <field field-name="override" type="int" />
+
+        <field field-name="foo" type="string" name="baz" />
+        <field field-name="bar" type="string" name="baz" not-saved="true" />
```
jmikola: Does `not-saved` need to be added to the XSD?

alcaeus: In addition to that, the XML mapping driver refuses to load mapping files that don't validate against the schema 🎉 

This was introduced in 2.0 and will generate notices/deprecation warnings in 1.3 to help people update.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
alcaeus: That's a good point, I think that's better than `if (! $notSaved)` 👍

docs/en/reference/annotations-reference.rst
alcaeus: I'll add a little more documentation for it in the `@Field` annotation and also document it in the "migrating schemas" documentation.

PR #1905, opened 2018-11-27
Don't dump to stdout in query command

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

The query command uses the Symfony VarDumper component to dump query output to the user. By default however, the dumper dumps this to stdout. This PR uses a custom dumper to instead dump to the console output.

This also adds missing documentation for the BC break introduced in #1896 (the `--depth` option was dropped).

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Tools/Console/Command/QueryCommand.php
+++ b/lib/Doctrine/ODM/MongoDB/Tools/Console/Command/QueryCommand.php
@@ -8,7 +8,8 @@ use LogicException;
 use Symfony\Component\Console;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputOption;
-use Symfony\Component\VarDumper\VarDumper;
+use Symfony\Component\VarDumper\Cloner\VarCloner;
+use Symfony\Component\VarDumper\Dumper\CliDumper;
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ODM/MongoDB/Tools/Console/Command/QueryCommand.php
```diff
@@ -89,8 +91,13 @@ protected function execute(Console\Input\InputInterface $input, Console\Output\O
...
         }
 
+        $cloner = new VarCloner();
+        $dumper = new CliDumper(static function (string $payload) use ($output) : void {
+            $output->write($payload);
+        });
+
         foreach ($qb->getQuery() as $result) {
-            VarDumper::dump($result);
+            $dumper->dump($cloner->cloneVar($result));
```
malarzm: I might have missed reasoning on chat yesterday, but what's the purpose of cloning result before dumping it?

alcaeus: In our case, it isn't entirely necessary, I copied the logic from the VarDumper component. In that case, the instance is cloned to not modify the original instance while dumping (e.g. initialising lazy collections).

PR #1908, opened 2018-12-01
Optimizations for documents loading

watari:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes/no
| Fixed issues | 

#### Summary
I am profiled library with XHProf and find out next:
1. Class metadata extraction from metadata factory is slow even when it is already loaded to process memory. 
2. Hydrators can work more efficiently if their logic will be slightly changed.

For speeding up metadata extraction I added field to `DocumentManager` in which all previously extracted metadata will be cached with class name as index. This approach allow me to hugely increase documents population speed.
For speeding up hydrators work I added to them fields for storing directly inside reflection fields, field mappings and persistent collection factory. Also I slightly updated conditions for checking if field should be populated (this not bring much but perfectionist inside of me was satisfied by that:) ).

As a result of my manipulations ODM in "documents reading mode" is on 45-50% faster than it was previously. Hydrators  now approximately on 10% faster (here major role played storing required information inside hydrator). 

In PR all changes that was made by me. Also you can check it by yourself - I created [repository](https://github.com/watari/mongodb-odm-benchmarks) with command. In it `composer.json.opt` contain dependency for fork with my changes.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/DocumentManager.php
+++ b/lib/Doctrine/ODM/MongoDB/DocumentManager.php
@@ -144,6 +144,9 @@ class DocumentManager implements ObjectManager
     /** @var ClassNameResolver */
     private $classNameResolver;
 
+    /** @var ClassMetadata[]  */
+    private $classMetadataCache = [];
+
     /**
      * Creates a new Document that operates on the given Mongo connection
      * and uses the given Configuration.
@@ -279,7 +282,11 @@ class DocumentManager implements ObjectManager
      */
     public function getClassMetadata($className) : ClassMetadata
     {
-        return $this->metadataFactory->getMetadataFor($className);
+        if (empty($this->classMetadataCache[$className])) {
+            $this->classMetadataCache[$className] = $this->metadataFactory->getMetadataFor($className);
+        }
+
+        return $this->classMetadataCache[$className];
     }
 
     /**
--- a/lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
+++ b/lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
@@ -205,7 +205,7 @@ EOF
         if (isset(\$data['%1\$s'])) {
             \$value = \$data['%1\$s'];
             %3\$s
-            \$this->class->reflFields['%2\$s']->setValue(\$document, clone \$return);
+            \$reflFields['%2\$s']->setValue(\$document, clone \$return);
             \$hydratedData['%2\$s'] = \$return;
         }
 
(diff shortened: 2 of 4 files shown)


alcaeus:
@watari I'll keep this PR open until after the release of Beta-1: I want to take some time to write benchmarks using phpbench to prevent future regressions. In this case, the caching of metadata seems useless since metadata factory already does this, so I'd like to figure out what's causing the performance issues.

watari:
@alcaeus , Factory indeed perform caching. But before this cache can be used, each time class name resolver called ([here](https://github.com/doctrine/mongodb-odm/blob/a7cccd72137ab1d054a53a6b6e789f92c1102cda/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataFactory.php#L61)). It seems to me that this service is to heavy.
You can see it by yourself in reports provided by XHProf
Caching on `DocumentManager` level prevent calling of `ClassNameResolver::getRealClass` each time

alcaeus:
Just an update on this: I've created #2023 to add some performance benchmarks. I can definitely see some regressions in 2.0, even from using `ext-mongodb` with `mongo-php-adapter` in 1.x, which definitely shouldn't happen. I'll test these changes against the benchmarks and will expand the suite soon.

@watari maybe you could walk me through the changes to hydrators some time? I'm not sure I understand why they make a performance difference, but then I may just be blind there. Thanks!

watari:
@alcaeus. My changes it is some sort of "micro optimizations". I did next things:
1. Reduce amount of references for data. Fields mappings and reflection information now is extracted in constructor of hydrator from `ClassMetadata` to corresponding fields. During each call of `hydrate` method used local copy of this information. In this way we a saving some time since we make less "references" for getting needed data. For example, `$this->class->fieldMappings['field']` make 3 "references" for data when line `$fieldMappings['field']` requires only 1 "reference". 
2. Reduce methods calling by extraction of factory to hydrator corresponding field. 

In code that is under "high load" this changes is make difference.

lib/Doctrine/ODM/MongoDB/DocumentManager.php
```diff
@@ -279,7 +282,11 @@ public function getClassNameResolver() : ClassNameResolver
      */
     public function getClassMetadata($className) : ClassMetadata
     {
-        return $this->metadataFactory->getMetadataFor($className);
+        if (empty($this->classMetadataCache[$className])) {
```
alcaeus: This change can be removed as the root cause for the performance regression has been covered in #2069.

alcaeus:
After some more investigation and trial-and-error, I wasn't able to consistently measure any performance improvements with the other changes. I considered applying changes to the `nullable` association logic in hydrators, but I don't think it makes the code more legible enough to warrant changing this.

@watari I want to thank you for taking the time to analysing the performance issues you've encountered and for creating the PR. In the future, it should hopefully be easier to measure improvements now that we have a limited set of benchmark tests. We can always expand on that test suite when encountering specific issues to better measure the impact of some changes.

I'm closing this PR in favour of #2069 which fixes the performance regression with proxy name resolution. If you can produce benchmarks that show a significant impact of the other changes, please create a PR for those.

Issue #841, opened 2014-04-01
Split index options between SchemaManager and ClassMetadata


jmikola:
Index options specific to the creation of indexes should be handled exclusively by SchemaManager and the related index commands. This includes:
- `background`
- `dropDups`
- `safe` (deprecated in favor of `w`)
- `timeout` (possibly deprecated in favor of `socketTimeoutMS`; see [PHP-1045](https://jira.mongodb.org/browse/PHP-1045)).
- `w`

Other index options that relate to the type/properties may be considered part of the schema and do belong in metadata (e.g. annotations, XML, YAML):
- `default_language` (for text indexes)
- `expireAfterSeconds` (for TTL indexes)
- `language_override` (for text indexes)
- `name`
- `order` (not a real option, but used for field-level mappings)
- `sparse`
- `unique`
- `weights` (for text indexes; not easily supported in XML)

For reference:
- http://docs.mongodb.org/manual/reference/method/db.collection.ensureIndex/
- http://php.net/manual/en/mongocollection.ensureindex.php
- https://jira.mongodb.org/browse/PHP-1045

jmikola:
Bumping this from 1.0.0, since it's not a trival refactoring. When implementing this in a minor version, we can then deprecate the execution options (e.g. `background`, `w`) from being specified in the class metadata. For BC, we should still have SchemaManager read those options; however, commands should also take these options at runtime.

There's room for debate on this, but I think the commands that create indexes in ODM should have their own defaults (e.g. `{w:1}`) and any execution options specified on the metadata should supersede those defaults. If there are options specified at runtime (e.g. console command parameters), those should supersede everything.

jacobchristensen79:
Jmikola, I have searched for a while how to define in XML the "weight" index property. DO you have the link for the doc or a sample? Thanks a lot.

jmikola:
I only alluded to the [`weights` option](https://docs.mongodb.com/manual/reference/method/db.collection.createIndex/#options-for-text-indexes) in my original post for completeness' sake. There is no reference to it in [`XMLDriver::loadMetadataForClass()`](https://github.com/doctrine/mongodb-odm/blob/1.1.0/lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php#L48) and I believe it still unsupported in XML mappings. This is currently supported in annotations, YAML, and PHP mappings because they're very liberal in what they accept for options.

[...]

PR #1910, opened 2018-12-19
Separate index creation options from index options

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | #841 

#### Summary

* Dropped the `dropDups` option for indexes since MongoDB no longer uses it
* Dropped the `timeout` command line option for schema commands
* Added `maxTimeMS` and `writeConcern` options to schema manager
* Exposed `maxTimeMS` as well as `w`, `wTimeout` and `journal` options to schema commands
* Changed index equivalency checks to consider indexes with different options as different unless we handle them separately. This fixes issues when using new index options that aren't known to MongoDB ODM as we wouldn't recreate indexes when those options change.

The last change might create issues down the line if new options are exposed in the `indexInfo` document from a `getIndexes` call that are missing from the mapped options. We'll have to consider this.


lib/Doctrine/ODM/MongoDB/Tools/Console/Command/Schema/AbstractCommand.php
```diff
@@ -50,4 +64,24 @@ protected function getMetadataFactory()
...
+        return $maxTimeMs !== null ? (int) $maxTimeMs : null;
+    }
+
+    protected function getWriteConcernFromInput(InputInterface $input) : ?WriteConcern
+    {
+        $w        = $input->getOption('w');
+        $wTimeout = $input->getOption('wTimeout');
+        $journal  = $input->getOption('journal');
+
+        if ($w === null && $wTimeout === null && $journal === null) {
```
alcaeus: @jmikola we may want to revisit this check, not sure if it's appropriate. Does it make sense to just pass `wTimeout` or `journal` options to the command without passing a `w` option, and if so, what default should we assume?

alcaeus:
Note: I extracted index equivalency checks from this PR to a separate branch and will be opening a PR for that shortly. I feel those changes warrant their own discussion.

lib/Doctrine/ODM/MongoDB/Tools/Console/Command/Schema/AbstractCommand.php
```diff
@@ -50,4 +65,26 @@ protected function getMetadataFactory()
...
+        if (! is_string($w)) {
+            return null;
+        }
+
+        $wTimeout = $input->getOption('wTimeout');
+        $wTimeout = is_string($wTimeout) ? (int) $wTimeout : 0;
+        $journal  = $input->getOption('journal');
+        $journal  = is_string($journal) ? (bool) $journal : false;
+
+        return new WriteConcern($w, $wTimeout, $journal);
```
jmikola: If you look at [`WriteConcern::__construct()`](https://github.com/mongodb/mongo-php-driver/blob/1.6.0alpha1/src/MongoDB/WriteConcern.c#L156), the first argument is parsed as an arbitrary zval and we inspect its type when creating the underlying write concern struct. If an integer is provided, it becomes the numeric mode you expect. Any strings are converted to either the special "majority" mode or a tag set (see: [`w` docs](https://docs.mongodb.com/manual/reference/write-concern/#w-option) in the MongoDB manual).

As-is, I think this function might inadvertently create a WriteConcern with a numeric string as its tag set, which will likely produce runtime errors since the user won't have a custom tag set defined on the replica set for something like "1" or "2".

Two options here:

 * Inspect the "w" value (e.g. [`ctype_digit`](http://php.net/manual/en/function.ctype-digit.php) and decide to cast it to an integer. This means that users have no way to enter a tag set that may be consist of only digit characters (an edge case I don't think we need to care about until such a user opens a bug report).
 * Split command line options so we solicit numeric and string "w" values differently. This covers all our bases, but it also pollutes the API a bit and is solving a problem before we have a legitimate use case.

I'll note that libmongoc actually supports `w = -1`; however, we don't [document that](http://php.net/manual/en/mongodb-driver-writeconcern.construct.php) and I don't think you need to handle it here. Effectively, it's worse than `w = 0` in that it also ignores all socket errors.

Lastly, to your question in Slack: it is possible the have a write concern with just `j` and no `w` and `wtimeout` values; however, that's only via URI options (connection string and/or array). When constructing a WriteConcern object, you're correct in that the driver does require `w`. I might actually suggest throwing errors if the user doesn't provide `w` rather than silently ignore them.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -307,34 +301,38 @@ public function createDocumentCollection(string $documentName) : void
...
+            $this->dm->getDocumentDatabase($documentName)->createCollection($class->getBucketName() . '.files', $options);
+            $this->dm->getDocumentDatabase($documentName)->createCollection($class->getBucketName() . '.chunks', $options);
 
             return;
         }
 
+        $options = [
+            'capped' => $class->getCollectionCapped(),
+            'size' => $class->getCollectionSize(),
+            'max' => $class->getCollectionMax(),
```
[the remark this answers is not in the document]
alcaeus: That is correct.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -343,34 +341,36 @@ public function dropCollections() : void
...
+        $options = $this->getWriteOptions($maxTimeMs, $writeConcern);
+
+        $this->dm->getDocumentCollection($documentName)->drop($options);
 
         if (! $class->isFile) {
             return;
         }
 
-        $this->dm->getDocumentBucket($documentName)->getChunksCollection()->drop();
+        $this->dm->getDocumentBucket($documentName)->getChunksCollection()->drop($options);
```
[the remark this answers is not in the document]
alcaeus: Correct, see https://github.com/doctrine/mongodb-odm/blob/8c0c5da468e324b474b4aca70c4766f01433aebc/lib/Doctrine/ODM/MongoDB/DocumentManager.php#L332.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -652,4 +646,19 @@ private function collectionIsSharded(string $documentName) : bool
         $stats = $database->command(['collstats' => $class->getCollection()])->toArray()[0];
         return (bool) ($stats['sharded'] ?? false);
     }
+
+    private function getWriteOptions(?int $maxTimeMs = null, ?WriteConcern $writeConcern = null, array $options = []) : array
+    {
+        unset($options['maxTimeMs'], $options['writeConcern']);
```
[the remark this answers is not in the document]
alcaeus: This was done to ignore any creation options users may have set in the options for an index.

lib/Doctrine/ODM/MongoDB/Tools/Console/Command/Schema/AbstractCommand.php
alcaeus: * Using `wTimeout` or `journal` without specifying a `w` option will now cause an exception to be thrown
* The `w` option is inspected using `is_numeric` and cast to `int` to fix treating `--w=2` as tag options. I avoided using `ctype_digit` as the `ctype` extension apparently needs to be installed separately on FreeBSD. I think `is_numeric` will work just fine, please let me know if you think otherwise.

jmikola: Interesting note about `ctype` not being included on FreeBSD. I would have assumed it's simple enough where it'd be standard (no reason it can't be implemented in pure PHP really). `is_numeric()` sounds fine. It has some edge cases where floats or exponents might be cast to an integer, but those are about as likely to be actual custom write concerns as a straight digit.

PR #1911, opened 2018-12-20
Rework index equivalency checks

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #841 

#### Summary

This was extracted from #1910 to allow for a more in-depth discussion. This PR reworks index equivalency checks. Previously, when encountering an unknown index option, we would always consider the indexes equivalent even though they weren't. This PR reverses this behaviour and considers indexes different if they have a different set of options. We exclude some options from this check (e.g. `unique`) to consider indexes equivalent when one specifies the option with the default value while the other one doesn't specify the option at all.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/SchemaManager.php
+++ b/lib/Doctrine/ODM/MongoDB/SchemaManager.php
@@ -395,81 +395,16 @@ final class SchemaManager
         $this->dm->getDocumentDatabase($documentName)->drop($this->getWriteOptions($maxTimeMs, $writeConcern));
     }
 
-    /**
-     * Determine if an index returned by MongoCollection::getIndexInfo() can be
-     * considered equivalent to an index in class metadata.
-     *
-     * Indexes are considered different if:
-     *
-     *   (a) Key/direction pairs differ or are not in the same order
-     *   (b) Sparse or unique options differ
-     *   (c) Geospatial options differ (bits, max, min)
-     *   (d) The partialFilterExpression differs
-     *
-     * The background option is only relevant to index creation and is not
-     * considered.
-     *
-     * @param array|IndexInfo $mongoIndex Mongo index data.
-     */
-    public function isMongoIndexEquivalentToDocumentIndex($mongoIndex, array $documentIndex) : bool
+    public function isMongoIndexEquivalentToDocumentIndex(IndexInfo $mongoIndex, array $documentIndex) : bool
     {
-        $documentIndexOptions = $documentIndex['options'];
-
-        if (! $this->isEquivalentIndexKeys($mongoIndex, $documentIndex)) {
-            return false;
-        }
-
-        if (empty($mongoIndex['sparse']) xor empty($documentIndexOptions['sparse'])) {
-            return false;
-        }
-
-        if (empty($mongoIndex['unique']) xor empty($documentIndexOptions['unique'])) {
-            return false;
-        }
-
-        foreach (['bits', 'max', 'min'] as $option) {
-            if (isset($mongoIndex[$option]) xor isset($documentIndexOptions[$option])) {
-                return false;
-            }
-
-            if (isset($mongoIndex[$option], $documentIndexOptions[$option]) &&
-                $mongoIndex[$option] !== $documentIndexOptions[$option]) {
-                return false;
-            }
-        }
-
-        if (empty($mongoIndex['partialFilterExpression']) xor empty($documentIndexOptions['partialFilterExpression'])) {
-            return false;
-        }
-
-        if (isset($mongoIndex['partialFilterExpression'], $documentIndexOptions['partialFilterExpression']) &&
-            $mongoIndex['partialFilterExpression'] !== $documentIndexOptions['partialFilterExpression']) {
-            return false;
-        }
-
-        if (isset($mongoIndex['weights']) && ! $this->isEquivalentTextIndexWeights($mongoIndex, $documentIndex)) {
-            return false;
-        }
-
-        foreach (['default_language', 'language_override', 'textIndexVersion'] as $option) {
-            /* Text indexes will always report defaults for these options, so
-             * only compare if we have explicit values in the document index. */
-            if (isset($mongoIndex[$option], $documentIndexOptions[$option]) &&
-                $mongoIndex[$option] !== $documentIndexOptions[$option]) {
-                return false;
-            }
-        }
-
-        return true;
+        return $this->isEquivalentIndexKeys($mongoIndex, $documentIndex) && $this->isEquivalentIndexOptions($mongoIndex, $documentIndex);
     }
 
     /**
      * Determine if the keys for a MongoDB index can be considered equivalent to
      * those for an index in class metadata.
-     *
-     * @param array|IndexInfo $mongoIndex Mongo index data.
      */
-    private function isEquivalentIndexKeys($mongoIndex, array $documentIndex) : bool
+    private function isEquivalentIndexKeys(IndexInfo $mongoIndex, array $documentIndex) : bool
     {
         $mongoIndexKeys    = $mongoIndex['key'];
         $documentIndexKeys = $documentIndex['keys'];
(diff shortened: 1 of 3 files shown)


lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -403,54 +448,51 @@ public function dropDocumentDatabase(string $documentName, ?int $maxTimeMs = nul
...
 
         foreach (['bits', 'max', 'min'] as $option) {
-            if (isset($mongoIndex[$option]) xor isset($documentIndexOptions[$option])) {
-                return false;
-            }
-
-            if (isset($mongoIndex[$option], $documentIndexOptions[$option]) &&
-                $mongoIndex[$option] !== $documentIndexOptions[$option]) {
+            if (isset($mongoIndexOptions[$option], $documentIndexOptions[$option]) &&
+                $mongoIndexOptions[$option] !== $documentIndexOptions[$option]) {
```
jmikola: This has the same logic as the `foreach` loop for text index options below. Should we combine them?

I see the logic of "isset(mongo option, document option) && mongo option != document option)" is also repeated elsewhere for `partialFilterExpression`. Perhaps we should just factor this out into a method and call it on single list of such options?

I'm not sure if the comment for text indexes also applies in other cases, with the server always reporting defaults for these options and that being the reason we only compare if we have explicit values in the document index.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -403,54 +448,51 @@ public function dropDocumentDatabase(string $documentName, ?int $maxTimeMs = nul
      *
      * The background option is only relevant to index creation and is not
      * considered.
-     *
-     * @param array|IndexInfo $mongoIndex Mongo index data.
      */
-    public function isMongoIndexEquivalentToDocumentIndex($mongoIndex, array $documentIndex) : bool
+    private function isEquivalentIndexOptions(IndexInfo $mongoIndex, array $documentIndex) : bool
```
jmikola: I forget, but does ODM allow arbitrary index options to be specified? If so, I assume that `indexOptionsAreMissing()` is the catch-all case for options that don't receive special comparison logic in this method.

I noted the comment above that `background` is not considered by this method but am curious if the addition of `indexOptionsAreMissing()` might break that old behavior. IIRC, `background` is only specified when the index is first created (i.e. ODM's options) but wouldn't show up when querying the database for index specs (i.e. IndexInfo). If so, we might need to add it to the list of ignored missing options.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
alcaeus: Yep, it probably makes sense to extract this to a common method 👍

lib/Doctrine/ODM/MongoDB/SchemaManager.php
alcaeus: I'll add tests for that use-case, specifically checking whether the `background` option is returned when fetching index information after creating an index with the `background` option set.

alcaeus: The comment about the `background` option not being considered was obsolete and left over from the behaviour in 1.x. This exception was removed as part of this refactoring, but I'm not sure if we're right in excluding it. Creating an index with the option set will also expose it when listing index info:
```
> db.indexTest.createIndex({foo: 1}, {background: true})
{
	"createdCollectionAutomatically" : false,
	"numIndexesBefore" : 1,
	"numIndexesAfter" : 2,
	"ok" : 1
}
> db.indexTest.getIndexes()
[
        ...,
	{
		"v" : 2,
		"key" : {
			"foo" : 1
		},
		"name" : "foo_1",
		"ns" : "test.indexTest",
		"background" : true
	}
]
```

While it would most certainly be better to always create indexes with correct options, changing the `background` option after the fact would cause an index to be dropped and recreated to have the correct `background` option. I'm not sure if that's feasible.

One possible solution would be to add an `index-background` command option to the `odm:schema:create` command as a creation option. If set, this would ensure that all indexes created through the command will have the `background` option set, regardless of the `background` option in the index. We could easily add this to 1.3 without causing a BC break.

From then on, we can decide whether we still want to allow specifying the `background` option on an index and how it should behave. Given that we're trying to wrap up 2.0 soon, I'd rather not make any significant BC breaks to this logic.

jmikola: Let's consider deprecating the `background` option in ODM 2.x if/when MongoDB server does away with the notion of background index builds.

In the meantime, we can continue to ignore `background` during index comparisons and add a basic "enable" flag to turn on background index builds when creating indexes from the the CLI command. We don't need a CLI option to force-disable background. That is to say, if a model class has `background: true` in its index spec, the CLI command will not be able to override that to a foreground index build.

alcaeus:
@jmikola I've added the `background` option to the index creation methods in SchemaManager and propagated it to the schema creation command. Since we're not deprecating the `background` index option yet, I've decided against adding it to 1.3 as it's a lot of extra work due to the additional refactoring done to SchemaManager in 2.0.

lib/Doctrine/ODM/MongoDB/Tools/Console/Command/Schema/CreateCommand.php
```diff
@@ -32,6 +32,7 @@ protected function configure()
             ->addOption('class', 'c', InputOption::VALUE_REQUIRED, 'Document class to process (default: all classes)')
             ->addOption(self::COLLECTION, null, InputOption::VALUE_NONE, 'Create collections')
             ->addOption(self::INDEX, null, InputOption::VALUE_NONE, 'Create indexes')
+            ->addOption('background', null, InputOption::VALUE_NONE, sprintf('Create indexes in background (requires "%s" option)', self::INDEX))
```
jmikola: Is the requirement enforced, or is this just a way of saying we'll ignore the option unless `self::INDEX` is also specified?

lib/Doctrine/ODM/MongoDB/Tools/Console/Command/Schema/CreateCommand.php
```diff
@@ -44,17 +45,18 @@ protected function execute(InputInterface $input, OutputInterface $output)
         // Default to the full creation order if no options were specified
         $create = empty($create) ? $this->createOrder : $create;
 
-        $class = $input->getOption('class');
+        $class      = $input->getOption('class');
+        $background = (bool) $input->getOption('background');
```
jmikola: Do you think there's value in respecting an unset `background` option and relying on the server default? I expect that would entail changing various type arguments above to `?bool $background` and adding logic to not stuff it into options arrays if null.

[SERVER-37270](https://jira.mongodb.org/browse/SERVER-37270) appears to be the JIRA issue where MongoDB 4.2 will ignore `background: true` (presumably `background: false` was always ignored). Going forward, all index builds will use a hybrid path, which makes the `background` option obsolete. This is what I was alluding to in https://github.com/doctrine/mongodb-odm/pull/1911/files#r264858582, although I didn't have the JIRA link at the time. For that reason, I think omitting the option by default makes the most sense.

If that's seems too complicated, I'm fine with passing `['background' => false]` by default, since it will achieve the same outcome.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -625,13 +649,13 @@ private function runShardCollectionCommand(string $documentName, ?WriteConcern $
         )->toArray()[0];
     }
 
-    private function ensureGridFSIndexes(ClassMetadata $class, ?int $maxTimeMs = null, ?WriteConcern $writeConcern = null) : void
+    private function ensureGridFSIndexes(ClassMetadata $class, ?int $maxTimeMs = null, ?WriteConcern $writeConcern = null, bool $background = false) : void
```
jmikola: Just to confirm, ODM users never specify index options for these GridFS indexes (like they do for normal indexes) so `background` isn't getting merged into anything here other than `unique` for the files index. Correct?

lib/Doctrine/ODM/MongoDB/Tools/Console/Command/Schema/CreateCommand.php
alcaeus: The latter - we'll only use the `background` option if indexes are being created.

lib/Doctrine/ODM/MongoDB/Tools/Console/Command/Schema/CreateCommand.php
alcaeus: With bool options there is no possibility to pass a value - the value is controlled by the option being passed or not. Thus, if we wanted to add a third state to respect the server's default value, we'd have to split this into a `background` and `no-background` option. Passing neither option would then always force server default, passing `background` would force background builds, passing `no-background` would pass `['background' => false]` and setting both would cause an error.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
alcaeus: Correct - users can't change options for these indexes at all. The only reason they're created here (as opposed to letting the library create them) is that I wanted to create all indexes for the schema. Hence the duplication of logic from the MongoDB library.

alcaeus:
Merging without updating background logic for now - we can always revisit that in a 2.x release once MongoDB Server starts treating the `background` option differently.

Issue #1864, opened 2018-09-21
PersistentCollection instance lost raw data during serialization/deserialization process.


watari:
### Bug Report

|    Q        |   A
|------------ | ------
| Version     | 1.2.5

#### Summary

By default instance keep raw data from the database. So if you not accesses items from this collection before serialization, they will be not created from this raw data. This will cause lost of all data about items in collection during serialization/deserialization process. 
I solved this problem for itself by redefining PersistentCollection class - I added call of 'initialize' method in '_sleep'.

#### How to reproduce

1. Create instance of PersistentCollection with raw data.
2. Serialize instance.
3. Unserialize instance.
4. At this point data will be lost.

malarzm:
Sorry this fell through the cracks, I'll close this issue to keep the conversation in one place since you've opened a PR :)

PR #1914, opened 2018-12-27
Fix raw mongo data loss during serialization.

watari:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #1864 

#### Summary

Added `initialize` method call in `PersistentCollection::__sleep` method for avoiding raw data loss.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionTrait.php
+++ b/lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionTrait.php
@@ -574,6 +574,7 @@ trait PersistentCollectionTrait
      */
     public function __sleep()
     {
+        $this->initialize();
         return ['coll', 'initialized'];
     }
 
--- a/tests/Doctrine/ODM/MongoDB/Tests/PersistentCollectionTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/PersistentCollectionTest.php
@@ -11,8 +11,10 @@ use Doctrine\ODM\MongoDB\DocumentManager;
 use Doctrine\ODM\MongoDB\Mapping\ClassMetadata;
 use Doctrine\ODM\MongoDB\PersistentCollection;
 use Doctrine\ODM\MongoDB\UnitOfWork;
+use Documents\Phonenumber;
 use Documents\User;
 use stdClass;
+use function count;
 use function serialize;
 use function unserialize;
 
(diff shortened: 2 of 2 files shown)


malarzm:
I'd say that if anything, we should serialize raw data to avoid initialization. Or PersistentCollection should be wise enough to refetch raw data if it's not initialized and `raw` is `null`. But I also think that `PersistentCollection` shouldn't really be serialized. What's the use case?

watari:
In my case is results caching serialization is performed. After deserialization we do not have references to UoW and cannot perform hydration. I choosed just to initialize collections before serialization will be performed. 
I think that missing data fething is bad idea since in this case we must make additional queries to database.
`PersistentCollection` created inside of document by Doctrine and we cannot replace it on something else transparently. We can redefine for each parent document `__sleep` method, but then there will be place for error, since developer can simply forget about redefinition.
Data loss during serialization was surprise for me when I in first time discovered it. It is not obvious behavior as for me)

malarzm:
I think that writing `__sleep` for documents that will be serialized is better that initializing each and every collection that gets serialized. As for forgetting to add a collection there, it can be covered with a fairly simple unit test :)

Also I'm not sure but auto-initializing collection may lead to tons of data being serialized (A has collection of Bs, B has collection of C, C has references to A and so on - everything will now be serialized due to auto-initialization) which is not something user would be expecting either.

watari:
> I think that writing __sleep for documents that will be serialized is better  that initializing each and every collection that gets serialized.

As for me it is expected behavior by default.
`PersistentCollection` implement `Collection` and `PersistentCollectionInterface`. `PersistentCollectionInterface` is marked as `internal` so we must pretend that we always deal with instances of `Collection`. `Collection` interface also implement `ArrayCollection` which you recommend to use for nested collections in newly created entities. So we pretend that `ArrayCollection` is everywhere. Instances of `ArrayCollection` does not loss anything when serialization is performed. So we must expect same from instances of `PersistentCollection` since it hidden under same interface (`Collection`).
If raw data is lost during serialization - we get corrupted document when it will be unserialized. This is not good.
I think that is better by default initialize every collection and keep all data. 
I do not propose force initialization of every collection when it was extracted from database. I propose doing this only when collection is serialized. 

About redefining of documents `__sleep` method. It is good when you want to customize document serialization. In case of persistent collection empty collection does not make any sense. I think that in this case better explicitly exclude unneeded collections from serialized document.

My point is next: serialize everything by default, if you want to exclude something from serialized document - exclude it explicitly.

Anyway, maybe my "expected behavior" is not indeed "expected" :)

malarzm:
> PersistentCollectionInterface is marked as internal so we must pretend that we always deal with instances of Collection.

`@internal` on the interface is there so we can add methods to the interface when needed, we won't remove anything from the interface without a major release. Since ODM is not fully transparent and has its quirks, I'd encourage people to check against `PersistentCollection` and `Proxy` in scenarios they know something might be lazy loaded and it shouldn't or vice versa, load something lazily). I do this all the time :)

> Anyway, maybe my "expected behavior" is not indeed "expected" :) What do you think?

Since you've asked, I think you shouldn't be serializing a document in the first place. You should explicitly transform it into `CachedDocument` and cache that, so there are no false assumptions after the object is unserialized. Also you control exactly what (and how) you want to have cached so you won't end up having vast object graph serialized without asking for it. If you really want to have a document serialized, use a decent serializer like JMS' or Symfony's one - they'll handle persistent collections just fine and prolly will work around issues with plain `serialize` I don't even know about.

alcaeus:
Closing here: initialising the collection before serialisation is not the way to fix this. I've reopened #1864 and will be addressing it in a future 1.2 release.

watari:
You are rigth that this is not solution of problem. I forgot about references - we do not use them, so I did not count them when I come with this solution :). References can realy increase amount of objects that must be serialized. 
BTW, this issue affect all versions of library, not only 1.2.x branch.

alcaeus:
Yes, the issue affects 1.2 and master - we’ll fix it in 1.2 and then cherry-pick to master.

PR #1917, opened 2019-01-03
Fix IDE warning about store-as deprecation warning

etshy:
Fix IDE warning about store-as deprecation warning

|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

"Fixing" the deprecation warning in PHPStorm on `store-as` attribute (instead of `simple` attribute) as discussed in #1895

### DIFF
--- a/doctrine-mongo-mapping.xsd
+++ b/doctrine-mongo-mapping.xsd
@@ -158,7 +158,8 @@
     <xs:attribute name="target-document" type="xs:string" />
     <xs:attribute name="field" type="xs:NMTOKEN" use="required" />
     <xs:attribute name="fieldName" type="xs:NMTOKEN" />
-    <xs:attribute name="simple" type="xs:boolean" /><!-- deprecated -->
+    <!-- deprecated -->
+    <xs:attribute name="simple" type="xs:boolean" />
     <xs:attribute name="store-as" type="odm:reference-store-as" default="dbRefWithDb" />
     <xs:attribute name="inversed-by" type="xs:NMTOKEN" />
     <xs:attribute name="mapped-by" type="xs:NMTOKEN" />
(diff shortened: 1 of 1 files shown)


malarzm:
I think it'd be good if we merge it to 1.2.x first and up from there, we still maintain that branch given 1.3 is not released yet

malarzm:
> I always do git rebase -i 1.2.x in a case like that, then drop all commits that are not supposed to be in there.

You can also use `git rebase --onto` :)

etshy:
I just did the changes quickly on github directly, so I don't know if I can rebase it from 1.2 (wthout to have to clone the repo locally).

Is it possible directly from github ?

malarzm:
There's no rebase via github as far as I know so the quickest will be closing this and doing change again in 1.2.x branch

greg0ire:
BTW, you should amend your commit message, you used the same sentence for both the subject and the body.

Please format your commit message according to these rules : 

- The commit message subject must be less than 50 characters and tell us *what* you did.
- The commit message body should tell us *why* you did it. It is optional but highly recommended.

Bad example :

```
Fixed bug #989 by removing call to defraculate()
```

Good example:

```
Remove call to defraculate()

Calling this function caused a bug because it interferes 
with calls to getSchmeckles().
Fixes #989
```

More details in CONTRIBUTING.md

etshy:
OK I'll redo a new PR when I could then.
I'll quickly try to add a message in the deprecated comment, to see if phpStorm can show it, too.

I'll make sure to write a good commit title/body.
I close this PR since it will not be merged

greg0ire:
> OK I'll redo a new PR when I could then.

You did not have to, you can use `git commit --amend`, `git push --force` and edit the target of the branch

Issue #1916, opened 2018-12-31
Schema commands error out with 'The "maxTimeMs" option does not exist.'


Rybbow:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.0.0-beta1

#### Summary

When trying to execute `doctrine:mongodb:schema:create`, `doctrine:mongodb:schema:update` or `doctrine:mongodb:schema:drop` from DoctrineMongoDBBundle an error 'The "maxTimeMs" option does not exist.' is printed out and the command fails.

#### Current behavior

Changes introduced in [this commit](https://github.com/doctrine/mongodb-odm/commit/0204d4079f1ace40a54a7b6c5755ae69db4c3809#diff-3c235a7aedeb6ba2bf96208a6e295210) introduced common options for schema related commands in `\Doctrine\ODM\MongoDB\Tools\Console\Command\Schema\AbstractCommand::configure()`. However, since `parent::configure()` call is missing from the `configure()` method in all of `\Doctrine\ODM\MongoDB\Tools\Console\Command\Schema\UpdateCommand`, `\Doctrine\ODM\MongoDB\Tools\Console\Command\Schema\CreateCommand` and `\Doctrine\ODM\MongoDB\Tools\Console\Command\Schema\DropCommand`, the options are not present and trying to use them causes an error.

#### How to reproduce

Install and configure `doctrine/mongodb-odm-bundle` in `dev-master` version in any Symfony project and execute any execute one of the mentioned commands.

#### Expected behavior

The command should not fail.

I guess adding the `parent::configure()` call to `AbstractCommand` will fix the problem. I can prepare a PR if that's the only required fix here. Also, I don't think `\Doctrine\ODM\MongoDB\Tools\Console\Command\Schema\ValidateCommand` or `\Doctrine\ODM\MongoDB\Tools\Console\Command\Schema\ShardCommand` use the new common schema command options anywhere and they don't really make much sense for them, so they should not be changed.

alcaeus:
Thanks for the report. I'll prepare a fix shortly 👍

Rybbow:
I can prepare a fix, it's not a problem. Just let me know if the solution I suggested is ok.

alcaeus:
[...]

PR #1918, opened 2019-01-03
Fix incorrect schema commands behaviour

Rybbow:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #1916
#### Summary

Fixes for broken schema commands, that did not use `parent::configure()` call, which is needed to steup all command options. Changed also `ShardCommand` and `SchemaManager` to allow using writeConcern when sharding.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/SchemaManager.php
+++ b/lib/Doctrine/ODM/MongoDB/SchemaManager.php
@@ -526,7 +526,7 @@ class SchemaManager
      *
      * @throws MongoDBException
      */
-    public function ensureSharding() : void
+    public function ensureSharding(?WriteConcern $writeConcern = null) : void
     {
         foreach ($this->metadataFactory->getAllMetadata() as $class) {
             assert($class instanceof ClassMetadata);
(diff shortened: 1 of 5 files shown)


lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -596,7 +596,8 @@ private function runShardCollectionCommand(string $documentName) : array
             [
                 'shardCollection' => $dbName . '.' . $class->getCollection(),
                 'key'             => $shardKey['keys'],
-            ]
+            ],
+            $this->getWriteOptions(null, $writeConcern)
```
alcaeus: The `$options` argument to `command` refers to generic command option, which only accepts a `readPreference` and `session` option. In this case, you have to just merge the array returned by `getWriteOptions` into the actual command.

Rybbow: Ah, sorry, fixed!

PR #1922, opened 2019-01-08
Fix wrong usage of sequence vs. choice in XSD

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

#### Summary

This fixes two wrong types in the mapping XSD that required people to write child elements in a certain order. This is inconsistent with the rest of the schema file.

### DIFF
--- a/doctrine-mongo-mapping.xsd
+++ b/doctrine-mongo-mapping.xsd
@@ -37,7 +37,7 @@
   </xs:complexType>
 
   <xs:complexType name="gridfs-file">
-    <xs:sequence>
+    <xs:choice minOccurs="0" maxOccurs="unbounded">
       <xs:element name="id" type="odm:id" minOccurs="0" maxOccurs="1"/>
 
       <xs:element name="length" type="odm:gridfs-length-field" minOccurs="0" maxOccurs="1" />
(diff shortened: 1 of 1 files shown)


doctrine-mongo-mapping.xsd
```diff
@@ -152,11 +152,11 @@
   </xs:complexType>
 
   <xs:complexType name="gridfs-metadata-field">
-    <xs:sequence>
+    <xs:choice minOccurs="0" maxOccurs="unbounded">
       <xs:element name="discriminator-field" type="odm:discriminator-field" minOccurs="0"/>
       <xs:element name="discriminator-map" type="odm:discriminator-map" minOccurs="0"/>
       <xs:element name="default-discriminator-value" type="odm:default-discriminator-value" minOccurs="0"/>
```
malarzm: Should we have `maxOccurs="1"` on elements?

alcaeus: Oh, yeah, that's a good point 👍

alcaeus: According to PhpStorm's highlighting, `minOccurs="1" maxOccurs="1"` is the default. However, even when `maxOccurs` is present, it's possible to add more than one element. I'll still add it for the sake of completeness.

alcaeus: So `minOccurs="1" maxOccurs="1"` is the default value. I've cleaned up the file in this regard. However, I couldn't get a file to break with this, probably because we've got these wrapped in a `choice`. Thus, while I can specify that only one `read-preference` element can be provided, there can be multiple `choice` elements, of which each can be a `read-preference` element. I'm not sure how to solve this in XSD land.

malarzm: probably there's nothing we can do with this. Anyways good to know about both things :)

alcaeus: I might play around with that some more in the future. For now, we're good.

PR #1733, opened 2018-01-12
[2.0] Merge ClassMetadataInfo into ClassMetadata

carusogabriel:
This closes #1727 by merge `ClassMetadataInfo` into `ClassMetadata` as [suggested](https://github.com/doctrine/mongodb-odm/pull/1714#issuecomment-356524003) by @alcaeus in #1714

PR #1924, opened 2019-01-09
Add deprecation layers

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

Note: PR is best reviewed commit-by-commit.

This PR brings in a first round of deprecations:
* #1734: Deprecates the YAML metadata drivers
* #1807: Add the `id` element to the mapping schema and deprecate passing `id="true"` to a `field` element
* #1812: Deprecate the `namespace` property and its accessor and trigger a deprecation warning when passing non-FQCN
* #1733: Move logic from `ClassMetadataInfo` to `ClassMetadata`, deprecate `ClassMetadataInfo` and provide an alias
* #1814: Validate XML mapping files against a schema and trigger a deprecation warning when passing an invalid mapping file
* #1834: Deprecate `writeConcern` and `fieldName` attributes in schema in favour of `write-concern` and `field-name`

I might have to revisit the BC layer for `ClassMetadata` once again. @greg0ire has tried the same for ORM in https://github.com/doctrine/orm/pull/6886. What I'd like to do: trigger a deprecation notice when someone uses `ClassMetadataInfo` but allow using `ClassMetadata` in place of it. Using `class_alias` in `ClassMetadata.php` means that we can't trigger a deprecation notice. We might have to split this out of this PR.


greg0ire:
See also: https://dev.to/greg0ire/how-to-deprecate-a-type-in-php-48cf

lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
```diff
@@ -135,6 +139,37 @@ public function loadMetadataForClass($className, ClassMetadata $class)
...
         }
+
+        if (isset($xmlRoot->id)) {
+            $field   = $xmlRoot->id;
+            $mapping = [
+                'id' => true,
+                'fieldName' => 'id',
+            ];
+
+            /** @var SimpleXMLElement $attributes */
```
greg0ire: Use `assert`?

lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
```diff
@@ -503,7 +546,7 @@ private function setShardKey(ClassMetadataInfo $class, \SimpleXmlElement $xmlSha
      *
      * list($readPreference, $tags) = $this->transformReadPreference($xml->{read-preference});
      *
-     * @param \SimpleXMLElement $xmlReadPreference
+     * @param SimpleXMLElement $xmlReadPreference
```
greg0ire: Why not use a type hint instead?

malarzm:
I'll take your word for `ClassMetadata` stuff, rest looks ok to me except for invalid schema that fails some tests :)

lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
alcaeus: This has been fixed in 2.0 and I'm not keen on introducing this in 1.x. We can't use PHPStan since it's PHP 5 code, so I'd rather just let this die peacefully.

lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
alcaeus: Don't know, I've changed it 👍

alcaeus:
@greg0ire updated code with your suggestions and changed `ClassMetadataInfo` deprecation to follow the advice in your blog post. Could you take another look and see if you spot anything that seems wrong?

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataInfo.php
```diff
@@ -19,2230 +19,23 @@
...
-     */
-    public function isIdGeneratorAuto()
-    {
-        return $this->generatorType == self::GENERATOR_TYPE_AUTO;
-    }
-
-    /**
-     * Checks whether the class will use a collection to generate incremented identifiers.
-     *
-     * @return boolean TRUE if the class uses the INCREMENT
```
greg0ire: More importantly, this makes the code compatible with `composer --classmap-authoritative`, which is mostly use in production systems. Please add a comment stating that so that we do not remove this if IDEs add support for `class_alias` (they might already)

greg0ire: The `extends ClassMetadata` is the part that is important for IDEs.

greg0ire: Use en_US spelling here? "favor" over "favour"?

alcaeus: Whoops 😉

PR #1929, opened 2019-01-14
Fix serialisation of uninitialised PersistentCollection instances

### LINKED ISSUE #1864: PersistentCollection instance lost raw data during serialization/deserialization process.

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | Fixes #1864, Replaces #1914

#### Summary

This ensures that any internal data from `PersistentCollection` instances are serialised to have a usable collection instance after serialisation. This affects the `mongoData` and `snapshot` properties along with `isDirty` and `hints`.

Ping @watari

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionTrait.php
+++ b/lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionTrait.php
@@ -574,15 +574,15 @@ trait PersistentCollectionTrait
     }
 
     /**
-     * Called by PHP when this collection is serialized. Ensures that only the
-     * elements are properly serialized.
+     * Called by PHP when this collection is serialized. Ensures that the
+     * internal state of the collection can be reproduced after serialization
      *
      * @internal Tried to implement Serializable first but that did not work well
      *           with circular references. This solution seems simpler and works well.
      */
     public function __sleep()
     {
-        return array('coll', 'initialized');
+        return array('coll', 'initialized', 'mongoData', 'snapshot', 'isDirty', 'hints');
     }
 
     /* ArrayAccess implementation */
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionTrait.php
```diff
@@ -582,7 +582,7 @@ public function slice($offset, $length = null)
      */
     public function __sleep()
     {
-        return array('coll', 'initialized');
+        return array('coll', 'initialized', 'mongoData', 'snapshot', 'isDirty', 'hints');
```
malarzm: Please update PHPDoc of the method as it's wrong now

malarzm:
I still think that using unserialised `PersistentCollection` is a can of worms but this is better than #1914 :)

lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionTrait.php
alcaeus: Fixed!

watari:
Hi there. Sorry for delay. First of all - thank you for PR with fix :)
But still `PersistentCollection` is unusable after serialization since It does not have link to UoW and cannot perform hydration of raw data. I agree with @malarzm , that `PersistentCollection` should not be serialized since it will be broken after unserialization.

I propose just to forget about this issue until it do not brake anything :) 

About my PR - it was bad in general and I did not count  references. In such way I fixed it for my projects and it was worked for me.
I think that this issue will be really actual when implementation of results caching in lib will begin.

malarzm:
@watari for the record, unserialised collection will have `UnitOfWork` if you call [`->setDocumentManager()`](https://github.com/doctrine/mongodb-odm/blob/9aae6d067ba6d8b16b922c78c72cfee6c54df4f0/lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionTrait.php#L95)

PR #1937, opened 2019-01-18
Removed PHP 5.5 workaround

juliusstoerrle:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | no

#### Summary

This line is not required anymore as we now have PHP 7.2 as min. version.
I did not test this apart from making sure the file can still be parsed. Not sure if that is required or how to accomplish that.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -1013,9 +1013,7 @@ class DocumentPersister
             }
 
             // No further preparation unless we're dealing with a simple reference
-            // We can't have expressions in empty() with PHP < 5.5, so store it in a variable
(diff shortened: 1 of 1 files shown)


malarzm:
> I did not test this apart from making sure the file can still be parsed. Not sure if that is required or how to accomplish that.

Existing tests will tell us 👍

alcaeus:
Thanks @juliusstoerrle!

MatTheCat:
`empty` would return `true` if `$value` is `null` or an empty array so why is it casted to an array? 🤔

malarzm:
Can't really tell why we did the cast in the first place but for instance if value would happen to be an empty `stdClass` it wouldn't work: https://3v4l.org/hh8Fj

MatTheCat:
Indeed thanks!

Issue #1696, opened 2017-12-21
[2.0] Document BC breaks and add deprecations in 1.3


alcaeus:
For every BC break we made in 2.0, there needs to be a deprecation warning and (if possible/sensible) a forward layer for people to prepare switching to ODM 2.0. This affects, but is not limited to BC breaks in #1553.

malarzm:
FWIW I was using labels for PRs breaking BC: https://github.com/doctrine/mongodb-odm/issues?utf8=%E2%9C%93&q=label%3A%22BC+break%22+milestone%3A2.0.0

alcaeus:
BC breaks have been documented in the `UPGRADE` document. Now we need to add BC layers (deprecations, possibly backports of new classes, etc.) in 1.3. All PRs that need this have been tagged with the [needs-1.3-compat-layer](https://github.com/doctrine/mongodb-odm/issues?utf8=✓&q=label%3Aneeds-1.3-compat-layer+) tag.

alcaeus:
Closing as current BC breaks have been documented in 1.3.

PR #1938, opened 2019-01-18
Renamed scheduledForDirtyCheck

juliusstoerrle:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes?
| Fixed issues | no, but TODO in code

#### Summary

Saw this @todo in the code and thought I might aswell just fix it.
These names are now the same as in the ORM.

I guess this is a BC break if people used the UOW directly, should this change be included in the Upgrade/Changelog? (#1696)


juliusstoerrle:
Oh, Thanks. I totally did not see those. Very good tip with the global .gitignore!

PR #1942, opened 2019-01-22
Finish renaming dirtyCheck methods

alcaeus:
Finishes work started in #1938. ORM may not have renamed that method yet, but it makes sense to keep method names similar instead of having multiple names for the same functionality.


alcaeus:
Ah, yes. Will do 👍

Issue #1943, opened 2019-01-23
GenerateProxiesCommand/GenerateHydratorsCommand: assert is_string on an array


thiver:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | dev-master

#### Summary

`$filter = $input->getOption('filter');`
`assert(is_string($filter));`

$filter returns an array because of InputOption::VALUE_IS_ARRAY, so the assertion fails always.

Class Methods:

Doctrine\ODM\MongoDB\Tools\Console\Command\GenerateProxiesCommand::execute
Doctrine\ODM\MongoDB\Tools\Console\Command\GenerateHydratorsCommand::execute

#### Current behavior

The commands cannot be executed because of the assertion fail.

#### How to reproduce

Run the GenerateProxiesCommand or GenerateHydratorsCommand command.

#### Expected behavior

The $filter should be asserted as an array.

malarzm:
Sounds about right, @thiver do you mind making a PR to fix the bug?

alcaeus:
🤦‍♂️ great find @thiver!

thiver:
No problem, I will create a PR for it.

malarzm:
Closing as #1944 was just merged

PR #1944, opened 2019-01-24
fix: replace string assertion with array assertion for $filter

thiver:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #1943

#### Summary

The $filter input option from the generate commands should be asserted as array instead of string.


malarzm:
@thiver thanks for the fix! Could you please squash your commits or do you prefer us doing so?

malarzm:
Thanks a lot @thiver!

PR #1947, opened 2019-02-01
Remove usage of deprecated ClassMetadataInfo class in hydrators

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

This removes a leftover usage of the `ClassMetadataInfo` in generated hydrator code.


malarzm:
@alcaeus There's something wrong with the code, all tests are failing

lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
```diff
@@ -369,7 +369,7 @@ private function generateHydratorClass(ClassMetadata $class, $hydratorClassName,
 use Doctrine\ODM\MongoDB\Hydrator\HydratorInterface;
 use Doctrine\ODM\MongoDB\Query\Query;
 use Doctrine\ODM\MongoDB\UnitOfWork;
-use Doctrine\ODM\MongoDB\Mapping\ClassMetadataInfo;
+use Doctrine\ODM\MongoDB\Mapping\ClassMetadata;
```
malarzm: This line is duplicated (it is just above the fold)

greg0ire: Consider sorting the use statements to avoid this.

alcaeus: That's done in master but we don't want to apply doctrine/cs in 1.x. I could also just run tests locally to avoid this, but this is a classic "this won't break" scenario 😂

PR #1950, opened 2019-02-04
[1.3] Deprecate GridFS mappings

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | deprecation
| BC Break     | no
| Fixed issues | 

#### Summary

This deprecates the current GridFS mapping to prepare users for the BC break in 2.0. A forward compatible layer can't be provided due to the differences in the underlying API (and because the work to do so would be unreasonable for the benefit provided).

### DIFF
--- a/UPGRADE-1.3.md
+++ b/UPGRADE-1.3.md
@@ -1,5 +1,13 @@
 # UPGRADE FROM 1.2 TO 1.3
 
+## GridFS
+
+GridFS support will be rewritten for 2.0 and was deprecated in its current form.
+A forward compatible layer cannot be provided due to differences in the
+underlying API.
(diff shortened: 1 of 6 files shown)


UPGRADE-1.3.md
```diff
@@ -1,5 +1,13 @@
 # UPGRADE FROM 1.2 TO 1.3
 
+## GridFS
+
+GridFS support will be rewritten for 2.0 and was deprecated in its current form.
```
malarzm: we tend to use "has been deprecated" across the file

alcaeus: Will fix 👍

alcaeus: Upon further evaluation, we use "was deprecated" a bunch of times. I'm not sure which is more correct, maybe @jmikola can help with that?

I'll merge this and fix it in a separate PR.

PR #1954, opened 2019-02-12
Document autoload requirement for generated proxies

### LINKED ISSUE #1920: Generating proxy classes only when file not exists
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | ?
| Version     | ?

#### Summary

The `Doctrine\ODM\MongoDB\Configuration::AUTOGENERATE_FILE_NOT_EXISTS` configuration option should prevent the proxy generator from generating a given proxy file if it already exists, but actually the proxy file is regenerated on each request.

#### Current behavior

Even though `Configuration::setAutoGenerateProxyClasses()` is called with `Configuration::AUTOGENERATE_FILE_NOT_EXISTS` the proxy generator regenerates each affected proxy file once per request.

#### How to reproduce

Make sure document proxies are generated. Then in a separate process use `DocumentManager::getReference()` with a document that is not in the identity map - it will causes proxies to be regenerated.

#### Expected behavior

Generated proxy file should not change?

If you take a look [here](https://github.com/doctrine/mongodb-odm/blob/master/lib/Doctrine/ODM/MongoDB/Configuration.php#L202) at what this configuration option in fact does when set, it turns out it simply changes proxy generator strategy. And inspecting the set [FileWriterGeneratorStrategy](https://github.com/Ocramius/ProxyManager/blob/master/src/ProxyManager/GeneratorStrategy/FileWriterGeneratorStrategy.php) and the [FileLocator](https://github.com/Ocramius/ProxyManager/blob/master/src/ProxyManager/FileLocator/FileLocator.php) it uses, it can be seen that there is no `file_exists` check in place (and it seems it has never existed, which means this never worked?).

I guess a custom decorator for `\ProxyManager\GeneratorStrategy\GeneratorStrategyInterface`, that would check the existence of the file and wrap the original strategy would solve the issue. However, this would probably mean that proxies are no longer automatically regenerated for dev/debug environments in projects using this library. Which would mean that proxies do not contain up-to-date metadata for document classes that were edited until they are manually regenerated or unless some alternative tracking method is introduced. 

What should be the course of action here?

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | Fixes #1920

### DIFF
--- a/docs/en/reference/introduction.rst
+++ b/docs/en/reference/introduction.rst
@@ -406,6 +406,15 @@ their mapping.
     $config->setHydratorNamespace('Hydrators');
     $config->setDefaultDB('doctrine_odm');
 
+    spl_autoload_register($config->getProxyManagerConfiguration()->getProxyAutoloader());
+
+.. note::
+
+    The last call to ``spl_autoload_regiter`` is necessary to autoload generated
+    proxy classes. Without this, the proxy library would re-generate proxy
+    classes for every request. See the `tuning for production`_ chapter in
+    ProxyManager's documentation.
+
 The easiest way to define mappings for our document classes is with annotations.
 We'll need to specify an annotation driver in our configuration (with one or
 more paths) and register the annotations for the driver:
@@ -496,3 +505,4 @@ please create separate clients for your application and ODM.
 .. _MongoDB: https://www.mongodb.com/
 .. _10gen: http://www.10gen.com
 .. _Composer: http://getcomposer.org/
+.. _tuning for production: https://ocramius.github.io/ProxyManager/docs/tuning-for-production.html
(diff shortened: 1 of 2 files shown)


docs/en/reference/introduction.rst
```diff
@@ -406,6 +406,14 @@ their mapping.
     $config->setHydratorNamespace('Hydrators');
     $config->setDefaultDB('doctrine_odm');
 
+    spl_autoload_register($config->getProxyManagerConfiguration()->getProxyAutoloader());
+
+.. note::
+
+    The last call to ``spl_autoload_regiter`` is necessary to autoload generated
+    proxy classes. Without this, the proxy library would re-generate proxy
+    classes for every request. See the `tuning for production`_ documentation.
```
malarzm: Should we mention that the link leads outside of our docs like "See the tuning for production chapter in `ProxyManager`'s documentation"?

PR #1956, opened 2019-02-14
Remove previously deprecated operations from Query Builder

jmikola:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes/no
| Fixed issues | 

#### Summary

The `TYPE_GEO_LOCATION` constant was already marked for removal in a comment. I assume this was an outstanding todo item.

While revising the `Builder::limit()` comment, I also realized that the ODM 2.0 Query Builder no longer supports `group` and `mapReduce` commands. I didn't find any open issues for implementing them, so I assume they were both intentionally dropped when porting over code from `doctrine/mongodb`. Let me know if you'd like to remove those as well and I can update this PR and create an entry in `CHANGELOG-2.0.md`.

As you know, [`group`](https://docs.mongodb.com/manual/reference/command/group/) has been deprecated since MongoDB 3.4. `mapReduce` is not yet deprecated, but I expect it will be once Javascript support is integrated into the aggregation pipeline (no ETA on that).


alcaeus:
> I didn't find any open issues for implementing them, so I assume they were both intentionally dropped when porting over code from `doctrine/mongodb`.

Correct, we dropped those for the reasons you mentioned below: `group` being deprecated and `mapReduce` facing the same fate soon(ish).

Double-checking I also just realised that we haven't documented these removals at all. Feel free to add that to this PR.

jmikola:
@alcaeus: I think this is ready for another look. `geoNear`'s removal was already documented, so I only added a changelog bullet for `group` and `mapReduce`. I also found a bunch of remnants in the public API, which were removed.

alcaeus:
This looks good. I've created #2001 to add the corresponding deprecations for 1.3. Since the large refactoring in the Query and QueryBuilder classes will cause conflicts either way, I'm also perfectly fine merging this PR to 2.0 before merge #2001 to 1.3. Since we didn't document all deprecations properly in the code when starting 2.0, there will be more instances like this where we add deprecations to 1.3 after dropping the corresponding code in 2.0.

malarzm:
FWIW I've just merged #2001 :)

Issue #1931, opened 2019-01-15
Mark internal classes as @internal in PHPDocs


alcaeus:
Mark classes that are internal and not covered by the BC promise as `@internal` via PHPDoc to let users know that they shouldn't rely on their API.

Method signatures may be changed in minor releases.

malarzm:
I think we could also add `final` to some classes like `CollectionPersister` or such

alcaeus:
Absolutely. We can prepare that and mark them as `@final` in 1.3 along with deprecation notices if they're extended.

malarzm:
Closing as I've just opened #1984 to mark classes as final which was last thing to do in this issue (PR has a needs compat label)

PR #1960, opened 2019-02-19
Mark methods and classes as internal

malarzm:
See #1931 for details. I've added a note that we can also remove internal stuff in minor releases as I saw a number of methods that should not be added in the first place and I thought that we could remove them in 2.1 (such as `HydratorFactory::setUnitOfWork` which should be passed in ctor).

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
+++ b/lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
@@ -107,6 +107,8 @@ class HydratorFactory
 
     /**
      * Sets the UnitOfWork instance.
+     *
+     * @internal
      */
     public function setUnitOfWork(UnitOfWork $uow) : void
     {
--- a/lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionInterface.php
+++ b/lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionInterface.php
@@ -71,14 +71,12 @@ interface PersistentCollectionInterface extends Collection
     public function setDirty($dirty);
 
     /**
-     * INTERNAL:
      * Sets the collection's owning entity together with the AssociationMapping that
      * describes the association between the owner and the elements of the collection.
      */
     public function setOwner(object $document, array $mapping);
 
     /**
-     * INTERNAL:
      * Tells this collection to take a snapshot of its current state reindexing
      * itself numerically if using save strategy that is enforcing BSON array.
      * Reindexing is safe as snapshot is taken only after synchronizing collection
@@ -87,14 +85,12 @@ interface PersistentCollectionInterface extends Collection
     public function takeSnapshot();
 
     /**
-     * INTERNAL:
      * Clears the internal snapshot information and sets isDirty to true if the collection
      * has elements.
      */
     public function clearSnapshot();
 
     /**
-     * INTERNAL:
      * Returns the last snapshot of the elements in the collection.
      *
      * @return array The last snapshot of the elements.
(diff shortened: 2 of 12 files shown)


malarzm:
This does not include adding `final` as it'll be easier to provide a deprecation notices for 1.3 basing on PR doing only `final`.

lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionInterface.php
```diff
@@ -71,14 +71,12 @@ public function isDirty();
     public function setDirty($dirty);
 
     /**
-     * INTERNAL:
```
malarzm: Entire interface is marked as internal

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -966,8 +974,6 @@ private function computeAssociationChanges(object $parentDocument, array $assoc,
      * whereby changes detected in this method prevail.
      *
      * @throws InvalidArgumentException If the passed document is not MANAGED.
-     *
-     * @ignore
```
malarzm: This method is advertised in documentation so it makes it pretty much not internal

lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
```diff
@@ -107,6 +107,8 @@ public function __construct(DocumentManager $dm, EventManager $evm, ?string $hyd
 
     /**
      * Sets the UnitOfWork instance.
+     *
+     * @internal
      */
     public function setUnitOfWork(UnitOfWork $uow) : void
```
alcaeus: Where is this used? Since `HydratorFactory` receives a `DocumentManager` instance it should be able to fetch the UoW from there 🤔

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AnnotationDriver.php
```diff
@@ -25,6 +25,8 @@
 
 /**
  * The AnnotationDriver reads the mapping metadata from docblock annotations.
+ *
+ * @internal
```
alcaeus: Not sure about marking the drivers as internal classes: they need to be set up manually unless you're using the Symfony Bundle, so they should be covered by the BC promise.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -1145,6 +1151,8 @@ private function executeDeletions(ClassMetadata $class, array $documents, array
      * identity map.
      *
      * @throws InvalidArgumentException
+     *
+     * @internal
```
alcaeus: Agree about making the `scheduleFor...` methods internal, but disagree about `isScheduledFor...`: these can be useful in event subscribers to check document state.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -2471,12 +2523,11 @@ public function getClassNameForAssociation(array $mapping, $data) : string
     }
 
     /**
-     * INTERNAL:
      * Creates a document. Used for reconstitution of documents during hydration.
      *
      * @internal Highly performance-sensitive method.
      *
-     * @ignore
+     * @internal
```
alcaeus: This method is already marked as internal, but I feel we should just drop the comment ;)

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -2707,6 +2770,8 @@ public function propertyChanged($document, $propertyName, $oldValue, $newValue)
 
     /**
      * Gets the currently scheduled document insertions in this UnitOfWork.
+     *
+     * @internal
```
alcaeus: The `getScheduledDocument...` methods are useful in event subscribers, so we shouldn't mark them as internal.

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AnnotationDriver.php
malarzm: Good point. I wasn't sure either and thought it's better to add and remove than to skip :)

lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
alcaeus: Ah. Ugly. Makes sense then, let's try to clean those interfaces up in 3.0.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
malarzm: `getScheduledDocument...` are no longer internal however `getScheduledCollection...` (and related `is...`) still are as I feel that's implementation detail of the ODM

alcaeus: I remember using those as well - let me try to refresh my memory. I remember there were some special things needed to be done when calling `clear` on a persistent collection as it entails a collection deletion followed by it being newly persisted. I'm not sure if changesets give enough information, could be that subscribers need to query UoW for information on collection operations as well.

PR #1962, opened 2019-02-21
Fix wrong usage of discriminator map in complex document inheritance chains

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | Maybe
| Fixed issues | 

#### Summary

When having multiple levels of inheritance in your document structure, discriminator maps are inherited incorrectly. They are currently inherited from their parent class and cannot be overridden. As highlighted in the attached tests, fetching all documents from the `GH1962BarSuperclass` repository should only yield subclasses of that superclass but also returns an unrelated `GH1962FooDocument`, which should not happen.

In my opinion, discriminator maps should be overridable in a child class (as showcased in the test). Optionally, me may want to invert inheritance completely, where a discriminator map is not passed on to child classes but to parent classes; that way, a class will always contain all possible classes in its discriminator map.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataInfo.php
@@ -784,6 +784,10 @@ class ClassMetadataInfo implements \Doctrine\Common\Persistence\Mapping\ClassMet
      */
     public function setDiscriminatorMap(array $map)
     {
+        $this->subClasses = [];
+        $this->discriminatorMap = [];
+        $this->discriminatorValue = null;
+
         foreach ($map as $value => $className) {
             if (strpos($className, '\\') === false && strlen($this->namespace)) {
                 $className = $this->namespace . '\\' . $className;
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1962Test.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1962Test.php
@@ -1,0 +1,81 @@
+<?php
+
+namespace Doctrine\ODM\MongoDB\Tests\Functional\Ticket;
+
+use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
+use Doctrine\ODM\MongoDB\Tests\BaseTest;
+
+class GH1962Test extends BaseTest
+{
+    public function testDiscriminatorMaps()
+    {
+        $metadata = $this->dm->getClassMetadata(GH1962Superclass::class);
+        self::assertCount(3, $metadata->discriminatorMap);
+
+        $metadata = $this->dm->getClassMetadata(GH1962BarSuperclass::class);
+        self::assertCount(2, $metadata->discriminatorMap);
+    }
+
+    public function testFetchingDiscriminatedDocuments()
+    {
+        $foo = new GH1962FooDocument();
+        $bar = new GH1962BarDocument();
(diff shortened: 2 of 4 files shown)


tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1962Test.php
```diff
@@ -0,0 +1,81 @@
...
+        $baz = new GH1962BazDocument();
+
+        $this->dm->persist($foo);
+        $this->dm->persist($bar);
+        $this->dm->persist($baz);
+
+        $this->dm->flush();
+        $this->dm->clear();
+
+        self::assertCount(3, $this->dm->getRepository(GH1962Superclass::class)->findAll());
```
malarzm: How come this is working? `@MappedSuperclass`es are not query-able. Or is it true only for queries through  `Query`?

alcaeus: I'd have to take another look at that. I can also reproduce this with a document, the problem stays the same. I'll check query builder for mapped superclasses later 👍

alcaeus: Took another look, there's no difference between mapped superclasses and documents. The only issue I found was the one fixed here, where having a class without a discriminator value in the inheritance chain (like a mapped superclass) causes the DocumentPersister to not add the discriminator field to the query. For the root class this may be fine, but when you have an inheritance chain like the one in the added test case, this obviously leads to wrong results.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataInfo.php
```diff
@@ -784,6 +784,10 @@ public function setDiscriminatorField($discriminatorField)
      */
     public function setDiscriminatorMap(array $map)
     {
+        $this->subClasses = [];
```
ostrolucky: Hey @alcaeus do you remember why this diff chunk was applied? It's causing some issues. You can see when just doing `return $preparedQuery;` in `\Doctrine\ODM\MongoDB\Persisters\DocumentPersister::addDiscriminatorToPreparedQuery` and running `GH1962Test`

alcaeus: The entire inheritance of discriminator maps from parent classes is wrong: before this patch, the full discriminator map from the parent class was applied to child classes. The problem is highlighted in `GH1962Test`, where a mapped superclass is supposed to have a different discriminator map than its parents class to correctly reflect class inheritance.

In this case, the discriminator map will contain values that shouldn't be in there, since there was no way to completely override a discriminator map (values are always added). Since discriminator maps don't have to be complete (e.g. contain all possible values) in 1.x, the `subClasses` array in a child class would contain classes that are only in there based on the wrong discriminator map. Maybe it's overkill, but I thought it best to reset all values involved in discriminator maps for safety when setting a new map.

If you found an issue with the code, please highlight it instead of hinting at it. Thanks.

ostrolucky: Problem is that if query returns more results than expected (record with discriminator value missing in discriminator map), this leads to obscure error being produced (`Doctrine\Common\Persistence\Mapping\MappingException : Class 'foo' does not exist`). This shouldn't be a problem in practice after this PR, since it guarantees such situations don't happen, but is a problem when working on tweaking logic for inserting discriminator values into query.

PR #1966, opened 2019-03-05
Fix embedding documents containing named indexes

### LINKED ISSUE #1344: Indexing embedded document in different fields fails due duplicate name
Hello,

I'll go right to the point. When embedding a document class in two different fields and attempting to set an index on that document class, index cannot be created due to name conflict:

```
Trying to create an index with same name test with different key spec { property2.property_sub: 1 } vs existing spec { property1.property_sub: 1 }
```

Here are two classes that allow to reproduce the error.  
Main class goes as follows:

``` php
<?php

namespace AppBundle\Document\Test;

/**
 * @ODM\Document
 */
class TestMain
{
    /**
     * @ODM\Id
     */
    private $id;

    /**
     * @ODM\EmbedMany(targetDocument="TestSub")
     */
    private $property1;

    /**
     * @ODM\EmbedMany(targetDocument="TestSub")
     */
    private $property2;
}
```

Embedded class:

``` php
<?php

namespace AppBundle\Document\Test;

/**
 * @ODM\Document
 * @ODM\Index(keys = {"property_sub" = "asc"}, name = "test")
 */
class TestSub
{
    /**
     * @ODM\Id
     */
    private $id;

    /**
     * @ODM\Field(type="string")
     */
    private $property_sub;

    /**
     * @ODM\Field(type="string")
     */
    private $property_sub2;
}
```

I've tested this a little, and issue goes away if:
- name attribute for index is not used
- document is embedded just once

In my case, I'm bound to use `name` because document is embedded deep in main document tree - otherwise I receive error regarding maximum index name length, which is already nicely mentioned in the docs.

As far as I'm concerned, I will probably create indexes manually, since I'm about to use a document in a "standalone" collection or embedded depending on specific application state, so specifying indexes in metadata won't really work for me (I actually do not want those deep indexes, only on main collection to guarantee uniqueness of specific data sets).  
However, I believe it is not an expected behaviour, so I'd like to share my case.

Thanks in advance, and thanks for a great library :)

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | fixes #1344 

#### Summary

This fixes an issue when embedding a document containing a named index multiple times. Since the index had a name, this caused an exception with MongoDB since the index would already exist when creating it for the second embed use (see test case). With this change we're prefixing indexes pulled up from embedded documents with the name of the property that's embedding them. While this can cause exceptions when an index name becomes too long, the current solution gives more control to the user.

One option would be to add an `indexPrefix` option to embedded relationships and using that instead of the field name. Since that would be a new feature, I will not include it in the bugfix for 1.2.x.

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1344Test.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1344Test.php
@@ -1,0 +1,57 @@
+<?php
+
+namespace Doctrine\ODM\MongoDB\Tests\Functional\Ticket;
+
+use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
+use Doctrine\ODM\MongoDB\Tests\BaseTest;
+
+class GH1344Test extends BaseTest
+{
+    public function testGeneratingIndexesDoesNotThrowException()
+    {
+        $indexes = $this->dm->getSchemaManager()->getDocumentIndexes(GH1344Main::class);
+        self::assertCount(4, $indexes);
+        self::assertSame('embedded1_embedded', $indexes[0]['options']['name']);
+        self::assertSame('embedded1_embedded_nested', $indexes[1]['options']['name']);
+        self::assertSame('embedded2_embedded', $indexes[2]['options']['name']);
+        self::assertSame('embedded2_embedded_nested', $indexes[3]['options']['name']);
+
+        $this->dm->getSchemaManager()->ensureDocumentIndexes(GH1344Main::class);
+    }
+}
+
+/** @ODM\Document */
+class GH1344Main
+{
+    /** @ODM\Id */
+    public $id;
+
+    /** @ODM\EmbedOne(targetDocument=GH1344Embedded::class) */
+    public $embedded1;
+
+    /** @ODM\EmbedOne(targetDocument=GH1344Embedded::class) */
+    public $embedded2;
+}
+
+/**
+ * @ODM\EmbeddedDocument
+ * @ODM\Index(keys={"property"="asc"}, name="embedded")
+ */
+class GH1344Embedded
+{
+    /** @ODM\Field */
+    public $property;
+
+    /** @ODM\EmbedOne(targetDocument=GH1344EmbeddedNested::class) */
+    public $embedded;
+}
+
+/**
+ * @ODM\EmbeddedDocument
+ * @ODM\Index(keys={"property"="asc"}, name="nested")
+ */
+class GH1344EmbeddedNested
+{
+    /** @ODM\Field */
+    public $property;
+}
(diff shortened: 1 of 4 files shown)


tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1344Test.php
```diff
@@ -0,0 +1,57 @@
...
+    public function testGeneratingIndexesDoesNotThrowException()
+    {
+        $indexes = $this->dm->getSchemaManager()->getDocumentIndexes(GH1344Main::class);
+        self::assertCount(4, $indexes);
+        self::assertSame('embedded1_embedded', $indexes[0]['options']['name']);
+        self::assertSame('embedded1_embedded_nested', $indexes[1]['options']['name']);
+        self::assertSame('embedded2_embedded', $indexes[2]['options']['name']);
+        self::assertSame('embedded2_embedded_nested', $indexes[3]['options']['name']);
+
+        $this->dm->getSchemaManager()->ensureDocumentIndexes(GH1344Main::class);
```
malarzm: What's the purpose of ensuring indexes here?

alcaeus: Just making sure that the indexes can be created successfully.

alcaeus:
> could you please add it and note that indexes on embedded documents will have a prefix?

Yeah, prefix should be documented regardless of previous docs. I'll add a chapter later today.

alcaeus:
Documentation added, mind taking another look @malarzm?

Steveb-p:
From what I remember maximum index name length in MongoDB should be 125 characters. It isn't specified in MongoDB docs, but is mentioned multiple times in MongoDB issue tracker (https://jira.mongodb.org/browse/DOCS-1820 for example) and on some other sites. I'd say mention of this limit might be added to that documentation note?

ODM might also check for index name length in the future?

alcaeus:
> From what I remember maximum index name length in MongoDB should be 125 characters. It isn't specified in MongoDB docs

The [index name length documentation](https://docs.mongodb.com/manual/reference/limits) explains this in more detail:
> Fully qualified index names, which includes the namespace and the dot separators (i.e. <database name>.<collection name>.$<index name>), cannot be longer than 128 characters.

We can check this at schema creation, but I believe any attempt to truncate index names can just cause issues down the road, so I prefer if the original exception was shown to the user. As mentioned previously, I believe it’s more sensible allowing users to configure this index prefix in future versions than trying to add more magic to the prefix.

alcaeus:
> It'd be great to add `name` to the list of optional attributes

I somehow completely missed the fact that the name wasn’t in the options. This should indeed be properly documented. 👍

Steveb-p:
My intention would not be to truncate or otherwise transform the index name, but rather prevent the attempt to create such an index or wrap exception with Doctrine's own one with some helpful message about how it can be resolved, and maybe what field it was about.

But that is wayyy out of scope :) Thank you @alcaeus

alcaeus:
@malarzm added a link to the indexes chapter to the annotation mapping, as the options are explained in more detail there. I've also added the note about index prefixing there so it won't get lost.

@Steveb-p let's fix the first issue - the reason I'm not really keen on doing too much with the exception is because either way, the user will be presented with a difficult problem: they have to either rename the index, the fields contained in the index path (as they are prefixed to the name), the collection name or even the database name.

Even without our exception, the error message provided by MongoDB will contain all necessary information:
> namespace name generated from index name "doctrine_odm_tests.GH1344TooLongIndexName.$embedded1_this_is_a_really_long_name_that_will_cause_problems_for_whoever_tries_to_use_it_whether_in_an_embedded_field_or_not" is too long (127 byte max)

Regardless, I've added a test case to cover this case to make sure we're always giving users at least the index name.

PR #1969, opened 2019-03-07
remove unnecessary files for production

rvitaliy:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | n.a.

#### Summary

- remove useless files for production from package
- configure git to automatic line endings normalized

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,17 @@
+# Auto-detect text files, ensure they use LF.
+* text=auto eol=lf
+
+# Exclude non-essential files from dist
+/.github export-ignore
+/docs export-ignore
+/tests export-ignore
+/tools export-ignore
+/.doctrine-project.json export-ignore
+/.gitattributes export-ignore
+/.gitignore export-ignore
+/.gitmodules export-ignore
+/.scrutinizer.yml export-ignore
+/.travis.yml export-ignore
+/phpcs.xml.dist export-ignore
+/phpstan.neon.dist export-ignore
+/phpunit.xml.dist export-ignore


.gitattributes
```diff
@@ -0,0 +1,17 @@
+# Auto-detect text files, ensure they use LF.
+* text=auto eol=lf
+
+# Exclude non-essential files from dist
+/.github export-ignore
+/docs export-ignore
```
alcaeus: I believe we should keep docs - opinions @doctrine/doctrinecore?

ostrolucky: Don't have strong preference, but I'm inclined to believe we should keep it rather than remove it

malarzm: I think we had a discussion sometime ago about docs and the consensus was they should stay

stof: @alcaeus I would be in favor of dropping them. People generally won't go read the doc inside their vendor folder, but would instead read the rendered file on the site.
On the other hand, people concerned about the size of their Docker images would hate such useless files in the vendor folder (I saw rants on Twitter a few weeks ago about the fact that Symfony ships tests in the archive, and that represents the same size than the actual Symfony codebase, doubling its size)

malarzm: On a second thought I wouldn't mind dropping the docs from the package, after all we now have docs online for each major version released and I think we won't be removing "old" ones as they're already nicely grayed out: https://www.doctrine-project.org/projects/mongodb-odm.html

ostrolucky: Online docs are not accessible when you are offline. And old stuff will eventually be removed from internet. That's how it works. After another 20 years, doctrine website (and project) might not even exist anymore.

malarzm: For the sake of having it around, doctrine1 was moved from SVN to git 9 years ago and the docs for it are still there. As long as we still have sources available online I think we're fine. Also I believe that being totally offline nowadays is hard to achieve :)

morozov: No need to keep the docs IMO. As for the tests, sometimes they may contain security vulnerabilities which are not accounted for in production.

Ocramius: I'm generally for shipping with docs: they are much smaller than the test tree anyway, and they are quite useful when working on a package while offline. On this particular diff, I'd say "let's ship it" and we can see if it has an effect on the ecosystem.

alcaeus:
After looking through the various Doctrine repositories, there's no clear "with" or "without" docs. Some repos exclude them, some don't. However, we've recently merged a few PRs that explicitly excluded the `docs` folder from exports. With that in mind, I'm ok with removing them here. Thanks for your opinions!

PR #1972, opened 2019-03-08
[1.3] Use build stages in .travis.yml

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

* PHP 5.6 will be tested as part of the build against lowest dependencies
* Tests the currently supported MongoDB versions (3.4, 3.6, 4.0)
* Adds PHP 7.3 and 7.4snapshot (with allowed failures) to the build matrix

This change allows us to add more build stages for 1.3, e.g. the upcoming performance test suite 🎉

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,33 +1,44 @@
-dist: trusty
+dist: xenial
 sudo: required
 language: php
 
 php:
-  - 5.6
   - 7.0
   - 7.1
   - 7.2
+  - 7.3
 
 env:
   global:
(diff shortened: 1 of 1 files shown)


malarzm:
FWIW something seems to be off with sharding

alcaeus:
Yep, need to investigate, just haven't gotten to it. Setting it up locally is always a pain 😉

PR #1974, opened 2019-03-11
Make addStage Aggregation Builder method public

Steveb-p:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | 

#### Summary

Makes `addStage` method public to allow easier usage of "custom" `Stage` class objects.

Following discussion in https://doctrine.slack.com/archives/CAAE6T66B/p1552300833003900 / https://doctrine.slack.com/messages/CAAE6T66B/convo/CAAE6T66B-1552300833.003900/

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
@@ -20,6 +20,7 @@
 namespace Doctrine\ODM\MongoDB\Aggregation;
 
 use Doctrine\MongoDB\Aggregation\Builder as BaseBuilder;
+use Doctrine\MongoDB\Aggregation\Stage;
 use Doctrine\MongoDB\Aggregation\Stage\GeoNear;
 use Doctrine\MongoDB\CommandCursor as BaseCommandCursor;
 use Doctrine\ODM\MongoDB\CommandCursor;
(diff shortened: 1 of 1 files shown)


alcaeus:
Technically, I could merge this, but while taking another look I noticed that this only partially helps. Any stage method in the builder class returns the newly added stage, which works because the base `Stage` class implements all methods also implemented by the builder. This means that you can simply chain all method calls. However, since the base class doesn't offer a public `addStage` class, this is not chain able like other methods:
```php
$builder
    ->match()
        ->field('foo')->equals('bar')
    ->addStage(new Match()) // <- fatal error because the match stage returned above does not have an addStage method
    ->group()
    // ...
;
```

I'm not sure this is a good idea, as it may be quite unexpected and confusing. I'm not sure this warrants another minor release of `doctrine/mongodb` that makes these methods public. With this in mind, I would suggest moving this back to 2.0 where we are in complete control of the API and can add the missing methods. What do you think @Steveb-p?

alcaeus:
Closing in favour of #2020, we can't really add this in 1.3 without making the fluent API of aggregation builder confusing, as `addStage` would only be available on the builder, not on stages.

Steveb-p:
@alcaeus Sorry I totally missed that comment in my mailbox :/ Just wanted to tell you I really appreciate your work :) Thanks for merging this to 2.0 :)

alcaeus:
Thank you for making the initial PR - sorry we couldn't get it into 1.3. 🍻

PR #1977, opened 2019-03-12
[1.2] Fix sharding commands for MongoDB 4.0

alcaeus:
Discovered this while working on #1972: the root problem is also present in 1.2, where we should also fix it.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/SchemaManager.php
+++ b/lib/Doctrine/ODM/MongoDB/SchemaManager.php
@@ -22,6 +22,7 @@ namespace Doctrine\ODM\MongoDB;
 use Doctrine\ODM\MongoDB\Mapping\ClassMetadata;
 use Doctrine\ODM\MongoDB\Mapping\ClassMetadataFactory;
 use Doctrine\ODM\MongoDB\Mapping\ClassMetadataInfo;
+use function in_array;
 use function sprintf;
 
 class SchemaManager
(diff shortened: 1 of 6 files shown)


lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -650,9 +651,10 @@ public function ensureDocumentSharding($documentName, array $indexOptions = arra
             }
         } while (! $done && $try < 2);
 
-        // Starting with MongoDB 3.2, this command returns code 20 when a collection is already sharded.
-        // For older MongoDB versions, check the error message
-        if ($result['ok'] == 1 || (isset($result['code']) && $result['code'] == 20) || $result['errmsg'] == 'already sharded') {
+        // For MongoDB 3.2 and newer, this command returns code 20 when a collection is already sharded.
+        // For MongoDB 4.0 and newer, this command returns code 23 when a collection is already sharded.
+        // For older MongoDB versions, we check the error message
+        if ($result['ok'] == 1 || (isset($result['code']) && in_array($result['code'], [20, 23])) || $result['errmsg'] == 'already sharded') {
```
jmikola: If you'd rather avoid magic numbers here, you could create private constants for `IllegalOperation` and `AlreadyInitialized`. See: https://github.com/mongodb/mongo/blob/master/src/mongo/base/error_codes.err

alcaeus: Constants added and marked as `@internal` due to lack of constant visibility. Will make them private in 2.0.

Issue #1970, opened 2019-03-07
Partial discriminator-maps not working anymore since 1.2.7


rieschl:
### BC Break Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 1.2.7

#### Summary

The changes [calculating the discriminator mapping](https://github.com/doctrine/mongodb-odm/pull/1962/files?utf8=%E2%9C%93&diff=unified#diff-c9b41d11b954b84d2c23a31bf8807b00L956) introduced a BC break in our mappings, because the discriminator map in the `mapped-superclass` only contained one of the documents. That may be "smelly", but it worked before.

#### Previous behavior

Having a partial discriminator map did automatically add the missing `discriminator-mapping`s using the FQCN as `discriminator-mapping value`.

#### Current behavior

The documents not in the `discriminator-map` can be saved but not fetched anymore.
If I add the other documents or remove the `discriminator-map` completely it works again.

###### Mapped superclass
```
<doctrine-mongo-mapping
        xmlns="http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping"
        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
        xsi:schemaLocation="http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping
        http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping.xsd"
>
    <mapped-superclass
            name="Eventjet\Ticket\AbstractTicket"
            collection="tickets"
            inheritance-type="SINGLE_COLLECTION"
    >
        <field name="id" id="true" type="uuid" strategy="NONE"/>
        <field name="number" fieldName="number" type="string"/>
        <field name="scan_code" fieldName="scanCode" type="string"/>
        <field name="revoked" type="boolean"/>
        <reference-one
                field="ticket_type"
                fieldName="ticketType"
                target-document="Eventjet\Ticket\TicketType\AbstractTicketType"
        />
        <discriminator-field name="type"/>
        <discriminator-map>
            <discriminator-mapping value="standing_room" class="Eventjet\Ticket\StandingRoomTicket"/>
        </discriminator-map>
        <indexes>
            <index name="tickets_revoked_seats">
                <key name="seat_id"/>
                <key name="revoked"/>
            </index>
        </indexes>
    </mapped-superclass>
</doctrine-mongo-mapping>
```
###### document mapping contained in `discriminator-map`:
```
[...]

PR #1978, opened 2019-03-13
Revert "Fix wrong usage of discriminator map in complex document inheritance chains"

alcaeus:
This PR reverts doctrine/mongodb-odm#1962 because of the regression highlighted in #1970. Since there is no useful way to fix this in 1.x due to partial discriminator maps being allowed, this will only be fixed in 2.0 where they have been removed.


malarzm:
Thanks 🎉

PR #1984, opened 2019-03-21
Mark classes as final

malarzm:
Several classes are now marked as final. Some things I was considering:
* is class marked as internal
* does it have an interface
* is there any way one could extend the class and use somehow inject it into ODM again

By the way I've removed two unused mock classes and a test that was testing nothing (interesting) and allowed me to make `ClassMetadataFactory` final


alcaeus:
With 1.3.x now being merged down into master, we no longer prepare separate pull requests to prepare deprecations. The course of action in this case would be to create a pull request to 1.3.x that marks the classes in question as `@final` and throws a deprecation when a class is extended. To do so, you can compare `self:class` against `static::class`: https://3v4l.org/jE9Ef.

alcaeus:
#2008 was merged into 1.3.x and further on to master, so this can be rebased. @malarzm you can remove changes to `phpstan.neon.dist` - I had to silence errors related to `@final` to get it stable: https://github.com/doctrine/mongodb-odm/commit/ba0798e1fa3a80a31465383a52b515973a0b8460#diff-0847156bd64b0de0f4d91da2394ec0e8R11.

PR #1989, opened 2019-04-02
Fix metadata storage when uploading GridFS files

### LINKED ISSUE #1985: Metadata not stored with File document
### Bug Report

I tried to follow [these instructions](https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/latest/reference/storing-files-with-mongogridfs.html) to store a file in the MongoDB. The file is stored just fine but the metadata is thrown away.

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.0.0-beta1

#### Summary

As stated above, the metadata are not stored in the document. They are, in fact, but are an empty object. If I specify the discriminator field it IS stored in the document, but the fields for metadata are not.

I traced the problem like this:

1. \Doctrine\ODM\MongoDB\Repository\DefaultGridFSRepository::uploadFromFile
1. \Doctrine\ODM\MongoDB\Repository\DefaultGridFSRepository::uploadFromStream
1. \Doctrine\ODM\MongoDB\Repository\DefaultGridFSRepository::prepareOptions
1. \Doctrine\ODM\MongoDB\Persisters\PersistenceBuilder::prepareInsertData
1. \Doctrine\ODM\MongoDB\UnitOfWork::getDocumentChangeSet

At point 3. the $uploadOptions->metadata contains the embedded document with metadata fields correctly set.

At point 5. the $changeSet is an empty array.

The metadata class is annotated simply as:

```
/**
 * @ODM\EmbeddedDocument()
 */
```

#### Current behavior

Metadata is not stored in the file document.

#### Expected behavior

Metadata are stored in the file document.

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | Fixes #1985 

#### Summary

This fixes a bug where metadata submitted while uploading a file to GridFS would not be stored correctly. The test only checked for the existence of the metadata object but did not check whether its information was stored correctly.

Unfortunately, due to the mechanics of GridFS uploads not all operations work as you'd normally expect. For example, uploading a file to GridFS with a metadata object that contains a reference will not cascade any operations to the referenced object as the UnitOfWork is not flushed during the update operation.

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/Repository/DefaultGridFSRepositoryTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Repository/DefaultGridFSRepositoryTest.php
@@ -5,11 +5,13 @@ declare(strict_types=1);
 namespace Doctrine\ODM\MongoDB\Tests\Repos;
 
 use DateTime;
+use Doctrine\ODM\MongoDB\Mapping\MappingException;
 use Doctrine\ODM\MongoDB\Repository\GridFSRepository;
 use Doctrine\ODM\MongoDB\Repository\UploadOptions;
 use Doctrine\ODM\MongoDB\Tests\BaseTest;
 use Documents\File;
 use Documents\FileMetadata;
+use Documents\FileWithoutMetadata;
 use Documents\User;
 use function fclose;
 use function filesize;
@@ -17,6 +19,7 @@ use function fopen;
 use function fseek;
 use function fstat;
 use function fwrite;
+use function sprintf;
 use function tmpfile;
 
 class DefaultGridFSRepositoryTest extends BaseTest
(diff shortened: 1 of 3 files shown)


tests/Doctrine/ODM/MongoDB/Tests/Repository/DefaultGridFSRepositoryTest.php
```diff
@@ -209,9 +221,29 @@ public function testDeletingFileAlsoDropsChunks() : void
...
+
+        $uploadOptions->metadata->getEmbedOne()->name = 'Foo';
+
+        $fileResource = fopen(__FILE__, 'r');
+
+        $this->expectException(MappingException::class);
+        $this->expectExceptionMessage(sprintf("No mapping found for field by DB name 'metadata' in class '%s'.", FileWithoutMetadata::class));
+
+        try {
+            /** @var File $file */
```
greg0ire: There is no such variable…

alcaeus: Nope, missed that while cleaning up 👍

Issue #1995, opened 2019-04-05
Proxy directory is created with wrong permissions


josefsabl:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.0.0-beta2

#### Summary

Relates to:

https://github.com/doctrine/mongodb-odm/issues/1890
https://github.com/doctrine/mongodb-odm/commit/9f4f4ce35beae0634973b7b32c9612e05008d869#diff-9502eb87a0cbee5c6090046e8f5494da

The original implementation created the dir with 775 permissions, the new implementation creates the dir with 755 (which breaks our server configuration)

PR #1996, opened 2019-04-05
Fix dir permissions for proxies

josefsabl:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | https://github.com/doctrine/mongodb-odm/issues/1995


alcaeus:
Thanks @josefsabl!

Issue #1993, opened 2019-04-05
Support metadata class name resolution in ResolveTargetDocumentListener


alcaeus:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | yes

#### Summary

The `ResolveTargetEntityListener` in ORM not only supports changing the `targetEntity` attribute for associations but also doing a secondary ClassMetadata lookup in case the original class was not found. This secondary behaviour is not supported in the `ResolveTargetDocumentListener` yet because of the missing `onClassMetadataNotFound` event handling.

This functionality should be added to ODM as well, but it requires several changes:
* Introduce the `onClassMetadataNotFound` event along with its object
* Add support for the `onClassMetadataNotFound` event in `ResolveTargetDocumentListener`
* Change `ResolveTargetDocumentListener` to be an event subscriber instead of an event listener (BC break)

emulienfou:
Hi @alcaeus just made the changes here:
* `ResolveTargetDocumentListener` has an event subscriber in this branch [here](https://github.com/emulienfou/mongodb-odm/tree/listener-event-subscriber)
* Introducing class `onClassMetadataNotFound` in this branch [here](https://github.com/emulienfou/mongodb-odm/tree/introducing-class-metadata-not-found)

I also update the **DoctrineMongoDBBundle** to register the service as an event_subscriber [here](https://github.com/emulienfou/DoctrineMongoDBBundle/tree/event-subscriber)

> I will made Pull Requests about this 3 modifications

emulienfou:
Hi @alcaeus however I think there is something I missed, I cannot figured out why the `getSubscribedEvents()` method is not triggered.

Here is the cache result for the service after updating the ODM bundle:
```
<service id="doctrine_mongodb.odm.listeners.resolve_target_document" class="Doctrine\ODM\MongoDB\Tools\ResolveTargetDocumentListener" public="false">
      <tag name="doctrine.event_subscriber"/>
      <call method="addResolveTargetDocument">
        <argument>Symfony\Component\Security\Core\User\UserInterface</argument>
        <argument>App\Document\User</argument>
        <argument type="collection"/>
      </call>
    </service>
```

When debugging the event manager, no listeners are registered. Cannot figured out why.
If you got some times to check that?

emulienfou:
Never mind, I figured out my issue

alcaeus:
[...]

PR #1997, opened 2019-04-07
Change `ResolveTargetDocumentListener` to be an event subscriber instead of an event listener

emulienfou:
|      Q       |   A
|------------- | -----------
| BC Break     | yes
| Fixed issues | #1993 

#### Summary

Note: needs changes from #1998. As long as that isn't merged, this PR will be unstable.

Change `ResolveTargetDocumentListener` to be an event subscriber instead of an event listener

### DIFF
--- a/UPGRADE-1.3.md
+++ b/UPGRADE-1.3.md
@@ -1,5 +1,31 @@
 # UPGRADE FROM 1.2 TO 1.3
 
+## Events
+
+### `onClassMetadataNotFound` event added
+
+The default ClassMetadataFactory now triggers an `onClassMetadataNotFound` event
+when no metadata was found for a given class. The event args contain information
+about the loaded class and can store a metadata object which can be used to load
+custom metadata when none was found.
+
+### `ResolveTargetDocumentListener` is now an event subscriber
+
+The `Doctrine\ODM\MongoDB\Tools\ResolveTargetDocumentListener` class now
+implements the `EventSubscriber` interface. You can still register it as a
+listener as was necessary previously, but in order to use new functionality you
+have to register it as an event subscriber.
+
+### `ResolveTargetDocumentListener` resolves class names on document load
+
+The `Doctrine\ODM\MongoDB\Tools\ResolveTargetDocumentListener` class not only
+resolves class names when looking up the `targetDoccumet` of an association in
+a document, but it also uses the new `onClassMetadataNotFound` event to resolve
+class names and trigger a secondary metadata load cycle if no metadata was found
+for the original class name. To use this functionality, either register an event
+listener for the `onClassMetadataNotFound` event or register the entire class as
+an event subscriber.
+
 ## GridFS
 
 GridFS support will be rewritten for 2.0 and was deprecated in its current form.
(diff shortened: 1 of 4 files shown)


lib/Doctrine/ODM/MongoDB/Tools/ResolveTargetDocumentListener.php
```diff
@@ -2,21 +2,34 @@
...
     private $resolveTargetDocuments = array();
 
+    /**
+     * {@inheritDoc}
+     */
+    public function getSubscribedEvents()
+    {
+        return [
+            Events::loadClassMetadata,
+            Events::onClassMetadataNotFound,
```
alcaeus: There is no method to handle this event.

tests/Doctrine/ODM/MongoDB/Tests/Tools/ResolveTargetDocumentListenerTest.php
```diff
@@ -18,11 +19,17 @@ class ResolveTargetDocumentListenerTest extends \Doctrine\ODM\MongoDB\Tests\Base
...
+     * @var ClassMetadataFactory
+     */
+    private $factory;
+
     public function setUp()
     {
         parent::setUp();
 
          $this->listener = new ResolveTargetDocumentListener();
+        $this->factory = $this->dm->getMetadataFactory();
```
malarzm: Seems like the line above has an extra space but I guess that's fixed in 2.0 so don't mind

alcaeus: Yep, wanted to avoid fixing it because a) it's in tests and b) it would just create another conflict in master 😉

lib/Doctrine/ODM/MongoDB/Tools/ResolveTargetDocumentListener.php
```diff
@@ -31,6 +45,26 @@ public function addResolveTargetDocument($originalDocument, $newDocument, array
         $this->resolveTargetDocuments[ltrim($originalDocument, "\\")] = $mapping;
     }
 
+    /**
+     * @param OnClassMetadataNotFoundEventArgs $args
```
malarzm: This phodoc is superfluous but it can be removed with tooling during merge to 2.0

UPGRADE-1.3.md
```diff
@@ -1,5 +1,31 @@
...
+when no metadata was found for a given class. The event args contain information
+about the loaded class and can store a metadata object which can be used to load
+custom metadata when none was found.
+
+### `ResolveTargetDocumentListener` is now an event subscriber
+
+The `Doctrine\ODM\MongoDB\Tools\ResolveTargetDocumentListener` class now
+implements the `EventSubscriber` interface. You can still register it as a
+listener as was necessary previously, but in order to use new functionality you
+have to register it as event subscriber.
```
malarzm: as an?

UPGRADE-1.3.md
```diff
@@ -1,5 +1,31 @@
...
+
+### `ResolveTargetDocumentListener` resolves class names on document load
+
+The `Doctrine\ODM\MongoDB\Tools\ResolveTargetDocumentListener` class not only
+resolves class names when looking up the `targetDoccumet` of an association in
+a document, but it also uses the new `onClassMetadataNotFound` event to resolve
+class names and trigger a secondary metadata load cycle if no metadata was found
+for the original class name. To use this functionality, either register an event
```
malarzm: ":" instead of ";"?

PR #2001, opened 2019-04-10
Provide deprecation notices for deprecated query types

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

With #1956, deprecated query types and their methods are removed. In order to prepare users for those, 1.3.x will properly deprecate these methods.


malarzm:
Thanks @alcaeus!

Issue #2002, opened 2019-04-10
Document class is unlisted in the discriminator map - using no discriminator map


josefsabl:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.0.0-beta2

#### Summary

Document class "Domain\XYZ\Document" is unlisted in the discriminator map although I am using no discriminator map.

Possibly related to this: https://github.com/doctrine/mongodb-odm/issues/867

#### Current behavior

Let's assume this code:

```
/**
 * @ODM\Document()
 * @ODM\InheritanceType("SINGLE_COLLECTION")
 * @ODM\DiscriminatorField("class")
 */
class DocumentA {
    /**
     * @ODM\ReferenceOne(targetDocument="DocumentA")
     * @ODM\Index
     * @var Document|null
     */
    public $parentDocument;
}

/**
 * @ODM\Document()
 */
class DocumentB extends DocumentA {
}

$docB = new DocumentB;
$docB->parentDocument = $docA;
```

Now I get the `Document class "DocumentA" is unlisted in the discriminator map` exception on a flush.

The interesting fact is that when I create the $docA in the same script as the $docB everything works.

However when I load it from the database prior setting it to the property, I get the exception.

alcaeus:
Could you by any chance reproduce this in a test case? You can find a whole bunch of such test cases in https://github.com/doctrine/mongodb-odm/tree/master/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket. Thanks!

josefsabl:
Will give it a shot.

josefsabl:
Done. I hope the clearing loaded metadata like that is okay. It simulates my problem where request that doesn't work with document A but works with document B fails.

josefsabl:
Until the bug is fixed, I forked the ODM and just plainly disabled the notification whatsoever to enable me to work on my project.

alcaeus:
Fix has been merged to 1.3 and master. Thanks for reporting, and more so for the failing test case!

josefsabl:
The exception went away, but a new problem appeared. See: #2024

PR #2004, opened 2019-04-11
Failing test for GH2002: Document class is unlisted in the discriminator map

josefsabl:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | yes
| Fixed issues | #2002 

#### Summary

Failing test to demonstrate the bug.

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH2002Test.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH2002Test.php
@@ -1,0 +1,71 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ODM\MongoDB\Tests\Functional\Ticket;
+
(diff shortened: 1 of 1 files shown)


alcaeus:
Thank you, I’ll take a look tomorrow morning 👍

alcaeus:
I've figured out why this only occurs after clearing metadata; this is unintended and will definitely be rolled back. We missed some changes in #1887 (https://github.com/doctrine/mongodb-odm/pull/1887/files#diff-79eacc807d7eb88891bcf592dd36cc33R104) that modify metadata when preparing update/upsert data for documents.

As for the rest, it looks like we were a bit too restrictive with the exception and didn't account for all cases (possibly because of the changes mentioned above). I'm currently expanding the test case you've created to cover all possible use cases of discriminators and discriminator maps. The issue at hand will definitely be fixed in the next beta!

I'm closing this PR as I've got the test I need and will be incorporating it into a different pull request. Thanks for your help reproducing this, it made the errors rather obvious and easy to spot!

PR #2005, opened 2019-04-12
Fix discriminator handling with partial discriminator maps

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | 

#### Summary

While working on #2002, I noticed that the detection logic wasn't quite right. So I expanded the failing test submitted by @josefsabl to cover more use cases to make sure we're getting the discriminator logic correct.

Note that this doesn't fix this bug for 2.0, as we need to adapt the logic there a little bit to cover for the (expected) exceptions. While debugging, I also noticed an issue where the `PersistenceBuilder` modifies the ClassMetadata which definitely shouldn't happen.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/DocumentManager.php
+++ b/lib/Doctrine/ODM/MongoDB/DocumentManager.php
@@ -749,53 +749,42 @@ class DocumentManager implements ObjectManager
     {
         $discriminatorField = null;
         $discriminatorValue = null;
-        $discriminatorData  = [];
+        $discriminatorMap   = null;
 
         if (isset($referenceMapping['discriminatorField'])) {
             $discriminatorField = $referenceMapping['discriminatorField'];
+
             if (isset($referenceMapping['discriminatorMap'])) {
-                $pos = array_search($class->name, $referenceMapping['discriminatorMap']);
+                $discriminatorMap = $referenceMapping['discriminatorMap'];
+            }
+        } else {
+            $discriminatorField = $class->discriminatorField;
+            $discriminatorValue = $class->discriminatorValue;
+            $discriminatorMap   = $class->discriminatorMap;
+        }
+
+        if ($discriminatorField === null) {
+            return [];
+        }
+
+        if ($discriminatorValue === null) {
+            if (! empty($discriminatorMap)) {
+                $pos = array_search($class->name, $discriminatorMap);
+
                 if ($pos !== false) {
                     $discriminatorValue = $pos;
                 }
             } else {
                 $discriminatorValue = $class->name;
             }
-        } else {
-            $discriminatorField = $class->discriminatorField;
-            $discriminatorValue = isset($class->discriminatorValue) ? $class->discriminatorValue : $class->name;
         }
 
-        if ($discriminatorField !== null) {
-            if ($discriminatorValue === null) {
-                @trigger_error(sprintf('Document class "%s" is unlisted in the discriminator map for reference "%s". This is deprecated and will throw an exception in 2.0.', $class->name, $referenceMapping['name']), E_USER_DEPRECATED);
-                $discriminatorValue = $class->name;
-            }
-
-            $discriminatorData = [$discriminatorField => $discriminatorValue];
-        } elseif (! isset($referenceMapping['targetDocument'])) {
-            $discriminatorField = $referenceMapping['discriminatorField'];
-
-            $discriminatorMap = null;
-            if (isset($referenceMapping['discriminatorMap'])) {
-                $discriminatorMap = $referenceMapping['discriminatorMap'];
-            }
-
-            if ($discriminatorMap === null) {
-                $discriminatorValue = $class->name;
-            } else {
-                $discriminatorValue = array_search($class->name, $discriminatorMap);
-
-                if ($discriminatorValue === false) {
-                    @trigger_error(sprintf('Document class "%s" is unlisted in the discriminator map for reference "%s". This is deprecated and will throw an exception in 2.0.', $class->name, $referenceMapping['name']), E_USER_DEPRECATED);
-                    $discriminatorValue = $class->name;
-                }
-            }
-
-            $discriminatorData = [$discriminatorField => $discriminatorValue];
+        if ($discriminatorValue === null) {
+            @trigger_error(sprintf('Document class "%s" is unlisted in the discriminator map for reference "%s". This is deprecated and will throw an exception in doctrine/mongodb-odm 2.0.', $class->name, $referenceMapping['name']), E_USER_DEPRECATED);
+            $discriminatorValue = $class->name;
         }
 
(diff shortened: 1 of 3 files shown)


lib/Doctrine/ODM/MongoDB/DocumentManager.php
```diff
@@ -749,53 +749,43 @@ private function getDiscriminatorData(array $referenceMapping, ClassMetadata $cl
...
+                    $discriminatorValue = $pos;
                 }
+            } else {
+                $discriminatorValue = $class->name;
             }
+        }
 
-            $discriminatorData = [$discriminatorField => $discriminatorValue];
+        if ($discriminatorValue === null) {
+            @trigger_error(sprintf('Document class "%s" is unlisted in the discriminator map for reference "%s". This is deprecated and will throw an exception in 2.0.', $class->name, $referenceMapping['name']), E_USER_DEPRECATED);
```
greg0ire: What do you think of including the package name in deprecations?

```suggestion
            @trigger_error(sprintf('Document class "%s" is unlisted in the discriminator map for reference "%s". This is deprecated and will throw an exception in doctrine/mongodb-odm 2.0.', $class->name, $referenceMapping['name']), E_USER_DEPRECATED);
```

lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
```diff
@@ -110,9 +110,17 @@ public function prepareInsertData($document)
...
         // add discriminator if the class has one
         if (isset($class->discriminatorField)) {
-            $insertData[$class->discriminatorField] = isset($class->discriminatorValue)
-                ? $class->discriminatorValue
-                : $class->name;
+            $discriminatorValue = $class->discriminatorValue;
+
+            if ($discriminatorValue === null) {
+                if (! empty($class->discriminatorMap)) {
+                    @trigger_error(sprintf('Document class "%s" is unlisted in the discriminator map of its inheritance chain. This is deprecated and will throw an exception in 2.0.', $class->name), E_USER_DEPRECATED);
```
greg0ire:
```suggestion
                    @trigger_error(sprintf('Document class "%s" is unlisted in the discriminator map of its inheritance chain. This is deprecated and will throw an exception in doctrine/mongodb-odm 2.0.', $class->name), E_USER_DEPRECATED);
```

lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
```diff
@@ -289,9 +297,17 @@ public function prepareUpsertData($document)
...
         // add discriminator if the class has one
         if (isset($class->discriminatorField)) {
-            $updateData['$set'][$class->discriminatorField] = isset($class->discriminatorValue)
-                ? $class->discriminatorValue
-                : $class->name;
+            $discriminatorValue = $class->discriminatorValue;
+
+            if ($discriminatorValue === null) {
+                if (! empty($class->discriminatorMap)) {
+                    @trigger_error(sprintf('Document class "%s" is unlisted in the discriminator map of its inheritance chain. This is deprecated and will throw an exception in 2.0.', $class->name), E_USER_DEPRECATED);
```
greg0ire:
```suggestion
                    @trigger_error(sprintf('Document class "%s" is unlisted in the discriminator map of its inheritance chain. This is deprecated and will throw an exception in doctrine/mongodb-odm 2.0.', $class->name), E_USER_DEPRECATED);
```

lib/Doctrine/ODM/MongoDB/DocumentManager.php
```diff
@@ -749,53 +749,43 @@ private function getDiscriminatorData(array $referenceMapping, ClassMetadata $cl
     {
         $discriminatorField = null;
         $discriminatorValue = null;
-        $discriminatorData  = [];
+        $discriminatorMap   = null;
 
-        if (isset($referenceMapping['discriminatorField'])) {
+        if (isset($referenceMapping['discriminatorField']) || ! isset($referenceMapping['targetDocument'])) {
             $discriminatorField = $referenceMapping['discriminatorField'];
```
malarzm: According to new `if` logic `$referenceMapping['discriminatorField']` can now be empty

lib/Doctrine/ODM/MongoDB/DocumentManager.php
```diff
@@ -749,53 +749,43 @@ private function getDiscriminatorData(array $referenceMapping, ClassMetadata $cl
...
+            return [];
+        }
 
-            if ($discriminatorMap === null) {
-                $discriminatorValue = $class->name;
-            } else {
-                $discriminatorValue = array_search($class->name, $discriminatorMap);
+        if ($discriminatorValue === null) {
+            if (! empty($discriminatorMap)) {
+                $discriminatorValue = null;
```
malarzm: `$discriminatorValue` is already `null`

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH2002Test.php
```diff
@@ -0,0 +1,232 @@
+<?php
+
+namespace Doctrine\ODM\MongoDB\Tests\Functional\Ticket;
+
+use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
+use Doctrine\ODM\MongoDB\Tests\BaseTest;
+
+class GH2002Test extends BaseTest
```
malarzm: Is it worth putting into more concrete test than a ticket one?

lib/Doctrine/ODM/MongoDB/DocumentManager.php
alcaeus: https://github.com/doctrine/mongodb-odm/blob/b03d693324015b65f432820ed6051f0cff056520/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php#L1500..L1502 should cover this by automatically adding a default value for `discriminatorField`.

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH2002Test.php
alcaeus: I'd leave it like this for now - this is cleanup that I was pushing off past 2.0: I want to clean up functional tests and move them to more concrete test classes.

lib/Doctrine/ODM/MongoDB/DocumentManager.php
malarzm: In that case `isset` in `if` doesn't make sense :)

alcaeus: Right. Removed. I had already added a test to cover against that 🙄

PR #2006, opened 2019-04-12
Run appropriate query operations when replacing documents

### LINKED ISSUE #1981: Query builder incorrectly handles updates with replacement documents
[`Query::runQuery()`](https://github.com/doctrine/mongodb-odm/blob/2.0.0-beta1/lib/Doctrine/ODM/MongoDB/Query/Query.php#L442) currently delegates `TYPE_UPDATE` operations to either [`Collection::updateOne()`](https://docs.mongodb.com/php-library/current/reference/method/MongoDBCollection-updateOne/) or [`updateMany()`](https://docs.mongodb.com/php-library/current/reference/method/MongoDBCollection-updateMany/) depending on the `multiple` option. If it's possible for users to provide a replacement document (i.e. no update operators) through the query builder, doing so will yield an exception. The query builder should delegate to [`replaceOne()`](https://docs.mongodb.com/php-library/current/reference/method/MongoDBCollection-replaceOne/) for such operations.

Likewise, the query builder should delegate `findAndModify` commands with replacement documents to [`findOneAndReplace()`](https://docs.mongodb.com/php-library/current/reference/method/MongoDBCollection-findOneAndReplace/). It currently only delegates to the update/delete variants.

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | Fixes #1981 

#### Summary

This ensures that when not using update operators, the query does not attempt to run an updateOne or findOneAndUpdate operation. Instead, the replaceOne or findOneAndReplace operation must be run to avoid errors.
When using the "multiple" option on an update query in combination with a replacement document, the query class will now throw an error as there is no replaceMany functionality.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Query/Query.php
+++ b/lib/Doctrine/ODM/MongoDB/Query/Query.php
@@ -32,6 +32,9 @@ use function array_values;
 use function assert;
 use function is_array;
 use function is_callable;
+use function is_string;
+use function key;
+use function reset;
 
 /**
  * ODM Query wraps the raw Doctrine MongoDB queries to add additional functionality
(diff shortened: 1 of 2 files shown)


alcaeus:
> Looks A-OK but shouldn't the tests actually assert the correct command is ran? Or would the library complain otherwise?

This is where it gets interesting: the actual command run is always `update` - this is just the high-level API that MongoDB provides. So while I could use a command listener to check the command, the absence of an error is all the assertion we can have at the moment.

malarzm:
👌

PR #2007, opened 2019-04-12
Loosen naming rules for discriminator values

### LINKED ISSUE #1980: less strict constraint for discriminator-mapping values in 2.0
It's not really a feature request but more retaining backwards compatibility.
When testing version 2 I noticed that the syntax checks for XML mappings are more strict than in version 1. Using a FQCN as `discriminator-mapping` `value` now throws a `Doctrine\ODM\MongoDB\Mapping\MappingException` for it not being a `xs:NMTOKEN` type.

While I totally understand the necessity of breaking BC from time to time, this break causes a whole other set of issues: Version 1 supported partial discriminator maps and automatically created the discriminator mapping when it's not defined in the mapping. Funny enough the automatically set value of the mapping is the FQCN 😄
So, when upgrading to 2.0 I don't just have to add the missing discriminator mappings but also migrate existing documents to use a legal name, because currently they're all FQCNs now.
That can be very complicated on busy environments so I kindly ask to rethink this current constrain for the upcoming version 2.
Also, (I didn't check) probably the annotation mapping doesn't have this kind of constraint, so it would be reasonable to have the same behavior in both mapping drivers. I'm sure lots of people want to upgrade to version 2 as soon as it's released because we don't have to use compatibility layers for the old PHP5 `mongo` driver anymore.

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | Fixes #1980 

#### Summary

With 2.0, loading mapping files that don't validate against the provided XSD file are no longer loaded and cause an exception. In 1.3, such files will trigger a deprecation notice to inform the user of the upcoming restriction. However, some of the values that were accepted before are not considered valid, most notably discriminator values. This PR loosens the restriction for those types.

Technically, we can most likely do the same for many other attributes that use `NMTOKEN` right now, especially since those checks are not done for annotation mappings. However, I'd rather not completely open up the types for field names, as we'd most likely want to restrict them to what MongoDB allows. Thus, only discriminator values are loosened right now.

### DIFF
--- a/doctrine-mongo-mapping.xsd
+++ b/doctrine-mongo-mapping.xsd
@@ -285,7 +285,7 @@
   </xs:simpleType>
 
   <xs:complexType name="discriminator-mapping">
-    <xs:attribute name="value" type="xs:NMTOKEN" use="required"/>
+    <xs:attribute name="value" type="xs:string" use="required"/>
     <xs:attribute name="class" type="xs:string" use="required"/>
   </xs:complexType>
 
@@ -300,7 +300,7 @@
   </xs:complexType>
 
   <xs:complexType name="default-discriminator-value">
-    <xs:attribute name="value" type="xs:NMTOKEN" use="required"/>
+    <xs:attribute name="value" type="xs:string" use="required"/>
   </xs:complexType>
 
   <xs:simpleType name="lifecycle-callback-type">


malarzm:
Thanks @alcaeus!

PR #2011, opened 2019-04-15
[2.1] Deprecate AbstractIdGenerator

malarzm:
There's no reason to provide an abstract class instead of an interface. Since this is a BC break I propose we do this for 2.0 although I'm eager to deprecate the class in 1.3 and remove the abstract in 2.0 already :)

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Id/AbstractIdGenerator.php
+++ b/lib/Doctrine/ODM/MongoDB/Id/AbstractIdGenerator.php
@@ -4,17 +4,9 @@ declare(strict_types=1);
 
 namespace Doctrine\ODM\MongoDB\Id;
 
-use Doctrine\ODM\MongoDB\DocumentManager;
-
 /**
- * AbstractIdGenerator
+ * @deprecated AbstractIdGenerator was deprecated in 2.1 and will be removed in 3.0. Implement IdGenerator interface instead.
  */
-abstract class AbstractIdGenerator
+abstract class AbstractIdGenerator implements IdGenerator
 {
-    /**
-     * Generates an identifier for a document.
-     *
-     * @return mixed
-     */
-    abstract public function generate(DocumentManager $dm, object $document);
 }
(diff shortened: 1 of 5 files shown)


lib/Doctrine/ODM/MongoDB/Id/AbstractIdGenerator.php
```diff
@@ -4,17 +4,10 @@
...
-abstract class AbstractIdGenerator
+abstract class AbstractIdGenerator implements IdGenerator
 {
-    /**
-     * Generates an identifier for a document.
-     *
-     * @return mixed
-     */
-    abstract public function generate(DocumentManager $dm, object $document);
+
```
malarzm: I was thinking about adding a `__construct` emitting `E_USER_DEPRECATED` but chances are that if somebody has overridden the constructor, the parent's one might be not called

alcaeus: Since we don't have a constructor to hook into, you may also trigger the deprecation on top of the file. It will alert people that they're using deprecated functionality, even though it won't point out every single occurrence where this is used.

alcaeus:
One minor thing: this needs to be mentioned in the upgrade notes.

lib/Doctrine/ODM/MongoDB/Id/AbstractIdGenerator.php
malarzm: Not sure if it's a problem, but since our generators are extending the abstract they'll also generate a deprecation notice. And we can't remove it from the inheritance chain as that'd be a BC break too :/

alcaeus: Ah, yeah. Difficult one, that 🤔

alcaeus:
Pushed back to 2.1 to not deprecate stuff in a 2.0 release 😉

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -1370,7 +1370,7 @@ public function isCollectionValuedEmbed(string $fieldName) : bool
     /**
      * Sets the ID generator used to generate IDs for instances of this class.
      */
-    public function setIdGenerator(AbstractIdGenerator $generator) : void
+    public function setIdGenerator(IdGenerator $generator) : void
```
alcaeus, line 1373: This is a BC break since `ClassMetadata` is not final - not sure how to best solve this at this time :|

alcaeus: To add a quick note, I think the impact of the BC break is small enough to risk doing this. I would also make the class `@final` to indicate that people really shouldn't be messing with it. We can then consider making it `final` in 3.0.

Issue #2009, opened 2019-04-15
Document remaining BC breaks


alcaeus:
|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.0.0

### BC Break Report

Running [roave/backward-compatibility-check](https://github.com/Roave/BackwardCompatibilityCheck) on MongoDB ODM 2.0 against 1.2.x yields a total of 3303 BC breaks. I suppose many of them can be ignored (e.g. default values or properties changed and such) and many have been documented, but there have been a few instances where we've forgotten to deprecate functionality, mainly related to the migration from `doctrine/mongodb` to `mongodb/mongodb` which brought along a bunch of implicit BC breaks (see #1988, #1959).

I've created a gist of the current state of BC breaks containing all BC breaks the tool identified: https://gist.github.com/alcaeus/b8c1b38e297cc957e991683bef069771. These BC breaks need to be checked and if relevant, properly documented in the `UPGRADE` files and `trigger_error` calls.

PR #2012, opened 2019-04-16
Document more BC breaks

alcaeus:
|      Q       |   A
|------------- | -----------
| BC Break     | no
| Fixed issues | Part of #2009

#### Summary

This documents all BC breaks from the `REMOVED` category of https://gist.github.com/alcaeus/b8c1b38e297cc957e991683bef069771. It includes a BC layer for the `DocumentRepository` deprecation similar to how we handle `ClassMetadata`. @greg0ire I'd appreciate it if you could take a close look at that particular commit, I'm not entirely sure I didn't mess it up.

I'll create a separate PR to go through the other BC breaks the tool reported; I'm not sure we want to document every type hint that we added.


alcaeus:
> The code looks good, not sure why 5.6 is unhappy with it 🤔

Coverage seems to have imported the `DocumentRepository` class, which means that importing `Repository\DocumentRepository` will not work without aliasing it.

malarzm:
> I'm not sure we want to document every type hint that we added.

IMO it's just not worth it. A sentence like "We've added scalar typehints nearly everywhere" will do :)

lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
```diff
@@ -64,6 +64,8 @@ public function field($fieldName)
     /**
      * @param mixed|self $expression
      * @return mixed
+     *
+     * @internal This method will be private in MongoDB ODM 2.0.
```
malarzm: Should this method emit a deprecation notice if it's used from a class extending this one?

lib/Doctrine/ODM/MongoDB/CommandCursor.php
```diff
@@ -29,6 +29,7 @@
  *
  * @since  1.1
  * @author alcaeus <alcaeus@alcaeus.org>
+ * @deprecated This class is deprecated and will be removed in 2.0. You should typehint against the {@see Iterator} interface instead.
```
malarzm: it'd be possible to emit a deprecation in the ctor

lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -86,6 +87,18 @@ class Configuration extends \Doctrine\MongoDB\Configuration
      */
     const AUTOGENERATE_EVAL = 3;
 
+    /**
+     * Array of attributes for this configuration instance.
+     *
+     * @var array
+     * @deprecated This property will be private in MongoDB ODM 2.0.
+     */
+    protected $attributes = [
```
malarzm: what's the point of this array? Was it protected in the extended configuration and it's easier to deprecate this way than to release new `doctrine/mongodb` with deprecation?

lib/Doctrine/ODM/MongoDB/Cursor.php
```diff
@@ -37,8 +38,9 @@
  * For compatibility, this class also extends Doctrine\MongoDB\Cursor.
  *
  * @since  1.0
+ * @deprecated This class is deprecated and will be removed in 2.0. You should typehint against the {@see Iterator} interface instead.
```
malarzm: Deprecation in the ctor?

lib/Doctrine/ODM/MongoDB/EagerCursor.php
```diff
@@ -27,16 +27,8 @@
...
  *
  * @since  1.0
- * @deprecated Deprecated in favor of using Cursor; will be removed in 2.0
+ * @deprecated This class is deprecated and will be removed in 2.0. You should typehint against the {@see Iterator} interface instead.
  */
 class EagerCursor extends Cursor
 {
-    public function __construct(CursorInterface $baseCursor, UnitOfWork $unitOfWork, ClassMetadata $class)
-    {
-        @trigger_error(
```
malarzm: Why is the notice removed?

lib/Doctrine/ODM/MongoDB/MongoDBException.php
```diff
@@ -31,6 +31,8 @@ class MongoDBException extends \Exception
      * @param string $fieldName
      * @param string $method
      * @return MongoDBException
+     *
+     * @deprecated This method was deprecated in 1.3 and will be dropped in ODM 2.0.
```
malarzm: Not sure if worth it in this case, but deprecation notices are missing in both deprecated methods

lib/Doctrine/ODM/MongoDB/Query/Query.php
```diff
@@ -45,8 +45,32 @@ class Query extends \Doctrine\MongoDB\Query\Query
...
+    /** @deprecated */
     const HINT_READ_PREFERENCE_TAGS = 4;
     const HINT_READ_ONLY = 5;
+    /** @deprecated */
+    const TYPE_GEO_NEAR = 10;
+
+    /**
+     * @inheritdoc
+     *
+     * @internal This field will be private in MongoDB ODM 2.0.
```
malarzm: Entire class will be `final`

lib/Doctrine/ODM/MongoDB/Query/Query.php
```diff
@@ -359,6 +383,8 @@ public function execute()
      * @see \Doctrine\MongoDB\Cursor::prepareCursor()
      * @param BaseCursor $cursor
      * @return CursorInterface
+     *
+     * @deprecated This method will be removed in MongoDB ODM 2.0.
```
malarzm: missing deprecation in the method itself

lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
alcaeus: That should work 👍

lib/Doctrine/ODM/MongoDB/CommandCursor.php
alcaeus: We still use this class extensively, which would cause deprecation notices users can't fix, which is why I haven't added that.

lib/Doctrine/ODM/MongoDB/Configuration.php
alcaeus: Exactly this - don't feel like doing a bunch of work in doctrine/mongodb.

lib/Doctrine/ODM/MongoDB/Cursor.php
alcaeus: See https://github.com/doctrine/mongodb-odm/pull/2012#discussion_r276080917.

lib/Doctrine/ODM/MongoDB/EagerCursor.php
alcaeus: See https://github.com/doctrine/mongodb-odm/pull/2012#discussion_r276080917.

lib/Doctrine/ODM/MongoDB/MongoDBException.php
alcaeus: I wasn't sure whether to throw them or not. I could add it, but a wrong call to `findBy` would cause two deprecation notices in that case. Do you think it'd be worth adding it?

lib/Doctrine/ODM/MongoDB/Query/Query.php
alcaeus: I'll remove the comment then

lib/Doctrine/ODM/MongoDB/Query/Query.php
alcaeus: I can add a conditional one if the class triggering it is not `Query\Query` itself.

alcaeus: Just realised that the class will be final, so that deprecation notice should tell people that they shouldn't be calling any protected methods anymore. I'll skip the deprecation notice in the method because of that.

alcaeus:
Added more deprecations from the list of BC breaks. BC breaks due to type hints are mostly relevant to people extending classes or implementing interfaces, so I'd like to provide this information in a separate UPGRADE document for people that are interested in those particular cases.

lib/Doctrine/ODM/MongoDB/MongoDBException.php
alcaeus: Added `trigger_error` calls to deprecated methods - people need to fix one deprecation anyways, that will also get rid of the secondary deprecation. The advantage is that people calling these methods themselves will also be warned 👍

PR #2014, opened 2019-04-21
Update to doctrine/cs v6

carusogabriel:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | -

#### Summary

Use the latest version of https://github.com/doctrine/coding-standard.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/GraphLookup.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/GraphLookup.php
@@ -93,12 +93,14 @@ class GraphLookup extends Stage
         // No targetClass mapping - simply use field name as is
         if (! $this->targetClass) {
             $this->connectFromField = $connectFromField;
+
             return $this;
         }
 
         // connectFromField doesn't have to be a reference - in this case, just convert the field name
         if (! $this->targetClass->hasReference($connectFromField)) {
             $this->connectFromField = $this->convertTargetFieldName($connectFromField);
+
             return $this;
         }
 
@@ -109,6 +111,7 @@ class GraphLookup extends Stage
         }
 
         $this->connectFromField = $this->getReferencedFieldName($connectFromField, $referenceMapping);
+
         return $this;
     }
 
@@ -119,6 +122,7 @@ class GraphLookup extends Stage
     public function connectToField(string $connectToField) : self
     {
         $this->connectToField = $this->convertTargetFieldName($connectToField);
+
         return $this;
     }
 
@@ -159,6 +163,7 @@ class GraphLookup extends Stage
             $this->targetClass = $this->dm->getClassMetadata($from);
         } catch (BaseMappingException $e) {
             $this->from = $from;
+
             return $this;
         }
 
@@ -167,6 +172,7 @@ class GraphLookup extends Stage
         }
 
         $this->from = $this->targetClass->getCollection();
+
         return $this;
     }
 
@@ -274,7 +280,9 @@ class GraphLookup extends Stage
     {
         if (is_array($expression)) {
             return array_map([$this, 'convertExpression'], $expression);
-        } elseif (is_string($expression) && substr($expression, 0, 1) === '$') {
+        }
+
+        if (is_string($expression) && substr($expression, 0, 1) === '$') {
             return '$' . $this->getDocumentPersister($this->class)->prepareFieldName(substr($expression, 1));
         }
 
@@ -317,8 +325,6 @@ class GraphLookup extends Stage
             case ClassMetadata::REFERENCE_STORE_AS_ID:
             case ClassMetadata::REFERENCE_STORE_AS_REF:
                 return ClassMetadata::getReferenceFieldName($mapping['storeAs'], $mapping['name']);
-                break;
-
             default:
                 throw MappingException::cannotLookupDbRefReference($this->class->name, $fieldName);
         }
--- a/lib/Doctrine/ODM/MongoDB/DocumentManager.php
+++ b/lib/Doctrine/ODM/MongoDB/DocumentManager.php
@@ -510,6 +510,7 @@ class DocumentManager implements ObjectManager
             throw new InvalidArgumentException(gettype($document));
         }
         $this->errorIfClosed();
+
         return $this->unitOfWork->merge($document);
     }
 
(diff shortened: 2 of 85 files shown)


lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionTrait.php
```diff
@@ -226,7 +227,7 @@ public function takeSnapshot()
     public function clearSnapshot()
     {
         $this->snapshot = [];
-        $this->isDirty  = $this->coll->count() ? true : false;
+        $this->isDirty  = $this->coll->count() !== 0;
```
carusogabriel: Is this change right?

lib/Doctrine/ODM/MongoDB/Repository/DocumentRepository.php
```diff
@@ -46,7 +46,6 @@ class DocumentRepository implements ObjectRepository, Selectable
 
     /**
      * Initializes this instance with the specified document manager, unit of work and
-     * class metadata.
```
carusogabriel: Caused by https://github.com/doctrine/coding-standard/pull/97. Maybe we should inline this comment?

lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionTrait.php
malarzm: Looks good 👍

lib/Doctrine/ODM/MongoDB/Repository/DocumentRepository.php
malarzm: Shouldn't that rule remove comments like "Class DocumentRepository" as that's the name of the file? If it removes more than I'd argue it's a bug in the rule/fixer

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/GraphLookup.php
```diff
@@ -317,6 +325,7 @@ private function getReferencedFieldName(string $fieldName, array $mapping) : str
             case ClassMetadata::REFERENCE_STORE_AS_ID:
             case ClassMetadata::REFERENCE_STORE_AS_REF:
                 return ClassMetadata::getReferenceFieldName($mapping['storeAs'], $mapping['name']);
+
                 break;
```
malarzm: `break` should be removed since we have a `return` just before

lib/Doctrine/ODM/MongoDB/DocumentManager.php
```diff
@@ -718,6 +720,7 @@ public function createReference(object $document, array $referenceMapping)
                 }
 
                 return $class->getDatabaseIdentifierValue($id);
+
                 break;
```
malarzm: Same here

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -1157,7 +1159,7 @@ public function setCollectionMax(int $max) : void
      */
     public function isMappedToCollection() : bool
     {
-        return $this->collection ? true : false;
+        return $this->collection !== '';
```
malarzm: Theoretically `$this->collection` can be null as [it has no default](https://github.com/doctrine/mongodb-odm/blob/485d84dc05401aea58db9d41335924475b50f7f1/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php#L189) value so this is wrong

lib/Doctrine/ODM/MongoDB/Query/Expr.php
```diff
@@ -661,7 +672,9 @@ public function includesReferenceTo(object $document) : self
         switch ($storeAs) {
             case ClassMetadata::REFERENCE_STORE_AS_ID:
                 $this->query[$mapping['name']] = $reference;
+
                 return $this;
+
                 break;
```
malarzm: `break` can be removed

lib/Doctrine/ODM/MongoDB/Query/Expr.php
```diff
@@ -1045,7 +1067,9 @@ public function references(object $document) : self
...
```
malarzm: `break` can be removed

lib/Doctrine/ODM/MongoDB/Repository/DocumentRepository.php
carusogabriel: I agree. I'll open a PR fixing it there 👍

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
carusogabriel: Should we check for both `''` and 'null' then?

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/GraphLookup.php
carusogabriel: Strange that this should be caught by [`Squiz.PHP.NonExecutableCode`](https://github.com/doctrine/coding-standard/blob/master/lib/Doctrine/ruleset.xml#L420). I'll fix it by now, but this should be report upstream ☺️

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
malarzm: I'd say yes and if you could update variable typehint to include `|null` that'd be great :)

lib/Doctrine/ODM/MongoDB/Repository/DocumentRepository.php
carusogabriel: @malarzm Is that okay to inline this comment to get this PR merged? I don't see this issue being fixed upstream :(

alcaeus: Yeah, let’s do that. Line length limits aren’t enforced anyways.

malarzm: Sure go ahead 👍

carusogabriel:
I'll come back to this during the week :)

malarzm:
@carusogabriel could you please resolve the conflicts?

carusogabriel:
@malarzm Done 👍

alcaeus:
Thanks Gabriel!

Issue #2015, opened 2019-04-23
Discriminator value is lost for reloaded documents


josefsabl:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | dev-master

#### Summary

Okay, so this is probably related or maybe duplicate to: https://github.com/doctrine/mongodb-odm/issues/2002

But not only I get the error message described in the ticket but I also lose the discriminator value for the references associations.

#### Current behavior

After reloading and setting new value for reference, the discriminator field in the ref is lost.

Setting discriminator map will fix the problem.

#### How to reproduce

https://github.com/doctrine/mongodb-odm/pull/2016

![2019-04-23_18h24_27](https://user-images.githubusercontent.com/1252386/56598695-47f2d400-65f5-11e9-9e3a-38065c184135.png)

josefsabl:
If I may humbly present my opinion. I don't like the discriminator map feature at all. Especially with the latest changes.

I believe it violates DRY and OCP quite obviously.

First of all I need to maintain lists like:

```
user => Document\User
userAdmin => Document\UserAdmin
userBfu => Document\UserBfu
userSuper => Document\UserSuper
...
```

We also search for documents by their type (`findBy(['class' => Document\User])`). And as we don't want to have string constants in the code we would have to create constants for strings like `user`, `userAdmin` etc... and they cannot be used in annotations afaik. So there is no way to solve ideally.

But more importantly, I need to add to this list each time any other developer adds a new class in his package that is merely dependent on the base package.

The latest change with the exception targets the later problem so that the user does not forget. But it is not only about forgetting but about architecture. Imagine you have installed the base class as a dependency, think of it as some company level framework.

I know about the argument that the implementation details should not leak into the database and not using a discriminator map prevents the code to be refactored (inability to move/rename classes). Well, I'd put aside the fact that I could just run a trivial script to update the database, but the discriminator map was useful for migration as well. I could rename the class and add the new name to the discriminator map, think of alsoLoad. It is impossible with the introduction of the exception.

Thanks for listening, or not :-)

malarzm:
I'll allow myself to answer to some of your points:

> We also search for documents by their type (findBy(['class' => Document\User])).

You shouldn't need to query by a discriminator field, if you get a repository for a specific class (say `Document\User`) it'll automatically get you instances of that class only. Discriminator and its value are ODM-specific, it shouldn't leak to your code.

[...]

PR #2016, opened 2019-04-23
Issue 2015 Test

josefsabl:
Display of buggy behaviour described in: https://github.com/doctrine/mongodb-odm/issues/2015

Sorry, I didn't notice you didn't merge my last test so it is displayed here again. I guess you won't merge this either so it doesn't matter :-)


alcaeus:
Thanks for the test case. With the fix from #2005 applied, this test no longer fails, so I'm closing this PR.

Issue #1808, opened 2018-06-05
Allow passing additional driver options


alcaeus:
Previously reported as https://github.com/alcaeus/mongo-php-adapter/issues/227. It looks like it's not currently possible to pass the `noCursorTimeout` to the driver when running a find query. This should be supported, along with any other options that were added to the new driver.

malarzm:
If it's a matter of adding `noCursorTimeout` method to query builder I'd add it to 1.3

alcaeus:
I'm not sure the legacy driver supports it - I'd have to check the source code or ask @jmikola.

jmikola:
In the legacy driver, this can be expressed via the [`MongoCursor::immortal()`](http://php.net/manual/en/mongocursor.immortal.php) option.

This is not to be confused with [`MongoCursor::$timeout`](http://php.net/manual/en/class.mongocursor.php) and [`MongoCursor::timeout()`](http://php.net/manual/en/mongocursor.timeout.php), both of which are used for setting client-side socket timeouts, which could leave the operation running on the server.

Note that the `mongodb` extension doesn't implement cursor-level socket timeouts. It's `noCursorTimeout` option corresponds to the [`find` command option of the same name](https://docs.mongodb.com/manual/reference/command/find/) and can be used to control the lifetime of a cursor on the server.

jmikola:
Note to self: the goal here is to support additional operation options (e.g. [`MongoDB\Collection::find()`](https://docs.mongodb.com/php-library/current/reference/method/MongoDBCollection-find/)) in the ODM query builder; however, we should refrain from dumping _all_ options into the already complex builder API.

jmikola:
I think we can get away with adding `array_merge()` logic to https://github.com/doctrine/mongodb-odm/blob/485d84dc05401aea58db9d41335924475b50f7f1/lib/Doctrine/ODM/MongoDB/Query/Query.php#L415 as we do for other operation types. This would allow users to specify additional `find` options in the call to `getQuery()` instead of introducing new builder methods for each option.

PR #2017, opened 2019-04-24
Allow specifying generic options for Query find operations

jmikola:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | #1808

#### Summary

Previously, find operations were unique in that they did not inherit generic options passed to the `Query::__construct()` via `Builder::getQuery()`. This makes find consistent with other operations and also adds tests for allowing query array options to override these generic options.

This also enabled users to pass driver-specific options to `Collection::find()` that may not be supported by the Builder API.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Query/Query.php
+++ b/lib/Doctrine/ODM/MongoDB/Query/Query.php
@@ -401,7 +401,7 @@ class Query implements IteratorAggregate
      *
      * @return Iterator|UpdateResult|InsertOneResult|DeleteResult|array|object|int|null
      */
-    public function runQuery()
+    private function runQuery()
     {
         $options = $this->options;
 
--- a/lib/Doctrine/ODM/MongoDB/Query/Query.php
+++ b/lib/Doctrine/ODM/MongoDB/Query/Query.php
@@ -16,10 +16,10 @@ use InvalidArgumentException;
 use IteratorAggregate;
 use MongoDB\Collection;
 use MongoDB\DeleteResult;
-use MongoDB\Driver\Cursor;
 use MongoDB\InsertOneResult;
 use MongoDB\Operation\FindOneAndUpdate;
 use MongoDB\UpdateResult;
+use Traversable;
 use UnexpectedValueException;
 use const E_USER_DEPRECATED;
 use function array_combine;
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ODM/MongoDB/Query/Query.php
```diff
@@ -347,7 +348,15 @@ static function ($value) {
         );
     }
 
-    private function makeIterator(Cursor $cursor) : Iterator
+    /**
+     * Decorate the cursor with caching, hydration, and priming behavior.
+     *
+     * Note: while this method could strictly take a MongoDB\Driver\Cursor, we
+     * accept Traversable for testing purposes since Cursor cannot be mocked.
```
jmikola: Note: The CursorInterface being introduced in 1.6.0 ([PHPC-1123](https://jira.mongodb.org/browse/PHPC-1123)) would help here, but it's likely you'll get ODM 2.0 released before that's available. I think we should consider swapping any Cursor typehints to CursorInterface for ODM 3.0, though (if you'd like to make a ticket now).

https://jira.mongodb.org/browse/PHPC-1281

tests/Doctrine/ODM/MongoDB/Tests/QueryTest.php
```diff
@@ -455,42 +456,64 @@ public function testFindAndModifyOptionsAreRenamed()
         $query->execute();
     }
 
-    public function testCountWithOptions()
+    public function testCountOptionInheritance()
     {
-        $collection = $this->getMockCollection();
+        $nearest = new ReadPreference('nearest');
+        $secondaryPreferred = new ReadPreference('secondaryPreferred');
```
jmikola: It would have been nice to refactor both of these tests to use data providers, but ReadPreference still cannot be serialized (pending [PHPC-1281](https://jira.mongodb.org/browse/PHPC-1281) for 1.6.0) and doing so might cause problems if the test suite is run with process isolation.

alcaeus: As a workaround, you could simply pass the readPreference string for now and change it once we've updated the requirements to require 1.6.0. I'll leave it up to you whether to change this.

jmikola: Ah, that's a very sensible idea. I was too hung up on passing the raw arguments to consider it :P

jmikola:
> no need to refactor the tests to use data providers unless you think it's absolutely worth it.

No need since we're only testing `count` and `find` here, but it's something we could revisit in the future if more tests are created to cover option inheritance across all operation types.

jmikola:
Revised to fix CS failures in https://travis-ci.org/doctrine/mongodb-odm/jobs/524008616. Will merge after the build passes.

Issue #2018, opened 2019-04-26
Aggregation builder $indexStats with discriminator fields mapping


aturki:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 1.2.8

#### Summary
Aggregation builder $indexStats with discriminator fields mapping

#### Current behavior
The aggregation builder fails to execute $indexStats stage when a discriminatorField is used in the class definition with error "$indexStats is only valid as the first stage in a pipeline." (ref: https://docs.mongodb.com/manual/reference/operator/aggregation/indexStats/#behavior)

#### How to reproduce
```
// Parent.php
/**
 * @ODM\InheritanceType("SINGLE_COLLECTION")
 * @ODM\DiscriminatorField("someField")
 *
 * @ODM\Document()
 *
 */

 class Parent
 {
 }

// Child.php 
/**
 * @ODM\Document()
 *
 */
 class Child extends Parent
 {
 }

// SomeService.php
$builder = $documentManager->createAggregationBuilder(Child::class);
$result = $builder->indexStats()->execute()->toArray(); // <- Fails here as the $indexStat is not the first stage of the pipeline (due to discriminator)
```

#### Expected behavior

IndexStats pipeline should be executed even when the collection contains a discriminator field. 
As a technical suggestion: maybe reordering the aggregation pipeline stages before executing them could solve this ? Any guidance on how to attack this (for a PR) is appreciated ;)

alcaeus:
Could you please post the output of `$builder->indexStats()->getPipeline()`? I suspect that the culprit is in https://github.com/doctrine/mongodb-odm/blob/4acb36955163b20e1ac3241c5fe354ea24337691/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php#L219..L221 - we presend a `$match` stage containing the appropriate filters, but may not be allowed to do so. In fact, we check for a `$geoNear` stage, so we might want to check for an `$indexStats` stage as well.

aturki:
@alcaeus, that's exactly it. Here is the output of `$builder->indexStats()->getPipeline()`
```
array (2) [
    array (1) [
        '$match' => array (1) [
            'type' => string (21) "my-discriminator-value"
        ]
    ]
    array (1) [
        '$indexStats' => stdClass (0)
    ]
]
```

PR #2019, opened 2019-04-26
Fixing  when used with discriminator field

aturki:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #2018 

#### Summary

Fixing IndexStats with used with discriminator field

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
@@ -21,6 +21,7 @@ namespace Doctrine\ODM\MongoDB\Aggregation;
 
 use Doctrine\MongoDB\Aggregation\Builder as BaseBuilder;
 use Doctrine\MongoDB\Aggregation\Stage\GeoNear;
+use Doctrine\MongoDB\Aggregation\Stage\IndexStats;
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
```diff
@@ -231,6 +231,9 @@ static function (Stage $stage) {
             $this->stages
         );
 
+        if ($this->getStage(0) instanceof Stage\IndexStats) {
```
alcaeus: I don’t think we should unset any existing stage.

lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
```diff
@@ -239,6 +242,10 @@ static function (Stage $stage) {
                 array_unshift($pipeline, ['$match' => $matchExpression]);
             }
         }
+        // Put back IndexStats stage as first element in pipeline
+        if ($this->getStage(0) instanceof Stage\IndexStats) {
```
alcaeus: This will cause wrong results, as we’re applying a match stage to the indexStats stage. I believe that if the first stage is indexStats, we should not apply any filters to the pipeline at all.

alcaeus:
Closing and reopening to nudge travis-ci to build this.

PR #2021, opened 2019-04-28
Dump xdebug filter to improve code coverage build performance

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

This makes use of a PHPUnit feature which dumps the code coverage whitelist in a format understandable by Xdebug, which in turn greatly speeds up code coverage builds. Locally, I was able to get build times with coverage from ~17 minutes down to ~11 minutes.

I've bumped the required PHPUnit version to the lowest version supporting the feature and have not checked if we may drop some more versions due to the PHP requirement.


malarzm:
Awesome, thanks @alcaeus!

PR #2023, opened 2019-04-29
Performance benchmark

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | -

#### Summary

This PR introduces a few basic performance benchmarks to compare the performance of `ext-mongo` vs. `ext-mongodb` + `mongo-php-adapter` in 1.3, and then `ext-mongodb` natively in 2.0. For now, we have three benchmarks around storing, loading and hydrating documents. I have yet to understand the reason for the change in https://github.com/doctrine/mongodb-odm/pull/1908/files#diff-2022fc4ba746afe6891859e465c4474bR219 but will add a benchmark for that as well.

At the moment, the `BaseBench` class duplicates code from the PHPUnit bootstrap - not sure if abstracting that makes sense at this time.

### DIFF
--- a/benchmark/BaseBench.php
+++ b/benchmark/BaseBench.php
@@ -1,0 +1,77 @@
+<?php
+
+namespace Doctrine\ODM\MongoDB\Benchmark;
+
+use Doctrine\Common\Cache\ArrayCache;
+use Doctrine\MongoDB\Connection;
+use Doctrine\ODM\MongoDB\Configuration;
+use Doctrine\ODM\MongoDB\DocumentManager;
+use Doctrine\ODM\MongoDB\Mapping\Driver\AnnotationDriver;
+use PhpBench\Benchmark\Metadata\Annotations\BeforeMethods;
+
+/**
+ * @BeforeMethods({"initDocumentManager", "clearDatabase"})
+ */
+abstract class BaseBench
+{
+    const DATABASE_NAME = 'doctrine_odm_performance';
+
+    /**
+     * @var DocumentManager
+     */
+    protected static $documentManager;
+
+    /**
+     * @return DocumentManager
+     */
+    protected function getDocumentManager()
+    {
+        return self::$documentManager;
+    }
+
+    public function initDocumentManager()
+    {
+        $config = new Configuration();
+
+        $config->setProxyDir(__DIR__ . '/../../tests/Proxies');
+        $config->setProxyNamespace('Proxies');
+        $config->setHydratorDir(__DIR__ . '/../../tests/Hydrators');
+        $config->setHydratorNamespace('Hydrators');
+        $config->setPersistentCollectionDir(__DIR__ . '/../../tests/PersistentCollections');
+        $config->setPersistentCollectionNamespace('PersistentCollections');
+        $config->setDefaultDB(self::DATABASE_NAME);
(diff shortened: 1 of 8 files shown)


alcaeus:
@jmikola @malarzm This is ready for now, I made sure to test some examples with embedded documents and references. Right now we check the performance of storing/loading documents as well as hydrating data. These are just basic tests, but I'm hoping that these will help me understand the impact of some of the changes we made in 2.0, but also those suggested in #1908.

We'll try to expand the performance test suite in future 2.0 work, but I don't think it's worth holding up 1.3 or 2.0 much longer just because of some tests.

malarzm:
Shall we add `benchmark` folder to `.gitattributes` so it's ignored during export?

malarzm:
Other than the previous comment this looks awesome!

benchmark/Document/HydrateDocumentBench.php
```diff
@@ -0,0 +1,134 @@
+<?php
+
+namespace Doctrine\ODM\MongoDB\Benchmark\Document;
+
+use Doctrine\ODM\MongoDB\Benchmark\BaseBench;
+use Doctrine\ODM\MongoDB\Hydrator\HydratorInterface;
+use Documents\User;
+use MongoDate;
+use MongoId;
```
jmikola: I assume this is the benchmark for `ext-mongo` and the adapter, which will be revised after merging this up to `master`.

benchmark/BaseBench.php
```diff
@@ -0,0 +1,77 @@
...
+        );
+
+        self::$documentManager = DocumentManager::create($connection, $config);
+    }
+
+    public function clearDatabase()
+    {
+        // Check if the database exists. Calling listCollections on a non-existing
+        // database in a sharded setup will cause an invalid command cursor to be
+        // returned
```
jmikola: I wouldn't expect anything less from mongos :smile:

composer.json
```diff
@@ -35,6 +36,14 @@
     "autoload": {
         "psr-0": { "Doctrine\\ODM\\MongoDB": "lib/" }
     },
+    "autoload-dev": {
```
jmikola: Why was this never needed before? Wouldn't the test suite's bootstrap file need to include `vendor/autoload.php`?

Also, I didn't spot a bootstrap file for the benchmark scripts. Is that because `vendor/bin/phpbench` knows to include `vendor/autoload.php` on its own?

benchmark/Document/HydrateDocumentBench.php
alcaeus: That is correct. I've tested this before in master to see what changes we need and will fix this once merging up to master.

benchmark/BaseBench.php
alcaeus: Yeah, that came up when we added tests against a sharded cluster. Took me a while to figure out 🙄

composer.json
alcaeus: The phpbench configuration uses the same bootstrap file as phpunit, which includes `vendor/autoload.php`. This change makes the manual adding of namespaces to the loader in said bootstrap file obsolete.

Ironically, I forgot to remove those calls and noticed while double-checking the bootstrap file to confirm that the autoloader is being used. This is now fixed.

Issue #2024, opened 2019-04-29
Query builder finds parent classes although explicitly asked to find only child documents (without discriminator map)


josefsabl:
### Bug Report

Relates to #2002, #2015.

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | dev-master

#### Summary

I followed @malarzm [comment](https://github.com/doctrine/mongodb-odm/issues/2015#issuecomment-485914361) about not needing to query by discriminator field and removed the query. And it worked well. But today, as I tested the fixed #2002 and removed the discriminator map (falling back to class name which best suits my needs) the query started returning the parent classes.

I am not sure if it is expected or a bug.

#### Current behavior

```
/**
 * NO DISCRIMINATOR MAP
 */
class A {
}

class B extends A {
}

$results = $dm->createQueryBuilder(B::class)->find()->...
```

$results contains instance of A amongst B's.

alcaeus:
This was initially fixed in #1962 but reverted in #1978 due to a regression (see #1970). I had this on my radar for 2.0 but lost track of it. I'll be providing a fix shortly.

alcaeus:
So this use-case is a bit different (and unfortunately significantly more complicated) than the original issue in #1962. For a bit of knowledge about internals, when you define a discriminator map on a document, you're basically telling ODM that there are child classes of this class that contain mapping information. ODM will not immediately load these, but it will know of them. This information is then used when you query for a class that has child classes: we go through the list of child classes and add all their discriminator values to the query. Reason why we had to revert the fix in 1.x was because you could have incomplete discriminator maps, e.g. tell ODM about some subclasses, but not all of them. Obviously, the fix breaks existing behaviour (since querying always returned all documents and suddenly only returned a subset of them).

[...]

PR #2025, opened 2019-05-02
Fix wrong usage of discriminator map in complex document inheritance chains

alcaeus:
|      Q       |   A
|------------- | -----------
| BC Break     | yes
| Fixed issues | #2024 (partially)

#### Summary

See #1962 for details. This brings the wanted functionality into 2.0 as we can't have it in 1.x due to a regression with partial discriminator maps. While #2024 wants the same behaviour for inheritance chains without discriminators, this isn't possible since the subclasses of a particular class aren't known.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -124,7 +124,7 @@ final class DocumentPersister
         $this->class           = $class;
         $this->cp              = $this->uow->getCollectionPersister();
 
-        if ($class->isMappedSuperclass || $class->isEmbeddedDocument || $class->isQueryResultDocument) {
+        if ($class->isEmbeddedDocument || $class->isQueryResultDocument) {
             return;
         }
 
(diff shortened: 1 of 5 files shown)


lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -918,12 +918,35 @@ public function prepareFieldName(string $fieldName) : string
...
+            return $preparedQuery;
+        }
+
+        if (count($discriminatorValues) === 1) {
+            $preparedQuery[$this->class->discriminatorField] = $discriminatorValues[0];
+        } else {
+            $preparedQuery[$this->class->discriminatorField] = ['$in' => $discriminatorValues];
+        }
+
+        return $preparedQuery;
```
malarzm: This `return` is inevitable, what's the point of code after it? Should it be removed?

alcaeus: That's me forgetting to clean up after myself 😉

alcaeus: Removed dead code.

malarzm:
Thanks @alcaeus!

PR #2031, opened 2019-06-09
add $preserveNullAndEmptyArrays to unwind function

Romaxx:
unwind function takes 2 parameters
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues |

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
@@ -193,10 +193,10 @@ class Builder extends BaseBuilder
     /**
      * {@inheritdoc}
      */
-    public function unwind($fieldName)
+    public function unwind($fieldName,$preserveNullAndEmptyArrays=true)
     {
         $fieldName = $this->getDocumentPersister()->prepareFieldName($fieldName);
-        return parent::unwind($fieldName);
+        return parent::unwind($fieldName,$preserveNullAndEmptyArrays);
     }
 
     /**


Steveb-p:
Thanks for the PR. However:
1. This will simply not work by introducing the change in this PR. `doctrine/mongodb` (that is currently being phased out afaik) does not contain `Builder` with `unwind` with two parameters.
https://github.com/doctrine/mongodb/blob/d4f4ea7caf7e5de6cc1ec1963d0a52429c03b129/lib/Doctrine/MongoDB/Aggregation/Builder.php#L420-L433
2. Proposed change goes against MongoDB defaults. In MongoDB `preserveNullAndEmptyArrays` defaults to `false`.
https://docs.mongodb.com/manual/reference/operator/aggregation/unwind/#pipe._S_unwind
3. Lastly, MongoDB already allows 3 options to be passed to `$unwind` operator. Using a second parameter for `preserveNullAndEmptyArrays` only would mean it will become difficult to pass all the others. Options array would be better suites for, well, options.
4. Code style...

Romaxx:
Thank you for the response, I have modified my doctrine/mongodb version long time ago so I thought it was merged.
how can I implement the stage without doctrine/mongodb ? Can I create new unwind stage in mongodb-odm ?
I think that implement the option at function level is useful when we dont want to read the docs to get all the option available for a specific stage.

alcaeus:
> Thank you for the response, I have modified my doctrine/mongodb version long time ago so I thought it was merged.

There's no PR from you in that repository. However, support for the `includeArrayIndex` and `preserveNullAndEmptyArray` options was added in https://github.com/doctrine/mongodb/pull/240, which was subsequently released as 1.3.0.

> I think that implement the option at function level is useful when we dont want to read the docs to get all the option available for a specific stage.

The options are available after adding an `unwind` stage to the builder:
```
$builder
    ->unwind('$fieldName')
        ->preserveNullAndEmptyArrays()
;
```

Since the functionality is present, I'm closing this PR.

PR #2035, opened 2019-06-25
Fix `order` syntax for YAML index examples

jmikola:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

The order syntax is only relevant for single-field indexes, where the index is specified within a field mapping. Multi-field indexes should simply specify the ordering (e.g. "asc") as a value for the corresponding key.

This came up in https://github.com/mongodb/mongo-php-library/issues/627.

This isn't relevant for `master`, since YamlDriver was removed in 2.0; however, this change should likely be merged up to the `1.3.x` branch. I'll defer to @malarzm and @alcaeus if it's worth back-porting to `1.0.x` and `1.1.x`, since the drop-down in the documentation does allow one to view content for those branches.

### DIFF
--- a/docs/en/reference/indexes.rst
+++ b/docs/en/reference/indexes.rst
@@ -136,7 +136,7 @@ For your convenience you can quickly specify a unique index with
         fields:
           username:
             unique: true
-            order: true
+            order: asc
 
 If you want to specify an index that consists of multiple fields
 you can specify them on the class doc block:
@@ -191,10 +191,8 @@ you can specify them on the class doc block:
               options:
                 unique: true
               keys:
-                accountId:
-                  order: asc
-                username:
-                  order: asc
+                accountId: asc
+                username: asc
 
 To specify multiple indexes you must use the ``@Indexes``
 annotation:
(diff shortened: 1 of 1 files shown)


malarzm:
I think we should include this even in unmaintained branches :+1: not sure if we can target PR to `1.0.x` and cleanly merge up, I'd be fine with merging this to `1.2.x` and cherry-picking to older branches.

jmikola:
This applies cleanly to 1.0.x, so I've rebased it. Will merge up through 1.3.x once it's in 1.0.x.

There were a few dangling commits in [1.0.x](https://github.com/doctrine/mongodb-odm/compare/1.1.x...1.0.x) and [1.1.x](https://github.com/doctrine/mongodb-odm/compare/1.2.x...1.1.x), due to another back-ported docs fix, but a `--strategy=ours` merge of 1.0.x into 1.1.x and 1.1.x into 1.2.x addressed that and will allow for a clean merge up after this PR is merged into 1.0.x.

PR #2038, opened 2019-07-01
Add Regex to upgrade guide

markitosgv:
Add Regex info to upgrade guide

|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

Add Regex info to upgrade guide

### DIFF
--- a/UPGRADE-2.0.md
+++ b/UPGRADE-2.0.md
@@ -264,6 +264,7 @@ ocramius. If you are checking for proxies, the following changed:
 * Querying for a mapped superclass in a complex inheritance chain will now only
   return children of that specific class instead of all classes in the
   inheritance tree.
+* The `MongoRegex` class has been deprecated, use and transform your regular expression to use `MongoDB\BSON\Regex`.
 
 ## Schema manager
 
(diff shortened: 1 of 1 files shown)


UPGRADE-2.0.md
```diff
@@ -264,6 +264,7 @@ ocramius. If you are checking for proxies, the following changed:
 * Querying for a mapped superclass in a complex inheritance chain will now only
   return children of that specific class instead of all classes in the
   inheritance tree.
+* The `MongoRegex` class is no longer exists in new driver, use and transform your regular expression to use `MongoDB\BSON\Regex`.
```
greg0ire:
```suggestion
* The `MongoRegex` class no longer exists in the new driver, use and transform your regular expression to use `MongoDB\BSON\Regex`.
```

markitosgv: thanks @greg0ire

malarzm:
I'll defer merge decision to @alcaeus as mostly he was writing the UPGRADE.md file but I think we weren't mentioning driver-related changes in the document on purpose as it's not exactly "our" upgrade path.

jmikola:
Just spoke to @alcaeus about this and we agreed that this is outside the scope of ODM. ODM itself does provide a `regex()` method in the query builder, which can be used with either a pattern string or BSON regex type to construct a `$regex` query operator.

Documentation for BSON class differences between `ext-mongo` and `ext-mongodb` would be better suited to the [Upgrade Guide](https://docs.mongodb.com/php-library/current/upgrade/) in the PHPLIB documentation, which ODM could then link to.

Issue #2047, opened 2019-08-15
$toInt aggregation operator unavailable?


arevai:
Is $toInt aggregation operator available?

jcaillot:
let's say our document has already a 'price' (decimal) field, we can extract some new fields:

```
$builder = $dm->createAggregationBuilder(OurDocument::class);

// Caution: the single quotes are absolutely required here to allow the use of the $ sign:

$builder->addFields()
             ->field("to_int")
             ->expression(  ['$toInt' => '$price' ]  )
            ->field("double")
            ->expression(  ['$multiply' => ['$price', 2 ] ]  );

$result = $builder->execute();

```
All. available operators are detailed here : https://docs.mongodb.com/manual/reference/operator/aggregation/

Steveb-p:
That would make a nice addition to docs actually.

arevai:
Awesome, thanks!

alcaeus:
Using `expression` is the preferred way to use new operators that don't have a native method yet.

As for the `$toInt` operator itself, we will work on aggregation improvements in 2.1, which includes adding all new operators and new stages added since we last worked on aggregation here.

PR #2048, opened 2019-08-16
Show how to pass expression as array

Steveb-p:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #2047

#### Summary

Show how one can pass expression as array to `AggregationBuilder` to make use of operators that did not yet receive a wrapper method.


alcaeus:
Thanks @Steveb-p!

PR #2053, opened 2019-09-03
Fix EmbedMany documentation samples

Steveb-p:
|      Q       |   A
|------------- | -----------
| BC Break     | no

#### Summary
EmbedMany uses `Collection` interface objects when loading, so to prevent issues with types I'd suggest always using `Collection`.


malarzm:
Thanks @Steveb-p!

PR #2054, opened 2019-09-10
Update sharded cluster to use replica set in background

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | 

#### Summary

With the new driver and retryable writes enabled by default, we get failure on sharded clusters backed by standalone servers. Since we only run on a single topology, changing the cluster to use replica sets fixes the issue without having to disable retryable writes by default via the connection string.

This will be merged to 2.0, but I think we may want to test additional topologies for 2.0 in the future.

### DIFF
--- a/tests/sharded.json
+++ b/tests/sharded.json
@@ -1,42 +1,134 @@
 {
-  "configsvrs": [
-    {
-      "dbpath": "$DBPATH/db27117",
-      "logpath": "$LOGPATH/configsvr27117.log",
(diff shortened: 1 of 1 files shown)


alcaeus:
Closing as I can't see what the failure is. I'll be updating the mongo-orchestration config for 1.3 and disable retryWrites for this fix.

PR #2056, opened 2019-09-10
Modernise tests

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

After fixing the build regarding retryable writes, it's time to modernise a few things here:
* PHP 7.4 build is now required to pass. This requires allowing PHPUnit 7.5, as well as fixing two instances where we didn't properly guard before an array access.
* MongoDB 3.4 is no longer tested (because it's no longer supported), MongoDB 4.2 is added as default server version
* The old mongo-orchestration setup was removed. Instead, we're using the same setup as mongo-php-library. This allows us to test against multiple topologies in the future. By default, we now test against a standalone server. There is also a test against a sharded cluster with replica sets which covers the sharding logic.


malarzm:
This is not going to 1.2?

alcaeus:
> This is not going to 1.2?

Nope. 1.3 is going to be released soon, so I don't want to spend time getting 1.2 to work.

malarzm:
All right, just need to keep in mind to merge 1.2.x up before cutting a 1.3 release :)

.travis.yml
```diff
@@ -7,42 +7,31 @@ php:
   - 7.1
   - 7.2
   - 7.3
+  - 7.4snapshot
 
 env:
   global:
-    - DOCTRINE_MONGODB_SERVER="mongodb://localhost:27017/?retryWrites=false"
+    - DEPLOYMENT=SHARDED_CLUSTER_RS
```
jmikola: Did you really mean for sharded cluster to be the default test topology (rather than a normal replica set)? I don't believe this is customized in any of the jobs below, as they only change the server version.

It looks like `SHARDED_CLUSTER`, `REPLICASET`, `STANDALONE`, and `STANDALONE_AUTH` aren't used at all.

phpunit.xml.dist
```diff
@@ -31,5 +30,8 @@
     <php>
         <const name="DOCTRINE_MONGODB_SERVER" value="mongodb://localhost:27017" />
         <const name="DOCTRINE_MONGODB_DATABASE" value="doctrine_odm_tests" />
+
```
jmikola: I'd suggest a comment noting that this is relevant for PHP 5.6 builds relying on `ext-mongo`. Perhaps you can include a link to https://stackoverflow.com/a/47578362/162228

.travis.yml
alcaeus: I didn't want to make too many changes. I picked `SHARDED_CLUSTER_RS` as it allows us to test all tests at once without running into issues with retryable writes. The other topologies will be used later once I refactor the build pipeline to test on multiple topologies.

I might give `STANDALONE` a quick test run to see if it makes the tests faster - I believe that the sharded cluster slows them down a fair bit.

phpunit.xml.dist
alcaeus: Good point, I'll add that in a separate PR 👍

.travis.yml
jmikola: Aye, if this is just about a basic sanity check, I think STANDALONE or REPLICASET would be preferred to anything sharded. REPLICASET probably mirrors what most users are running.

PR #2057, opened 2019-09-10
Add storage strategy in xsd

etshy:
For ease to code xml mapping (on PHPStorm, at least), I just added the storage strategies in xsd file
https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/1.2/reference/storage-strategies.html#storage-strategies

I don't know if this should be added into 1.1.x too.
Let me know what to do if it should.

Edit : Also, I'm not sure what doeas the `xs:restriction` element yet,so I just based my changed on the existing `reference-store-as` `simpleType`


malarzm:
> I don't know if this should be added into 1.1.x too.

Merging it to 1.2 is sufficient, 1.1 is not maintained anyway. Thanks @etshy!

PR #2062, opened 2019-09-17
Optimise build configuration

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

* Simplify usage of build steps to make this more consistent with the master branch
* Fix wrong version number in 4.0 build stage. This resulted in the default MongoDB version (currently 4.0.7) to be used on travis-ci
* Extract coverage to a separate build to speed up build times
* Add builds against all MongoDB versions supported by the PHP driver

This makes the build files for 1.3.x and master more consistent, with master then using this version of the build file and just adding the new code quality stages.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,18 +16,20 @@ env:
     - ADAPTER_VERSION="^1.0.0"
     - SERVER_DISTRO=ubuntu1604
     - SERVER_VERSION=4.2.0
+  matrix:
+    - SERVER_VERSION="3.0.15"
+    - SERVER_VERSION="3.2.22"
+    - SERVER_VERSION="3.4.23"
+    - SERVER_VERSION="3.6.14"
+    - SERVER_VERSION="4.0.12"
+    - SERVER_VERSION="4.2.0"
 
 jobs:
   include:
     # Test against lowest dependencies, including driver and server versions
     - stage: Test
       php: 5.6
-      env: DRIVER_VERSION="1.6.7" COMPOSER_FLAGS="--prefer-lowest" SERVER_VERSION="3.6.14"
-
-    # Test against MongoDB 4.0
-    - stage: Test
-      php: 7.3
-      env: SERVER_VERSION="4.0.12"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -16,32 +16,51 @@ env:
     - ADAPTER_VERSION="^1.0.0"
     - SERVER_DISTRO=ubuntu1604
     - SERVER_VERSION=4.2.0
+  matrix:
+    - SERVER_VERSION="3.0.15"
+    - SERVER_VERSION="3.2.22"
```
malarzm: 3.0 and 3.2 are EOLed, is there any reason to test them?

alcaeus: The driver still supports them, which is why I added them to the pipeline. I wouldn't go out of my way to support them, but if we can verify that tests pass without additional work then I'm good for it

.travis.yml
```diff
@@ -16,32 +16,48 @@ env:
...
     - stage: Test
       php: 7.3
       env: DEPLOYMENT=REPLICASET
+
+    # Run tests with coverage
+    - stage: Code Quality
+      php: 7.2
+      env: DEPLOYMENT=SHARDED_CLUSTER_RS
+      before_script:
+        - mv ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/xdebug.ini{.disabled,}
```
jmikola: Is this a clever one-liner to rename the `xdebug.ini.disabled` INI file to `xdebug.ini`?

Would it make sense to add `|| echo "xdebug not available"` here, as is done in `before_install` below?

alcaeus: Yep, this expands to `mv .../xdebug.ini.disabled .../xdebug.ini`. In this case, there is no `echo` as the next line fails the build if xdebug was not found, providing a more detailed message. The `mv` in `before_install` disables the xdebug extension if present, printing an info message if it wasn't available to begin with (which is not a failure condition as we don't want it there for most builds).

Issue #2032, opened 2019-06-12
Show hint what to add to discriminator map for partial discriminator maps


rieschl:
### Feature Request

`1.3-RC1` introduced deprecation warnings for the upcoming version `2.0`. One of them is a warning when a partial discriminator map is found ([e.g. here](https://github.com/doctrine/mongodb-odm/blob/1.3.x/lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php#L308)).

As discussed on Slack, , it would be nice if this warning included what should be added to the discriminator map to stay backwards compatible, i.e. it should output the FQCN of the class, exactly what is currently being added automatically. If I have existing documents with the automatically generated discriminators, I have to use the same value, because otherwise those documents would break.

alcaeus:
I finally managed to take a look at this today. The three exception cases all affect the class that is being printed in the deprecation message, or its inheritance chain. Due to the way mapping is set up, we don't know what class defines the discriminator mapping, so it's best to check the class that is printed and go up the inheritance chain if it doesn't have a discriminator map.

While looking at this, I noticed that we didn't change this for one case: associations without a target document can optionally have a discriminator map, which currently can be incomplete. I'll have to add a deprecation warning there as well, but I have to think about how to get full information there. We currently only know the field name of the association, but not the class it's contained in. Since this is public API, I have to find a backwards compatible way of passing the information to the method.

alcaeus:
Closing now that we've merged #2064. We've added what we can add without breaking BC. Unfortunately, there are a couple of instances where we can't show you all the info, but in general you should be able to deduce what's going on here.

PR #2064, opened 2019-09-18
Add missing deprecation notice for incomplete embedded associations

alcaeus:
While looking at #2032, I realised that we don't trigger a deprecation warning when using an incomplete discriminator map for an embedded relationship. This PR rectifies this.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
@@ -410,17 +410,15 @@ class PersistenceBuilder
          */
         if ( ! isset($embeddedMapping['targetDocument'])) {
             $discriminatorField = $embeddedMapping['discriminatorField'];
-            $discriminatorValue = isset($embeddedMapping['discriminatorMap'])
-                ? array_search($class->name, $embeddedMapping['discriminatorMap'])
-                : $class->name;
+            if (! empty($embeddedMapping['discriminatorMap'])) {
+                $discriminatorValue = array_search($class->name, $embeddedMapping['discriminatorMap']);
 
-            /* If the discriminator value was not found in the map, use the full
-             * class name. In the future, it may be preferable to throw an
-             * exception here (perhaps based on some strictness option).
-             *
-             * @see DocumentManager::createDBRef()
-             */
-            if ($discriminatorValue === false) {
+                if ($discriminatorValue === false) {
+                    @trigger_error(sprintf('Document class "%s" is unlisted in the discriminator map for embedded field "%s". This is deprecated and will throw an exception in doctrine/mongodb-odm 2.0.', $class->name, $mapping['name']), E_USER_DEPRECATED);
+
+                    $discriminatorValue = $class->name;
+                }
+            } else {
                 $discriminatorValue = $class->name;
             }
 


malarzm:
No test required adjustment or will you be doing that while merging up when an exception is thrown?

alcaeus:
`master` already throws an exception, I guess we just missed adding the deprecation. We didn't have many tests with partial discriminator maps, which is why there are no test changes.

lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
```diff
@@ -410,17 +410,15 @@ public function prepareEmbeddedDocumentValue(array $embeddedMapping, $embeddedDo
...
-
-            /* If the discriminator value was not found in the map, use the full
-             * class name. In the future, it may be preferable to throw an
-             * exception here (perhaps based on some strictness option).
-             *
-             * @see DocumentManager::createDBRef()
-             */
-            if ($discriminatorValue === false) {
+            if (! empty($embeddedMapping['discriminatorMap'])) {
+                $discriminatorValue = array_search($class->name, $embeddedMapping['discriminatorMap']);
```
jmikola: Should we use strict mode here? I imagine that may be something `master` might already sniff for.

lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
```diff
@@ -410,17 +410,15 @@ public function prepareEmbeddedDocumentValue(array $embeddedMapping, $embeddedDo
...
-                : $class->name;
-
-            /* If the discriminator value was not found in the map, use the full
-             * class name. In the future, it may be preferable to throw an
-             * exception here (perhaps based on some strictness option).
-             *
-             * @see DocumentManager::createDBRef()
-             */
-            if ($discriminatorValue === false) {
+            if (! empty($embeddedMapping['discriminatorMap'])) {
```
jmikola: Any concern using `empty()` here in lieu of `isset()`? I suppose ClassMetadata wouldn't allow this to get a falsey value otherwise, so maybe there's no practical difference.

lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
alcaeus: We currently don't use strict mode for any `array_search` calls and apparently don't have a phpcs sniff for this either. I've created #2067 to track and implement this, pending potential BC issues.

lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
alcaeus: While we can't guarantee types in the mapping due to public properties and using arrays for association mappings, we always assume the discriminator map to be an array, even when no map is in use. That's why we use `empty` in all places where we check for a discriminator map being used instead of `isset` as that would cause false positives in classes with no discriminator map.

Since ClassMetadata is being changed for ORM 3.0, I think we'll go a similar route for ODM 3.0 and use stricter types for all kinds of mappings, in which case we can revisit this and eventually not store an empty array for a discriminator map, but rather `null` (and also contain that logic within whatever class stores discriminator information so we don't have to do this check in multiple places).

PR #2065, opened 2019-09-19
Add command example to create a capped collection

melkamar:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

This PR adds the command to use when using Doctrine with Symfony to create a _capped collection_. 

I spent some time on debugging what exactly needs to be done to have Doctrine listen to the capped collection configuration and it was missing from the documentation.

### DIFF
--- a/docs/en/reference/capped-collections.rst
+++ b/docs/en/reference/capped-collections.rst
@@ -90,3 +90,13 @@ You can drop the collection too if it already exists:
     <?php
 
     $documentManager->getSchemaManager()->dropDocumentCollection('Category');
+
+Use the ``odm:schema:create`` command to create the database schema:
+
+.. code-block:: console
+
+    $ php mongodb.php odm:schema:create
+
+It must not exist when the command is invoked. It is not possible to convert an existing collection to a capped one using Doctrine. It is, however, possible to convert it by calling a `direct Mongo command`_.
+
+.. _`direct Mongo command`: https://docs.mongodb.com/manual/core/capped-collections/#convert-a-collection-to-capped


docs/en/reference/capped-collections.rst
```diff
@@ -90,3 +90,13 @@ You can drop the collection too if it already exists:
     <?php
 
     $documentManager->getSchemaManager()->dropDocumentCollection('Category');
+
+If using Symfony in addition to Doctrine, use the following command to create the capped collection:
```
alcaeus: The console commands are available even if not using Symfony, but people need to add them to their console application themselves. I'd reword this:
```suggestion
Use the ``odd:schema:create`` command to create the database schema:
```

docs/en/reference/capped-collections.rst
```diff
@@ -90,3 +90,13 @@ You can drop the collection too if it already exists:
...
+
+.. code-block:: bash
```
alcaeus: Elsewhere, we use `console` instead of `bash`
```suggestion
.. code-block:: console
```

docs/en/reference/capped-collections.rst
```diff
@@ -90,3 +90,13 @@ You can drop the collection too if it already exists:
...
+
+    php app/console --env=<env> doctrine:mongodb:schema:create
```
alcaeus: This is how we show console examples elsewhere, as we don't assume people to use the Symfony Bundle:
```suggestion
    $ php mongodb.php odm:schema:create
```

docs/en/reference/capped-collections.rst
Steveb-p: Watch out for typo in ``odd:schema:create`` in suggested changes ;)

alcaeus: Whoops. Thanks @Steveb-p, I've fixed it.

melkamar:
Applied the suggestions, thanks for them!

alcaeus:
@melkamar can you squash commits? Happy to merge afterwards!

alcaeus:
Thanks @melkamar!

PR #2068, opened 2019-09-26
fix an error in the XML schema

rieschl:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

#### Summary

A small bubu in the XML mapping schema file caused `\DOMDocument::schemaValidate` to fail and `\Doctrine\ODM\MongoDB\Mapping\Driver\XmlDriver::validateSchema` therefore triggers a deprecation warning.

This is already fixed in the `2.0.0-RC2` branch.

### DIFF
--- a/doctrine-mongo-mapping.xsd
+++ b/doctrine-mongo-mapping.xsd
@@ -447,7 +447,7 @@
 
   <xs:complexType name="read-preference-tag-set">
     <xs:choice minOccurs="0" maxOccurs="unbounded">
-      <xs:element name="tag" type="read-preference-tag" minOccurs="0" maxOccurs="unbounded" />
+      <xs:element name="tag" type="odm:read-preference-tag" minOccurs="0" maxOccurs="unbounded" />
     </xs:choice>
   </xs:complexType>
 


doctrine-mongo-mapping.xsd
```diff
@@ -447,7 +447,7 @@
 
   <xs:complexType name="read-preference-tag-set">
     <xs:choice minOccurs="0" maxOccurs="unbounded">
-      <xs:element name="tag" type="read-preference-tag" minOccurs="0" maxOccurs="unbounded" />
+        <xs:element name="tag" type="odm:read-preference-tag" minOccurs="0" maxOccurs="unbounded"/>
```
malarzm: Could you please remove these extra spaces (and amend original commit)?

rieschl: Ah sure. Sorry. Stupid PhpStorm autoformatting ;)

rieschl: I just saw the space before the closing tag. It seems that it's not used consistently. I just amended again to keep the old format in this line (space)

malarzm:
Thanks a lot @rieschl!

PR #2069, opened 2019-09-30
Class name resolver performance optimisation

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | 

#### Summary

cc @watari

#1908 suggests a number of changes to improve document performance. I was especially curious about the changes to the document manager, where an additional cache for `ClassMetadata` is added. Looking at that change in isolation, it turns out that most of the improvements to the performance test suite come from that single change.

Upon further investigation, I was able to pinpoint the performance regression to the class name resolution for proxy objects, which now was added to MongoDB ODM until the doctrine/persistence layer supports multiple proxy implementations.

This PR removes this issue by adding a caching version of a class name resolver which removes the need for repeatedly resolving the same class names. Since the original resolver was not marked as `@internal`, it is only deprecated and not removed entirely. The new implementations are marked as `@internal` as users shouldn't need to instantiate those directly. The associated interface is not marked as `@internal` and is covered by the BC promise.

I'll take a look at the other changes in #1908 in separate PRs or merge it directly with the DocumentManager change removed.

For reference, the following times are from before and after the change:
```
⅀T: 9,645.960μs μSD/r 0.000μs μRSD/r: 0.000%
⅀T: 8,144.470μs μSD/r 0.000μs μRSD/r: 0.000%
```

This is an improvement of ~15.5% just for this change alone!

### DIFF
--- a/UPGRADE-2.0.md
+++ b/UPGRADE-2.0.md
@@ -231,8 +231,13 @@ ocramius. If you are checking for proxies, the following changed:
   resolver from the document manager:
   ```php
   $dm->getClassNameResolver()->getRealClass($className);
-  $dm->getClassNameResolver()->getClass($object);
   ```
+* The return value of the `getClassNameResolver` method was updated in
+  2.0.0. The previously returned
+  `Doctrine\ODM\MongoDB\Proxy\ClassNameResolver` class was dropped in favour of
+  the `Doctrine\ODM\MongoDB\Proxy\Resolver\ClassNameResolver` interface. This
+  BC break was necessary to mitigate a performance regression. The `getClass`
+  method was dropped from the interface as it wasn't being used.
 
 ## Repository
 
--- a/lib/Doctrine/ODM/MongoDB/DocumentManager.php
+++ b/lib/Doctrine/ODM/MongoDB/DocumentManager.php
@@ -13,6 +13,7 @@ use Doctrine\ODM\MongoDB\Mapping\ClassMetadataFactory;
 use Doctrine\ODM\MongoDB\Mapping\MappingException;
 use Doctrine\ODM\MongoDB\Proxy\Factory\ProxyFactory;
 use Doctrine\ODM\MongoDB\Proxy\Factory\StaticProxyFactory;
+use Doctrine\ODM\MongoDB\Proxy\Resolver\CachingClassNameResolver;
 use Doctrine\ODM\MongoDB\Proxy\Resolver\ClassNameResolver;
 use Doctrine\ODM\MongoDB\Proxy\Resolver\ProxyManagerClassNameResolver;
 use Doctrine\ODM\MongoDB\Query\FilterCollection;
@@ -184,7 +185,7 @@ class DocumentManager implements ObjectManager
         $this->schemaManager     = new SchemaManager($this, $this->metadataFactory);
         $this->proxyFactory      = new StaticProxyFactory($this);
         $this->repositoryFactory = $this->config->getRepositoryFactory();
-        $this->classNameResolver = new ProxyManagerClassNameResolver($this->config);
+        $this->classNameResolver = new CachingClassNameResolver(new ProxyManagerClassNameResolver($this->config));
     }
 
     /**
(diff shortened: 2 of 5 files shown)


lib/Doctrine/ODM/MongoDB/DocumentManager.php
```diff
@@ -11,9 +11,11 @@
 use Doctrine\ODM\MongoDB\Mapping\ClassMetadata;
 use Doctrine\ODM\MongoDB\Mapping\ClassMetadataFactory;
 use Doctrine\ODM\MongoDB\Mapping\MappingException;
-use Doctrine\ODM\MongoDB\Proxy\ClassNameResolver;
 use Doctrine\ODM\MongoDB\Proxy\Factory\ProxyFactory;
 use Doctrine\ODM\MongoDB\Proxy\Factory\StaticProxyFactory;
+use Doctrine\ODM\MongoDB\Proxy\Resolver\CachingClassNameResolver;
+use Doctrine\ODM\MongoDB\Proxy\Resolver\ClassNameResolver;
```
malarzm: This still is a BC break. Consider userland code `function foo(Doctrine\ODM\MongoDB\Proxy\ClassNameResolver\ClassNameResolved $resolver)` which then was executed with `foo($dm->getClassNameResolver());`. With current change this will break as document manager is now returning an instance of `Doctrine\ODM\MongoDB\Proxy\Resolver\ClassNameResolver`. I think we just need to break the compatibility without false promise of not removing old class to keep compatibility.

alcaeus: You've got a point. I can also integrate the functionality into the original resolver, just thought it'd make more sense to abstract this out :|

malarzm: Yeah, interface and caching decorator were nice and normally I'd call it way to go :)

alcaeus: Changes made and BC break documented in UPGRADE-2.0 document. I've also added the `BC Break` label to this PR so it shows up properly in the changelog. Will release RC-3 once I'm done with performance improvements.

lib/Doctrine/ODM/MongoDB/Proxy/Resolver/CachingClassNameResolver.php
```diff
@@ -0,0 +1,44 @@
...
+            $this->resolvedNames[$class] = $this->resolver->getRealClass($class);
+        }
+
+        return $this->resolvedNames[$class];
+    }
+
+    /**
+     * Gets the real class name of an object (even if its a proxy).
+     */
+    public function getClass(object $object) : string
```
stof: should we even keep this method in the interface ? The implementation is always `$this->getRealClass(get_class($object))` to benefit from the decorated behavior of `getRealClass`.

alcaeus: I just checked and we don't even use it. @malarzm what do you think? I'd have no problem dropping it TBH

malarzm: Good point, calling `\get_class` on one's own is easy enough, I think we can drop it 👍

alcaeus: Removed the method. Thanks @stof!

UPGRADE-2.0.md
```diff
@@ -231,8 +231,13 @@ ocramius. If you are checking for proxies, the following changed:
...
   ```php
   $dm->getClassNameResolver()->getRealClass($className);
-  $dm->getClassNameResolver()->getClass($object);
   ```
+* The return value of the `getClassNameResolver` method was updated in
+  2.0.0. The previously returned
+  `Doctrine\ODM\MongoDB\Proxy\ClassNameResolver` class was dropped in favour of
+  the `Doctrine\ODM\MongoDB\Proxy\Resolver\ClassNameResolver` interface. This
+  BC break was necessary to mitigate a performance regression. The `getClass`
+  method was dropped from the interface as it wasn't being used.
```
jmikola: Out of curiosity, how was `getClass()` different from `getRealClass()`?

EDIT: I see below it was just a convenience method that accepted an `object` instead of a class string. Good riddance!

Issue #2044, opened 2019-07-12
Type error when data from database is incompatible with mapping


alcaeus:
This is the root cause for the issue described in https://github.com/doctrine/DoctrineMongoDBBundle/issues/564: a property was mapped as an embedded document, but the data received from the database was a scalar value. In order to avoid type errors, the hydration logic needs to better assert expected types.

malarzm:
FWIW have no idea we can fail gracefully if we expect an object and get a scalar. Unless more gracefully as throw an exception instead of a type error? :)

alcaeus:
Fixed in #2073

PR #2073, opened 2019-10-01
Throw hydrator exceptions when encountering invalid types

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #2044 

#### Summary

This adds type checks for associations to the hydrator, to prevent errors when receiving invalid data from the database. For most associations, this resulted in a `TypeError` being thrown, but for references this could result in invalid offsets when accessing the reference "array".

To fix this problem, there are new `HydratorException` methods to account for these errors. We print the class name, field name of the offending data, as well as the expected and actual types. Since we are in the middle of hydration, we can't reliably print a full field path (we don't know if this is a root hydration or already within an embedded document), which also means that we can't print an identifier for the document being hydrated. However, the exception message better explains the problem than a PHP error does. More importantly, this makes it look less like a bug in ODM and more like an issue with the data (which it actually is).

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
+++ b/lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
@@ -245,6 +245,11 @@ EOF
         /** @ReferenceOne */
         if (isset(\$data['%1\$s'])) {
             \$reference = \$data['%1\$s'];
+
+            if (\$this->class->fieldMappings['%2\$s']['storeAs'] !== ClassMetadata::REFERENCE_STORE_AS_ID && ! is_array(\$reference)) {
+                throw HydratorException::associationTypeMismatch('%3\$s', '%1\$s', 'array', gettype(\$reference));
+            }
+
             \$className = \$this->unitOfWork->getClassNameForAssociation(\$this->class->fieldMappings['%2\$s'], \$reference);
             \$identifier = ClassMetadata::getReferenceId(\$reference, \$this->class->fieldMappings['%2\$s']['storeAs']);
             \$targetMetadata = \$this->dm->getClassMetadata(\$className);
@@ -257,7 +262,8 @@ EOF
 EOF
                     ,
                     $mapping['name'],
-                    $mapping['fieldName']
+                    $mapping['fieldName'],
+                    $class->getName()
                 );
             } elseif ($mapping['association'] === ClassMetadata::REFERENCE_ONE && $mapping['isInverseSide']) {
                 if (isset($mapping['repositoryMethod']) && $mapping['repositoryMethod']) {
@@ -305,6 +311,11 @@ EOF
 
         /** @Many */
         \$mongoData = isset(\$data['%1\$s']) ? \$data['%1\$s'] : null;
+
+        if (\$mongoData !== null && ! is_array(\$mongoData)) {
+            throw HydratorException::associationTypeMismatch('%3\$s', '%1\$s', 'array', gettype(\$mongoData));
+        }
+
         \$return = \$this->dm->getConfiguration()->getPersistentCollectionFactory()->create(\$this->dm, \$this->class->fieldMappings['%2\$s']);
         \$return->setHints(\$hints);
         \$return->setOwner(\$document, \$this->class->fieldMappings['%2\$s']);
@@ -318,7 +329,8 @@ EOF
 EOF
                     ,
                     $mapping['name'],
-                    $mapping['fieldName']
+                    $mapping['fieldName'],
+                    $class->getName()
                 );
             } elseif ($mapping['association'] === ClassMetadata::EMBED_ONE) {
                 $code .= sprintf(
@@ -327,6 +339,11 @@ EOF
         /** @EmbedOne */
         if (isset(\$data['%1\$s'])) {
             \$embeddedDocument = \$data['%1\$s'];
+
+            if (! is_array(\$embeddedDocument)) {
+                throw HydratorException::associationTypeMismatch('%3\$s', '%1\$s', 'array', gettype(\$embeddedDocument));
+            }
+        
             \$className = \$this->unitOfWork->getClassNameForAssociation(\$this->class->fieldMappings['%2\$s'], \$embeddedDocument);
             \$embeddedMetadata = \$this->dm->getClassMetadata(\$className);
             \$return = \$embeddedMetadata->newInstance();
@@ -347,7 +364,8 @@ EOF
 EOF
                     ,
                     $mapping['name'],
-                    $mapping['fieldName']
+                    $mapping['fieldName'],
+                    $class->getName()
                 );
             }
         }
@@ -360,6 +378,7 @@ EOF
 namespace $namespace;
 
 use Doctrine\ODM\MongoDB\DocumentManager;
+use Doctrine\ODM\MongoDB\Hydrator\HydratorException;
 use Doctrine\ODM\MongoDB\Hydrator\HydratorInterface;
 use Doctrine\ODM\MongoDB\Query\Query;
 use Doctrine\ODM\MongoDB\UnitOfWork;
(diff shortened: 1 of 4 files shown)


lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
```diff
@@ -245,6 +245,11 @@ private function generateHydratorClass(ClassMetadata $class, string $hydratorCla
         /** @ReferenceOne */
         if (isset(\$data['%1\$s'])) {
             \$reference = \$data['%1\$s'];
+
+            if (\$this->class->fieldMappings['%2\$s']['storeAs'] !== ClassMetadata::REFERENCE_STORE_AS_ID && ! is_array(\$reference)) {
+                throw HydratorException::associationTypeMismatch('%3\$s', '%1\$s', 'array', gettype(\$reference));
```
jmikola: I was originally confused as to the value of `%3\$s`, and assumed it was literal PHP code based on the previous usage. Further inspection revealed that this is in a completely separate `sprintf()` call and it's the class name string, which is also added 18 lines below. :+1:

It's a shame we can't use more meaningful labels for these things.

lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
```diff
@@ -245,6 +245,11 @@ private function generateHydratorClass(ClassMetadata $class, string $hydratorCla
         /** @ReferenceOne */
         if (isset(\$data['%1\$s'])) {
             \$reference = \$data['%1\$s'];
+
+            if (\$this->class->fieldMappings['%2\$s']['storeAs'] !== ClassMetadata::REFERENCE_STORE_AS_ID && ! is_array(\$reference)) {
```
jmikola: If I understand correctly, `REFERENCE_STORE_AS_ID` is the only strategy that stores an arbitrary value. All others nest the reference in an embedded document (either a DBRef or ODM's newer style that omits the $-prefixed field names), correct?

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -664,15 +666,24 @@ private function loadEmbedManyCollection(PersistentCollectionInterface $collecti
         $embeddedDocuments = $collection->getMongoData();
         $mapping           = $collection->getMapping();
         $owner             = $collection->getOwner();
+
         if (! $embeddedDocuments) {
             return;
         }
 
+        if ($owner === null) {
+            throw PersistentCollectionException::ownerRequiredToLoadCollection();
```
jmikola: This exception already existed but just wasn't used here?

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -693,12 +704,22 @@ private function loadReferenceManyCollectionOwningSide(PersistentCollectionInter
...
+        }
+
         $sorted = isset($mapping['sort']) && $mapping['sort'];
 
         foreach ($collection->getMongoData() as $key => $reference) {
-            $className  = $this->uow->getClassNameForAssociation($mapping, $reference);
+            $className = $this->uow->getClassNameForAssociation($mapping, $reference);
```
jmikola: Noting this is just a whitespace change.

tests/Doctrine/ODM/MongoDB/Tests/HydratorTest.php
```diff
@@ -65,6 +67,92 @@ public function testReadOnly()
...
+    public function testEmbedOneWithWrongType()
+    {
+        $user = new HydrationClosureUser();
+
+        $this->expectException(HydratorException::class);
+        $this->expectExceptionMessage('Expected association for field "embedOne" in document of type "' . HydrationClosureUser::class . '" to be of type "array", "string" received.');
+
+        $this->dm->getHydratorFactory()->hydrate($user, [
+            '_id' => 1,
+            'embedOne' => 'jon',
```
jmikola: Glad to see we're still using @jwage as a data fixture.

tests/Doctrine/ODM/MongoDB/Tests/HydratorTest.php
```diff
@@ -65,6 +67,92 @@ public function testReadOnly()
...
+    public function testReferenceOneWithWrongType()
+    {
+        $user = new HydrationClosureUser();
+
+        $this->expectException(HydratorException::class);
+        $this->expectExceptionMessage('Expected association for field "referenceOne" in document of type "' . HydrationClosureUser::class . '" to be of type "array", "string" received.');
+
+        $this->dm->getHydratorFactory()->hydrate($user, [
+            '_id' => 1,
+            'referenceOne' => 'jon',
```
jmikola: Out of curiosity, do we bother testing for malformed DBRef or ODM-style reference objects? I realize that's probably out of scope for this PR.

lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
malarzm: Correct :)

tests/Doctrine/ODM/MongoDB/Tests/HydratorTest.php
malarzm: I believe we have no tests for such scenarios

lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
alcaeus: Yeah, that is very unfortunate. We may be able to improve this using Heredoc, but that's a change for a different release.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
alcaeus: Yep. This was introduced whenever we require the owner to be set (PHPStan found a few instances where we assumed it would be set). Since we now need the owner to generate the correct exception message, we have to also throw it here.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
alcaeus: Yep - assignment operator alignment strikes again. See https://github.com/doctrine/coding-standard/pull/107.

tests/Doctrine/ODM/MongoDB/Tests/HydratorTest.php
alcaeus: The beauty of copy/paste 😅

tests/Doctrine/ODM/MongoDB/Tests/HydratorTest.php
alcaeus: We don't have these, but adding them would probably make sense. Although I'd argue that we may want to be less strict when reading these references (e.g., accept `['id' => 'foo', 'collection' => 'foo']`) when reading a `DBRef` object.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
malarzm: Oh yeah, I totally forgot about that PR

tests/Doctrine/ODM/MongoDB/Tests/HydratorTest.php
malarzm: On one hand it'd be nice but on the other it could give false impression that changing one storage strategy to another works "just like that" which can bite back when using aggregations

jmikola: Opened https://github.com/doctrine/mongodb-odm/issues/2077 to track this.

PR #2076, opened 2019-10-02
Use expectException(Message) instead of annotations

olvlvl:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

The `@expectedException` and `@expectedExceptionMessage` annotations are deprecated. They will be removed in PHPUnit 9. This PR replaces `@expectedException` and `@expectedExceptionMessage` annotations with `expectedException()` and `expectedExceptionMessage()`.

My only concern is with [OutTest](https://github.com/doctrine/mongodb-odm/blob/a007696489d09435c332bec34140ca7310287785/tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/OutTest.php#L44) because `builder->out()` is throwing the exception, not `builder->getPipeline()`.

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/Aggregation/ExprTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Aggregation/ExprTest.php
@@ -7,6 +7,7 @@ namespace Doctrine\ODM\MongoDB\Tests\Aggregation;
 use BadMethodCallException;
 use Doctrine\ODM\MongoDB\Aggregation\Expr;
 use Doctrine\ODM\MongoDB\Tests\BaseTest;
+use LogicException;
 
 class ExprTest extends BaseTest
 {
@@ -68,9 +69,6 @@ class ExprTest extends BaseTest
         );
     }
 
-    /**
-     * @expectedException \LogicException
-     */
     public function testExpressionWithoutField()
     {
         $nestedExpr = $this->createExpr();
@@ -82,6 +80,7 @@ class ExprTest extends BaseTest
 
         $expr = $this->createExpr();
 
+        $this->expectException(LogicException::class);
         $expr->expression($nestedExpr);
     }
 
--- a/tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/OutTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/OutTest.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Doctrine\ODM\MongoDB\Tests\Aggregation\Stage;
 
+use Doctrine\ODM\MongoDB\Mapping\MappingException;
 use Doctrine\ODM\MongoDB\Tests\BaseTest;
 use Documents\Sharded\ShardedUser;
 use Documents\SimpleReferenceUser;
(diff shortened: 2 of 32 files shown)


malarzm:
> My only concern is with OutTest because builder->out() is throwing the exception, not builder->getPipeline().

I think that `getPipeline()` call could be removed in this case

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/ExprTest.php
```diff
@@ -82,6 +79,7 @@ public function testExpressionWithoutField()
 
         $expr = $this->createExpr();
 
+        $this->expectException(\LogicException::class);
```
malarzm: Personally I prefer putting expectException as a first thing in a test but I'm not sure what's the preference of us as a team :)

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/OutTest.php
```diff
@@ -38,16 +39,13 @@ public function testOutStageWithCollectionName()
     }
 
     /**
-     * @expectedException \Doctrine\ODM\MongoDB\Mapping\MappingException
      * @expectedExceptionMessage Cannot use class 'Documents\Sharded\ShardedUser' as collection for out stage.
```
malarzm: `@expectedExceptionMessage` is not deprecated? Even if not I think that if we switch to methods for expecting exception, we shouldn't keep annotations for the message assertion

olvlvl: I wanted this PR to focus on `@expectedException` but I can update the other annotations too.

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/ExprTest.php
olvlvl: We put the `expectException()` just before the thing that is expected to throw an exception. Let me know what to do.

alcaeus: Your usage is correct - test setup should not trigger the expected exception by accident. Please leave it as is 👍

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/OutTest.php
greg0ire: I think that's the only other type of annotation that needs to be converted to code, isn't it?

olvlvl: I added a commit for `@expectedExceptionMessage`.

olvlvl:
@greg0ire Do you want me to do the squash, or will you squash+merge?

malarzm:
@olvlvl please squash, we're not using squash+merge as it's not doing an actual merge (or it wasn't doing this last time we were checking)

greg0ire:
Also, we would rather have you be in charge of the final commit message ;)

olvlvl:
Squashed.

greg0ire:
Thanks @olvlvl !

PR #2078, opened 2019-10-04
Use alternatives for features deprecated in PHPUnit 8

olvlvl:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

- The method `assertInternalType()` is deprecated, use alternatives such as `assertIsInt()`. https://github.com/sebastianbergmann/phpunit/issues/3369
- The `$delta` parameter of `assertEquals()` is deprecated, use `assertWithDelta()` instead. https://github.com/sebastianbergmann/phpunit/issues/3341

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/EmbeddedTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/EmbeddedTest.php
@@ -615,7 +615,7 @@ class EmbeddedTest extends BaseTest
             ->getQuery()
             ->getSingleResult();
 
-        $this->assertInternalType('array', $test1Data);
+        $this->assertIsArray($test1Data);
 
         $this->assertArrayHasKey('m_id', $test1Data['embedOne']);
     }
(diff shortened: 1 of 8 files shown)


olvlvl:
🤔 This test was not modified.

```
1) Doctrine\ODM\MongoDB\Tests\Functional\LockTest::testLockTimestampThrowsException

Failed asserting that exception of type "Doctrine\ODM\MongoDB\LockException" is thrown.
```

greg0ire:
Should you really target master since there is no BC-break (disclaimer, I don't know where we are regarding the merge down vs merge up policy, but IIRC only the ORM still uses a merge down policy)

alcaeus:
We definitely merge up in ODM. I’m not going to update 2.0 for PHPUnit 9, so this is fine for master.

As for the failing test, this one fails once in a while. I’ve restarted the offending build.

greg0ire:
Still failing. My turn to restart it :)

malarzm:
Thanks @olvlvl!

PR #2082, opened 2019-10-08
Fix wrong initialisation of proxies with public properties

### LINKED ISSUE #2080: Undefined property error from proxy when EmbedOne field is not set
### BC Break Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.0.0

#### Summary
When an EmbedOne field is not set in database, it now causes an `Undefined property` error.

#### Previous behavior
Previously in 1.2.x the field was simply left `null`.

#### Current behavior
An `Undefined property` is caused by generated proxy `__get($name)` method.

#### How to reproduce

For instance, I've got 2 models `Lab` and `Address`:

```php
<?php

namespace App\Doctrine;

/**
 * @ODM\Document(collection="labs")
 */
class Lab extends BaseModel
{
    /**
     * @ODM\Id
     * @var string
     */
    public $_id;
    /**
     * @ODM\EmbedOne(targetDocument=Address::class)
     * @var Address
     */
    public $address;
}
```
```php
<?php

namespace App\Doctrine;

/** @ODM\EmbeddedDocument */
class Address extends BaseModel
{
    /**
     * @ODM\Field(type="string")
     * @var string
     */
    public $country;
}
```
Now if I insert a new `Lab` document without touching the `address` field, it's also not set in the database as it used to be. However, when I retrieve the document, in the hydrator it's got
```php
    public function hydrate(object $document, array $data, array $hints = array()): array
    {
        /** @EmbedOne */
        if (isset($data['address'])) {
            // omitted
        }
    }
```
...and in the proxy it's got
```php
    /**
     * @var \Closure|null initializer responsible for generating the wrapped object
     */
    private $initializerc3072 = null;

    /**
     * @var bool tracks initialization status - true while the object is initializing
     */
    private $initializationTrackerce44e = false;

    /**
     * @var bool[] map of public properties of the parent class
     */
    private static $publicPropertiesda3cc = [
        'address' => true,
    ];

    public function __get($name)
    {
        $this->initializerc3072 && ! $this->initializationTrackerce44e && $this->callInitializer728ea('__get', array('name' => $name));

        if (isset(self::$publicPropertiesda3cc[$name])) {
            return $this->$name;  // BREAKS HERE
        }
        // omitted
    }
```
Since the `address` field is not set in the hydrator, when the proxy tries to get its value `return $this->$name` it causes an `Undefined property` error.

PS: Not sure whether it's relevant but `BaseModel` is
```php
<?php

namespace App\Doctrine;

use Carbon\Carbon;
use MyCLabs\Enum\Enum;

class BaseModel
{
    protected $excludedFromArray = ['_id', 'lazyPropertiesDefaults'];
    private $preserveOriginalName = false;

    public function __construct(array $data = null)
    {
        if (!empty($data)) {
            foreach ($data as $key => $value) {
                $this->$key = $value;
            }
        }
    }

    public function __get($name)
    {
        $accessor = 'get_' . $this->snakeCase($name);
        if (method_exists($this, $accessor)) {
            return $this->$accessor();
        }
        return $this->$name ?? null;
    }

    public function __set($name, $value)
    {
        $mutator = 'set_' . $this->snakeCase($name);
        if (method_exists($this, $mutator)) {
            // an additional bool parameter ($toRemove = false) is provided
            $this->$mutator($value, false);
        } else {
            $this->$name = $value;
        }
    }

    public function __unset($name)
    {
        $this->__set($name, null);
    }

    public function __isset($name)
    {
        return $this->__get($name) !== null;
    }

    public function __clone()
    {
        foreach (get_object_vars($this) as $key => $value) {
            $this->$key = $this->cloneRecursively($value);
        }
    }

    /**
     * @return array
     */
    public function getExcludedFromArray(): array
    {
        return $this->excludedFromArray;
    }

    /**
     * @param array $excludedFromArray
     */
    public function setExcludedFromArray(array $excludedFromArray): void
    {
        $this->excludedFromArray = $excludedFromArray;
    }

    /**
     * @param mixed $item
     * @return mixed
     */
    private function cloneRecursively($item)
    {
        if (is_object($item)) {
            return clone $item;
        }
        if (is_array($item)) {
            return array_map([self::class, 'cloneRecursively'], $item);
        }
        return $item;
    }

    public function toArray(): array
    {
        $reflect = new \ReflectionClass($this);
        $result = [];
        foreach ($reflect->getProperties(\ReflectionProperty::IS_PUBLIC) as $key) {
            $key = $key->getName();
            if (strpos($key, '__') === 0
                || in_array($key, $this->getExcludedFromArray(), true)
                || method_exists($this, 'get_' . $this->snakeCase($key))) {
                continue;
            }
            $result[$key] = $this->toArrayRecursively($this->$key);
        }
        foreach ($reflect->getMethods() as $method) {
            $method = $method->getName();
            if (preg_match('/^get_(.+)$/', $method, $matches)) {
                $key = $matches[1];
                if (strpos($key, '__') === 0 || in_array($key, $this->getExcludedFromArray(), true)) {
                    continue;
                }
                $result[$key] = $this->toArrayRecursively($this->{$matches[0]}());
            }
        }
        return $result;
    }

    /**
     * @param mixed $item
     * @return mixed
     */
    private function toArrayRecursively($item)
    {
        if ($item instanceof Carbon) {
            return $item->format('c');
        }
        if ($item instanceof \DateTimeInterface) {
            return Carbon::instance($item)->format('c');
        }
        if ($item instanceof Enum) {
            return $item->getValue();
        }
        if ($item instanceof \IteratorAggregate) {
            $item = $item->getIterator();
        }
        if ($item instanceof \Iterator) {
            $item = iterator_to_array($item);
        }
        if (is_object($item) && method_exists($item, 'toArray')) {
            return $item->toArray();
        }
        if (is_array($item)) {
            return array_map([self::class, 'toArrayRecursively'], $item);
        }
        return $item;
    }

    protected function snakeCase($name): string
    {
        if ($this->preserveOriginalName) {
            return $name;
        }
        return strtolower(preg_replace('/[^A-Z0-9]+/i', '_', $name));
    }
}
```

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | Fixes #2080 

#### Summary

This fixes an issue where hydrating a proxy object will not initialise any properties that aren't passed to the hydrator. With this fix, we invoke proxy initialisation with an empty initialiser so we initialise an empty object before adding any data.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
+++ b/lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
@@ -465,8 +465,19 @@ EOF
             }
         }
 
-        if ($document instanceof GhostObjectInterface) {
-            $document->setProxyInitializer(null);
+        if ($document instanceof GhostObjectInterface && $document->getProxyInitializer() !== null) {
+            // Inject an empty initialiser to not load any object data
+            $document->setProxyInitializer(static function (
+                GhostObjectInterface $ghostObject,
+                string $method, // we don't care
+                array $parameters, // we don't care
+                &$initializer,
+                array $properties // we currently do not use this
+            ) : bool {
+                $initializer = null;
+
+                return true;
+            });
         }
 
         $data = $this->getHydratorFor($metadata->name)->hydrate($document, $data, $hints);
--- a/tests/Doctrine/ODM/MongoDB/Tests/HydratorTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/HydratorTest.php
@@ -40,12 +40,35 @@ class HydratorTest extends BaseTest
         $this->assertEquals('jon', $user->name);
         $this->assertInstanceOf(DateTime::class, $user->birthdate);
         $this->assertInstanceOf(HydrationClosureReferenceOne::class, $user->referenceOne);
+        $this->assertInstanceOf(GhostObjectInterface::class, $user->referenceOne);
         $this->assertInstanceOf(PersistentCollection::class, $user->referenceMany);
         $this->assertInstanceOf(GhostObjectInterface::class, $user->referenceMany[0]);
         $this->assertInstanceOf(GhostObjectInterface::class, $user->referenceMany[1]);
+        $this->assertInstanceOf(HydrationClosureEmbedOne::class, $user->embedOne);
         $this->assertInstanceOf(PersistentCollection::class, $user->embedMany);
         $this->assertEquals('jon', $user->embedOne->name);
         $this->assertEquals('jon', $user->embedMany[0]->name);
+    }
+
+    public function testHydrateProxyWithMissingAssociations()
+    {
+        $user = $this->dm->getReference(HydrationClosureUser::class, 1);
+        $this->assertInstanceOf(GhostObjectInterface::class, $user);
+
+        $this->dm->getHydratorFactory()->hydrate($user, [
+            '_id' => 1,
+            'title' => null,
+            'name' => 'jon',
+        ]);
+
+        $this->assertEquals(1, $user->id);
+        $this->assertNull($user->title);
+        $this->assertEquals('jon', $user->name);
+        $this->assertNull($user->birthdate);
+        $this->assertNull($user->referenceOne);
+        $this->assertInstanceOf(PersistentCollection::class, $user->referenceMany);
+        $this->assertNull($user->embedOne);
+        $this->assertInstanceOf(PersistentCollection::class, $user->embedMany);
     }
 
     public function testReadOnly()


lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
```diff
@@ -466,6 +466,10 @@ public function hydrate(object $document, array $data, array $hints = []) : arra
         }
 
         if ($document instanceof GhostObjectInterface) {
+            // Inject an empty initialiser to not load any object data
+            $document->setProxyInitializer(static function () {
```
Ocramius: This should return `: bool` (`true`)

lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
```diff
@@ -466,6 +466,10 @@ public function hydrate(object $document, array $data, array $hints = []) : arra
...
+            });
+            $document->initializeProxy();
```
Ocramius: This is probably not needed: initialization will be triggered by any access (including reflection)

lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
Ocramius: This initializer should remove itself from the proxy once triggered

tests/Doctrine/ODM/MongoDB/Tests/HydratorTest.php
```diff
@@ -40,14 +40,37 @@ public function testHydrator()
...
         $this->assertInstanceOf(PersistentCollection::class, $user->referenceMany);
         $this->assertInstanceOf(GhostObjectInterface::class, $user->referenceMany[0]);
         $this->assertInstanceOf(GhostObjectInterface::class, $user->referenceMany[1]);
+        $this->assertInstanceOf(HydrationClosureEmbedOne::class, $user->embedOne);
         $this->assertInstanceOf(PersistentCollection::class, $user->embedMany);
         $this->assertEquals('jon', $user->embedOne->name);
         $this->assertEquals('jon', $user->embedMany[0]->name);
     }
 
+    public function testHydrateProxyWithMissingAssociations()
```
Ocramius: How was this failing before the patch? Undefined properties?

alcaeus: Correct.

lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
alcaeus: You are correct. I've also changed the logic to only inject an empty initialiser if the object is not in the initialisation process. Otherwise this could lead to recursive initialisation when calling `initializeProxy()` directly (or accessing any property).

Ocramius: Recursive initialization should be disabled by ProxyManager itself - it's a common scenario that I think I have covered with tests too.

lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
malarzm: Out of curiosity: what black magic is involved in including reflection into "any access"?

alcaeus: Unsetting all properties and handling access to everything in magic methods. Take a look at the generated proxies and you'll be shocked.

This also means that calling a method that doesn't access a lazy property no longer triggers proxy initialisation (which is a big plus).

malarzm: Cool, I thought that reflection access data on some lower level that couldn't be caught on PHP level 👍

Ocramius: I actually had to add test cases to php-src to prevent regressions on that 😅

Frederick888:
Thanks for the fix! And it'd be nice if these changes can land in release asap since it's blocking my upgrade process :P

malarzm:
Thanks @alcaeus!

alcaeus:
> And it'd be nice if these changes can land in release asap since it's blocking my upgrade process :P

2.0.1 has been released: https://github.com/doctrine/mongodb-odm/releases/tag/2.0.1

Frederick888:
@alcaeus Thanks for the release!
Btw do you think [the test I put together](https://github.com/Frederick888/mongodb-odm/blob/issue-2080/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH2080Test.php) is still worth merging here, or the updated `HydratorTest.php` already gets it covered (so I don't need to bother cleaning up my test case, either lol)?

alcaeus:
From what I can see the test I added covers yours, so you don't need to bother cleaning it up ;)

PR #2084, opened 2019-10-11
Cast also-load xml attribute to string

SSHVersionControl:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

#### Summary
Need to cast also-xml to string before calling explode as it is a SimpleXmlElement object otherwise a fatal error is thrown


malarzm:
@SSHVersionControl thank you very much for your contribution! Could you please either update one of our existing tests or write a new one to prevent regressions in the future?

alcaeus:
One more thing: this bug is already present in 1.3. Please target the 1.3 branch with your bugfix. Thanks!

alcaeus:
@SSHVersionControl I saw you changed the base to 1.3.x and back to master. Do you need help rebasing the changes?

PR #2085, opened 2019-10-14
Cast also-load xml attribute to string

SSHVersionControl:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

#### Summary
Need to cast also-xml to string before calling explode as it is a SimpleXmlElement object otherwise a fatal error is thrown

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/Mapping/Driver/XmlDriverTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Mapping/Driver/XmlDriverTest.php
@@ -4,10 +4,9 @@ namespace Doctrine\ODM\MongoDB\Tests\Mapping\Driver;
 
 use Doctrine\ODM\MongoDB\Mapping\ClassMetadata;
 use Doctrine\ODM\MongoDB\Mapping\Driver\XmlDriver;
+use TestDocuments\AlsoLoadDocument;
 use TestDocuments\UserCustomIdGenerator;
 use TestDocuments\UserCustomIdGeneratorWithIdField;
-
-require_once 'fixtures/AlsoLoadDocument.php';
 
 class XmlDriverTest extends AbstractDriverTest
 {
@@ -90,10 +89,10 @@ class XmlDriverTest extends AbstractDriverTest
         ], $classMetadata->getIndexes());
     }
 
-    public function testDriver()
+    public function testAlsoLoadFieldMapping()
     {
-        $classMetadata = new ClassMetadata('TestDocuments\AlsoLoadDocument');
-        $this->driver->loadMetadataForClass('TestDocuments\AlsoLoadDocument', $classMetadata);
+        $classMetadata = new ClassMetadata(AlsoLoadDocument::class);
+        $this->driver->loadMetadataForClass(AlsoLoadDocument::class, $classMetadata);
 
         $this->assertEquals(array(
             'fieldName' => 'createdAt',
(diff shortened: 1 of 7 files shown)


SSHVersionControl:
Related to #2084

tests/Doctrine/ODM/MongoDB/Tests/Mapping/Driver/XmlDriverTest.php
```diff
@@ -87,6 +89,29 @@ public function testInvalidPartialFilterExpressions()
             ],
         ], $classMetadata->getIndexes());
     }
+
+    public function testDriver()
+    {
+        $classMetadata = new ClassMetadata('TestDocuments\AlsoLoadDocument');
```
alcaeus: Please use the `::class` constants instead f string names (needs a `use` statement on top of the file):

```suggestion
        $classMetadata = new ClassMetadata(AlsoLoadDocument::class);
```

tests/Doctrine/ODM/MongoDB/Tests/Mapping/Driver/XmlDriverTest.php
```diff
@@ -87,6 +89,29 @@ public function testInvalidPartialFilterExpressions()
...
+        $this->driver->loadMetadataForClass('TestDocuments\AlsoLoadDocument', $classMetadata);
```
alcaeus:
```suggestion
        $this->driver->loadMetadataForClass(AlsoLoadDocument::class, $classMetadata);
```

tests/Doctrine/ODM/MongoDB/Tests/Mapping/Driver/XmlDriverTest.php
```diff
@@ -87,6 +89,29 @@ public function testInvalidPartialFilterExpressions()
             ],
         ], $classMetadata->getIndexes());
     }
+
+    public function testDriver()
```
alcaeus:
```suggestion
    public function testAlsoLoadFieldMapping()
```

alcaeus: This test would also profit from being in `AbstractDriverTest` so we test the same functionality for the YAML driver.

SSHVersionControl: I originally had it in the AbstractDriverTest but the YAML mapping is different from the XML for the also-load attribute. The YAML mapping is missing the data
`'also-load' => 'createdOn,creation_date',`

tests/Doctrine/ODM/MongoDB/Tests/Mapping/Driver/XmlDriverTest.php
```diff
@@ -7,6 +7,8 @@
 use TestDocuments\UserCustomIdGenerator;
 use TestDocuments\UserCustomIdGeneratorWithIdField;
 
+require_once 'fixtures/AlsoLoadDocument.php';
```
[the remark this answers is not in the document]
SSHVersionControl: I agree except that this namespace is not configured in the composer.json file. I'll add it to the composer.json if that is ok?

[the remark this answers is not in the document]
alcaeus: Yes, that will work fine. Feel free to remove any other manual `require` statements to these files.

tests/Doctrine/ODM/MongoDB/Tests/Mapping/Driver/XmlDriverTest.php
alcaeus: YAML has a different syntax which is worth testing. On the other hand, since we've dropped this in 2.0 anyways you can leave this as is.

SSHVersionControl:
Hi @alcaeus, anything other changes need ed for this pull request or is it all ok?

alcaeus:
Thanks @SSHVersionControl!

PR #2091, opened 2019-10-24
Change returned type references in query builder documentation

sarcher:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | n/a

#### Summary

Looks like something that was missed in the 2.x changes; `execute()` no longer returns a cursor but an `Iterator`, and `distinct()` returns a plain array.

I do wonder if there should be an additional note for `distinct()` calling out that the returned type is different, since that can be a bit jarring (I was migrating a project and it took me a minute to understand exactly what was different here).


sarcher:
Should be all set; let me know if anything else is needed. Thanks!

greg0ire:
> change the base of this PR to be the 2.0.x branch

I did it for you.

malarzm:
Thank you @sarcher!

PR #2093, opened 2019-10-30
Proposal: check entity type (instanceof) in subscriber and minor formatting of the article

AntoniusGolly:
Before I work on this PR I would like to ask on your opinion. I find the way the subscriber is written misleading. Currently, the subsriber will try to link the ODM Document to *every* ORM Entity. With other words the subscriber is  called post load of *every* entity defined in the project. What we want is something like

```
public function postLoad(LifecycleEventArgs $eventArgs)
    {
        $entity = $eventArgs->getEntity();

        if ($entity instanceof Order) {
            // business here
        }
}
```

Am I correct or am I missing a piece of information?

Also, in this PR I would add an example of a .yaml configuration of the listener.

|      Q       |   A
|------------- | -----------
| BC Break     | yes/no
| Fixed issues |

### DIFF
--- a/docs/en/cookbook/blending-orm-and-mongodb-odm.rst
+++ b/docs/en/cookbook/blending-orm-and-mongodb-odm.rst
@@ -120,12 +120,7 @@ or in .yaml
         tags:
             - { name: doctrine.event_listener, connection: default, event: postLoad}
 
-
-So now we need to define a class named `MyEventSubscriber` and pass a dependency to the `DocumentManager`. 
-It will have a `postLoad()` method that sets the product document reference. 
-
-
-So now we need to define a class named `MyEventSubscriber` and pass a dependency to the `DocumentManager`. It will have a `postLoad()` method that sets the product document reference:
+So now we need to define a class named `MyEventSubscriber` and `DocumentManager` as a dependency. It will have a `postLoad()` method that sets the product document reference:
 
 .. code-block:: php
 
@@ -145,8 +140,8 @@ So now we need to define a class named `MyEventSubscriber` and pass a dependency
         {
             $order = $eventArgs->getEntity();
 
-            if (!($order instanceof Order)) {
-                // fails silently if entity other then Order is loaded
+            if (!$order instanceof Order) {
+                // early return if entity other then Order is loaded
                 return;
             }
 
@@ -163,7 +158,7 @@ So now we need to define a class named `MyEventSubscriber` and pass a dependency
 The `postLoad` method will be invoked after an ORM entity is loaded from the database. This allows us 
 to use the `DocumentManager` to set the `$product` property with a reference to the `Product` document 
 with the product id we previously stored. Please note, that the event subscriber will be called on 
-postLoad of any given entity that is loaded by doctrine. Thus, it is recommended to check for the current 
+postLoad for all entities that are loaded by doctrine. Thus, it is recommended to check for the current 
 entity.  
 
 Working with Products and Orders
--- a/docs/en/cookbook/blending-orm-and-mongodb-odm.rst
+++ b/docs/en/cookbook/blending-orm-and-mongodb-odm.rst
@@ -112,6 +112,19 @@ Now we need to setup an event subscriber that will set the `$product` property o
         array(\Doctrine\ORM\Events::postLoad), new MyEventSubscriber($dm)
     );
 
+or in .yaml
+
+.. code-block:: yaml    
+    
+    App\Listeners\MyEventSubscriber:
+        tags:
+            - { name: doctrine.event_listener, connection: default, event: postLoad}
+
+
+So now we need to define a class named `MyEventSubscriber` and pass a dependency to the `DocumentManager`. 
+It will have a `postLoad()` method that sets the product document reference. 
+
+
 So now we need to define a class named `MyEventSubscriber` and pass a dependency to the `DocumentManager`. It will have a `postLoad()` method that sets the product document reference:
 
 .. code-block:: php
@@ -131,6 +144,12 @@ So now we need to define a class named `MyEventSubscriber` and pass a dependency
         public function postLoad(LifecycleEventArgs $eventArgs)
         {
             $order = $eventArgs->getEntity();
+
+            if (!($order instanceof Order)) {
+                // fails silently if entity other then Order is loaded
+                return;
+            }
+
             $em = $eventArgs->getEntityManager();
             $productReflProp = $em->getClassMetadata('Entities\Order')
                 ->reflClass->getProperty('product');
@@ -141,7 +160,11 @@ So now we need to define a class named `MyEventSubscriber` and pass a dependency
         }
     }
 
-The `postLoad` method will be invoked after an ORM entity is loaded from the database. This allows us to use the `DocumentManager` to set the `$product` property with a reference to the `Product` document with the product id we previously stored.
+The `postLoad` method will be invoked after an ORM entity is loaded from the database. This allows us 
+to use the `DocumentManager` to set the `$product` property with a reference to the `Product` document 
+with the product id we previously stored. Please note, that the event subscriber will be called on 
+postLoad of any given entity that is loaded by doctrine. Thus, it is recommended to check for the current 
+entity.  
 
 Working with Products and Orders
 --------------------------------
--- a/docs/en/cookbook/blending-orm-and-mongodb-odm.rst
+++ b/docs/en/cookbook/blending-orm-and-mongodb-odm.rst
@@ -171,7 +171,7 @@ Now create a new `Order` and link it to a `Product` in MySQL:
 Later we can retrieve the entity and lazily load the reference to the document in MongoDB:
 
 .. code-block:: php
-
+ 
     <?php
 
     $order = $em->find('Order', $order->getId());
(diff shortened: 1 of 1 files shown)


malarzm:
That sounds sensible :+1:. I'd just suggest to use early return instead:

public function postLoad(LifecycleEventArgs $eventArgs)
    {
        $entity = $eventArgs->getEntity();

        if (!$entity instanceof Order) {
            return;
        }

        // business here
}

AntoniusGolly:
Alright, I continue then. Do you like to write this as a replace of the current example or as an addition and leave the current example untouched?

Steveb-p:
> Alright, I'll continue then. Do you like to write this as a replace of the current example or as an addition and leave the current example untouched?

You can probably go either way. Most of documentation was written by people that was writing code responsible for certain functionalities, hence it's usually "better" to partially rewrite or readjust docs instead of full-on rewrite (to maintain the "vision" of original issue that was resolved) and then expanding it with additional samples / use-cases or explaining stuff that you encountered that wasn't clear. This prevents loss of information that might happen when you discard previous docs.

malarzm:
@AntoniusGolly if you're talking about that specific code example feel free to adjust original one. There's no point in having two examples of same subscriber

AntoniusGolly:
I finished the editing.

docs/en/cookbook/blending-orm-and-mongodb-odm.rst
```diff
@@ -131,6 +144,12 @@ So now we need to define a class named `MyEventSubscriber` and pass a dependency
         public function postLoad(LifecycleEventArgs $eventArgs)
         {
             $order = $eventArgs->getEntity();
+
+            if (!($order instanceof Order)) {
```
Steveb-p: Negation operator is lower in operand priorities, so wrapping braces can be dropped.

docs/en/cookbook/blending-orm-and-mongodb-odm.rst
```diff
@@ -131,6 +144,12 @@ So now we need to define a class named `MyEventSubscriber` and pass a dependency
...
+                // fails silently if entity other then Order is loaded
+                return;
```
Steveb-p: It's not really "failing". It's an early exit - we are ignoring any and all other classes.

docs/en/cookbook/blending-orm-and-mongodb-odm.rst
```diff
@@ -112,6 +112,19 @@ Now we need to setup an event subscriber that will set the `$product` property o
...
+
+.. code-block:: yaml    
+    
+    App\Listeners\MyEventSubscriber:
+        tags:
+            - { name: doctrine.event_listener, connection: default, event: postLoad}
+
+
+So now we need to define a class named `MyEventSubscriber` and pass a dependency to the `DocumentManager`. 
+It will have a `postLoad()` method that sets the product document reference.
```
greg0ire: In RST, you have to use double backticks. I see that it was like that before you worked on that file, and you can see the end result here: https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/latest/cookbook/blending-orm-and-mongodb-odm.html#blending-the-orm-and-mongodb-odm

docs/en/cookbook/blending-orm-and-mongodb-odm.rst
```diff
@@ -112,6 +112,19 @@ Now we need to setup an event subscriber that will set the `$product` property o
...
+or in .yaml
+
+.. code-block:: yaml    
+    
+    App\Listeners\MyEventSubscriber:
+        tags:
+            - { name: doctrine.event_listener, connection: default, event: postLoad}
+
+
+So now we need to define a class named `MyEventSubscriber` and pass a dependency to the `DocumentManager`.
```
Steveb-p: Didn't you mean "pass `DocumentManager` as dependency"?

docs/en/cookbook/blending-orm-and-mongodb-odm.rst
```diff
@@ -141,7 +160,11 @@ So now we need to define a class named `MyEventSubscriber` and pass a dependency
         }
     }
 
-The `postLoad` method will be invoked after an ORM entity is loaded from the database. This allows us to use the `DocumentManager` to set the `$product` property with a reference to the `Product` document with the product id we previously stored.
+The `postLoad` method will be invoked after an ORM entity is loaded from the database. This allows us 
+to use the `DocumentManager` to set the `$product` property with a reference to the `Product` document 
+with the product id we previously stored. Please note, that the event subscriber will be called on 
+postLoad of any given entity that is loaded by doctrine. Thus, it is recommended to check for the current
```
Steveb-p: `that the event subscriber will be called on 
postLoad of any given entity that is loaded by doctrine` should probably be:
`that the event subscriber will be called on 
postLoad for all entities that are loaded by Doctrine`

docs/en/cookbook/blending-orm-and-mongodb-odm.rst
AntoniusGolly: What do you mean? I don't think your comment makes sense.

docs/en/cookbook/blending-orm-and-mongodb-odm.rst
AntoniusGolly: I didn't change that line. It appears green because I added a line break.

docs/en/cookbook/blending-orm-and-mongodb-odm.rst
greg0ire: @AntoniusGolly I mean this:

```suggestion
It will have a ``postLoad()`` method that sets the product document reference. 
```

docs/en/cookbook/blending-orm-and-mongodb-odm.rst
Steveb-p: Oh, sorry then. However, since it is clearly invalid can we adjust it here? It really is the other way around - `DocumentManager` is passed as dependency to `EventSubscriber`.

docs/en/cookbook/blending-orm-and-mongodb-odm.rst
Steveb-p: It happens a lot in this file. So could you adjust it everywhere @AntoniusGolly? Pretty please? :sweat_smile:

AntoniusGolly: Only on this line? I haven't changed any of the backticks, have I?

greg0ire: Is that a no? :cry:

docs/en/cookbook/blending-orm-and-mongodb-odm.rst
AntoniusGolly: As you are right anyways I'll change that.

docs/en/cookbook/blending-orm-and-mongodb-odm.rst
AntoniusGolly: In the file I checked out there were single backticks.

Steveb-p: image

Here's a motivational cat :smiley_cat:

AntoniusGolly: Right, of course ;) Thought you were talking about changes I introduced. Now is all clear.

docs/en/cookbook/blending-orm-and-mongodb-odm.rst
```diff
@@ -1,14 +1,14 @@
 Blending the ORM and MongoDB ODM
 ================================
 
-Since the start of the `Doctrine MongoDB Object Document Mapper`_ project people have asked how it can be integrated with the `ORM`_. This article will demonstrates how you can integrate the two transparently, maintaining a clean domain model.
+Since the start of the ``Doctrine MongoDB Object Document Mapper``_ project people have asked how it can be integrated with the ``ORM``_. This article will demonstrates how you can integrate the two transparently, maintaining a clean domain model.
```
greg0ire: I think this line should be unchanged. This is the correct syntax for links, it works great as you can see here: https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/latest/cookbook/blending-orm-and-mongodb-odm.html#blending-the-orm-and-mongodb-odm

AntoniusGolly:
bump

docs/en/cookbook/blending-orm-and-mongodb-odm.rst
```diff
@@ -131,6 +139,12 @@ So now we need to define a class named `MyEventSubscriber` and pass a dependency
         public function postLoad(LifecycleEventArgs $eventArgs)
         {
             $order = $eventArgs->getEntity();
+
+            if (!$order instanceof Order) {
+                // early return if entity other then Order is loaded
```
malarzm: I think you can remove the comment altogether, it barely says what's written in the `if` statement. Or if we're interested in explaining things we may state something like "Subscriber's `postLoad` method will be called for each document but here we're interested with only instances of `Order`. One way or another, current line has a typo (then -> than), and in ODM we're using "document" instead of "entity" :)

malarzm:
@AntoniusGolly I'm sorry this relatively small PR took that much time and amendments. Sadly, parts of our documentation are really outdated and we try to use every opportunity to make them better. This time I believe we wanted that too much. I took the liberty and applied my last remark in #2143 preserving you as the original author. Thank you very much for starting this and we're looking forward to more PRs from you!

PR #2097, opened 2019-11-06
Fix wrong handling of custom identifier names

### LINKED ISSUE #2096: Custom Identifier name isn't allowed to be custom.
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no idea
| Version     | 2.0.1

#### Summary

In XML configuration, attempting to give custom field_name to IDs doesn't work. ClassMetadata fails.
```<id field-name="uuid" strategy="UUID" type="bin_uuid" />```

#### Current behavior

```<id field-name="uuid" strategy="UUID" type="bin_uuid" />```

Field-names ~~both~~ in document class, ~~and in Db are~~ is "uuid". Yet, code here:
https://github.com/doctrine/mongodb-odm/blob/master/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php#L1953
resolves ```$mapping['fieldName']``` value to ```id``` instead of intended ```uuid```.

#### How to reproduce

Use this as ID field configuration in your XML document odm file:
```<id field-name="uuid" strategy="UUID" type="bin_uuid" />```

#### Expected behavior

For ClassMetadata to have ```$mapping['fieldName']``` to actually be equal to ```$mapping['field_name']``` (the latter having the correct value), being equal to ```uuid``` not ```id```, as this example expects.

_UPDATE: We can't have custom identifier name in Db, it is always ```_id```. This was realized to be the case and the fact established in our chat below._

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | Fixes #2096 

#### Summary

This fixes an error where the `field-name` attribute in `id` XML mappings was not correctly mapped.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
@@ -159,6 +159,10 @@ class XmlDriver extends FileDriver
                 $mapping[$key] = (string) $value;
             }
 
+            if (isset($attributes['field-name'])) {
+                $mapping['fieldName'] = (string) $attributes['field-name'];
+            }
+
             if (isset($mapping['strategy'])) {
                 $mapping['options'] = [];
                 if (isset($field->{'generator-option'})) {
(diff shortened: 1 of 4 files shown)


shehi:
Any idea when we will have a release including this fix?

alcaeus:
I'll tag a release later today.

PR #2100, opened 2019-11-07
Remove reference to generators on iterator destruction

### LINKED ISSUE #2092: Memory leaking when used query builder in the loop
### Bug Report

I try to use ODM in daemon, but have problem with memory consumption.

|    Q        |   A
|------------ | ------
| Version     | 2.0.1

I created minimal example that leads to an error:

```
        ini_set('memory_limit', '16m');
        gc_enable();
        $i = 0;
        while (true) {
            /** @var File $file */
            $file = $this->dm
                ->createQueryBuilder(File::class)
                ->getQuery()
                ->getSingleResult();

            $this->dm->flush();
            $this->dm->clear();
            unset($file);
            gc_collect_cycles();
            echo $i++, PHP_EOL;
        }
``` 

The code crushes regardless of the size allocated memory after some iterations (the more memory, the more iterations)

```
PHP Fatal error:  Allowed memory size of 16777216 bytes exhausted (tried to allocate 20480 bytes) in /usr/local/app/vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/Iterator/HydratingIterator.php on line 91
PHP Stack trace:
PHP   1. {main}() /usr/local/app/yii:0
PHP   2. yii\console\Application->run() /usr/local/app/yii:32
PHP   3. yii\console\Application->handleRequest($request = *uninitialized*) /usr/local/app/vendor/yiisoft/yii2/base/Application.php:386
PHP   4. yii\console\Application->runAction($route = *uninitialized*, $params = *uninitialized*) /usr/local/app/vendor/yiisoft/yii2/console/Application.php:147
PHP   5. yii\console\Application->runAction($route = *uninitialized*, $params = *uninitialized*) /usr/local/app/vendor/yiisoft/yii2/console/Application.php:180
PHP   6. app\commands\TestController->runAction($id = *uninitialized*, $params = *uninitialized*) /usr/local/app/vendor/yiisoft/yii2/base/Module.php:528
PHP   7. app\commands\TestController->runAction($id = *uninitialized*, $params = *uninitialized*) /usr/local/app/vendor/yiisoft/yii2/console/Controller.php:164
PHP   8. yii\base\InlineAction->runWithParams($params = *uninitialized*) /usr/local/app/vendor/yiisoft/yii2/base/Controller.php:157
PHP   9. call_user_func_array:{/usr/local/app/vendor/yiisoft/yii2/base/InlineAction.php:57}(*uninitialized*, *uninitialized*) /usr/local/app/vendor/yiisoft/yii2/base/InlineAction.php:57
PHP  10. app\commands\TestController->actionIndex() /usr/local/app/vendor/yiisoft/yii2/base/InlineAction.php:57
PHP  11. Doctrine\ODM\MongoDB\Query\Query->getSingleResult() /usr/local/app/commands/TestController.php:39
PHP  12. Doctrine\ODM\MongoDB\Query\Query->getIterator() /usr/local/app/vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/Query/Query.php:277
PHP  13. Doctrine\ODM\MongoDB\Query\Query->execute() /usr/local/app/vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/Query/Query.php:249
PHP  14. Doctrine\ODM\MongoDB\Query\Query->runQuery() /usr/local/app/vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/Query/Query.php:181
PHP  15. Doctrine\ODM\MongoDB\Query\Query->makeIterator($cursor = *uninitialized*) /usr/local/app/vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/Query/Query.php:418
PHP  16. Doctrine\ODM\MongoDB\Iterator\CachingIterator->__construct($iterator = *uninitialized*) /usr/local/app/vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/Query/Query.php:356
PHP  17. Doctrine\ODM\MongoDB\Iterator\CachingIterator->storeCurrentItem() /usr/local/app/vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/Iterator/CachingIterator.php:47
PHP  18. Generator->key() /usr/local/app/vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/Iterator/CachingIterator.php:128
PHP  19. Doctrine\ODM\MongoDB\Iterator\CachingIterator->wrapTraversable($traversable = *uninitialized*) /usr/local/app/vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/Iterator/CachingIterator.php:128
PHP  20. Doctrine\ODM\MongoDB\Iterator\HydratingIterator->rewind() /usr/local/app/vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/Iterator/CachingIterator.php:139
PHP  21. Generator->rewind() /usr/local/app/vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/Iterator/HydratingIterator.php:73
PHP  22. Doctrine\ODM\MongoDB\Iterator\HydratingIterator->wrapTraversable($traversable = *uninitialized*) /usr/local/app/vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/Iterator/HydratingIterator.php:73
2019-10-29 07:44:48 [-][-][-][error][yii\base\ErrorException:1] yii\base\ErrorException: Allowed memory size of 16777216 bytes exhausted (tried to allocate 20480 bytes) in /usr/local/app/vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/Iterator/HydratingIterator.php:91
```

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | Fixes #2092 

#### Summary

The `CachingIterator` and `HydratingIterator` classes all wrap a given traversable in a generator. However, even when these iterators are garbage collected, the generator itself sometimes stays around and thus causes a memory leak. Thus, we remove our reference to the generator when destructing the iterators which allows freeing the underlying MongoDB cursor object.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Iterator/HydratingIterator.php
+++ b/lib/Doctrine/ODM/MongoDB/Iterator/HydratingIterator.php
@@ -8,6 +8,7 @@ use Doctrine\ODM\MongoDB\Mapping\ClassMetadata;
 use Doctrine\ODM\MongoDB\UnitOfWork;
 use Generator;
 use Iterator;
+use RuntimeException;
 use Traversable;
 
 /**
@@ -17,7 +18,7 @@ use Traversable;
  */
 final class HydratingIterator implements Iterator
 {
-    /** @var Generator */
+    /** @var Generator|null */
     private $iterator;
 
     /** @var UnitOfWork */
@@ -37,6 +38,11 @@ final class HydratingIterator implements Iterator
         $this->unitOfWorkHints = $unitOfWorkHints;
     }
 
+    public function __destruct()
+    {
+        $this->iterator = null;
+    }
+
     /**
      * @see http://php.net/iterator.current
      *
@@ -44,7 +50,7 @@ final class HydratingIterator implements Iterator
      */
     public function current()
     {
-        return $this->hydrate($this->iterator->current());
+        return $this->hydrate($this->getIterator()->current());
     }
 
     /**
@@ -54,7 +60,7 @@ final class HydratingIterator implements Iterator
      */
     public function key()
     {
-        return $this->iterator->key();
+        return $this->getIterator()->key();
     }
 
     /**
@@ -62,7 +68,7 @@ final class HydratingIterator implements Iterator
      */
     public function next() : void
     {
-        $this->iterator->next();
+        $this->getIterator()->next();
     }
 
     /**
@@ -70,7 +76,7 @@ final class HydratingIterator implements Iterator
      */
     public function rewind() : void
     {
-        $this->iterator->rewind();
+        $this->getIterator()->rewind();
     }
 
     /**
@@ -79,6 +85,15 @@ final class HydratingIterator implements Iterator
     public function valid() : bool
     {
         return $this->key() !== null;
+    }
+
+    private function getIterator() : Generator
+    {
+        if ($this->iterator === null) {
+            throw new RuntimeException('Iterator has already been destroyed');
+        }
+
+        return $this->iterator;
     }
 
     private function hydrate($document)
--- a/lib/Doctrine/ODM/MongoDB/Iterator/CachingIterator.php
+++ b/lib/Doctrine/ODM/MongoDB/Iterator/CachingIterator.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Doctrine\ODM\MongoDB\Iterator;
 
 use Generator;
+use RuntimeException;
 use Traversable;
 use function current;
 use function key;
@@ -25,7 +26,7 @@ final class CachingIterator implements Iterator
     /** @var array */
     private $items = [];
 
-    /** @var Generator */
+    /** @var Generator|null */
     private $iterator;
 
     /** @var bool */
@@ -45,6 +46,11 @@ final class CachingIterator implements Iterator
     {
         $this->iterator = $this->wrapTraversable($iterator);
         $this->storeCurrentItem();
+    }
+
+    public function __destruct()
+    {
+        $this->iterator = null;
     }
 
     public function toArray() : array
@@ -80,7 +86,7 @@ final class CachingIterator implements Iterator
     public function next() : void
     {
         if (! $this->iteratorExhausted) {
-            $this->iterator->next();
+            $this->getIterator()->next();
             $this->storeCurrentItem();
         }
 
@@ -118,6 +124,17 @@ final class CachingIterator implements Iterator
         while (! $this->iteratorExhausted) {
             $this->next();
         }
+
+        $this->iterator = null;
+    }
+
+    private function getIterator() : Generator
+    {
+        if ($this->iterator === null) {
+            throw new RuntimeException('Iterator has already been destroyed');
+        }
+
+        return $this->iterator;
     }
 
     /**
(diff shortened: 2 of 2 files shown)


alcaeus:
> Any chance to have this tested? Few iterations with checking memory consumption? Or maybe phpbench reports memory usage?

I was able to test this by setting a low memory limit and creating iterators in a loop, then allowing it a certain number of iterations to run out of memory. However, a lot of this varies from platform to platform. I also ran some tests on different PHP versions (see https://3v4l.org/UIfIb), and you can see that GC behaviour varies wildly. I don't think there's any way to test this automatically.

lib/Doctrine/ODM/MongoDB/Iterator/HydratingIterator.php
```diff
@@ -70,6 +88,10 @@ public function next() : void
      */
     public function rewind() : void
     {
+        if ($this->iterator === null) {
```
TheHett: This check can be taken out in a separate method.

TheHett:
```
public function checkIterator()
{
        if ($this->iterator === null) {
            throw new BadMethodCallException('Iterator has already been destroyed');
        }
}
```

TheHett: or better incapsulate text into exception

```
throw BadMethodCallException::iterratorHasAlreadyBeenDestoyed();
```

alcaeus: I have to check if that works, since this is mainly there to silence PHPStan. Moving this to a separate method probably will introduce the PHPStan failures again and given that this is a very theoretical error (it can only appear if a user manually calls `__destruct` on the object) I'm ok with not creating a separate exception class to encapsulate the exception message.

TheHett: Ops, you are used SPL exception, not ODM. Why?
Besides phpdoc of `BadMethodCallException` says:

>  Exception thrown if a callback refers to an undefined method or if some arguments are missing.

alcaeus: Switched to using `RuntimeException`.

For the sake of completeness, this is the result of extracting to a private method:
```
------ ------------------------------------------------- 
  Line   HydratingIterator.php                            
 ------ ------------------------------------------------- 
  55     Cannot call method current() on Generator|null.  
  67     Cannot call method key() on Generator|null.      
  77     Cannot call method next() on Generator|null.     
  87     Cannot call method rewind() on Generator|null.   
 ------ ------------------------------------------------- 
```

I've left the checks in to satisfy the strict checks.

alcaeus:
@jmikola @malarzm I've decided to keep the generators here for now. From what I can see, `IteratorIterator` works similarly, but there are some minor differences with respect to rewinding, especially in `CachingIterator`. I'd like to make those changes in a minor release and have scheduled #2105 for 2.1.

lib/Doctrine/ODM/MongoDB/Iterator/HydratingIterator.php
TheHett: Many make it so that internal method ensure iterator:
```
private function ensureIterator(): Iterator 
{
    if ($this->iterator === null) {
        throw new RuntimeException('Iterator has already been destroyed');
    }
    return $this->iterator;
}
```
PhpStan will protect against NPE because it not allow to use property directly.

alcaeus: Good point, I've created a `getIterator` method that does this. Thanks for suggesting this!

Issue #2103, opened 2019-11-10
No ability to query by custom mapping types


petr-buchin:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.0.2

#### Summary

Currently ODM does not support querying by any PHP object values, except `DateTimeInterface` and `ObjectId`, even if there are custom mapping types to convert this values to their DB representation.
This makes using of ODM pretty weird, when you can use your value objects in your documents (entities), but you must query this documents by database values.

After a little research I found that problem is in the `Type::getTypeFromPHPVariable()` method, which looks unfinished:

```
if (is_object($variable)) {
      if ($variable instanceof DateTimeInterface) {
          return self::getType('date');
      }

      if ($variable instanceof ObjectId) {
          return self::getType('id');
      }
} else {
//...
}
```

Here ODM explicitly ignores any objects except `DateTimeInterface` and `ObjectId`.

I believe that maybe this behavior is a result of ODM not having an explicit way to map PHP value to custom mapping type. However, there is a mapping from field name to field type, and this info may be used to properly convert value.

#### Current behavior

No ability to search by PHP object values, used in custom mapping types.

#### How to reproduce

1. Create custom mapping types, which converts php object to some scalar value for storing in DB.
2. Create document, which uses this custom mapping type.
3. Try to query by a "php value" used in custom type.
4. Query will not find anything ever,

#### Expected behavior

I expect to be able to query by PHP objects, used as a type of a field in my document.

alcaeus:
That is unfortunately known. I’ll put this on the roadmap, but can’t say when we’ll get to it.

petr-buchin:
@alcaeus from what I see, this is pretty straightforward to add this feature. Actually, you can see my PR, which adds this feature at least for non-array values.

If you have any comments on that PR, I would be happy to fix it.

alcaeus:
Fixed in #2104.

PR #2104, opened 2019-11-10
Added conversion from PHP values to DB values during query preparation

petr-buchin:
|      Q       |   A
|------------- | -----------
| BC Break     | no
| Fixed issues | #2103 

#### Summary

This PR adds conversion of PHP values in queries to their DB representations using types functionality, and thus allows to query by php values rather that db values.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -1030,10 +1030,10 @@ final class DocumentPersister
                 $preparedValue = Type::convertPHPToDatabaseValue($preparedValue);
 
                 if ($this->class->hasField($key)) {
-                    $mapping = $this->class->fieldMappings[$key];
+                    $mapping  = $this->class->fieldMappings[$key];
                     $typeName = $mapping['type'];
-                    if (Type::hasType($typeName) && !in_array($typeName, ['collection', 'hash'])) {
-                        $type = Type::getType($mapping['type']);
+                    if (Type::hasType($typeName) && ! in_array($typeName, ['collection', 'hash'])) {
+                        $type          = Type::getType($mapping['type']);
                         $preparedValue = $type->convertToDatabaseValue($preparedValue);
                     }
                 }
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -1023,10 +1023,12 @@ final class DocumentPersister
             foreach ($preparedQueryElements as [$preparedKey, $preparedValue]) {
                 if (is_array($preparedValue)) {
                     $preparedValue = array_map('\Doctrine\ODM\MongoDB\Types\Type::convertPHPToDatabaseValue', $preparedValue);
-                } else {
+                } elseif(isset($this->class->fieldMappings[$key])) {
                     $mapping = $this->class->fieldMappings[$key];
                     $type = Type::getType($mapping['type']);
                     $preparedValue = $type->convertToDatabaseValue($preparedValue);
+                } else {
+                    $preparedValue = Type::convertPHPToDatabaseValue($preparedValue);
                 }
 
                 $preparedQuery[$preparedKey] = $preparedValue;
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -1029,7 +1029,7 @@ final class DocumentPersister
 
                 $preparedValue = Type::convertPHPToDatabaseValue($preparedValue);
 
-                if ($this->class->hasField($key) && !$this->class->isIdentifier($key)) {
+                if ($this->class->hasField($key)) {
                     $mapping = $this->class->fieldMappings[$key];
                     $typeName = $mapping['type'];
                     if (Type::hasType($typeName) && !in_array($typeName, ['collection', 'hash'])) {
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -1025,13 +1025,13 @@ final class DocumentPersister
                     $preparedValue = array_map('\Doctrine\ODM\MongoDB\Types\Type::convertPHPToDatabaseValue', $preparedValue);
                 } elseif ($this->class->hasField($key) && ! $this->class->isIdentifier($key)) {
                     $mapping = $this->class->fieldMappings[$key];
-                    if (empty($mapping['reference'])) {
+                    $typeName = $mapping['type'];
+                    if (!in_array($typeName, ['collection', 'hash']) && Type::hasType($typeName)) {
                         $type = Type::getType($mapping['type']);
                         $preparedValue = $type->convertToDatabaseValue($preparedValue);
                     } else {
                         $preparedValue = Type::convertPHPToDatabaseValue($preparedValue);
                     }
-
                 } else {
                     $preparedValue = Type::convertPHPToDatabaseValue($preparedValue);
                 }
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -1022,18 +1022,20 @@ final class DocumentPersister
             $preparedQueryElements = $this->prepareQueryElement((string) $key, $value, null, true, $isNewObj);
             foreach ($preparedQueryElements as [$preparedKey, $preparedValue]) {
                 if (is_array($preparedValue)) {
-                    $preparedValue = array_map('\Doctrine\ODM\MongoDB\Types\Type::convertPHPToDatabaseValue', $preparedValue);
-                } elseif ($this->class->hasField($key) && ! $this->class->isIdentifier($key)) {
+                    $preparedQuery[$preparedKey] = array_map('\Doctrine\ODM\MongoDB\Types\Type::convertPHPToDatabaseValue', $preparedValue);
+
+                    continue;
+                }
+
+                $preparedValue = Type::convertPHPToDatabaseValue($preparedValue);
+
+                if ($this->class->hasField($key) && !$this->class->isIdentifier($key)) {
                     $mapping = $this->class->fieldMappings[$key];
                     $typeName = $mapping['type'];
-                    if (!in_array($typeName, ['collection', 'hash']) && Type::hasType($typeName)) {
+                    if (Type::hasType($typeName) && !in_array($typeName, ['collection', 'hash'])) {
                         $type = Type::getType($mapping['type']);
                         $preparedValue = $type->convertToDatabaseValue($preparedValue);
-                    } else {
-                        $preparedValue = Type::convertPHPToDatabaseValue($preparedValue);
                     }
-                } else {
-                    $preparedValue = Type::convertPHPToDatabaseValue($preparedValue);
                 }
 
                 $preparedQuery[$preparedKey] = $preparedValue;
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -1021,28 +1021,57 @@ final class DocumentPersister
 
             $preparedQueryElements = $this->prepareQueryElement((string) $key, $value, null, true, $isNewObj);
             foreach ($preparedQueryElements as [$preparedKey, $preparedValue]) {
-                if (is_array($preparedValue)) {
-                    $preparedQuery[$preparedKey] = array_map('\Doctrine\ODM\MongoDB\Types\Type::convertPHPToDatabaseValue', $preparedValue);
-
-                    continue;
-                }
-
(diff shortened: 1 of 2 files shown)


alcaeus:
At a first glance this looks good, thanks! Can you please add some tests for this feature? I'd love to have this in 2.1!

petr-buchin:
@alcaeus thanks for the review! Ok, I will add tests this week!

alcaeus:
Ignore that for now please - I just haven't found the time to introduce that test to the tip of my boot. It just needs a good kicking here and there. I'll try to fix it later today.

petr-buchin:
@alcaeus Thanks to the simple test I added for this feature, I fixed a bug.

If you want me to cover some other cases with tests, just tell me what cases I should consider.

Thanks!

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -1021,9 +1021,24 @@ public function prepareQueryOrNewObj(array $query, bool $isNewObj = false) : arr
...
+
+                    continue;
+                }
+
+                $preparedValue = Type::convertPHPToDatabaseValue($preparedValue);
+
+                if ($this->class->hasField($key)) {
+                    $mapping  = $this->class->fieldMappings[$key];
+                    $typeName = $mapping['type'];
+                    if (Type::hasType($typeName) && ! in_array($typeName, ['collection', 'hash'])) {
```
alcaeus: I would suggest throwing an exception if we can't find the type: this should've been caught during the metadata mapping phase and not be silently ignored here.

tests/Doctrine/ODM/MongoDB/Tests/Functional/DocumentPersisterTest.php
```diff
@@ -176,6 +183,46 @@ public function testPrepareQueryOrNewObjWithHashIdAndInOperators($hashId)
         $this->assertEquals($expected, $documentPersister->prepareQueryOrNewObj($value));
     }
 
+    /**
+     * @dataProvider provideCustomIds
```
alcaeus: Why does this need a data provider with three identifiers?

tests/Doctrine/ODM/MongoDB/Tests/Functional/DocumentPersisterTest.php
```diff
@@ -176,6 +183,46 @@ public function testPrepareQueryOrNewObjWithHashIdAndInOperators($hashId)
...
+     */
+    public function testPrepareQueryOrNewObjWithCustomTypedId(string $objectIdString)
+    {
+        $class             = DocumentPersisterTestDocumentWithCustomId::class;
+        $documentPersister = $this->uow->getDocumentPersister($class);
+
+        $customId = DocumentPersisterCustomTypedId::fromString($objectIdString);
+
+        $value    = ['id' => $customId];
```
alcaeus: Please also add tests for the following queries:
* `['id' => ['$ne' => $customId]]`
* `['id' => ['$in' => [$customId1, $customId2]]]`

I suggest using a data provider for this:
```php
public static function queryProvider() : Generator
{
    yield 'Direct comparison' => ['query' => ['id' => $customId], 'expected' => ['_id' => ...]];
    yield 'Operator with single value' => ...;
}
```

petr-buchin:
Hi @alcaeus, thanks for the review!
> I think we also need to apply the same transformation to arrays
Do you mean we need to convert arrays the same way as we do with a single values?

If yes, how do you want to do this?
From what I see, there is no way to specify item type for a `collection` mapping (except reference collections), thus there is no way to know the type of an item in an array.

petr-buchin:
@alcaeus I got your point: I updated tests as you requested and added exception for non-existent types.

I also found and fixed the bug: at the moment ODM user can query by providing target document for reference field like that:

```
['referenceField' => $targetDocumentValue]
```

but user cannot query by providing target document with a query operators, like that:
```
['referenceField' => ['$ne' => $targetDocumentValue]]
```

For now ODM will try to convert target document value in query using target document's id type.
I fixed this behavior by passing retrieving id from target document (see changes in `DocumentPersister::prepareQueryExpression()` method).

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
petr-buchin: Done

tests/Doctrine/ODM/MongoDB/Tests/Functional/DocumentPersisterTest.php
petr-buchin: I removed this data provider and replaced it by the data provider you've suggested below

tests/Doctrine/ODM/MongoDB/Tests/Functional/DocumentPersisterTest.php
petr-buchin: Done

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -1021,15 +1021,59 @@ public function prepareQueryOrNewObj(array $query, bool $isNewObj = false) : arr
...
 
+    /**
+     * Converts a single value to its database representation based on the mapping type
+     *
+     * @param string $fieldName
+     * @param mixed  $value
+     *
+     * @return mixed
+     */
+    private function convertToDatabaseValue($fieldName, $value)
```
alcaeus: For new code, please use type declarations where possible:

```suggestion
    private function convertToDatabaseValue(string $fieldName, $value)
```

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -1021,15 +1021,59 @@ public function prepareQueryOrNewObj(array $query, bool $isNewObj = false) : arr
...
             }
         }
 
         return $preparedQuery;
     }
 
+    /**
+     * Converts a single value to its database representation based on the mapping type
+     *
+     * @param string $fieldName
```
alcaeus: Obsolete due to the type declaration requested below.
```suggestion
```

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -1021,15 +1021,59 @@ public function prepareQueryOrNewObj(array $query, bool $isNewObj = false) : arr
...
+     * @param mixed  $value
```
alcaeus:
```suggestion
     * @param mixed $value
```

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -1244,6 +1288,11 @@ private function prepareQueryExpression(array $expression, ClassMetadata $class)
             // Process query operators whose argument arrays need type conversion
             if (in_array($k, ['$in', '$nin', '$all']) && is_array($v)) {
                 foreach ($v as $k2 => $v2) {
+                    if ($v2 instanceof $class->name) {
```
alcaeus: I assume that this is the bugfix that you mentioned in your comment:

> For now ODM will try to convert target document value in query using target document's id type.
I fixed this behavior by retrieving id from target document (see changes in `DocumentPersister::prepareQueryExpression()` method).

Can you please add a comment as to why this conditional is necessary? This code is already quite loosely commented, so any comment explaining why we're doing things will help when we will eventually have to touch this piece of code again. Thanks!

tests/Doctrine/ODM/MongoDB/Tests/Functional/DocumentPersisterTest.php
```diff
@@ -176,6 +184,70 @@ public function testPrepareQueryOrNewObjWithHashIdAndInOperators($hashId)
         $this->assertEquals($expected, $documentPersister->prepareQueryOrNewObj($value));
     }
 
+    /**
+     * @param array $testCase
+     *
+     * @dataProvider queryProviderForCustomTypeId
+     */
+    public function testPrepareQueryOrNewObjWithCustomTypedId(array $testCase)
```
alcaeus:
```suggestion
    public function testPrepareQueryOrNewObjWithCustomTypedId(array $expected, array $query)
```

tests/Doctrine/ODM/MongoDB/Tests/Functional/DocumentPersisterTest.php
```diff
@@ -176,6 +184,70 @@ public function testPrepareQueryOrNewObjWithHashIdAndInOperators($hashId)
...
+    {
+        $class             = DocumentPersisterTestDocumentWithCustomId::class;
+        $documentPersister = $this->uow->getDocumentPersister($class);
+
+        Type::registerType('DocumentPersisterCustomId', DocumentPersisterCustomIdType::class);
+
+        $this->assertEquals(
+            $testCase['expected'],
```
alcaeus:
```suggestion
            $expected,
```

tests/Doctrine/ODM/MongoDB/Tests/Functional/DocumentPersisterTest.php
```diff
@@ -176,6 +184,70 @@ public function testPrepareQueryOrNewObjWithHashIdAndInOperators($hashId)
...
+            $documentPersister->prepareQueryOrNewObj($testCase['query'])
```
alcaeus:
```suggestion
            $documentPersister->prepareQueryOrNewObj($query)
```

tests/Doctrine/ODM/MongoDB/Tests/Functional/DocumentPersisterTest.php
```diff
@@ -176,6 +184,70 @@ public function testPrepareQueryOrNewObjWithHashIdAndInOperators($hashId)
...
+        $documentWithCustomId  = $this->dm->getReference(
+            DocumentPersisterTestDocumentWithCustomId::class,
+            $customId
+        );
+        $documentWithCustomId2 = $this->dm->getReference(
+            DocumentPersisterTestDocumentWithCustomId::class,
+            $customId2
+        );
+
+        // We need to have test cases inside this method, since we cannot access `$this->dm` in data provider
```
alcaeus: We can work around this by using a closure for the `getReference` logic above. Consider this short example:
```php
public function dataProvider(): array
{
    return [[function (DocumentManager $dm) { return $dm->getReference(...); }]];
}

/** @dataProvider dataProvider */
public function testCase(Closure $getReference)
{
    $reference = $getReference($this->dm);
    // ...
}
```

tests/Doctrine/ODM/MongoDB/Tests/Functional/DocumentPersisterTest.php
```diff
@@ -223,6 +295,36 @@ public function testPrepareQueryOrNewObjWithSimpleReferenceToTargetDocumentWithN
...
+    {
+        $objectIdString  = (string) new ObjectId();
+        $objectIdString2 = (string) new ObjectId();
+
+        $customId  = DocumentPersisterCustomTypedId::fromString($objectIdString);
+        $customId2 = DocumentPersisterCustomTypedId::fromString($objectIdString2);
+
+        yield 'Direct comparison' => [
+            [
+                'query' => ['id' => $customId],
```
alcaeus: With the suggestion above, you can drop one array level:
```php
yield 'Direct comparison' => [
    'query' => ...,
    'expected' => ...,
];
```

The key names for "query" and "expected" aren't necessary since PHPUnit passes the values of the array as arguments to the test. However, they make working with the data provider easier since you don't need to think which array is which and instead can use the parameter names.

petr-buchin:
@alcaeus thanks for the suggestions!
I will fix this later today

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
petr-buchin: Done.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
petr-buchin: Done.

tests/Doctrine/ODM/MongoDB/Tests/Functional/DocumentPersisterTest.php
petr-buchin: Done.

tests/Doctrine/ODM/MongoDB/Tests/Functional/DocumentPersisterTest.php
petr-buchin: I've refactored this test using your advice, but I'm not sure that I understood you correctly.
Please look at the code and confirm if that's what you wanted.

The downside of closure is that we don't get `$expected` and `$query` as arguments of the test method as in the test above.

tests/Doctrine/ODM/MongoDB/Tests/Functional/DocumentPersisterTest.php
petr-buchin: Done, but only for the test method that does not use closure for `getReference()` logic

petr-buchin:
@alcaeus sorry about not getting back with fixes sooner.

I did all the fixes as you've requested, but I'm not sure about one single fix: please confirm if fix with `getReference()` logic, data provider and closure is what you wanted.

Thanks!

alcaeus:
Thanks @petr-buchin!

petr-buchin:
@alcaeus thanks for review and for help with this!

PR #2108, opened 2019-11-12
Fix merging of documents when not cascading references

### LINKED ISSUE #2106: Ununitialized data in relation after merge entity
### Bug Report

A certain set of actions can lead to null fields in related entity.

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.0.1

#### Summary

This occurs when working with more than one DocumentManager`s.

#### How to reproduce

```
        $dm = $this->dmFactory->createDocumentManager();

        $entity = $dm->find(Upload::class, '5dca8412a3610c000774748d');

        // initialize relation by accessing at the field, it`s important
        $entity->getProject()->getName();

        $dm2 = $this->dmFactory->createDocumentManager();

        /** @var Upload $entity2 */
        $entity2 = $dm2->merge($entity);

        // error here: `getName` gives null
        $entity2->getProject()->getName();
```

Where the `Upload` entity has `project` relation:
```
    /**
     * @var Project
     * @ODM\ReferenceOne(name="project_id", targetDocument=Project::class, storeAs="id")
     */
    private $project;
```

#### Expected behavior

Related data in `entity2` must be correctly initialized

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | Fixes #2106 

#### Summary

The described issue only appears when not cascading merge operations. The problem is that the call to `registerManaged` after creating a new proxy object will return early since a document with that ID is already registered (the original proxy loaded in `$managedCopy`. This leads to the new value for the association being different from the document that is being managed by the entity manager, which again causes issues on subsequent proxy initialisations.

On a related note, I'm not sure why the merge logic includes creating a new proxy object for the managed copy and discarding the changes from the original. However, this is logic that was introduced 8 years ago to ORM, was ported to MongoDB ODM and wasn't changed in ORM since, so I don't feel comfortable changing this particular thing about it.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
+++ b/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
@@ -1938,6 +1938,11 @@ class UnitOfWork implements PropertyChangedListener
                                 $targetClass = $this->dm->getClassMetadata($targetDocument);
                                 $relatedId = $targetClass->getIdentifierObject($other);
 
+                                $current = $prop->getValue($managedCopy);
+                                if ($current !== null) {
+                                    $this->removeFromIdentityMap($current);
+                                }
+
                                 if ($targetClass->subClasses) {
                                     $other = $this->dm->find($targetClass->name, $relatedId);
                                 } else {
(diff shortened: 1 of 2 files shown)


malarzm:
> On a related note, I'm not sure why the merge logic includes creating a new proxy object for the managed copy and discarding the changes from the original.

I believe the merge logic is "take everything as I present you and consider it your state as it was read from db" hence, I believe, discarding of original document state as the merge is cascaded.

malarzm:
Please disregard my previous comment, I should have first read the code and where the discard/create is located :)

PR #2111, opened 2019-11-13
remove repeated sidebar section

hamidrezabstn:
Type ->bug

#### Summary

remove basic mapping section which it was repeated


alcaeus:
Thanks @hamidrezabstn!

PR #2114, opened 2019-11-18
Ignore index name changes

notrix:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | -

#### Summary

As mongoDB generates index name automatically and document index has no forced index name, `doctrine/mongodb-odm` sees changes and always recreates indexes. I think that index name is not important change and could be ignored.


alcaeus:
@notrix as this is a bugfix for 2.0, can you please re-target 2.0 with this branch? You might have to drop commits from 2.1 from this branch, please let me know if you need help with that.

alcaeus:
Thanks @notrix!

Issue #2113, opened 2019-11-14
Query/Aggregate Builders no longer allow incremental results / always cache


Steveb-p:
|    Q        |   A
|------------ | -----
| Version     | 2.0.2

### Support Question

In ODM 1.3 I was able to build and run queries with large result sets. This allowed me to process incoming data incrementally, by yielding data in generator.

In 2.0 with new Builders I've noticed that both builders wrap `Cursor` with `CachingIterator`, which causes high memory usage in such cases. I've hit my memory limit (obviously didn't want to raise it, I had it relatively high in the first place) and had to work around it - I'm getting pipeline built and passed directly into `aggregate` method of `Collection`.

```php
$builder = $this->dm->createAggregationBuilder($collectionName);

$collection = $this->dm->getClient()->selectCollection('xxx', $collectionName);
yield from $collection->aggregate($builder->getPipeline(), [
    'allowDiskUse' => true,
    'cursor' => true,
]);
``` 

I'd like to be able to select which class caches the results, maybe discarding the ability to rewind/count when I know I won't be using them?

alcaeus:
Ah, I see. We originally made the decision to wrap the MongoDB cursor in a `CachingIterator` instance to allow rewinding over the result set. I think adding an explicit way to disable the `CachingIterator` functionality should be possible for query and aggregation builders:

```php
$builder
    ->cache(false) // true is the default
    ->getQuery()
    ->getIterator(); // Would no longer return a `CachingIterator`
```

For the aggregation builder I'd have to think about where to put this option. I'm a bit vary of putting it on the root level (like we do in query builder) since we only have aggregation pipeline stages there. On the other hand, aggregation options in `execute` might be wrong as well, as those options are all passed to the underlying `aggregate` method.

If you want to give implementing this a shot, please feel free to reach out on Slack or here to discuss specifics.

malarzm:
FWIW I'd name the method `rewindable` or something alike. Cache might imply we finally implemented 2nd level cache for queries :)

alcaeus:
Good point. Let's not call it `cache` ;)

Steveb-p:
> If you want to give implementing this a shot, please feel free to reach out on Slack or here to discuss specifics.

Gladly, however I've still got another issue I've reported here that I promised to take care of, so I don't want to promise more :) I'll try though. I'll say so here if I start doing anything about it :smile:

Steveb-p:
[...]

PR #2116, opened 2019-11-20
Provide rewindable option to Aggregation and Query builders/queries

Steveb-p:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #2113 

#### Summary

Proof of concept for allowing user control over wrapping resulting `Traversable`/`Iterator` with `CachingIterator` - to prevent memory issues when iterating over large result sets.

Note: This PR is incomplete - still needs ~tests, relevant comment blocks and~ documentation adjustments.

- [x] Add tests
- [x] Add comments to code
- [x] Adjust documentation

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
@@ -8,6 +8,7 @@ use Doctrine\ODM\MongoDB\DocumentManager;
 use Doctrine\ODM\MongoDB\Iterator\CachingIterator;
 use Doctrine\ODM\MongoDB\Iterator\HydratingIterator;
 use Doctrine\ODM\MongoDB\Iterator\Iterator;
+use Doctrine\ODM\MongoDB\Iterator\UnrewindableIterator;
 use Doctrine\ODM\MongoDB\Mapping\ClassMetadata;
 use Doctrine\ODM\MongoDB\Persisters\DocumentPersister;
 use Doctrine\ODM\MongoDB\Query\Expr as QueryExpr;
@@ -53,6 +54,9 @@ class Builder
 
     /** @var Stage[] */
     private $stages = [];
+
+    /** @var bool */
+    private $rewindable = true;
 
     /**
      * Create a new aggregation builder.
@@ -431,6 +435,16 @@ class Builder
     }
 
     /**
+     * Controls if resulting iterator should be wrapped with CachingIterator.
+     */
+    public function rewindable(bool $rewindable = true) : self
+    {
+        $this->rewindable = $rewindable;
+
+        return $this;
+    }
+
+    /**
      * Randomly selects the specified number of documents from its input.
      *
      * @see https://docs.mongodb.org/manual/reference/operator/aggregation/sample/
@@ -550,6 +564,8 @@ class Builder
             $cursor = new HydratingIterator($cursor, $this->dm->getUnitOfWork(), $class);
         }
 
-        return new CachingIterator($cursor);
+        $cursor = $this->rewindable ? new CachingIterator($cursor) : new UnrewindableIterator($cursor);
+
+        return $cursor;
     }
 }
--- a/lib/Doctrine/ODM/MongoDB/Query/Query.php
+++ b/lib/Doctrine/ODM/MongoDB/Query/Query.php
@@ -10,6 +10,7 @@ use Doctrine\ODM\MongoDB\Iterator\CachingIterator;
 use Doctrine\ODM\MongoDB\Iterator\HydratingIterator;
 use Doctrine\ODM\MongoDB\Iterator\Iterator;
 use Doctrine\ODM\MongoDB\Iterator\PrimingIterator;
+use Doctrine\ODM\MongoDB\Iterator\UnrewindableIterator;
 use Doctrine\ODM\MongoDB\Mapping\ClassMetadata;
 use Doctrine\ODM\MongoDB\MongoDBException;
 use InvalidArgumentException;
@@ -84,6 +85,9 @@ final class Query implements IteratorAggregate
      */
     private $primers = [];
 
+    /** @var bool */
+    private $rewindable = true;
+
     /**
      * Hints for UnitOfWork behavior.
      *
@@ -115,7 +119,7 @@ final class Query implements IteratorAggregate
      */
     private $options;
 
-    public function __construct(DocumentManager $dm, ClassMetadata $class, Collection $collection, array $query = [], array $options = [], bool $hydrate = true, bool $refresh = false, array $primers = [], bool $readOnly = false)
+    public function __construct(DocumentManager $dm, ClassMetadata $class, Collection $collection, array $query = [], array $options = [], bool $hydrate = true, bool $refresh = false, array $primers = [], bool $readOnly = false, bool $rewindable = true)
     {
         $primers = array_filter($primers);
 
@@ -144,6 +148,7 @@ final class Query implements IteratorAggregate
 
         $this->setReadOnly($readOnly);
         $this->setRefresh($refresh);
+        $this->setRewindable($rewindable);
 
         if (! isset($query['readPreference'])) {
             return;
@@ -316,6 +321,14 @@ final class Query implements IteratorAggregate
     }
 
     /**
+     * Set to enable wrapping of resulting Iterator with CachingIterator
+     */
+    public function setRewindable(bool $rewindable = true) : void
+    {
+        $this->rewindable = $rewindable;
+    }
+
+    /**
      * Execute the query and return its results as an array.
      *
      * @see IteratorAggregate::toArray()
@@ -344,8 +357,8 @@ final class Query implements IteratorAggregate
      *
      * Note: while this method could strictly take a MongoDB\Driver\Cursor, we
      * accept Traversable for testing purposes since Cursor cannot be mocked.
-     * HydratingIterator and CachingIterator both expect a Traversable so this
-     * should not have any adverse effects.
+     * HydratingIterator, CachingIterator, and BaseIterator expect a Traversable
+     * so this should not have any adverse effects.
      */
     private function makeIterator(Traversable $cursor) : Iterator
     {
@@ -353,7 +366,7 @@ final class Query implements IteratorAggregate
             $cursor = new HydratingIterator($cursor, $this->dm->getUnitOfWork(), $this->class, $this->unitOfWorkHints);
         }
 
-        $cursor = new CachingIterator($cursor);
+        $cursor = $this->rewindable ? new CachingIterator($cursor) : new UnrewindableIterator($cursor);
 
         if (! empty($this->primers)) {
             $referencePrimer = new ReferencePrimer($this->dm, $this->dm->getUnitOfWork());
--- a/docs/en/reference/aggregation-builder.rst
+++ b/docs/en/reference/aggregation-builder.rst
@@ -191,6 +191,26 @@ specified document.
     can map embedded documents, define references, and even use discriminators
     to get different result documents according to the aggregation result.
 
+Disabling Result Caching
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+Due to MongoDB cursors not being rewindable, ODM uses a caching iterator when
+returning results from aggregation pipelines. This cache allows you to iterate a
+result cursor multiple times without re-executing the original aggregation
+pipeline. However, in long-running processes or when handling a large number of
+results, this can lead to high memory usage. To disable this result cache, you
+can tell the query builder to not return a caching iterator:
+
+.. code-block:: php
+
+    <?php
+
+    $builder = $dm->createAggregationBuilder(\Documents\Orders::class);
+    $builder->setRewindable(false);
+
+When setting this option to ``false``, attempting a second iteration will result
+in an exception.
+
 Aggregation pipeline stages
 ---------------------------
 
--- a/docs/en/reference/query-builder-api.rst
+++ b/docs/en/reference/query-builder-api.rst
@@ -287,6 +287,28 @@ get the raw results directly back from mongo by using the
 
     print_r($users);
 
+Disabling Result Caching
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+Due to MongoDB cursors not being rewindable, ODM uses a caching iterator when
+returning results from queries. This cache allows you to iterate a result cursor
+multiple times without re-executing the original query. However, in long-running
+processes or when handling a large number of results, this can lead to high
+memory usage. To disable this result cache, you can tell the query builder to
+not return a caching iterator:
+
+.. code-block:: php
+
+    <?php
+
+    $blogPosts = $dm->createQueryBuilder(BlogPost::class)
+        ->setRewindable(false)
+        ->getQuery()
+        ->execute();
+
+When setting this option to ``false``, attempting a second iteration will result
+in an exception.
+
 Limiting Results
 ~~~~~~~~~~~~~~~~
 
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Stage.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Stage.php
@@ -265,6 +265,16 @@ abstract class Stage
     }
 
     /**
+     * Controls if resulting iterator should be wrapped with CachingIterator.
+     */
+    public function rewindable(bool $rewindable = true) : self
+    {
+        $this->builder->rewindable($rewindable);
+
+        return $this;
+    }
+
+    /**
      * Randomly selects the specified number of documents from its input.
      *
      * @see https://docs.mongodb.org/manual/reference/operator/aggregation/sample/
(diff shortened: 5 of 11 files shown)


malarzm:
So for the heads up, @alcaeus wants to evaluate one other approach/idea (but I'll let him do the explaining) and if it doesn't make sense, we'll go with this :)

Steveb-p:
@malarzm I've talked a little with @alcaeus over on slack about it, and he didn't seem sure about his approach. 

As far as I understood it, it revolved around creating a `WindowIterator` that would cache a specific amount of results instead of the whole result set.

However, either way I believe some way of allowing control over what iterators are added to iterator chain will be probably needed, whether it will be `rewindable` method for both `Builder`s or something else is up for discussion.

I've added commits with new `BaseIterator` class, since `PrimingIterator` requires `Iterator` instance as it's constructor argument and reducing it to `iterable` or `Traversable` might not do the trick - after all, there are plenty of instances where `Iterator` is specifically expected, not just any `Traversable`.  
Depending on `rewindable` option either `BaseIterator` or `CachingIterator` is used to wrap `Cursor`.

I've also slightly changed the way `CachingIterator` works. I've removed the `getIterator` private method as it was only used in `storeCurrentResult` method - which in turn was called in constructor once and on each `next` call. Now instead the wrapping `Generator` stores results directly in `$this->items` just before `yield`ing them (`storeCurrentResult` is still called in constructor).

```php
private function wrapTraversable(Traversable $traversable) : Generator
{
    foreach ($traversable as $key => $value) {
        $this->items[$key] = $value;
        yield $key => $value;
        $this->iteratorAdvanced = true;
    }

    $this->iterator = null;
}
```
 
I've removed `iteratorExhausted` property - now `Generator` removes it's reference immediately just before final return, which also signifies iterator exhaustion (`$this->iterator === null`). This also means that `Generator` is freed earlier - immediately after exhaustion - in case PHP messes up it's reference counts.

I've wondered if those two iterators - `BaseIterator` and `CachingIterator` - should have a common parent class, but almost every method differs and I'm not sure it is worth it creating such a class for them.

There are still some code quality checks / comment updates that I need to address, but I'd like to know your opinion if this approach is a valid one.

See new commits: https://github.com/doctrine/mongodb-odm/pull/2116/files/6e26070f7ad7a89f33bc5b1042133e0886118c4d..a312708b991c7c36d8bdaa1b581a70e007dff620

lib/Doctrine/ODM/MongoDB/Iterator/BaseIterator.php
```diff
@@ -0,0 +1,142 @@
...
+/**
+ * Iterator for wrapping a Traversable and caching its results.
+ *
+ * By caching results, this iterators allows a Traversable to be counted and
+ * rewound multiple times, even if the wrapped object does not natively support
+ * those operations (e.g. MongoDB\Driver\Cursor).
+ *
+ * @internal
+ */
+final class BaseIterator implements Iterator
```
alcaeus: `BaseIterator` makes it sounds like this is designed to be extended, which it isn't due to it being `final` and `@internal`. The SPL brings a `NoRewindIterator`, which would be a good name except that it behaves differently (it ignores calls to `rewind` instead of throwing an exception. What do you think of calling this `UnrewindableIterator`?

```suggestion
final class UnrewindableIterator implements Iterator
```

lib/Doctrine/ODM/MongoDB/Iterator/BaseIterator.php
```diff
@@ -0,0 +1,142 @@
...
+     */
+    public function __construct(Traversable $iterator)
+    {
+        $this->iterator = $this->wrapTraversable($iterator);
+        $this->storeCurrentItem();
+    }
+
+    public function toArray() : array
+    {
+        if ($this->iteratorAdvanced) {
```
alcaeus: Suggest extracting this to a method to avoid duplicating logic here and in `rewind`:
```suggestion
        $this->preventRewinding(__METHOD__);
```

lib/Doctrine/ODM/MongoDB/Iterator/BaseIterator.php
```diff
@@ -0,0 +1,142 @@
...
+     * Initialize the iterator and stores the first item in the cache. This
+     * effectively rewinds the Traversable and the wrapping Generator, which
+     * will execute up to its first yield statement. Additionally, this mimics
+     * behavior of the SPL iterators and allows users to omit an explicit call
+     * to rewind() before using the other methods.
+     */
+    public function __construct(Traversable $iterator)
+    {
+        $this->iterator = $this->wrapTraversable($iterator);
+        $this->storeCurrentItem();
```
alcaeus: I feel like this call is superfluous: `wrapTraversable` stores the current key and value in `$this->key` and `$this->value` in the foreach loop, so I'm wondering why we still need this here.

lib/Doctrine/ODM/MongoDB/Iterator/BaseIterator.php
```diff
@@ -0,0 +1,142 @@
...
+    public function toArray() : array
+    {
+        if ($this->iteratorAdvanced) {
+            throw new \LogicException(sprintf(
+                'Cannot call %s for iterator that already yielded results',
+                __METHOD__
+            ));
+        }
+        
+        $wrapper = function () {
```
alcaeus:
```suggestion
        $toArray = function () {
```

lib/Doctrine/ODM/MongoDB/Iterator/BaseIterator.php
```diff
@@ -0,0 +1,142 @@
...
+            yield $this->key => $this->current;
+            yield from $this->iterator;
+        };
+        
+        return iterator_to_array($wrapper());
```
alcaeus:
```suggestion
        return iterator_to_array($toArray());
```

lib/Doctrine/ODM/MongoDB/Iterator/BaseIterator.php
```diff
@@ -0,0 +1,142 @@
...
+            return;
+        }
+
+        $this->current = $this->iterator->current();
+    }
+
+    private function wrapTraversable(Traversable $traversable) : Generator
+    {
+        foreach ($traversable as $key => $value) {
+            $this->key = $key;
```
alcaeus: Why do we store the key and value instead of accessing `$this->iterator->key()` and `$this->iterator->current()`?

lib/Doctrine/ODM/MongoDB/Iterator/CachingIterator.php
```diff
@@ -121,43 +117,35 @@ public function valid() : bool
...
     {
-        while (! $this->iteratorExhausted) {
+        while ($this->iterator) {
             $this->next();
         }
 
         $this->iterator = null;
     }
 
-    private function getIterator() : Generator
```
alcaeus: This method is necessary to prevent PHPStan from throwing an error in our face: https://travis-ci.org/doctrine/mongodb-odm/jobs/623925578#L1441.

lib/Doctrine/ODM/MongoDB/Iterator/CachingIterator.php
```diff
@@ -121,43 +117,35 @@ public function valid() : bool
...
         }
 
-        $this->items[$key] = $this->getIterator()->current();
+        $this->items[$key] = $this->iterator->current();
     }
 
     private function wrapTraversable(Traversable $traversable) : Generator
     {
         foreach ($traversable as $key => $value) {
+            $this->items[$key] = $value;
```
alcaeus: I'd rather reuse `storeCurrentItem` here:
```suggestion
            $this->storeCurrentItem();
```

That said, I'm not sure why this refactoring is necessary, could you elaborate on that please?

lib/Doctrine/ODM/MongoDB/Query/Query.php
```diff
@@ -344,16 +359,16 @@ static function ($value) {
      *
      * Note: while this method could strictly take a MongoDB\Driver\Cursor, we
      * accept Traversable for testing purposes since Cursor cannot be mocked.
-     * HydratingIterator and CachingIterator both expect a Traversable so this
-     * should not have any adverse effects.
+     * HydratingIterator, CachingIterator and BaseIterator expect a Traversable
```
alcaeus: Please use an [Oxford comma](https://en.wikipedia.org/wiki/Serial_comma) here:
```suggestion
     * HydratingIterator, CachingIterator, and BaseIterator expect a Traversable 
```

lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
```diff
@@ -430,6 +434,18 @@ public function replaceRoot($expression = null) : Stage\ReplaceRoot
         return $stage;
     }
 
+    /**
+     * Controls if resulting iterator should be wrapped with CachingIterator.
+     * 
+     * @see https://github.com/doctrine/mongodb-odm/issues/2113
+     */
+    public function rewindable(bool $rewindable = true) : self
```
alcaeus: Due to the work the aggregation builder is set up, this also requires a proxy method in `Doctrine\ODM\MongoDB\Aggregation\Stage`:
```php
public function rewindable(bool $rewindable = true) : self
{
    $this-builder->rewindable($rewindable);

    return $this;
}

File: lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php [hunk 11]
```diff
@@ -430,6 +434,18 @@ public function replaceRoot($expression = null) : Stage\ReplaceRoot
         return $stage;
     }
 
+    /**
+     * Controls if resulting iterator should be wrapped with CachingIterator.
+     * 
+     * @see https://github.com/doctrine/mongodb-odm/issues/2113
```
alcaeus: I believe we can omit this comment.

lib/Doctrine/ODM/MongoDB/Iterator/BaseIterator.php
```diff
@@ -0,0 +1,142 @@
...
+            $this->iterator->next();
+        }
+    }
+
+    /**
+     * @see http://php.net/iterator.rewind
+     */
+    public function rewind() : void
+    {
+        /* If the iterator has advanced, exhaust it now so that future iteration
```
alcaeus: This comment is obsolete.

lib/Doctrine/ODM/MongoDB/Iterator/BaseIterator.php
```diff
@@ -0,0 +1,142 @@
...
+ *
+ * @internal
+ */
+final class BaseIterator implements Iterator
+{
+    private $key;
+    
+    private $current;
+
+    /** @var Generator */
```
alcaeus:
```suggestion
    /** @var Generator|null */
```

This will fix the current PHPStan errors in this file, but most likely cause others. You can use the `getIterator()` method from `CachingIterator` to prevent these errors.

alcaeus (changes requested):
Thanks for working on this logic! I've left some feedback in the code itself. Just a couple of points here:
* I'm not sure if the refactoring in `CachingIterator` is necessary. If it isn't, I'd prefer leaving it out as long as it doesn't have a performance impact.
* The code currently produces a lot of phpcs errors: https://travis-ci.org/doctrine/mongodb-odm/jobs/623925579#L2854. If you've installed dependencies locally, you can run `vendor/bin/phpcbf` to fix most of them. Then, run `vendor/bin/phpcs` to see what errors couldn't be fixed and fix them manually.
* PHPStan also complains about your code: https://travis-ci.org/doctrine/mongodb-odm/jobs/623925578#L1426. Reverting the refactoring in `CachingIterator` will fix some of them, while applying the `getIterator()` method to `BaseIterator` and updating the `$iterator` type hint as suggested below will fix the rest of them.
* This PR currently contains commits from another branch. I'll merge up through the branches soon, which should take care of that. We can worry about rebasing and squashing commits once the implementation is solid.

lib/Doctrine/ODM/MongoDB/Iterator/BaseIterator.php
Steveb-p: It's required to cause generator to reach first `yield`. Otherwise `Cursor` is not starting it's execution, which differs from what `CachingIterator` does. When creating this I worked under the assumption that this should not be changed.

Should I change the method name to something else, and if yes, what would you prefer?

lib/Doctrine/ODM/MongoDB/Query/Query.php
Steveb-p: Applied in 87fc9210df0a412b36be475e3e9f7b925aa8201a

lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
Steveb-p: Applied in 74ddf548fa0d781ebba8adcefaf40f79609016d6

lib/Doctrine/ODM/MongoDB/Iterator/CachingIterator.php
Steveb-p: Reverted changes to `CachingIterator` in 3dd3423525672225acbebd4c4ebc72d5ff326edd

lib/Doctrine/ODM/MongoDB/Iterator/BaseIterator.php
Steveb-p: Applied alongside other comments related to `UnrewindableIterator` (previously `BaseIterator`) in 907e08d5bee727860eac12ef51dc6392a654c613.

lib/Doctrine/ODM/MongoDB/Iterator/BaseIterator.php
Steveb-p: I do not know why I had it set up this way. It's probably a result of me using `CachingIterator` as base. Adjusted.

lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
Steveb-p: Only the `@see` part, right? Adjusted :smile:

lib/Doctrine/ODM/MongoDB/Iterator/BaseIterator.php
Steveb-p: I've removed the method completely and only left `$this->iterator->key()` in `__construct` which - as far as I understand it, I may be wrong - starts up the generator causes `Cursor` to query and execute.

alcaeus: Wouldn't the initial `foreach` already initialise the cursor and start iterating?

Steveb-p:
```php
<?php

function wrapTraversable(Traversable $traversable) {
    echo("ITERATOR EXECUTING\n");
    foreach ($traversable as $key => $value) {
        yield $key => $value;
    }
};

$generator = wrapTraversable(new ArrayIterator([1, 2, 3]));

echo "BEFORE KEY\n";
$generator->key();
echo "AFTER KEY\n";

foreach ($generator as $key => $value) {
    echo $value;
}
```

Output:
```
/usr/bin/php /home/steveb/PhpstormProjects/adpanel/adpanel/test.php
BEFORE KEY
ITERATOR EXECUTING
AFTER KEY
123
```

Therefore, because `$this->iterator` is an instance of a `Generator`, it does not start executing (even until the first `yield` statement) until any call to the usual `Iterator` methods.

`CachingIterator` is - as far as I understand it - used to immediately start the `Cursor` and this call does exactly that, similarly to how `storeCurrentItem` was doing.

alcaeus:
Sorry for the long delay. Logic looks good now. Could you please add tests for the builder/query functionality to make sure no The [QueryTest](https://github.com/doctrine/mongodb-odm/blob/master/tests/Doctrine/ODM/MongoDB/Tests/Functional/QueryTest.php) and BuilderTest classes (for [Aggregation](https://github.com/doctrine/mongodb-odm/blob/master/tests/Doctrine/ODM/MongoDB/Tests/Aggregation/BuilderTest.php) and [Query](https://github.com/doctrine/mongodb-odm/blob/master/tests/Doctrine/ODM/MongoDB/Tests/Query/BuilderTest.php)) serve as examples. After that, this is ready for a merge into 2.1 :)

Steveb-p:
@alcaeus I'll try to wrap this up tonight/tomorrow.

alcaeus:
No worries, caught up in things as well, so I can only finish it up next week anyway.

Steveb-p:
I've added tests in 9f347c6. If I misunderstood their purpose - being too generic or too narrow - please correct me.

They should not pass since I've deliberately marked them as expecting an exception to be thrown to show a little quirk/issue that those tests uncovered: empty resultsets will incorrectly **not** mark iterator as exhausted. This is due to the fact that foreach loop in wrapping `Generator` is never entered. I've tried setting `iteratorAdvanced` property to `true` after the `foreach` but since we expect the iterator to be started immediately this causes it to become unusable. I'm wondering if there is a graceful solution for this. WDYT?

alcaeus:
@Steveb-p once again, sorry for the long delay here, but I was not able to do any significant work over the past few weeks due to the current situation.

I updated the tests so they pass, and also added paragraphs to the query builder and aggregation builder documentation explaining this feature. Once I've got the other branches stable and merge up, this is good to merge.

Thanks for highlighting this issue, fixing it, and bringing along more patience a maintainer could ask for! ❤️

alcaeus (approved):

Issue #1975, opened 2019-03-11
add DateImmutableType


rieschl:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

It would be nice if `DateTimeImmutable` would be supported natively, as using immutable dates is usually better than mutable ones.
So I suggest implementing a `DateImmutableType`.
It's probably as easy as using the current `DateType` type and converting the Date to DateTimeImmutable.
That's my current implementation:

```
final class DateImmutableType extends DateType
{
    public const DATE_IMMUTABLE = 'date_immutable';

    public static function getDateTime($value): \DateTimeInterface
    {
        $dateTime = parent::getDateTime($value);
        if ($dateTime instanceof \DateTimeImmutable) {
            return $dateTime;
        }
        if (!$dateTime instanceof \DateTime) {
            throw new \InvalidArgumentException(
                \sprintf(
                    'Value has to be instance of DateTime, got %s',
                    get_class($dateTime))
            );
        }
        return \DateTimeImmutable::createFromMutable($dateTime);
    }
```
It's not yet fully tested, so use at your own risk.

malarzm:
Closing as #2118 was merged

PR #2118, opened 2019-11-23
Add DateImmutable type

malarzm:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | #1975

#### Summary

As per title :)

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -6,6 +6,7 @@ namespace Doctrine\ODM\MongoDB\Persisters;
 
 use BadMethodCallException;
 use DateTime;
+use DateTimeImmutable;
 use Doctrine\Common\Persistence\Mapping\MappingException;
 use Doctrine\ODM\MongoDB\DocumentManager;
 use Doctrine\ODM\MongoDB\Hydrator\HydratorException;
@@ -211,11 +212,11 @@ final class DocumentPersister
             if ($this->class->isVersioned) {
                 $versionMapping = $this->class->fieldMappings[$this->class->versionField];
                 $nextVersion    = null;
-                if ($versionMapping['type'] === 'int') {
+                if ($versionMapping['type'] === Type::INT || $versionMapping['type'] === Type::INTEGER) {
                     $nextVersion = max(1, (int) $this->class->reflFields[$this->class->versionField]->getValue($document));
                     $this->class->reflFields[$this->class->versionField]->setValue($document, $nextVersion);
-                } elseif ($versionMapping['type'] === 'date') {
-                    $nextVersionDateTime = new DateTime();
+                } elseif ($versionMapping['type'] === Type::DATE || $versionMapping['type'] === Type::DATE_IMMUTABLE) {
+                    $nextVersionDateTime = $versionMapping['type'] === Type::DATE ? new DateTime() : new DateTimeImmutable();
                     $nextVersion         = Type::convertPHPToDatabaseValue($nextVersionDateTime);
                     $this->class->reflFields[$this->class->versionField]->setValue($document, $nextVersionDateTime);
                 }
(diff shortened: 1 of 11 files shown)


lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -211,11 +212,11 @@ public function executeInserts(array $options = []) : void
             if ($this->class->isVersioned) {
                 $versionMapping = $this->class->fieldMappings[$this->class->versionField];
                 $nextVersion    = null;
-                if ($versionMapping['type'] === 'int') {
+                if ($versionMapping['type'] === Type::INT || $versionMapping['type'] === Type::INTEGER) {
```
malarzm: I'm not sure why we haven't used consts in the first place so I've used them everywhere and kept it as a separate commit

alcaeus: We should deprecate and remove the `boolean` and `integer` types to be consistent with PHP, which only allows `bool` and `int`. For now this works. 👍

lib/Doctrine/ODM/MongoDB/Types/DateImmutableType.php
```diff
@@ -0,0 +1,19 @@
...
+
+namespace Doctrine\ODM\MongoDB\Types;
+
+use DateTime;
+use DateTimeImmutable;
+use DateTimeInterface;
+
+class DateImmutableType extends DateType
+{
+    public static function getDateTime($value) : DateTimeInterface
```
alcaeus:
```suggestion
    /** @return DateTimeImmutable */
    public static function getDateTime($value) : DateTimeInterface
```

lib/Doctrine/ODM/MongoDB/Types/DateImmutableType.php
```diff
@@ -0,0 +1,19 @@
...
+    {
+        $datetime = parent::getDateTime($value);
+
+        return $datetime instanceof DateTime ? DateTimeImmutable::createFromMutable($datetime) : $datetime;
```
alcaeus: This can lead to returning something that isn't a `DateTimeImmutable` if so given from the parent class. I would suggest checking for that.

tests/Doctrine/ODM/MongoDB/Tests/Functional/LockTest.php
```diff
@@ -79,14 +79,14 @@ public function testMultipleFlushesDoIncrementalUpdates()
             $this->dm->persist($test);
             $this->dm->flush();
 
-            $this->assertIsInt($test->getVersion());
+            $this->assertInternalType('int', $test->getVersion());
```
alcaeus: Why the change? This method is deprecated in PHPUnit 8.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
malarzm: Created https://github.com/doctrine/mongodb-odm/issues/2124

lib/Doctrine/ODM/MongoDB/Types/DateImmutableType.php
malarzm: `DateTimeInterface` has only two implementations and IIRC it's forbidden to write a new one in userland. But if you prefer so I can check what's returned and throw an exception if it's neither `DateTime` or `DateTimeImmutable` 👍

tests/Doctrine/ODM/MongoDB/Tests/Functional/LockTest.php
malarzm: Somehow I had newest PHPUnit running and it complained while I was running entire `LockTest` class. Can revert it and keep the change for the upgrade PR 👍

lib/Doctrine/ODM/MongoDB/Types/DateImmutableType.php
malarzm: Shall I make similar change to the `DateType`?

tests/Doctrine/ODM/MongoDB/Tests/Functional/LockTest.php
alcaeus: Yes please, let's refactor that separately.

lib/Doctrine/ODM/MongoDB/Types/DateImmutableType.php
alcaeus: Yeah, that should help static inspection

malarzm: FTR: I'm leaving the original DateType as if it detects DateTimeInterface it's returning it right away and DateTime::createFromImmutable was introduced in PHP 7.3 which leaves me with no convenient way of transforming immutable datetime

PR #2123, opened 2019-11-25
Remove unneceesary internal designations

### LINKED ISSUE #2122: [deprecations] Create a DocumentManagerInterface to prevent deprecations on @interal
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes*
| Version     | 2.0.2
* but only for automated testing without deprecations, which breaks our CI.

#### Summary

Since updating to Symfony 4.4 I have this deprecation message in the automated tests:

The "Doctrine\ODM\MongoDB\DocumentManager::getClassMetadata()" method is considered internal Performance-sensitive method. It may change without further notice. You should not extend it from "Mockery_11_Doctrine_ODM_MongoDB_DocumentManager".

This is due to DocumentManager.php line 280:
"@internal Performance-sensitive method."

#### How to reproduce

basic Symfony 4.4 project with a controller and a test like this: (Note could have been any service, not necessarily a controller)

```
class SomeController {
  public function test(DocumentManager $dm) {
    $dm->flush();
  }
}
```

```
class SomeControllerTest {
  public function test() {
    $om = \Mockery::mock(DocumentManager::class);
    (new SomeController)->test($om);
  }
}
```

#### Expected behavior

No deprecations:

### Suggested solution

Easily fixed by creating an `DocumentManagerInterface`, just like the `EntityManagerInterface`. This way we can write automated tests using the interface and have no problems with internal annotations like this. This solves other issues too.

I know there is a `interface ObjectManager` that could be used for this case, but in my specific use-case, I'm using both the ODM and the ORM in the same Symfony application, and due to the nature of Dependency Injection it's not logical to use the ObjectManager, which is ambiguous in this case.

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | (partly) fixes #2122

#### Summary

This removes a few `@internal` designations for methods that aren't really internal, but where the comment was used to leave notes that shouldn't be exposed in an API documentation.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/DocumentManager.php
+++ b/lib/Doctrine/ODM/MongoDB/DocumentManager.php
@@ -283,7 +283,6 @@ class DocumentManager implements ObjectManager
      *
      * @param string $className The class name.
      * @return \Doctrine\ODM\MongoDB\Mapping\ClassMetadata
-     * @internal Performance-sensitive method.
      */
     public function getClassMetadata($className)
     {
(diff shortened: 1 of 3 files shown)


malarzm:
Thanks for taking the time to finally remove these comments :)

Issue #1835, opened 2018-07-27
Support Decimal128 type


alcaeus:
See #1815: we want to support the Decimal128 type, converting values to strings for use with `bcmath`. This requires updating some operators and logic (e.g. when using `increment` strategy) to handle large decimal values.

malarzm:
I've started working on this but I need to have https://github.com/doctrine/mongodb-odm/pull/2118 merged first as I'm sure I'll have conflicts.

alcaeus:
Fixed in #2128.

PR #2128, opened 2019-11-29
Add Decimal128 type

malarzm:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | #1835

#### Summary

As per title, again ;)

### DIFF
--- a/docs/en/reference/basic-mapping.rst
+++ b/docs/en/reference/basic-mapping.rst
@@ -150,6 +150,7 @@ Here is a quick overview of the built-in mapping types:
 -  ``custom_id``
 -  ``date``
 -  ``date_immutable``
+-  ``decimal128``
 -  ``file``
 -  ``float``
 -  ``hash``
@@ -182,6 +183,7 @@ This list explains some of the less obvious mapping types:
 -  ``collection``: numerically indexed array to MongoDB array
 -  ``date``: DateTime to ``MongoDB\BSON\UTCDateTime``
 -  ``date_immutable``: DateTimeImmutable to ``MongoDB\BSON\UTCDateTime``
+-  ``decimal128``: string to ``MongoDB\BSON\Decimal128``, requires ``ext-bcmath``
 -  ``hash``: associative array to MongoDB object
 -  ``id``: string to ObjectId by default, but other formats are possible
 -  ``timestamp``: string to ``MongoDB\BSON\Timestamp``
--- a/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
@@ -25,6 +25,7 @@ use function array_pop;
 use function class_exists;
 use function constant;
 use function count;
+use function extension_loaded;
 use function get_class;
 use function in_array;
 use function is_array;
@@ -1974,6 +1975,7 @@ class ClassMetadata implements BaseClassMetadata
 
         $this->applyStorageStrategy($mapping);
         $this->checkDuplicateMapping($mapping);
+        $this->typeRequirementsAreMet($mapping);
 
         $this->fieldMappings[$mapping['fieldName']] = $mapping;
         if (isset($mapping['association'])) {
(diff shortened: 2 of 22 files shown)


malarzm:
Please ignore CS issues for now, there's something fishy going on with my local instance of `phpcbf`. Also https://github.com/doctrine/mongodb-odm/pull/2128/commits/3d6ac962f446a987a0c6fc9a2f66a58e2fc21c8d is a good candidate for a PR on its own and can be extracted as such. Other than that we have a working increment support for decimals (and user-land types) :)

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -1196,9 +1191,18 @@ private function applyStorageStrategy(array &$mapping) : void
                 ];
                 break;
 
+            case $mapping['type'] === 'one':
```
alcaeus: This doesn't need to be extracted: it is still covered by the `default` condition and will apply there.

docs/en/reference/basic-mapping.rst
```diff
@@ -182,6 +183,7 @@ This list explains some of the less obvious mapping types:
 -  ``collection``: numerically indexed array to MongoDB array
 -  ``date``: DateTime to ``MongoDB\BSON\UTCDateTime``
 -  ``date_immutable``: DateTimeImmutable to ``MongoDB\BSON\UTCDateTime``
+-  ``decimal128``: string to ``MongoDB\BSON\Decimal128``
```
alcaeus: Please add a note that this type requires the `ext-bcmath` extension.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -380,6 +388,12 @@ public function update(object $document, array $options = []) : void
                 $nextVersion                             = $versionMapping['type'] === Type::DATE ? new DateTime() : new DateTimeImmutable();
                 $update['$set'][$versionMapping['name']] = Type::convertPHPToDatabaseValue($nextVersion);
                 $query[$versionMapping['name']]          = Type::convertPHPToDatabaseValue($currentVersion);
+            } elseif ($versionMapping['type'] === Type::DECIMAL128) {
+                $current = $this->class->reflFields[$this->class->versionField]->getValue($document);
+                $nextVersion = bcadd($current, '1');
```
alcaeus: Similar to the incrementable type, this could be simplified by adding an interface for types suitable for versioning. I'm happy to defer this to a separate PR though.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
malarzm: I've added a call to `Type::getType($mapping['type']);` in `default` branch which fails miserably when handed `one` :) Likely we shouldn't allow registering such named types in userland.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
malarzm: That was my intention before I realized the original issue requested `$inc` and not versioning :) Can do that in next PR tho

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
alcaeus: Ah, I see. No need to change then.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
alcaeus: 👍

docs/en/reference/basic-mapping.rst
malarzm: Done 👍

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -2083,6 +2089,13 @@ private function isAllowedGridFSField(string $name) : bool
         return in_array($name, self::ALLOWED_GRIDFS_FIELDS, true);
     }
 
+    private function typeRequirementsAreMet(string $type) : void
+    {
+        if ($type === Type::DECIMAL128 && ! extension_loaded('bcmath')) {
+            throw MappingException::typeRequirementsNotFulfilled(Type::DECIMAL128, 'ext-bcmath is missing');
```
alcaeus: Adding the field and class names to the exception message will make it easier for users to spot which piece of their code is causing these errors.

alcaeus:
Thanks Maciej!

PR #2131, opened 2019-12-03
FIX - info message must be lowercase

floverdevel:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

Some messages are rendered as `Hydrator classes generated to "path/to/my/project/var/pre-cached/Hydrators</INFO>"`.
Because the opening and closing tags are not case sensitive.
So i use `</info>` instead of `</INFO>` for these messages.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Tools/Console/Command/GenerateHydratorsCommand.php
+++ b/lib/Doctrine/ODM/MongoDB/Tools/Console/Command/GenerateHydratorsCommand.php
@@ -97,7 +97,7 @@ EOT
             $dm->getHydratorFactory()->generateHydratorClasses($metadatas, $destPath);
 
             // Outputting information message
-            $output->write(PHP_EOL . sprintf('Hydrator classes generated to "<info>%s</INFO>"', $destPath) . PHP_EOL);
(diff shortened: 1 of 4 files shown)


floverdevel:
Allright, I will do that tonight 😁

Steveb-p:
> 3\. Force push to overwrite all this : `git push --force`.

`--force-with-lease` is usually the safer version that ensures no commits from remote are lost if someone pushes mid `fetch-push` operation.

alcaeus:
@steveb-p: while I normally agree, that entire point is moot if you’ve just fetched and you’re pushing to a repository/branch where you’re the only one pushing to 😉

floverdevel:
I rebased on `1.3.x` but i supposed I should also change the destination of my PR to `1.3.x` instead of `master`.

alcaeus:
> I rebased on `1.3.x` but i supposed I should also change the destination of my PR to `1.3.x` instead of `master`.

Yep - that's the part that most people don't have trouble with, which is why I omitted it. You are right of course.

Thanks for fixing this, @floverdevel!

PR #2135, opened 2019-12-10
Add md5 to allowed gridfs fields

derzkiy:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

Read md5 file info for some checks


alcaeus:
The `md5`, `contentType` and `aliases` fields in the GridFS document are deprecated in the [GridFS spec](https://github.com/mongodb/specifications/blob/master/source/gridfs/gridfs-spec.rst). For that reason, we have decided to no longer allow mapping those fields in files.

webdevilopers:
If I see this correctly then `md5` is deprecated but still inserted by MongoDB by default no matter the mapping. Can you confirm?

alcaeus:
It is inserted by the MongoDB library unless you specify the `disableMD5` option when uploading the file. Since it's deprecated we have to decide it not to allow mapping it. I'm not sure when (or even if) the setting will be deprecated in the driver, but I wouldn't rely on the field being there at all. If you want to store a hash of the file, please create your own (non-md5) hash and store it in metadata (which it is).

PR #2150, opened 2020-01-10
Address deprecations from persistence

greg0ire:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | n/a

#### Summary

A backwards-compatibility layer has been added to persistence to help
consumers move to the new namespacing. It is based on class aliases,
which means the type declaration changes should not be a BC-break: types
are the same.
See doctrine/persistence#71

This means:
- using the new namespaces
- adding autoload calls for new types to types that may be extended and
use persistence types in type declarations of non-constructor methods,
so that signature compatibility is recognized by old versions of php.
More details on this at
https://dev.to/greg0ire/how-to-deprecate-a-type-in-php-48cf


greg0ire:
The build fails because persistence requires php 7.1 . Could we drop php 7.0?

malarzm:
I think we can safely do that in 2.1 👍

greg0ire:
:thinking: why not 2.0.4 ? Also, should we forbid persistence 1.4 and up on this branch? That way people don't get deprecations they cannot fix?

malarzm:
Oh this is targeting `1.3.x`... Here we're still allowing 5.6 so no way we can bump to 7.1. I guess it'll deprecations from the common will be one more poke for folks to upgrade ODM.

> why not 2.0.4 ?

PHP version can be bumped on a minor release, we even made a blog post about it :) https://www.doctrine-project.org/2017/07/25/php-7.1-requirement-and-composer.html

greg0ire:
But… 2.0.x already requires `7.2`, doesn't it?

malarzm:
Yeah, this is what happens when I'm replying in the night. I'm sorry @greg0ire you're right, we can totally include this in 2.0.4. I was confused by you saying we need to drop 7.0 and thought 2.0 is still supporting that version.

docs/en/cookbook/blending-orm-and-mongodb-odm.rst
```diff
@@ -3,12 +3,12 @@ Blending the ORM and MongoDB ODM
 
 Since the start of the `Doctrine MongoDB Object Document Mapper`_ project people have asked how it can be integrated with the `ORM`_. This article will demonstrates how you can integrate the two transparently, maintaining a clean domain model.
 
-This example will have a `Product` that is stored in MongoDB and the `Order` stored in a MySQL database.
+This example will have a ``Product`` that is stored in MongoDB and the ``Order`` stored in a MySQL database.
```
greg0ire: I don't recall doing these changes but they are great.

greg0ire:
There! Got it to a mergeable state again :P

lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -475,3 +476,5 @@ public function getProxyManagerConfiguration() : ProxyManagerConfiguration
         return $this->proxyManagerConfiguration;
     }
 }
+
+interface_exists(MappingDriver::class);
```
malarzm: Why do we need this line? Do we need to autoload something?

greg0ire: Yes exactly. Please read the blog post I linked above, it's fairly complicated.

TL;DR: it's because of `public function setMetadataDriverImpl(MappingDriver $driverImpl) : void`. People extending `Configuration` and overriding this method with the old type declaration might get a crash in some situations.

alcaeus:
Thanks @greg0ire!

lib/Doctrine/ODM/MongoDB/Configuration.php
rubenrua: IMHO Adding a comment would be a good idea

greg0ire: It's a bit late now, this has been merged. This is explained at length in the blog post, which is linked in the commit message though.

Steveb-p: Since almost everyone wonders why it is used I'd say a comment should be added. Three months from now someone will look on this code and remove it, since "it doesn't do anything" :smile:

greg0ire: Do you think they would really do so without taking a look at the git history? Anyway, feel free to make a PR with comments, I will not block it, I think I might even have seen such comments in other places where we have similar autoload calls.

Steveb-p: Even if they do, they will simply find messages that it's related to deprecations. It will not hint why a particular code was introduced and what side effects it brings.

IMO the cost of this comment outweights the potential time that one would spend trying to figure out what it does - but it's just that, an opinion :smile:

greg0ire: > Even if they do, they will simply find messages that it's related to deprecations.

What about the link to the blog post? Didn't you read it?

Steveb-p: I sure did. But I doubt someone looking at the code will visit this PR or check it's description / comments. There were even people in the comments to your blog post that tried to invalidate this approach.

I don't mean to criticize if that's how you feel and I agree with you. However, it isn't obvious what code does, and only that. Git history is helpful only if you already know why it might be used.

greg0ire: > But I doubt someone looking at the code will visit this PR or check it's description / comments

First, they don't need to, using blame on the file is enough to find the commit: https://github.com/doctrine/mongodb-odm/blame/2.0.x/lib/Doctrine/ODM/MongoDB/Mapping/Driver/AnnotationDriver.php

and the commit itself has a comment and a link to the blog post.

Second, why wouldn't they? I do that all the time… but if you still feel otherwise, like I said, you can absolutely create a PR with the comment. It certainly won't hurt.

PR #2159, opened 2020-01-26
Deprecation and forward compat for flush

malarzm:
We missed this originally. Patch also includes forward compatibility layer so consumers may pass `$options` as a first argument already. Also there's a chance I should have placed similar code in `UnitOfWork::commit` although I haven't thought about any legit reason to call it directly instead of `DocumentManager::flush` so I didn't bother

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/DocumentManager.php
+++ b/lib/Doctrine/ODM/MongoDB/DocumentManager.php
@@ -531,17 +531,32 @@ class DocumentManager implements ObjectManager
      * This effectively synchronizes the in-memory state of managed objects with the
      * database.
      *
-     * @param object $document
+     * @param object|array $documentOrOptions
      * @param array $options Array of options to be used with batchInsert(), update() and remove()
      * @throws \InvalidArgumentException
      */
-    public function flush($document = null, array $options = array())
+    public function flush($documentOrOptions = null, array $options = array())
     {
-        if (null !== $document && ! is_object($document) && ! is_array($document)) {
-            throw new \InvalidArgumentException(gettype($document));
+        if (func_num_args() === 1 && is_array($documentOrOptions) && ! $this->arrayContainsDocuments($documentOrOptions)) {
+            $options = $documentOrOptions;
+            $documentOrOptions = null;
+        }
+        if (null !== $documentOrOptions && ! is_object($documentOrOptions) && ! is_array($documentOrOptions)) {
+            throw new \InvalidArgumentException(gettype($documentOrOptions));
+        }
+        if (!empty($documentOrOptions)) {
+            @trigger_error(
+                'Flushing selected documents has been deprecated and will be removed in doctrine/mongodb-odm 2.0.',
+                E_USER_DEPRECATED
+            );
+        } elseif (func_num_args() === 2) {
+            @trigger_error(
+                sprintf('Flushing selected documents has been deprecated and will be removed in doctrine/mongodb-odm 2.0. You can omit the first argument to %s.', __METHOD__),
+                E_USER_DEPRECATED
+            );
         }
         $this->errorIfClosed();
-        $this->unitOfWork->commit($document, $options);
+        $this->unitOfWork->commit($documentOrOptions, $options);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


lib/Doctrine/ODM/MongoDB/DocumentManager.php
```diff
@@ -531,17 +531,33 @@ public function getRepository($documentName)
...
+        }
+        if ($documentOrOptions !== null) {
+            @trigger_error(
+                'Flushing selected documents has been deprecated and will be removed in doctrine/mongodb-odm 2.0.',
+                E_USER_DEPRECATED
+            );
+        }
+        if (func_num_args() === 2) {
+            @trigger_error(
+                '$options will become a first argument in doctrine/mongodb-odm 2.0, you may pass it as first now.',
```
malarzm: I'm not sure about wording here

lib/Doctrine/ODM/MongoDB/DocumentManager.php
```diff
@@ -531,17 +531,33 @@ public function getRepository($documentName)
...
+        if (null !== $documentOrOptions && ! is_object($documentOrOptions) && ! is_array($documentOrOptions)) {
+            throw new \InvalidArgumentException(gettype($documentOrOptions));
+        }
+        if ($documentOrOptions !== null) {
+            @trigger_error(
+                'Flushing selected documents has been deprecated and will be removed in doctrine/mongodb-odm 2.0.',
+                E_USER_DEPRECATED
+            );
+        }
+        if (func_num_args() === 2) {
```
alcaeus: We should only trigger the deprecation notice if documents were passed:

```suggestion
        elseif (func_num_args() === 2) {
```

lib/Doctrine/ODM/MongoDB/DocumentManager.php
alcaeus:
```suggestion
                sprintf('Flushing selected documents has been deprecated and will be removed in doctrine/mongodb-odm 2.0. You can omit the first argument to %s.', __METHOD__),
```

lib/Doctrine/ODM/MongoDB/DocumentManager.php
```diff
@@ -842,4 +858,18 @@ public function getFilterCollection()
 
         return $this->filterCollection;
     }
+
+    private function arrayContainsDocuments(array $documentOrOptions)
+    {
+        foreach ($documentOrOptions as $documentPerhaps) {
+            if (! is_object($documentPerhaps)) {
+                return false;
```
alcaeus: Any reason why you decided to treat the array as options upon encountering a non-object value? Given what we expect, I'd actually say that unless none of the given values is a transient document, we may want to treat it as such and keep the original behaviour.

```suggestion
                continue;
```

lib/Doctrine/ODM/MongoDB/DocumentManager.php
malarzm: Not necessarily, `$dm->flush(null, $options);` will stop working as well and folks can call `$dm->flush($options);` now

lib/Doctrine/ODM/MongoDB/DocumentManager.php
malarzm: Reasoning is that any and all documents must be an object in the first place. So if it's not an object, something would have failed down the line anyway as unit of work would be trying to compute change set or an array or scalar

lib/Doctrine/ODM/MongoDB/DocumentManager.php
alcaeus: As is, `$dm->flush([$document], $options);` throws two deprecation warnings. We should make sure this doesn't happen as it might confuse the user.

As I write this, we should also make sure to gracefully handle `$dm->flush([], $options);` to make sure we don't treat it as the "user passed documents" case.

lib/Doctrine/ODM/MongoDB/DocumentManager.php
alcaeus: Ok, that makes sense. We can keep it this way then

lib/Doctrine/ODM/MongoDB/DocumentManager.php
rieschl: It seems that passing an array of documents is an undocumented feature. I didn't find anything in the documentation and the method signature mentions just _one_ object. Only [deep down in the UoW](https://github.com/doctrine/mongodb-odm/blob/1.3.x/lib/Doctrine/ODM/MongoDB/UnitOfWork.php#L404) is the check if the parameter is an array. So did anyone even use that?

malarzm: Hard to say, but there's a chance :)

Issue #2157, opened 2020-01-22
DiscriminatorMap is breaking Facet Aggregation Builders


Ludo444:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.0.2

#### Summary

If there is a `Doctrine\ODM\MongoDB\Aggregation\Stage\Facet` used on a pipeline with document using `DiscriminatorMap`, when calling `getPipeline()` method inside `Facet`, it prepends the pipeline with filtering `$match` step that queries `DiscriminatorField`. This filtering may fail in certain pipelines as the field may be missing.

#### Current behavior

During later stages of a pipeline the `DiscriminatorField` may not be present in document list and `Facet` stage can contain empty results because of the filtering.

#### How to reproduce

Have a document that contains similar annotation:
```
 * @ODM\DiscriminatorField("type")
 * @ODM\DiscriminatorMap({"typ1"=Type1Document::class, "bracket"=Type2Document::class})
 */
abstract class Document
```

Use a `Facet` stage on an aggregation `Builder` that queries above type of `Document`.

#### Suggested solution

I think in case when `Builder` is used inside the `Facet` stage, it should ignore `DiscriminatorMap` filtering as the `DiscriminatorField` may not be present in pipeline anymore. Either by passing a param to `getPipeline()` method or let the `Builder` instance somehow know it should not add those extra filters there.

malarzm:
@Ludo444 thanks for the report! Could you maybe create a PR with a failing test case? If you don't feel like diving too much into our tests, a functional one [like one of these](https://github.com/doctrine/mongodb-odm/tree/master/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket) will suffice.

Ludo444:
@malarzm hopefully i can look to it sometime this week

Ludo444:
I created PR with test case and possible fix. Not sure if it's the most elegant solution but it seems to solve the problem fine.

malarzm:
@Ludo444 thank you very much! I've asked Andreas to take a look there :)

alcaeus:
Fixed in #2160. Thanks for the PR, @Ludo444 and sorry for the delay!

PR #2160, opened 2020-01-28
Bug/2157 Test (and fix) Facet's aggregation pipeline with DiscriminatorMap documents

Ludo444:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #2157 

#### Summary

Provide a bool flag for `getPipeline()` if it's being called from inside the `Facet` and test case with `DiscriminatorMap` annotated document.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
@@ -15,11 +15,16 @@ use GeoJson\Geometry\Point;
 use MongoDB\Collection;
 use MongoDB\Driver\Cursor;
 use OutOfRangeException;
+use TypeError;
 use function array_map;
 use function array_merge;
 use function array_unshift;
 use function assert;
+use function func_get_arg;
+use function func_num_args;
+use function gettype;
 use function is_array;
+use function is_bool;
 use function sprintf;
 
 /**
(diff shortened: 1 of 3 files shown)


lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
```diff
@@ -214,36 +219,63 @@ public function geoNear($x, $y = null) : Stage\GeoNear
...
      * the geoNear operation. For all other pipelines, it prepends a $match stage
      * containing the required query.
+     *
+     * For aggregation pipelines that will be nested (e.g. in a facet stage),
+     * you should not apply filters as this may cause wrong results to be
+     * given.
      */
-    public function getPipeline() : array
+    // phpcs:enable Squiz.Commenting.FunctionComment.ExtraParamComment
+    public function getPipeline(/* bool $applyFilters = true */) : array
```
alcaeus, line 239: Since we need to preserve BC, we can't add a new method argument. Instead, we add it commented and use `func_num_args` to check for it being given. The `if` below then emulates strict typing, throwing an exception if we get anything but a bool.

Note that I've also changed the logic of the argument: instead of calling it `isFacet`, we call it `applyFilters`. This is mainly because upcoming changes to the `$lookup` stage will require us using the same functionality, so `isFacet` would be the wrong name.

lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
```diff
@@ -214,36 +219,63 @@ public function geoNear($x, $y = null) : Stage\GeoNear
...
             static function (Stage $stage) {
                 return $stage->getExpression();
             },
             $this->stages
         );
 
-        if ($this->getStage(0) instanceof Stage\GeoNear) {
-            $pipeline[0]['$geoNear']['query'] = $this->applyFilters($pipeline[0]['$geoNear']['query']);
-        } elseif ($this->getStage(0) instanceof Stage\IndexStats) {
+        if ($this->getStage(0) instanceof Stage\IndexStats) {
```
alcaeus, line 258: Applied some formatting and early exits here to avoid unnecessary nesting.

lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
malarzm: I think we'd do just fine with new argument with default value, but since you've gone the extra mile with this approach let's keep it :+1: The only possible downside I see is PHPStan complaining about passing a parameter that is not there. Unless it'll take phpdoc's word?

alcaeus: It should take PHPDoc's word. However, adding a new parameter, even with default value, is a BC break: https://3v4l.org/jobIX

malarzm: It sure is, my point was more "who extends a builder" :) Sorry for not making myself clear

malarzm: OTOH the builder is not final so my point was invalid

PR #2166, opened 2020-02-18
Views

### LINKED ISSUE #1925: Support for MongoDB Views
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

With MongoDB 3.4, MongoDB introduced read-only views based on aggregation pipelines. Read [the documentation](https://docs.mongodb.com/manual/core/views/) for basic information on views.

Support for views should be added to ODM:
* Allow mapping a class as a view
* Any persist/remove operations on objects of this class will cause errors
* View objects can be proxied (the aggregation pipeline could build a reference to another view)
* The mapping needs to contain the aggregation pipeline used to build the view

The last item is the most difficult point: for annotations, it'd be relatively easy to just write an aggregation pipeline stage into the mapping, but for XML mappings this would be more difficult. One option would be to require a repository that provides a static method that returns an aggregation builder instance. We could then use that instance to create the view in schema commands.

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | Fixes #1925.

#### Summary

As per the ticket, this is an implementation of MongoDB views in ODM. Views are treated as a separate mapped object class, similar to files, documents, and query results. Views cannot be proxied (for now), and their repository must implement a specific `ViewRepository` interface. Thus, a view cannot be mapped without a custom repository.

The `ViewRepository` interface defines a single method:
```php
public function createViewAggregation(Builder $builder) : void;
```

This method receives an aggregation builder instance based on the document specified in the `viewOn` mapping attribute. This was done as the view needs to be static, so any dynamic changes (e.g. by choosing a different class to base the view on) must be avoided. The aggregation builder instance is then used to build the aggregation pipeline that defines the view. Exposing a builder instance will allow for more flexibility down the line, as we could leverage this builder to optimise queries, e.g. if there is an inverse reference pointing to a view).

As for always requiring a custom repository, I had considered letting the mapped class define the pipeline, but I thought that the repository was a better fit than a static method on the entity (potentially with a custom name).

This PR covers:
* [x] Annotation and XML mappings for views
* [x] Schema manager logic to create and drop views
* [x] Refactoring `AbstractRepositoryFactory` to throw exceptions when trying to get the repository for an invalid class (e.g. an embedded document)
* [x] Refactoring the mapping driver test to use the document manager created: this allows us to better test mappings
* [x] Documentation
* [x] Allow using views as targetDocument for references, including proxy handling

### DIFF
--- a/docs/en/reference/annotations-reference.rst
+++ b/docs/en/reference/annotations-reference.rst
@@ -1110,6 +1110,67 @@ encouraged to use the :ref:`atomicSet <atomic_set>` or
 will ensure that collections are updated in the same write operation as the
 versioned parent document.
 
+@View
+-----
+
+Required annotation to mark a PHP class as a view. Views are created from
+aggregation pipelines, which are returned from a special repository method.
+Views can be used like collections for any read operations. Result documents are
+not managed and cannot be referenced using the :ref:`reference-many <reference_many>`
+and :ref:`reference-one <reference_one>` mappings.
+
+Required attributes:
+
+-
+   ``rootClass`` - this is the base collection that the view is created from
+-
+   ``repositoryClass`` - a repository class is required. This repository must
+   implement the ``MongoDB\ODM\MongoDB\Repository\ViewRepository`` interface.
+
+Optional attributes:
+
+-
+   ``db`` - By default, the document manager will use the MongoDB database
+   defined in the configuration, but this option may be used to override the
+   database for a particular document class.
+-
+   ``view`` - By default, the view name is derived from the document's class
+   name, but this option may be used to override that behavior.
+
+.. code-block:: php
+
+    <?php
+
+    /**
+     * @View(
+     *     db="documents",
+     *     rootClass=User::class,
+     *     repositoryClass=UserNameRepository::class,
+     * )
+     */
+    class UserName
+    {
+        //...
+    }
+
+    class UserNameRepository implements \Doctrine\ODM\MongoDB\Repository\ViewRepository
+    {
+        public function createViewAggregation(Builder $builder) : void
+        {
+            $builder->project()
+                ->includeFields(['username']);
+        }
+    }
+
+The ``createViewAggregation`` method can add any aggregation pipeline stage,
+except for the ``$out`` and ``$merge`` stages. The pipeline is created for the
+root class specified in the view mapping.
+
+.. note::
+
+    Views must be created before they can be queried. This can be done using the
+    ``odm:schema:create`` command.
+
 .. _BSON specification: http://bsonspec.org/spec.html
 .. _DBRef: https://docs.mongodb.com/manual/reference/database-references/#dbrefs
 .. _geoNear command: https://docs.mongodb.com/manual/reference/command/geoNear/
--- a/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
@@ -434,6 +434,13 @@ class ClassMetadata implements BaseClassMetadata
     public $isQueryResultDocument = false;
 
     /**
+     * READ-ONLY: Whether this class describes the mapping of a database view.
+     *
+     * @var bool
+     */
+    private $isView = false;
+
+    /**
      * READ-ONLY: Whether this class describes the mapping of a gridFS file
      *
      * @var bool
@@ -500,6 +507,9 @@ class ClassMetadata implements BaseClassMetadata
 
     /** @var InstantiatorInterface */
     private $instantiator;
+
+    /** @var string|null */
+    private $rootClass;
 
     /**
      * Initializes a new ClassMetadata instance that will hold the object-document mapping
@@ -1719,6 +1729,22 @@ class ClassMetadata implements BaseClassMetadata
         $this->isReadOnly = true;
     }
 
+    public function getRootClass() : ?string
+    {
+        return $this->rootClass;
+    }
+
+    public function isView() : bool
+    {
+        return $this->isView;
+    }
+
+    public function markViewOf(string $rootClass) : void
+    {
+        $this->isView    = true;
+        $this->rootClass = $rootClass;
+    }
+
     /**
      * {@inheritDoc}
      */
@@ -2020,6 +2046,11 @@ class ClassMetadata implements BaseClassMetadata
 
         if ($this->isQueryResultDocument) {
             $serialized[] = 'isQueryResultDocument';
+        }
+
+        if ($this->isView()) {
+            $serialized[] = 'isView';
+            $serialized[] = 'rootClass';
         }
 
         if ($this->isFile) {
--- a/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
@@ -659,6 +659,10 @@ class ClassMetadata implements BaseClassMetadata
      */
     public function invokeLifecycleCallbacks(string $event, object $document, ?array $arguments = null) : void
     {
+        if ($this->isView()) {
+            return;
+        }
+
         if (! $document instanceof $this->name) {
             throw new InvalidArgumentException(sprintf('Expected document class "%s"; found: "%s"', $this->name, get_class($document)));
         }
--- a/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataFactory.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataFactory.php
@@ -136,7 +136,7 @@ final class ClassMetadataFactory extends AbstractClassMetadataFactory
     {
         assert($class instanceof ClassMetadata);
 
-        return ! $class->isMappedSuperclass && ! $class->isEmbeddedDocument && ! $class->isQueryResultDocument;
+        return ! $class->isMappedSuperclass && ! $class->isEmbeddedDocument && ! $class->isQueryResultDocument && ! $class->isView();
     }
 
     /**
@@ -223,7 +223,7 @@ final class ClassMetadataFactory extends AbstractClassMetadataFactory
      */
     protected function validateIdentifier(ClassMetadata $class) : void
     {
-        if (! $class->identifier && ! $class->isMappedSuperclass && ! $class->isEmbeddedDocument && ! $class->isQueryResultDocument) {
+        if (! $class->identifier && $this->isEntity($class)) {
             throw MappingException::identifierRequired($class->name);
         }
     }
--- a/lib/Doctrine/ODM/MongoDB/Mapping/MappingException.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/MappingException.php
@@ -255,4 +255,14 @@ final class MappingException extends BaseMappingException
             $expectedRepositoryClass
         ));
     }
+
+    public static function viewWithoutRootClass(string $className) : self
+    {
+        return new self(sprintf('Class "%s" mapped as view without must have a root class.', $className));
+    }
+
+    public static function viewRootClassNotFound(string $className, string $rootClass) : self
+    {
+        return new self(sprintf('Root class "%s" for view "%s" could not be found.', $rootClass, $className));
+    }
 }
--- a/lib/Doctrine/ODM/MongoDB/Mapping/MappingException.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/MappingException.php
@@ -245,4 +245,14 @@ final class MappingException extends BaseMappingException
     {
         return new self(sprintf("Class '%s' cannot be discriminated because it is marked as a GridFS file", $className));
     }
+
+    public static function invalidRepositoryClass(string $className, string $repositoryClass, string $expectedRepositoryClass) : self
+    {
+        return new self(sprintf(
+            'Invalid repository class "%s" for mapped class "%s". It must be an instance of "%s".',
+            $repositoryClass,
+            $className,
+            $expectedRepositoryClass
+        ));
+    }
 }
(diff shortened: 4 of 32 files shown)


lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -320,6 +329,28 @@ public function createDocumentCollection(string $documentName, ?int $maxTimeMs =
...
+        if ($class->isView()) {
+            $options = $this->getWriteOptions($maxTimeMs, $writeConcern);
+
+            $rootClass = $class->getRootClass();
+            assert(is_string($rootClass));
+
+            $builder    = $this->dm->createAggregationBuilder($rootClass);
+            $repository = $this->dm->getRepository($class->name);
+            assert($repository instanceof ViewRepository);
+            $repository->createViewAggregation($builder);
```
malarzm: Should we have an assert that nobody changed builder's root class or am I paranoid?

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -127,7 +127,7 @@ public function __construct(
         $this->class           = $class;
         $this->cp              = $this->uow->getCollectionPersister();
 
-        if ($class->isEmbeddedDocument || $class->isQueryResultDocument) {
+        if ($class->isEmbeddedDocument || $class->isQueryResultDocument || $class->isView()) {
```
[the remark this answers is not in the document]
alcaeus: This was done years ago when there was a notable impact. IIRC, a single method call does not have such a heavy impact. Ping @beberlei, as he might know better.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -433,6 +433,13 @@ class ClassMetadata implements BaseClassMetadata
      */
     public $isQueryResultDocument = false;
 
+    /**
+     * READ-ONLY: Whether this class describes the mapping of a database view.
+     *
+     * @var bool
+     */
+    private $isView = false;
```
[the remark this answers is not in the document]
alcaeus: I made it private for a more consistent mapping: a view without root class is no longer possible.

lib/Doctrine/ODM/MongoDB/Repository/AbstractRepositoryFactory.php
```diff
@@ -44,18 +47,43 @@ public function getRepository(DocumentManager $documentManager, string $document
...
+                if (! is_a($repositoryClassName, ViewRepository::class, true)) {
+                    throw MappingException::invalidRepositoryClass($documentName, $repositoryClassName, ViewRepository::class);
+                }
+                break;
+
+            case $metadata->isEmbeddedDocument:
+                throw MongoDBException::cannotCreateRepository($documentName);
+                break;
+
+            case $metadata->isMappedSuperclass:
```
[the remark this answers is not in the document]
alcaeus: That's correct.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
alcaeus: Not sure how that would be possible? The metadata class is set in the constructor and can't be changed.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -300,10 +302,17 @@ public function createCollections(?int $maxTimeMs = null, ?WriteConcern $writeCo
     {
         foreach ($this->metadataFactory->getAllMetadata() as $class) {
             assert($class instanceof ClassMetadata);
-            if ($class->isMappedSuperclass || $class->isEmbeddedDocument || $class->isQueryResultDocument) {
-                continue;
+
+            switch (true) {
```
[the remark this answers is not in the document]
alcaeus: I can change it back. Previous iteration had different handling for views, so it made sense. It's not necessary right now.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
malarzm: Consider it resolved then, thought it'd be somehow possible :)

lib/Doctrine/ODM/MongoDB/Repository/AbstractRepositoryFactory.php
```diff
@@ -44,18 +47,43 @@ public function getRepository(DocumentManager $documentManager, string $document
...
+
+            case $metadata->isView():
+                if (! is_a($repositoryClassName, ViewRepository::class, true)) {
+                    throw MappingException::invalidRepositoryClass($documentName, $repositoryClassName, ViewRepository::class);
+                }
+                break;
+
+            case $metadata->isEmbeddedDocument:
+                throw MongoDBException::cannotCreateRepository($documentName);
+                break;
```
malarzm: Isn't this reported by PHPStan as unreachable code?

lib/Doctrine/ODM/MongoDB/SchemaManager.php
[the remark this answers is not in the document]
malarzm: Yes please, it's more concise

lib/Doctrine/ODM/MongoDB/Repository/AbstractRepositoryFactory.php
alcaeus: Nope 🤔

lib/Doctrine/ODM/MongoDB/SchemaManager.php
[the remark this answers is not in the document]
alcaeus: Most of these were changed back.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -2022,6 +2048,10 @@ public function __sleep()
             $serialized[] = 'isQueryResultDocument';
         }
 
+        if ($this->isView()) {
+            $serialized[] = 'isView';
```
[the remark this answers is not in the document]
alcaeus: Oh, yes.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -649,6 +659,10 @@ public function setCustomRepositoryClass(?string $repositoryClassName) : void
      */
     public function invokeLifecycleCallbacks(string $event, object $document, ?array $arguments = null) : void
     {
+        if ($this->isView()) {
```
malarzm: What's the reason this is blocked for views but not for `QueryResultDocuments`?

alcaeus: QueryResultDocuments are not persisted at all, so no lifecycle tracking is done by the UnitOfWork.

docs/en/reference/annotations-reference.rst
```diff
@@ -1110,6 +1110,67 @@ encouraged to use the :ref:`atomicSet <atomic_set>` or
 will ensure that collections are updated in the same write operation as the
 versioned parent document.
 
+@View
+-----
+
+Required annotation to mark a PHP class as a view. Views are created from
+aggregation pipelines, which is built in a special repository method. Views can
```
greg0ire: Not sure I understand the sentence properly. The pipelines are built in a method?

```suggestion
aggregation pipelines, which are built in a special repository method. Views can
```

docs/en/reference/annotations-reference.rst
```diff
@@ -1110,6 +1110,67 @@ encouraged to use the :ref:`atomicSet <atomic_set>` or
...
+        }
+    }
+
+The ``createViewAggregation`` method can add any aggregation pipeline stage,
+except for the ``$out`` and ``$merge`` stages. The pipeline is created for the
+root class specified in the view mapping.
+
+.. note::
+
+    Views must be created before using. This can be done using the
```
greg0ire: "before they can be used" / "before usage" / "before you can use them"? Dunno why, but this feels awkard to me.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataFactory.php
```diff
@@ -223,7 +223,7 @@ protected function doLoadMetadata($class, $parent, $rootEntityFound, array $nonS
      */
     protected function validateIdentifier(ClassMetadata $class) : void
     {
-        if (! $class->identifier && ! $class->isMappedSuperclass && ! $class->isEmbeddedDocument && ! $class->isQueryResultDocument) {
+        if (! $class->identifier && ! $class->isMappedSuperclass && ! $class->isEmbeddedDocument && ! $class->isQueryResultDocument && ! $class->isView()) {
```
greg0ire: It might be time to consider using linebreaks :P
```suggestion
        if (! $class->identifier &&
            ! $class->isMappedSuperclass &&
            ! $class->isEmbeddedDocument &&
            ! $class->isQueryResultDocument &&
            ! $class->isView()) {
```

lib/Doctrine/ODM/MongoDB/Mapping/MappingException.php
```diff
@@ -245,4 +245,19 @@ public static function discriminatorNotAllowedForGridFS(string $className) : sel
     {
         return new self(sprintf("Class '%s' cannot be discriminated because it is marked as a GridFS file", $className));
     }
+
+    public static function invalidRepositoryClass(string $className, string $repositoryClass, string $expectedRepositoryClass) : self
+    {
+        return new self(sprintf('Invalid repository class "%s" for mapped class "%s". It must be an instance of "%s".', $repositoryClass, $className, $expectedRepositoryClass));
```
greg0ire:
```suggestion
        return new self(sprintf(
            'Invalid repository class "%s" for mapped class "%s". It must be an instance of "%s".',
            $repositoryClass,
            $className,
            $expectedRepositoryClass
        ));
```

lib/Doctrine/ODM/MongoDB/Utility/LifecycleEventManager.php
```diff
@@ -183,4 +184,13 @@ private function cascadePostPersist(ClassMetadata $class, object $document) : vo
             }
         }
     }
+
+    private function dispatchEvent(ClassMetadata $class, string $eventName, ?EventArgs $eventArgs = null)
```
greg0ire:
```suggestion
    private function dispatchEvent(ClassMetadata $class, string $eventName, ?EventArgs $eventArgs = null) : void
```

tests/Doctrine/ODM/MongoDB/Tests/SchemaManagerTest.php
```diff
@@ -391,6 +407,44 @@ public function testCreateDocumentCollectionForFile(array $expectedWriteOptions,
...
+                            '$project' => ['username' => true],
+                        ],
+                    ],
+                ],
+                new ArraySubset($options + $expectedWriteOptions)
+            );
+
+        $rootCollection = $this->documentCollections['CmsUser'];
+        $rootCollection
+            ->expects($this->any())
```
greg0ire: This line can be removed entirely.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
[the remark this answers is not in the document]
beberlei: method calls are still a lote more expensive than public property access. Its not a problem in this constructor here though, not the hotpath.

docs/en/reference/annotations-reference.rst
alcaeus: Ok, I'll reword this a bit. To clarify, the repository method receives an aggregation builder instance and can add its aggregation pipeline that defines the view.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataFactory.php
alcaeus: 😅

docs/en/reference/annotations-reference.rst
alcaeus: Reworded, better now?

docs/en/reference/annotations-reference.rst
alcaeus: Reworded, better now?

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataFactory.php
```diff
@@ -223,7 +223,11 @@ protected function doLoadMetadata($class, $parent, $rootEntityFound, array $nonS
      */
     protected function validateIdentifier(ClassMetadata $class) : void
     {
-        if (! $class->identifier && ! $class->isMappedSuperclass && ! $class->isEmbeddedDocument && ! $class->isQueryResultDocument) {
+        if (! $class->identifier &&
+            ! $class->isMappedSuperclass &&
+            ! $class->isEmbeddedDocument &&
+            ! $class->isQueryResultDocument &&
+            ! $class->isView()) {
```
malarzm: could this be reduced to `! $class->identifier && $this->isEntity($class)`?

lib/Doctrine/ODM/MongoDB/Repository/AbstractRepositoryFactory.php
malarzm: Could you remove it? :)

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -1307,6 +1319,10 @@ public function scheduleForDelete(object $document) : void
             return;
         }
 
+        if ($isView) {
+            return;
```
malarzm: Why is this happening so late? I'd expect removing a view document would be NOP if not a `\LogicException`. Do we want to have a view removed from identity map?

alcaeus: Removing a document needs to remove it from the identity map, which is done a few lines earlier. We just don't want to register it as a deletion.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataFactory.php
alcaeus: Absolutely, that makes this a little easier to digest.

PR #2180, opened 2020-05-01
Update phpunit

malarzm:
As per title. For now I've decided to use https://packagist.org/packages/dms/phpunit-arraysubset-asserts as it's easier than changing test for views :)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,11 +36,12 @@
         "symfony/var-dumper": "^3.4|^4.1|^5.0"
     },
     "require-dev": {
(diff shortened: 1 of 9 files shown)


malarzm:
@alcaeus we need to either drop PHP 7.2 or postpone this update to 2.2

alcaeus:
Dropping 7.2 is out of the question for now. We can change the constraint to `^8.2 || ^9.0`.

malarzm:
I'll see what needs to happen with array subsets. It's not critical for 2.1 so I'll take it off the roadmap

alcaeus:
Taken care of in #2248.

PR #2183, opened 2020-05-17
Favour bool and int over boolean and integer

### LINKED ISSUE #2124: Deprecate integer and boolean types
And suggest to use `int` and `bool` instead. FTR I'm not sure if we should switch those types at runtime as somebody might have overridden built-in type (or is that too far fetched?)

malarzm:
Closes #2124

### DIFF
--- a/UPGRADE-2.1.md
+++ b/UPGRADE-2.1.md
@@ -5,3 +5,6 @@ the `Doctrine\ODM\MongoDB\Id\IdGenerator` interface.
 
 The `Doctrine\ODM\MongoDB\Mapping\ClassMetadata` class has been marked final. The class will no longer be extendable 
 in ODM 3.0.
+
+The `boolean` and `integer` mapping types have been deprecated. Use their shorthand counterparts: `bool` and `int` 
+respectively.
--- a/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
@@ -36,6 +36,7 @@ use function ltrim;
 use function sprintf;
 use function strtolower;
 use function strtoupper;
+use function trigger_error;
 
 /**
  * A <tt>ClassMetadata</tt> instance holds all the object-document mapping metadata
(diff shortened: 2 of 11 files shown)


UPGRADE-2.1.md
```diff
@@ -5,3 +5,5 @@ the `Doctrine\ODM\MongoDB\Id\IdGenerator` interface.
 
 The `Doctrine\ODM\MongoDB\Mapping\ClassMetadata` class has been marked final. The class will no longer be extendable 
 in ODM 3.0.
+
+`boolean` and `integer` types have been deprecated. Use their shorthand counterparts: `bool` and `int` respectively.
```
alcaeus, line 9:
```suggestion
The `boolean` and `integer` mapping types have been deprecated. Use their shorthand counterparts: `bool` and `int` respectively.
```

malarzm:
@alcaeus also while I'm here there's also oddly looking `\Doctrine\ODM\MongoDB\Types\IntIdType` type registered under `int_id`. Do you recall anything special about it that couldn't be achieved with `int`? Shall I also deprecate it?

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -1983,6 +1984,18 @@ public function mapField(array $mapping) : array
         $this->checkDuplicateMapping($mapping);
         $this->typeRequirementsAreMet($mapping);
 
+        $deprecatedTypes = [
+            Type::BOOLEAN => Type::BOOL,
+            Type::INTEGER => Type::INT,
+        ];
+        if (isset($deprecatedTypes[$mapping['type']])) {
+            @trigger_error(sprintf(
+                '"%s" type was deprecated in 2.1 and will be removed in 3.0. Use "%s" instead.',
```
alcaeus, line 1993: Missed this the first time around: please provide the package name in the message:
```suggestion
                '"%s" type was deprecated in doctrine/mongodb-odm 2.1 and will be removed in 3.0. Use "%s" instead.',
```

alcaeus:
I think deprecating it makes sense. The class itself provides no value, so deprecating it in favour of `int` (and using `IntType` in the mean time) makes sense. I decided to dig a bit and found that there's only one place where we actually use `int_id`: https://github.com/doctrine/mongodb-odm/blob/efdde4cb35a4b34426e92acc51dbc870e1aa05ed/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php#L1898

I'm not sure why this uses `int_id`, but it may as well use `int`. That could easily be tested by changing it and seeing how much breaks. If that works, we can deprecate `int_id` entirely, just like we deprecate `integer` and `boolean`. Feel free to leave that to a separate PR since this is good to go.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
malarzm: Added in const as well 👍

PR #2184, opened 2020-05-17
Introduce Versionable interface

### LINKED ISSUE #2140: Extract Versionable interface for types
Following in the footsteps of https://github.com/doctrine/mongodb-odm/pull/2128 which provided `Incrementable` interface for types.

malarzm:
Closes #2140

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Types/Versionable.php
+++ b/lib/Doctrine/ODM/MongoDB/Types/Versionable.php
@@ -1,0 +1,20 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ODM\MongoDB\Types;
+
+/**
+ * Types implementing this interface can be used for version fields.
+ */
+interface Versionable
+{
+    /**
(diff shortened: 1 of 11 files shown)


lib/Doctrine/ODM/MongoDB/Types/Versionable.php
```diff
@@ -0,0 +1,20 @@
...
+
+namespace Doctrine\ODM\MongoDB\Types;
+
+/**
+ * Types implementing this interface can be used for version fields.
+ */
+interface Versionable
+{
+    /**
+     * Calculates next version.
```
alcaeus, line 13: Maybe we should add a hint that `null` will be passed if no current version exists?

malarzm: Good idea

malarzm: I put it into `@param` WDYT?

PR #2185, opened 2020-05-18
Favour int over int_id

malarzm:
Follow up of #2183

### DIFF
--- a/UPGRADE-2.1.md
+++ b/UPGRADE-2.1.md
@@ -6,5 +6,5 @@ the `Doctrine\ODM\MongoDB\Id\IdGenerator` interface.
 The `Doctrine\ODM\MongoDB\Mapping\ClassMetadata` class has been marked final. The class will no longer be extendable 
 in ODM 3.0.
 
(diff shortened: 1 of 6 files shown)


UPGRADE-2.1.md
```diff
@@ -6,5 +6,5 @@ the `Doctrine\ODM\MongoDB\Id\IdGenerator` interface.
 The `Doctrine\ODM\MongoDB\Mapping\ClassMetadata` class has been marked final. The class will no longer be extendable 
 in ODM 3.0.
 
-The `boolean` and `integer` mapping types have been deprecated. Use their shorthand counterparts: `bool` and `int` 
-respectively.
+The `boolean`, `integer`, and `int_id` mapping types have been deprecated. Use their shorthand counterparts: `bool` and 
+`int` respectively.
```
malarzm, line 10: Not sure if this still makes sense?

alcaeus: You could add a second sentence explaining that the `int_id` and `int` types are equivalent.

PR #2186, opened 2020-05-19
Append ODM version info to driver metadata

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

This uses a new option in the upcoming MongoDB driver to provide additional driver metadata. The new feature allows custom drivers (such as ODMs) to add their own metadata to the initial MongoDB handshake. This allows for a more detailed analysis of driver versions being used.

Note that this PR doesn't bump the required driver version on purpose: if the driver doesn't support this option, it doesn't throw an error but ignore the option. Furthermore, as ext-mongodb 1.8 is not released yet, I don't want to tie the ODM release schedule to that of the MongoDB driver.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,6 +30,7 @@
         "doctrine/event-manager": "^1.0",
         "doctrine/instantiator": "^1.1",
         "doctrine/persistence": "^1.3.5",
+        "jean85/pretty-package-versions": "^1.3.0",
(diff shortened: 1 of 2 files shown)


alcaeus:
~Because someone can't fill out a PR template~ I have no idea what you're talking about 😅

PR #2188, opened 2020-05-23
Typed properties support

### LINKED ISSUE #2161: Fix typed property without default value support
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.0.5

#### Summary

We are having same (or similar) issue described in this [issue](https://github.com/doctrine/orm/issues/7854) and [PR](https://github.com/doctrine/orm/pull/7857) on ORM albeit with Mongo ODM although the doctrine/persistence is on 1.3.6.

Aside from embedded documents, we get the `Error: Typed property Document::$field must not be accessed before initialization.` error for scalar types, probably for all fields actually.

malarzm:
Fixes #2161

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,6 +30,7 @@
         "doctrine/event-manager": "^1.0",
         "doctrine/instantiator": "^1.1",
         "doctrine/persistence": "^1.3.5",
+        "doctrine/reflection": "1.2.1",
         "jean85/pretty-package-versions": "^1.3.0",
         "mongodb/mongodb": "^1.2.0",
         "ocramius/proxy-manager": "^2.2",
@@ -56,6 +57,7 @@
             "Doctrine\\ODM\\MongoDB\\Benchmark\\": "benchmark",
             "Doctrine\\ODM\\MongoDB\\Tests\\": "tests/Doctrine/ODM/MongoDB/Tests",
             "Documents\\": "tests/Documents",
+            "Documents74\\": "tests/Documents74",
             "Stubs\\": "tests/Stubs",
             "TestDocuments\\" :"tests/Doctrine/ODM/MongoDB/Tests/Mapping/Driver/fixtures"
         }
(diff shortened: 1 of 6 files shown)


alcaeus:
Haven’t checked the source entirely yet, but given that we’re changing a bunch of internals I’d suggest fixing this in 2.0 instead of 2.1. I’m not sure we should consider this a bug fix for 2.0.

tests/Doctrine/ODM/MongoDB/Tests/Mapping/AnnotationDriverTest.php
```diff
@@ -92,34 +92,6 @@ public function testColumnWithMissingTypeDefaultsToString()
         $this->assertEquals('id', $cm->fieldMappings['id']['type']);
     }
 
-    /**
-     * @group DDC-318
```
malarzm, line 96: I've allowed myself to delete these two tests because:
- the original patch I've found in https://gist.github.com/doctrinebot/3b78c5c5f390ea819bf3 (via https://github.com/doctrine/orm/issues/3961) was meant for thing currently existing in `\Doctrine\Persistence\Mapping\Driver\AnnotationDriver::getAllClassNames` and I believe it should be tested there
- I'm eager to remove other tests with `DDC-318` group for the same reason
- these two tests are doing the same thing now - prolly their true meaning is lost in time anyway
- most importantly these were the only tests that are trying to load each and every document thus failing with PHP 7.4 specific documents in global `Documents` folder. With them gone the build should be green now

But if you believe these tests are still relevant I can move typed documents someplace else :)

malarzm:
@alcaeus the true part of the fix is couple of lines long, this won't be too much hassle to include in 2.0.x IMO

tests/Doctrine/ODM/MongoDB/Tests/Mapping/AnnotationDriverTest.php
malarzm: Eh nevermind, these were just reported first. I'm gonna move the documents someplace else

PR #2189, opened 2020-05-29
Fix prepareQueryOrNewObj with object values

alanpoulain:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | N/A

#### Summary

Following https://github.com/doctrine/mongodb-odm/pull/2104.

In API Platform, since the release of the 2.1 version, the following test was failing: https://github.com/api-platform/core/blob/d83d50228e2686b7a2f76bfbadf2dec3fd83fdcd/tests/Bridge/Doctrine/MongoDbOdm/Filter/DateFilterTest.php#L264-L289

You can see it here: https://github.com/api-platform/core/pull/3571/checks?check_run_id=718581504 and here: https://github.com/api-platform/core/pull/3571/checks?check_run_id=718581453.

It's because https://github.com/doctrine/mongodb-odm/pull/2104 has removed the `array_map` on the `$preparedValue`: `DateTime` is not converted anymore to the DB type if it is the type of a value in an array and if the field is not present in the class (for instance with a lookup).

To fix this, I've changed the `convertToDatabaseValue` to fallback to `Type::convertPHPToDatabaseValue` if the field doesn't have a mapping.

It's also a better behavior than the previous one (before https://github.com/doctrine/mongodb-odm/pull/2104): the previous one didn't work if the value was too deep nested.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -1012,10 +1012,7 @@ final class DocumentPersister
 
             $preparedQueryElements = $this->prepareQueryElement((string) $key, $value, null, true, $isNewObj);
             foreach ($preparedQueryElements as [$preparedKey, $preparedValue]) {
-                $preparedValue = Type::convertPHPToDatabaseValue($preparedValue);
-                if ($this->class->hasField($key)) {
-                    $preparedValue = $this->convertToDatabaseValue($key, $preparedValue);
-                }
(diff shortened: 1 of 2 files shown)


malarzm:
Thanks a lot @alanpoulain for taking time to fix this! 2.1.1 now available :)

alanpoulain:
It was very quick, thank you!

However, I've made some CS changes, I think it will be better like this: https://github.com/doctrine/mongodb-odm/pull/2192.

malarzm:
Yeah, looks a bit better :) But important thing is that bug is gone, CS can wait for next patch release

PR #2202, opened 2020-06-21
Add missing parent::configure() call

theyough:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | yes/no
| Fixed issues | doctrine/DoctrineMongoDBBundle#639

#### Summary

see issue details

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Tools/Console/Command/Schema/ShardCommand.php
+++ b/lib/Doctrine/ODM/MongoDB/Tools/Console/Command/Schema/ShardCommand.php
@@ -18,6 +18,8 @@ class ShardCommand extends AbstractCommand
 {
     protected function configure()
     {
+        parent::configure();
+
         $this
             ->setName('odm:schema:shard')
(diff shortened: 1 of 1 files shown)


malarzm:
@theyough thank you for the PR? Would you mind targeting the 1.3.x branch as we need to fix this but in the lowest supported branch?

theyough:
@malarzm , sorry i was a bit in rush, yesterday.i will do. Thank you

theyough:
@malarzm it appears from what i see, that this bug beginning from branch 2.X

malarzm:
@theyough the call is missing in `1.3.x` as well: https://github.com/doctrine/mongodb-odm/blob/1.3.x/lib/Doctrine/ODM/MongoDB/Tools/Console/Command/Schema/ShardCommand.php#L29

theyough:
Yes but there is no parent::configure() method

malarzm:
Ah sorry, my bad!

PR #2204, opened 2020-06-25
Allow doctrine/persistence 2.x

fruitwasp:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues |

#### Summary

Adds support for doctrine/persistence 2.x


malarzm:
BC break in 2.x is a no go. But this is not a lost cause, IIRC migration plan for doctrine/common was for libraries to require `1.x || 2.x` to allow smooth transition for the ecosystem. Sadly I never had time to fully grasp the idea but there are better informed people :) @greg0ire would you mind chiming in?

greg0ire:
You're correct, `|| ^2.0` can be added on the stable branch, here is how you should probably do that: https://github.com/doctrine/orm/pull/8158

After that, people will have an upgrade path, and you can drop the `^1.0 ||` part (and the weird code artefacts I added in #2150 ), like this: https://github.com/doctrine/orm/pull/8166

fruitwasp:
done; targets the 2.1.x branch now too

malarzm:
Thanks @fruitwasp!

Issue #2199, opened 2020-06-15
Fatal error: Uncaught TypeError: MongoDB\BSON\ObjectId::__construct()


nationdata-it:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes/no
| Version     | 2.1.1

#### Summary
when querying the database it generates error

`file.php
        /** @ODM\Field(type="object_id") */
	private $collection_object_id;

	/** @ODM\Field(type="object_id") */
	private $form_cabinet_id;

	/** @ODM\Field(type="object_id") */
	private $employee_id;
`

`$files = $dm->createQueryBuilder(File::class)
				->field('collection_object_id')->equals(new ObjectId($collection_object_id))
				->field('employee_id')->equals(new ObjectId($session_user))
				->field('form_cabinet_id')->exists(false)
				->getQuery()
				->execute();`

or
`
$criteria = [
				'collection_object_id' => new ObjectId($collection_object_id),
				'employee_id' => new ObjectId($session_user),
				'form_cabinet_id' => ['$exists' => false]
];
$dm->getRepository(self::getClassName())->findBy($criteria, $sort)
`

#### Current behavior

Fatal error: Uncaught TypeError: MongoDB\BSON\ObjectId::__construct() expects parameter 1 to be string, boolean given in /var/www/project/vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/Types/ObjectIdType.php on line 20

#### Expected behavior

should return records from db

Steveb-p:
@nationdata-it please provide a stacktrace.

malarzm:
@nationdata-it at a glance it seems that either `$collection_object_id` or `$session_user` is a bool.

nationdata-it:
yes initial i was thinking that i'm sending wrong values to confirm that i changed the call to something like this.

`
$collection_object_id = new ObjectId('5ecd55047ac60e4c2c1f266e');
		$session_user = new ObjectId('5e6843962aca3953e42e2d99');
		$files = File::getSingleResult([
				'collection_object_id' => $collection_object_id,
				'employee_id' => $session_user,
				'form_cabinet_id' => ['$exists' => false]
		]);
`

nationdata-it:
stack trace @Steveb-p 

[...]

Issue #2194, opened 2020-05-31
exists(false); is transformed to exists('')


Plopix:
This appears between 2.0.6 and 2.1.0

### BC Break Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.1.0

#### Summary

The preparation does not work as expected for `$exists`, the current system look at the type of the property and convert it. But for `$exists` that should not be converted according the the type of the property tested.

**it should stay a boolean.**

#### Previous behavior

`false` was still `false` and not `""`

#### Current behavior

false is converted to `""` in the `StringType` convertToDatabaseValue

#### How to reproduce

Pretty easy:

```php
        $qb = $this->getDocumentManager()->createAggregationBuilder($this->getClassMetadata()->name);
        $qb->match()->field('assessment')->exists(false);
        dump($qb->getPipeline());
```

Observe

```
^ array:1 [
  0 => array:1 [
    "$match" => array:1 [
      "assessment" => array:1 [
        "$exists" => false
      ]
    ]
  ]
]
```
then update to 2.1.0 or 2.1.1 using composer update

```
^ array:1 [
  0 => array:1 [
    "$match" => array:1 [
      "assessment" => array:1 [
        "$exists" => ""
      ]
    ]
  ]
]

```

malarzm:
@Plopix would you mind checking if #2205 fixes your issue?

Plopix:
Reproduced on `2.1.1` and I confirm the fix is working well for me.

malarzm:
Closing as #2205 was merged

Issue #2193, opened 2020-05-29
The Doctrine\ODM\MongoDB\Persisters\DocumentPersister::prepareQueryOrNewObj() method does not support certain expressions such as $currentDate


CedrickOka:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.1.1

#### Summary

When using the expression `Expr::currentDate()` with the mongo query builder we get the following error :

```bash
In DateType.php line 65:

  [InvalidArgumentException]                
  Could not convert "date" to a date value  

Exception trace:
  at /.../vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/Types/DateType.php:65
 Doctrine\ODM\MongoDB\Types\DateType::getDateTime()
at /.../vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/Types/DateType.php:92
 Doctrine\ODM\MongoDB\Types\DateType->convertToDatabaseValue()
at /.../vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php:1071
 Doctrine\ODM\MongoDB\Persisters\DocumentPersister->convertToDatabaseValue()
at /.../vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php:1047
 Doctrine\ODM\MongoDB\Persisters\DocumentPersister->convertToDatabaseValue()
at /.../vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php:1017
 Doctrine\ODM\MongoDB\Persisters\DocumentPersister->prepareQueryOrNewObj()
at /.../vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php:1010
 Doctrine\ODM\MongoDB\Persisters\DocumentPersister->prepareQueryOrNewObj()
at /.../vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/Query/Expr.php:593
 Doctrine\ODM\MongoDB\Query\Expr->getNewObj()
at /.../vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/Query/Builder.php:677
```

#### Current behavior

[...]

PR #2205, opened 2020-06-26
Stop type conversion for certain keys in query

malarzm:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #2193, #2194, #2199

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -1031,6 +1031,10 @@ final class DocumentPersister
     {
         if (is_array($value)) {
             foreach ($value as $k => $v) {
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -1031,6 +1031,10 @@ private function convertToDatabaseValue(string $fieldName, $value)
     {
         if (is_array($value)) {
             foreach ($value as $k => $v) {
+                if ($k === '$exists' || $k === '$type' || $k === '$currentDate') {
```
alcaeus, line 1034: Looking at the query operators in MongoDB, there are a few more that we'll want to take another look at (`$expr`, `$text`, `$mod`, etc.) as they don't require type conversion based on the field type. However, I'm fine taking care of those in a separate PR.

Issue #2195, opened 2020-06-02
Doctrine\ODM\MongoDB\Persisters\CollectionPersister::deleteCollections don't remove all necessary elements


TvitiDev:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 1.3.7

#### Summary

In upgrading version up to 1.3.* in  DocumentPersister::handleCollections() method you made optimisation. Now you update the list collections in method CollectionPersister::updateAll() and method CollectionPersister::deleteCollections() remove all elements as one "$unset". But you lost the some paths for remove.

#### Current behavior and how to reproduce

For example, I have next document:

```
{
  "_id" : ObjectId("5ed602e89bfe8563bd61c302"),
  "property" : [
    {
      "value" : 100,
      "items" : [
        {
          "value" : 1,
          "is_flag" : false
        },
        {
          "value" : 2,
          "is_flag" : false
        },
        {
          "value" : 3,
          "is_flag" : false
        }
      ]
    },
    {
      "value" : 101,
      "items" : [
        {
          "value" : 1,
          "is_flag" : false
        },
        {
          "value" : 2,
          "is_flag" : false
        },
        {
          "value" : 3,
          "is_flag" : false
        }
      ]
    }
  ],
  "property_old" : [
    {
      "value" : 200,
      "items" : [
        {
          "value" : 1,
          "is_flag" : false
        },
        {
          "value" : 2,
          "is_flag" : false
        },
        {
          "value" : 3,
          "is_flag" : false
        }
      ]
    }
  ]
}
```

me need remove:
```
'property.1',
'property.0.items.2',
'property_old.0'
```

I expect, what your code remove this element, but your code remove only 'property.1', because variable $paths (line 368 of file CollectionPersister.php) contains next elements:
before execute method excludeSubPaths():
```
'property',
'property.0.items',
'property_old'
```
after execute method excludeSubPaths():
```
'property'
```
This is not correct. You lost 2 path for remove and will remove only 'property.1'

#### Expected behavior

I see 2 errors:
1) Incorrect list paths before execute method excludeSubPaths(), he should contains:
```
'property.1',
'property.0.items.2',
'property_old.0'
```
but contains:
```
'property',
'property.0.items',
'property_old'
```
[...]

PR #2208, opened 2020-06-30
Fix updating collections with similar names

malarzm:
This PR includes refined test from #2195 and a fix for first scenario. Second is problematic as query tries to remove from `property_old` and `property_old.1` which fails. PR targets master just because it doesn't matter and it's easier for me to try out stuff :) If we reach a conclusion on how to fix this should be retargeted to 1.3.x.


alcaeus:
@malarzm I think we need to re-educate that bot...

PR #2210, opened 2020-07-14
Add Document Manager Decorator like doctrine entity manager decorator fo ORM

babarinde:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | 

#### Summary

Add a decorator for the object/document manager, quite similar to the entity manager decorator
This is quite useful when using libraries like Swoole, allowing us to update/switch the object manager when existing connection has been closed.
Tests are missing and would appreciate if I can be directed on how to start
Thanks

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Decorator/DocumentManagerDecorator.php
+++ b/lib/Doctrine/ODM/MongoDB/Decorator/DocumentManagerDecorator.php
@@ -6,13 +6,13 @@ namespace Doctrine\ODM\MongoDB\Decorator;
 
 use Doctrine\Common\EventManager;
 use Doctrine\ODM\MongoDB\Aggregation\Builder as AggregationBuilder;
+use Doctrine\ODM\MongoDB\DocumentManagerInterface;
 use Doctrine\ODM\MongoDB\Hydrator\HydratorFactory;
 use Doctrine\ODM\MongoDB\Proxy\Resolver\ClassNameResolver;
 use Doctrine\ODM\MongoDB\Query\Builder;
 use Doctrine\ODM\MongoDB\Query\FilterCollection;
 use Doctrine\ODM\MongoDB\SchemaManager;
 use Doctrine\ODM\MongoDB\UnitOfWork;
-use Doctrine\ODM\MongoDB\DocumentManagerInterface;
 use Doctrine\Persistence\ObjectManagerDecorator;
 use MongoDB\Client;
 use MongoDB\Collection;
@@ -35,7 +35,7 @@ abstract class DocumentManagerDecorator extends ObjectManagerDecorator implement
     /**
      * {@inheritdoc}
      */
-    public function createQueryBuilder($documentName = null): Builder
+    public function createQueryBuilder($documentName = null) : Builder
     {
         return $this->wrapped->createQueryBuilder($documentName);
     }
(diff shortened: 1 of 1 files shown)


lib/Doctrine/ODM/MongoDB/Decorator/DocumentManagerDecorator.php
```diff
@@ -0,0 +1,234 @@
...
+     */
+    public function find($documentName, $id, $lockMode = null, $lockVersion = null)
+    {
+        return $this->wrapped->find($documentName, $id, $lockMode, $lockVersion);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function flush()
```
alcaeus, line 110: If you're extending the abstract `ObjectManagerDecorator`, why are you implementing this method when it is already implemented in the base class?

babarinde: > Could you please show a use-case for this? I'm not sure why you'd need to switch the object manager, especially with the given example of closing a connection.

In Swoole for instance The App is boostrapped once, If the Instance of the ODM is injected into services, middleware, etc.directly, it is possible for the DB connection to fail/timeout and since the ODM is shared across services, ODM responsibilities to DB will fail.
The Decorator is instead injected to the services, upon such failure/connection timeout, then it is easier to use an ODM factory into our extension of this decorator receives to switch the `$this->wrapped` of the Decorator
`
use Closure;
use Doctrine\ODM\MongoDB\Decorator\DocumentManagerDecorator as DecoratorDocumentManagerDecorator;

class MyDocumentManagerDecorator extends DecoratorDocumentManagerDecorator
{
    private Closure $createDM;
    /**
     * Class constructor.
     */
    public function __construct(callable $createDM)
    {
        parent::__construct($createDM());
        $this->createDM = Closure::fromCallable($createDM);
    }
    /**
     * @return void
     */
    public function open(): void
    {
        if (!$this->isOpen()) {
            $this->wrapped = ($this->createDM)();
        }
    }
}
`

malarzm:
I believe this is only half-baked solution without introducing an interface for `DocumentManager`

babarinde:
> I believe this is only half-baked solution without introducing an interface for `DocumentManager`

@malarzm Please, how do you mean,
There was an corresponding Decorator implementation of the `Doctrine\Persistence\ObjectManagerDecorator` in the ORM (though that may not follow), this has pretty useful.

malarzm:
ORM's https://github.com/doctrine/orm/blob/master/lib/Doctrine/ORM/Decorator/EntityManagerDecorator.php#L14 is implementing `EntityManagerInterface` effectively allowing people to typehint against the interface and receive either a normal EM or a decorator, without their code caring/knowing. To allow same thing we need to introduce similar interface first to ODM as otherwise people can only typehint against `ObjectManagerInterface` and in some cases it's not sufficient.

babarinde:
I noticed that as well, but was not sure why there was no equivalent DocumentManagerInterface and what the consequence of introducing such interface will be beyond this Decorator.
Having said that, Is there a way forward?

babarinde:
@malarzm created a [PR](https://github.com/doctrine/mongodb-odm/pull/2217) to fix the deficiency

PR #2211, opened 2020-07-22
Introduce aggregation API to replicate query API

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | 

#### Summary

This introduces a new `Aggregation` class, that can be created from an aggregation builder using the `getAggregation` method. The `execute` method on the aggregation builder has been deprecated in favour of the new API.

#### Rationale

The query builder returns a query instance that can be passed around without executing the query, but also disallowing changes to the query. The aggregation builder on the other hand only returns an iterator that may or may not have already executed something on the database. Passing the aggregation builder around opens it up to modification that may not be intended.

### DIFF
--- a/UPGRADE-3.0.md
+++ b/UPGRADE-3.0.md
@@ -1,0 +1,23 @@
+# UPGRADE FROM 2.x to 3.0
+
+## Aggregation
+
+The new `Doctrine\ODM\MongoDB\Aggregation\Builder::getAggregation()` method
+returns an `Doctrine\ODM\MongoDB\Aggregation\Aggregation` instance, comparable
+to the `Query` class.
+
+The `Doctrine\ODM\MongoDB\Aggregation\Builder::execute()` method was removed.
+
+## ID generators
+
+The `Doctrine\ODM\MongoDB\Id\AbstractIdGenerator` class has been removed. Custom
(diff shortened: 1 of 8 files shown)


alcaeus:
@malarzm updated UPGRADE documents and updated documentation. Mind giving it another look please?

UPGRADE-3.0.md
```diff
@@ -0,0 +1,24 @@
+# UPGRADE FROM 2.x to 3.0
+
+## Aggregation
+
+The new `Doctrine\ODM\MongoDB\Aggregation\Builder::getAggregation()` method.
+This will return an `Doctrine\ODM\MongoDB\Aggregation\Aggregation` instance,
+comparable to the `Query` class.
+
+The `Doctrine\ODM\MongoDB\Aggregation\Builder::execute()` method was deprecated
+and will be removed in ODM 3.0.
```
malarzm, line 10: In context of 3.0 this (and other sentences) should be written with the past tense I think :) But I have no problem with doing so in another PR

alcaeus: That was the original intention, but I goofed up. It's fixed now.

malarzm:
Thanks @alcaeus!

Issue #2203, opened 2020-06-22
Embedded documents with targetDocument and discriminatorMap property do not save discriminatorField


Steveb-p:
|    Q        |   A
|------------ | -----
| Version     | 2.1.0

### Support Question

tl;dr
When using `targetDocument` and `discriminatorMap` annotation property for `EmbedMany` (I believe `EmbedOne` as well) embedded documents do not save discriminator property.

I believe it should be mentioned in the docs that those two should not be mixed.  
If possible I would either allow `targetDocument`to coexist with `discriminatorMap` as a "sanity check" when document passed is expected to be a specific class / interface.  
Otherwise, I would throw an exception if they are used together.

Context:
I was extending an existing document to include a new type of embedded document with new properties. I had a document looking like this:
```php
use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;

/**
 * @ODM\Document
 */
class Order 
{
    /**
     * @ODM\EmbedMany(targetDocument=LegacyClass::class)
     */
    private $embedded;
}
```

then I wanted to introduce another possible target document:

```php
use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;

/**
 * @ODM\Document
 */
class Order 
{
    /**
     * @ODM\EmbedMany(
     *     targetDocument=AbstractClass::class),
     *     strategy="atomicSetArray",
     *     discriminatorField="type",
     *     discriminatorMap={
     *         "legacy"=LegacyClass::class,
     *         "statistic_aggregate"=NewClass::class,
     *     },
     *     defaultDiscriminatorValue="legacy",
     */
    private $embedded;
}
```

However, this resulted in `type` property to never be set.

While documents inserted into this embedded collection were `NewClass` instances and have it's property, when trying to read ODM was not finding `type` property and tried to instantiate `LegacyClass`.

Therefore I think that docs should explicitly warn about mixing those two annotations (`targetDocument` & `discriminatorMap`).

malarzm:
> If possible I would either allow targetDocumentto coexist with discriminatorMap as a "sanity check" when document passed is expected to be a specific class / interface.

IIRC we've made discriminator map strict if it's present (i.e. ODM will throw an exception if class is not listed in the map), so a sanity check is already there. My vote goes for an exception given it's not working anyway, a note in the docs will be also a nice to have.

Steveb-p:
@malarzm so basically an additional check should be added around here:

[...]

PR #2213, opened 2020-07-25
Throw exception for targetDocument & discriminatorMap field mapping

Steveb-p:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #2203

#### Summary

Throws an exception when `discriminatorMap` is used alongside with `targetDocument`.

When used together their behavior was to discard object different than `targetDocument` specified, which lead to potential data loss.

I'm not sure about the tests, or placement of functions / checks, so correct me please :smiley_cat:

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
@@ -1935,6 +1935,10 @@ use function trigger_error;
         if (isset($mapping['repositoryMethod']) && ! (empty($mapping['skip']) && empty($mapping['limit']) && empty($mapping['sort']))) {
             throw MappingException::repositoryMethodCanNotBeCombinedWithSkipLimitAndSort($this->name, $mapping['fieldName']);
         }
+        
+        if (isset($mapping['targetDocument']) && isset($mapping['discriminatorMap'])) {
+            throw MappingException::targetDocumentCanNotBeCombinedWithDiscriminatorMap($this->name, $mapping['fieldName']);
+        }
 
         if (isset($mapping['reference']) && $mapping['type'] === 'one') {
             $mapping['association'] = self::REFERENCE_ONE;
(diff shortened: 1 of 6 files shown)


malarzm:
I'm adding this to 1.3.8 milestone as the original issue so stalebot will leave this PR be while we wait for resolution

alcaeus:
Sorry for the delay here, this totally slipped under my radar. Typically, we'd fix this in 2.1 as that's the current bug fix version. The only concern I have about the patch is that it would start throwing exceptions in code that worked "fine" before. "Fine" meaning that it would store data and ignore the discriminator map. Due to this, we may want to add a deprecation warning about this in 2.2 and only throw an exception about this invalid use-case in 3.0. What do you think @Steveb-p @malarzm?

malarzm:
I'm on board 👍

Steveb-p:
@alcaeus the reason I considered this important enough to throw an exception is that it I think it resulted in some data loss for sometime ago.

I don't remember the original issue in the application I've worked on that caused me to notice this, so I'm not really sure now.

Of course, I absolutely understand that due to the fact that this changes the behavior of the lib in a particular case it should wait for a major release. It's just that weighting data loss against BC break lead me to prefer the exception at that point.

alcaeus:
@Steveb-p agree - the prospect of potential data loss is something I considered as well, but I'm not sure that potentially breaking something that may just have worked (including in our own tests) is worth it. We could lock this down a bit in the validation command to point out that the mapping is invalid and only throw a deprecation for the time being. We can then always decide to accept the BC break if we realise that this causes significant problems for people. We'll include one or the other in 2.2 - I just quite haven't made up my mind yet.

alcaeus:
@Steveb-p I've taken your work and changed it to only emit a deprecation notice, see #2271. I'm hesitant to make this an exception right away, so I feel deprecating this at least gives people that are affected some pause to think about what they're doing.

On a different note, I noticed that the tests you added only check for an exception, which was still happening despite removing the `throw` statement you added: this was because the mapping was invalid to begin with due to a missing `fieldName` field.

PR #2217, opened 2020-08-16
Refactor: introduce DocumentManagerInterface

babarinde:
|      Q       |   A
|------------- | -----------
| BC Break     | no

#### Summary

Add DocumentManagerInterface and use across library

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/DocumentManager.php
+++ b/lib/Doctrine/ODM/MongoDB/DocumentManager.php
@@ -214,7 +214,7 @@ class DocumentManager implements DocumentManagerInterface
      * Creates a new Document that operates on the given Mongo connection
      * and uses the given Configuration.
      */
-    public static function create(?Client $client = null, ?Configuration $config = null, ?EventManager $eventManager = null) : DocumentManagerInterface
+    public static function create(?Client $client = null, ?Configuration $config = null, ?EventManager $eventManager = null) : DocumentManager
     {
         return new static($client, $config, $eventManager);
     }
--- a/lib/Doctrine/ODM/MongoDB/DocumentManager.php
+++ b/lib/Doctrine/ODM/MongoDB/DocumentManager.php
@@ -17,7 +17,6 @@ use Doctrine\ODM\MongoDB\Proxy\Resolver\ProxyManagerClassNameResolver;
 use Doctrine\ODM\MongoDB\Query\FilterCollection;
 use Doctrine\ODM\MongoDB\Repository\DocumentRepository;
 use Doctrine\ODM\MongoDB\Repository\RepositoryFactory;
-use Doctrine\Persistence\ObjectManager;
 use Doctrine\Persistence\ObjectRepository;
 use InvalidArgumentException;
 use Jean85\PrettyVersions;
--- a/lib/Doctrine/ODM/MongoDB/DocumentManager.php
+++ b/lib/Doctrine/ODM/MongoDB/DocumentManager.php
@@ -45,7 +45,7 @@ use function sprintf;
  *     $config = new Configuration();
  *     $dm = DocumentManager::create(new Connection(), $config);
  */
-class DocumentManager implements ObjectManager
+class DocumentManager implements DocumentManagerInterface
 {
     public const CLIENT_TYPEMAP = ['root' => 'array', 'document' => 'array'];
 
@@ -215,7 +215,7 @@ class DocumentManager implements ObjectManager
      * Creates a new Document that operates on the given Mongo connection
      * and uses the given Configuration.
      */
-    public static function create(?Client $client = null, ?Configuration $config = null, ?EventManager $eventManager = null) : DocumentManager
+    public static function create(?Client $client = null, ?Configuration $config = null, ?EventManager $eventManager = null) : DocumentManagerInterface
     {
         return new static($client, $config, $eventManager);
     }
--- a/lib/Doctrine/ODM/MongoDB/DocumentManagerInterface.php
+++ b/lib/Doctrine/ODM/MongoDB/DocumentManagerInterface.php
@@ -1,0 +1,213 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ODM\MongoDB;
+
+use Doctrine\Common\EventManager;
+use Doctrine\ODM\MongoDB\Hydrator\HydratorFactory;
+use Doctrine\ODM\MongoDB\Mapping\ClassMetadata;
+use Doctrine\ODM\MongoDB\Mapping\MappingException;
+use Doctrine\ODM\MongoDB\Proxy\Factory\ProxyFactory;
+use Doctrine\ODM\MongoDB\Proxy\Resolver\ClassNameResolver;
+use Doctrine\ODM\MongoDB\Query\FilterCollection;
+use Doctrine\Persistence\ObjectManager;
+use InvalidArgumentException;
+use MongoDB\Client;
+use MongoDB\Collection;
+use MongoDB\Database;
+use MongoDB\GridFS\Bucket;
+use RuntimeException;
+
+/**
+ * The DocumentManager Interface 
+ */
+interface DocumentManagerInterface extends ObjectManager
+{
+    /**
+     * Gets the proxy factory used by the DocumentManager to create document proxies.
+     */
+    public function getProxyFactory() : ProxyFactory;
+
+    /**
+     * Creates a new Document that operates on the given Mongo connection
+     * and uses the given Configuration.
+     */
+    public static function create(?Client $client = null, ?Configuration $config = null, ?EventManager $eventManager = null) : DocumentManagerInterface;
+
+    /**
+     * Gets the EventManager used by the DocumentManager.
+     */
+    public function getEventManager() : EventManager;
+
+    /**
+     * Gets the MongoDB client instance that this DocumentManager wraps.
+     */
+    public function getClient() : Client;
+
+    /**
+     * Gets the UnitOfWork used by the DocumentManager to coordinate operations.
+     */
+    public function getUnitOfWork() : UnitOfWork;
+
+    /**
+     * Gets the Hydrator factory used by the DocumentManager to generate and get hydrators
+     * for each type of document.
+     */
+    public function getHydratorFactory() : HydratorFactory;
+
+    /**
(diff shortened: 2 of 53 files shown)


lib/Doctrine/ODM/MongoDB/DocumentManager.php
```diff
@@ -215,7 +214,7 @@ public function getProxyFactory() : ProxyFactory
      * Creates a new Document that operates on the given Mongo connection
      * and uses the given Configuration.
      */
-    public static function create(?Client $client = null, ?Configuration $config = null, ?EventManager $eventManager = null) : DocumentManager
+    public static function create(?Client $client = null, ?Configuration $config = null, ?EventManager $eventManager = null) : DocumentManagerInterface
```
alcaeus, line 217: This return type should stay the same: for one, changing this is a BC break. More importantly, in this case we can say that `create` will always return an instance of itself, so there's no need for the interface here.

lib/Doctrine/ODM/MongoDB/DocumentManagerInterface.php
```diff
@@ -0,0 +1,211 @@
...
+use MongoDB\Database;
+use MongoDB\GridFS\Bucket;
+use RuntimeException;
+
+interface DocumentManagerInterface extends ObjectManager
+{
+    /**
+     * Gets the proxy factory used by the DocumentManager to create document proxies.
+     */
+    public function getProxyFactory() : ProxyFactory;
```
alcaeus, line 28: The proxy factory is a dependency of the document manager. Any other class that needs it should not fetch it from the manager, but rather receive it via dependency injection itself.

lib/Doctrine/ODM/MongoDB/DocumentManagerInterface.php
```diff
@@ -0,0 +1,211 @@
...
+
+    /**
+     * Creates a new Document that operates on the given Mongo connection
+     * and uses the given Configuration.
+     */
+    public static function create(?Client $client = null, ?Configuration $config = null, ?EventManager $eventManager = null) : DocumentManagerInterface;
```
alcaeus, line 34: I see no advantage in the interface requiring a factory: the advantage of different managers is that they can have different ways to be constructed. For example, a simple factory for a manager would be `create(string $connectionUri, array $options = []) : DocumentManagerInterface;`. Having this method as part of the interface forbids you from creating custom factories in implementations of the contract.

lib/Doctrine/ODM/MongoDB/DocumentManagerInterface.php
```diff
@@ -0,0 +1,211 @@
...
+     * @return Collection[]
+     */
+    public function getDocumentCollections() : array;
+
+    /**
+     * Create a new Query instance for a class.
+     *
+     * @param string[]|string|null $documentName (optional) an array of document names, the document name, or none
+     */
+    public function createQueryBuilder($documentName = null) : Query\Builder;
```
alcaeus, line 103: Making my life easier: all methods above this one should not be part of the interface IMO, as they mostly expose implementation details like dependencies or driver classes (which may not always be the same). Please start a discussion on any of the methods if you want to make a case for including them in the interface.

lib/Doctrine/ODM/MongoDB/DocumentManagerInterface.php
```diff
@@ -0,0 +1,211 @@
...
+     */
+    public function createAggregationBuilder(string $documentName) : Aggregation\Builder;
+
+    /**
+     * Acquire a lock on the given document.
+     *
+     * @throws InvalidArgumentException
+     * @throws LockException
+     */
+    public function lock(object $document, int $lockMode, ?int $lockVersion = null) : void;
```
alcaeus, line 116: For the time being, I'd like to remove locking from the interface. I'm not convinced there are enough use-cases to warrant keeping it.

lib/Doctrine/ODM/MongoDB/DocumentManagerInterface.php
```diff
@@ -0,0 +1,211 @@
...
+     * Closes the DocumentManager. All documents that are currently managed
+     * by this DocumentManager become detached. The DocumentManager may no longer
+     * be used after it is closed.
+     */
+    public function close();
+
+    /**
+     * Gets the Configuration used by the DocumentManager.
+     */
+    public function getConfiguration() : Configuration;
```
alcaeus, line 176: This method should not be part of the interface: any service relying on manager configuration should be created with a configuration instance itself.

lib/Doctrine/ODM/MongoDB/DocumentManagerInterface.php
```diff
@@ -0,0 +1,211 @@
...
+
+    /**
+     * Returns a reference to the supplied document.
+     *
+     * @return mixed The reference for the document in question, according to the desired mapping
+     *
+     * @throws MappingException
+     * @throws RuntimeException
+     */
+    public function createReference(object $document, array $referenceMapping);
```
alcaeus, line 186: This method should not be part of the interface, as it only contains internal functionality.

lib/Doctrine/ODM/MongoDB/Event/LifecycleEventArgs.php
```diff
@@ -19,7 +20,7 @@ public function getDocument() : object
         return $this->getObject();
     }
 
-    public function getDocumentManager() : DocumentManager
+    public function getDocumentManager() : DocumentManagerInterface
```
alcaeus, line 23: This is a BC break since the `DocumentManagerInterface` doesn't contain all `DocumentManager` methods.

Please note that this applies to all event classes.

lib/Doctrine/ODM/MongoDB/Id/IdGenerator.php
```diff
@@ -13,5 +13,5 @@ interface IdGenerator
      *
      * @return mixed
      */
-    public function generate(DocumentManager $dm, object $document);
+    public function generate(DocumentManagerInterface $dm, object $document);
```
alcaeus, line 16: This is a BC break: widening the parameter type in the interface is not allowed, as it makes existing implementations incompatible.

lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionInterface.php
```diff
@@ -19,7 +19,7 @@ interface PersistentCollectionInterface extends Collection
     /**
      * Sets the document manager and unit of work (used during merge operations).
      */
-    public function setDocumentManager(DocumentManager $dm);
+    public function setDocumentManager(DocumentManagerInterface $dm);
```
alcaeus, line 22: This change is a BC break (see explanation above).

lib/Doctrine/ODM/MongoDB/Repository/DocumentRepository.php
```diff
@@ -173,7 +173,7 @@ public function getDocumentName() : string
         return $this->documentName;
     }
 
-    public function getDocumentManager() : DocumentManager
+    public function getDocumentManager() : DocumentManagerInterface
```
alcaeus, line 176: This is a BC break.

lib/Doctrine/ODM/MongoDB/Tools/Console/Command/Schema/AbstractCommand.php
```diff
@@ -53,7 +53,7 @@ protected function getSchemaManager()
     }
 
     /**
-     * @return DocumentManager
+     * @return DocumentManagerInterface
```
alcaeus, line 56: Strictly speaking, this is also a BC break. We may want to consider marking this class as `@internal` to avoid this.

alcaeus (changes requested):
@babarinde I've finally had enough time to take a look at this. I've left comments on the look of the interface itself (as I don't believe all methods should be part of it) and left comments where switching from the `DocumentManager` implementation to the `DocumentManagerInterface` interface constitutes a BC break that needs to be avoided.

@malarzm any opinion on the look of the interface?

lib/Doctrine/ODM/MongoDB/DocumentManagerInterface.php
babarinde: @alcaeus do you think getDocumentDatabase, getDocumentCollection and getDocumentBucket should not be part of the interface?

lib/Doctrine/ODM/MongoDB/DocumentManagerInterface.php
babarinde: i guess unlock should also be removed as well

lib/Doctrine/ODM/MongoDB/DocumentManagerInterface.php
babarinde: @alcaeus , While i agree that some of those methods be removed.
using the interface here https://github.com/doctrine/mongodb-odm/pull/2217/files#diff-ba2f525e9ff863e4a04dde560eaf81a6R69 will now affect https://github.com/doctrine/mongodb-odm/pull/2217/files#diff-ba2f525e9ff863e4a04dde560eaf81a6R72
We may have to revert  most constructors using the interface back to its original DM

lib/Doctrine/ODM/MongoDB/DocumentManagerInterface.php
alcaeus: Yes, I forgot to mention that. Sorry.

malarzm:
First, I am sorry it took me this long to get to this PR. I think we won't get away from typehinting driver's classes in the interface without introducing an abstraction layer. We've been there with doctrine/mongodb and it wasn't that great. We need to accept the fact that we (and folks using ODM) depend directly on MongoDB's driver and library and live with it. Should there be any breaking change in said dependencies will require us to bump major version now and I think it's ok.

Also I think that for the interface to make sense we need to include all public methods of the DocumentManager as otherwise the implementations are not really interchangeable. If we come to an agreement that a method should not be part of an interface it most probably should be moved elsewhere (and eventually removed with 3.0).

Finally, more as a note to us maintainers, it'd be great to add some aliasing to the bundle so people can start depending on the interface in their codebases.

alcaeus:
After further discussion, we've decided to postpone this indefinitely as this will be a whole lot of BC breaks for a doubtful result. Thanks @babarinde for tackling this and we'll pick up where you left off if we decide to introduce an interface further down the line.

PR #2223, opened 2020-09-26
Document requirement on PHP extension

greg0ire:
The extension is required in composer.json, and its installation should
be mentioned because it will not even be possible to install the ODM
without it.


malarzm (approved):

malarzm:
Thanks @greg0ire!

PR #2231, opened 2020-10-18
Update PHPStan

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

The PHPStan version was almost 2 year old. With the new version I had to lower the level to 5 because more than 900 errors appear in level 7.

I didn't know how to solve some of the errors so I ignored them, I'll be happy to solve them and remove them from the ignore list.

PR with the Github Action: https://github.com/franmomu/mongodb-odm/pull/5

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/DocumentManager.php
+++ b/lib/Doctrine/ODM/MongoDB/DocumentManager.php
@@ -296,7 +296,10 @@ class DocumentManager implements ObjectManager
      */
     public function getClassMetadata($className) : ClassMetadata
     {
-        return $this->metadataFactory->getMetadataFor($className);
+        $metadata = $this->metadataFactory->getMetadataFor($className);
+        assert($metadata instanceof ClassMetadata);
+
+        return $metadata;
     }
 
     /**
@@ -340,9 +343,9 @@ class DocumentManager implements ObjectManager
     {
         $className = $this->classNameResolver->getRealClass($className);
 
-        /** @var ClassMetadata $metadata */
         $metadata = $this->metadataFactory->getMetadataFor($className);
         assert($metadata instanceof ClassMetadata);
+
         if ($metadata->isFile) {
             return $this->getDocumentBucket($className)->getFilesCollection();
         }
@@ -376,8 +379,9 @@ class DocumentManager implements ObjectManager
     {
         $className = $this->classNameResolver->getRealClass($className);
 
-        /** @var ClassMetadata $metadata */
         $metadata = $this->metadataFactory->getMetadataFor($className);
+        assert($metadata instanceof ClassMetadata);
+
         if (! $metadata->isFile) {
             throw MongoDBException::documentBucketOnlyAvailableForGridFSFiles($className);
         }
@@ -587,8 +591,8 @@ class DocumentManager implements ObjectManager
      */
     public function getReference(string $documentName, $identifier) : object
     {
-        /** @var ClassMetadata $class */
-        $class    = $this->metadataFactory->getMetadataFor(ltrim($documentName, '\\'));
+        $class = $this->metadataFactory->getMetadataFor(ltrim($documentName, '\\'));
+        assert($class instanceof ClassMetadata);
         $document = $this->unitOfWork->tryGetById($identifier, $class);
 
         // Check identity map first, if its already in there just return it.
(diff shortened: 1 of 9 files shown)


lib/Doctrine/ODM/MongoDB/Query/ReferencePrimer.php
```diff
@@ -232,6 +226,8 @@ private function parseDotSyntaxForPrimer(string $fieldName, ClassMetadata $class
 
             return ['fieldName' => $fieldName, 'class' => $class, 'documents' => $documents, 'mapping' => $mapping];
         }
+
+        return ['fieldName' => $fieldName, 'class' => $class, 'documents' => $documents, 'mapping' => $mapping];
```
franmomu, line 230: I have not knowledge enough about this, I've just added the same line than https://github.com/doctrine/mongodb-odm/blob/6e373600959a6a50296724995b2315c51563071f/lib/Doctrine/ODM/MongoDB/Query/ReferencePrimer.php#L171.

malarzm:
@franmomu would it be possible for you to split this PR into two? First adding PHPStan (version currently used by us) as a GA and a second PR updating PHPStan, levels, and code to fix new faults? I'd definitely would like to merge PHPStan as a GA faster and solve new issues later :)

malarzm:
All righty with #2236 merged we can now move this one forward. Please ping me when the branch is rebased, we'll be able to see results in GA now thanks to you :)

franmomu:
hi @malarzm, can you please take a look whenever you can? I've also saw that PHPStan was not totally happy, I've created https://github.com/doctrine/mongodb-odm/pull/2240 in the meanwhile if you want to have it green.

lib/Doctrine/ODM/MongoDB/DocumentManager.php
```diff
@@ -310,8 +313,7 @@ public function getDocumentDatabase(string $className) : Database
             return $this->documentDatabases[$className];
         }
 
-        $metadata = $this->metadataFactory->getMetadataFor($className);
-        assert($metadata instanceof ClassMetadata);
+        $metadata = $this->getClassMetadata($className);
```
malarzm, line 316: Not using `getClassMetadata` was an intentional micro-optimization to avoid an additional method call. I'm not sure it is still relevant in 2020 but better safe than sorry and I'd prefer keeping it

lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionTrait.php
```diff
@@ -463,15 +463,17 @@ public function count()
...
 
     /**
      * {@inheritdoc}
      */
     public function add($value)
     {
-        return $this->doAdd($value, false);
+        $this->doAdd($value, false);
+
+        return true;
```
malarzm, line 476: it'd be easier to adjust `doAdd`'s return type :) Now it's just nitpicking, feel free to ignore this comment

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -222,14 +222,12 @@ public function executeInserts(array $options = []) : void
             $inserts[] = $data;
         }
 
-        if ($inserts) {
```
malarzm, line 225: Out of curiosity: PHPStan detected this `if` as redundant as its value is basing on `queuedInserts` which is already checked at the beginning?

lib/Doctrine/ODM/MongoDB/Query/ReferencePrimer.php
malarzm: You're right, an exception would be better if this should never be reached. This should read just fine:
```suggestion
        throw new \LogicException('Unable to parse property path for ReferencePrimer. Please report an issue in Doctrine\'s MongoDB ODM.');
```

malarzm (commented):
Left few remarks. Thanks @franmomu for all the work you do for ODM!

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
franmomu: Looks like it! https://phpstan.org/r/07684e38-4e79-4aa7-b6ba-500c27da21c3

malarzm: nice :)

malarzm (approved):
Thanks for the fixes!

lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionTrait.php
franmomu: Totally true!

franmomu:
You've very welcome @malarzm! Happy to help

PR #2234, opened 2020-10-25
Remove unused dependency

greg0ire:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | n/a

#### Summary
It does not seem to be used in the code, nor in the commit that
introduced it, b05a8255124fe7c6ab3be8a53a816771a7b50ab7 . I picked improvement as the type because having an extra dependency will not cause any issue, and removing it might cause some if you use the lib in your application but forget to declare it in your Composer manifest.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,6 @@
         "doctrine/event-manager": "^1.0",
         "doctrine/instantiator": "^1.1",
         "doctrine/persistence": "^1.3.5|^2.0",
-        "doctrine/reflection": "1.2.1",
         "jean85/pretty-package-versions": "^1.3.0",
         "mongodb/mongodb": "^1.2.0",
         "ocramius/proxy-manager": "^2.2",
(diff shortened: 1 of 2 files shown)


greg0ire:
Blocked by https://github.com/doctrine/coding-standard/pull/230

malarzm:
IIRC I was importing the dependency explicitly because of https://github.com/doctrine/reflection/blob/1.2.x/lib/Doctrine/Common/Reflection/TypedNoDefaultReflectionProperty.php. While the class is now in doctrine/persistence it's only in 2.1.x branch while we're still allowing 1.3.x to be installed

greg0ire:
That's what confuses me: `git log --all -S TypedNoDefaultReflectionProperty` returns nothing

malarzm:
That could have been to force `1.2.1` of reflection because for some reason I was getting `1.2.0` so I made an explicit requirement. Can't really say after all the time, if everything will be working OK I'm totally fine with removal.

greg0ire:
Let's merge then.

PR #2237, opened 2020-10-28
Rename Match classes for PHP 8 compatibility

### LINKED ISSUE #2222: Aggregation\Stage\Match class won't work with PHP8
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | no
| RFC         | no
| BC Break    | yes

#### Summary

Just a heads up if no one noticed it before, but [`Match`](https://github.com/doctrine/mongodb-odm/blob/master/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Match.php) will be a reserved word in PHP8 so that won't work with PHP8 anymore 🙂

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | nah
| Fixed issues | Fixes #2222.

#### Summary

This renames the `Match` classes to `MatchStage` as `match` is a reserved word in PHP 8. Note that the signature change should not cause BC issues when using PHP 7. Users wanting to use PHP 8 will have to change their code anyways if they are using the `Match` classes directly.

Note that testing ODM with PHP 8 only works with some composer shenanigans until the MongoDB driver is fully compatible (`ext-mongodb` compiles, but `mongodb/mongodb` needs some more love before we're ready).


lib/Doctrine/ODM/MongoDB/Aggregation/Stage.php
```diff
@@ -227,7 +228,7 @@ public function lookup(string $from) : Stage\Lookup
      *
      * @see http://docs.mongodb.org/manual/reference/operator/aggregation/match/
      */
-    public function match() : Stage\Match
+    public function match() : MatchStage
```
malarzm, line 231: `Stage\MatchStage` would be better to be consistent with other signatures

malarzm (approved):
Looks great beside that one nitpick 👍

malarzm:
Actually one more thing I noticed while looking at my other PR: could you add a note to the UPGRADE file so we won't forget about it later?

SenseException (approved):

alcaeus:
Closing this PR as I'll fold these changes into the PR where PHP 8 support will be added.

Issue #2233, opened 2020-10-23
Collection validator: JSON Schema


alexandre-abrioux:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

Hi,

I don't think it is currently possible to define a JSON schema to validate, on the database side, inserted or updated data. I am talking about the `$jsonSchema` operator of the `validator` expression: https://docs.mongodb.com/manual/core/schema-validation/#json-schema

Do you think that it would be something worth adding?

We could even generate that schema with all the metadata that we already have, with an option on the collection:

```
/**
 * @ODM\Document(
 *      collection="users",
 *      jsonSchemaValidator=true
 * )
 */
```

; or let the user specify the schema manually if needed:

```
/**
 * @ODM\Document(
 *      collection="users",
 *      jsonSchemaValidator=true,
 *      jsonSchema="{
            name: {
               bsonType: "string",
               description: "must be a string and is required"
            }
        }"
 * )
 */
```

I think it could be a nice feature to have. If it is something that you would be interested in I could write a PR, I'm just asking for your opinion about that idea first. Let me know! :)

Thank you!

malarzm:
@alexandre-abrioux this sound like a nice addition :) If you'd need any help feel free to ping me on our Slack's `#mongodb-odm` channel.

> We could even generate that schema with all the metadata that we already have, with an option on the collection

I think for a first step we should go with user-defined `jsonSchema`, generating schema based on metadata can be added later. Also I expect schema-generation to be a can of worms if we get to single collection inheritance or discriminated embedded documents so I'd rather defer this until later date and not get you into bigger troubles :)

alexandre-abrioux:
Thanks for the answer! I agree, let's do it by step. I'll start working on adding an user-defined `jsonSchema` and get back to you as soon as I have a working solution with tests. I'll check Slack if needed :)

PR #2238, opened 2020-10-31
JSON Schema Validation

alexandre-abrioux:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | #2233 

# Summary

This pull request introduces a new `@Validation` class annotation containing three options:
- `jsonSchema`
- `action`
- `level`

; corresponding to the three options of MongoDB collections:
- [`validator.$jsonSchema`](https://docs.mongodb.com/manual/core/schema-validation/#json-schema)
- [`validationAction`](https://docs.mongodb.com/manual/core/schema-validation/#accept-or-reject-invalid-documents)
- [`validationLevel`](https://docs.mongodb.com/manual/core/schema-validation/#existing-documents)

# The JSON Extension

The MongoDB Driver requires an array for the `validator` option of the `createCollection` method and `collMod` command.

We therefore have to decode the user-defined JSON schema, thus the need to add the `ext-json` to the `require-dev` section of the `composer.json`.

The `ext-json` is still not required for users to install the package, it is only added to the suggestion list.

# Adding a new Validation annotation vs. adding new options to the Document annotation

When I started working on this feature I chose to add three new properties to the `@Document` annotation:
- `validationJsonSchema`
- `validationAction`
- `validationLevel`

But I found that it would be convenient for the `validationAction` and `validationLevel` options to be enumerations as it would help the user to realize directly during the mapping process that they made a mistake in filling those.

A wrong value would trigger `doctrine/annotations`'s inner logic and raise an `AnnotationException`.

When used with Symfony for instance users wouldn't have to wait until execution of `odm:schema:update` command to realize they made a mistake, they would know right away on cache clear which happens on every request by default on dev environment.

Using enumerations was only possible by creating a whole new annotation class as the `@Document` annotation has a constructor and therefore its properties' annotations are not parsed, see https://github.com/doctrine/annotations/blob/e2623fd97c136cc7a18b467da8d8331c01de051c/lib/Doctrine/Common/Annotations/DocParser.php#L548

To sum up adding `@Enum` on the annotation properties was only possible by creating a new class, thus the choice of a new `@Validation` annotation. It makes the annotation-driven configuration similar to the XML-driven configuration which also contains enumerations.

# The validationAction / validationLevel default values issue

MongoDB version ≥ 4.4.0 allows to reset the `validationAction` and `validationLevel` options of the collections to their default values by passing empty strings to the respective arguments of the `collMod` command.

However I found that versions 4.2.0 and bellow do not care about empty strings in the `validationAction` /  `validationLevel` arguments of the `collMod` command. The command options will just get ignored and the collection options will not be modified.

In order to reset those options during `odm:schema:update`, in case for instance the user removed the `action` / `level` options of the `@Validation` annotation, we need to have default values defined for those options in Doctrine.

Those default values are documented and are equal the default values defined by MongoDB.

However if the MongoDB team decides to change those default values in the future it could cause confusion for our users, but it's the only solution I found to be able to support MongoDB versions < 4.4.0.

# Documentation

I updated the `annotations-reference` and added a new section on the `validation-of-documents` cookbook, I hope that's OK. Feel free to proofread it and recommend some adjustments as English is not my mother tong.

# TODO
- [x] Annotation
- [x] AnnotationDriver
- [x] XmlDriver
- [x] ClassMetadata
- [x] SchemaManager
- [x] UpdateCommand
- [x] Tests: Annotation
- [x] Tests: AnnotationDriver
- [x] Tests: XmlDriver
- [x] Tests: ClassMetadata
- [x] Tests: SchemaManager
- [x] Tests : Functionnal
- [x] Tests : UpdateCommand
- [x] Documentation
- [x] Clean Commit History

### DIFF
--- a/docs/en/cookbook/validation-of-documents.rst
+++ b/docs/en/cookbook/validation-of-documents.rst
@@ -1,6 +1,9 @@
 Validation of Documents
 =======================
 
+Validation of Documents - Application Side
+------------------------------------------
+
 .. sectionauthor:: Benjamin Eberlei <kontakt@beberlei.de>
 
 Doctrine does not ship with any internal validators, the reason
@@ -128,3 +131,123 @@ the Versionable extension, which requires another type of event
 called "onFlush".
 
 Further readings: :doc:`Lifecycle Events <../reference/events>`
+
+Validation of Documents - Database Side
+---------------------------------------
+
+.. sectionauthor:: Alexandre Abrioux <alexandre-abrioux@users.noreply.github.com>
+
+.. note::
+
+    This feature has been introduced in version 2.3.0
+
+MongoDB ≥ 3.6 offers the capability to validate documents during
+insertions and updates through a schema associated to the collection
+(cf. `MongoDB documentation <https://docs.mongodb.com/manual/core/schema-validation/>`_).
+
+Doctrine MongoDB ODM now provides a way to take advantage of this functionality
+thanks to the new :doc:`@Validation <../reference/annotations-reference#validation>`
+annotation and its properties (also available with XML mapping):
+
+-
+  ``validator`` - The schema that will be used to validate documents.
+  It is a string representing a BSON document under the
+  `Extended JSON specification <https://github.com/mongodb/specifications/blob/master/source/extended-json.rst>`_.
+-
+  ``action`` - The behavior followed by MongoDB to handle documents that
+  violate the validation rules.
+-
+  ``level`` - The threshold used by MongoDB to filter operations that
+  will get validated.
+
+Once defined, those options will be added to the collection after running
+the ``odm:schema:create`` or ``odm:schema:update`` command.
+
+.. configuration-block::
+
+    .. code-block:: php
+
+        <?php
+
+        namespace Documents;
+
+        use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
+        use Doctrine\ODM\MongoDB\Mapping\ClassMetadata;
+
+        /**
+         * @ODM\Document
+         * @ODM\Validation(
+         *     validator=SchemaValidated::VALIDATOR,
+         *     action=ClassMetadata::SCHEMA_VALIDATION_ACTION_WARN,
+         *     level=ClassMetadata::SCHEMA_VALIDATION_LEVEL_MODERATE,
+         * )
+         */
+        class SchemaValidated
+        {
+            public const VALIDATOR = <<<'EOT'
+        {
+            "$jsonSchema": {
+                "required": ["name"],
+                "properties": {
+                    "name": {
+                        "bsonType": "string",
+                        "description": "must be a string and is required"
+                    }
+                }
+            },
+            "$or": [
+                { "phone": { "$type": "string" } },
+                { "email": { "$regex": { "$regularExpression" : { "pattern": "@mongodb\\.com$", "options": "" } } } },
+                { "status": { "$in": [ "Unknown", "Incomplete" ] } }
+            ]
+        }
+        EOT;
+
+            /** @ODM\Id */
+            private $id;
+
+            /** @ODM\Field(type="string") */
+            private $name;
+
+            /** @ODM\Field(type="string") */
+            private $phone;
+
+            /** @ODM\Field(type="string") */
+            private $email;
+
+            /** @ODM\Field(type="string") */
+            private $status;
+        }
+
+    .. code-block:: xml
+
+        <?xml version="1.0" encoding="UTF-8"?>
+        <doctrine-mongo-mapping xmlns="http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping"
+                          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                          xsi:schemaLocation="http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping
+                          http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping.xsd">
+
+            <document name="SchemaValidated">
+                <schema-validation action="warn" level="moderate">
+                    {
+                        "$jsonSchema": {
+                            "required": ["name"],
+                            "properties": {
+                                "name": {
+                                    "bsonType": "string",
+                                    "description": "must be a string and is required"
+                                }
+                            }
+                        },
+                        "$or": [
+                            { "phone": { "$type": "string" } },
+                            { "email": { "$regex": { "$regularExpression" : { "pattern": "@mongodb\\.com$", "options": "" } } } },
+                            { "status": { "$in": [ "Unknown", "Incomplete" ] } }
+                        ]
+                    }
+                </schema-validation>
+            </document>
+        </doctrine-mongo-mapping>
+
+Please refer to the :doc:`@Validation <../reference/annotations-reference#document>` annotation reference
+for more details on how to use this feature.
--- a/docs/en/reference/annotations-reference.rst
+++ b/docs/en/reference/annotations-reference.rst
@@ -139,7 +139,7 @@ and it does not contain the class name of the persisted document, a
 @Document
 ---------
 
-Required annotation to mark a PHP class as a document, whose peristence will be
+Required annotation to mark a PHP class as a document, whose persistence will be
 managed by ODM.
 
 Optional attributes:
@@ -1091,6 +1091,100 @@ Alias of `@Index`_, with the ``unique`` option set by default.
     private $email;
 
 .. _annotations_reference_version:
+
+@Validation
+-----------
+
+This annotation may be used at the class level to specify the validation schema
+for the related collection.
+
+-
+   ``validator`` - Specifies a schema that will be used by
+   MongoDB to validate data inserted or updated in the collection.
+   Please refer to the following
+   `MongoDB documentation (Schema Validation ¶) <https://docs.mongodb.com/manual/core/schema-validation/>`_
+   for more details. The value should be a string representing a BSON document under the
+   `Extended JSON specification <https://github.com/mongodb/specifications/blob/master/source/extended-json.rst>`_.
+   The recommended way to fill up this property is to create a class constant
+   (eg. ``::VALIDATOR``) using the
+   `HEREDOC/NOWDOC syntax <https://www.php.net/manual/en/language.types.string.php#language.types.string.syntax.nowdoc>`_
+   for clarity and to reference it as the annotation value.
+   Please note that if you decide to insert the schema directly in the annotation without
+   using a class constant then double quotes ``"`` have to be escaped by doubling them ``""``.
+   This method also requires that you don't prefix multiline strings by the Docblock asterisk symbol ``*``.
+-
+   ``action`` - Determines how MongoDB handles documents that violate
+   the validation rules. Please refer to the related
+   `MongoDB documentation (Accept or Reject Invalid Documents ¶) <https://docs.mongodb.com/manual/core/schema-validation/#accept-or-reject-invalid-documents>`_
+   for more details. The allowed values are the following:
+
+       - ``error``
+       - ``warn``
+
+   If it is not defined then the default behavior (``error``) will be used.
+   Those values are also declared as constants for convenience:
+
+      - ``\Doctrine\ODM\MongoDB\Mapping\ClassMetadata::SCHEMA_VALIDATION_ACTION_ERROR``
+      - ``\Doctrine\ODM\MongoDB\Mapping\ClassMetadata::SCHEMA_VALIDATION_ACTION_WARN``
+
+   Import the ``ClassMetadata`` namespace to use those constants in your annotation.
+-
+   ``level`` - Determines which operations MongoDB applies the
+   validation rules. Please refer to the related
+   `MongoDB documentation (Existing Documents ¶) <https://docs.mongodb.com/manual/core/schema-validation/#existing-documents>`_
+   for more details. The allowed values are the following:
+
+      - ``off``
+      - ``strict``
+      - ``moderate``
+
+   If it is not defined then the default behavior (``strict``) will be used.
+   Those values are also declared as constants for convenience:
+
+      - ``\Doctrine\ODM\MongoDB\Mapping\ClassMetadata::SCHEMA_VALIDATION_LEVEL_OFF``
+      - ``\Doctrine\ODM\MongoDB\Mapping\ClassMetadata::SCHEMA_VALIDATION_LEVEL_STRICT``
+      - ``\Doctrine\ODM\MongoDB\Mapping\ClassMetadata::SCHEMA_VALIDATION_LEVEL_MODERATE``
+
+   Import the ``ClassMetadata`` namespace to use those constants in your annotation.
+
+.. code-block:: php
+
+    <?php
+
+    use Doctrine\ODM\MongoDB\Mapping\ClassMetadata;
+    // ... other imports
+
+    /**
+     * @Document
+     * @Validation(
+     *     validator=SchemaValidated::VALIDATOR,
+     *     action=ClassMetadata::SCHEMA_VALIDATION_ACTION_WARN,
+     *     level=ClassMetadata::SCHEMA_VALIDATION_LEVEL_MODERATE,
+     * )
+     */
+    class SchemaValidated
+    {
+        public const VALIDATOR = <<<'EOT'
+    {
+        "$jsonSchema": {
+            "required": ["name"],
+            "properties": {
+                "name": {
+                    "bsonType": "string",
+                    "description": "must be a string and is required"
+                }
+            }
+        },
+        "$or": [
+            { "phone": { "$type": "string" } },
+            { "email": { "$regex": { "$regularExpression" : { "pattern": "@mongodb\\.com$", "options": "" } } } },
+            { "status": { "$in": [ "Unknown", "Incomplete" ] } }
+        ]
+    }
+    EOT;
+
+        // rest of the class code...
+    }
 
 @Version
 --------
--- a/lib/Doctrine/ODM/MongoDB/Mapping/Annotations/Validation.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/Annotations/Validation.php
@@ -1,0 +1,34 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ODM\MongoDB\Mapping\Annotations;
+
+/**
+ * @Annotation
+ * @Target({"CLASS"})
+ */
+class Validation
+{
+    /** @var string */
+    public $validator;
+
+    /**
+     * @var string
+     * @Enum({
+     *     \Doctrine\ODM\MongoDB\Mapping\ClassMetadata::SCHEMA_VALIDATION_ACTION_ERROR,
+     *     \Doctrine\ODM\MongoDB\Mapping\ClassMetadata::SCHEMA_VALIDATION_ACTION_WARN,
+     *     })
+     */
+    public $action;
+
+    /**
+     * @var string
+     * @Enum({
+     *     \Doctrine\ODM\MongoDB\Mapping\ClassMetadata::SCHEMA_VALIDATION_LEVEL_OFF,
+     *     \Doctrine\ODM\MongoDB\Mapping\ClassMetadata::SCHEMA_VALIDATION_LEVEL_STRICT,
+     *     \Doctrine\ODM\MongoDB\Mapping\ClassMetadata::SCHEMA_VALIDATION_LEVEL_MODERATE,
+     *     })
+     */
+    public $level;
+}
--- a/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
@@ -125,6 +125,23 @@ use function trigger_deprecation;
     public const REFERENCE_STORE_AS_DB_REF_WITH_DB = 'dbRefWithDb';
     public const REFERENCE_STORE_AS_REF            = 'ref';
 
+    /**
+     * The collection schema validationAction values
+     *
+     * @see https://docs.mongodb.com/manual/core/schema-validation/#accept-or-reject-invalid-documents
+     */
+    public const SCHEMA_VALIDATION_ACTION_ERROR = 'error';
+    public const SCHEMA_VALIDATION_ACTION_WARN  = 'warn';
+
+    /**
+     * The collection schema validationLevel values
+     *
+     * @see https://docs.mongodb.com/manual/core/schema-validation/#existing-documents
+     */
+    public const SCHEMA_VALIDATION_LEVEL_OFF      = 'off';
+    public const SCHEMA_VALIDATION_LEVEL_STRICT   = 'strict';
+    public const SCHEMA_VALIDATION_LEVEL_MODERATE = 'moderate';
+
     /* The inheritance mapping types */
     /**
      * NONE means the class does not participate in an inheritance hierarchy
@@ -270,6 +287,27 @@ use function trigger_deprecation;
      * @var array<string, array>
      */
     public $shardKey = [];
+
+    /**
+     * Allows users to specify a validation schema for the collection.
+     *
+     * @var array|object|null
+     */
+    private $validator;
+
+    /**
+     * Determines whether to error on invalid documents or just warn about the violations but allow invalid documents to be inserted.
+     *
+     * @var string
+     */
+    private $validationAction = self::SCHEMA_VALIDATION_ACTION_ERROR;
+
+    /**
+     * Determines how strictly MongoDB applies the validation rules to existing documents during an update.
+     *
+     * @var string
+     */
+    private $validationLevel = self::SCHEMA_VALIDATION_LEVEL_STRICT;
 
     /**
      * READ-ONLY: The name of the document class.
@@ -991,6 +1029,42 @@ use function trigger_deprecation;
     public function isSharded(): bool
     {
         return $this->shardKey !== [];
+    }
+
+    /**
+     * @return array|object|null
+     */
+    public function getValidator()
+    {
+        return $this->validator;
+    }
+
+    /**
+     * @param array|object|null $validator
+     */
+    public function setValidator($validator): void
+    {
+        $this->validator = $validator;
+    }
+
+    public function getValidationAction(): string
+    {
+        return $this->validationAction;
+    }
+
+    public function setValidationAction(string $validationAction): void
+    {
+        $this->validationAction = $validationAction;
+    }
+
+    public function getValidationLevel(): string
+    {
+        return $this->validationLevel;
+    }
+
+    public function setValidationLevel(string $validationLevel): void
+    {
+        $this->validationLevel = $validationLevel;
     }
 
     /**
@@ -2159,6 +2233,12 @@ use function trigger_deprecation;
 
         if ($this->isReadOnly) {
             $serialized[] = 'isReadOnly';
+        }
+
+        if ($this->validator !== null) {
+            $serialized[] = 'validator';
+            $serialized[] = 'validationAction';
+            $serialized[] = 'validationLevel';
         }
 
         return $serialized;
--- a/lib/Doctrine/ODM/MongoDB/Mapping/Driver/AnnotationDriver.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/Driver/AnnotationDriver.php
@@ -13,6 +13,7 @@ use Doctrine\ODM\MongoDB\Mapping\Annotations\ShardKey;
 use Doctrine\ODM\MongoDB\Mapping\ClassMetadata;
 use Doctrine\ODM\MongoDB\Mapping\MappingException;
 use Doctrine\Persistence\Mapping\Driver\AnnotationDriver as AbstractAnnotationDriver;
+use MongoDB\Driver\Exception\UnexpectedValueException;
 use ReflectionClass;
 use ReflectionMethod;
 
@@ -25,6 +26,8 @@ use function count;
 use function get_class;
 use function interface_exists;
 use function is_array;
+use function MongoDB\BSON\fromJSON;
+use function MongoDB\BSON\toPHP;
 use function trigger_deprecation;
 
 /**
@@ -99,6 +102,25 @@ class AnnotationDriver extends AbstractAnnotationDriver
                 $metadata->setDefaultDiscriminatorValue($annot->value);
             } elseif ($annot instanceof ODM\ReadPreference) {
                 $metadata->setReadPreference($annot->value, $annot->tags ?? []);
+            } elseif ($annot instanceof ODM\Validation) {
+                if (isset($annot->validator)) {
+                    try {
+                        $validatorBson = fromJSON($annot->validator);
+                    } catch (UnexpectedValueException $e) {
+                        throw MappingException::schemaValidationError($e->getCode(), $e->getMessage(), $className, 'validator');
+                    }
+
+                    $validator = toPHP($validatorBson, []);
+                    $metadata->setValidator($validator);
+                }
+
+                if (isset($annot->action)) {
+                    $metadata->setValidationAction($annot->action);
+                }
+
+                if (isset($annot->level)) {
+                    $metadata->setValidationLevel($annot->level);
+                }
             }
         }
 
--- a/lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
@@ -11,6 +11,7 @@ use Doctrine\Persistence\Mapping\Driver\FileDriver;
 use DOMDocument;
 use InvalidArgumentException;
 use LibXMLError;
+use MongoDB\Driver\Exception\UnexpectedValueException;
 use SimpleXMLElement;
 
 use function array_keys;
@@ -29,6 +30,8 @@ use function iterator_to_array;
 use function libxml_clear_errors;
 use function libxml_get_errors;
 use function libxml_use_internal_errors;
+use function MongoDB\BSON\fromJSON;
+use function MongoDB\BSON\toPHP;
 use function next;
 use function preg_match;
 use function simplexml_load_file;
@@ -193,6 +196,28 @@ class XmlDriver extends FileDriver
 
         if (isset($xmlRoot->{'shard-key'})) {
             $this->setShardKey($metadata, $xmlRoot->{'shard-key'}[0]);
+        }
+
+        if (isset($xmlRoot->{'schema-validation'})) {
+            $xmlSchemaValidation = $xmlRoot->{'schema-validation'};
+
+            if (isset($xmlSchemaValidation['action'])) {
+                $metadata->setValidationAction((string) $xmlSchemaValidation['action']);
+            }
+
+            if (isset($xmlSchemaValidation['level'])) {
+                $metadata->setValidationLevel((string) $xmlSchemaValidation['level']);
+            }
+
+            $validatorJson = (string) $xmlSchemaValidation;
+            try {
+                $validatorBson = fromJSON($validatorJson);
+            } catch (UnexpectedValueException $e) {
+                throw MappingException::schemaValidationError($e->getCode(), $e->getMessage(), $className, 'schema-validation');
+            }
+
+            $validator = toPHP($validatorBson, []);
+            $metadata->setValidator($validator);
         }
 
         if (isset($xmlRoot['read-only']) && (string) $xmlRoot['read-only'] === 'true') {
--- a/lib/Doctrine/ODM/MongoDB/SchemaManager.php
+++ b/lib/Doctrine/ODM/MongoDB/SchemaManager.php
@@ -301,6 +301,64 @@ final class SchemaManager
     }
 
     /**
+     * Ensure collection validators are up to date for all mapped document classes.
+     */
+    public function updateValidators(?int $maxTimeMs = null, ?WriteConcern $writeConcern = null): void
+    {
+        foreach ($this->metadataFactory->getAllMetadata() as $class) {
+            assert($class instanceof ClassMetadata);
+            if ($class->isMappedSuperclass || $class->isEmbeddedDocument || $class->isQueryResultDocument || $class->isView() || $class->isFile) {
+                continue;
+            }
+
+            $this->updateDocumentValidator($class->name, $maxTimeMs, $writeConcern);
+        }
+    }
+
+    /**
+     * Ensure collection validators are up to date for the mapped document class.
+     */
+    public function updateDocumentValidator(string $documentName, ?int $maxTimeMs = null, ?WriteConcern $writeConcern = null): void
+    {
+        $class = $this->dm->getClassMetadata($documentName);
+        if ($class->isMappedSuperclass || $class->isEmbeddedDocument || $class->isQueryResultDocument || $class->isView() || $class->isFile) {
+            throw new InvalidArgumentException('Cannot update validators for files, views, mapped super classes, embedded documents or aggregation result documents.');
+        }
+
+        $validator = [];
+        if ($class->getValidator() !== null) {
+            $validator = $class->getValidator();
+        }
+
+        $collection       = $this->dm->getDocumentCollection($class->name);
+        $database         = $this->dm->getDocumentDatabase($class->name);
+        $collections      = $database->listCollections();
+        $collectionExists = false;
+        foreach ($collections as $existingCollection) {
+            if ($collection->getCollectionName() === $existingCollection->getName()) {
+                $collectionExists = true;
+                break;
+            }
+        }
+
+        if (! $collectionExists) {
+            $this->createDocumentCollection($documentName, $maxTimeMs, $writeConcern);
+
+            return;
+        }
+
+        $database->command(
+            [
+                'collMod' => $class->collection,
+                'validator' => $validator,
+                'validationAction' => $class->getValidationAction(),
+                'validationLevel' => $class->getValidationLevel(),
+            ],
+            $this->getWriteOptions($maxTimeMs, $writeConcern)
+        );
+    }
+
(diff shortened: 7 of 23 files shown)


malarzm:
Just wanted to say thanks for a really well-thought PR! It's great you have already included documentation. Looking forward to leaving the "draft" state :)

docs/en/reference/annotations-reference.rst
```diff
@@ -177,6 +197,17 @@ Optional attributes:
...
+                ""required"": [""name""],
+                ""properties"": {
+                    ""name"": {
+                        ""bsonType"": ""string"",
+                        ""description"": ""must be a string and is required""
+                    }
+                }
+            }",
+     *     validationAction=@ValidationAction(ClassMetadata::VALIDATION_ACTION_WARN),
+     *     validationLevel=@ValidationLevel(ClassMetadata::VALIDATION_LEVEL_MODERATE),
```
malarzm: If it's feasible it'd be better to use consts already for `validationAction` and `validationLevel` as we're already (https://github.com/doctrine/mongodb-odm/pull/2235) deprecating annotations inside other annotations to allow PHP 8's native attributes in the next version

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -122,6 +122,23 @@
     public const REFERENCE_STORE_AS_DB_REF_WITH_DB = 'dbRefWithDb';
     public const REFERENCE_STORE_AS_REF            = 'ref';
 
+    /**
+     * The collection schema validationAction values
+     *
+     * @see https://docs.mongodb.com/manual/core/schema-validation/#accept-or-reject-invalid-documents
+     */
+    public const VALIDATION_ACTION_ERROR = 'error';
```
malarzm: Let's prefix the consts with `SCHEMA_` so the name will tell us already it's for the schema validation

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -2112,6 +2186,16 @@ public function __sleep()
             $serialized[] = 'isReadOnly';
         }
 
+        if (is_array($this->validationJsonSchema)) {
+                $serialized[] = 'validationJsonSchema';
+            if ($this->validationAction !== self::VALIDATION_ACTION_ERROR) {
```
malarzm: let's serialize all 3 fields not looking at their default values - this way if we ever change the default we won't need to remember about this place

lib/Doctrine/ODM/MongoDB/Utility/EnvironmentHelper.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ODM\MongoDB\Utility;
+
+use function extension_loaded;
+use function version_compare;
+
+class EnvironmentHelper
```
malarzm: I think we'll be better off using functions natively instead of wrapping them in a class which merely forwards arguments. We're doing that already doing that in number of places

tests/Doctrine/ODM/MongoDB/Tests/Functional/DatabasesTest.php
```diff
@@ -19,6 +19,11 @@ public function testDefaultDatabase()
         $this->assertEquals('test_default', $this->dm->getDocumentDatabase(DefaultDatabaseTest::class)->getDatabaseName());
     }
 
+    public function testDatabaseVersion()
```
malarzm: what's the point of this test?

lib/Doctrine/ODM/MongoDB/Utility/EnvironmentHelper.php
malarzm: If it's added as a class to allow `testMissingJsonExtensionShouldThrowException` test then feel free to drop it. We're already doing our best to let users know they need JSON, we don't need to even test it :)

malarzm:
I let myself go through the review initially, hope you don't mind :)

alexandre-abrioux:
@malarzm Thanks ! I just wanted to increase the coverage before leaving the draft stage, but overall it was almost ready. I don't have much time this week but I will look into your review very soon. Thank you again :)

docs/en/reference/annotations-reference.rst
alexandre-abrioux: Nice, I didn't know about this feature! Would it be OK if we declared those annotations directly on the class? Because by removing them completely we would lose the enumerations and value checking. I'm thinking of something like this:
```php
<?php

        /**
         * Document
         * @ValidationAction(ClassMetadata::VALIDATION_ACTION_WARN)
         * @ValidationLevel(ClassMetadata::VALIDATION_LEVEL_MODERATE)
         * @ValidationJsonSchema("{
                ""required"": [""name""],
                ""properties"": {
                    ""name"": {
                        ""bsonType"": ""string"",
                        ""description"": ""must be a string and is required""
                    }
                }
            }")
         */
        class JsonSchemaValidated
        {
           ...
        }
```

lib/Doctrine/ODM/MongoDB/Utility/EnvironmentHelper.php
alexandre-abrioux: I created this class to allow mocking of built-in functions. Two purposes:

### 1 - Covering missing JSON extension

Like you said, to allow:
- `AnnotationDriverTest::testMissingJsonExtensionShouldThrowException`
- `XmlDriverTest::testMissingJsonExtensionShouldThrowException`

; which covers the raise of a `MappingException` here:

https://github.com/doctrine/mongodb-odm/blob/20ba50f8b3bd4d6c01dbced1fe09290af8da241e/lib/Doctrine/ODM/MongoDB/Mapping/Driver/AnnotationDriver.php#L150-L152

; and here:

https://github.com/doctrine/mongodb-odm/blob/20ba50f8b3bd4d6c01dbced1fe09290af8da241e/lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php#L195-L197

I created this `MappingException::jsonExtensionMissing` because `ext-json` is not required in `composer.json` and we could end up with uncaught exceptions.

I could remove this logic but I suggest we require `ext-json` in the `composer.json` at some point, what do you think?

### 2 - Covering wrong MongoDB version

Most importantly, we need to cover this:

https://github.com/doctrine/mongodb-odm/blob/20ba50f8b3bd4d6c01dbced1fe09290af8da241e/lib/Doctrine/ODM/MongoDB/SchemaManager.php#L336-L345

; because this method will be called each time the schema is updated, regardless of whether or not schema validation is used for the Document class: if no annotation is used on the class, we need to reset the collection `validator`.

The issue is that schema validation is only available for MongoDB version ≥ 3.6.

To cover this I found easy to mock the `version_compare` function, but I think I can actually mock the database version by adding an argument to the `getMockDatabase` method:

https://github.com/doctrine/mongodb-odm/blob/20ba50f8b3bd4d6c01dbced1fe09290af8da241e/tests/Doctrine/ODM/MongoDB/Tests/SchemaManagerTest.php#L1087

tests/Doctrine/ODM/MongoDB/Tests/Functional/DatabasesTest.php
alexandre-abrioux: This test covers the new `SchemaManager::getDocumentDatabaseVersion` method (see below), and ensures that the method returns an expected value on each of our supported MongoDB version.

https://github.com/doctrine/mongodb-odm/blob/20ba50f8b3bd4d6c01dbced1fe09290af8da241e/lib/Doctrine/ODM/MongoDB/SchemaManager.php#L309-L319

lib/Doctrine/ODM/MongoDB/Utility/EnvironmentHelper.php
alexandre-abrioux: Removed the `EnvironmentHelper` class in https://github.com/doctrine/mongodb-odm/pull/2238/commits/33693adb2c3a587028211cd78182a6e6a5c3e0a8

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -268,6 +285,27 @@
      */
     public $shardKey = [];
 
+    /**
+     * READ-ONLY: Allows users to specify a validation schema for the collection.
+     *
+     * @var array|null
+     */
+    public $validationJsonSchema;
```
alcaeus: I've started making these properties private and having accessors for them: the method call overhead here isn't severe enough to justify a public property that we "trust" to be read-only. Same goes for the other new properties.

docs/en/reference/annotations-reference.rst
alexandre-abrioux: I extracted the properties to a new class annotation to still be able to use enumerations in https://github.com/doctrine/mongodb-odm/pull/2238/commits/41ba49a3494e50f182089efc7c10748dd7274d60

alexandre-abrioux:
Hi, thanks again @malarzm for the draft review.
I took your advice into consideration, made the according changes and edited the documentation to reflect those modifications.
Please take a look again at the PR description because it was updated to explain some of the choices I made.
Some more tests were added to increase coverage.
It should now be ready to be reviewed :)
Thank you.

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/Validation.php
```diff
@@ -0,0 +1,34 @@
...
+ */
+class Validation
+{
+    /** @var string */
+    public $jsonSchema;
+
+    /**
+     * @var string
+     * @Enum({
+     *     \Doctrine\ODM\MongoDB\Mapping\ClassMetadata::SCHEMA_VALIDATION_ACTION_ERROR,
```
alcaeus, line 19: Mu suggestion would be to add a `use` statement for `ClassMetadata` to make the enumerations a bit more concise.

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AnnotationDriver.php
```diff
@@ -99,6 +100,20 @@ public function loadMetadataForClass($className, \Doctrine\Persistence\Mapping\C
                 $class->setDefaultDiscriminatorValue($annot->value);
             } elseif ($annot instanceof ODM\ReadPreference) {
                 $class->setReadPreference($annot->value, $annot->tags ?? []);
+            } elseif ($annot instanceof ODM\Validation) {
+                if (isset($annot->jsonSchema)) {
+                    $validationJsonSchema = json_decode($annot->jsonSchema, true);
+                    if ($validationJsonSchema === null) {
+                        throw MappingException::jsonSchemaValidationError($className, 'jsonSchema');
```
alcaeus: This may become obsolete if we don't pass in JSON strings, but in case we keep JSON strings I'd suggest to use `json_last_error` and `json_last_error_msg` to provide more detail about the error that occurred.

lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
```diff
@@ -180,6 +181,20 @@ public function loadMetadataForClass($className, \Doctrine\Persistence\Mapping\C
         if (isset($xmlRoot->{'shard-key'})) {
             $this->setShardKey($class, $xmlRoot->{'shard-key'}[0]);
         }
+        if (isset($xmlRoot->{'validation-json-schema'})) {
+            $validationJsonSchema = (string) $xmlRoot->{'validation-json-schema'};
+            $validationJsonSchema = json_decode($validationJsonSchema, true);
+            if ($validationJsonSchema === null) {
+                throw MappingException::jsonSchemaValidationError($className, 'validation-json-schema');
```
alcaeus: Same here: we should expose error details in the exception.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -295,6 +297,79 @@ public function deleteDocumentIndexes(string $documentName, ?int $maxTimeMs = nu
         $this->dm->getDocumentCollection($documentName)->dropIndexes($this->getWriteOptions($maxTimeMs, $writeConcern));
     }
 
+    /**
+     * Returns the MongoDB Database version for the document class.
+     */
+    public function getDocumentDatabaseVersion(string $documentName, ?int $maxTimeMs = null, ?WriteConcern $writeConcern = null) : string
```
alcaeus: A few issues with this method:
* I don't think this needs to be part of the public API. Consider making it protected.
* The `$writeConcern` option is not supported for reading commands (as `buildInfo` is), so you can drop it.
* `maxTimeMS` is not an option for the command helper (see https://docs.mongodb.com/php-library/current/reference/method/MongoDBDatabase-command/#definition). Furthermore, the `buildInfo` command doesn't need it, so you can also drop that and remote the call to `getWriteOptions` below.

alcaeus (commented):
First of all, thank you for taking the time to work on such a feature, I really appreciate the thought you've put into this.

I've left a few comments below from a first reading I did a little while back, but wanted to also discuss how we can make use of this feature. I see two general ways on how users might want their schema definitions:
* Set manually as implemented in this PR
* Automatically determined based upon mapping

The second is definitely more complex and IMO way outside the scope of the implementation. However, I believe that we should keep it in mind when defining how to map the feature. The current mapping might work with an automatic JSON schema generation, for example by omitting the `jsonSchema` property of the annotation, or by omitting the `validation-json-schema` tag in XML mappings. I don't think there's anything to change, but I wanted to bring it up if only to solicit feedback.

Speaking of annotations, I'm a little bit worried about using a JSON string in annotations. I know the idea is to be able to copy/paste the schema from somewhere else, but this point is moot if people need to quote it. Looking at it, JSON schema is (almost) compatible with the annotations, so that you could do the following:
```php
/**
 * @ODM\Document
 * @ODM\Validation(
 *     jsonSchema={
 *          "required": {"name"},
 *          "properties": {
 *              "name": {
 *                  "bsonType": "string",
 *                  "description": "must be a string and is required"
 *              }
 *          }
 *      },
 *     action=ClassMetadata::SCHEMA_VALIDATION_ACTION_WARN,
 *     level=ClassMetadata::SCHEMA_VALIDATION_LEVEL_MODERATE,
 * )
 */
class JsonSchemaValidated
```

The only change needed to make the JSON schema valid for an annotation is to change square brackets for arrays to curly brackets, as doctrine/annotations (currently) does not support using square brackets for arrays (this was suggested in https://github.com/doctrine/annotations/pull/122 but closed as 2.0 was thought to make this easier but it never materialised).

However, the advantage of pasting the schema in annotations is moot if we look at PHP 8 attributes (which will eventually replace annotations):
```php
#[ODM\Document()]
#[ODM\Validation(
    jsonSchema=[
        "required" => ["name"],
        "properties" => [
            "name" => [
                "bsonType" => "string",
                "description" => "must be a string and is required"
            ]
        ]
    ],
    action=ClassMetadata::SCHEMA_VALIDATION_ACTION_WARN,
    level=ClassMetadata::SCHEMA_VALIDATION_LEVEL_MODERATE,
])
class JsonSchemaValidated
```

In that case, pasting a string may be easier since we can single-quote it (or maybe even use HEREDOC?). Just wanted to check if you've given this some thought to make it a bit more user-friendly.

One concern I did have was that any validation errors may lead to a broken UnitOfWork that could then cause issues. This would be similar to an issue we recently found if we were to implement transaction support, but on second thought this doesn't seem to be an issue any more than it already is: a validation error would be comparable to a duplicate key exception, leading to a situation we already know and are aware of. I'll do another review of this PR and focus on test cases to make sure everything is covered.

alexandre-abrioux:
@alcaeus Thank you very much for that detailed PR review. I will work on your suggestions as soon as I get some time off during the following days.

I agree, we should definitely keep in mind the fact that we could generate that schema from metadata. That's what I was resolved to try and do when I opened #2233. But it was for sure a good suggestion by @malarzm to implement it manually first.

About the annotation's specs, this is a tough choice.

I actually started to implement this feature the "array way" first (without strings). I chose to fallback to the "string way" in the process of implementing it as I found some drawbacks. I'll try to summarize my thought process with a table (more details below):
| spec | advantages | inconveniences |
|-|-|-|
| array | - No need for `json_decode` in `AnnotationDriver`. | - `XMLDriver` cannot be implemented in a similar way: `ext-json` is still needed in the end.<br>- With PHP8 we will need to transform the JSON schema to PHP arrays. It is not as easy as a simple search/replace but there are online tools that can take care of it ([exemple](https://wtools.io/convert-json-to-php-array)).<br>- With PHP7 and below we need to replace square brackets to curly braces for arrays, complexity increases with the one of the JSON schema.<br>- If we extend, in the future, the validation implementation to the whole MongoDB `validator`, not just the `$jsonSchema` part of it, we will need more types than just the JSON valid data types (ex. regex) and thus it won't be possible to implement it with PHP arrays. But this will probably be done in another annotation anyway.<br>  |
| string | - Implementation is similar in `AnnotationDriver` and `XMLDriver`.<br>- With PHP8 Attributes and HEREDOC we will probably not need to escape characters anymore. | - Need for `ext-json`.<br>- With PHP7 and below we need to escape double quotes, but it is as simple as search/replace all double quotes in the JSON string. |

## The "array way"

It would be great to implement it this way. As you said:

> JSON schema is (almost) compatible with the annotations

; and we wouldn't need to `json_decode` the schema to PHP arrays. We could get rid of the dependency between this feature and `ext-json`.

However I found that this dependency was still needed for the `XMLDriver` as it would be terrible to ask users to write their JSON schema directly as an XML tree. Plus we would need to use some kind of `<xs:any>` hack in the `doctrine-mongo-mapping.xsd` which I didn't feel comfortable with.

In the end the `XMLDriver` implementation and the `AnnotationDriver` implementation would have needed to be different which I found kind of confusing for the user.

## The "string way"

It has a major flaw: we need to escape double quotes. However it is as simple as a search/replace so it is acceptable in my opinion.

Plus, as you mentioned, we probably won't need to escape in PHP8 anymore, which is great. I haven't got to test the new PHP8 Attributes still tho so I'm not 100% sure that they accept HEREDOC, but I guess they should.

## $jsonSchema vs. whole validator

Furthermore, but this is probably far-fetched, I was trying at that time to interface the whole `validator` component of MongoDB with doctrine, not just the `$jsonSchema` part of it. Citing MongoDB docs:

> In addition to JSON Schema validation that uses the $jsonSchema query operator, MongoDB supports validation with other query operators...

For instance:

```
{
    $jsonSchema: { ... },
    $or: [
         { phone: { $type: "string" } },
         { email: { $regex: /@mongodb\.com$/ } },
         { status: { $in: [ "Unknown", "Incomplete" ] } }
    ]
}
```

This could be done in doctrine if we decide to extend the feature it in the future. However those validators are not JSON typed. As I understand it they could be composed of any Javascript type (see the `$regex` in this example).

The validation would have to be passed as a string in the annotation (no other choice here). This is when I fell back to the "string way".

But the validation content would also need to be parsed with a Javascript object parser of some kind... That's when I understood it wasn't an easy task and I should focus on `$jsonSchema` first. But I didn't revert my "string way" choice since then.

Anyways, that's how I decided to use strings in the end. Sorry for the long explanation, just wanted to expose my whole thought process.

But it could still be reverted, you tell me :)

## Validation errors

I actually didn't think too much about it as I thought the exception raised by the mongodb driver was enough, like a unique index exception. Feel free to guide me if there is some test I could add to make sure that this is enough.

Thank you again for taking the time to review this PR!

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/Validation.php
alexandre-abrioux: This does not work for annotations located on `@Annotation` classes' properties unfortunately.

`doctrine/annotations` does not parse `use` statements in its `DocParser` for `@Annotation` classes:
https://github.com/doctrine/annotations/blob/88fb6fb1dae011de24dd6b632811c1ff5c2928f5/lib/Doctrine/Common/Annotations/DocParser.php#L486

The mechanism that allows this feature for user-land annotations is located here:
https://github.com/doctrine/annotations/blob/71abee25eb5e83e3c95565136710bb874eef1986/lib/Doctrine/Common/Annotations/AnnotationReader.php#L141

But the `getClassImports` method is not used in the `DocParser::collectAnnotationMetadata` for some reason.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
alexandre-abrioux: Thanks!

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/Validation.php
alcaeus: Ah, I missed that this was an annotations annotation - those are indeed special (a refactoring of those was planned for 2.0 but we don't have a compatibility layer for this in 1.x).

alexandre-abrioux:
branch rebased on master

alexandre-abrioux:
rebased (fixed conflicts)

alcaeus:
@alexandre-abrioux sorry for the delay here. I'll revisit this PR in the coming days.

alexandre-abrioux:
@alcaeus Thanks! Just finished rebasing to upstream, I got an error on the `sharded_cluster` pipeline and I'm having a hard time figuring out what could be the cause. I understand it's a new pipeline you added this week in https://github.com/doctrine/mongodb-odm/pull/2272. Could you check it out? I'm sure you can guide me into the right direction to fix it! Thanks again :)

alcaeus:
> I got an error on the `sharded_cluster` pipeline and I'm having a hard time figuring out what could be the cause.

Taking a short look I wasn't able to reproduce this locally, but I'll take another look when I return to review this.

alexandre-abrioux:
@alcaeus rebased the branch off of 2.3.x, and edited the doc:

> This feature has been introduced in version 2.2.0

to

> This feature has been introduced in version 2.3.0

Tests are green again :)

docs/en/cookbook/validation-of-documents.rst
```diff
@@ -127,4 +130,100 @@ instances. This was already discussed in the previous blog post on
...
+
+.. note::
+
+    This feature has been introduced in version 2.3.0
+
+MongoDB ≥ 3.6 offers the capability to validate documents during
+insertions and updates through a JSON schema associated to the collection
+(cf. `MongoDB documentation <https://docs.mongodb.com/manual/core/schema-validation/#json-schema>`_).
+
+Doctrine now provides a way to take advantage of this functionality
```
alcaeus:
```suggestion
Doctrine MongoDB ODM now provides a way to take advantage of this functionality
```

docs/en/cookbook/validation-of-documents.rst
```diff
@@ -127,4 +130,100 @@ instances. This was already discussed in the previous blog post on
...
+thanks to the new :doc:`@Validation <../reference/annotations-reference#validation>`
+annotation and its properties (also available with XML mapping):
+
+-
+  ``jsonSchema`` - The JSON Schema that will be used against documents
+  for validation.
```
alcaeus:
```suggestion
  ``jsonSchema`` - The JSON Schema that will be used to validate documents.
```

docs/en/cookbook/validation-of-documents.rst
```diff
@@ -127,4 +130,100 @@ instances. This was already discussed in the previous blog post on
...
+-
+  ``action`` - The behavior followed by MongoDB to handle documents that
+  violate the validation rules.
+-
+  ``level`` - The threshold used by MongoDB to filter operations that
+  will get validated.
+
+Once defined those options will be added to the collection after running
```
alcaeus:
```suggestion
Once defined, those options will be added to the collection after running
```

docs/en/cookbook/validation-of-documents.rst
```diff
@@ -127,4 +130,100 @@ instances. This was already discussed in the previous blog post on
...
+         * @ODM\Validation(
+         *     jsonSchema="{
+                    ""required"": [""name""],
+                    ""properties"": {
+                        ""name"": {
+                            ""bsonType"": ""string"",
+                            ""description"": ""must be a string and is required""
+                        }
+                    }
+                }",
```
alcaeus: Side note: this still has me worried, as people will be inclined to add `*` characters here and it's very unusual to have such longs strings, let alone strings that look like annotation code here.

Thinking about this, I believe it may make more sense to change this from a JSON string to a parsed representation. This would allow people to define the schema as a PHP array in a constant, then reference this in the annotation. Consider this:
```php
/** 
  * @ODM\Document
  * @ODM\Validation(jsonSchema=SomeDocument::VALIDATION_SCHEMA)
  */
class SomeDocument
{
    public const VALIDATION_SCHEMA = [
        'required' => ['name'],
        'properties' => [
            'name' => [...],
        ],
    ];
}
```

This would also allow people to paste JSON directly into the annotation if the library gets the ability to read arrays with square brackets.

docs/en/reference/annotations-reference.rst
```diff
@@ -1115,6 +1115,75 @@ encouraged to use the :ref:`atomicSet <atomic_set>` or
...
+   the validation rules. Please refer to the following
+   `MongoDB documentation <https://docs.mongodb.com/manual/core/schema-validation/#accept-or-reject-invalid-documents>`_
+   for more details. The allowed values are the following:
+      - ``error``
+      - ``warn``
+   If it is not defined then the default behavior (``error``) will be used.
+   Those values are also declared as constants for convenience:
+      - ``\Doctrine\ODM\MongoDB\Mapping\ClassMetadata::SCHEMA_VALIDATION_ACTION_ERROR``
+      - ``\Doctrine\ODM\MongoDB\Mapping\ClassMetadata::SCHEMA_VALIDATION_ACTION_WARN``
+   Just import the ``ClassMetadata`` namespace to use those constants in your annotation.
```
alcaeus:
```suggestion
   Import the ``ClassMetadata`` namespace to use those constants in your annotation.
```

docs/en/reference/annotations-reference.rst
```diff
@@ -1115,6 +1115,75 @@ encouraged to use the :ref:`atomicSet <atomic_set>` or
...
+   for more details. The allowed values are the following:
+      - ``off``
+      - ``strict``
+      - ``moderate``
+   If it is not defined then the default behavior (``strict``) will be used.
+   Those values are also declared as constants for convenience:
+      - ``\Doctrine\ODM\MongoDB\Mapping\ClassMetadata::SCHEMA_VALIDATION_LEVEL_OFF``
+      - ``\Doctrine\ODM\MongoDB\Mapping\ClassMetadata::SCHEMA_VALIDATION_LEVEL_STRICT``
+      - ``\Doctrine\ODM\MongoDB\Mapping\ClassMetadata::SCHEMA_VALIDATION_LEVEL_MODERATE``
+   Just import the ``ClassMetadata`` namespace to use those constants in your annotation.
```
alcaeus:
```suggestion
   Import the ``ClassMetadata`` namespace to use those constants in your annotation.
```

docs/en/cookbook/validation-of-documents.rst
```diff
@@ -127,4 +130,100 @@ instances. This was already discussed in the previous blog post on
...
+                        "required": ["name"],
+                        "properties": {
+                            "name": {
+                                "bsonType": "string",
+                                "description": "must be a string and is required"
+                            }
+                        }
+                    }
+                </validation-json-schema>
+            </document>
```
alcaeus: I missed this before, but on second reading this seems a little more intuitive as it captures all validation-related features into a single XML element:

```suggestion
            <document name="TestDocuments\JsonSchemaValidatedDocument">
                <json-schema-validation action="warn" level="moderate">
                    {
                        "required": ["name"],
                        "properties": {
                            "name": {
                                "bsonType": "string",
                                "description": "must be a string and is required"
                            }
                        }
                    }
                </json-schema-validation>
            </document>
```

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -300,6 +302,82 @@ public function deleteDocumentIndexes(string $documentName, ?int $maxTimeMs = nu
...
+    /**
+     * Returns the MongoDB Database version for the document class.
+     */
+    protected function getDocumentDatabaseVersion(string $documentName): string
+    {
+        $class = $this->dm->getClassMetadata($documentName);
+
+        return $this->dm->getDocumentDatabase($class->name)
+            ->command(['buildInfo' => 1])
+            ->toArray()[0]['version'];
```
alcaeus: A couple of notes on the version check in general:
* `buildInfo` should not be used as it won't reliably work to determine server version in the future (not all servers will support `buildInfo`
* The server version of one server in a cluster should not be indicative of all servers in the cluster.

With this in mind, I'd suggest dropping the version check entirely and letting the server fail if it does not support validation.

alcaeus (commented):
First of all, I'm really sorry for the time it has taken me to review this. Second of all, thank you so much for sticking with this and not giving up on me. Third of all, thank you for taking the time to add tests for the update command which was previously untested.

From a functional standpoint this PR is good to go, but I still have some reservations about the mapping. I'm not convinced that the annotation mapping is the best we can do, and I also think that we can make some minor improvements to the XML mapping to make it a little more contained. I'd appreciate your opinion on the suggestions, even if it seems like we're running in circles.

One more note, while looking through the changes I realised that this PR requires people to incorporate the schema for embedded documents into the documents that embed them. While this is not entirely optimal, I believe this is a sane choice for this feature, as it exposes low-level collection options. As a follow-up feature, we could think about a generator that is able to generate validation rules from other metadata (e.g. constraints from symfony/validator) and import validation rules for embedded documents.

alexandre-abrioux:
Hi @alcaeus,

Thank you for taking the time to review this, and for the very thorough and detailed response.

I am going to implement the changes, but before I start digging, I have a suggestion regarding your point about the annotations.
You're right: it doesn't make sense to place those long schemas inside them.

But I think we overlooked something. After re-reading all your comments, I thought later on: why not do it like this?

```php
/**
 * @ODM\Document
 * @ODM\Validation(
 *     jsonSchema=JsonSchemaValidated::JSON_SCHEMA,
 *     action=ClassMetadata::SCHEMA_VALIDATION_ACTION_WARN,
 *     level=ClassMetadata::SCHEMA_VALIDATION_LEVEL_MODERATE,
 * )
 */
class JsonSchemaValidated
{

    public const JSON_SCHEMA = <<<'EOT'
{
    "required": ["name"],
    "properties": {
        "name": {
            "bsonType": "string",
            "description": "must be a string and is required"
        }
    }
}
EOT;

    /** @ODM\Id */
    private $id;

    private $name;
}
```

I'm really convinced it's the best of both worlds. We would get the same experience with both the XML driver and the Annotation driver: just copy-paste your schema, and you're done!

I just quickly tested it in the test suite and that works. What do you think about this?

alcaeus:
> You're right: it doesn't make sense to place those long schemas inside them.
> 
> But I think we overlooked something. After re-reading all your comments, I thought later on: why not do it like this?
> [...]
> 
> I'm really convinced it's the best of both worlds. We would get the same experience with both the XML driver and the Annotation driver: just copy-paste your schema, and you're done!
> 
> I just quickly tested it in the test suite and that works. What do you think about this?

I like it - this could also be modified to include both worlds: if the schema given to the annotation is a string, assume it's JSON and try to decode it. Afterwards, we're always expecting a valid schema. This would allow people to either copy-paste the schema from an external source without any modification at all using HEREDOC, or if they want to create the schema by writing up a PHP array they can do that as well. Feel free to implement only one of these methods, I'd be happy to expanding support in a later PR.

tests/Doctrine/ODM/MongoDB/Tests/Functional/DatabasesTest.php
alexandre-abrioux: Test removed following instructions in https://github.com/doctrine/mongodb-odm/pull/2238#discussion_r582615832

lib/Doctrine/ODM/MongoDB/SchemaManager.php
alexandre-abrioux: Done in https://github.com/doctrine/mongodb-odm/pull/2238/commits/46b6dce71d8f835a45ea91e4a1193fbb44e1c423

docs/en/cookbook/validation-of-documents.rst
alexandre-abrioux: Done in https://github.com/doctrine/mongodb-odm/pull/2238/commits/46b6dce71d8f835a45ea91e4a1193fbb44e1c423

docs/en/cookbook/validation-of-documents.rst
alexandre-abrioux: Used class constants with Heredoc as proposed in https://github.com/doctrine/mongodb-odm/pull/2238#issuecomment-790123937 (done in https://github.com/doctrine/mongodb-odm/pull/2238/commits/46b6dce71d8f835a45ea91e4a1193fbb44e1c423)

alexandre-abrioux:
Hi @alcaeus 

I'm very sorry for the time it took me to take action after our last conversation, I got caught in other projects.

I finally implemented the requested changes:

- added your edits to the documentation ;
- modified the XML way of providing validation constraints, like so:

```xml
    <document name="TestDocuments\JsonSchemaValidatedDocument">
        <json-schema-validation action="warn" level="moderate">
            {
                "required": ["name"],
                "properties": {
                    "name": {
                        "bsonType": "string",
                        "description": "must be a string and is required"
                    }
                }
            }
        </json-schema-validation>
    </document>
```

- removed the test of the database version before updating the collection ;
- used HEREDOC in class constants for the JSON schema, like so:

```php
/**
 * @ODM\Document
 * @ODM\Validation(
 *     jsonSchema=JsonSchemaValidated::JSON_SCHEMA,
 *     action=ClassMetadata::SCHEMA_VALIDATION_ACTION_WARN,
 *     level=ClassMetadata::SCHEMA_VALIDATION_LEVEL_MODERATE,
 * )
 */
class JsonSchemaValidated
{

    public const JSON_SCHEMA = <<<'EOT'
{
    "required": ["name"],
    "properties": {
        "name": {
            "bsonType": "string",
            "description": "must be a string and is required"
        }
    }
}
EOT;

    /** @ODM\Id */
    private $id;

    private $name;
}
```
- and rebased the PR to the latest 2.3.x branch commit. 

I only implemented the JSON schema provided by the the user as a HEREDOC string. As you stated before, we can expand the feature to a schema provided as a PHP array in a later PR.

Feel free to review the code one last time if needed! Thank you again for the time spent on this :)

alcaeus (approved):
Awesome work @alexandre-abrioux, and thank you for your patience with me while I struggled to find time to review this. I'll release this in the coming days. I really appreciate the time you took to build this great feature!

jmikola:
@alcaeus pinged me to take a look at this the other day and I just got through reading the PR from top to bottom.

The first thing that stood out to me was that this was focused only on `$jsonSchema` and should probably more generically support defining the `validator` document in its entirety. `$jsonSchema` is just a single query operator and from a technical standpoint I don't see any reason the feature can't support a full query (e.g. mixing `$jsonSchema` with `$or`) out of the gate.

I see this was previously alluded to in https://github.com/doctrine/mongodb-odm/pull/2238#issuecomment-735960199:

> Furthermore, but this is probably far-fetched, I was trying at that time to interface the whole validator component of MongoDB with doctrine, not just the $jsonSchema part of it. Citing MongoDB docs:
>
>> In addition to JSON Schema validation that uses the $jsonSchema query operator, MongoDB supports validation with other query operators...
>
> For instance:
>
> ```
> {
>    $jsonSchema: { ... },
>    $or: [
>         { phone: { $type: "string" } },
>         { email: { $regex: /@mongodb\.com$/ } },
>         { status: { $in: [ "Unknown", "Incomplete" ] } }
>    ]
> }
> ```
> This could be done in doctrine if we decide to extend the feature it in the future. However those validators are not JSON typed. As I understand it they could be composed of any Javascript type (see the $regex operator in this example).

The Javascript types you're seeing there (e.g. regex) are really just due to documentation using the MongoDB shell for its example. If this page showed examples in multiple languages/drivers (as some others do), we might expect to see native language types used (or classes like [`MongoDB\BSON\Regex`](https://www.php.net/manual/en/class.mongodb-bson-regex.php).

As a general rule, I think it makes sense to use [Extended JSON](https://github.com/mongodb/specifications/blob/master/source/extended-json.rst) whenever we're asking users to provide JSON that needs to be converted to MongoDB data. That syntax is both valid JSON _and_ capable of expressing all special BSON types.

With respect to this PR, I think it'd make sense to rename "json-schema-validation" to "schema-validation" (consistent with the [MongoDB feature name](https://docs.mongodb.com/manual/core/schema-validation/)) and then solicit the following options:

 * `validator` as a document
 * `validationAction` (or just `action`) as a string
 * `validationLevel` (or just `level`) as a string

This would require anyone using a JSON schema to provide the top-level `$jsonSchema` operator for `validator` on their own, but I think that's a small ask (easily communicated in documentation) for having this feature support _all_ aspects of MongoDB's schema validation.

alcaeus:
Ah, that's a good point. @alexandre-abrioux I can make those changes if you want so you don't have to do another pass. Using extended JSON for this makes sense, as it saves us from re-declaring a good amount of syntax in the XSD (like I did for partial index filters).

alexandre-abrioux:
@jmikola Awesome! That is exactly what I had in mind when I first though about this feature, but I had no knowledge of the extended JSON specification, so couldn't wrap my head around how to solve this easily!

@alcaeus I'll take a shot at it right away! I'd love contribute and finish this feature if I can.

Just to be clear, this is what I'm headed to, using [MongoDB's documentation example](https://docs.mongodb.com/manual/core/schema-validation/#other-query-expressions) and using the extended JSON specification for the regex part. Please confirm me that's what you had in mind (also consider the annotation name and properties) and I'll make the changes.

For the XML part:
```xml
    <document name="TestDocuments\SchemaValidatedDocument">
        <schema-validation action="warn" level="moderate">
          {
            "$jsonSchema": { ... }
            "$or": [
              { "phone": { "$type": "string" } },
              { "email": { "$regex": { "$regularExpression" : { "pattern": "@mongodb\.com$", "options": "" } } } },
              { "status": { "$in": [ "Unknown", "Incomplete" ] } }
            ]
          }
        </schema-validation>
    </document>
```

For the annotation part:

```php
/**
 * @ODM\Document
 * @ODM\Validation(
 *     validator=SchemaValidated::VALIDATOR,
 *     action=ClassMetadata::SCHEMA_VALIDATION_ACTION_WARN,
 *     level=ClassMetadata::SCHEMA_VALIDATION_LEVEL_MODERATE,
 * )
 */
class SchemaValidated
{

    public const VALIDATOR = <<<'EOT'
{
  "$jsonSchema": { ... }
  "$or": [
    { "phone": { "$type": "string" } },
    { "email": { "$regex": { "$regularExpression" : { "pattern": "@mongodb\\.com$", "options": "" } } } },
    { "status": { "$in": [ "Unknown", "Incomplete" ] } }
  ]
}
EOT;

   // rest of the class code...
}
```

alcaeus:
Yep, that looks lovely!

@jmikola if I'm not mistaken we should use `BSON\toPHP(BSON\fromJson($validator))` ([see docs])(https://www.php.net/manual/en/ref.bson.functions.php) to avoid reading plain old JSON, right? BSON types are safe to serialise in Metadata, and this would give the driver the correct representation for whatever the user gives us. We need to make sure to link the extended JSON documentation in the docs here as well.

alexandre-abrioux:
Implemented the changes, looks good to me! Notes:

- The schema is parsed with `BSON\fromJSON` in the driver, so that the user could directly realize _during the mapping process_ if they made a mistake creating their class' schema.
- It is therefore a simple string that's being saved to the `ClassMetaData::validator` property: a serialized BSON document represented as a binary string. 
- The schema is then converted to its PHP representation with `BSON\toPhp` in the `SchemaManager`.

Don't hesitate to make edits the documentation if needed. Thanks!

jmikola:
> if I'm not mistaken we should use BSON\toPHP(BSON\fromJson($validator)) ([see docs])(https://www.php.net/manual/en/ref.bson.functions.php) to avoid reading plain old JSON, right?

Correct. `fromJSON` is only going to yield a binary string, so we need `toPHP` to decode the binary string back to a PHP value representing a document (e.g. `stdClass`).

> It is therefore a simple string that's being saved to the ClassMetaData::validator property: a serialized BSON document represented as a binary string.

I think it may make more sense to immediately chain `fromJson()` with `toPHP()` and only ever store the PHP value in ClassMetadata. In the event the user ever dumps ClassMetadata, the binary string may ultimately be more confusing. Also, IIRC there is mechanism for defining mappings purely with PHP (as opposed to annotations or XML), which entails explicitly constructing and configuring ClassMetadata objects. Sticking to familiar PHP types would help with that (admittedly uncommon) use case.

alexandre-abrioux:
> I think it may make more sense to immediately chain fromJson() with toPHP() and only ever store the PHP value in ClassMetadata. In the event the user ever dumps ClassMetadata, the binary string may ultimately be more confusing. Also, IIRC there is mechanism for defining mappings purely with PHP (as opposed to annotations or XML), which entails explicitly constructing and configuring ClassMetadata objects. Sticking to familiar PHP types would help with that (admittedly uncommon) use case.

@jmikola Thanks, it makes sense! I was afraid the PHP value wouldn't get serialized properly in `ClassMetadata:__sleep` method, but I tested it and it works perfectly.

I made the appropriate edits. However the static analysis test seems to fail now, but on part of the code I didn't modify. Would it be related to the recent upgrade of `doctrine/coding-standard`?

alcaeus:
> the static analysis test seems to fail now, but on part of the code I didn't modify

I'll take a look and fix this before merging. Again, thanks for your work here!

alcaeus (approved):

jmikola:
> I was afraid the PHP value wouldn't get serialized properly in ClassMetadata:__sleep method, but I tested it and it works perfectly.

Serialization support for the driver's BSON classes was implemented in 1.2.0 ([PHPC-460](https://jira.mongodb.org/browse/PHPC-460)), so there should be nothing to worry about there. :+1:

PR #2239, opened 2020-11-03
Assert array value is an instance of the key 

malarzm:
This PR cherry-picks part of #2234 to make PHPStan happy in 2.1.x branch.


malarzm:
I hope the PHPStan's error will go away as we upgrade it.

franmomu:
> I hope the PHPStan's error will go away as we upgrade it.

Looks like it's because the error formatter, I'll make a PR.

PR #2240, opened 2020-11-04
Remove phpstan error formatter

franmomu:
The phpstan version used (0.10.x)  does not work with the error
formatter, this just removes it and it will be added in another
commit upgrading PHPStan.

|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

I used _improvement_ as I'm not sure about the type.


malarzm:
> I used improvement as I'm not sure about the type.

No worries, that table just helps us during triage, so it's not that important in case of continued work on something :)

malarzm:
Thanks @franmomu!

PR #2243, opened 2020-11-09
Use NamedArgumentConstructorAnnotation for annotations

malarzm:
Brings https://github.com/doctrine/annotations/pull/369 to the ODM, here's what can be done later: https://github.com/doctrine/orm/pull/8266

|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | kind of but not really (see UPGRADE notes)
| Fixed issues | n/a


UPGRADE-2.2.md
```diff
@@ -9,9 +9,20 @@ to the `Query` class.
 The `Doctrine\ODM\MongoDB\Aggregation\Builder::execute()` method was deprecated
 and will be removed in ODM 3.0.
 
-## Document indexes (annotations)
+## Aannotations
+
+### Base class
+
+All ODM's annotations no longer extend the `Doctrine\Common\Annotations\Annotation`
+and may require passing arguments to their constructors.
```
malarzm, line 17: missing `class`

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/AbstractField.php
```diff
@@ -4,28 +4,44 @@
...
 
-abstract class AbstractField extends Annotation
+abstract class AbstractField implements NamedArgumentConstructorAnnotation
 {
     /** @var string */
     public $name;
 
-    /** @var string */
-    public $type = 'string';
+    /** @var string|null */
```
malarzm, line 15: `|null` comes from `@Id`

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/AlsoLoad.php
```diff
@@ -4,15 +4,24 @@
...
  * @Annotation
  */
-final class AlsoLoad extends Annotation
+final class AlsoLoad implements NamedArgumentConstructorAnnotation
 {
-    /** @var string */
+    /** @var string|string[] */
+    public $value;
+
+    /** @var string|null */
```
malarzm, line 19: I tried to fix all the wrong docblocks by the way

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/ChangeTrackingPolicy.php
```diff
@@ -4,13 +4,20 @@
...
  */
-final class ChangeTrackingPolicy extends Annotation
+final class ChangeTrackingPolicy implements NamedArgumentConstructorAnnotation
 {
+    /** @var string */
+    public $value;
+
+    public function __construct(string $value)
+    {
+        $this->value = $value;
```
malarzm, line 21: Idea for later PR (or commit): add validation to ctors

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/Document.php
```diff
@@ -4,17 +4,19 @@
...
  * @Annotation
  */
-final class Document extends AbstractDocument
+final class Document extends AbstractDocument implements NamedArgumentConstructorAnnotation
 {
     /** @var string|null */
     public $db;
 
-    /** @var string|null */
+    /** @var array|string|null */
```
malarzm, line 19: `array` is used for capped collection's config

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/File/ChunkSize.php
```diff
@@ -11,12 +11,14 @@
...
-    public $name = 'chunkSize';
-
-    /** @var string */
-    public $type = 'int';
-
-    /** @var bool */
-    public $notSaved = true;
+    public function __construct(
+        string $name = 'chunkSize',
+        string $type = 'int',
```
malarzm, line 16: Does it even make sense to allow overriding anything but `$name` here?

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/ReferenceMany.php
```diff
@@ -35,7 +32,7 @@ final class ReferenceMany extends AbstractField
     /** @var string|null */
     public $defaultDiscriminatorValue;
 
-    /** @var string[]|null */
+    /** @var string|string[]|null */
```
malarzm, line 35: turns out using string is also fine, it's converted to an array later while mapping field

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/File/ChunkSize.php
alcaeus: No: only the name can be configured, the rest is specified in the [GridFS Specification](https://github.com/mongodb/specifications/blob/master/source/gridfs/gridfs-spec.rst).

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/Inheritance.php
```diff
@@ -1,27 +0,0 @@
...
-namespace Doctrine\ODM\MongoDB\Mapping\Annotations;
-
-use Doctrine\Common\Annotations\Annotation;
-
-/**
- * Specifies inheritance mapping for a document
- *
- * @Annotation
- */
-final class Inheritance extends Annotation
```
alcaeus, line 14: Given that this class is `final` and never checked for, it should be safe to remove it. I'm sure there's someone out there using it, not realising that it didn't do anything.

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/MappedSuperclass.php
```diff
@@ -4,14 +4,21 @@
...
 /**
  * Specifies a parent class that other documents may extend to inherit mapping
  * information
  *
  * @Annotation
  */
-final class MappedSuperclass extends AbstractDocument
+final class MappedSuperclass extends AbstractDocument implements NamedArgumentConstructorAnnotation
 {
     /** @var string */
```
alcaeus, line 17:
```suggestion
    /** @var string|null */
```

alcaeus (commented):
As a general theme, I'd advocate to rename the `$value` arguments in constructors to something more meaningful. For example, in the `AlsoLoad` annotation, `$value` doesn't really tell me what this should be, especially with `$name` being the other field. This is especially true with these names being somewhat relevant to BC when people use named arguments.

malarzm:
Getting back to the PR got me so long somebody opened https://github.com/doctrine/annotations/issues/396 which actually is great :)

IonBazan:
@malarzm Since #2344 has been merged and contains similar changes, should we close this one or would you like to backport the `@NamedArgumentConstructorAnnotation` changes to `2.2.x`?

malarzm:
Nah, let's close this one 👍

PR #2244, opened 2020-11-11
Fix trigger_error level when deprecated type is used

rieschl:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | _none_

`trigger_error` should emit a E_USER_DEPRECATED error when a deprecated mapping type is used instead of the default error level.


alcaeus (approved):

malarzm:
Thanks @rieschl!

PR #2250, opened 2020-11-26
Move CI to GitHub Actions

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

This PR moves a number of CI tests to GitHub Actions, reducing the number of builds run on travis-ci. Since the recent plan changes, travis-ci has been slow and unreliable, and with 10000 credits available for the entire Doctrine organisation, builds have been blocked.

In this PR:
* The performance tests and tests against standalone are moved from travis-ci to GitHub Actions
* Tests with coverage have been removed as none of us really bothered to check it. We can bring it back if we find we want to take a look, but until then it's better to reduce the build load
* Actions run on pull requests to `master` and stable branches and all pushes. This is to ensure that we test branches, releases, and it gives contributors the opportunity to have actions run in their forks when they push a branch, allowing them to fix errors before creating a pull request. Note that this mirrors our setup from travis-ci.

This PR will be followed by a separate PR to run tests on sharded clusters and replica sets. These require additional work as they currently use mongo-orchestration to start clusters. I'll see how to best migrate these to GitHub Actions.


.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,79 @@
...
+
+      - name: "Cache dependencies installed with composer"
+        uses: "actions/cache@v2"
+        with:
+          path: "~/.composer/cache"
+          key: "php-${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock') }}"
+          restore-keys: "php-${{ matrix.php-version }}-composer-locked-"
+
+      - name: "Install dependencies with composer"
+        run: "composer update --no-interaction --no-progress --no-suggest"
```
greg0ire, line 71:
```suggestion
        run: "composer update --no-interaction --no-progress"
```

`--no-suggest` is deprecated: https://github.com/doctrine/mongodb-odm/pull/2250/checks#step:5:4

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,79 @@
...
+        if: "${{ matrix.deps == 'normal' }}"
+
+      - name: "Install lowest possible dependencies with composer"
+        run: "composer update --no-interaction --no-progress --no-suggest --prefer-dist --prefer-lowest"
```
greg0ire, line 75:
```suggestion
        run: "composer update --no-interaction --no-progress --prefer-dist --prefer-lowest"
```

.github/workflows/performance.yml
```diff
@@ -0,0 +1,49 @@
...
+
+      - name: "Cache dependencies installed with Composer"
+        uses: "actions/cache@v2"
+        with:
+          path: "~/.composer/cache"
+          key: "php-${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock') }}"
+          restore-keys: "php-${{ matrix.php-version }}-composer-locked-"
+
+      - name: "Install dependencies with Composer"
+        run: "composer install --no-interaction --no-progress --no-suggest"
```
greg0ire, line 45:
```suggestion
        run: "composer install --no-interaction --no-progress"
```

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,79 @@
...
+        uses: "actions/checkout@v2"
+        with:
+          fetch-depth: 2
+
+      - name: "Install PHP"
+        uses: "shivammathur/setup-php@v2"
+        with:
+          php-version: "${{ matrix.php-version }}"
+          tools: "pecl"
+          extensions: "mongodb-${{ matrix.driver-version }}, bcmath"
```
greg0ire, line 59: I have yet to try it out myself, but it seems this might be the reason why that step is taking so long, and a solution would to that would be to use https://github.com/shivammathur/cache-extensions

alcaeus: I'll give that a shot. Compiling the extension takes a while.

alcaeus:
@greg0ire Added extension caching, we'll see what that does for performance.

greg0ire:
Ok, the cache is warm now, so I restarted the build to see how performance will improve.

greg0ire:
I noticed that performance tests were much faster.

Sadly, there are segmentation faults here: https://github.com/doctrine/mongodb-odm/runs/1462657688?check_suite_focus=true#step:6:21

alcaeus:
There is an issue with `pecl` where `pecl install ...` segfaults at the end of the install process due to replacing the .so file while it's in use (by the CLI process that is started by `pecl install`): https://github.com/mongodb/mongo-php-driver/issues/1177.

I'll take another look and add another build step that prints information about the installed extension version (`php --ri mongodb`) so we can confirm that the correct version was installed and is available.

greg0ire:
I reported the issue at https://github.com/shivammathur/setup-php/issues/348

alcaeus:
Added, and it's showing the correct versions for the time being. I'll expand the build pipeline over time and keep an eye on this.

greg0ire:
I just restarted the jobs, the bugfix seems to work great, and the speedup is significant.

Screenshot_2020-11-27 doctrine mongodb-odm

Cc @ostrolucky , the setup on DoctrineBundle could be improved, but it would be a harder because as long as we support PHP 7.1, we need 2 PHP installs (one with XDebug, one with PCOV).

greg0ire (approved):

Issue #2257, opened 2020-12-03
Compound Index not recognized for deletion if order changed


cgrabenstein:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.1.2

#### Summary

We are optimizing indexes in our project and discovered, that switching the order of keys of a compound index does not trigger a deletion of the old index with the previous order. 

#### Current behavior

Say we have a class `User` with an index as following:

```
/**
 * @Document(collection="User")
 *
 * @UniqueIndex(keys={"firstname"="asc", "creation_date"="desc", "lastname"="asc"})
 */
class User {
...
}
```

Running `bin/console doctrine:mongodb:schema:update` will create an index `firstname_1_creation_date_-1_lastname_1`.

When we change the Index to
```
@UniqueIndex(keys={"firstname"="asc", "lastname"="asc", "creation_date"="desc"})
```
and run the same command again, the previously created index will not be deleted. Instead we have a second index `firstname_1_lastname_1_creation_date_-1` now.

#### Expected behavior

I would expect that the old index is being detected as no longer in use and deleted by the update command. 

The reason for this behaviour is that the Method `isEquivalentIndexKeys` in the `SchemaManager` does not consider the order of keys. It uses a weak equality check (`==`) between the key arrays which ignores order.

malarzm:
This sounds like a bug indeed. Would you like to try fixing this @cgrabenstein? Re using `==` for the check we tend to use strict checks wherever possible so there had to be good reason to use a weak one this time. If possible, I'd add an additional check for key order instead of changing the equality check.

cgrabenstein:
I'll gladly give it a shot. However I need some help concerning compound text indexes, especially if text index are mixed with non-text indexes. Consider the following cases:

Should `@Index({'foo': 'text', 'bar': 'text'})` be considered equivalent to `@Index({'bar': 'text', 'foo': 'text'})`? [`foo` and `bar` switched]
Should `@Index({'foo': 'text', 'a': 1, 'b': 1})` be considered equivalent to `@Index({'foo': 'text', 'b': 1, 'a': 1})`? [`a` and `b` switched]
Should `@Index({'foo': 'text', 'a': 1, 'b': 1})` be considered equivalent to `@Index({'a': 1, 'foo': 'text', 'b': 1})`? [`a` and `foo` switched]

If you can help me and tell me, which of the cases are considered equivalent and which are not, I will work on a fix for the bug!

[...]

PR #2260, opened 2020-12-07
Added missing check for order of compound keys

cgrabenstein:
Signed-off-by: Christoph Grabenstein <christoph.grabenstein@sensiolabs.de>

|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #2257

#### Summary

The fix consist of two main changes:
* added a strict check of the `array_key`s of the index-key-arrays.
* added a method that checks if the text indexes are at the same position.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/SchemaManager.php
+++ b/lib/Doctrine/ODM/MongoDB/SchemaManager.php
@@ -14,9 +14,13 @@ use MongoDB\Driver\WriteConcern;
 use MongoDB\Model\IndexInfo;
 use function array_diff_key;
 use function array_filter;
+use function array_keys;
 use function array_merge;
+use function array_search;
 use function array_unique;
+use function array_values;
 use function assert;
+use function in_array;
 use function is_string;
 use function iterator_count;
 use function iterator_to_array;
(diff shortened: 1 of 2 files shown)


alcaeus (approved):
LGTM. I've rebased your PR to make this a bug fix for 2.1. Thanks for contributing this fix! 🎉

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -460,11 +464,35 @@ private function isEquivalentIndexKeys(IndexInfo $mongoIndex, array $documentInd
...
         // phpcs:disable SlevomatCodingStandard.Operators.DisallowEqualOperators.DisallowedEqualOperator
-        return $mongoIndexKeys == $documentIndexKeys;
+        return $this->hasTextIndexesAtSamePosition($mongoIndex, $documentIndex) &&
+            array_keys($mongoIndexKeys) === array_keys($documentIndexKeys) &&
+            $mongoIndexKeys == $documentIndexKeys;
+    }
+
+    private function hasTextIndexesAtSamePosition(IndexInfo $mongoIndex, array $documentIndex) : bool
+    {
+        $mongoIndexKeys    = $mongoIndex['key'];
```
malarzm, line 479: Personally I'd prefer `$mongoIndex->getKey()` but as we're dealing with array structures here anyway that's a nitpick

malarzm:
Thanks a a lot @cgrabenstein for providing the fix!

lib/Doctrine/ODM/MongoDB/SchemaManager.php
alcaeus: Ah, I missed this. While we do implement `ArrayAccess` for `IndexInfo`, I believe it's more sensible to rely on the public API of the class over accessing data as array elements. However, since I'm not sure where we're still doing this we may want to do a separate PR to fix this.

PR #2261, opened 2020-12-20
Handle void return type in custom collection's methods

### LINKED ISSUE #2256: Void methods in collections causing fatal error
### Bug Report

I have a collection that extends Doctrine\Common\Collections\ArrayCollection which is persisted to database. This collection has methods that returns void, for example:

```
public function removeProduct(ProductId $productId): void
    {
        $index = $this->indexOf($productId);

        if ($index !== false) {
            $this->remove($index);
        }
    }
```
The generated persistent collection is trying to return value from void method:

```
/**
     * {@inheritDoc}
     */
    public function removeProduct(ProductId $productId): void
    {
        $this->initialize();
        if ($this->needsSchedulingForSynchronization()) {
            $this->changed();
        }
        return $this->coll->removeProduct($productId);
    }
```

I updated symfony recently, and during that process doctrine/common was also updated. I think that with doctrine/common in version 2.12 it was working, but I'm not 100% sure if I had outdated cache. I can verify that if needed.

|    Package   |   Version
|------------ | ------
| doctrine/mongodb-odm-bundle | 4.2.2
| doctrine/common | 3.0.2

#### How to reproduce

1. Create collection that is persisted to database
2. Add void method
3. Try to store collection

#### Expected behavior

Void methods are supported by proxy classes

malarzm:
Fixes #2256

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/PersistentCollection/DefaultPersistentCollectionGenerator.php
+++ b/lib/Doctrine/ODM/MongoDB/PersistentCollection/DefaultPersistentCollectionGenerator.php
@@ -190,6 +190,11 @@ CODE;
         $parametersString = $this->buildParametersString($method);
         $callParamsString = implode(', ', $this->getParameterNamesForDecoratedCall($method->getParameters()));
 
+        $return = 'return ';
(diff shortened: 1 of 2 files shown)


malarzm:
Eh, this was supposed to go into 2.1.x branch. I'll be cherry-picking this manually.

PR #2268, opened 2021-01-08
Run psalm with php 8

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

Psalm is failing because if there is no php version specified, it uses the lower bound of composer, so if a library uses PHP 7.4+ syntax, it fails.

Ref: https://github.com/laminas/laminas-code/issues/67


malarzm:
Dang I thought that psalm would infer version from the runtime, not from composer. At least that's what it looked like to me lately when running analysis on 8.0 instead of 7.4 fixed fails. Thanks @franmomu 👍

PR #2271, opened 2021-01-18
Deprecate mapping relations with targetDocument and discriminatorMap

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #2203

#### Summary

Based on #2213, this deprecates this invalid mapping. The discriminatorMap will be ignored entirely, but the mapping still works (but maybe not as people would expect). Despite that, I'm hesitant to make this an exception, so we're deprecating this for removal in 3.0.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
@@ -1946,6 +1946,16 @@ use function trigger_deprecation;
             throw MappingException::repositoryMethodCanNotBeCombinedWithSkipLimitAndSort($this->name, $mapping['fieldName']);
         }
 
+        if (isset($mapping['targetDocument']) && isset($mapping['discriminatorMap'])) {
+            trigger_deprecation(
+                'doctrine/mongodb-odm',
+                '2.2',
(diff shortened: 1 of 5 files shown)


lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -1946,6 +1946,16 @@ public function mapField(array $mapping): array
             throw MappingException::repositoryMethodCanNotBeCombinedWithSkipLimitAndSort($this->name, $mapping['fieldName']);
         }
 
+        if (isset($mapping['targetDocument']) && isset($mapping['discriminatorMap'])) {
+            trigger_deprecation(
+                'doctrine/mongodb-odm',
+                '2.2',
+                'Mapping both "targetDocument" and "discriminatorMap" on field "%s" in class "%s" is deprecated.',
```
malarzm, line 1953:
```suggestion
                'Mapping both "targetDocument" and "discriminatorMap" on field "%s" in class "%s" is deprecated. Only one of them can be used at a time',
```

I'd suggest to use one of them so the deprecation message is actionable

malarzm (approved):

PR #2272, opened 2021-01-18
Run tests on sharded clusters

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

This part of the pipeline was dropped when we moved from travis-ci to GH actions. This now uses the (definitely unsupported) mongodb-labs/drivers-evergreen-tools repository to launch MongoDB clusters of any version/topology.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -11,14 +11,14 @@ env:
   fail-fast: true
 
 jobs:
-  phpunit-standalone:
-    name: "PHPUnit on standalone servers"
+  phpunit:
+    name: "PHPUnit"
     runs-on: "${{ matrix.os }}"
 
     strategy:
       matrix:
         os:
-          - "ubuntu-20.04"
+          - "ubuntu-18.04"
         php-version:
           - "7.2"
           - "7.3"
@@ -28,23 +28,25 @@ jobs:
           - "4.4"
           - "4.2"
           - "4.0"
-          - "3.6"
         driver-version:
           - "stable"
+        topology:
+          - "server"
         deps:
           - "normal"
         include:
           - deps: "low"
-            os: "ubuntu-20.04"
+            os: "ubuntu-16.04"
             php-version: "7.2"
             mongodb-version: "3.6"
             driver-version: "1.5.0"
-
-    services:
-      mongodb:
-        image: "mongo:${{ matrix.mongodb-version }}"
-        ports:
-          - "27017:27017"
+            topology: "server"
+          - topology: "sharded_cluster"
+            os: "ubuntu-18.04"
+            php-version: "8.0"
+            mongodb-version: "4.4"
+            driver-version: "stable"
+            deps: "normal"
 
     steps:
       - name: "Checkout"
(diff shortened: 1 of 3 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -11,14 +11,14 @@ env:
...
-    name: "PHPUnit on standalone servers"
+  phpunit:
+    name: "PHPUnit"
     runs-on: "${{ matrix.os }}"
 
     strategy:
       matrix:
         os:
-          - "ubuntu-20.04"
+          - "ubuntu-18.04"
```
alcaeus, line 21: Note: going back to 18.04 as we drivers-evergreen-tools only supports MongoDB 4.4 on 20.04. 18.04 supports MongoDB 4.0, 4.2, and 4.4.

tests/Doctrine/ODM/MongoDB/Tests/Functional/ReadPreferenceTest.php
```diff
@@ -46,6 +46,8 @@ public function testHintIsNotSetByDefault()
      */
     public function testHintIsSetOnQuery($readPreference, array $tags = [])
     {
+        $this->skipTestIfSharded(User::class);
```
alcaeus, line 49: The sharded cluster uses a single-member replica set, so a functional test that runs a `find` with a secondary read preference always fails.

.github/workflows/continuous-integration.yml
```diff
@@ -28,23 +28,25 @@ jobs:
...
         driver-version:
           - "stable"
+        topology:
+          - "server"
         deps:
           - "normal"
         include:
           - deps: "low"
-            os: "ubuntu-20.04"
+            os: "ubuntu-16.04"
```
malarzm, line 39: Hopefully this image won't disappear before MongoDB ends support for 3.6 in April :)

tests/Doctrine/ODM/MongoDB/Tests/BaseTest.php
```diff
@@ -144,6 +144,20 @@ protected function skipTestIfNotSharded($className)
         $this->markTestSkipped('Test skipped because server does not support sharding');
     }
 
+    protected function skipTestIfSharded($className)
+    {
+        $result = $this->dm->getDocumentDatabase($className)->command(['listCommands' => true])->toArray()[0];
+        if (! $result['ok']) {
+            $this->markTestSkipped('Could not check whether server supports sharding');
```
malarzm, line 151: Shouldn't this one be marked as `risky` as we can't determine if the test should run or not? This will be easier to spot than `S` which already is showing up in the results.

malarzm (approved):
I have no clue what's `topology` doing in GH actions but I know you know what you're doing :)

tests/Doctrine/ODM/MongoDB/Tests/BaseTest.php
alcaeus: I can also throw an exception to let the test fail: if we can't run `listCommands` this warrants further investigation.

alcaeus:
As for `topology`, this is one of the inputs for the workflow. See the README for the full list: https://github.com/mongodb-labs/drivers-evergreen-tools#using-with-github-actions

tests/Doctrine/ODM/MongoDB/Tests/BaseTest.php
malarzm: Sound good 👍

alcaeus: I've removed the check entirely: the `command` method already throws an exception for failed commands, so the checks are total unnecessary. I've removed them.

PR #2276, opened 2021-02-02
Maintenance work

greg0ire:
Todo for the maintainers:

- [x] rename `master` to `2.3.x` (if that's indeed what should be done)
- [x] tweak settings to change the required jobs
- [x] merge this
- [x] merge up `2.2.x` into the new `2.3.x`

Also, I see that there is still a `.travis.yml`, I left it but it does not seem to be in use in PRs.

### DIFF
--- a/.doctrine-project.json
+++ b/.doctrine-project.json
@@ -6,19 +6,29 @@
     "docsSlug": "doctrine-mongodb-odm",
     "versions": [
         {
-            "name": "2.2",
-            "branchName": "master",
+            "name": "2.3",
+            "branchName": "2.3.x",
             "slug": "latest",
             "upcoming": true
+        },
+        {
+            "name": "2.2",
+            "branchName": "2.2.x",
+            "slug": "2.2",
+            "current": true,
+            "aliases": [
+                "current",
+                "stable",
+                "2.2.x"
+            ]
         },
         {
             "name": "2.1",
             "branchName": "2.1.x",
             "slug": "2.1",
             "current": true,
+            "maintained": false,
             "aliases": [
-                "current",
-                "stable",
                 "2.1.x"
             ]
         },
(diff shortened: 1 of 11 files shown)


.doctrine-project.json
```diff
@@ -7,7 +7,7 @@
     "versions": [
         {
             "name": "2.2",
-            "branchName": "master",
+            "branchName": "2.3.x",
```
SenseException, line 10: Then the name should be `2.3`, right? 2.2 should get its own entry by now.

SenseException (dismissed):

.doctrine-project.json
malarzm: From what I saw in the past (last release) the changes to `.doctrine-project.json` were applied to `master` after the release. There indeed only branchName would need to change: https://github.com/doctrine/mongodb-odm/blob/4269eea27ffdbdc77c07161d59b8fbc06f1ce39d/.doctrine-project.json#L9-L12

greg0ire: Ah right, a different workflow is used on the ODM repos. I'll do the changes here anyway so as to have only one PR and let the changes bubble up with a merge up, WDYT?

malarzm: FWIW we can change this, I just had no clue what to do and followed what was done last time :)

malarzm:
@greg0ire could you please elaborate on "tweak settings to change the required jobs" part? Is this about travis that somehow still tries to run?

greg0ire:
See the "Some checks haven't completed yet" at the bottom? They mention 2 jobs that have different names now. That can and should be changed at https://github.com/doctrine/mongodb-odm/settings/branch_protection_rules/16523510

greg0ire:
Alternatively, if the PR is fine with you, approve it and I will take care of the rest.

malarzm:
Ah I see it now, removed them altogether. We're not going to merge with fails anyway 👍

malarzm (approved):

greg0ire:
Taking your approval as a go ahead, let's do this!

greg0ire:
Ah I see you're doing it already, I'll let you proceed then :)

malarzm:
Ah yeah, I just started clicking since I have some time on my hands :) Prolly we should have coordinated this on Slack :D

malarzm:
Thanks @greg0ire!

PR #2277, opened 2021-02-02
2.2.x merge up to 2.3.x

malarzm:
Merge-up as requested in #2276


malarzm:
Conflicted `.doctrine-project.json` looks as it's supposed to after resolving so I'm merging

Issue #2278, opened 2021-02-03
Locking is broken when ClassMetadata is cached


buffcode:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | all

#### Summary

Lock functionality is broken as soon as a non-volatile ClassMetadata cache is used.

#### Current behavior

When a `metadata_cache_driver` is defined, Doctrine MongoDB caches the `ClassMetadata` using the given driver by serializing all relevant fields. The fields `isLockable` and `lockField` are **not** serialized.

On a second request (or after the cache has been warmed), `ClassMetadata` has no information about locking-behaviour anymore and calls to `->lock(...)` will fail with `undefined index`, as `$this->class->lockField` is empty [[1]] and there is no field mapping for this empty value.

```
#0 vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/UnitOfWork.php(2325): Doctrine\ODM\MongoDB\Persisters\DocumentPersister->lock(Object(MyDocument), 4)
#1 vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/DocumentManager.php(501): Doctrine\ODM\MongoDB\UnitOfWork->lock(Object(MyDocument), 4, NULL)
```

#### How to reproduce

1. Configure a persistent metadata cache driver:

```yaml
# config/packages/doctrine_mongodb.yaml 
services:
    app.doctrine_mongodb.redis.cache:
        class: Doctrine\Common\Cache\PredisCache
        arguments:
            - "@app.doctrine_mongodb.redis.inner"

    app.doctrine_mongodb.redis.inner:
        class: Predis\Client
        arguments:
           - "%env(REDIS_URL)%"

doctrine_mongodb:
    document_managers:
        default:
            auto_mapping: true
            metadata_cache_driver:
                type: service
                id:   app.doctrine_mongodb.redis.cache
```

2. Warm cache (`cache:warmup` or trigger application)
3. Try to lock the document by calling
```php
$documentManager->lock($myDocument, LockMode::PESSIMISTIC_WRITE);
```
4. Notice is thrown, no locking occurs.

#### Expected behavior

Lock functionality should work, even when `ClassMetadata` is cached.

#### Solution

Include `isLockable` and `lockField` in serialization (PRs incoming).

[...]

PR #2280, opened 2021-02-03
[1.3.x] Fix locking when ClassMetadata is unserialized

buffcode:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #2278

#### Summary

Caching / unserializing ClassMetadata broke locking functionality


SenseException:
Version 1 isn't supported anymore. Your PR on 2.2.x will be the main one we'll work with.

PR #2284, opened 2021-02-04
Fix documentation for uploadFromFile

### LINKED ISSUE #2283: Doc Error : GridFSRepository::uploadFromFile
In https://github.com/doctrine/mongodb-odm/blob/2.3.x/docs/en/reference/storing-files-with-gridfs.rst, the documentation describes  the call to `uploadFromFile` as having for parameters:
- 1st parameter : filename
- 2nd Parameter : path to file
```php
$repository = $documentManager->getRepository(Documents\Image::class);
$file = $repository->uploadFromFile('image.jpg', '/tmp/path/to/image');
```

But looking at the method documentation, it is the other way around
```php
    /**
     * Writes the contents of a file to a GridFS file.
     *
     * @param string|null $filename The filename to upload the file with. If no filename is provided, the name of the source file will be used.
     *
     * @return object The newly created GridFS file
     */
    public function uploadFromFile(string $source, ?string $filename = null, ?UploadOptions $uploadOptions = null);
```

so the example in the documentation should be 
```php
$repository = $documentManager->getRepository(Documents\Image::class);
$file = $repository->uploadFromFile('/tmp/path/to/image', 'image.jpg');
```

rrajkomar:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #2283 

#### Summary

Fixed argument order
Fixed example with Metadata object used instead of UpdloadOptions containing the metadata

### DIFF
--- a/docs/en/reference/storing-files-with-gridfs.rst
+++ b/docs/en/reference/storing-files-with-gridfs.rst
@@ -163,7 +163,7 @@ can also open an upload stream and write contents yourself.
     <?php
 
     $repository = $documentManager->getRepository(Documents\Image::class);
-    $file = $repository->uploadFromFile('image.jpg', '/tmp/path/to/image');
+    $file = $repository->uploadFromFile('/tmp/path/to/image', 'image.jpg');
 
 When using the default GridFS repository implementation, the ``uploadFromFile``
 and ``uploadFromStream`` methods return a proxy object of the file you just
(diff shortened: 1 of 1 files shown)


SenseException (changes requested):
That's a good catch. 

Please change the base branch to `2.2.x`, because as far as I understood @malarzm. 2.2.x is still the minimum supported branch despite not being the default branch.

There also a call of `$repository->uploadFromFile()` on line 239 you've missed.

malarzm:
@SenseException I'm going to change the default branch now, forgot to do that yesterday :)

@rrajkomar I tried to change the target branch for you but it produced additional commits. You'll need to rebase your work atop `2.2.x`

rrajkomar:
@malarzm. This should do it.

SenseException:
Please also fix the other `$repository->uploadFromFile('image.jpg', ` parts.

rrajkomar:
@SenseException I think I fixed them all this time. Coudn't find any other concerned reference of the method.

alcaeus (approved):

malarzm:
@SenseException with @alcaeus' approval I'm marking this as good to go.

Thanks @rrajkomar!

PR #2288, opened 2021-02-25
Set type maps on collection level

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

This PR improves the experience for users that create their own `MongoDB\Client` instance for the document manager. Instead of having to set the appropriate type map, ODM will now apply the type map on the collection level when calling `getDocumentCollection`. The end result is the same, as the type map gets applied to every operation (unless the operation has is own type map).

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/DocumentManager.php
+++ b/lib/Doctrine/ODM/MongoDB/DocumentManager.php
@@ -166,15 +166,12 @@ class DocumentManager implements ObjectManager
             'mongodb://127.0.0.1',
             [],
             [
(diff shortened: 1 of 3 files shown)


malarzm (approved):
Nice 👍 I'm eager to make this a bugfix for 2.2.x to make it publicly available in a timely manner :) WDYT?

SenseException (approved):

alcaeus:
@malarzm I went for the next minor version since it's a behavioural change. No rush on this one 😊

PR #2290, opened 2021-02-26
Fix invalid strict comparison when validating mappings

### LINKED ISSUE #2289: Doctrine\ODM\MongoDB\Tools\Console\Command\Schema\ValidateCommand marks each mapping got issues
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.2.0

#### Summary

The ValidateCommand uses `$meta === unserialize(serialize($meta)` to find mapping issues (no idea why it solved like that).
The original code used `$meta == unserialize(serialize($meta)` which could be correct, but the current one is not, cause you'll never get the same object reference again using `serialize` and `unserialize`

https://github.com/doctrine/mongodb-odm/commit/ef8670ece916b40acb5593a22148ebc33893956a#diff-54c6e453a03312a38081f1e6ce97d10c51e2543e9b40c5b0e5f05d9ae922b0c4 changed that.

#### Current behavior

Each mapping got mapping issue in depend if they are mapping issues

#### How to reproduce

Run the command with any mapping

#### Expected behavior

Only the mappings contain issues get reported.

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | Fixes #2289

#### Summary

Automated checkstyle fixes strike again. Strict comparisons will always fail when comparing original metadata to the unserialised variant.

cc @dominikzogg

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Tools/Console/Command/Schema/ValidateCommand.php
+++ b/lib/Doctrine/ODM/MongoDB/Tools/Console/Command/Schema/ValidateCommand.php
@@ -44,7 +44,9 @@ EOT
 
         $errors = 0;
         foreach ($metadataFactory->getAllMetadata() as $meta) {
-            if ($meta === unserialize(serialize($meta))) {
+            // Don't use === to compare as that will always evaluate to false since we receive a different object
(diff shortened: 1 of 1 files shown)


dominikzogg (approved):

malarzm (approved):

malarzm:
Thanks @alcaeus!

Issue #2292, opened 2021-03-05
WriteConcern option never passes into MongoDB


ossinkine:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.2.0

#### Summary

I'm sorry in advance if I've just not figured out.
I want to be sure the transaction was processed on majority replicas, I've set up `w: majority` option in `doctrine_mongodb.default_commit_options`, and try to debug where this option passed into MongoDB and did not find any cases. [`DocumentPersister::getWriteOptions`](https://github.com/doctrine/mongodb-odm/blob/a88db3e46d335190ea0395d44d30d580a345141e/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php#L1491) returns `['w' => 'majority']` and then options passew to [`MongoDB\Collection::updateOne`](https://github.com/mongodb/mongo-php-library/blob/master/src/Collection.php#L1066) which never usew `w` key, it expects `writeConcern` key with object instance of `MongoDB\Driver\WriteConcern`.

#### Current behavior

WriteConcern option never passes into MongoDB.

#### How to reproduce

Just try to pass `w` as write options and check it's never used in `mongodb/mongodb`.

#### Expected behavior

WriteConcern correctly passes info `mongodb/mongodb` and exists in request.

alcaeus:
Now that you mention it...updating the commit options to use `WriteConcern` has completely escaped my mind when updating to the new driver. Marking as a bug and will start fixing soon.

ossinkine:
@alcaeus passing `w` option in `DocumentManager::flush` or specifying in Document annotation also does not work. What is a true way to pass it?

alcaeus:
The true way is to pass a [write concern option](https://www.php.net/manual/en/class.mongodb-driver-writeconcern.php):
```php
$dm->flush(['writeConcern' => new \MongoDB\Driver\WriteConcern('majority')])
```

ossinkine:
Looks like [this method](https://github.com/doctrine/mongodb-odm/blob/a88db3e46d335190ea0395d44d30d580a345141e/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php#L1491) should be removed or refactored because it works wrong and consfusing

alcaeus:
You've found the source of the problem. Want to create a PR to fix the issue? :)

ossinkine:
Yes I can if you describe the correct behavior in your opinion

[...]

PR #2294, opened 2021-03-09
Correctly handle write concern specified in defaultCommitOptions

ossinkine:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #2292

#### Summary

Fix passing `w` option to `DocumentManager::flush` or default commit options

- [X] Bugfix
- [x] Tests

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Configuration.php
+++ b/lib/Doctrine/ODM/MongoDB/Configuration.php
@@ -21,6 +21,7 @@ use Doctrine\ODM\MongoDB\Repository\RepositoryFactory;
 use Doctrine\Persistence\Mapping\Driver\MappingDriver;
 use Doctrine\Persistence\ObjectRepository;
 use InvalidArgumentException;
+use MongoDB\Driver\WriteConcern;
 use ProxyManager\Configuration as ProxyManagerConfiguration;
 use ProxyManager\Factory\LazyLoadingGhostFactory;
 use ProxyManager\GeneratorStrategy\EvaluatingGeneratorStrategy;
@@ -347,7 +348,11 @@ class Configuration
 
     public function getDefaultCommitOptions(): array
     {
-        return $this->attributes['defaultCommitOptions'] ?? ['w' => 1];
+        if (! isset($this->attributes['defaultCommitOptions'])) {
+            $this->attributes['defaultCommitOptions'] = ['writeConcern' => new WriteConcern(1)];
+        }
+
+        return $this->attributes['defaultCommitOptions'];
     }
 
     public function setDefaultCommitOptions(array $defaultCommitOptions): void
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -32,6 +32,7 @@ use MongoDB\Collection;
 use MongoDB\Driver\Cursor;
 use MongoDB\Driver\Exception\Exception as DriverException;
 use MongoDB\Driver\Exception\WriteException;
+use MongoDB\Driver\WriteConcern;
 use MongoDB\GridFS\Bucket;
 use ProxyManager\Proxy\GhostObjectInterface;
 use stdClass;
@@ -39,6 +40,7 @@ use stdClass;
 use function array_combine;
 use function array_fill;
 use function array_intersect_key;
+use function array_key_exists;
 use function array_keys;
 use function array_map;
 use function array_merge;
@@ -61,6 +63,7 @@ use function spl_object_hash;
 use function sprintf;
 use function strpos;
 use function strtolower;
+use function trigger_deprecation;
 
 /**
  * The DocumentPersister is responsible for persisting documents.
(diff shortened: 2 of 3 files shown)


lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -347,7 +348,7 @@ public function getClassMetadataFactoryName(): string
 
     public function getDefaultCommitOptions(): array
     {
-        return $this->attributes['defaultCommitOptions'] ?? ['w' => 1];
+        return $this->attributes['defaultCommitOptions'] ?? ['writeConcern' => new WriteConcern(1)];
```
alcaeus: Even though a write concern is a lightweight object, I feel like we don't need to instantiate it every time:

```suggestion
        if (! isset($this->attributes['defaultCommitOptions'])) {
            $this->attributes['defaultCommitOptions'] = ['writeConcern' => new WriteConcern(1)];
        }

        return $this->attributes['defaultCommitOptions'];
```

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -1493,10 +1498,17 @@ private function getWriteOptions(array $options = []): array
         $defaultOptions  = $this->dm->getConfiguration()->getDefaultCommitOptions();
         $documentOptions = [];
         if ($this->class->hasWriteConcern()) {
-            $documentOptions['w'] = $this->class->getWriteConcern();
+            $documentOptions['writeConcern'] = new WriteConcern($this->class->getWriteConcern());
```
alcaeus, line 1499: I believe this should cover us for now since the write concern option is documented as `string|int|null`. We'll have to refactor the metadata handling to no longer accept a `w` option alone but rather find a way to specify actual write concern objects in metadata.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -61,6 +63,9 @@
 use function sprintf;
 use function strpos;
 use function strtolower;
+use function trigger_error;
+
+use const E_USER_DEPRECATED;
```
alcaeus:
```suggestion
use function trigger_deprecation;
```

alcaeus (commented):
Thanks for tackling this. I've added some suggestions on how to improve the changes you've made. The tests look fine. I think we could profit from some more functional tests (e.g. insert a duplicate document with an unacknowledged write, run into a timeout waiting for the write to be acknowledged, etc.)

lib/Doctrine/ODM/MongoDB/Configuration.php
Steveb-p: Actually, right hand side of this operator is not executed unless the isset test that is part of this operator returns false.

You can check it in the following script:

```php
<?php
declare(strict_types=1);

class Test {
    public function __construct()
    {
        die('x');
    }
}

$test = 1;

$test2 = $test ?? new Test();

echo 'done';

// output will be "done"
```

so you don't need to wrap it like suggested.

alcaeus: What I'm saying is that if you repeatedly call `getDefaultCommitOptions` without any options set, it will always return a new instance of `WriteConcern` with the next call once again instantiating a new object since the attributes array was not modified.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -1493,10 +1498,17 @@ private function getWriteOptions(array $options = []): array
         $defaultOptions  = $this->dm->getConfiguration()->getDefaultCommitOptions();
         $documentOptions = [];
         if ($this->class->hasWriteConcern()) {
-            $documentOptions['w'] = $this->class->getWriteConcern();
+            $documentOptions['writeConcern'] = new WriteConcern($this->class->getWriteConcern());
+        }
+
+        $writeOptions = array_merge($defaultOptions, $documentOptions, $options);
+        if (array_key_exists('w', $writeOptions)) {
+            @trigger_error(sprintf('The "w" option as commit option is deprecated, please pass "%s" object in "writeConcern" option.', WriteConcern::class), E_USER_DEPRECATED);
```
alcaeus: My apologies, this comment was swallowed in my earlier review:

```suggestion
            trigger_deprecation(
                'doctrine/mongodb-odm',
                '2.2',
                'The "w" option as commit option is deprecated, please pass "%s" object in "writeConcern" option.',
                WriteConcern::class
            );
```

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
ossinkine: I didn't get what you mean

ossinkine:
What about some more tests, undoubtedly it will be much better with them, but I'm not sure this bugfix requires them.

alcaeus:
I believe we are good on tests for this change. I'll circle back to add more tests when the handling of write concern in metadata is fixed.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
alcaeus: You are able to specify a write concern in metadata:
```
@ODM\Document(writeConcern="majority")
```

However, there's more to a write concern than just the `w`, but it also includes a `wtimeout` option to specify how long you're willing to wait for the write concern to be achieved. At this time, it's not possible to specify this option, which means that we'll have to refactor this handling in general, at which point we'll no longer have to deal with `w`.

alcaeus (approved):

alcaeus:
Thanks @ossinkine!

PR #2297, opened 2021-03-15
Fix mapping of the nullable option for XML driver

### LINKED ISSUE #2295: XML mapping doesn't take account of the attribute `nullable` for embedded document
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.2.0

#### Summary

I use XML to map the documents.
When I set the attribute `nullable=true` to a property, which is an embedded document (`embed-*`), the `ClassMetadata` resulted always set this property to `ClassMetadata->isNullable() === false`.
By the way, mapping with annotations works well.

Here is an example of the mapping:
```
<document name="Model\Cart" collection="carts">
        <embed-one field="shippingAddress" target-document="Model\ShippingAddress" nullable="true"/>
```

#### Current behavior

Whatever the attribute `nullable` is, when it's set to an embedded document property, the `ClassMetadata->isNullable()` always returns `false`.
For annotation mapping, there is no problem because all the metadata provided in annotations are merged to build the `ClassMetadata`.
But for XML, the metadata are built attribute by attribute and there is no part dealing with `nullable`.
See https://github.com/doctrine/mongodb-odm/blob/8069ded252ff457c481b4da7464ff9db00be41ba/lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php#L354-L395

#### Expected behavior

If the `nullable` attribute is set to `true` to an embedded document property, the `ClassMetadata->isNullable()` must returns `true`.

wuchen90:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | https://github.com/doctrine/mongodb-odm/issues/2295

#### Summary

- [x] Fix XML driver to map the attribute `nullable` for basic fields.
- [x] Fix for embed documents.
- [x] Fix for reference documents.
- [x] Add tests.

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/Mapping/Driver/AbstractDriverTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Mapping/Driver/AbstractDriverTest.php
@@ -12,6 +12,7 @@ use Documents\Phonenumber;
 use Documents\Profile;
 use PHPUnit\Framework\TestCase;
 use TestDocuments\EmbeddedDocument;
+use TestDocuments\NullableFieldsDocument;
 use TestDocuments\PartialFilterDocument;
 use TestDocuments\PrimedCollectionDocument;
 use TestDocuments\QueryResultDocument;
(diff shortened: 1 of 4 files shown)


tests/Doctrine/ODM/MongoDB/Tests/Mapping/Driver/AbstractDriverTest.php
```diff
@@ -392,4 +393,119 @@ public function testCollectionPrimers()
...
+            'isCascadeRefresh' => false,
+            'isCascadeRemove' => false,
+            'isInverseSide' => false,
+            'isOwningSide' => true,
+            'nullable' => true,
+            'strategy' => ClassMetadata::STORAGE_STRATEGY_SET,
+        ], $classMetadata->fieldMappings['username']);
+
+        $this->assertEquals([
+            'association' => 3,
```
malarzm: Could you please use constants for `association` fields and squash your commits later? Everything else is A-OK 👍

tests/Doctrine/ODM/MongoDB/Tests/Mapping/Driver/AbstractDriverTest.php
```diff
@@ -392,4 +393,119 @@ public function testCollectionPrimers()
...
+            'fieldName' => 'address',
+            'name' => 'address',
+            'type' => 'one',
```
malarzm: ah and `one`/`many` also have their constants in `ClassMetadata`

malarzm (approved):

malarzm:
Thanks a lot @wuchen90!

PR #2299, opened 2021-03-31
Fix query preparation when in elemMatch

### LINKED ISSUE #1674: Wrong mapping of $id field.
I think this is a wrong behaviour about mapping field in an embedded document.
I'm using version: 3.1.0

This is my example:

**Main document**
```
/**
 * @ODM\Document
 */
class Main
{
    /**
     * @ODM\Id
     */
    private $id;

    /**
     * @ODM\EmbedMany(targetDocument="Embedded")
     */
    protected $embedded;
}
```

**Embedded document**

```
/**
 * @ODM\EmbeddedDocument
 */
class Embedded
{
    /**
     * @ODM\Field(type="string")
     */
    protected $id;

    /**
     * @ODM\Field(type="boolean")
     */
    protected $active;
}
```

In the embedded document there is a field called $id that is a string type, not @ODM\Id.

The problem comes when i try to query that field.

```
$qb
      ->field('embedded')->elemMatch($qb->expr()
          ->field('id')->equals(1)
     )
;

$qb->getQuery()->execute();
```

This will result in this query:
```"embedded":{"$elemMatch":{"_id":{"$id":"5a0d7aee26567e3c36166391"}}}```

But i suppose to have:
```"embedded":{"$elemMatch":{"id":1}}```

So I investigate in the mapField function and i noticed that:

**Doctrine\ODM\MongoDB\Mapping\ClassMetadataInfo** line 1063: public function mapField(array $mapping)

```
  if (isset($mapping['id']) && $mapping['id'] === true) {
            $mapping['name'] = '_id';
            $this->identifier = $mapping['fieldName'];
            ...
 }
```

### PR DESCRIPTION
**Pull request by alcaeus**
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | Fixes #1674

#### Summary

As [outlined previously](https://github.com/doctrine/mongodb-odm/issues/1674#issuecomment-482079760), fixing the issue requires a little bit of work in the `Expr` class. This PR changes the logic to no longer evaluate `Expr` instances right away, but defer until they are built themselves. This allows us to look at the field mapping and update the `ClassMetadata` instance of `Expr` objects with the necessary information. Despite my previous suggestion to refactor the persister, I decided against it due to the complexity involved.

Note that this logic only applies to associations with a `targetDocument`. If no association of that kind exists, the behaviour is as before and the "parent" `ClassMetadata` instance is used.

There is a separate issue where you can't use `references` or `includesReferenceTo` on an `Expr` instance that will later be used. I'll fix that in a separate PR as that may have to wait until the next minor release due to API changes.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Query/Expr.php
+++ b/lib/Doctrine/ODM/MongoDB/Query/Expr.php
@@ -95,12 +95,7 @@ class Expr
 
         $this->query['$and'] = array_merge(
             $this->query['$and'],
-            array_map(
-                static function ($expression) {
-                    return $expression instanceof Expr ? $expression->getQuery() : $expression;
-                },
-                func_get_args()
-            )
+            func_get_args()
         );
 
         return $this;
@@ -123,9 +118,7 @@ class Expr
 
         $this->query['$nor'] = array_merge(
             $this->query['$nor'],
-            array_map(static function ($expression) {
-                return $expression instanceof Expr ? $expression->getQuery() : $expression;
-            }, func_get_args())
+            func_get_args()
         );
 
         return $this;
@@ -148,9 +141,7 @@ class Expr
 
         $this->query['$or'] = array_merge(
             $this->query['$or'],
-            array_map(static function ($expression) {
-                return $expression instanceof Expr ? $expression->getQuery() : $expression;
-            }, func_get_args())
+            func_get_args()
         );
 
         return $this;
@@ -175,12 +166,8 @@ class Expr
      */
     public function addToSet($valueOrExpression): self
     {
-        if ($valueOrExpression instanceof Expr) {
-            $valueOrExpression = $valueOrExpression->getQuery();
-        }
-
         $this->requiresCurrentField();
-        $this->newObj['$addToSet'][$this->currentField] = $valueOrExpression;
+        $this->newObj['$addToSet'][$this->currentField] = static::convertExpression($valueOrExpression, $this->class);
 
         return $this;
     }
@@ -414,7 +401,7 @@ class Expr
      */
     public function elemMatch($expression): self
     {
-        return $this->operator('$elemMatch', $expression instanceof Expr ? $expression->getQuery() : $expression);
+        return $this->operator('$elemMatch', $expression);
     }
 
     /**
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: lib/Doctrine/ODM/MongoDB/Query/Expr.php [hunk 1]
```diff
@@ -1420,4 +1403,48 @@ private function wrapEqualityCriteria(): void
...
+        return $convertedQuery;
+    }
+
+    /**
+     * Converts expression objects to query arrays. Non-expression values are
+     * returned unmodified.
+     *
+     * @param Expr|mixed $expression
+     */
+    private static function convertExpression($expression, ClassMetadata $classMetadata)
```
malarzm, line 1440: I might be tired but it's confusing we have a non-static convertExpressions and static convertExpression. Don't have any better idea or objections, just pointing this out :)

lib/Doctrine/ODM/MongoDB/Query/Expr.php
```diff
@@ -1420,4 +1403,48 @@ private function wrapEqualityCriteria(): void
...
+
+    private function convertExpressions(array $query, ?ClassMetadata $classMetadata = null): array
+    {
+        if ($classMetadata === null) {
+            $classMetadata = $this->class;
+        }
+
+        $convertedQuery = [];
+        foreach ($query as $key => $value) {
+            if (is_string($key) && $classMetadata->hasAssociation($key)) {
```
malarzm, line 1415: Can a mapped field be named `9` (for whatever reasons) or will it end as a string in the array anyway?

malarzm (approved):
Good idea with deferring the conversion 👍

lib/Doctrine/ODM/MongoDB/Query/Expr.php
alcaeus: I made it static to not be tempted to use `$this->class` for the conversion and instead relying on what the previous logic decided was the right class metadata instance.

lib/Doctrine/ODM/MongoDB/Query/Expr.php
alcaeus: Hmm, that's something PHP doesn't like at all anyways: https://3v4l.org/GoKLK. Reason for this is to reduce the performance impact of the logic so we don't have to check for associations with numeric indexes. Since PHP converts numeric strings to numbers I'm not sure what would happen, but I'm willing to take my chances on this and wait for someone's stuff to break.

PR #2300, opened 2021-04-13
Fix using null values in partial filter expressions

### LINKED ISSUE #2254: PartialFilterExpression with value set to null creates an index with value set to the string "null"
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes/no
| Version     | any

#### Summary

I wanted to create an index with a partial filter expression to index only the documents having a specific field set to `null`. Unfortunately, when using the XML mapping, the values are cast as `string`, therefore the index is created for the documents having the specific field set to the string `"null"`.

#### Current behavior

Currently, only booleans and numbers are handled as specific types. https://github.com/doctrine/mongodb-odm/blob/3bdfe97be472160ebbc924d5882ce61bd640a515/lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php#L525-L531

#### How to reproduce

```
<index>
    <key name="parent" />
    <partial-filter-expression>
        <field name="parent" operator="eq" value="null" />
    </partial-filter-expression>
</index>
```

#### Expected behavior

Adding an exception for null values might be useful.
```php
if ($value === 'true') { 
    $value = true; 
} elseif ($value === 'false') { 
    $value = false; 
} elseif (is_numeric($value)) { 
    $value = preg_match('/^[-]?\d+$/', $value) ? (int) $value : (float) $value; 
} elseif (strtolower($value) === 'null') {
    $value = null;
}
```

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | Fixes #2254

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
@@ -505,16 +505,7 @@ class XmlDriver extends FileDriver
 
         if (isset($xmlIndex->{'option'})) {
             foreach ($xmlIndex->{'option'} as $option) {
-                $value = (string) $option['value'];
-                if ($value === 'true') {
-                    $value = true;
-                } elseif ($value === 'false') {
-                    $value = false;
-                } elseif (is_numeric($value)) {
-                    $value = preg_match('/^[-]?\d+$/', $value) ? (int) $value : (float) $value;
-                }
-
-                $options[(string) $option['name']] = $value;
+                $options[(string) $option['name']] = $this->convertXMLElementValue((string) $option['value']);
             }
         }
 
@@ -624,16 +615,7 @@ class XmlDriver extends FileDriver
 
         if (isset($xmlShardkey->{'option'})) {
             foreach ($xmlShardkey->{'option'} as $option) {
-                $value = (string) $option['value'];
-                if ($value === 'true') {
-                    $value = true;
-                } elseif ($value === 'false') {
-                    $value = false;
-                } elseif (is_numeric($value)) {
-                    $value = preg_match('/^[-]?\d+$/', $value) ? (int) $value : (float) $value;
-                }
-
-                $options[(string) $option['name']] = $value;
+                $options[(string) $option['name']] = $this->convertXMLElementValue((string) $option['value']);
             }
         }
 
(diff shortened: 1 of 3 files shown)


lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
```diff
@@ -581,6 +573,37 @@ private function getPartialFilterExpression(SimpleXMLElement $fields): array
...
+     *
+     * Special strings "false", "true", and "null" are converted to their
+     * respective values. Numeric strings are cast to int or float depending on
+     * whether they contain decimal separators or not.
+     *
+     * @param mixed $value
+     *
+     * @return mixed
+     */
+    private function convertFilterExpressionValue($value)
```
malarzm: Telling by PHPDoc it's not only for filter expressions but rather generic parsing and could be used in some other places in this driver as well I believe :)

malarzm (approved):

lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
alcaeus: Good point. Let me check if we have similar logic for other bool values and consolidate if so.

Steveb-p: On the other hand this private method is used only here and it's always used with a `string`, so maybe enforce `string` type?

You could also use `string|float|int|boolean` (or just `scalar`, it's usually considered as union of the types before in static analysis) so wrong use could be picked up?

SenseException (approved):

lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
alcaeus: @Steveb-p thanks for pointing this out. I had the arg typed as `mixed` due to how the logic was handled before. I've now changed the return type to be `scalar|null` as that's more specific than just `mixed` and made the argument a `string`.

@malarzm replaced a couple more instances and renamed the method to `convertXMLElementValue` - I hope that's specific enough for others to understand.

malarzm (approved):
Awesome, thank you!

alcaeus:
I'll check out the performance test failures separately, but those seem unrelated at first glance.

PR #2302, opened 2021-04-15
Fix errors with nullable typed associations

### LINKED ISSUE #2301: Nullable embedded documents not initialized
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.2.x (498478488c7d8539239e697b2260b4944b2d7b7a)

#### Summary

When using `<embed-one ... nullable="true"/>`, hydration fails due to a missing nullable check in the generated hydrator.

#### Current behavior

When using `<embed-one ... nullable="true"/>` in combination with typed, uninitialized properties, then these properties are not correctly initialized to `null`. This works for regular fields due to the following generated condition in the hydrator:

```php
if (isset($data['theField']) || (! empty($this->class->fieldMappings['theField']['nullable']) && array_key_exists('theField', $data))) {
```

For embedded fields, however, the following condition is generated:

```php
if (isset($data['theField'])) {
```

As you can see, the check for the `nullable` property is missing, hence the field is not initialized.

#### How to reproduce

```php
class Container {
    private int $id;
    private ?Embedded $embedded;

    public function __construct(int $id, ?Embedded $embedded)
    {
        $this->id = $id;
        $this->embedded = $embedded;
    }

    public function getId(): int
    {
        return $this->id;
    }

    public function getEmbedded(): ?Embedded
    {
        return $this->embedded;
    }
}

class Embedded {
    public int $value;

    public function __construct(int $value)
    {
        $this->value = $value;
    }

    public function getValue(): int
    {
        return $this->value;
    }
}
```

```xml
<?xml version="1.0"?>
<doctrine-mongo-mapping xmlns="http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping"
                        xsi:schemaLocation="http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping
                    https://doctrine-project.org/schemas/odm/doctrine-mongo-mapping.xsd">
    <document name="Container">
        <id type="integer" strategy="NONE"/>
        <embed-one field="embedded" target-document="Embedded" nullable="true"/>
    </document>
    <embedded-document name="Embedded">
        <field name="value" type="integer"/>
    </embedded-document>
</doctrine-mongo-mapping>
```

```php
$container = new Container(10, null);

$documentManager->persist($container);
$documentManager->flush();
$documentManager->clear();

$container = $documentManager->find(Container::class, 10);

var_dump($container->getEmbedded());

// => Fatal error: Typed property Container::$embedded must not be accessed before initialization
```

#### Expected behavior

```php
var_dump($container->getEmbedded());

// => null
```

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | Fixed #2301

#### Summary

This fixes errors when loading documents with uninitialised nullable association fields (embedOne and referenceOne were affected).

While fixing this issue, I uncovered an additional bug with public typed properties: when using `$dm->merge` with a document that has uninitialised public properties, the properties are accessed. This may be due to a bug in the reflection code, as the `TypedNoDefaultValueReflectionProperty` class is only used for non-public properties. However, I believe that our workaround is correct, as the class mentioned above returns `null` for uninitialised properties, while we actually don't want to do anything with the property if it wasn't initialised.

### DIFF
--- a/tests/Documents74/TypedDocument.php
+++ b/tests/Documents74/TypedDocument.php
@@ -14,13 +14,28 @@ use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
 class TypedDocument
 {
     /** @ODM\Id() */
-    private string $id;
+    public string $id;
 
     /** @ODM\Field(type="string") */
-    private string $name;
+    public string $name;
 
     /** @ODM\EmbedOne(targetDocument=TypedEmbeddedDocument::class) */
-    private TypedEmbeddedDocument $embedOne;
+    public TypedEmbeddedDocument $embedOne;
+
+    /** @ODM\EmbedOne(targetDocument=TypedEmbeddedDocument::class, nullable=true) */
+    public ?TypedEmbeddedDocument $nullableEmbedOne;
+
+    /** @ODM\EmbedOne(targetDocument=TypedEmbeddedDocument::class, nullable=true) */
+    public ?TypedEmbeddedDocument $initializedNullableEmbedOne = null;
+
+    /** @ODM\ReferenceOne(targetDocument=TypedDocument::class) */
+    public TypedDocument $referenceOne;
+
+    /** @ODM\ReferenceOne(targetDocument=TypedDocument::class, nullable=true) */
+    public ?TypedDocument $nullableReferenceOne;
+
+    /** @ODM\ReferenceOne(targetDocument=TypedDocument::class, nullable=true) */
+    public ?TypedDocument $initializedNullableReferenceOne = null;
 
     /** @ODM\EmbedMany(targetDocument=TypedEmbeddedDocument::class) */
     private Collection $embedMany;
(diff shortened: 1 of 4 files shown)


tests/Documents74/TypedDocument.php
```diff
@@ -14,50 +14,35 @@
 class TypedDocument
 {
     /** @ODM\Id() */
-    private string $id;
+    public string $id;
```
malarzm, line 17: I'm not sure if it's a good idea to change everything to `public` due to logic in https://github.com/doctrine/persistence/blob/fdef2dd7da053bc97841fb45b866a7756bfe0f2f/lib/Doctrine/Persistence/Mapping/RuntimeReflectionService.php#L78-L82. On the other hand it's our reflection mechanism's responsibility to work with both public and private properties so I'm just pointing this out as a potential future problem.

malarzm (approved):

tests/Documents74/TypedDocument.php
alcaeus: This uncovered the problem in merge, and if there's one advantage of types properties it's that you can drop accessors if they should be publicly writable and readable. We'll see what breaks and will have to fix that 🤷‍♂️

malarzm: Not denying the value of the change itself, I was more wondering if separate `PublicTypedDocument` would be a good idea :)

PR #2304, opened 2021-05-15
Use proxy class name resolvers instead of local workaround

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues |

#### Summary

With the release of doctrine/persistence 2.2, we can now leverage class name resolvers in the metadata factory directly and can remove our workaround wherever we fetched metadata.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/DocumentManager.php
+++ b/lib/Doctrine/ODM/MongoDB/DocumentManager.php
@@ -4,7 +4,6 @@ declare(strict_types=1);
 
 namespace Doctrine\ODM\MongoDB;
 
-use Doctrine\Common\Cache\Psr6\DoctrineProvider;
 use Doctrine\Common\EventManager;
 use Doctrine\ODM\MongoDB\Hydrator\HydratorFactory;
 use Doctrine\ODM\MongoDB\Mapping\ClassMetadata;
@@ -36,7 +35,6 @@ use function get_class;
 use function gettype;
 use function is_object;
 use function ltrim;
-use function method_exists;
 use function sprintf;
 
 /**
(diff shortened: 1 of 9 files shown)


lib/Doctrine/ODM/MongoDB/DocumentManager.php
```diff
@@ -288,7 +283,11 @@ public function getSchemaManager(): SchemaManager
         return $this->schemaManager;
     }
 
-    /** Returns the class name resolver which is used to resolve real class names for proxy objects. */
+    /**
+     * Returns the class name resolver which is used to resolve real class names for proxy objects.
+     *
+     * @deprecated
```
malarzm: Shall we point user to what they should be using now? What about a runtime deprecation?

lib/Doctrine/ODM/MongoDB/Proxy/Resolver/CachingClassNameResolver.php
```diff
@@ -4,18 +4,20 @@
 
 namespace Doctrine\ODM\MongoDB\Proxy\Resolver;
 
+use Doctrine\Persistence\Mapping\ProxyClassNameResolver;
+
 /**
  * @internal
  */
-final class CachingClassNameResolver implements ClassNameResolver
+final class CachingClassNameResolver implements ClassNameResolver, ProxyClassNameResolver
```
malarzm, line 12: Shouldn't this implementation be also moved to doctrine/common? I imagine all out libraries will want to have one

lib/Doctrine/ODM/MongoDB/Proxy/Resolver/ClassNameResolver.php
```diff
@@ -4,6 +4,7 @@
 
 namespace Doctrine\ODM\MongoDB\Proxy\Resolver;
 
+/** @deprecated */
```
malarzm: This also lacka information when this will be removed and what to use instead

lib/Doctrine/ODM/MongoDB/DocumentManager.php
alcaeus: That should've been marked internal and people shouldn't use it. I'll add a note in the upgrade file

lib/Doctrine/ODM/MongoDB/Proxy/Resolver/CachingClassNameResolver.php
alcaeus: I can move it to doctrine/common, but that would delay this change. Since this class is internal, I'll mess with it in as many minor releases as I can.

lib/Doctrine/ODM/MongoDB/Proxy/Resolver/ClassNameResolver.php
alcaeus: Yep, will add documentation around this deprecation as well.

malarzm (approved):
Thanks!

PR #2306, opened 2021-05-15
Run stable versions of phpbench

alcaeus:
This will remove the pesky build failure.


malarzm (approved):
Not doing so for 2.2.x?

PR #2307, opened 2021-05-20
Allow mixed value in $not operator

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | 

#### Summary

[When working with regex and `$not`](https://docs.mongodb.com/v4.0/reference/operator/query/not/#not-and-regular-expressions), `$not` operator allows to pass a `MongoDB\BSON\Regex` instance.

Right now it's limited in phpdoc and when using static analysis tools you get a:

`Parameter #1 $expression of method Doctrine\ODM\MongoDB\Query\Builder::not() expects array|Doctrine\ODM\MongoDB\Query\Expr, MongoDB\BSON\Regex given.`

**Update**: should I target `2.3.x`? 🤔

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Query/Builder.php
+++ b/lib/Doctrine/ODM/MongoDB/Query/Builder.php
@@ -1028,11 +1028,11 @@ class Builder
      * @see Expr::not()
      * @see https://docs.mongodb.com/manual/reference/operator/not/
      *
-     * @param array|Expr $expression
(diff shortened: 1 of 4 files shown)


alcaeus:
2.2.x is fine since this is a bugfix. Can you add a test for this please? Thanks!

alcaeus (approved):

alcaeus:
Thanks @franmomu!

PR #2308, opened 2021-05-20
Make ClassMetadata generic

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

Following the changes from `doctrine/persistence`, this PR adds generics to `ClassMetadata` and to `Iterator`.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -13,6 +13,12 @@ parameters:
                  - lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
                  - lib/Doctrine/ODM/MongoDB/DocumentManager.php
 
+        # This cannot be solved the way it is, see https://github.com/vimeo/psalm/issues/5788
+        -
+            message: "#^Return type \\(Doctrine\\\\ODM\\\\MongoDB\\\\Mapping\\\\ClassMetadataFactory\\) of method Doctrine\\\\ODM\\\\MongoDB\\\\DocumentManager\\:\\:getMetadataFactory\\(\\) should be compatible with return type \\(Doctrine\\\\Persistence\\\\Mapping\\\\ClassMetadataFactory\\<Doctrine\\\\Persistence\\\\Mapping\\\\ClassMetadata\\<object\\>\\>\\) of method Doctrine\\\\Persistence\\\\ObjectManager\\:\\:getMetadataFactory\\(\\)$#"
+            count: 1
+            path: lib/Doctrine/ODM/MongoDB/DocumentManager.php
+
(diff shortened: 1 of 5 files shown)


phpstan-baseline.neon
```diff
@@ -13,6 +13,12 @@ parameters:
                  - lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
                  - lib/Doctrine/ODM/MongoDB/DocumentManager.php
 
+        # This cannot be solved the way it is, see https://github.com/vimeo/psalm/issues/5788
+        -
+            message: "#^Return type \\(Doctrine\\\\ODM\\\\MongoDB\\\\Mapping\\\\ClassMetadataFactory\\) of method Doctrine\\\\ODM\\\\MongoDB\\\\DocumentManager\\:\\:getMetadataFactory\\(\\) should be compatible with return type \\(Doctrine\\\\Persistence\\\\Mapping\\\\ClassMetadataFactory\\<Doctrine\\\\Persistence\\\\Mapping\\\\ClassMetadata\\<object\\>\\>\\) of method Doctrine\\\\Persistence\\\\ObjectManager\\:\\:getMetadataFactory\\(\\)$#"
+            count: 1
+            path: lib/Doctrine/ODM/MongoDB/DocumentManager.php
```
alcaeus, line 20: I'm not even going to pretend I understand half of what's going on here, but just to confirm I got this somewhat right: the problem is that we can't correctly say "this manager returns a class metadata factory which creates ORM metadata instances", is that correct?

alcaeus (approved):
LGTM, I appreciate all the work you do here and in our other projects to document these generic classes! 💚

What do you think @franmomu, should we do another check for other generics in ODM before 2.3, or should we defer that to after our release?

phpstan-baseline.neon
franmomu: Yeah, exactly, but in this case if I try to make `ClassMetadataFactory` generic, [I get an error from phpstan I guess because of this generic template of generic maybe.](https://github.com/franmomu/mongodb-odm/pull/11/checks?check_run_id=2637247964)

franmomu:
> What do you think @franmomu, should we do another check for other generics in ODM before 2.3, or should we defer that to after our release?

From the top of the head maybe https://github.com/doctrine/mongodb-odm/blob/2.2.x/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php and all iterators, but that could be some work, so maybe it's better to defer it after the release.

Let me make my last PR in this branch for upgrading `doctrine/coding-standard`.

alcaeus:
Thanks @franmomu!

PR #2311, opened 2021-05-21
Analyze tests directory with psalm

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

Adding static analysis to `tests` directory allow us to spot some issues.

In this case with psalm since with phpstan there are quite a lot issues, I'll be creating PRs to try to fix them little by little (there were more than 300).

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/EcommerceTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/EcommerceTest.php
@@ -78,6 +78,7 @@ class EcommerceTest extends BaseTest
         $this->dm->flush();
         $this->dm->detach($product);
         unset($product);
+        /** @psalm-suppress UndefinedVariable */
         $this->assertFalse(isset($product));
 
         $product = $this->getProduct();
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/EcommerceTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/EcommerceTest.php
@@ -78,8 +78,6 @@ class EcommerceTest extends BaseTest
         $this->dm->flush();
         $this->dm->detach($product);
         unset($product);
-        /** @psalm-suppress UndefinedVariable */
-        $this->assertFalse(isset($product));
 
         $product = $this->getProduct();
         $this->assertCount(2, $product->getOptions());
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/EmbeddedTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/EmbeddedTest.php
@@ -21,6 +21,7 @@ use Documents\Phonenumber;
 use Documents\User;
 use MongoDB\BSON\ObjectId;
 
+use function assert;
 use function get_class;
 
 class EmbeddedTest extends BaseTest
@@ -112,6 +113,7 @@ class EmbeddedTest extends BaseTest
             ->field('id')->equals($user->getId())
             ->getQuery()
             ->getSingleResult();
+        assert($user instanceof User);
         $this->assertNotNull($user);
         $this->assertEquals($addressClone, $user->getAddress());
 
@@ -148,6 +150,7 @@ class EmbeddedTest extends BaseTest
             ->field('id')->equals($user->getId())
             ->getQuery()
             ->getSingleResult();
+        assert($user instanceof User);
         $this->assertNotNull($user);
         $this->assertNull($user->getAddress());
     }
@@ -166,6 +169,7 @@ class EmbeddedTest extends BaseTest
             ->field('id')->equals($user->getId())
             ->getQuery()
             ->getSingleResult();
+        assert($user2 instanceof User);
         $this->assertNotNull($user2);
         $this->assertEquals($user->getPhonenumbers()->toArray(), $user2->getPhonenumbers()->toArray());
     }
@@ -187,10 +191,11 @@ class EmbeddedTest extends BaseTest
         $this->dm->clear();
 
         // retrieve test
-        $test   = $this->dm->createQueryBuilder(get_class($test))
+        $test = $this->dm->createQueryBuilder(get_class($test))
             ->field('id')->equals($test->id)
             ->getQuery()
             ->getSingleResult();
+        assert($test instanceof EmbeddedTestLevel0b);
         $level1 = $test->level1[0];
 
         // $test->level1[0] is available
@@ -227,6 +232,7 @@ class EmbeddedTest extends BaseTest
             ->getQuery()
             ->getSingleResult();
 
+        assert($test instanceof EmbeddedTestLevel0b);
         // $test->level1[0] is available
         $this->assertEquals('test level1 #1', $test->level1[0]->name);
 
@@ -243,6 +249,8 @@ class EmbeddedTest extends BaseTest
             ->field('id')->equals($test->id)
             ->getQuery()
             ->getSingleResult();
+
+        assert($test instanceof EmbeddedTestLevel0b);
 
         $this->assertInstanceOf(PersistentCollection::class, $test->level1);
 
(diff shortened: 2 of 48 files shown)


tests/Doctrine/ODM/MongoDB/Tests/Functional/TestTargetDocument.php
```diff
@@ -30,8 +30,15 @@ class TargetDocumentTestDocument
...
+abstract class AbstractTargetDocumentTestReference
 {
     /** @ODM\Id */
     public $id;
 }
+
+/**
+ * @ODM\Document
+ */
+class TargetDocumentTestReference extends AbstractTargetDocumentTestReference
```
franmomu, line 42: It was instantiating an abstract class in https://github.com/doctrine/mongodb-odm/pull/2311/files#diff-6f1c94e0061bd56051708e36273680c748d5dba9ac84f2cc319533a46ac78063R15

tests/Doctrine/ODM/MongoDB/Tests/Performance/UnitOfWorkPerformanceTest.php
```diff
@@ -22,8 +22,9 @@ class UnitOfWorkPerformanceTest extends BaseTest
      */
     public function testComputeChanges()
     {
+        $n     = 10000;
```
franmomu, line 25: `$n` is used in https://github.com/doctrine/mongodb-odm/pull/2311/files#diff-f3c0f4bbc68f314a5f5a2f8c28f67b508f8bb9c452f14ab29203a2a01cc28f15R48

tests/Documents/Bars/Location.php
```diff
@@ -17,11 +17,6 @@ public function __construct($name = null)
         $this->name = $name;
     }
 
-    public function getId()
-    {
-        return $this->id;
```
franmomu, line 22: `id` property is not in the class.

franmomu:
I got caught up making changes 😬 , I've tried to split it in commits and just change _simple_ things that should be easy to review, otherwise let me know and I can split it in PRs.

After this there are _just_ 130 remaining issues in phpstan in tests.

tests/Doctrine/ODM/MongoDB/Tests/Functional/EcommerceTest.php
```diff
@@ -78,6 +78,7 @@ public function testRemoveOption()
         $this->dm->flush();
         $this->dm->detach($product);
         unset($product);
+        /** @psalm-suppress UndefinedVariable */
```
alcaeus: I'm not sure why the assertion even has to happen. is this test trying to test PHP behaviour? IMO we can the `assertFalse` call instead.

tests/Doctrine/ODM/MongoDB/Tests/Functional/EmbeddedTest.php
```diff
@@ -112,6 +113,7 @@ public function testOneEmbedded()
             ->field('id')->equals($user->getId())
             ->getQuery()
             ->getSingleResult();
+        assert($user instanceof User);
```
alcaeus, line 116: This shows me that we're still missing some psalm template magic. I'll see if I can fix that somehow.

tests/Doctrine/ODM/MongoDB/Tests/Functional/TestTargetDocument.php
alcaeus: The link doesn't go anywhere for me, could you give me a link? This might be a bug since we shouldn't try to instantiate abstract classes.

tests/Doctrine/ODM/MongoDB/Tests/UnitOfWorkTest.php
```diff
@@ -518,7 +518,6 @@ public function testCommitsInProgressIsUpdatedOnException()
         $this->expectExceptionMessage('This should not happen');
 
         $this->dm->flush();
-        $this->assertAttributeSame(0, 'commitsInProgress', $this->dm->getUnitOfWork());
```
alcaeus, line 521: Can we replace this assertion in a different way, maybe by accessing the attribute in a bound closure?

alcaeus (approved):
Great work. You mentioned there being a number of errors left. How would one go about fixing those?

tests/Doctrine/ODM/MongoDB/Tests/Functional/TestTargetDocument.php
franmomu: Oops sorry, as I've split them into commits I guess it got lost, here: https://github.com/doctrine/mongodb-odm/blob/f225a0cfc5ba878aba3e32b7f5fd70facc2d4eee/tests/Doctrine/ODM/MongoDB/Tests/Functional/TestTargetDocument.php#L15

alcaeus: Ok, I'll take a look and see what that test was doing. Thanks!

tests/Doctrine/ODM/MongoDB/Tests/Functional/EmbeddedTest.php
franmomu: Yeah, exactly, it would be great to be able to remove those and eventually I hope we will. `Builder` and `Query` should be generic I guess, but I didn't try yet. (**Edit**: When generics would be added, psalm would complain about this being always true.)

I've also added template to [`Cursor` and `CursorInterface`](https://github.com/vimeo/psalm/blob/master/stubs/mongodb.phpstub) just in case we want to go deeper.

tests/Doctrine/ODM/MongoDB/Tests/UnitOfWorkTest.php
franmomu: Hmmm the thing about this line is that it never gets executed because the `flush()` method throws an exception, but reading the test name: `testCommitsInProgressIsUpdatedOnException` maybe that exception should be catch and check the `commitsInProgress` property?

franmomu:
> You mentioned there being a number of errors left. How would one go about fixing those?

I've created https://github.com/doctrine/mongodb-odm/pull/2317

tests/Doctrine/ODM/MongoDB/Tests/Mapping/ClassMetadataTest.php
```diff
@@ -64,6 +64,7 @@ public function testClassMetadataInstanceSerialization()
         $cm->setVersioned(true);
         $cm->setVersionField('version');
         $validatorJson = '{ "$and": [ { "email": { "$regex": { "$regularExpression" : { "pattern": "@mongodb\\\\.com$", "options": "" } } } } ] }';
+        /** @psalm-suppress TooFewArguments */
```
alcaeus, line 67: What's the cause for this error? Is the documentation here incorrect or is it the code that's wrong?

tests/Doctrine/ODM/MongoDB/Tests/UnitOfWorkTest.php
alcaeus: On second thought, yes. The method should catch the exception, check the `commitsInProgress` property and then return. If no exception was caught, the test should fail using `$this->fail()`.

tests/Doctrine/ODM/MongoDB/Tests/Functional/TestTargetDocument.php
alcaeus: Yes, that change makes absolute sense, thank you!

alcaeus (commented):
One question about the new `@psalm-suppress`, but the rest looks good!

tests/Doctrine/ODM/MongoDB/Tests/UnitOfWorkTest.php
franmomu: perfect, done.

tests/Doctrine/ODM/MongoDB/Tests/Mapping/ClassMetadataTest.php
franmomu: The code is fine, I think the problem is in Psalm that doesn't have a stub specifying that the second parameter is optional. I've created https://github.com/vimeo/psalm/pull/5864

alcaeus (approved):

alcaeus:
Thanks @franmomu!

PR #2314, opened 2021-05-23
Describe iterable type arrays

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

This describes better every `array` type in `ClassMetadata`. 

In the `FieldMapping` psalm type, I've put together all the options I've seen (maybe there are more), the good thing about creating a type is that we can reuse it in more places (like `UnitOfWork` which some of its methods receive an `array $mapping`).

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
@@ -57,6 +57,128 @@ use function trigger_deprecation;
  *    get the whole class name, namespace inclusive, prepended to every property in
  *    the serialized representation).
  *
+ * @psalm-type FieldMappingConfig = array{
+ *      type?: string,
+ *      fieldName?: string,
+ *      name?: string,
+ *      strategy?: string,
+ *      notSaved?: bool,
+ *      association?: int,
+ *      id?: bool,
+ *      collectionClass?: class-string,
+ *      cascade?: list<string>,
+ *      embedded?: bool,
+ *      orphanRemoval?: bool,
+ *      options?: array<string, mixed>,
+ *      nullable?: bool,
+ *      reference?: bool,
+ *      storeAs?: string,
+ *      targetDocument?: class-string,
+ *      mappedBy?: string|null,
+ *      inversedBy?: string|null,
+ *      discriminatorField?: string,
+ *      defaultDiscriminatorValue?: string,
+ *      discriminatorMap?: array<string, class-string>,
+ *      repositoryMethod?: string|null,
+ *      sort?: array<string, string>,
+ *      limit?: int|null,
+ *      skip?: int|null,
+ *      version?: bool,
+ *      lock?: bool,
+ *      notSaved?: bool,
+ *      inherited?: string,
+ *      declared?: class-string
+ * }
+ * @psalm-type FieldMapping = array{
+ *      type: string,
+ *      fieldName: string,
+ *      name: string,
+ *      isCascadeRemove: bool,
+ *      isCascadePersist: bool,
+ *      isCascadeRefresh: bool,
+ *      isCascadeMerge: bool,
+ *      isCascadeDetach: bool,
+ *      isOwningSide: bool,
+ *      isInverseSide: bool,
+ *      strategy?: string,
+ *      notSaved?: bool,
+ *      association?: int,
+ *      id?: bool,
+ *      collectionClass?: class-string,
+ *      cascade?: list<string>,
+ *      embedded?: bool,
+ *      orphanRemoval?: bool,
+ *      options?: array<string, mixed>,
+ *      nullable?: bool,
+ *      reference?: bool,
+ *      storeAs?: string,
+ *      targetDocument?: class-string,
+ *      mappedBy?: string|null,
+ *      inversedBy?: string|null,
+ *      discriminatorField?: string,
+ *      defaultDiscriminatorValue?: string,
+ *      discriminatorMap?: array<string, class-string>,
+ *      repositoryMethod?: string|null,
+ *      sort?: array<string, string>,
+ *      limit?: int|null,
+ *      skip?: int|null,
+ *      version?: bool,
+ *      lock?: bool,
+ *      notSaved?: bool,
+ *      inherited?: string,
+ *      declared?: class-string
+ * }
+ * @psalm-type AssociationMappingField = array{
+ *      type: string,
+ *      fieldName: string,
+ *      name: string,
+ *      isCascadeRemove: bool,
+ *      isCascadePersist: bool,
+ *      isCascadeRefresh: bool,
+ *      isCascadeMerge: bool,
+ *      isCascadeDetach: bool,
+ *      isOwningSide: bool,
+ *      isInverseSide: bool,
+ *      targetDocument: class-string,
+ *      association: int,
+ *      strategy?: string,
+ *      notSaved?: bool,
+ *      id?: bool,
+ *      collectionClass?: class-string,
+ *      cascade?: list<string>,
+ *      embedded?: bool,
+ *      orphanRemoval?: bool,
+ *      options?: array<string, mixed>,
+ *      nullable?: bool,
+ *      reference?: bool,
+ *      storeAs?: string,
+ *      mappedBy?: string|null,
+ *      inversedBy?: string|null,
+ *      discriminatorField?: string,
+ *      defaultDiscriminatorValue?: string,
+ *      discriminatorMap?: array<string, class-string>,
+ *      repositoryMethod?: string|null,
+ *      sort?: array<string, string>,
+ *      limit?: int|null,
+ *      skip?: int|null,
+ *      version?: bool,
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -57,6 +57,43 @@
  *    get the whole class name, namespace inclusive, prepended to every property in
  *    the serialized representation).
  *
+ * @psalm-type FieldMapping = array{
```
franmomu: I've used this psalm type to define the property `fieldMappings` and also as argument for some methods. Maybe this is not totally accurated since for `fieldMappings` I guess there are more mandatory keys than `type`, but I think for now is fine. 

We could maybe create a `FieldMappingParameter` with all the optional keys and then `FieldMapping` restricting the mandatory ones, something like this:

```php
/**
 * @psalm-type FieldMappingParameter = array{
 *     type: string, 
 *     fieldName?: string, 
 *     name?: string, 
 *     strategy?: string, 
 *     ...
 * }
 * @psalm-type FieldMapping = FieldMappingParameter&array{
 *     fieldName: string, 
 *     name: string
 * }
*/
```

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -260,14 +299,16 @@
     /**
      * READ-ONLY: The array of indexes for the document collection.
      *
-     * @var array
+     * @var array<array<string, mixed>>
+     * @psalm-var array<array{keys: array<string, mixed>, options: array<string, mixed>}>
```
alcaeus: I'm wondering if we should define types for this. Consider:
```suggestion
     * @psalm-var array<IndexMapping>
```

Defining an `IndexMapping` psalm type elsewhere that makes use of other types (e.g. `IndexKeys` and `IndexOptions`) would make for easier reading and understanding of what is actually stored.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -260,14 +299,16 @@
...
      */
     public $indexes = [];
 
     /**
      * READ-ONLY: Keys and options describing shard key. Only for sharded collections.
      *
      * @var array<string, array>
+     * @psalm-var array{keys?: array<string, mixed>, options?: array<string, mixed>}
```
alcaeus: Are the keys optional because the shardKey defaults to an empty array instead of `null`? If so, would it make sense to change it?

This is a public property, so we may have to get creative here. Happy to defer it to a future minor release to save time and leave as is for now.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -914,6 +971,8 @@ public function addIndex(array $keys, array $options = []): void
 
     /**
      * Returns the array of indexes for this Document.
+     *
+     * @psalm-return array<array{keys: array<string, mixed>, options: array<string, mixed>}>
```
alcaeus: Defining types for indexes as explained above would also remove the duplication of many of those type definitions.

alcaeus (commented):
Wow, this is cool stuff! I've left some questions below.

As for required and optional keys in the field mapping, there are most likely more required fields other than type, but I'd have to go through this myself to figure that out. I can do so once you're done making changes, but if you feel brave, I'll gladly let you figure out what the code assumes to be there ;)

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
franmomu: Good idea! Done, I've defined them in `ClassMetadata`, but I guess maybe should belong somewhere else.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
franmomu: Exactly, it's because it can be empty. I think there is no special representation for that in phpstan based on https://github.com/phpstan/phpstan/issues/2739#issuecomment-565797975.

IMHO I would leave it as is for now, working with array shapes can become tricky with static analysis (to me at least), we can revisit this when raising static analysis level in other PRs.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -57,6 +57,128 @@
...
+ *      type: string,
+ *      fieldName: string,
+ *      name: string,
+ *      isCascadeRemove: bool,
+ *      isCascadePersist: bool,
+ *      isCascadeRefresh: bool,
+ *      isCascadeMerge: bool,
+ *      isCascadeDetach: bool,
+ *      isOwningSide: bool,
+ *      isInverseSide: bool,
```
franmomu, line 102: This is how far I've arrived, I had to repeat the keys as defining them as intersection of `FielMappingConfig&{mandatory_field: type}` doesn't seem to work properly.

I've copied the mandatory keys at the beginning of the definitions.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
Steveb-p: @franmomu PHPStan supports similar array checks as psalm: https://phpstan.org/writing-php-code/phpdoc-types#array-shapes.

franmomu: Yeah, so here the issue is that `$shardKey` could be either an array with `keys` and `options` keys or an empty array, but doesn't make sense to only have `options` or `keys`. 

In Psalm you could use [`array<empty, empty>`](https://psalm.dev/docs/annotating_code/type_syntax/atomic_types/#empty), so it would look like:

```php
/**
 * @psalm-var array{keys: array<string, mixed>, options: array<string, mixed>}|array<empty, empty>
 */
public $shardKey = [];
```

but PHPStan based on that link:

> PHPStan currently performs a simplification - array where all keys are optional is represented as an empty array + an array where all the keys are present.

you should define all the keys as optional.

alcaeus: Let's leave this as-is for now. I have an idea for a first metadata refactoring and will tackle this then 👍

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -57,6 +57,128 @@
  *    get the whole class name, namespace inclusive, prepended to every property in
  *    the serialized representation).
  *
+ * @psalm-type FieldMappingConfig = array{
+ *      type: string,
+ *      fieldName?: string,
+ *      name?: string,
```
alcaeus, line 63: Looking at this, I believe we need two types:
* The `FieldMapping` type: this type defines how we store the field metadata internally
* The `FieldMappingConfig` type: this type defines what users can pass to the `mapField` method.

For example, the `mapField` method requires either `fieldName` or `name` property to be given, but once the mapping is stored internally, both fields will be set. Having them optional in the internal type requires more type checks than necessary, hence the suggestion to define separate types.

alcaeus (approved):
I left a note on the type, but I believe that what we have here is an improvement worth merging already. I was wanting to experiment with Psalm types a little more, so that will be a nice exercise when refactoring the Metadata class as alluded to in another PR.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
franmomu: Yep, this is how is done, `mapField` is defined like:

```php
/**     
 * Map a field.     
 *     
 * @psalm-param FieldMappingConfig $mapping     
 *     
 * @psalm-return FieldMapping     
 */
```

alcaeus (approved):

alcaeus:
@franmomu can you rebase to resolve the conflicts?

SenseException (approved):

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -57,6 +57,128 @@
  *    get the whole class name, namespace inclusive, prepended to every property in
  *    the serialized representation).
  *
+ * @psalm-type FieldMappingConfig = array{
+ *      type?: string,
```
franmomu, line 61: I had to make `type` also optional in `FieldMappingConfig` because these methods

https://github.com/doctrine/mongodb-odm/blob/f225a0cfc5ba878aba3e32b7f5fd70facc2d4eee/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php#L1252-L1257

do not require `type`.

alcaeus:
Thanks @franmomu!

PR #2315, opened 2021-05-24
Ignore parent parameter name mismatch

franmomu:
~This could lead to problems using named arguments in PHP 8~

|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

Psalm was failing because of https://psalm.dev/docs/running_psalm/issues/ParamNameMismatch/

**UPDATE**: It would be BC break to change the parameter as pointed out in https://github.com/doctrine/mongodb-odm/pull/2315#issuecomment-847253049, so ignoring the issue for now.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/DocumentManager.php
+++ b/lib/Doctrine/ODM/MongoDB/DocumentManager.php
@@ -443,18 +443,18 @@ class DocumentManager implements ObjectManager
      * NOTE: The persist operation always considers documents that are not yet known to
      * this DocumentManager as NEW. Do not pass detached documents to the persist operation.
      *
-     * @param object $document The instance to make managed and persistent.
+     * @param object $object The instance to make managed and persistent.
      *
-     * @throws InvalidArgumentException When the given $document param is not an object.
+     * @throws InvalidArgumentException When the given $object param is not an object.
      */
-    public function persist($document)
+    public function persist($object)
     {
-        if (! is_object($document)) {
-            throw new InvalidArgumentException(gettype($document));
+        if (! is_object($object)) {
+            throw new InvalidArgumentException(gettype($object));
         }
 
         $this->errorIfClosed();
-        $this->unitOfWork->persist($document);
+        $this->unitOfWork->persist($object);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


SenseException:
Psalm is listing the mismatch because of interchangeable classes that should use the same variable names, but the current renaming would AFAIK also be a BC break.

franmomu:
> the current renaming would AFAIK also be a BC break.

Makes sense, I've ignored the psalm issues instead.

SenseException (dismissed):
I'm not sure how to handle this in the next major release though.

malarzm:
Not sure if this is a standard, but lately I've seen `@no-named-arguments` in PHPUnit (https://github.com/sebastianbergmann/phpunit/blob/564e78a6c088916467cd4bc0af2e1bb9da9b1a0c/src/TextUI/ResultPrinter.php#L16). Personally I'd be happy to slap this on most of our public API, named arguments should be used where it makes sense, not everywhere

franmomu:
In fact, in this case are methods with just one argument which I'm not sure if it makes sense to use named parameters. 

About `@no-named-arguments` 👍, to read more about it: https://github.com/Roave/BackwardCompatibilityCheck/issues/264, but I guess that should be decided internally if apply it or not in general, I'll open an issue in `doctrine/persistence`.

alcaeus (changes requested):
IMO:
* Named parameters are a new feature, which highlights inconsistencies in APIs
* Such inconsistencies should either be fixed before named args are widely used, or fixed in a backwards compatible way once thy are

Considering that all of these arguments are single, required method arguments, I'll argue that most people won't be using named arguments for that; and if they are then the impact would be caught early (static analysis should catch using named args that don't exist).

Sorry for the back and forth, but it's preferable to rename the arguments for consistency and be done with this. I've left additional comments regarding `no-named-arguments` in the linked persistence PR, but for ODM I'd definitely prefer cleaning up the API rather than telling people not to use named args.

alcaeus (approved):

alcaeus:
Thanks @franmomu!

franmomu:
> Sorry for the back and forth, but it's preferable to rename the arguments for consistency and be done with this. I've left additional comments regarding `no-named-arguments` in the linked persistence PR, but for ODM I'd definitely prefer cleaning up the API rather than telling people not to use named args.

No worries! It's something that has to be discussed.

PR #2316, opened 2021-05-25
Provide tests to validate issue #2310

wuchen90:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #2310 #2301

PR #2318, opened 2021-05-31
Fix wrong handling for nullable fields in upsert and update

### LINKED ISSUE #2310: Nullable embedded document not initialized when using aggregations
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.2.1

#### Summary

This bug is the same as https://github.com/doctrine/mongodb-odm/issues/2301 but with AggregationBuilder.

#### How to reproduce

```php
class Container
{
    private int $id;
    private ?Embedded $embedded;

    public function __construct(int $id, ?Embedded $embedded)
    {
        $this->id = $id;
        $this->embedded = $embedded;
    }

    public function getId(): int
    {
        return $this->id;
    }

    public function getEmbedded(): ?Embedded
    {
        return $this->embedded;
    }
}

class Embedded
{
    public int $value;

    public function __construct(int $value)
    {
        $this->value = $value;
    }

    public function getValue(): int
    {
        return $this->value;
    }
}
```

```xml
<?xml version="1.0"?>
<doctrine-mongo-mapping xmlns="http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping"
                        xsi:schemaLocation="http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping
                    https://doctrine-project.org/schemas/odm/doctrine-mongo-mapping.xsd">
    <document name="Container">
        <id type="integer" strategy="NONE"/>
        <embed-one field="embedded" target-document="Embedded" nullable="true"/>
    </document>
    <embedded-document name="Embedded">
        <field name="value" type="integer"/>
    </embedded-document>
</doctrine-mongo-mapping>
```

```php
$container = new Container(10, null);

$documentManager->persist($container);
$documentManager->flush();
$documentManager->clear();

$aggBuilder = $documentManager->createAggregationBuilder(Container::class);
$aggBuilder->match()->field('id')->equals(10);

$container = $aggBuilder->hydrate(Container::class)->getAggregation()->getIterator()->current();

var_dump($container->getEmbedded());

// => Fatal error: Typed property Container::$embedded must not be accessed before initialization
```

#### Expected behavior

```php
var_dump($container->getEmbedded());

// => null
```

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | Fixes #2310

#### Summary

#2310 reported another instance of #2301, but on closer inspection it turns out that this was not the same bug but the same symptoms. The bug was wrong handling of `null` values during an `upsert`. I found two separate issues:

1. Nullable embedded fields are not set to `null` when the upsert results in a new document, which causes the bug described in #2310,
2. Nullable embedded fields are not set to `null` on update but rather unset, which also causes the bug described in #2310,
3. Nullable references are always set to `null` on upsert when they shouldn't.

I've kept the tests added in #2316 and expanded them to cover all insert, upsert, and update scenarios. I've also added new tests to `UpsertTest` to specifically test for the behaviour unrelated to typed properties.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
@@ -68,14 +68,11 @@ final class PersistenceBuilder
         foreach ($class->fieldMappings as $mapping) {
             $new = $changeset[$mapping['fieldName']][1] ?? null;
 
-            if ($new === null && $mapping['nullable']) {
-                $insertData[$mapping['name']] = null;
-            }
-
-            /* Nothing more to do for null values, since we're either storing
-             * them (if nullable was true) or not.
-             */
             if ($new === null) {
+                if ($mapping['nullable']) {
+                    $insertData[$mapping['name']] = null;
+                }
+
                 continue;
             }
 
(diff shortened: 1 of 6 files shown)


alcaeus:
I'll have to give the CI pipeline for 2.x a look...

alcaeus:
@malarzm finally managed to fix the bug and the build, so this is ready for a review.

malarzm (approved):
This one ended up bigger than I've expected :)

PR #2320, opened 2021-06-01
Add generics to UoW methods and `DocumentManager::getReference`

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

While checking some errors in https://github.com/doctrine/mongodb-odm/pull/2317, these ones:

https://github.com/doctrine/mongodb-odm/pull/2317/files#diff-995edee38ad4f8387e58ebd52c31bcc04c56cc2448d331b1cf5e0b35c57b9efaR228-R266

are because `DocumentManager::getReference` was not generic.

This PR adds generics to that method and by extension to `UnitOfWork::tryGetById`. Since I was adding generics to UoW, I've also added so other methods like https://github.com/doctrine/orm/commit/1e832a67823a9b56dacedd04d726b121b5711478.

PR #2321, opened 2021-06-02
Fix phpstan build

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

After https://github.com/doctrine/mongodb-odm/pull/2320, there are some issues that are resolved, so there should be removed from the baseline.


alcaeus (approved):

alcaeus:
Thanks @franmomu!

PR #2326, opened 2021-06-21
Add more specific phpdoc to `DocumentRepository`

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

There have been releases of phpstan and psalm fixing the `toPHP` signature.

The phpdoc added to the `DocumentRepository` methods is because Psalm does not work well inheriting phpdoc templates in methods.


alcaeus (approved):

alcaeus:
As so many times in the past few weeks, thank you @franmomu! 🎉

PR #2327, opened 2021-06-21
Fix some issues of Psalm in level 6

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

Right now we are using level 7 of psalm, I raised it locally to level 6 and this PR fixes some of the issues that I think they are easy to review and only in `tests` directory.

The first commits are not SA related, just some polishing.

For the remaning ones I'll open another PR.

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/AtomicSetTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/AtomicSetTest.php
@@ -274,18 +274,18 @@ class AtomicSetTest extends BaseTest
 
         $user = $this->dm->getRepository(get_class($user))->find($user->id);
         $this->assertCount(1, $user->inception);
-        $this->assertEquals($user->inception[0]->value, 'start');
+        $this->assertEquals('start', $user->inception[0]->value);
         $this->assertNotNull($user->inception[0]->one);
-        $this->assertEquals($user->inception[0]->one->value, 'start.one');
+        $this->assertEquals('start.one', $user->inception[0]->one->value);
         $this->assertCount(2, $user->inception[0]->one->many);
-        $this->assertEquals($user->inception[0]->one->many[0]->value, 'start.one.many.0');
-        $this->assertEquals($user->inception[0]->one->many[1]->value, 'start.one.many.1');
+        $this->assertEquals('start.one.many.0', $user->inception[0]->one->many[0]->value);
+        $this->assertEquals('start.one.many.1', $user->inception[0]->one->many[1]->value);
         $this->assertNotNull($user->inception[0]->one->one);
-        $this->assertEquals($user->inception[0]->one->one->value, 'start.one.one');
+        $this->assertEquals('start.one.one', $user->inception[0]->one->one->value);
         $this->assertCount(1, $user->inception[0]->one->one->many);
-        $this->assertEquals($user->inception[0]->one->one->many[0]->value, 'start.one.one.many.0');
+        $this->assertEquals('start.one.one.many.0', $user->inception[0]->one->one->many[0]->value);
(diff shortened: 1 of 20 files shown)


tests/Doctrine/ODM/MongoDB/Tests/SchemaManagerTest.php
```diff
@@ -181,20 +181,24 @@ function (string $fqcn) {
         }
 
         foreach ($this->documentBuckets as $class => $bucket) {
-            $bucket->getFilesCollection()
-                ->expects($this->any())
+            $filesCollection = $bucket->getFilesCollection();
+            assert($filesCollection instanceof Collection && $filesCollection instanceof MockObject);
```
franmomu, line 185: here, `$bucket` is a mock and `getFilesCollection` returns also a mock, not sure how that can be expressed other than this or we can just ignore it.

malarzm: I'm fine with this assert 👍

tests/Doctrine/ODM/MongoDB/Tests/Types/DateImmutableTypeTest.php
```diff
@@ -129,7 +129,10 @@ public function testClosureToPHP($input, $output)
             return $return;
         })($input);
 
-        $this->assertInstanceOf('DateTimeImmutable', $return);
+        // @phpstan-ignore-next-line
```
malarzm, line 132: Why do we need to ignore assert below?

malarzm (approved):

tests/Doctrine/ODM/MongoDB/Tests/Types/DateImmutableTypeTest.php
franmomu: Because of the `eval` in the closure, phpstan says:

```
 ------ ------------------------------------------------------------------------------
  Line   tests/Doctrine/ODM/MongoDB/Tests/Types/DateImmutableTypeTest.php
 ------ ------------------------------------------------------------------------------
  131    Call to function assert() with false will always evaluate to false.
  131    Instanceof between null and DateTimeImmutable will always evaluate to false.
 ------ ------------------------------------------------------------------------------
```
I've been trying other ways, but none worked 😞

malarzm: Yeah, indeed the eval makes things worse... thanks for explaining!

malarzm (approved):

malarzm:
Thanks @franmomu!

tests/Doctrine/ODM/MongoDB/Tests/Functional/AtomicSetTest.php
```diff
@@ -274,18 +274,18 @@ public function testWeNeedToGoDeeper()
 
         $user = $this->dm->getRepository(get_class($user))->find($user->id);
         $this->assertCount(1, $user->inception);
-        $this->assertEquals($user->inception[0]->value, 'start');
+        $this->assertEquals('start', $user->inception[0]->value);
```
alcaeus, line 277: The signature of `assertEquals` is `assertEquals($expected, $actual)` - why are we flipping these?

franmomu: hmm yeah, so here we are checking that the (`actual`) value of `$user->inception[0]->value` is `start` (`expected`).

alcaeus: Sorry. Idiot me decided to comment before having coffee. You are correct.

PR #2328, opened 2021-06-26
Improve type declarations in phpdoc

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

This PR tries to add more specific type declarations in phpdoc to `UnitOfWork` and `Configuration`.

I've added some of psalm types:

#### `CommitOptions`: 
```php
 * @psalm-type CommitOptions array{
 *      fsync?: bool,
 *      safe?: int,
 *      w?: int,
 *      writeConcern?: WriteConcern
 * }
```
These options are from:
 - https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/2.2/reference/working-with-objects.html#flush-options
 - https://github.com/doctrine/mongodb-odm/blob/1db4828c0f3d34e9bd2adbe7459fd1027fcd5848/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php#L1518

#### `Hints`
```php
* @psalm-type Hints = array<int, mixed>
```
I've extracted this one because it might be reused.

#### `Changeset`
```php
 * @psalm-type ChangeSet = array{
 *      0: mixed,
 *      1: mixed
 * }
```
To represent each changeset of a document.

I've created `ClassMetadataTestUtil::getFieldMapping` to be able to create valid `FieldMapping` passing the relevant options, any other name or approach is welcome.

I'll add comments where I think it needs further explanation


tests/Doctrine/ODM/MongoDB/Tests/Mapping/ClassMetadataTest.php
```diff
@@ -636,22 +637,12 @@ public function testAddInheritedAssociationMapping()
...
             'storeAs' => ClassMetadata::REFERENCE_STORE_AS_ID,
-            'isCascadeRemove' => false,
-            'isCascadePersist' => false,
-            'isCascadeRefresh' => false,
-            'isCascadeMerge' => false,
-            'isCascadeDetach' => false,
-            'isOwningSide' => false,
-            'isInverseSide' => false,
-            'targetDocument' => null,
-            'association' => ClassMetadata::REFERENCE_ONE,
```
franmomu, line 653: I've added these in a previous commit

tests/Doctrine/ODM/MongoDB/Tests/PersistentCollectionTest.php
```diff
@@ -42,19 +42,9 @@ public function testExceptionForGetTypeClassWithoutDocumentManager()
...
-            'type' => ClassMetadata::ONE,
-            'name' => 'name',
-            'fieldName' => 'fieldName',
-            'isCascadeRemove' => false,
-            'isCascadePersist' => false,
-            'isCascadeRefresh' => false,
-            'isCascadeMerge' => false,
-            'isCascadeDetach' => false,
-            'isOwningSide' => false,
-            'isInverseSide' => false,
```
franmomu, line 55: The same here, I added them in https://github.com/doctrine/mongodb-odm/commit/5e57d0e033660b3bebe17e718348c8e454639066

phpstan-baseline.neon
```diff
@@ -252,3 +252,9 @@ parameters:
             message: "#^Property Doctrine\\\\ODM\\\\MongoDB\\\\PersistentCollection\\<TKey of \\(int\\|string\\),T of object\\>\\:\\:\\$mapping \\(Doctrine\\\\ODM\\\\MongoDB\\\\PersistentCollection\\\\FieldMapping\\|null\\) does not accept array\\<string, array\\<int\\|string, mixed\\>\\|bool\\|int\\|string\\|null\\>\\.$#"
             count: 1
             path: lib/Doctrine/ODM/MongoDB/PersistentCollection.php
+
+        # There is no way to define a generic array as attribute
```
franmomu, line 256: So there is no way (that I'm aware of) to use `@template` in properties to define something like:

```php
/**
 * @psalm-var array<class-string<T>, array<string, T>>
 *
 * @template T of object
 */
private $identityMap = [];
```

lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -287,6 +314,8 @@ public function getHydratorDir(): ?string
     /**
      * Gets an int flag that indicates whether hydrator classes should always be regenerated
      * during each script execution.
+     *
+     * @psalm-return self::AUTOGENERATE_*
```
alcaeus, line 318: The fact that this works blows my mind. That is nice! 🤯

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -46,6 +47,20 @@
  * The UnitOfWork is responsible for tracking changes to objects during an
  * "object-level" transaction and for writing out changes to the database
  * in the correct order.
+ *
+ * @psalm-import-type FieldMapping from ClassMetadata
+ * @psalm-import-type AssociationFieldMapping from ClassMetadata
+ * @psalm-type ChangeSet = array{
```
alcaeus, line 53: I honestly feel like the work you're doing is taking care of a lot of heavy lifting for a metadata refactoring. These types were all implicitly declared, but with these explicit type declarations we can start designing interfaces to no longer depend on arrays for storing data. I can't thank you enough for doing this work! 💚

phpstan-baseline.neon
alcaeus: @muglug sorry for the shameless ping, but would you have any suggestions on how to achieve this?

alcaeus (approved):

alcaeus:
Thanks @franmomu!

PR #2329, opened 2021-06-26
Psalm level 6

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes/no
| Fixed issues | 

#### Summary

~Just to track all the issues for level 6, I'll create separated PR to fix them since there are _just_ 48, I think they can be solved,~ this is not a blocker for 2.3 release.

Needs https://github.com/doctrine/mongodb-odm/pull/2338

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/GeoNear.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/GeoNear.php
@@ -113,6 +113,8 @@ class GeoNear extends MatchStage
 
     /**
      * The maximum number of documents to return.
+     *
+     * @psalm-suppress InvalidReturnStatement, InvalidReturnType it should return Stage\Limit
      */
     public function limit(int $limit): self
     {
--- a/lib/Doctrine/ODM/MongoDB/Proxy/Factory/StaticProxyFactory.php
+++ b/lib/Doctrine/ODM/MongoDB/Proxy/Factory/StaticProxyFactory.php
@@ -75,8 +75,9 @@ final class StaticProxyFactory implements ProxyFactory
                 ->proxyFactory
                 ->createProxy(
                     $metadata->getName(),
-                    static function () {
+                    static function (): bool {
                         // empty closure, serves its purpose, for now
+                        return true;
                     },
                     [
                         'skippedProperties' => $this->skippedFieldsFqns($metadata),
--- a/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
@@ -1743,6 +1743,8 @@ use function trigger_deprecation;
      */
     public function getIdentifierValues($object): array
     {
+        assert($this->identifier !== null);
+
         return [$this->identifier => $this->getIdentifierValue($object)];
     }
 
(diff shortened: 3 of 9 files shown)


lib/Doctrine/ODM/MongoDB/Aggregation/Stage/GeoNear.php
```diff
@@ -113,6 +113,8 @@ public function includeLocs(string $includeLocs): self
 
     /**
      * The maximum number of documents to return.
+     *
+     * @psalm-suppress InvalidReturnStatement, InvalidReturnType it should return Stage\Limit
```
franmomu, line 117: `GeoNear::num` and `GeoNear::limit` should probably be deprecated https://docs.mongodb.com/manual/reference/operator/aggregation/geoNear/#mongodb-pipeline-pipe.-geoNear in favor of the `Limit` stage.

lib/Doctrine/ODM/MongoDB/Proxy/Factory/StaticProxyFactory.php
```diff
@@ -75,8 +75,9 @@ public function generateProxyClasses(array $classes): int
                 ->proxyFactory
                 ->createProxy(
                     $metadata->getName(),
-                    static function () {
+                    static function (): bool {
                         // empty closure, serves its purpose, for now
+                        return true;
```
franmomu, line 80: I don't know about this if it has some implications.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/GeoNear.php
malarzm: I'm not sure if we were deprecating features deprecated by MongoDB. @alcaeus remember anything?

alcaeus: There are two parts at play here:
- The `num` option to `$geoNear` is deprecated, so we should also deprecate the `num` helper here.
- To take advantage of the `num` option, this stage overrides the `limit` method to use the `num` option instead of adding a new `limit` stage. We can drop the `limit` override in this class, which will keep current functionality (by appending a `$limit` stage instead of using `num` with `$geoNear`)

franmomu: > * To take advantage of the `num` option, this stage overrides the `limit` method to use the `num` option instead of adding a new `limit` stage. We can drop the `limit` override in this class, which will keep current functionality (by appending a `$limit` stage instead of using `num` with `$geoNear`)

This would've been ideal, but tests say the opposite 😞 , `GeoNear::limit` is returning `self` and `Stage::limit` is returning `Limit` stage, so code like this:

https://github.com/doctrine/mongodb-odm/blob/76eac013332cabb39ea4defd82b1b90bfdb93318/tests/Doctrine/ODM/MongoDB/Tests/Aggregation/BuilderTest.php#L117-L123

will fail because `Limit::field` does not exist

lib/Doctrine/ODM/MongoDB/Proxy/Factory/StaticProxyFactory.php
alcaeus: For the proxies we generate, there shouldn't be any difference due to this.

franmomu:
Closing as https://github.com/doctrine/mongodb-odm/pull/2408 was merged

PR #2331, opened 2021-06-26
Fix phpdoc for repositories

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes/no
| Fixed issues | 

#### Summary

Ref: https://github.com/doctrine/mongodb-odm/pull/2329

Removes ObjectRepository as return type since it will always return a more specific one (DocumentRepository, GridFSRepository or ViewRepository)

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Repository/AbstractRepositoryFactory.php
+++ b/lib/Doctrine/ODM/MongoDB/Repository/AbstractRepositoryFactory.php
@@ -11,7 +11,6 @@ use Doctrine\ODM\MongoDB\MongoDBException;
 use Doctrine\Persistence\ObjectRepository;
 
 use function is_a;
-use function ltrim;
 use function spl_object_hash;
 
 /**
@@ -42,7 +41,7 @@ abstract class AbstractRepositoryFactory implements RepositoryFactory
             return $this->repositoryList[$hashKey];
         }
 
-        $repository = $this->createRepository($documentManager, ltrim($documentName, '\\'));
+        $repository = $this->createRepository($documentManager, $documentName);
 
         $this->repositoryList[$hashKey] = $repository;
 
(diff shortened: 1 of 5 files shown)


lib/Doctrine/ODM/MongoDB/Repository/AbstractRepositoryFactory.php
```diff
@@ -42,7 +42,10 @@ public function getRepository(DocumentManager $documentManager, string $document
             return $this->repositoryList[$hashKey];
         }
 
-        $repository = $this->createRepository($documentManager, ltrim($documentName, '\\'));
+        /** @psalm-var class-string<T> $trimmedDocumentName */
+        $trimmedDocumentName = ltrim($documentName, '\\');
```
franmomu: Not sure if the `ltrim` call is really needed, in line 38, before this, there is this call with `$documentName`: 

https://github.com/doctrine/mongodb-odm/blob/1db4828c0f3d34e9bd2adbe7459fd1027fcd5848/lib/Doctrine/ODM/MongoDB/Repository/AbstractRepositoryFactory.php#L38

And then inside `createRepository` the `$documentName` is also used for the same call:

https://github.com/doctrine/mongodb-odm/blob/1db4828c0f3d34e9bd2adbe7459fd1027fcd5848/lib/Doctrine/ODM/MongoDB/Repository/AbstractRepositoryFactory.php#L64

So I guess we could remove the `ltrim` call.

alcaeus: I agree with this, the upstream logic is able to handle this kind of thing. Good catch!

alcaeus (commented):
LGTM with the `ltrim` call dropped.

lib/Doctrine/ODM/MongoDB/Repository/AbstractRepositoryFactory.php
franmomu: Perfect, done!

alcaeus (approved):

malarzm (approved):

malarzm:
Dunno what's up with these 6 pending checks, all seems green from the "Checks" tab :)

malarzm:
Thanks @franmomu!

franmomu:
> Dunno what's up with these 6 pending checks, all seems green from the "Checks" tab :)

I think that is configured in Settings > Branches > Branch protection rules, but no idea why those have showed up 😕

malarzm:
Yup, they were exactly there :)

PR #2333, opened 2021-06-26
Add return static

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

One of the issues from https://github.com/doctrine/mongodb-odm/pull/2329 and also from phpstan is that in

https://github.com/doctrine/mongodb-odm/blob/1db4828c0f3d34e9bd2adbe7459fd1027fcd5848/tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/GraphLookupTest.php#L81-L86

`equals()` method is from `MatchStage` and is returning `self` and `connectFromField` is from `GraphLookup` (which extends `MatchStage`), so returning `static` I think it would be more appropriate.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/MatchStage.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/MatchStage.php
@@ -48,6 +48,8 @@ class MatchStage extends Stage
      *
      * @param array|Expr $expression
      * @param array|Expr ...$expressions
+     *
+     * @return static
      */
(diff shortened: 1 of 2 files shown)


alcaeus (approved):
This being `static` was always the intention, but it probably got lost when we cleaned up phpdoc annotations. Looking forward to PHP 8 so we no longer have this problem.

alcaeus:
Thanks @franmomu!

PR #2334, opened 2021-06-27
Fix handling of upserts during scheduling for deletion

webmozart:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | 

#### Summary

Currently, when a document is scheduled for upsert and later scheduled for deletion, the flush will fail, since the document will be *both* marked for upsert and deletion.

When scheduling for deletion, scheduled insertions and updates are already reverted. This PR adds reverting of scheduled upserts.


alcaeus (approved):

alcaeus:
Thanks @webmozart!

webmozart:
@alcaeus Thanks for merging this so quickly!

PR #2336, opened 2021-07-01
Add missing return types

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | task
| BC Break     | no
| Fixed issues | 

#### Summary

Adds return types (thanks to phpstorm and [psalter](https://psalm.dev/docs/manipulating_code/fixing/)). There are some other things that can be improved based on phpstorm findings, but these changes are enough 😬 .

I wanted to add just simple return types, so I've used phpstorm first because psalter was adding also some specific types for data providers in tests and then use psalter for the ones that phpstorm didn't caught and some manual changes.

I've removed `Group::expression` and `Group::field` (found because psalter was trying to add `return static`):

https://github.com/doctrine/mongodb-odm/blob/4d34a81cd99caea8b99a6df15cb2e47f4d43e803/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Group.php#L76-L101

Since they are in `Operator`:

https://github.com/doctrine/mongodb-odm/blob/4d34a81cd99caea8b99a6df15cb2e47f4d43e803/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Operator.php#L394-L419

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
@@ -44,9 +44,6 @@ class Expr
     /** @var array|null */
     private $switchBranch;
 
-    /**
-     * @inheritDoc
-     */
     public function __construct(DocumentManager $dm, ClassMetadataInterface $class)
     {
         assert($class instanceof ClassMetadata);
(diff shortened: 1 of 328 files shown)


malarzm (approved):
I've skimmed the changes in `lib` and they look OK, those in `tests` I'm gonna take for granted ;)

alcaeus (approved):

SenseException (approved):

PR #2338, opened 2021-07-06
Fix more psalm findings

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

Now there are 23 issues in psalm level 6, this PR fixes 15 of them and for the rest I'll update https://github.com/doctrine/mongodb-odm/pull/2329 so I think it would be easier to review.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/DocumentManager.php
+++ b/lib/Doctrine/ODM/MongoDB/DocumentManager.php
@@ -309,6 +309,7 @@ class DocumentManager implements ObjectManager
      */
     public function getClassMetadata($className): ClassMetadata
     {
+        /** @psalm-var ClassMetadata<T> $metadata */
         $metadata = $this->metadataFactory->getMetadataFor($className);
         assert($metadata instanceof ClassMetadata);
 
--- a/lib/Doctrine/ODM/MongoDB/DocumentManager.php
+++ b/lib/Doctrine/ODM/MongoDB/DocumentManager.php
@@ -306,14 +306,12 @@ class DocumentManager implements ObjectManager
      * @psalm-return ClassMetadata<T>
      *
      * @template T of object
+     *
+     * @psalm-suppress InvalidReturnType, InvalidReturnStatement see https://github.com/vimeo/psalm/issues/5788
      */
     public function getClassMetadata($className): ClassMetadata
     {
-        /** @psalm-var ClassMetadata<T> $metadata */
-        $metadata = $this->metadataFactory->getMetadataFor($className);
-        assert($metadata instanceof ClassMetadata);
-
-        return $metadata;
+        return $this->metadataFactory->getMetadataFor($className);
     }
 
     /**
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -1037,6 +1037,8 @@ final class DocumentPersister
         $preparedQuery = [];
 
         foreach ($query as $key => $value) {
+            $key = (string) $key;
+
             // Recursively prepare logical query clauses
             if (in_array($key, ['$and', '$or', '$nor'], true) && is_array($value)) {
                 foreach ($value as $k2 => $v2) {
@@ -1046,14 +1048,14 @@ final class DocumentPersister
                 continue;
             }
 
-            if (strpos((string) $key, '$') === 0 && is_array($value)) {
+            if (isset($key[0]) && $key[0] === '$' && is_array($value)) {
                 $preparedQuery[$key] = $this->prepareQueryOrNewObj($value, $isNewObj);
                 continue;
             }
 
-            $preparedQueryElements = $this->prepareQueryElement((string) $key, $value, null, true, $isNewObj);
+            $preparedQueryElements = $this->prepareQueryElement($key, $value, null, true, $isNewObj);
             foreach ($preparedQueryElements as [$preparedKey, $preparedValue]) {
-                $preparedValue               = $this->convertToDatabaseValue((string) $key, $preparedValue);
+                $preparedValue               = $this->convertToDatabaseValue($key, $preparedValue);
                 $preparedQuery[$preparedKey] = $preparedValue;
             }
         }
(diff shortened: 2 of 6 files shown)


lib/Doctrine/ODM/MongoDB/DocumentManager.php
```diff
@@ -306,13 +306,12 @@ public function getClassNameResolver(): ClassNameResolver
      * @psalm-return ClassMetadata<T>
      *
      * @template T of object
+     *
+     * @psalm-suppress InvalidReturnType, InvalidReturnStatement see https://github.com/vimeo/psalm/issues/5788
      */
     public function getClassMetadata($className): ClassMetadata
     {
-        $metadata = $this->metadataFactory->getMetadataFor($className);
-        assert($metadata instanceof ClassMetadata);
```
franmomu, line 313: I added this `assert` in another PR to make SA tools happy, but since this is a performance sensitive method, I think it's better to just ignore the issue which is related to https://github.com/vimeo/psalm/issues/5788

lib/Doctrine/ODM/MongoDB/Proxy/Factory/StaticProxyFactory.php
```diff
@@ -124,7 +124,7 @@ private function createInitializer(
     }
 
     /**
-     * @return string[]
+     * @return array<int, string>
```
franmomu, line 127: Change to be in sync with `$proxyOptions` of https://github.com/Ocramius/ProxyManager/blob/21e2b4aa7d7661e7641cc6362fc8635ddcfa8464/src/ProxyManager/Factory/LazyLoadingGhostFactory.php#L85

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -742,7 +742,7 @@ public function enableShardingForDbByDocumentName(string $documentName): void
         }
     }
 
-    private function runShardCollectionCommand(string $documentName, ?WriteConcern $writeConcern = null): array
+    private function runShardCollectionCommand(string $documentName, ?WriteConcern $writeConcern = null): void
```
franmomu, line 745: This value was used prior to https://github.com/doctrine/mongodb-odm/pull/1848, but after that it is not used anymore.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -213,7 +213,7 @@ final class UnitOfWork implements PropertyChangedListener
      * At the end of the UnitOfWork all these collections will make new snapshots
      * of their data.
      *
-     * @psalm-var array<string, PersistentCollectionInterface<array-key, object>>
+     * @psalm-var array<string, array<PersistentCollectionInterface<array-key, object>>>
```
franmomu, line 216: my fault, it was an array containing another one:

https://github.com/doctrine/mongodb-odm/blob/37ac6436cc0a668b21dcb4f9bc5e41688c4ac508/lib/Doctrine/ODM/MongoDB/UnitOfWork.php#L981

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -1626,6 +1626,8 @@ public function removeFromIdentityMap(object $document): bool
      * @throws InvalidArgumentException If the class does not have an identifier.
      *
      * @template T of object
+     *
+     * @psalm-suppress InvalidReturnStatement, InvalidReturnType because of the inability of defining a generic property map
```
franmomu, line 1630: these two suppresses are because the same cause than in https://github.com/doctrine/mongodb-odm/pull/2328#discussion_r659146336

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -1046,7 +1046,7 @@ public function prepareQueryOrNewObj(array $query, bool $isNewObj = false): arra
                 continue;
             }
 
-            if (isset($key[0]) && $key[0] === '$' && is_array($value)) {
+            if (strpos((string) $key, '$') === 0 && is_array($value)) {
```
franmomu: We were getting:

```
ERROR: InvalidArrayAccess - lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php:1049:23 - Cannot access array value on non-array variable $key of type array-key (see https://psalm.dev/005)
            if (isset($key[0]) && $key[0] === '$' && is_array($value)) {
```

lib/Doctrine/ODM/MongoDB/DocumentManager.php
malarzm: People are meant to disable `assert` in production mode anyway :)

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
malarzm: Current code is less performant, `strpos` will go through entire string looking for `$` while what we had previously were taking a first char and checking whether it's a `$`. Instead we should ensure that `$key` is a string and we'll be ok psalm-wise: https://psalm.dev/r/8e2a12f216

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -1409,8 +1409,8 @@ private function hasQueryOperators($value): bool
             $value = get_object_vars($value);
         }
 
-        foreach ($value as $key => $value) {
-            if (isset($key[0]) && $key[0] === '$') {
+        foreach ($value as $key => $notUsedValue) {
+            if (strpos((string) $key, '$') === 0) {
```
malarzm: same as above

lib/Doctrine/ODM/MongoDB/DocumentManager.php
malarzm: But given we need `@psalm-suppress` for two other things I'm fine with this 👍

malarzm:
Future me: please go and do #2329 once we merge this one

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
franmomu: Makes sense 👍, I had used `strpos` to avoid the `isset` check, but didn't think about not having `$` and check the entire string.

> Instead we should ensure that $key is a string and we'll be ok psalm-wise: https://psalm.dev/r/8e2a12f216

I've casted it to `string` since it was also done afterwards.

malarzm (approved):

malarzm:
Thanks @franmomu!

Issue #1824, opened 2018-07-13
Support for 3.6+ $expr in queries


chrisdejager:
I would love to be able to compare fields with each other in a query without needing to fall back on javascript in the where operator.

Since MongoDB 3.6+ we can use expressions within queries: https://docs.mongodb.com/v3.6/reference/operator/query/expr/
Example: db.monthlyBudget.find( { $expr: { $gt: [ "$spent" , "$budget" ] } } )

Using $expr instead of javascript would obviously perform better.

Is this planned for a future release?

Thanx!
Chris

alcaeus:
Yes - we've got some catching up on those. Current focus is on ODM 2.0, which will (finally) get rid of the legacy driver and thus allow us to start working on new features introduced in more recent MongoDB server versions. I'm triaging this for the 2.x series as 2.0 won't bring many new features of that kind - I'd rather start following up with more rapid minor releases in short succession.

chrisdejager:
I expected something like that :) Thanx for the heads up.

Thanks a million for this library!

PR #2343, opened 2021-07-18
Support expr in queries

vincent-le-henaff:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | #1824 

#### Summary

It's my first contribution.. so i'm not sure i'm doing it the right way ... but anyway here what i've done :

- due to naming confusing between actual expr methods (that create new instance of expression helpers) and naming convention of expression operators that implies a future naming collision with $expr operator implementation, i've implemented new methods "createExpr" and add depreciation notice for actual expr methods and avoid BC break. Obviously with the idea that new exprOp method will then get renamed as expr in a futher release.
- So i've implemented "exprOp" method throught QueryBuilder & Expr for both Query and Aggregation parts that implement usage of mongodb $expr operator
- also added a test for this new implemented exprOp method/operator.

ps: thx for your work/time on this lib, that's really helpfull ;)


malarzm:
@Protocteur thanks for your PR and kind words! This one will need to wait for @alcaeus' input as he's the one with the vision for aggregations and operators :) In the meantime could you please rebase your work atop the `2.3.x` branch? This will be a new feature and as such needs to go to the next minor version. Also a small heads up that if we'll go down the deprecation route there'll be couple more administrative things to do

lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
```diff
@@ -162,6 +162,14 @@ public function count(string $fieldName): Stage\Count
         return $stage;
     }
 
+    /**
+     * Create a new Expr instance that can be used as an expression with the Builder
+     */
+    public function createExpr(): Expr
```
alcaeus: If we're introducing new API, I would suggest naming this `createAggregationExpression` to avoid similar mistakes that we made when creating the `expr()` method.

lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
```diff
@@ -180,9 +188,12 @@ public function execute(array $options = []): Iterator
         return $this->getAggregation($options)->getIterator();
     }
 
+    /**
+     * @deprecated use createExpr instead
+     */
     public function expr(): Expr
     {
-        return new Expr($this->dm, $this->class);
+        return $this->createExpr();
```
alcaeus: We also trigger a deprecation notice that users can react to according to their implementation. To do so, use the `trigger_deprecation` helper:
```suggestion
        trigger_deprecation(
            'doctrine/mongodb-odm',
            '2.3',
            'The "%s" method is deprecated. Please use "%s::createAggregationExpr" instead.',
            __METHOD__,
            static::class
        );
        
        return $this->createAggregationExpression();
```

lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
```diff
@@ -333,6 +333,14 @@ public static function convertExpression($expression)
         return $expression;
     }
 
+    /**
+     * Returns a new expression object
+     */
+    public function createExpr(): self
```
alcaeus: `createAggregationExpression` for clarity:
```suggestion
    /**
     * Returns a new expression object
     *
     * @return static
     */
    public function createAggregationExpression(): self
```

lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
```diff
@@ -162,6 +162,14 @@ public function count(string $fieldName): Stage\Count
         return $stage;
     }
 
+    /**
+     * Create a new Expr instance that can be used as an expression with the Builder
+     */
```
alcaeus, line 165:
```suggestion
    /**
     * Create a new Expr instance that can be used as an expression with the Builder
     *
     * @return static
     */
```

lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
```diff
@@ -462,10 +470,25 @@ public function exp($exponent): self
 
     /**
      * Returns a new expression object
+     *
+     * @deprecated use createExpr instead
      */
     public function expr(): self
     {
-        return new static($this->dm, $this->class);
+        return $this->createExpr();
```
alcaeus: See previous suggestion regarding `trigger_deprecation` as it also applies here.

lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
```diff
@@ -462,10 +470,25 @@ public function exp($exponent): self
...
+
+    /**
+     * Specify $expr criteria for the current field.
+     *
+     * @see Builder::expr()
+     * @see https://docs.mongodb.com/manual/reference/operator/query/expr/
+     *
+     * @param array|self $expression
+     */
+    public function exprOp($expression): self
```
alcaeus: `$expr` is a query operator that can't be used as an aggregation operator. This method can be dropped from this class.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/MatchStage.php
```diff
@@ -107,6 +107,15 @@ public function all(array $values): self
         return $this;
     }
 
+    /**
+     * Create a new Expr instance that can be used to build partial expressions
+     * for other operator methods.
+     */
+    public function createExpr(): Expr
```
alcaeus: I'd suggest renaming this to `createQueryExpression` for better distinction from `createAggregationExpression`.

```suggestion
    public function createQueryExpression(): Expr
```

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/MatchStage.php
```diff
@@ -153,12 +162,26 @@ public function exists(bool $bool): self
...
 
     /**
-     * Create a new Expr instance that can be used to build partial expressions
-     * for other operator methods.
+     * @deprecated use createExpr instead
      */
     public function expr(): Expr
     {
-        return $this->builder->matchExpr();
+        return $this->createExpr();
```
alcaeus: This method should also trigger a deprecation notice.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/MatchStage.php
```diff
@@ -153,12 +162,26 @@ public function exists(bool $bool): self
...
+
+    /**
+     * Specify $expr criteria for the current field.
+     *
+     * @see Expr::exprOp()
+     * @see https://docs.mongodb.com/manual/reference/operator/query/expr/
+     *
+     * @param array|\Doctrine\ODM\MongoDB\Aggregation\Expr $expression
+     */
+    public function exprOp($expression): self
```
alcaeus: Naming suggestion:
```suggestion
    public function aggregationExpression($expression): self
```

lib/Doctrine/ODM/MongoDB/Query/Builder.php
```diff
@@ -366,6 +366,17 @@ public function count(): self
         return $this;
     }
 
+    /**
+     * Create a new Expr instance that can be used as an expression with the Builder
+     */
+    public function createExpr(): Expr
```
alcaeus:
```suggestion
    public function createQueryExpression(): Expr
```

lib/Doctrine/ODM/MongoDB/Query/Builder.php
```diff
@@ -493,13 +504,27 @@ public function exists(bool $bool): self
...
     /**
      * Create a new Expr instance that can be used as an expression with the Builder
+     *
+     * @deprecated use createExpr instead
      */
     public function expr(): Expr
     {
-        $expr = new Expr($this->dm);
-        $expr->setClassMetadata($this->class);
+        return $this->createExpr();
```
alcaeus: This method should also trigger a deprecation notice.

lib/Doctrine/ODM/MongoDB/Query/Builder.php
```diff
@@ -493,13 +504,27 @@ public function exists(bool $bool): self
...
-        return $expr;
+    /**
+     * Specify $expr criteria for the current field.
+     *
+     * @see Expr::exprOp()
+     * @see https://docs.mongodb.com/manual/reference/operator/query/expr/
+     *
+     * @param array|Expr $expression
+     */
+    public function exprOp($expression): self
```
alcaeus:
```suggestion
    public function aggregationExpression($expression): self
```

lib/Doctrine/ODM/MongoDB/Query/Builder.php
```diff
@@ -493,13 +504,27 @@ public function exists(bool $bool): self
...
+    }
 
-        return $expr;
+    /**
+     * Specify $expr criteria for the current field.
+     *
+     * @see Expr::exprOp()
+     * @see https://docs.mongodb.com/manual/reference/operator/query/expr/
+     *
+     * @param array|Expr $expression
```
alcaeus: This should not take an `Expr` object but rather an `Aggregation\Expr` as we're expecting an aggregation expression.

lib/Doctrine/ODM/MongoDB/Query/Expr.php
```diff
@@ -404,6 +404,19 @@ public function elemMatch($expression): self
...
 
+    /**
+     * Specify $expr criteria for the current field.
+     *
+     * @see Builder::exprOp()
+     * @see https://docs.mongodb.com/manual/reference/operator/query/expr/
+     *
+     * @param array|Expr $expression
+     */
+    public function exprOp($expression): self
```
alcaeus:
```suggestion
    public function aggregationExpression($expression): self
```

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/AggregationOperatorsProviderTrait.php
```diff
@@ -194,6 +194,11 @@ public static function provideExpressionOperators()
                 'operator' => 'exp',
                 'args' => ['$field'],
             ],
+            'exprOp' => [
+                'expected' => ['$expr' => ['$eq' => ['$field', '$otherField']]],
+                'operator' => 'exprOp',
+                'args' => [['$eq' => ['$field', '$otherField']]],
+            ],
```
alcaeus: Suggestion to account for the new method names:

```suggestion
            'aggregationExpression' => [
                'expected' => ['$expr' => ['$eq' => ['$field', '$otherField']]],
                'operator' => 'aggregationExpression',
                'args' => [['$eq' => ['$field', '$otherField']]],
            ],
```

In the long run, we may want to also pass expression objects to these operator methods, but that's something we should incorporate better than creating separate items in the data provider. Just making a mental note in case anyone wants to take a crack in a separate PR.

alcaeus (changes requested):
Thank you @Protocteur for your contribution! I took a look through the code, and I agree with the way you've decided to solve the problem. I'd suggest using this opportunity to use clearer method names to better allow distinguishing the two different expression classes. To that end, we may want to follow up with a separate PR to rename the expression classes themselves, but that's something I want to discuss with @greg0ire first (there be dragons).

I've left some naming suggestions and some notes about deprecation notices which we provide for engineers. Feel free to apply suggestions from the code review and fix changes in a separate PR. I'm not sure if there's any value in testing the individual methods since we're covering this logic in multiple other tests - I'll let you decide how much test coverage you want to add for this.

Please let me know if you need any help, and please do follow up if I lose track of this PR.

alcaeus:
@Protocteur I changed the base of this PR to 2.3.x since that's where new features go. This unfortunately picked up a number of commits. You can rebase your branch yourself, or let us know if you require assistance with it.

lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
vincent-le-henaff: It's been a while this the begining of that PR :p (i'm on a huge project and finaly almost finished so, i'm cleaning my code now, and it's time to clean that part too .. ^^)

So i made all the changed suggested. :)

vincent-le-henaff:
I've updated method naming like suggested createAggregationExpression and createQueryExpression.

Removed unnecesssary exprOp inside Aggregation\Expr

added depreciation triggers

update testing

I had a lot of work on my main project that used this update, sorry for the delay ;)

vincent-le-henaff (commented):
i merged my branch with doctrine:2.3.x and made requested changes.

Does it need more change now ? I saw my PR request fail at the checking process, but it seems that come from 2.3.x branch more than the change i've made. So this PR will be shelve for now ?

malarzm:
> Does it need more change now ? I saw my PR request fail at the checking process, but it seems that come from 2.3.x branch more than the change i've made. So this PR will be shelve for now ?

For sure you need to take a look at coding standards fail, but most probably that can be easily fixed by running `vendor/bin/phpcbf` locally. I'll look into fixing static analysis fails hopefully this week and will let you know once your branch can be rebased atop new 2.3.x.

vincent-le-henaff (commented):
I've matched your naming suggestions and others changed asked.

I also rebased my branch on 2.5.x because i need the lookup pipeline that has beend included in this branch.

> I'm not used to review process, and didn't took time before to finish the PR/review process. But now i finished my 2 year of crunch (ouch), and i felt i had done everything right (include suggestions in commits) but the PR was still "in stand by". Now i understand (?) that i need to submit my own review to "close" the request and not only commit and comment each suggests from the request. I'm i right ?

composer.json
```diff
@@ -22,7 +22,7 @@
         { "name": "Andreas Braun", "email": "alcaeus@alcaeus.org" }
     ],
     "require": {
-        "php": "^7.4 || ^8.0",
+        "php": "^8.1",
```
malarzm: Welcome back! 🎉 

What happened here? We're definitely not going to make this bump for 2.5.x

vincent-le-henaff: i roll it back, i skipped a bit of digging. And tought you were aiming 8.1 by using ReturnTypeWillChange attribute. And then i read a ref about this and saw it's design for backward compatibility xD

But i got the same error trigger localy as in the workflow about the "ReturnTypeWillChange does not exist", but after updating my php interpreter and extensions used for running phpstan/psalm localy thoses errors aren't triggered anymore. I doesn't have much knowledge around coding standard etc .. (i'm working alone for 15 years now, and unfortunatly doesn't had the occasion to pratice that part) So i'm reading a lot and try to finish this PR the correct way, but i'm really new to team work with this "level" of standard/workflow, and it's a bit rough :p, i'm currently reading psalm docs to not fucked up things. But i feel a bit out of my league on this whole subjects. Not sure i'm doing the correct way (and spam commits while i understand what i'm doing wrong or good xD)

PR #2344, opened 2021-07-23
Add PHP 8.0 Attribute mapping support

### LINKED ISSUE #2337: PHP 8 attributes
### Feature Request

Please implement PHP 8 attributes like Doctrine ORM did.

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

I would love to get rid of all my annotations

IonBazan:
Hi guys, this is my first contribution here 👾  

|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | minor, due to `@NamedArgumentConstructor` - see `UPGRADE-2.3.md`
| Fixed issues | #2337 
| Conflicts with | #2243 (some changes are duplicated there but we need ti check for conflicts or differences)

#### Summary

This PR adds AttributeDriver to allow mapping documents using PHP 8 Attributes to align with ORM functionality.
The driver is actually same as the AnnotationDriver and using custom Reader. 

I noticed some annotations are not documented in the [Annotations Reference page](https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/latest/reference/annotations-reference.html):
- `@DiscriminatorValue` - not documented at all
-  `@Inheritance` - ~~not documented at all, seems not used anywhere~~ Deleted in https://github.com/doctrine/mongodb-odm/pull/2344/commits/5b4a286b382cae356d8f2e29a6200cb596de216c
- `@QueryResultDocument` - fixed in #2347 

#### Progress

- [x] Add the Driver
- [x] Annotate all Annotations/Attributes with `@NamedArgumentConstructor` and add constructors
- [x] Test the mapping against existing drivers
- [x] Mention the new driver in the documentation
- [x] Review Annotations that are not present in the Annotation Reference

#### Notes

- All annotations are now using `@NamedArgumentConstructor` and thus `strict_types` and `Enum` checks are skipped (see https://github.com/doctrine/annotations/blob/f96567ba9e3848f59256b473d730e5ba39d81bfe/lib/Doctrine/Common/Annotations/DocParser.php#L569). I removed the tests relying on these features as they are no longer relevant.
- `#[Document]`'s `indexes` property cannot be used with PHP Attributes as nested Attributes are not supported. This property is deprecated anyway and top level `#[Index]` attribute should be used instead.
- `#[Indexes]` attribute cannot be used for the same reason but it is deprecated too.


alcaeus:
Thank you for the work done so far! @beberlei could I ask you to help out with reviewing this since you made the same change in ORM?

I'm a little strapped for time right now, but we'll try to give feedback as quick as possible. We definitely want this feature in ODM, so thank you again for getting this started!

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/AbstractField.php
```diff
@@ -4,9 +4,7 @@
 
 namespace Doctrine\ODM\MongoDB\Mapping\Annotations;
 
-use Doctrine\Common\Annotations\Annotation;
-
-abstract class AbstractField extends Annotation
+abstract class AbstractField implements Annotation
 {
     /** @var string */
```
franmomu:
```suggestion
    /** @var string|null */
```
It is not related to this PR, but it would be nice to fix these types while modifying these classes (it can be done in another PR).

tests/Doctrine/ODM/MongoDB/Tests/PersistentCollection/CollWithPHP80Types.php
```diff
@@ -9,6 +9,9 @@
 
 class CollWithPHP80Types extends ArrayCollection
 {
+    /**
+     * @psalm-suppress ReservedWord
```
franmomu: There is no need of this suppress because the CI is being executed with php 8, you can run it locally with:

```
vendor/bin/psalm --php-version=8.0
```

IonBazan: It seems that this is an expected behavior - we want the code to be compatible with 7.2 (psalm uses `composer.json` to determine the lowest supported version), it's just this fixture that uses PHP 8 capabilities for testing purposes. I think we should put these files in separate directory (`Tests/Fixtures/`) and exclude from static analysis completely.

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/AbstractField.php
IonBazan: Adding type hints to constructor arguments made it easier to spot some errors - many tests started to fail when I added  type hint for `cascade` which accepts both `array` and `string` while phpdoc stated it's `array` only 😉 
After I finish work on this PR, I can do one more round to check for more inconsistencies in a separate PR.

tests/Doctrine/ODM/MongoDB/Tests/PersistentCollection/CollWithPHP80Types.php
franmomu: > psalm uses `composer.json` to determine the lowest supported version

That is correct, but you can specify the version through `--php-version` option which is what we are using right now:

https://github.com/doctrine/mongodb-odm/blob/e2c6dc41b8b52441eb69d0d79697f3fcc435561e/.github/workflows/static-analysis.yml#L79

and psalm is executing only with PHP 8:

https://github.com/doctrine/mongodb-odm/blob/e2c6dc41b8b52441eb69d0d79697f3fcc435561e/.github/workflows/static-analysis.yml#L60-L62

Ideally, I think it should be executed using several versions, but that would require to have several config files (for phpstan and psalm) since in 7.2 we should exclude files which use PHP 8 syntax.

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/AbstractField.php
beberlei: Its fixing doc bugs that are needed for this PR, from my pov this can be combined here, we did the same for the ORM patch

beberlei (commented):
I see this extends the AnnoationDriver with no extra changes. This will not work or you must require PHP 8.1, because nested attributes are not supported in 8.0. 

I am not super familiar with the ODM, one example i saw #[Indexes] nesting index attributes. 

What we did in ORM is copy adapted the Attribute driver and flattened the attribute hierachy

IonBazan:
@beberlei You are right but the `#[Indexes]` and `#[Document(indexes: ...)]` are the only places we use nested annotations and both are deprecated. That's why I'd suggest to keep it simple and keep the attribute hierarchy flat in the future. I saw the ORM implementation where mapping relies on nested annotations but since we are dropping existing nested annotations soon, I don't think it's worth implementing that in AttributeDriver.

beberlei:
That makes sense then, i never used the ODM before and am clueless about use of nested annotations there, just wanted to make sure that its adressed

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/DiscriminatorField.php
```diff
@@ -4,13 +4,23 @@
...
+use Doctrine\Common\Annotations\Annotation\NamedArgumentConstructor;
 
 /**
  * Specify a field name to store a discriminator value
  *
  * @Annotation
+ * @NamedArgumentConstructor
  */
-final class DiscriminatorField extends Annotation
+#[Attribute(Attribute::TARGET_CLASS | Attribute::TARGET_PROPERTY)]
```
franmomu:
```suggestion
#[Attribute(Attribute::TARGET_CLASS)]
```
Based on https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/2.2/reference/annotations-reference.html#discriminatorfield looks like it's only used at class level (same for `DefaultDiscriminatorValue` and `DiscriminatorMap`).

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/DiscriminatorValue.php
```diff
@@ -4,13 +4,21 @@
...
+use Doctrine\Common\Annotations\Annotation\NamedArgumentConstructor;
 
 /**
  * Use the specified discriminator for this class
  *
  * @Annotation
+ * @NamedArgumentConstructor
  */
-final class DiscriminatorValue extends Annotation
+final class DiscriminatorValue implements Annotation
```
franmomu, line 17: Should this one have `Attribute`? Apparently the annotation is not documented, but it's used

https://github.com/doctrine/mongodb-odm/blob/85d2b9f9664517ca0ece48547be3e68f7a2a2e8f/lib/Doctrine/ODM/MongoDB/Mapping/Driver/AnnotationDriver.php#L94-L95

IonBazan: Good catch, thanks!

IonBazan:
OK I revisited all annotations and compared with changes from #2243 - there were a few places where I missed the default values or properties that shouldn't be set in constructor.

The only problematic one is `Indexes` which I can't make work with `NamedArgumentConstructor` because `$value` can be an array or a single item and can be constructed either by passing an array of indexes or a list of arguments like: 
`@Indexes(@Index, @Index, @Index` so I left it intact as it's deprecated anyway.

alcaeus:
I think we're fine not supporting `#[Indexes]` until initialisers are done. In the meantime, people have other ways to map indexes using attributes. Let's just make sure to call this out in our docs so that people know they can't just change annotation syntax into an attribute syntax without making some minor tweaks.

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/Document.php
```diff
@@ -4,31 +4,59 @@
...
+ * @NamedArgumentConstructor
  */
+#[Attribute(Attribute::TARGET_CLASS)]
 final class Document extends AbstractDocument
 {
     /** @var string|null */
     public $db;
 
-    /** @var string|null */
+    /** @var string|array|null */
```
alcaeus, line 22: Interesting, where do we accept an array as collection? I assumed this would only ever be a string 🤔

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/Indexes.php
```diff
@@ -13,6 +14,7 @@
  *
  * @Annotation
  */
-final class Indexes extends Annotation
+#[Attribute(Attribute::TARGET_CLASS)]
+final class Indexes extends BaseAnnotation implements Annotation
```
alcaeus, line 18: Unlike other annotation classes, this still extends the original `Annotation` class. Is that on purpose or an oversight?

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AttributeDriver.php
```diff
@@ -0,0 +1,32 @@
...
+    {
+        if ($reader === null) {
+            $reader = new AttributeReader();
+        }
+
+        return new self($reader, $paths);
+    }
+}
+
+interface_exists(ClassMetadata::class);
```
alcaeus: Why do we need to load this interface?

tests/Doctrine/ODM/MongoDB/Tests/Mapping/AbstractAnnotationDriverTest.php
```diff
@@ -0,0 +1,352 @@
...
+use Generator;
+use stdClass;
+
+use function get_class;
+
+abstract class AbstractAnnotationDriverTest extends AbstractMappingDriverTest
+{
+    public function testFieldInheritance(): void
+    {
+        // @TODO: This can be a generic test for all drivers
```
alcaeus, line 21: Can you create a refactoring ticket for this and mention it here? That way, people will know which ticket to pick up if they're willing to work on this.

tests/Doctrine/ODM/MongoDB/Tests/Mapping/AbstractAnnotationDriverTest.php
```diff
@@ -0,0 +1,352 @@
...
+use Doctrine\ODM\MongoDB\Mapping\ClassMetadata;
+use Doctrine\ODM\MongoDB\Mapping\Driver\AnnotationDriver;
+use Doctrine\ODM\MongoDB\Mapping\MappingException;
+use Documents\CmsUser;
+use Generator;
+use stdClass;
+
+use function get_class;
+
+abstract class AbstractAnnotationDriverTest extends AbstractMappingDriverTest
```
alcaeus, line 17: Nit: not sure I like the `annotation` part of the name, since this is also used for attributes. I was considering a name that indicates that these mappings are defined in the code itself, but couldn't come up with a name that rolls off the tongue. Happy to leave as is if you can't find a better name.

tests/Doctrine/ODM/MongoDB/Tests/Mapping/AbstractAnnotationDriverTest.php
```diff
@@ -0,0 +1,352 @@
...
+{
+    public function testFieldInheritance(): void
+    {
+        // @TODO: This can be a generic test for all drivers
+        $super  = $this->dm->getClassMetadata(AnnotationDriverTestSuper::class);
+        $parent = $this->dm->getClassMetadata(AnnotationDriverTestParent::class);
+        $child  = $this->dm->getClassMetadata(AnnotationDriverTestChild::class);
+
+        $this->assertFalse($super->hasField('id'), 'MappedSuperclass does not register its own public field');
```
alcaeus, line 26: Huge props for adding messages to all these assertions - those are quite neat!

tests/Doctrine/ODM/MongoDB/Tests/Mapping/XmlMappingDriverTest.php
```diff
@@ -17,7 +18,7 @@
 
 class XmlMappingDriverTest extends AbstractMappingDriverTest
 {
-    protected function loadDriver(): XmlDriver
+    protected function loadDriver(): MappingDriver
```
alcaeus, line 21: I think this change is unnecessary: we're always returning an `XmlDriver`, so the return type was correct.
```suggestion
    protected function loadDriver(): XmlDriver
```

alcaeus (commented):
Gave this a review, but I'd still appreciate someone else taking a look. Great work @IonBazan!

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/AbstractIndex.php
```diff
@@ -4,12 +4,10 @@
...
-use Doctrine\Common\Annotations\Annotation;
-
-abstract class AbstractIndex extends Annotation
+abstract class AbstractIndex implements Annotation
 {
     /** @var string[] */
-    public $keys = [];
+    public $keys;
 
     /** @var string */
```
franmomu:
```suggestion
    /** @var string|null */
```

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/File/ChunkSize.php
```diff
@@ -4,19 +4,19 @@
...
-
-    /** @var string */
-    public $type = 'int';
-
-    /** @var bool */
-    public $notSaved = true;
+    public function __construct(?string $name = 'chunkSize')
+    {
+        parent::__construct($name, 'int', false, [], null, true);
+    }
```
franmomu:
```suggestion
    public function __construct()
    {
        parent::__construct('chunkSize', 'int', false, [], null, true);
    }
```
Never used GridFS, but based on https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/2.2/reference/annotations-reference.html#file-chunksize and https://github.com/mongodb/specifications/blob/master/source/gridfs/gridfs-spec.rst looks like name is not configurable.

It that's right, the same applies to:

- [Filename](https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/2.2/reference/annotations-reference.html#file-filename)
- [Length](https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/2.2/reference/annotations-reference.html#file-length)
- [Metadata](https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/2.2/reference/annotations-reference.html#file-metadata)
- [UploadDate](https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/2.2/reference/annotations-reference.html#file-uploaddate)

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/Document.php
IonBazan: I was surprised too but yes, we support passing the collection as array of parameters: https://github.com/doctrine/mongodb-odm/blob/6e42eed57998ad96f3c868ac5b1329a724024463/tests/Doctrine/ODM/MongoDB/Tests/Functional/CollectionsTest.php#L119-L126

alcaeus: Oh wow, TIL. We can revisit this some other time. Thanks for fixing this!

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/Inheritance.php
```diff
@@ -4,14 +4,17 @@
...
 
 /**
  * Specifies inheritance mapping for a document
  *
  * @Annotation
+ * @NamedArgumentConstructor
  */
-final class Inheritance extends Annotation
+#[Attribute(Attribute::TARGET_CLASS)]
+final class Inheritance implements Annotation
```
franmomu: Is this annotation used? The type `NONE` sounds strange, I couldn't find any usage or example.

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/ReferenceMany.php
```diff
@@ -63,11 +64,62 @@ final class ReferenceMany extends AbstractField
     public $skip;
 
     /** @var string */
-    public $strategy = CollectionHelper::DEFAULT_STRATEGY;
+    public $strategy;
```
franmomu: I think this can be removed because it's in `AbstractField`.

tests/Doctrine/ODM/MongoDB/Tests/Mapping/XmlMappingDriverTest.php
IonBazan: @alcaeus Return type narrowing (covariance) was introduced in PHP 7.4 while we are supporting PHP 7.2+. This method is declared as `abstract` in `AbstractMappingDriverTest` with `MappingDriver` return typehint. The type hint was missing before.

tests/Doctrine/ODM/MongoDB/Tests/Mapping/AbstractAnnotationDriverTest.php
IonBazan: Thanks but I just renamed the `AnnotationDriverTest` and changed it to an abstract test.

tests/Doctrine/ODM/MongoDB/Tests/Mapping/AbstractAnnotationDriverTest.php
IonBazan: Same here but I couldn't find any better name here too. Ideally we should move the whole logic into `AbstractMappingDriverTest` to make sure all drivers have same assertions.

Edit: Tracked in #2354

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/Indexes.php
IonBazan: I decided not to touch this class as it's deprecated and cannot be used as Attribute due to nesting but you are right, it should be amended too.

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/Inheritance.php
IonBazan: This annotation is not used and not documented anywhere. According to the comments in other PRs, it appears that this class is safe to remove then.

tests/Doctrine/ODM/MongoDB/Tests/Mapping/AbstractAnnotationDriverTest.php
IonBazan: Here you go: #2354

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/Indexes.php
IonBazan: @alcaeus I think I know why... It seems that older versions of `doctrine/annotations` won't handle extra arguments as array when using `@NamedArgumentConstructor` (see failing tests):
```php
/**
 * @ODM\Indexes({@ODM\Index(),@ODM\Index()}) - passing an array of indexes
 * @ODM\Indexes(@ODM\Index(),@ODM\Index()) - passing indexes as extra arguments
 */
```

IonBazan: I've reverted the changes applied to this class to bring back support for dynamic constructor arguments. While I don't have any better idea to fix that, I'd appreciate any suggestions here.

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/File/Filename.php
```diff
@@ -4,19 +4,19 @@
...
-
-    /** @var string */
-    public $type = 'string';
-
-    /** @var bool */
-    public $notSaved = true;
+    public function __construct(?string $name = 'filename')
+    {
+        parent::__construct($name, 'string', false, [], null, true);
+    }
```
franmomu:
```suggestion
    public function __construct()
    {
        parent::__construct('filename', 'string', false, [], null, true);
    }
```

The same reason than https://github.com/doctrine/mongodb-odm/pull/2344#discussion_r683185514

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/File/Length.php
```diff
@@ -4,19 +4,19 @@
...
-
-    /** @var string */
-    public $type = 'int';
-
-    /** @var bool */
-    public $notSaved = true;
+    public function __construct(?string $name = 'length')
+    {
+        parent::__construct($name, 'int', false, [], null, true);
+    }
```
franmomu:
```suggestion
    public function __construct()
    {
        parent::__construct('length', 'int', false, [], null, true);
    }
```

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/File/Metadata.php
```diff
@@ -32,4 +30,23 @@ final class Metadata extends AbstractField
...
+        bool $nullable = false,
+        array $options = [],
+        ?string $strategy = null,
+        bool $notSaved = false,
+        ?string $targetDocument = null,
+        ?string $discriminatorField = null,
+        ?array $discriminatorMap = null,
+        ?string $defaultDiscriminatorValue = null
+    ) {
+        parent::__construct($name, ClassMetadata::ONE, $nullable, $options, $strategy, $notSaved);
```
franmomu:
```suggestion
    public function __construct(
        bool $nullable = false,
        array $options = [],
        ?string $strategy = null,
        bool $notSaved = false,
        ?string $targetDocument = null,
        ?string $discriminatorField = null,
        ?array $discriminatorMap = null,
        ?string $defaultDiscriminatorValue = null
    ) {
        parent::__construct('metadata', ClassMetadata::ONE, $nullable, $options, $strategy, $notSaved);
```

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/File/UploadDate.php
```diff
@@ -4,19 +4,19 @@
...
-
-    /** @var string */
-    public $type = 'date';
-
-    /** @var bool */
-    public $notSaved = true;
+    public function __construct(?string $name = 'uploadDate', string $type = 'date')
+    {
+        parent::__construct($name, $type, false, [], null, true);
+    }
```
franmomu:
```suggestion
    public function __construct()
    {
        parent::__construct('uploadDate', 'date', false, [], null, true);
    }
```

tests/Doctrine/ODM/MongoDB/Tests/Mapping/XmlMappingDriverTest.php
alcaeus: Oh, I thought we got that in 7.2. Never mind then...

malarzm:
Just wanted to let you know I'll get to this PR somewhen next week. I'm a bit swamped at work before going on vacation ;)

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AttributeDriver.php
```diff
@@ -0,0 +1,27 @@
...
+declare(strict_types=1);
+
+namespace Doctrine\ODM\MongoDB\Mapping\Driver;
+
+use Doctrine\Common\Annotations\Reader;
+
+/**
+ * The AnnotationDriver reads the mapping metadata from docblock annotations.
+ */
+class AttributeDriver extends AnnotationDriver
```
franmomu, line 12: I was trying the branch and I had to add the constructor like in https://github.com/doctrine/orm/blob/106ed8009a0c581a1566775cbd4d2c4cc93976b1/lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php#L33-L39

Looks like the `create` method was introduced for testing purposes in https://github.com/doctrine/mongodb-odm/commit/5d783c3274186e22ba7a3914ff93184e76c7f9e0

.github/workflows/continuous-integration.yml
```diff
@@ -35,7 +35,7 @@ jobs:
           - "highest"
         include:
           - deps: "lowest"
-            os: "ubuntu-16.04"
+            os: "ubuntu-18.04"
```
Ahummeling: Is this change related to the support for php attributes in some way? Or is this just the point in time the ancient ubuntu version is updated to something less ancient? Just out of curiousity really

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/File/ChunkSize.php
```diff
@@ -4,19 +4,19 @@
...
 {
-    /** @var string */
-    public $name = 'chunkSize';
-
-    /** @var string */
-    public $type = 'int';
-
-    /** @var bool */
-    public $notSaved = true;
+    public function __construct()
```
Ahummeling, line 18: At the risk of sounding stupid, I'm not really getting the idea behind a constructor with no arguments, that calls its parent with fixed arguments. Is this some attribute related pattern? <br>
I see the same thing being done in `Filename` and `Length`, but not in `Metadata` and `Id`. Can you explain the purpose of these classes? I

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AttributeDriver.php
```diff
@@ -0,0 +1,27 @@
...
+{
+    /**
+     * Factory method for the Attribute Driver
+     *
+     * @param string[]|string $paths
+     */
+    public static function create($paths = [], ?Reader $reader = null): AnnotationDriver
```
Ahummeling, line 21: Should the `$reader` not be an instance of `AttributeReader`? Also the return type seems incorrect to me, I would expect `AttributeDriver::create` to return an instance of `AttributeDriver`.  <br>
If not, then perhaps this method is probably better placed in the parent class, with the introduction of a `createDefaultReader` method, you'd leave the generic stuff in the parent chill, with implementing only the class-specific stuff in the child classes. <br>
Just a suggestion really, interested in your thoughts on this

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AttributeReader.php
```diff
@@ -0,0 +1,95 @@
...
+
+    public function getClassAnnotation(ReflectionClass $class, $annotationName)
+    {
+        foreach ($this->getClassAnnotations($class) as $annotation) {
+            if ($annotation instanceof $annotationName) {
+                return $annotation;
+            }
+        }
+
+        return null;
```
Ahummeling, line 36: Would it not be better to through an exception at this point? I'm not really sure how the doctrine projects handle returning null vs throwing an exception..

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AttributeReader.php
```diff
@@ -0,0 +1,95 @@
...
+            $attributeName = $attribute->getName();
+            assert(is_string($attributeName));
+            // Make sure we only get Doctrine Annotations
+            if (! is_subclass_of($attributeName, Annotation::class)) {
+                continue;
+            }
+
+            $instance = $attribute->newInstance();
+            assert($instance instanceof Annotation);
+            $instances[] = $instance;
```
Ahummeling, line 90: Consider `$instances[$attributeName] = $instance;` That would allow you to have the `getXAnnotation($reflection, $name)` methods simply do `return $this->getXAnnotations($reflection)[$name] ?? nulll;` preventing you from having to loop over them and instance checking the.

Ahummeling (approved):
So I feel like I've already gone into too much technical detail already, feel free to ignore / resolve my comments at will. This is my first code review I've done outside of my own company. If my comments seem strange or out of place, please let me know, so I can work on it.

.github/workflows/continuous-integration.yml
IonBazan: This is because Setup-PHP action dropped support for Ubuntu 16.04 and CI started to fail. This is no longer relevant as 2.2.x has been merged to 2.3.x which refactored the whole CI workflow.

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/File/ChunkSize.php
IonBazan: This is because these annotations don't accept any parameters and they just extend `AbstractField` setting some hardcoded parameters. Consider `ChunkSize ` as an alias of a field that always named `'chunkSize'` and is an `int`.

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AttributeDriver.php
IonBazan: This class is extending `AnnotationDriver` and thus cannot alter the argument nor return type unless we upgrade to PHP 7.4+.

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AttributeReader.php
IonBazan: According to https://github.com/doctrine/annotations/blob/1.14.x/lib/Doctrine/Common/Annotations/Reader.php interface, it should return `null` when the annotation is not found. In fact, these methods are not used in metadata driver at all as i relies on `getClassAnnotations` to get all attributes for class.

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AttributeReader.php
IonBazan: Please note that some attributes like `#[Index]` may be repeated in the class context. This is why we would need to make the structure like `$instances[$attributeName][] = $instance` but we would still need to loop to support polymorphism - `getClassAnnotation(AbstractIndex::class)` should return either `Index` or `UniqueIndex`.

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AttributeDriver.php
IonBazan: I've added the constructor but it seems that it doesn't make much sense here. I think we should use the one from parent class.

franmomu: The thing here is that using the bundle it creates the service definition for the driver here I think:

https://github.com/symfony/symfony/blob/83bdd2eacf593ec8ab4511afccba75bcaf92ef66/src/Symfony/Bridge/Doctrine/DependencyInjection/AbstractDoctrineExtension.php#L210-L212

And it only receives the array of paths.

IonBazan: Thanks for clarification - in that case, the current constructor should work with that. I will try adding the integration to MongoDBBundle just to make sure it works.

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/Indexes.php
alcaeus: Leaving as is is fine for the time being. @beberlei any suggestion on how to solve this?

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AttributeDriver.php
alcaeus: > I will try adding the integration to MongoDBBundle just to make sure it works.

For prior art, you can check out the same change in DoctrineBundle: https://github.com/doctrine/DoctrineBundle/pull/1322. Changes should be pretty much the same, since the bundles share the same base architecture.

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/File/ChunkSize.php
alcaeus: For some background on GridFS mappings, the [GridFS specification](https://github.com/mongodb/specifications/blob/master/source/gridfs/gridfs-spec.rst) defines what the documents look like:
```
{
  "_id" : <TFileId>,
  "length" : <Int64>,
  "chunkSize" : <Int32>,
  "uploadDate" : <BSON datetime, ms since Unix epoch in UTC>,
  "md5" : <hex string>,
  "filename" : <string>,
  "contentType" : <string>,
  "aliases" : <string array>,
  "metadata" : <Document>
}
```

This means that when mapping a GridFS bucket to a document class, all of the "system fields" always have the same field names, types, properties, etc. Only the `metadata` field can be customised by the user, and only by specifying an `EmbeddedDocument` to be used for this purpose.

That's why these annotations take an empty constructor: there's nothing to configure for these fields at all, we only need to tell ODM that they should put the filename of a file into a specific property. The only difference is the `File\Metadata` annotation, which allows the user to set the `targetDocument` being used for metadata storage.

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AttributeDriver.php
alcaeus: Can we add `@return` and `@param` doc entries to rectify this for static analysis?

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AttributeReader.php
alcaeus: Ah, interesting idea. There are indeed some annotations and attributes that can be repeated, but this is a good idea to keep in the back of your head. I'll also add that the metadata readers should only be used once in the entire deployment. After code is deployed, we can read all metadata and store it in a persistent metadata cache (e.g. a file cache, but not a cache shared between multiple machines). In the context of ODM this makes the metadata classes less performance sensitive, as they won't be used in production usage.

alcaeus:
@Ahummeling:
> So I feel like I've already gone into too much technical detail already, feel free to ignore / resolve my comments at will. This is my first code review I've done outside of my own company. If my comments seem strange or out of place, please let me know, so I can work on it.

If you hadn't mentioned it, I wouldn't have been able to tell. For the benefit of everyone reading this, reviewing code for OSS projects follows the same principles as every other code review you do for your job: you check if the implementation is sound, test coverage is sufficient, and whether edge cases are covered. There's no additional magic involved, except for the butterflies in your stomach after merging and potentially breaking things for many developers :)

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/AlsoLoad.php
```diff
@@ -4,15 +4,30 @@
...
  * @Annotation
+ * @NamedArgumentConstructor
  */
-final class AlsoLoad extends Annotation
+#[Attribute(Attribute::TARGET_PROPERTY | Attribute::TARGET_METHOD)]
+final class AlsoLoad implements Annotation
 {
-    /** @var string */
+    /** @var string|string[] */
+    public $value;
```
malarzm, line 20: I wonder if we can type all the properties now given we'll have ctors everywhere. That can left for another PR tho

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/AbstractDocument.php
```diff
@@ -4,8 +4,6 @@
 
 namespace Doctrine\ODM\MongoDB\Mapping\Annotations;
 
-use Doctrine\Common\Annotations\Annotation;
-
-abstract class AbstractDocument extends Annotation
+abstract class AbstractDocument implements Annotation
```
malarzm, line 7: I believe the fact that our annotations no longer extend `Doctrine\Common\Annotations\Annotation` should be noted in UPGRADE.md. I'm fine with the change on its own 👍

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/Document.php
malarzm: IIRC cappedncollections are the only use case

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/Indexes.php
malarzm: FWIW this class was deprecated as a preparation for attribute mapping, I think we can leave it as is. Whoever will want to use attributes for indices can do so as per deprecation suggests :)

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/Inheritance.php
```diff
@@ -1,27 +0,0 @@
...
-namespace Doctrine\ODM\MongoDB\Mapping\Annotations;
-
-use Doctrine\Common\Annotations\Annotation;
-
-/**
- * Specifies inheritance mapping for a document
- *
- * @Annotation
- */
-final class Inheritance extends Annotation
```
malarzm, line 14: IIRC this was unused but its removal still warrants a note in the UPGRADE.md file

lib/Doctrine/ODM/MongoDB/Types/DateImmutableType.php
```diff
@@ -21,19 +17,11 @@ public static function getDateTime($value): DateTimeInterface
...
 
-        if ($datetime instanceof DateTimeImmutable) {
-            return $datetime;
-        }
-
         if ($datetime instanceof DateTime) {
             return DateTimeImmutable::createFromMutable($datetime);
         }
 
-        throw new RuntimeException(sprintf(
```
malarzm: These changes look unrelated and last time we talked about them we decided the code's fine as is. Since both versions are fine I'll leave the decision to the rest of maintainers

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH816Test.php
```diff
@@ -2,9 +2,10 @@
 
 declare(strict_types=1);
 
-namespace Doctrine\ODM\MongoDB\Tests;
+namespace Doctrine\ODM\MongoDB\Tests\Functional\Ticket;
```
malarzm, line 5: Oh boy 😅 thanks for fixing these!

tests/Doctrine/ODM/MongoDB/Tests/Mapping/AttributeDriverTest.php
```diff
@@ -0,0 +1,19 @@
...
+
+declare(strict_types=1);
+
+namespace Doctrine\ODM\MongoDB\Tests\Mapping;
+
+use Doctrine\ODM\MongoDB\Mapping\Driver\AttributeDriver;
+use Doctrine\Persistence\Mapping\Driver\MappingDriver;
+
+/**
+ * @requires PHP 8.0
```
malarzm, line 12: Cool, didn't know about the `@requires` annotation. It'd be great to go through our tests and look for our old manual skips because of PHP's version.

malarzm (approved):
Thanks for the amazing work here!

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/AlsoLoad.php
IonBazan: As we still support PHP 7.2, the properties can't be typed yet. However, the type hints were introduced to all constructors where possible.

lib/Doctrine/ODM/MongoDB/Types/DateImmutableType.php
IonBazan: It is actually an edge case that PHP doesn't support implementing the `DateTimeInterface` directly in userland so I think I will just revert the change in case this behavior ever changes in future PHP versions.

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH816Test.php
IonBazan: These tests were failing while I was working on this feature and IDE kept complaining about the namespaces 😄

tests/Doctrine/ODM/MongoDB/Tests/Mapping/AttributeDriverTest.php
IonBazan: Good idea, tracked as #2357

IonBazan:
@malarzm @alcaeus @franmomu @beberlei @Ahummeling Thank you for your inputs, I feel like we are close to make it final! Can you do another round just to make sure it's all working properly?

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AnnotationDriver.php
```diff
@@ -218,11 +218,9 @@ public function loadMetadataForClass($className, \Doctrine\Persistence\Mapping\C
                 }
 
                 if ($annot instanceof ODM\Indexes) {
-                    // Setting the type to mixed is a workaround until https://github.com/doctrine/annotations/pull/209 is released.
-                    /** @var mixed $value */
                     $value = $annot->value;
                     foreach (is_array($value) ? $value : [$value] as $index) {
-                        $indexes[] = $index;
+                        $this->addIndex($metadata, $index);
```
alcaeus: What's the reason for calling `addIndex` here? Indexes are added in L243 and below, so I'm wondering if this change is correct.

lib/Doctrine/ODM/MongoDB/Types/DateImmutableType.php
```diff
@@ -29,6 +29,7 @@ public static function getDateTime($value): DateTimeInterface
             return DateTimeImmutable::createFromMutable($datetime);
         }
 
+        // @phpstan-ignore-next-line
```
alcaeus: IIRC this was only necessary in 2.2.x but was removed in 2.3.x. @franmomu can you verify please?

alcaeus (commented):
Changes LGTM, only a question about the handling of the `@Indexes` annotation which seems to have changed slightly.

alcaeus:
@IonBazan the test failure on a sharded cluster seems unrelated - this might be a case of us not using the correct write concern for those tests on a sharded cluster. I'll investigate separately but have kicked the build to see if it passes now.

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AnnotationDriver.php
IonBazan: Good catch, the change was incorrect indeed. I changed it while experimenting with `@Indexes` annotation and accidentally left it here. The reason it should be added to `$indexes` array is that it's being normalized below according to field name.

lib/Doctrine/ODM/MongoDB/Types/DateImmutableType.php
franmomu: ah right, it's because it was removed from [`phpstan-baseline.neon`](https://github.com/doctrine/mongodb-odm/pull/2344/files#diff-995edee38ad4f8387e58ebd52c31bcc04c56cc2448d331b1cf5e0b35c57b9efa), can you please revert that file and this line won't be needed.

IonBazan: Reverted, thanks!

alcaeus (approved):
LGTM. Great work @IonBazan, this is very much appreciated! 💚

alcaeus:
@franmomu @malarzm happy to merge this if one of you gives it another seal of approval :)

franmomu (approved):
Awesome work! thank you! 🚀

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/AlsoLoad.php
malarzm: Ah right, silly me :)

malarzm:
@IonBazan one last thing: could you please rebase your branch atop current `2.3.x`? I'd like to al least get rid of merge commits from the history, also there were commits for CI that are no longer needed. Ideally the commit history should make sense on its own, but probably it's too late now so we may squash everything together :)

IonBazan:
@malarzm I managed to squash and edit some of the commits to get rid the ones that have been reverted so the history makes more sense now. I kept co-author info where @franmomu helped. Let me know if it needs further squashing.

alcaeus:
Thanks @IonBazan! Great work here!

PR #2345, opened 2021-07-23
Ignore phpstan error

franmomu:
It appeared in https://github.com/doctrine/mongodb-odm/pull/2344/checks?check_run_id=3141873310

https://github.com/doctrine/mongodb-odm/blob/e2c6dc41b8b52441eb69d0d79697f3fcc435561e/lib/Doctrine/ODM/MongoDB/Types/DateImmutableType.php#L20-L37

`parent::getDateTime($value);` returns a `DateTimeInterface`, since we are checking against `DateTimeImmutable` and `DateTime` and those are the only implementations of `DateTimeInterface`, phpstan reports the `throw new RuntimeException` as unreachable.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -233,3 +233,9 @@ parameters:
             message: "#^Method Doctrine\\\\ODM\\\\MongoDB\\\\UnitOfWork\\:\\:getById\\(\\) should return T of object but returns object\\.$#"
             count: 1
             path: lib/Doctrine/ODM/MongoDB/UnitOfWork.php
+
+        # That statement is never reached because DateTimeInterface is either DateTimeImmutable or DateTime
+        -
+            message: "#^Unreachable statement \\- code above always terminates\\.$#"
+            count: 1
+            path: lib/Doctrine/ODM/MongoDB/Types/DateImmutableType.php


Steveb-p:
Isn't it a correct issue reported by phpstan? It's not possible afaik to have an implementation different than those two?

> DateTimeInterface is meant so that both DateTime and DateTimeImmutable can be type hinted for. It is not possible to implement this interface with userland classes.

https://www.php.net/manual/en/class.datetimeinterface.php

franmomu:
> Isn't it a correct issue reported by phpstan? It's not possible afaik to have an implementation different than those two?

As stated in the documentation:

> It is not possible to implement this interface with userland classes.

So users cannot provide an implementation. With php 8 we can replace the `ifs` with:

```php
return DateTimeImmutable::createFromInterface($datetime); 
```

and remove the exception.

Steveb-p:
Or an `else` can be used instead of the second condition.

That exception is impossible to reach because parent would throw a type error due to type mismatch if anything other than `DateTimeInterface` is returned. Only two classes can be used, so it's either one or the other.

franmomu:
> Or an `else` can be used instead of the second condition.

yes, that can be done too, but following https://github.com/doctrine/mongodb-odm/pull/2118#discussion_r350459241 I would say it is fine the way it is written right now.

malarzm (approved):
> Or an else can be used instead of the second condition.

IIRC our coding standards are forbidding `else` statements favouring early returns over them.

malarzm:
Thanks @franmomu!

PR #2347, opened 2021-07-24
add @QueryResultDocument annotation reference

IonBazan:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | -

#### Summary

The `@QueryResultDocument` was missing from the Annotation Reference doc page. It has been documented on Aggregation Builder page only. Targeting `1.3.x` branch because it seems as oldest supported version.

BTW the Annotation Reference page seems to be missing from the sidebar in v2.2.x, even though it's defined in `sidebar.rst` file 🤔 .

### DIFF
--- a/docs/en/reference/annotations-reference.rst
+++ b/docs/en/reference/annotations-reference.rst
@@ -1266,7 +1266,7 @@ These documents can use all features regular documents can use but they will not
     $builder
         ->hydrate(\Documents\CompanyTransactions::class)
         ->group()
-            ->field('id')
+            ->field('_id')
             ->expression('$company')
             ->field('totalAmount')
             ->sum('$amount');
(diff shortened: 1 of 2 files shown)


docs/en/reference/annotations-reference.rst
```diff
@@ -1240,6 +1240,39 @@ the method to be registered.
...
+
+        /** @Field(type="float") */
+        private $totalAmount;
+    }
+
+    $builder = $dm->createAggregationBuilder(\Documents\Transaction::class);
+    $builder
+        ->hydrate(\Documents\CompanyTransactions::class)
+        ->group()
+            ->field('id')
```
alcaeus: I'd use the database name here, as defining a new field based on existing mappings can be risky: it will fail if the _id field is mapped to a different property.
```suggestion
            ->field('_id')
```

alcaeus (approved):

alcaeus:
Thanks @IonBazan!

PR #2350, opened 2021-08-02
Merge 2.2.x into 2.3.x

alcaeus:
This should also allow me to finally fix the required build config for protected branches.


psalm.xml
```diff
@@ -6,6 +6,7 @@
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xmlns="https://getpsalm.org/schema/config"
     xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd"
+    errorBaseline="psalm-baseline.xml"
```
franmomu: 🤔 I think I fixed these `ParamNameMismatch` some time ago (in `2.3.x`), probably this is not needed.

This is what I don't like about ignoring issues in the psalm baseline, that it doesn't complain about non existing ignored issues.

alcaeus: I'll recreate the baseline and see what happens.

alcaeus: See https://github.com/doctrine/mongodb-odm/pull/2350/commits/f995aadc4dac6adfdb1f60fdb37fdc76a06c18a1 - this now has an empty baseline. Do you want me to drop the file and config?

franmomu: I would remove it if it is not needed and add it when we need it.

franmomu (approved):

PR #2351, opened 2021-08-02
Fix missing hint query option during find

### LINKED ISSUE #2346: Index hints not being passed from QueryBuilder to final Find
### BC Break Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.2.0

#### Summary

The database for our project is actually AWS's DocumentDB service.  I will say off the bat that I do understand that it's not technically MongoDB, and there are some feature differences - however, we recently upgraded our Symfony project from v 1.1 to v2.0 (technically we upgraded doctrine/mongo-odm-bundle from ^3.5 to ^4.3) and things were working much better before the upgrade.

We are experiencing very long query times on some particular queries, that are definitely covered by indexes.  Performance logging in AWS indicates that Document DB is choosing to not use any indexes for the query, and so in this case the AWS documentation recommends using index hints on the queries to force DocumentDB to use the index. 

#### Previous behavior

Before we upgraded, the hints were being added to the find query

#### Current behavior

Index hints are not passing through to the query.

#### How to reproduce

I'm actually not totally sure how to describe how to reproduce the behavior, but I've identified where in the code this is going wrong. 

First, we use DoctrineQueryBuilder->hint() method to create a query, and add an index hint.   
https://github.com/doctrine/mongodb-odm/blob/2.2.x/lib/Doctrine/ODM/MongoDB/Query/Builder.php#L782

In that MongoDB/Query/Query object that is created, the hint is definitely present, but when it comes time to pass the data to the collection->find(), the hint is not added to the $queryOptions
https://github.com/doctrine/mongodb-odm/blob/2.2.x/lib/Doctrine/ODM/MongoDB/Query/Query.php#L427

For some reason the hint IS being added to count queries later in that same switch, so I'm not sure if it's being intentionally left off of the finds:
https://github.com/doctrine/mongodb-odm/blob/2.2.x/lib/Doctrine/ODM/MongoDB/Query/Query.php#L508

In my test server, if I manually add 'hint' to line#427 above, my page load time goes from around 14803ms to 505ms.

TimoBakx:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixes | #2346

#### Summary
This PR fixes the missing `hint` query option during a `find` query.

I added two tests: one to make sure the hint gets correctly propagated from the query array to the find method, and one to make sure the hint is correctly carried from the query builder to the query.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Query/Query.php
+++ b/lib/Doctrine/ODM/MongoDB/Query/Query.php
@@ -424,7 +424,7 @@ final class Query implements IteratorAggregate
 
         switch ($this->query['type']) {
             case self::TYPE_FIND:
-                $queryOptions = $this->getQueryOptions('select', 'sort', 'skip', 'limit', 'readPreference');
+                $queryOptions = $this->getQueryOptions('select', 'sort', 'skip', 'limit', 'readPreference', 'hint');
                 $queryOptions = $this->renameQueryOptions($queryOptions, ['select' => 'projection']);
 
                 $cursor = $this->collection->find(
--- a/tests/Doctrine/ODM/MongoDB/Tests/Query/BuilderTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Query/BuilderTest.php
@@ -321,6 +321,18 @@ class BuilderTest extends BaseTest
         $this->assertEquals($expected, $qb->getQueryArray());
     }
 
+    public function testFindWithHint(): void
+    {
+        $qb = $this->getTestQueryBuilder()
+            ->find(User::class)
+            ->hint('foo');
+
+        $expected = 'foo';
+
+        $this->assertEquals($expected, $qb->debug('hint'));
+        $this->assertEquals($expected, $qb->getQuery()->debug('hint'));
+    }
+
     public function testUpsertUpdateQuery()
     {
         $qb = $this->getTestQueryBuilder()
--- a/tests/Doctrine/ODM/MongoDB/Tests/QueryTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/QueryTest.php
@@ -490,6 +490,30 @@ class QueryTest extends BaseTest
         $this->assertSame(100, $query->execute());
     }
 
+    public function testFindWithHint(): void
+    {
+        $cursor = $this->createMock(Traversable::class);
+
+        $collection = $this->getMockCollection();
+        $collection->expects($this->once())
+            ->method('find')
+            ->with(['foo' => 'bar'], ['hint' => 'foo'])
+            ->will($this->returnValue($cursor));
+
+        // Using QueryBuilder->find adds hint to the query array
+        $queryArray = [
+            'type' => Query::TYPE_FIND,
+            'query' => ['foo' => 'bar'],
+            'hint' => 'foo',
+        ];
+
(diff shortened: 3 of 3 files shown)


lib/Doctrine/ODM/MongoDB/Query/Query.php
```diff
@@ -424,7 +424,7 @@ private function runQuery()
 
         switch ($this->query['type']) {
             case self::TYPE_FIND:
-                $queryOptions = $this->getQueryOptions('select', 'sort', 'skip', 'limit', 'readPreference');
+                $queryOptions = $this->getQueryOptions('select', 'sort', 'skip', 'limit', 'readPreference', 'hint');
```
alcaeus, line 427: For now this fixes the bug, but I'm wondering if we should be more lenient with the options we accept. `find` take a lot more options than these few, and we may do well to allow people to utilise them. @jmikola any suggestions on this?

alcaeus (approved):
Code LGTM, and the PhpStan failure seems unrelated. @franmomu could you please help out here?

franmomu:
I've created https://github.com/doctrine/mongodb-odm/pull/2352 for the phpstan issue

lib/Doctrine/ODM/MongoDB/Query/Query.php
TimoBakx: I can work out a list of options from the Builder and Collection classes and write a bunch of tests for those in a separate PR.

alcaeus: Let's hold off on what Jeremy says - not sure we need to test for every single one of them, but I'm rather thinking to just pass everything that we know isn't from ODM on to the driver.

alcaeus:
Thanks @TimoBakx!

lib/Doctrine/ODM/MongoDB/Query/Query.php
TimoBakx: From what I can see, only `maxTimeMS` and `snapshot` are still missing. With `snapshot` being deprecated in the `Find` class, so I'm not sure if that should be added.

jmikola: @TimoBakx: I'm not sure what documentation you're referring to, but there are many more options supported by the extension and library:

 * https://www.php.net/manual/en/mongodb-driver-query.construct.php
 * https://docs.mongodb.com/php-library/current/reference/method/MongoDBCollection-find/

I think it makes sense for the builder to provide some API to add arbitrary options that may not be explicitly handled in ODM but are known to be supported in the extension/library. However, that would likely require some internal refactoring to the Query class, since many of these fields appear to be stuffed directly into the internal `$query` property. Handling this might require segregating options into `$query['options']` so we no longer need to rely on `getQueryOptions` to extract specific keys.

TimoBakx: @jmikola I looked at all the options that have their own method in the query `Builder` class (and are added to the `$this->query` array there and checked which of those were used in the `Find` class code and were also missing in the `$this->getQueryOptions` call (which means that the instructions to the `Builder` would be ignored).

Moving things to separate options would indeed be better, but I'm not sure how much of a BC break that would be. It might be easier to add the missing values first. 🙂

PR #2359, opened 2021-08-11
flip attribute driver constructor arguments

IonBazan:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

#### Summary

Because `AbstractDoctrineExtension` expects the `AttributeDriver` constructor to accept `$paths` as first argument, while `AnnotationDriver` expects `Reader` to be passed there, we need to shift the argument order to make it compatible.
https://github.com/symfony/symfony/blob/1d196159fde010590b857059b5219922031c284b/src/Symfony/Bridge/Doctrine/DependencyInjection/AbstractDoctrineExtension.php#L209-L212

While ORM's `AttributeDriver` simply skips that argument and does not allow to customize the reader, I decided to simply shift the argument order to maintain the compatibility with `AbstractDoctrineExtension` and allow to customize the reader.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Mapping/Driver/AnnotationDriver.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/Driver/AnnotationDriver.php
@@ -340,11 +340,7 @@ class AnnotationDriver extends AbstractAnnotationDriver
      */
     public static function create($paths = [], ?Reader $reader = null): AnnotationDriver
     {
-        if ($reader === null) {
-            $reader = new AnnotationReader();
-        }
-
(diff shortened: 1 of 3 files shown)


franmomu (approved):
I thought we had added the constructor in https://github.com/doctrine/mongodb-odm/pull/2344/files#r683695547

IonBazan:
@franmomu You are right, I think I reverted that change at some point as I thought it is not necessary but when I was trying to use it in a Symfony app, it turned out that we have to keep the constructor this way, at least for now 🤷🏻‍♂️

alcaeus:
Thanks @IonBazan!

PR #2362, opened 2021-09-02
Add some property typehints

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

This add some property type hints to `tests` directory,  I've changed a couple of things that I would point out in the code. Since there are a lot missing, I'll try to make manageable PRs.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Configuration.php
+++ b/lib/Doctrine/ODM/MongoDB/Configuration.php
@@ -87,11 +87,14 @@ class Configuration
      * @psalm-var array{
      *      autoGenerateHydratorClasses?: self::AUTOGENERATE_*,
      *      autoGeneratePersistentCollectionClasses?: self::AUTOGENERATE_*,
+     *      classMetadataFactoryName?: class-string<ClassMetadataFactory>,
      *      defaultCommitOptions?: CommitOptions,
-     *      defaultDocumentRepositoryClassName?: string,
+     *      defaultDocumentRepositoryClassName?: class-string<ObjectRepository>,
+     *      defaultGridFSRepositoryClassName?: class-string<GridFSRepository>,
+     *      defaultDB?: string,
      *      documentNamespaces?: array<string, string>,
      *      filters?: array<string, array{
-     *          class: string,
+     *          class: class-string,
      *          parameters: array<string, mixed>
      *      }>,
      *      hydratorDir?: string,
@@ -101,6 +104,7 @@ class Configuration
      *      persistentCollectionFactory?: PersistentCollectionFactory,
      *      persistentCollectionGenerator?: PersistentCollectionGenerator,
      *      persistentCollectionDir?: string,
+     *      persistentCollectionNamespace?: string,
      *      repositoryFactory?: RepositoryFactory
      * }
      */
@@ -402,11 +406,17 @@ class Configuration
         return $this->attributes['defaultDB'] ?? null;
     }
 
+    /**
+     * @psalm-param class-string<ClassMetadataFactory> $cmfName
+     */
     public function setClassMetadataFactoryName(string $cmfName): void
     {
         $this->attributes['classMetadataFactoryName'] = $cmfName;
     }
 
+    /**
+     * @psalm-return class-string<ClassMetadataFactory>
+     */
     public function getClassMetadataFactoryName(): string
     {
         if (! isset($this->attributes['classMetadataFactoryName'])) {
@@ -450,6 +460,9 @@ class Configuration
         ];
     }
 
+    /**
+     * @psalm-return class-string|null
+     */
     public function getFilterClassName(string $name): ?string
     {
         return isset($this->attributes['filters'][$name])
@@ -468,6 +481,8 @@ class Configuration
     }
 
     /**
+     * @psalm-param class-string<ObjectRepository> $className
+     *
      * @throws MongoDBException If not is a ObjectRepository.
      */
     public function setDefaultDocumentRepositoryClassName(string $className): void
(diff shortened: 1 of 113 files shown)


tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1294Test.php
```diff
@@ -43,8 +43,12 @@ class GH1294User
...
     public $id;
 
-    /** @ODM\Field(type="string") */
-    public $name = false;
+    /**
+     * @ODM\Field(type="string")
+     *
+     * @var string|null
+     */
+    public $name = '';
```
franmomu, line 51: `name` is always initialized:

https://github.com/doctrine/mongodb-odm/blob/15773364718445ce999cdf501d297b5be31e25af/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1294Test.php#L15-L21

tests/Doctrine/ODM/MongoDB/Tests/Persisters/PersistenceBuilderTest.php
```diff
@@ -76,7 +76,7 @@ public function testFindWithOrOnCollectionWithDiscriminatorMap(): void
 
         $sameCollection1->id   = $ids[0];
         $sameCollection1->name = 'First entry in SameCollection1';
-        $sameCollection1->test = 1;
+        $sameCollection1->test = 'test';
```
franmomu, line 79: It is used as `string`:

https://github.com/doctrine/mongodb-odm/blob/15773364718445ce999cdf501d297b5be31e25af/tests/Doctrine/ODM/MongoDB/Tests/Persisters/PersistenceBuilderTest.php#L55-L63

lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -402,11 +406,15 @@ public function getDefaultDB(): ?string
...
     }
 
+    /** @psalm-param class-string $cmfName */
     public function setClassMetadataFactoryName(string $cmfName): void
     {
         $this->attributes['classMetadataFactoryName'] = $cmfName;
     }
 
+    /**
+     * @psalm-return class-string
```
malarzm: Let's be consistent with multi-line vs single-lined PHPDocs :)

malarzm (commented):
I'm gonna trust you and all our tests with this :)

malarzm (approved):

lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -87,8 +87,11 @@ class Configuration
      * @psalm-var array{
      *      autoGenerateHydratorClasses?: self::AUTOGENERATE_*,
      *      autoGeneratePersistentCollectionClasses?: self::AUTOGENERATE_*,
+     *      classMetadataFactoryName?: class-string,
```
IonBazan: Let's try to limit allowed classes here. It appears that it should only accept `\Doctrine\ODM\MongoDB\Mapping\ClassMetadataFactory`. In fact, the class is marked as `final` so I can't imagine anyone extending it but you can still create your own implementation that adds `setDocumentManager` and `getDocumentManager` methods (there is no common interface for that).

lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -87,8 +87,11 @@ class Configuration
...
      *      defaultCommitOptions?: CommitOptions,
      *      defaultDocumentRepositoryClassName?: string,
+     *      defaultGridFSRepositoryClassName?: class-string,
```
IonBazan: This should implement `\Doctrine\ODM\MongoDB\Repository\GridFSRepository`

tests/Doctrine/ODM/MongoDB/Tests/Events/LifecycleListenersTest.php
```diff
@@ -68,6 +72,7 @@ public function testLifecycleListeners(): void
         $this->listener->called = [];
 
         $document = $dm->find(TestDocument::class, $test->id);
+        assert($document->embedded instanceof PersistentCollectionInterface);
```
IonBazan: Can we use `self::assertInstanceOf()` together with `phpstan-phpunit` plugin instead?

lib/Doctrine/ODM/MongoDB/Configuration.php
franmomu: done, I've used multiline for PHPDoc in methods (I've found a way to force this from phpcs), the single-lined one is forced by phpcs when only has one single line content.

lib/Doctrine/ODM/MongoDB/Configuration.php
franmomu: sure, I've used `class-string<ClassMetadataFactory>` and modified also `defaultGridFSRepositoryClassName` and `classMetadataFactoryName`, thanks!

tests/Doctrine/ODM/MongoDB/Tests/Events/LifecycleListenersTest.php
franmomu: yep, done.

lib/Doctrine/ODM/MongoDB/Configuration.php
IonBazan: Can you also apply that to the getters and setters where applicable?

IonBazan (approved):

lib/Doctrine/ODM/MongoDB/Configuration.php
franmomu: yeah, should be better now, thanks.

PR #2363, opened 2021-09-02
Add more missing property type hints

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

Follow-up of https://github.com/doctrine/mongodb-odm/pull/2362

Apart from tests, given we also have the `tests` directory analysed by psalm and phpstan, they check if these new values are right for the properties used.


malarzm (approved):
Amazing work 👍

PR #2368, opened 2021-09-14
Finish adding return type declarations to phpdoc

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

This is the last PR adding return type declarations (I'll create a couple for parameter type declarations), there is one change in `GraphLookup:: convertTargetFieldName`, this method is called in:

https://github.com/doctrine/mongodb-odm/blob/38fc4947c81f8bce6ac2224f7883d76684c36758/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/GraphLookup.php#L92-L106

and

https://github.com/doctrine/mongodb-odm/blob/38fc4947c81f8bce6ac2224f7883d76684c36758/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/GraphLookup.php#L123-L128

In both calls the argument is a `string`, so it will never be an `array`.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/AbstractBucket.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/AbstractBucket.php
@@ -57,9 +57,6 @@ abstract class AbstractBucket extends Stage
         return $this;
     }
 
-    /**
-     * {@inheritdoc}
-     */
     public function getExpression(): array
     {
         $stage = [
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/AbstractBucket.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/AbstractBucket.php
@@ -82,6 +82,11 @@ abstract class AbstractBucket extends Stage
      */
     abstract protected function getStageName(): string;
 
+    /**
+     * @param array|mixed|string $expression
+     *
+     * @return array|mixed|string
+     */
     private function convertExpression($expression)
     {
         if (is_array($expression)) {
--- a/lib/Doctrine/ODM/MongoDB/Query/Expr.php
+++ b/lib/Doctrine/ODM/MongoDB/Query/Expr.php
@@ -1436,6 +1436,8 @@ class Expr
      * returned unmodified.
      *
      * @param Expr|mixed $expression
+     *
+     * @return array|mixed
      */
     private static function convertExpression($expression, ClassMetadata $classMetadata)
     {
--- a/lib/Doctrine/ODM/MongoDB/Types/DateType.php
+++ b/lib/Doctrine/ODM/MongoDB/Types/DateType.php
@@ -84,9 +84,6 @@ class DateType extends Type implements Versionable
         return $datetime;
     }
 
-    /**
-     * {@inheritDoc}
-     */
     public function convertToDatabaseValue($value)
     {
         if ($value === null || $value instanceof UTCDateTime) {
@@ -98,9 +95,6 @@ class DateType extends Type implements Versionable
         return new UTCDateTime((int) $datetime->format('Uv'));
     }
 
-    /**
-     * {@inheritDoc}
-     */
     public function convertToPHPValue($value)
     {
         if ($value === null) {
--- a/lib/Doctrine/ODM/MongoDB/Types/DateType.php
+++ b/lib/Doctrine/ODM/MongoDB/Types/DateType.php
@@ -84,6 +84,9 @@ class DateType extends Type implements Versionable
         return $datetime;
     }
 
+    /**
+     * {@inheritDoc}
+     */
     public function convertToDatabaseValue($value)
     {
         if ($value === null || $value instanceof UTCDateTime) {
@@ -95,6 +98,9 @@ class DateType extends Type implements Versionable
         return new UTCDateTime((int) $datetime->format('Uv'));
     }
 
+    /**
+     * {@inheritDoc}
+     */
     public function convertToPHPValue($value)
     {
         if ($value === null) {
@@ -114,6 +120,11 @@ class DateType extends Type implements Versionable
         return 'if ($value === null) { $return = null; } else { $return = \\' . static::class . '::getDateTime($value); }';
     }
 
+    /**
+     * @param mixed $current
+     *
+     * @return DateTimeInterface
+     */
     public function getNextVersion($current)
     {
         return new DateTime();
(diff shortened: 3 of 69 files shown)


lib/Doctrine/ODM/MongoDB/Types/DateType.php
```diff
@@ -114,6 +120,11 @@ public function closureToPHP(): string
         return 'if ($value === null) { $return = null; } else { $return = \\' . static::class . '::getDateTime($value); }';
     }
 
+    /**
+     * @param mixed $current
+     *
+     * @return DateTimeInterface
```
franmomu, line 120: I had to use `DateTimeInterface` here since `DateImmutableType` overrides this method

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/AbstractBucket.php
```diff
@@ -82,6 +82,11 @@ abstract protected function getExtraPipelineFields(): array;
      */
     abstract protected function getStageName(): string;
 
+    /**
+     * @param array|mixed|string $expression
+     *
+     * @return array|mixed|string
```
IonBazan, line 85: Isn't `mixed` already containing `array` and `string`?

lib/Doctrine/ODM/MongoDB/Types/DateType.php
IonBazan: Are we able to add type hint here or PHP 7.2 won't support that?

IonBazan (approved):

lib/Doctrine/ODM/MongoDB/Types/DateType.php
franmomu: 🤔 that would mean break BC for someone overriding this method.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/AbstractBucket.php
franmomu: Yes, this is what PHPStorm suggested to add, I didn't dislike it since it explicitly shows the different types, but I can change it to only `mixed` if we agree to that.

IonBazan: That's what I thought - let's leave it this way.

lib/Doctrine/ODM/MongoDB/Iterator/PrimingIterator.php
```diff
@@ -44,6 +44,9 @@ public function toArray(): array
         return iterator_to_array($this);
     }
 
+    /**
+     * {@inheritDoc}
```
malarzm: why do we need these?

lib/Doctrine/ODM/MongoDB/Query/Expr.php
```diff
@@ -1436,6 +1436,8 @@ private function convertExpressions(array $query, ?ClassMetadata $classMetadata
      * returned unmodified.
      *
      * @param Expr|mixed $expression
+     *
+     * @return array
```
malarzm: This can be added as a return type as we're dealing with private function

lib/Doctrine/ODM/MongoDB/Iterator/PrimingIterator.php
franmomu: For the `inheritDoc` cases, I was getting these errors:

```
FILE: lib/Doctrine/ODM/MongoDB/Iterator/PrimingIterator.php
-----------------------------------------------------------------------------------------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
-----------------------------------------------------------------------------------------------------------------------------------------------------------
 47 | ERROR | Method \Doctrine\ODM\MongoDB\Iterator\PrimingIterator::current() does not have return type hint nor @return annotation for its return value.
-----------------------------------------------------------------------------------------------------------------------------------------------------------
```
I remember reading about phpcs that performs analysis by file, so it doesn't get that information from the parent, I'll see if there is a way to avoid this.

lib/Doctrine/ODM/MongoDB/Query/Expr.php
franmomu: I was wrong about the return type, it could be `mixed` as well 🤦 , thanks!

lib/Doctrine/ODM/MongoDB/Iterator/PrimingIterator.php
franmomu: https://github.com/slevomat/coding-standard/issues/1149 🤔 the only thing I can think of is that we can try to reach level 5 of phpstan that checks missing type declarations and remove this sniff.

malarzm: The good thing is that we're already on level 5 :) https://github.com/doctrine/mongodb-odm/blob/2.2.x/phpstan.neon.dist#L5

franmomu: [Level 6](https://phpstan.org/user-guide/rule-levels) sorry 😞

SenseException (approved):

lib/Doctrine/ODM/MongoDB/Iterator/PrimingIterator.php
malarzm: Dang, so close. Personally I'd rather not clutter the code with useless `inheritdoc` just to satisfy CS Fixer. How about merging all the other changes, still keep the rule as excluded, and work towards level 6? I'd also much appreciate @alcaeus input on this

franmomu: I don't like `inheritdoc` either, I'm fine keeping the rule excluded.  Should I remove all the `inheritdoc`? (apart from the ones added in this PR) 🤔

malarzm: 👍 for removing all `@inheritDocs`

franmomu:
friendly ping @malarzm if you could take a look, the `@inheritDoc` blocks have been removed 

PS: if you have more time (and feel like reviewing) you can also take a look at https://github.com/doctrine/mongodb-odm/pull/2369 😬

malarzm (approved):

malarzm:
Thanks @franmomu and sorry for keeping you waiting 🎉

PR #2369, opened 2021-09-14
Add parameter type hints to tests

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

I can split it in several PRs if it's necessary.

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/MatchStageTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/MatchStageTest.php
@@ -61,9 +61,6 @@ class MatchStageTest extends BaseTest
         $this->assertSame($stage, $stage->$method(...$args));
     }
 
-    /**
-     * @return array<array{string, mixed[]}>
-     */
     public function provideProxiedExprMethods(): array
     {
         return [
--- a/tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/MatchStageTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/MatchStageTest.php
@@ -42,7 +42,7 @@ class MatchStageTest extends BaseTest
     /**
      * @dataProvider provideProxiedExprMethods
      */
-    public function testProxiedExprMethods($method, array $args = []): void
+    public function testProxiedExprMethods(string $method, array $args = []): void
     {
         $expr = $this->getMockQueryExpr();
         $expr
@@ -61,6 +61,9 @@ class MatchStageTest extends BaseTest
         $this->assertSame($stage, $stage->$method(...$args));
     }
 
+    /**
+     * @return array<array{string, mixed[]}>
+     */
     public function provideProxiedExprMethods(): array
     {
         return [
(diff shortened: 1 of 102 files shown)


tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/MatchStageTest.php
```diff
@@ -40,9 +41,11 @@ public function testMatchFromBuilder(): void
     }
 
     /**
+     * @param Closure|array $args
+     *
      * @dataProvider provideProxiedExprMethods
      */
-    public function testProxiedExprMethods($method, array $args = []): void
+    public function testProxiedExprMethods(string $method, $args = []): void
```
SenseException: It doesn't look like there is a Closure in use. The previous `array` type seemed to be fitting. What do I miss?

franmomu: Nothing 😬 , I do missed this because of the name of the function is used in other tests, thanks!

SenseException (approved):

tests/Doctrine/ODM/MongoDB/Tests/Functional/AtomicSetTest.php
```diff
@@ -119,6 +121,9 @@ public function testAtomicCollectionUnset($clearWith): void
         $this->assertCount(0, $user->phonenumbers);
     }
 
+    /**
+     * @return array<array{array|ArrayCollection|null}>
```
malarzm: This looks like an overkill and I myself as a maintainer wouldn't like to write this, let alone contributors. Each array gets passed as arguments to a function that can be typed so I'd argue that's enough. Is there any way of opting out of defining types for data providers?

malarzm (changes requested):

tests/Doctrine/ODM/MongoDB/Tests/Functional/AtomicSetTest.php
franmomu: I was trying to add this information to also fix issues with phpcs sniff `SlevomatCodingStandard.TypeHints.ReturnTypeHint.MissingTraversableTypeHintSpecification` and fix issues with static analysis tools (phpstan level 6) where you should define the iterable type.

But I understand this may seem overkill, we could do the same here as the return types I guess, keeping the rule and for phpstan we could use something like https://github.com/phpstan/phpstan-phpunit/issues/63#issuecomment-592515871.

malarzm: That'd be great 🙏

franmomu: I think I already removed those phpdoc (I may have left some, I'll check tomorrow)

malarzm: Awesome, feel free to merge 👍

malarzm (approved):

IonBazan (approved):

PR #2372, opened 2021-09-28
support wildcard index mapping

### LINKED ISSUE #2371: Support wildcard indexes
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

Mongodb supports adding [wildcard Indexes](https://docs.mongodb.com/v4.4/core/index-wildcard/) but it is not possible to configure it via mongodb-odm mappings.

Using following mapping:
```xml
<indexes>
    <index name="field.$**">
        <key name="field_wildcard"/>
    </index>
</indexes>
```

Results in error `Line 27:0: Element '{http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping}index', attribute 'name': 'tenants.$**.configured' is not a valid value of the atomic type 'xs:NMTOKEN'.`

IonBazan:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | Fix #2371 

#### Summary

This enables us to map indexes with `field.$**` name.

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/Mapping/Driver/fixtures/WildcardIndexDocument.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Mapping/Driver/fixtures/WildcardIndexDocument.php
@@ -1,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+namespace TestDocuments;
+
+class WildcardIndexDocument
+{
+    protected $id;
+}
--- a/doctrine-mongo-mapping.xsd
+++ b/doctrine-mongo-mapping.xsd
@@ -449,7 +449,7 @@
       <xs:element name="partial-filter-expression" type="odm:partial-filter-expression" minOccurs="0" />
     </xs:choice>
 
-    <xs:attribute name="name" type="xs:NMTOKEN" />
+    <xs:attribute name="name" type="xs:string" />
     <xs:attribute name="drop-dups" type="xs:boolean" />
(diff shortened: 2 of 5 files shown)


alcaeus (approved):

alcaeus:
We could add a mapping test to see if the wildcard index can be correctly mapped in both annotations and XML mappings. Other than that, I'm not sure if we even need to make any logic changes.

tests/Doctrine/ODM/MongoDB/Tests/Mapping/Driver/fixtures/WildcardIndexDocument.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+namespace TestDocuments;
+
+class WildcardIndexDocument
+{
+    protected $id;
```
franmomu, line 9:
```suggestion
    private $id;
```

franmomu (approved):

tests/Doctrine/ODM/MongoDB/Tests/Mapping/Driver/fixtures/WildcardIndexDocument.php
IonBazan: PHPCS is reporting this property as unused when it's private so I will just revert this change. That explains why all other models in this namespace use `protected` properties too.

PR #2374, opened 2021-10-05
Deprecate document namespace short aliases

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | deprecations
| BC Break     | no
| Fixed issues | 

#### Summary

This PR deprecates the usage of document namespace short aliases as in https://github.com/doctrine/orm/issues/8818

I haven't found any usage in docs.


IonBazan (approved):

SenseException (approved):

lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -140,6 +141,13 @@ public function addDocumentNamespace(string $alias, string $namespace): void
      */
     public function getDocumentNamespace(string $documentNamespaceAlias): string
     {
+        Deprecation::trigger(
```
malarzm, line 144: Definitely late to the party but are we changing the way how deprecations are being triggered? Asking since we tend to use `trigger_deprecation` before: https://github.com/doctrine/mongodb-odm/search?q=trigger_deprecation

PR #2378, opened 2021-10-10
Avoid PHP 8.1 deprecations with ReturnTypeWillChange

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

~We should probably add `#[\ReturnTypeWillChange]` in some methods, let's see what happens~

Added some `ReturnTypeWillChange`

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Iterator/PrimingIterator.php
+++ b/lib/Doctrine/ODM/MongoDB/Iterator/PrimingIterator.php
@@ -7,6 +7,7 @@ namespace Doctrine\ODM\MongoDB\Iterator;
 use Doctrine\ODM\MongoDB\Mapping\ClassMetadata;
 use Doctrine\ODM\MongoDB\Query\ReferencePrimer;
 use Doctrine\ODM\MongoDB\UnitOfWork;
+use ReturnTypeWillChange;
 
 use function is_callable;
 use function iterator_to_array;
@@ -61,6 +62,10 @@ final class PrimingIterator implements Iterator
         return iterator_to_array($this);
     }
 
+    /**
+     * @return TValue|null
+     */
+    #[ReturnTypeWillChange]
     public function current()
     {
         $this->primeReferences();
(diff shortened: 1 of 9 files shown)


franmomu:
https://github.com/doctrine/mongodb-odm/blob/8e45176ac1ca4f01bd00c1c26596e68755621e57/tests/Doctrine/ODM/MongoDB/Tests/PersistentCollectionTest.php#L130

[was failing with php 8.1](https://github.com/doctrine/mongodb-odm/pull/2378/checks?check_run_id=3852784886)

https://github.com/doctrine/mongodb-odm/blob/8e45176ac1ca4f01bd00c1c26596e68755621e57/lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionTrait.php#L205

was returning `true`.

My knowledge about internals is quite limited, I guessed is something related with not defining the object in a variable, so the reference gets lost when serializing and unserializing, but I don't really know.

~Apart from this I don't know why there isn't any deprecation warning about not defining declaration types or adding `#[\ReturnTypeWillChange]` for implemented core functions like [these](https://github.com/doctrine/mongodb-odm/blob/8e45176ac1ca4f01bd00c1c26596e68755621e57/lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionTrait.php#L611-L667).~

~🤔~

UPDATE: it wasn't configured the error level

lib/Doctrine/ODM/MongoDB/Iterator/PrimingIterator.php
```diff
@@ -73,6 +78,10 @@ public function next(): void
         $this->iterator->next();
     }
 
+    /**
+     * @return TKey
```
IonBazan:
```suggestion
     * @return TKey|null
```

`key()` returns `null` on failure.

IonBazan (approved):
LGTM 🚀

PR #2375, opened 2021-10-07
Allow ^6.0 for symfony/*

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

Follow-up of https://github.com/doctrine/mongodb-odm/pull/2370 adding a job to test it: https://github.com/doctrine/mongodb-odm/pull/2375/checks?check_run_id=3824035674

PR #2380, opened 2021-10-14
Drop support of unsupported Symfony versions

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

Ref: https://github.com/doctrine/mongodb-odm/pull/2375#discussion_r728641746

I've looked for: `method_exists`, `interface_exists`, `symfony 3.4` and I haven't found anything, so I guess there are no BC layers.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,9 +35,9 @@
         "jean85/pretty-package-versions": "^1.3.0 || ^2.0.1",
         "mongodb/mongodb": "^1.2.0",
         "psr/cache": "^1.0 || ^2.0 || ^3.0",
-        "symfony/console": "^3.4 || ^4.1 || ^5.0",
+        "symfony/console": "^4.4 || ^5.3",
         "symfony/deprecation-contracts": "^2.2",
-        "symfony/var-dumper": "^3.4 || ^4.1 || ^5.0"
+        "symfony/var-dumper": "^4.4 || ^5.3"
     },
     "require-dev": {
         "ext-bcmath": "*",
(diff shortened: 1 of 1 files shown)


IonBazan:
Initially I proposed that in #2375 but now when I think about it, ORM is supporting all versions from 3.0 to 6.0 without any maintenance overhead as Doctrine does not touch any Console nor VarDumper features that have changed recently. Since this is not a Symfony Bundle but a library, providing support for wide range of dependencies' versions is desired.

franmomu:
I'm fine with that. Checking this, I've realised that we are not testing those versions though.

`phpbench/phpbench` requires symfony/console ^4.2: https://github.com/phpbench/phpbench/blob/b1ea39a7a3cf996065638fd416f3653bcb5a5aa3/composer.json#L26

`symfony/cache` 4.4 requires `symfony/var-dump` ^4.2: https://github.com/symfony/cache/blob/8810b1addbd271245af891f5ffca90a5ab5eb8f2/composer.json#L31

IonBazan:
@franmomu should we use the `vendor-bin` plugin then? Or maybe try allowing super old versions of these tools like ORM did: https://github.com/doctrine/orm/blob/416aa1d2d71c8d1796781c3f57b17ba26debde26/composer.json#L41

franmomu:
In https://github.com/doctrine/mongodb-odm/pull/2375 I've removed `phpbench/phpbench` when executing PHPUnit tests that allows to use Symfony 6 and console 3.4

PR #2383, opened 2021-10-28
add missing type declarations

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

We can individually set up some checks to force type declarations (extracted from https://github.com/phpstan/phpstan-src/blob/master/conf/config.level6.neon)

### DIFF
--- a/benchmark/Document/HydrateDocumentBench.php
+++ b/benchmark/Document/HydrateDocumentBench.php
@@ -35,7 +35,7 @@ final class HydrateDocumentBench extends BaseBench
     /** @var HydratorInterface */
     private static $hydrator;
 
-    public function init()
+    public function init(): void
     {
         self::$data = [
             '_id' => new ObjectId(),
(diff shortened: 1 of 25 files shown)


IonBazan (approved):
Nice 👍🏻 Are we targeting to get level 6 in v2.3.0? If so, can you create an issue with checkboxes representing each rule to track the progress on that?

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Out.php
```diff
@@ -37,6 +37,7 @@ public function getExpression(): array
     public function out(string $collection): Stage\Out
     {
         try {
+            // @phpstan-ignore-next-line if $from is not a class-string is handled by the try-catch
```
SenseException: The PHPStan issues should be put into the config file. Same with the other files.

SenseException (changes requested):

SenseException (approved):

franmomu:
merged in https://github.com/doctrine/mongodb-odm/pull/2385

PR #2385, opened 2021-11-02
Use PHPStan 1

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

It's made on top of https://github.com/doctrine/mongodb-odm/pull/2383, once that is merge I'll rebase this one


lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -1356,7 +1363,7 @@ public function getCollection(): string
      * Sets the collection this Document is mapped to.
      *
      * @param array|string $name
-     * @psalm-param array{name: string, capped?: bool, size?: int, max?: int}|string $name
+     * @psalm-param array{name?: string, capped?: bool, size?: int, max?: int}|string $name
```
malarzm: Why this? `name` is actually required to be present in the array :)

lib/Doctrine/ODM/MongoDB/Repository/DocumentRepository.php
```diff
@@ -243,6 +243,7 @@ public function matching(Criteria $criteria): ArrayCollection
         }
 
         // @TODO: wrap around a specialized Collection for efficient count on large collections
+        /** @var Iterator<int, T> $iterator */
         $iterator = $queryBuilder->getQuery()->execute();
         assert($iterator instanceof Iterator);
```
malarzm, line 248: I guess the `assert` can be removed now?

tests/Doctrine/ODM/MongoDB/Tests/Functional/DiscriminatorsDefaultValueTest.php
```diff
@@ -185,28 +185,28 @@ class ParentDocumentWithoutDiscriminator extends ParentDocument
...
-     * @var ChildDocumentWithDiscriminator|null
+     * @var ChildDocumentWithDiscriminator
      */
     protected $referencedChild;
 
     /**
      * @ODM\ReferenceMany(targetDocument=ChildDocumentWithoutDiscriminator::class)
      *
-     * @var Collection<int, ChildDocumentWithDiscriminator>
+     * @var Collection<int, ChildDocumentWithDiscriminator>|array<ChildDocumentWithDiscriminator>
```
malarzm, line 195: Personally I'm more for `ChildDocumentWithDiscriminator[]`. How do we tend to use these across the repository?

malarzm (approved):

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
franmomu: It was because of the `isset` check below, I've added `treatPhpDocTypesAsCertain` option as `false` which seems cleaner.

tests/Doctrine/ODM/MongoDB/Tests/Functional/DiscriminatorsDefaultValueTest.php
franmomu: I was using `[]` at first and then I changed them to this because I was getting (not in this case):

https://phpstan.org/r/3d620f33-b686-4cea-ad37-510b638b2845

Apparently that only happens when there is a default value of `[]` 🤔

lib/Doctrine/ODM/MongoDB/Repository/DocumentRepository.php
franmomu: yeah we can, I left it because if we reach level 6 of PHPStan or achieve to make query builder generic, we could get rid of the `@var` annotation, but we can remove it in the meanwhile.

tests/Doctrine/ODM/MongoDB/Tests/Functional/DiscriminatorsDefaultValueTest.php
malarzm: Weird, empty array should be good enough for anything array-ish :) Mind reporting it to PHPStan?

lib/Doctrine/ODM/MongoDB/Repository/DocumentRepository.php
malarzm: No need in this case :+1:

malarzm:
Thanks @franmomu!

tests/Doctrine/ODM/MongoDB/Tests/Functional/DiscriminatorsDefaultValueTest.php
franmomu: I was looking at this and I found in https://phpstan.org/writing-php-code/phpdoc-types#iterables

> These iterable rules are applied only when the Collection type isn’t generic. When it’s generic, generics rules for class-level type variables are applied.

> If PHP encounters Collection|Foo[], two possible paths are taken:

> Collection implements Traversable so Collection|Foo[] is interpreted as a Collection object that iterates over Foo. The array part isn’t applied.
> Collection does not implement Traversable so Collection|Foo[] is interpreted as a Collection object or an array of Foo objects.
> If Collection|Foo[] means "Collection or array" in your case even if Collection implements Traversable, you need to disambiguate the type by using `Collection|array<Foo>` instead.

So this last paragraph explains this case, good to know.

PR #2386, opened 2021-11-03
Remove key template for iterators

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

I think there is no need for specifying the key as template when using iterators, otherwise we (and users) always have to use: `Iterator<key, User>` instead of `Iterator<User>`.

I've used `mixed` as key because of [`Iterator::key` definition](https://www.php.net/manual/en/iterator.key.php). In the tests directory, all the occurrences I've found are using `int`.


malarzm (approved):

malarzm:
Thanks @franmomu!

PR #2388, opened 2021-11-05
Use array_diff_key to get diff for new or deleted documents

### LINKED ISSUE #2387: Doctrine deletes embedded documents
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 1.0.3-2.2.3

#### Current behavior

In current [PersistentCollectionTrait::getDeletedDocuments](https://github.com/doctrine/mongodb-odm/blob/2.2.2/lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionTrait.php#L256) implementation array_udiff used, and in some cases it returns wrong results so doctrine removes valid documents from collection.

#### How to reproduce
It is not so easy to reproduce with doctrine code with test cases because it uses spl_object_hash, but here is example with real object hashes from real application
```php
<?php

$arr1 = [
"000000001602268d0000000049345143",
"00000000160224e90000000049345143",
"000000001602235e0000000049345143",
"00000000160223480000000049345143",
"00000000160223ba0000000049345143",
"00000000160223b40000000049345143",
"00000000160223a60000000049345143",
"00000000160223900000000049345143",
"00000000160223820000000049345143",
"00000000160223fc0000000049345143",
"00000000160223ee0000000049345143",
"00000000160223d80000000049345143",
"00000000160223ca0000000049345143",
"00000000160223c40000000049345143",
"00000000160222360000000049345143",
"00000000160222200000000049345143",
"00000000160222120000000049345143",
"000000001602220c0000000049345143",
"000000001602227e0000000049345143",
"00000000160222680000000049345143",
"000000001602225a0000000049345143",
"00000000160222540000000049345143",
"00000000160222460000000049345143",
"00000000160222b00000000049345143",
"00000000160222a20000000049345143",
"000000001602229c0000000049345143",
"000000001602228e0000000049345143",
"00000000160222f80000000049345143",
"00000000160222ea0000000049345143",
"00000000160222e40000000049345143",
"00000000160222d60000000049345143",
"00000000160222c00000000049345143",
"00000000160221320000000049345143",
"000000001602212c0000000049345143",
"000000001602211e0000000049345143",
"00000000160221080000000049345143",
"000000001602217a0000000049345143",
"00000000160221740000000049345143",
"00000000160221660000000049345143",
"00000000160221500000000049345143",
];

$arr2 = $arr1;
$arr2[] = "00000000160220a30000000049345143";

var_dump(array_diff(
    $arr1,
    $arr2
));

var_dump(array_udiff(
    $arr1,
    $arr2,
    function ($a, $b) {
        return $a === $b ? 0 : ($a > $b ? 1 : -1);
    }
));

var_dump(array_udiff(
    $arr1,
    $arr2,
    function ($a, $b) {
        return $a <=> $b;
    }
));
```

Result:
```
array(0) {
}
array(2) {
  [14]=>
  string(32) "00000000160222360000000049345143"
  [15]=>
  string(32) "00000000160222200000000049345143"
}
array(2) {
  [14]=>
  string(32) "00000000160222360000000049345143"
  [15]=>
  string(32) "00000000160222200000000049345143"
}

```
Here we execute snippet: https://sandbox.onlinephpfunctions.com/code/9476593e43d067fe72cf8e05f9b993112ce7e15b

#### Expected behavior

Doctrine should not delete valid objects from collection

#### Possible solutions:

We can use some other methods to get documents diff. For example in doctrine ORM 
array_diff_key is used https://github.com/doctrine/orm/blob/2.10.x/lib/Doctrine/ORM/PersistentCollection.php#L695
or we can try to use strcmp to compare two hashes: https://sandbox.onlinephpfunctions.com/code/6f3a86594e0e86dba6199e45392a4bbfb7e70cd3

pepeh:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | fixes #2387

#### Summary

Avoid of array_udiff usage for getDeletedDocuments and getInsertedDocuments functions

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionTrait.php
+++ b/lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionTrait.php
@@ -256,10 +256,10 @@ trait PersistentCollectionTrait
     public function getDeletedDocuments()
     {
         $coll = $this->coll->toArray();
-        $loadedObjectsByOid = \array_combine(\array_map('spl_object_hash', $this->snapshot), $this->snapshot);
-        $newObjectsByOid = \array_combine(\array_map('spl_object_hash', $coll), $coll);
+        $loadedObjectsByOid = array_combine(array_map('spl_object_hash', $this->snapshot), $this->snapshot);
+        $newObjectsByOid = array_combine(array_map('spl_object_hash', $coll), $coll);
 
-        return array_values(\array_diff_key($loadedObjectsByOid, $newObjectsByOid));
+        return array_values(array_diff_key($loadedObjectsByOid, $newObjectsByOid));
     }
 
     /** {@inheritdoc} */
@@ -278,10 +278,10 @@ trait PersistentCollectionTrait
     public function getInsertedDocuments()
     {
         $coll = $this->coll->toArray();
-        $loadedObjectsByOid = \array_combine(\array_map('spl_object_hash', $this->snapshot), $this->snapshot);
-        $newObjectsByOid = \array_combine(\array_map('spl_object_hash', $coll), $coll);
+        $loadedObjectsByOid = array_combine(array_map('spl_object_hash', $this->snapshot), $this->snapshot);
+        $newObjectsByOid = array_combine(array_map('spl_object_hash', $coll), $coll);
 
-        return array_values(\array_diff_key($newObjectsByOid, $loadedObjectsByOid));
+        return array_values(array_diff_key($newObjectsByOid, $loadedObjectsByOid));
     }
 
     /** {@inheritdoc} */
--- a/lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionTrait.php
+++ b/lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionTrait.php
@@ -11,13 +11,14 @@ use Doctrine\ODM\MongoDB\MongoDBException;
 use Doctrine\ODM\MongoDB\UnitOfWork;
 use Doctrine\ODM\MongoDB\Utility\CollectionHelper;
 
-use function array_udiff;
+use function array_combine;
+use function array_diff_key;
+use function array_map;
 use function array_udiff_assoc;
 use function array_values;
 use function count;
 use function get_class;
 use function is_object;
-use function spl_object_hash;
 
 /**
  * Trait with methods needed to implement PersistentCollectionInterface.
@@ -255,9 +256,9 @@ trait PersistentCollectionTrait
     /** {@inheritdoc} */
     public function getDeletedDocuments()
     {
-        $coll = $this->coll->toArray();
+        $coll               = $this->coll->toArray();
         $loadedObjectsByOid = array_combine(array_map('spl_object_hash', $this->snapshot), $this->snapshot);
-        $newObjectsByOid = array_combine(array_map('spl_object_hash', $coll), $coll);
+        $newObjectsByOid    = array_combine(array_map('spl_object_hash', $coll), $coll);
 
         return array_values(array_diff_key($loadedObjectsByOid, $newObjectsByOid));
     }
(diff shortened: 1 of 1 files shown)


IonBazan:
@pepeh thanks for reporting and solving the problem. Can you add some tests to verify the new behaviour? Please also run `vendor/bin/phpcbf` to fix code style.

pepeh:
@IonBazan Thank you. I have fixed code style.

We already have test cases
[testGetDeletedDocuments](https://github.com/doctrine/mongodb-odm/blob/064f9052e6605d8af30556cc86997b434b6c5422/tests/Doctrine/ODM/MongoDB/Tests/PersistentCollectionTest.php#L137)
[testGetInsertedDocuments](https://github.com/doctrine/mongodb-odm/blob/064f9052e6605d8af30556cc86997b434b6c5422/tests/Doctrine/ODM/MongoDB/Tests/PersistentCollectionTest.php#L211)
and i'm not sure what can I add more.

IonBazan (approved):

lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionTrait.php
```diff
@@ -255,18 +257,15 @@ static function ($a, $b) {
     /** {@inheritdoc} */
     public function getDeletedDocuments()
     {
-        $compare = static function ($a, $b) {
-            $compareA = is_object($a) ? spl_object_hash($a) : $a;
-            $compareb = is_object($b) ? spl_object_hash($b) : $b;
-
-            return $compareA === $compareb ? 0 : ($compareA > $compareb ? 1 : -1);
+        $callback = static function ($val) {
+            return is_object($val) ? spl_object_hash($val) : $val;
```
franmomu: I'm wondering if it could be possible to be something different than `object`.

franmomu (approved):

lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionTrait.php
pelekhw2: getDeletedDocuments is used only on 1 place. I don't know why on old variant `is_object($val) ? spl_object_hash($val) : $val` instead just `spl_object_hash($val)` is used.

https://github.com/doctrine/mongodb-odm/blob/e01fe301d6fbda274001cfeaec60b2d642db2643/lib/Doctrine/ODM/MongoDB/UnitOfWork.php#L926

IonBazan: I wonder if we could also rely on `spl_object_id()` instead which is apparently slightly faster.

pepeh: It works in the same way, just returns numbers so we can use it.

```PHP
<?php
var_dump(spl_object_hash(new StdClass()), spl_object_hash(new StdClass()));
var_dump(spl_object_id(new StdClass()), spl_object_id(new StdClass()));

$a = new StdClass();
$b = new StdClass();

var_dump(spl_object_id($a), spl_object_id($b));

```
Result
```
string(32) "000000004e83240c000000005159bace"
string(32) "000000004e83240c000000005159bace"
int(1)
int(1)
int(1)
int(2)
```

pepeh: Should I change it?
And should I keep "is_object($val)" condition?

IonBazan: `PersistentCollection` shouldn't be used with anything that is not managed by UoW so I think we should be safe to skip that `is_object()` check and replace `spl_object_hash()` with `spl_object_id()` just like in ORM. 

This piece of code is really old (pre-PHP 7) and intended to be used only by UoW internally so this doesn't seem like a breaking change to me unless @malarzm or @alcaeus disagree.

pepeh: ok, changes are pushed

malarzm (approved):
IIRC the `is_object` check could come from some ooold times when PersistentCollection was meant to be used with other collection thingies. As it's only used for holding objects I think it's safe to drop it (but I'm still in vacation mode so I might be wrong 👼)

All in all, the change looks good to me 👍

IonBazan (approved):

IonBazan:
Thank you @pepeh! ❤️

Issue #2182, opened 2020-05-10
schema:create fails with single collection inheritance


malarzm:
### Bug Report

As per https://stackoverflow.com/questions/61615412/odm-inheritance-document-with-type-single-collection-schema-create-error the `odm:schema:create` fails with an error if the collection already exists. This is the case when single collection inheritance is used.

#### Expected behavior

The command finishes successfully.

Vladnev:
@alcaeus Hi it also actual for 2.2.0 version. As i see 2.1 Roadmap not supproted.

malarzm:
@Vladnev thanks for noticing! It must have somehow stay in 2.1 when we were retriaging issues :) Would you like to try fixing the issue since it seems to affect you as well?

malarzm:
Fixed with #2392

PR #2392, opened 2021-11-23
Schema:create fails with single collection inheritance

JJarrie:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #2182 

#### Summary

Maintain in memory already processed collection in case of single collection inheritance and skip next procession of these collections.

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/SchemaManagerTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/SchemaManagerTest.php
@@ -33,6 +33,7 @@ use PHPUnit\Framework\Constraint\Constraint;
 use PHPUnit\Framework\Constraint\IsEqual;
 use PHPUnit\Framework\MockObject\MockObject;
 
+use function array_count_values;
 use function array_map;
 use function assert;
 use function class_exists;
@@ -457,11 +458,15 @@ class SchemaManagerTest extends BaseTest
      */
     public function testCreateCollections(array $expectedWriteOptions, ?int $maxTimeMs, ?WriteConcern $writeConcern)
     {
-        foreach ($this->documentDatabases as $class => $database) {
+        $createdCollections = [];
+        foreach ($this->documentDatabases as $database) {
             $database
                 ->expects($this->atLeastOnce())
                 ->method('createCollection')
-                ->with($this->anything(), $this->writeOptions($expectedWriteOptions));
+                ->with($this->anything(), $this->writeOptions($expectedWriteOptions))
+                ->willReturnCallback(static function (string $collectionName) use (&$createdCollections): void {
+                    $createdCollections[] = $collectionName;
+                });
 
             $database
                 ->expects($this->atLeastOnce())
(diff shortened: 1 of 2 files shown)


malarzm (changes requested):
Thanks @JJarrie for the PR! :tada: The solution you've provided is simple, we're requiring PRs to have tests not only to prove the solution is working, but to prevent regressions in the future. Could you please write one? Don't hesitate to ask should you need any help :)

JJarrie:
> Thanks @JJarrie for the PR! tada The solution you've provided is simple, we're requiring PRs to have tests not only to prove the solution is working, but to prevent regressions in the future. Could you please write one? Don't hesitate to ask should you need any help :)

Hello, I need some help. I tried to add the same structure as the one who raised the error in the issue ticket.

```php
<?php

declare(strict_types=1);

namespace Documents\Functional;

/**
 * @ODM\MappedSuperclass
 * @ODM\DiscriminatorField("@type")
 * @ODM\DiscriminatorMap(SingleInheritanceCollectionSuperClass::TYPES)
 */
class SingleInheritanceCollectionSuperClass
{
    public const TYPES = [
        'subType1' => SingleInheritanceCollectionSubType1::class,
        'subType2' => SingleInheritanceCollectionSubType2::class,
    ];

    /** @ODM\Id */
    private $id;

    private $name;
}
```

```php
<?php

declare(strict_types=1);

namespace Documents\Functional;

/**
 * @ODM\Document
 */
class SingleInheritanceCollectionSubType1 extends SingleInheritanceCollectionSuperClass
{
    private $someFieldA;
}

```

```php
<?php

declare(strict_types=1);

namespace Documents\Functional;

/**
 * @ODM\Document
 */
class SingleInheritanceCollectionSubType2 extends SingleInheritanceCollectionSuperClass
{
    private $someFieldB;
}
```

And as the database is mocked in SchemaManagerTest::testCreateCollections, the original error (coming from database) is not raised. And I have some difficulties to understand how to handle my case into the functionnal suite. If someone have an idea or an explanation, I take it.

:)

IonBazan:
@JJarrie I've added a test which makes sure one of our existing documents using single collection inheritance trigger the collection creation only once.

JJarrie:
> @JJarrie I've added a test which makes sure one of our existing documents using single collection inheritance trigger the collection creation only once.

@IonBazan thank you for your help.

tests/Doctrine/ODM/MongoDB/Tests/SchemaManagerTest.php
```diff
@@ -470,6 +475,7 @@ public function testCreateCollections(array $expectedWriteOptions, ?int $maxTime
         }
 
         $this->schemaManager->createCollections($maxTimeMs, $writeConcern);
+        self::assertSame(1, array_count_values($createdCollections)['Tournament']);
```
malarzm, line 482: To prevent this test from becoming obsolete there should be an assertion in the beginning that Tournament has inheritance and is using single collection. Otherwise the change in existing entities can render this test useless

malarzm (approved):

malarzm:
Thanks @JJarrie!

PR #2394, opened 2021-11-30
Remove specific code for testing Symfony 6

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | task
| BC Break     | no
| Fixed issues | 

#### Summary

There is no need for this since it has been released and it will install Symfony 6 using `highest` dependencies.


malarzm (approved):

malarzm:
Thanks @franmomu!

PR #2398, opened 2021-12-21
PHP 8.1 types in custom collections

### LINKED ISSUE #2397: Handle intersection types in custom collections
Originally reported in https://github.com/doctrine/common/issues/948 and while collection generation was based off doctrine/common's proxy generation it'll need an update too.

malarzm:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | fixes #2397


franmomu (approved):
👍 nice

PR #2399, opened 2021-12-26
Skip non-mapped typed properties with default values when using Proxies

### LINKED ISSUE #2349: Cannot access uninitialized non-nullable property when accessing a non-db field property of a document
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.2.1

#### Summary
This happens on PHP >= 7.4 where the Document properties have been type hinted.

A fatal error is thrown when you access a property of a Proxy document where the property itself is NOT a field managed by the ODM, even if that property has a default value. This only seems to happen when the Proxy document has been loaded directly from the ODM (i.e. NOT from a __get call from accessing $document->someProperty).

#### Current behaviour
Throws a fatal error when property is accessed:
```
  [Error]                                                                                              
  Cannot access uninitialized non-nullable property App\Document\Customer::$domainEvents by reference  
```

#### How to reproduce
Reproduced on this repo: https://github.com/andythorne/doctrine-odm-example-error

Key bit copied from above repo. See it's [documents](https://github.com/andythorne/doctrine-odm-example-error/tree/main/src/Document) and [command to reproduce the error](https://github.com/andythorne/doctrine-odm-example-error/blob/main/src/Command/RunTestCommand.php)
```php
// The Order is loaded directly via the DM. The returned document is an Order object, as expected. The Order
// object has a customer property that contains a Proxy<Customer> object.
/** @var Order $order */
$order = $this->documentManager->getRepository(Order::class)->find($orderId);

// Fetch a list of Customers from the DB. Any customer object that was referenced in the above order is still
// a proxy object, however it will not have the defaults set for the un-managed $domainEvents property.
$customers = $this->documentManager->getRepository(Customer::class)->findAll();

/** @var Customer $customer */
foreach ($customers as $customer) {
    $customer->doSomeUpdate(); // This will trigger an error as we try to append to the non-existing $domainEvents property
}
```

#### Expected behaviour
Any properties should have their default values set from the Document class.

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | yes
| Fixed issues | #2349

#### Summary

Taking a look at #2349 seems like the problem is that the proxy is not initialized when it's managed. 

~Looking at the code, if the document is an instance of `GhostObjectInterface` and it not initialized, it calls to `$document->setProxyInitializer(null);` which removes the initialization.~

To be honest I'm not sure if it's the right fix, maybe that code was there for some reason.

The test is from https://github.com/andythorne/doctrine-odm-example-error

**Update**: I was taking a look again to this issue, apparently one way to solve it is to skip those properties when using a proxy.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Proxy/Factory/StaticProxyFactory.php
+++ b/lib/Doctrine/ODM/MongoDB/Proxy/Factory/StaticProxyFactory.php
@@ -14,9 +14,11 @@ use Doctrine\ODM\MongoDB\Utility\LifecycleEventManager;
 use Doctrine\Persistence\NotifyPropertyChanged;
 use ProxyManager\Factory\LazyLoadingGhostFactory;
 use ProxyManager\Proxy\GhostObjectInterface;
+use ReflectionClass;
 use ReflectionProperty;
 
 use function array_filter;
+use function array_merge;
 use function count;
 
 /**
@@ -131,6 +133,19 @@ final class StaticProxyFactory implements ProxyFactory
      */
     private function skippedFieldsFqns(ClassMetadata $metadata): array
     {
+        $skippedIdFieldsFqns        = $this->getIdFieldsFqns($metadata);
+        $skippedNonMappedFieldsFqns = $this->getUnmappedPropertyFqns($metadata);
+
+        return array_merge($skippedIdFieldsFqns, $skippedNonMappedFieldsFqns);
+    }
+
+    /**
+     * @param ClassMetadata<object> $metadata
+     *
+     * @return array<int, string>
+     */
+    private function getIdFieldsFqns(ClassMetadata $metadata): array
+    {
         $idFieldFqcns = [];
 
         foreach ($metadata->getIdentifierFieldNames() as $idField) {
@@ -138,6 +153,30 @@ final class StaticProxyFactory implements ProxyFactory
         }
 
         return $idFieldFqcns;
+    }
+
+    /**
+     * @param ClassMetadata<object> $metadata
+     *
+     * @return array<int, string>
+     */
+    private function getUnmappedPropertyFqns(ClassMetadata $metadata): array
+    {
+        $nonMappedFieldFqcns = [];
+
+        $reflectionClass = new ReflectionClass($metadata->getName());
+
+        foreach ($reflectionClass->getProperties() as $property) {
+            $propertyName = $property->getName();
+
+            if ($metadata->hasField($propertyName)) {
+                continue;
+            }
+
+            $nonMappedFieldFqcns[] = $this->propertyFqcn($property);
+        }
+
+        return $nonMappedFieldFqcns;
     }
 
     private function propertyFqcn(ReflectionProperty $property): string
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH2349Test.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH2349Test.php
@@ -1,0 +1,36 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ODM\MongoDB\Tests\Functional\Ticket;
+
+use Doctrine\ODM\MongoDB\Tests\BaseTestCase;
+use Documents74\GH2349Customer;
+use Documents74\GH2349Order;
+
+/** @requires PHP 7.4 */
+class GH2349Test extends BaseTestCase
+{
+    public function testAccessingUnitializedProperties(): void
+    {
+        $customer = new GH2349Customer('Some Place');
+        $order    = new GH2349Order($customer);
+
+        $this->dm->persist($customer);
+        $this->dm->persist($order);
+        $this->dm->flush();
+        $this->dm->clear();
+
+        $orderId = GH2349Order::ID;
+
+        $order = $this->dm->getRepository(GH2349Order::class)->find($orderId);
+
+        // Fetch a list of Customers from the DB. Any customer object that was referenced in the above order is still
+        // a proxy object, however it will not have the defaults set for the un-managed $domainEvents property.
+        $customers = $this->dm->getRepository(GH2349Customer::class)->findAll();
+
+        foreach ($customers as $customer) {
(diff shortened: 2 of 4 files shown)


lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -2839,7 +2839,7 @@ public function getOrCreateDocument(string $className, array $data, array &$hint
             $document = $this->identityMap[$class->name][$serializedId];
             $oid      = spl_object_hash($document);
             if ($document instanceof GhostObjectInterface && ! $document->isProxyInitialized()) {
-                $document->setProxyInitializer(null);
+                $document->initializeProxy();
```
malarzm: This seems to do the trick, but we shouldn't initialize the object at this point as this will lead to a db query while we already have data waiting for hydration. The correct fix would be to somehow initialize properties not managed by the ODM

malarzm (changes requested):

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
franmomu: Thanks @malarzm! that makes sense, I'll try to take a look how the ORM does this if I can. In the meantime I leave the test here just in case someone wants to give it a try

franmomu:
I was taking a look again at this issue and one possible solution would be to skip these properties.

I've skipped the typed non-mapped properties that have a default value. What about this? maybe is there something else to take into account? 🤔

carlos-granados:
@malarzm any chance that this might be merged? We are experiencing this issue with Doctrine 2.3 and were hoping to see it solved when we upgrade to 2.4

alcaeus:
2.4 is already released. I'll take a look at this after the weekend for inclusion in a 2.5 patch release.

alcaeus:
I made some adjustments to the tests to no longer rely on the absence of an error (and instead running an assertion), as well as applying the logic to all unmapped properties, not only those with default values.

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH2349Test.php
```diff
@@ -0,0 +1,36 @@
...
+        $orderId = GH2349Order::ID;
+
+        $order = $this->dm->getRepository(GH2349Order::class)->find($orderId);
+
+        // Fetch a list of Customers from the DB. Any customer object that was referenced in the above order is still
+        // a proxy object, however it will not have the defaults set for the un-managed $domainEvents property.
+        $customers = $this->dm->getRepository(GH2349Customer::class)->findAll();
+
+        foreach ($customers as $customer) {
+            self::assertSame([], $customer->getEvents()); // This would trigger an error if unmapped properties are unset
```
malarzm, line 33: This test seems fragile:

- we don't check whether `$customer` is indeed a proxy
- and is it still uninitialized (the test would still pass)

Maybe a db query count would be also good to assert so we never get into situation like the one found in original PR?

lib/Doctrine/ODM/MongoDB/Proxy/Factory/StaticProxyFactory.php
```diff
@@ -130,6 +132,19 @@ private function createInitializer(
      * @return array<int, string>
      */
     private function skippedFieldsFqns(ClassMetadata $metadata): array
+    {
+        $skippedIdFieldsFqns        = $this->getIdFieldsFqns($metadata);
+        $skippedNonMappedFieldsFqns = $this->getUnmappedPropertyFqns($metadata);
```
malarzm, line 137: This looks more expensive than the original identifier exclusion and is called each and every time we create a proxy object, pretty much a hot path. We should memoize the result of `skippedFieldsFqns` per `$metadata`

lib/Doctrine/ODM/MongoDB/Proxy/Factory/StaticProxyFactory.php
```diff
@@ -140,6 +155,30 @@ private function skippedFieldsFqns(ClassMetadata $metadata): array
...
+    private function getUnmappedPropertyFqns(ClassMetadata $metadata): array
+    {
+        $nonMappedFieldFqcns = [];
+
+        $reflectionClass = new ReflectionClass($metadata->getName());
+
+        foreach ($reflectionClass->getProperties() as $property) {
+            $propertyName = $property->getName();
+
+            if ($metadata->hasField($propertyName)) {
```
malarzm, line 172: Just a note as field with different PHP and DB field always confused me: metadata is operating on PHP's object properties, and DB field name is used only in hydrators/persisters, right? If yes then we're A-OK here

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH2349Test.php
```diff
@@ -0,0 +1,36 @@
...
+
+declare(strict_types=1);
+
+namespace Doctrine\ODM\MongoDB\Tests\Functional\Ticket;
+
+use Doctrine\ODM\MongoDB\Tests\BaseTestCase;
+use Documents74\GH2349Customer;
+use Documents74\GH2349Order;
+
+/** @requires PHP 7.4 */
```
malarzm, line 11: If I'm seeing correctly, 7.4 was already minimal version for ODM 2.5.x :) No harm done given we haven't cleaned this up yet, worth doing before 2.6.x :)

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH2349Test.php
malarzm: Also i believe this would be better placed in a generic proxy test, where we don't even need database operations. Just creating a proxy and asserting whether unmapped stuff can be used without initialization - this should be easier to maintain in future

lib/Doctrine/ODM/MongoDB/Proxy/Factory/StaticProxyFactory.php
malarzm: Or put into metadata itself?

alcaeus:
Fixed by #2698 - I had forgotten about this PR until after merging the other PR. Thanks for your efforts @franmomu!

PR #2407, opened 2022-02-14
Fix PHPStan issues

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | task
| BC Break     | no
| Fixed issues | 

#### Summary

It's nice to see some ignored errors go :)

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/GraphLookup.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/GraphLookup.php
@@ -22,25 +22,25 @@ use function substr;
 
 class GraphLookup extends Stage
 {
-    /** @var string */
+    /** @var string|null */
     private $from;
 
-    /** @var string|Expr|array */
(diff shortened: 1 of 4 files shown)


lib/Doctrine/ODM/MongoDB/Aggregation/Stage/GraphLookup.php
```diff
@@ -37,10 +37,10 @@ class GraphLookup extends Stage
     /** @var string */
     private $as;
 
-    /** @var int */
+    /** @var int|null */
```
malarzm, line 40: at a glance fields above should get `|null` as well, no?

franmomu: yep, I have to try to keep increasing static analysis level when I find some time

SenseException (approved):

malarzm (approved):

malarzm:
Thanks @franmomu!

PR #2408, opened 2022-02-16
Psalm level 5

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | task
| BC Break     | no
| Fixed issues | 

#### Summary

Replaces https://github.com/doctrine/mongodb-odm/pull/2329

Since it's quite difficult and time consuming to fix all possible issues found by static analysis, I think it's better if we use a baseline to ignore some of the issues for now and we set a higher level so new code has to comply with that level.

I've targeted `2.3.x` since there are some phpdoc fixed and it's not adding new features.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
@@ -571,7 +571,7 @@ class XmlDriver extends FileDriver
     }
 
     /**
-     * @return array<string, array<string, mixed>|scalar>
+     * @return array<string, array<string, mixed>|scalar|null>
      */
     private function getPartialFilterExpression(SimpleXMLElement $fields): array
     {
--- a/lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionTrait.php
+++ b/lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionTrait.php
@@ -428,6 +428,7 @@ trait PersistentCollectionTrait
 
     /**
      * @return Traversable
+     * @psalm-return Traversable<TKey, T>
      */
     #[ReturnTypeWillChange]
     public function getIterator()
@@ -540,6 +541,7 @@ trait PersistentCollectionTrait
      * @param mixed $offset
      *
      * @return mixed
+     * @psalm-return T|null
      */
     #[ReturnTypeWillChange]
     public function offsetGet($offset)
@@ -660,6 +662,11 @@ trait PersistentCollectionTrait
      * @param mixed $offset
      *
      * @return bool|T|null
+     * @psalm-return (
+     *      $arrayAccess is false
+     *      ? T|null
+     *      : T|null|true
+     * )
      */
     private function doRemove($offset, bool $arrayAccess)
     {
--- a/composer.json
+++ b/composer.json
@@ -48,7 +48,7 @@
         "phpunit/phpunit": "^8.5 || ^9",
         "squizlabs/php_codesniffer": "^3.5",
         "symfony/cache": "^4.4 || ^5.0 || ^6.0",
-        "vimeo/psalm": "^4.8.1"
+        "vimeo/psalm": "^4.20.0"
     },
     "suggest": {
         "ext-bcmath": "Decimal128 type support"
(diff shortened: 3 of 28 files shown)


lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
```diff
@@ -571,7 +571,7 @@ private function addIndex(ClassMetadata $class, SimpleXMLElement $xmlIndex): voi
     }
 
     /**
-     * @return array<string, array<string, mixed>|scalar>
+     * @return array<string, array<string, mixed>|scalar|null>
```
malarzm, line 574: `null` doesn't look like an acceptable value for the array but I guess it comes from `convertXMLElementValue` so let's live with that

lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionTrait.php
```diff
@@ -660,6 +662,11 @@ private function doAdd($value, $arrayAccess)
      * @param mixed $offset
      *
      * @return bool|T|null
+     * @psalm-return (
+     *      $arrayAccess is false
+     *      ? T|null
+     *      : T|null|true
```
malarzm, line 668: oh boy, now we're programming in comments :D

lib/Doctrine/ODM/MongoDB/Proxy/Resolver/ProxyManagerClassNameResolver.php
```diff
@@ -27,13 +26,6 @@ public function getRealClass(string $class): string
         return $this->resolveClassName($class);
     }
 
-    /**
-     * @psalm-param class-string<RealClassName>|class-string<ProxyInterface<RealClassName>> $className
```
malarzm: These are not needed anymore or just problematic?

malarzm (approved):
Awesome 🎉

lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
franmomu: yep, exactly

lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionTrait.php
franmomu: 😬 this can be really interesting when adding generics for example to `Query`, that depending on the `hydrate` value, maybe we change the return type of `Query::execute()` to just `Iterator<TDocument>` which would mean removing quite `assert` calls.

lib/Doctrine/ODM/MongoDB/Proxy/Resolver/ProxyManagerClassNameResolver.php
franmomu: This is interesting, we were getting:

```
ERROR: MoreSpecificImplementedParamType - lib/Doctrine/ODM/MongoDB/Proxy/Resolver/ProxyManagerClassNameResolver.php:37:45 - Argument 1 of Doctrine\ODM\MongoDB\Proxy\Resolver\ProxyManagerClassNameResolver::resolveClassName has the more specific type 'class-string<ProxyManager\Proxy\ProxyInterface<RealClassName>>|(class-string<RealClassName:fn-doctrine\odm\mongodb\proxy\resolver\proxymanagerclassnameresolver::resolveclassname as object>)', expecting 'class-string<Doctrine\Persistence\Proxy<T>>|(class-string<T:fn-doctrine\persistence\mapping\proxyclassnameresolver::resolveclassname as object>)' as defined by Doctrine\Persistence\Mapping\ProxyClassNameResolver::resolveClassName (see https://psalm.dev/140)
    public function resolveClassName(string $className): string
```
so I've removed it, but didn't pay attention that in [`doctrine/persistence`](https://github.com/doctrine/persistence/blob/0130920efd696cf5480744f499672b2d4ca47c85/lib/Doctrine/Persistence/Mapping/ProxyClassNameResolver.php#L7-L16), the interface is using `Proxy`:
```php
<?php

namespace Doctrine\Persistence\Mapping;

use Doctrine\Persistence\Proxy;

interface ProxyClassNameResolver
{
    /**
     * @psalm-param class-string<Proxy<T>>|class-string<T> $className
     *
     * @psalm-return class-string<T>
     *
     * @template T of object
     */
    public function resolveClassName(string $className): string;
}
```
we are using`ocramius/proxy-manager`, so I'll leave it and ignore it since it adds the correct type.

SenseException (approved):

malarzm:
I think I've removed correct checks from the list of required ones, let's hope next PRs can be merged not only by me :+1:

PR #2410, opened 2022-02-19
Update .doctrine-project.json

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | task
| BC Break     | no
| Fixed issues | 

#### Summary

~We need to create 2.4.x~, I don't don't if the release job should have created it


malarzm (approved):

malarzm:
Thanks @franmomu! I wonder if we could automate changing the JSON file somehow 🤔

PR #2411, opened 2022-02-20
Use typed properties for default metadata

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

~I'll retarget to `2.4.x` once is created.~

This is kind of based on https://github.com/doctrine/orm/pull/8439 and https://github.com/doctrine/orm/pull/8589.

The idea is that if you use typed properties, it gets the type from that properties. For `array`, I've chosen `hash`, I thought it was more right.

### DIFF
--- a/UPGRADE-2.4.md
+++ b/UPGRADE-2.4.md
@@ -1,0 +1,15 @@
+# UPGRADE FROM 2.3 to 2.4
+
+## Typed properties as default mapping metadata
+
+When using typed properties on Document classes, Doctrine will use these types to set defaults mapping types.
+
+If you have defined some properties like:
+
+```php
+#[Field]
+private int $myProp;
+```
+
+This property will be stored in DB as `string` but casted back to `int`. Please note that at this
+time, due to backward compatibility reasons, nullable type does not imply `nullable` mapping.
--- a/docs/en/reference/basic-mapping.rst
+++ b/docs/en/reference/basic-mapping.rst
@@ -198,6 +198,25 @@ This list explains some of the less obvious mapping types:
     suitable you should either use an embedded document or use formats provided
     by the MongoDB driver (e.g. ``\MongoDB\BSON\UTCDateTime`` instead of ``\DateTime``).
 
+.. _reference-php-mapping-types:
+
+PHP Types Mapping
+_________________
+
+Since version 2.4 Doctrine can determine usable defaults from property types
+on document classes. Doctrine will map PHP types to ``type`` attribute as
+follows:
+
+- ``DateTime``: ``date``
+- ``DateTimeImmutable``: ``date_immutable``
+- ``array``: ``hash``
+- ``bool``: ``bool``
+- ``float``: ``float``
+- ``int``: ``int``
+- ``string``: ``string``
+
+Please note that at this time, due to backward compatibility reasons, nullable type does not imply `nullable` mapping.
+
 Property Mapping
 ----------------
 
--- a/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
@@ -5,6 +5,8 @@ declare(strict_types=1);
 namespace Doctrine\ODM\MongoDB\Mapping;
 
 use BadMethodCallException;
+use DateTime;
+use DateTimeImmutable;
 use Doctrine\Instantiator\Instantiator;
 use Doctrine\Instantiator\InstantiatorInterface;
 use Doctrine\ODM\MongoDB\Id\IdGenerator;
@@ -20,6 +22,7 @@ use InvalidArgumentException;
 use LogicException;
 use ProxyManager\Proxy\GhostObjectInterface;
 use ReflectionClass;
+use ReflectionNamedType;
 use ReflectionProperty;
 
 use function array_filter;
@@ -42,6 +45,8 @@ use function sprintf;
 use function strtolower;
 use function strtoupper;
 use function trigger_deprecation;
+
+use const PHP_VERSION_ID;
 
 /**
  * A <tt>ClassMetadata</tt> instance holds all the object-document mapping metadata
@@ -1525,6 +1530,11 @@ use function trigger_deprecation;
     {
         $mapping['embedded'] = true;
         $mapping['type']     = self::ONE;
+
+        if ($this->isTypedProperty($mapping['fieldName'])) {
+            $mapping = $this->validateAndCompleteTypedAssociationMapping($mapping);
+        }
+
         $this->mapField($mapping);
     }
 
@@ -1549,6 +1559,11 @@ use function trigger_deprecation;
     {
         $mapping['reference'] = true;
         $mapping['type']      = self::ONE;
+
+        if ($this->isTypedProperty($mapping['fieldName'])) {
+            $mapping = $this->validateAndCompleteTypedAssociationMapping($mapping);
+        }
+
         $this->mapField($mapping);
     }
 
@@ -2195,6 +2210,19 @@ use function trigger_deprecation;
             unset($this->generatorOptions['type']);
         }
 
+        if ($this->isTypedProperty($mapping['fieldName'])) {
+            $mapping = $this->validateAndCompleteTypedFieldMapping($mapping);
+
+            if (isset($mapping['type']) && ($mapping['type'] === self::ONE || $mapping['type'] === self::MANY)) {
+                $mapping = $this->validateAndCompleteTypedAssociationMapping($mapping);
+            }
+        }
+
+        if (! isset($mapping['type'])) {
+            // Default to string
+            $mapping['type'] = Type::STRING;
+        }
+
         if (! isset($mapping['nullable'])) {
             $mapping['nullable'] = false;
         }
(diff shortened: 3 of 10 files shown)


IonBazan:
Do you think we could also add support for References and embedded documents there?

docs/en/reference/basic-mapping.rst
```diff
@@ -130,6 +130,23 @@ to use with the ``setDefaultDB`` method:
...
+Since version 2.4 Doctrine can determine usable defaults from property types
+on document classes. Doctrine will map PHP types to ``type`` attribute as
+follows:
+
+- ``DateTime``: ``date``
+- ``DateTimeImmutable``: ``date_immutable``
+- ``array``: ``hash``
+- ``bool``: ``boolean``
+- ``float``: ``float``
+- ``int``: ``integer``
```
IonBazan: `boolean` and `integer` are deprecated since v2.1.
```suggestion
- ``bool``: ``bool``
- ``float``: ``float``
- ``int``: ``int``
```

franmomu:
> Do you think we could also add support for References and embedded documents there?

Right now it sets `targetDocument` for `EmbedOne` and `ReferenceOne` if that's what you mean.

IonBazan (approved):
Nice, I think I missed that during review 👍🏻  This looks great and good to go from my side 🚀 

I'm just wondering if we need to add the `changelog` entry, as it could potentially introduce a BC break for people who relied on non-strict PHP type casting:

```php
#[Field] // Will be stored in DB as `string` but casted back to `int`
private int $myProp;
```
which doesn't seem like a big deal though.

docs/en/reference/basic-mapping.rst
```diff
@@ -130,6 +130,23 @@ to use with the ``setDefaultDB`` method:
...
+on document classes. Doctrine will map PHP types to ``type`` attribute as
+follows:
+
+- ``DateTime``: ``date``
+- ``DateTimeImmutable``: ``date_immutable``
+- ``array``: ``hash``
+- ``bool``: ``bool``
+- ``float``: ``float``
+- ``int``: ``int``
+- ``string`` or any other type: ``string``
```
malarzm:
```suggestion
- ``string``: ``string``
```

I'd remove the "or any other type" part as this paragraph is before we teach user what types we have. Now I wrote that I think the section should be below listing all available mapping types :)

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -2505,4 +2533,73 @@ private function checkDuplicateMapping(array $mapping): void
...
+     *
+     * @psalm-param FieldMappingConfig $mapping
+     *
+     * @return FieldMappingConfig
+     */
+    private function validateAndCompleteTypedFieldMapping(array $mapping): array
+    {
+        $type = $this->reflClass->getProperty($mapping['fieldName'])->getType();
+
+        if ($type instanceof ReflectionNamedType && ! isset($mapping['type'])) {
```
malarzm: It'd be better to have an early return instead of a nested switch.

tests/Doctrine/ODM/MongoDB/Tests/Mapping/ClassMetadataTest.php
```diff
@@ -136,6 +138,44 @@ public function testFieldIsNullable(): void
...
+        $cm->mapField(['fieldName' => 'array']);
+        self::assertEquals(Type::HASH, $cm->getTypeOfField('array'));
+
+        // bool
+        $cm->mapField(['fieldName' => 'boolean']);
+        self::assertEquals(Type::BOOL, $cm->getTypeOfField('boolean'));
+
+        // float
+        $cm->mapField(['fieldName' => 'float']);
+        self::assertEquals(Type::FLOAT, $cm->getTypeOfField('float'));
```
malarzm, line 171: `int` seems to be missing compared to the list of what is autocompleted :)

malarzm (commented):
Awesome idea and job porting the feature from ORM! Two additional ideas:
1. Do you know why ORM is not inferring nullable typed properties? I think we should do that, seems pretty straightforward, but I wonder if there are any dragons down the road.
2. In the same vein as EmbedOne we could autocomplete `collectionClass` attribute for EmbedMany/ReferenceMany. I can take care of it in a follow-up PR if you'd like me to

franmomu:
> Awesome idea and job porting the feature from ORM! Two additional ideas:
> 
> 1. Do you know why ORM is not inferring nullable typed properties? I think we should do that, seems pretty straightforward, but I wonder if there are any dragons down the road.
> 2. In the same vein as EmbedOne we could autocomplete `collectionClass` attribute for EmbedMany/ReferenceMany. I can take care of it in a follow-up PR if you'd like me to

I'll take a look at those maybe later today, with the nullable type I remember there were some BC breaks as https://github.com/doctrine/orm/issues/8723, but I haven't checked the reason.

franmomu:
> I'm just wondering if we need to add the `changelog` entry, as it could potentially introduce a BC break for people who relied on non-strict PHP type casting:
> 
> ```
> #[Field] // Will be stored in DB as `string` but casted back to `int`
> private int $myProp;
> ```
> 
> which doesn't seem like a big deal though.

Interesting! I'll add a comment for that case

franmomu:
About the nullable type property see https://github.com/doctrine/orm/issues/8723#issuecomment-851229328

I'm not sure if that would introduce some unwanted behaviour 🤔

tests/Doctrine/ODM/MongoDB/Tests/Mapping/AbstractMappingDriverTest.php
```diff
@@ -194,6 +198,28 @@ public function testIdentifier(ClassMetadata $class): ClassMetadata
...
+        $this->assertSame(Type::DATE_IMMUTABLE, $class->getTypeOfField('dateTimeImmutable'));
+        $this->assertSame(Type::HASH, $class->getTypeOfField('array'));
+        $this->assertSame(Type::BOOL, $class->getTypeOfField('boolean'));
+        $this->assertSame(Type::FLOAT, $class->getTypeOfField('float'));
+
+        $this->assertSame(TypedEmbeddedDocument::class, $class->getAssociationTargetClass('embedOne'));
+        $this->assertSame(UserTyped::class, $class->getAssociationTargetClass('referenceOne'));
+
+        $this->assertSame(CustomCollection::class, $class->getAssociationCollectionClass('embedMany'));
+        $this->assertSame(CustomCollection::class, $class->getAssociationCollectionClass('referenceMany'));
```
IonBazan, line 220: What happens if you simply typehint `Collection`? I think we should be checking if it's an interface or class and skip the custom collection logic for `ArrayCollection` and `PersistentCollection`.

IonBazan: I think we should also mention this change in docs.

IonBazan:
Regarding nullable properties, I think we should leave it in current state. The behaviour in ODM is slightly different - `nullable` parameter decides whether to store the fields that were set to `null` and I would expect it to require explicit `nullable: true` declaration to make it store the field when using typed properties.

tests/Doctrine/ODM/MongoDB/Tests/Mapping/AbstractMappingDriverTest.php
franmomu: > What happens if you simply typehint `Collection`? I think we should be checking if it's an interface or class and skip the custom collection logic for `ArrayCollection` and `PersistentCollection`.

With `Collection` it uses `ArrayCollection` as `collectionClass`, the thing here is that it uses the `collectionClass` to [generate another class that extends from it and implements `PersistentCollectionInterface`](https://github.com/doctrine/mongodb-odm/blob/d15c8579022497f26c2eb75eaefc9fcef542db5c/lib/Doctrine/ODM/MongoDB/PersistentCollection/DefaultPersistentCollectionGenerator.php#L139), so we cannot use an interface as `collectionClass`.

UPGRADE-2.4.md
```diff
@@ -0,0 +1,14 @@
...
+When using typed properties on Document classes, Doctrine will use these types to set defaults mapping types.
+
+If you have defined some properties like:
+
+```php
+#[Field]
+private int $myProp;
+```
+
+This property will be stored in DB as `string` but casted back to `int`.
```
malarzm:
```suggestion
This property will be stored in DB as `string` but casted back to `int`. Please note that at this 
time, due to backward compatibility reasons, nullable type does not imply `nullable` mapping.
```

docs/en/reference/basic-mapping.rst
```diff
@@ -198,6 +198,23 @@ This list explains some of the less obvious mapping types:
...
+follows:
+
+- ``DateTime``: ``date``
+- ``DateTimeImmutable``: ``date_immutable``
+- ``array``: ``hash``
+- ``bool``: ``bool``
+- ``float``: ``float``
+- ``int``: ``int``
+- ``string``: ``string``
+
```
malarzm, line 217: I'd add similar note about `nullable` as in upgrade docs as the behaviour is not fully intuitional

malarzm (approved):
As for nullable types I'd suggest to leave this PR as is and merge it once checks are passing. Then create a follow-up PR that will emit a deprecation once it detect typed nullable property that is not on par with mapping with intention to throw an exception in 3.0.

tests/Documents74/UserTyped.php
```diff
@@ -0,0 +1,105 @@
...
+
+    /** @ODM\EmbedMany  */
+    #[ODM\EmbedMany]
+    public CustomCollection $embedMany;
+
+    /** @ODM\ReferenceMany  */
+    #[ODM\ReferenceMany]
+    public CustomCollection $referenceMany;
+
+    public static function loadMetadata(ClassMetadata $metadata): void
```
malarzm: Just noticed: what's this?

malarzm (approved):
Are you fixing the PHPStan failure here or in a follow-up PR as it seems unrelated?

franmomu:
> Are you fixing the PHPStan failure here or in a follow-up PR as it seems unrelated?

I was creating https://github.com/doctrine/mongodb-odm/pull/2417 👍 to target `2.3.x`

tests/Documents74/UserTyped.php
franmomu: 🤔 It was copied from the ORM thinking about us having support for [Static PHP Mapping](https://github.com/doctrine/persistence/blob/2.3.x/lib/Doctrine/Persistence/Mapping/Driver/StaticPHPDriver.php), but we are not, so I'll remove it, thanks!

tests/Doctrine/ODM/MongoDB/Tests/Mapping/AbstractMappingDriverTest.php
IonBazan: Tested and it looks like it works perfectly. Thanks!

malarzm:
Thanks @franmomu 🚀

malarzm:
About the deprecation thing: https://doctrine.slack.com/archives/CAAE6T66B/p1647934766611479?thread_ts=1647885533.164609&cid=CAAE6T66B

> @IonBazan some may prefer to keep these null values in DB, but some could decide to skip them and this is controlled by nullable: true/false accordingly. The name is a bit unfortunate for someone who comes from ORM background because it actually means “store null values” but that’s another story. I chose not to store my nulls in the DB on purpose.

> @malarzm Actually I never used ODM in a way that I have a null in the object, not-nullable mapping, and these two factors combined end with field not being written or $unset. Frankly I thought that'd throw an exception somewhere in the reflection land

Issue #2404, opened 2022-01-13
Support for PHP 8.1 enums


IonBazan:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes
| BC Break    | no

#### Summary

We could add native support for PHP 8.1+ enums by adding a new parameter to the mapping that allows to define the enum class name like it's done in ORM: https://github.com/doctrine/orm/pull/9304

ozahorulia:
I'm already using this feature in the latest doctrine ORM. It would be great to have the same functionality for the ODM.

malarzm:
Referencing ORM's PR we could base our work on: https://github.com/doctrine/orm/pull/9304

malarzm:
Merged in #2440

PR #2412, opened 2022-02-20
add BackedEnum support

IonBazan:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | #2404 

#### Summary

~Provides native support for `BackedEnum`s. This could be later integrated with #2411 to allow even better DX.~
Already integrated in this PR 🚀 

#### Additional changes:

- Coding Standards workflow runs on PHP 8.1 now and ignores the enum `case` indentation due to https://github.com/squizlabs/PHP_CodeSniffer/issues/3474
- Locks PHPStan < 1.5.0 as it reports some false positives due to a bug in template intersection types


IonBazan:
This is now ready for review. Please note that the "required" Coding Standards run will not be executed because I changed the PHP version there so the required name no longer match repository settings.

Also, it will produce conflicts with #2411 - depending on which PR is merged first, the other one must be updated to bring typed property guessing for enums.

.github/workflows/coding-standards.yml
```diff
@@ -10,4 +10,6 @@ on:
 jobs:
   coding-standards:
     name: "Coding Standards"
-    uses: "doctrine/.github/.github/workflows/coding-standards.yml@1.4.0"
+    uses: "doctrine/.github/.github/workflows/coding-standards.yml@1.4.1"
```
malarzm: I think the part changing CS could go into 2.3 as a separate PR so we can have one list of jobs required to pass for a PR. Thoughts?

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/Field.php
```diff
@@ -16,4 +17,24 @@
...
+    /** @var class-string<BackedEnum>|null */
+    public $enumType;
+
+    /**
+     * @param mixed[]                       $options
+     * @param class-string<BackedEnum>|null $enumType
+     */
+    public function __construct(
+        ?string $name = null,
+        ?string $type = 'string',
```
malarzm: We need to remember to change the default value here as #2411 did

lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
```diff
@@ -275,6 +275,10 @@ public function loadMetadataForClass($className, \Doctrine\Persistence\Mapping\C
                     $mapping['notSaved'] = ((string) $attributes['not-saved'] === 'true');
                 }
 
+                if (isset($mapping['enum-type'])) {
```
malarzm: This looks fishy given the check above and below are checking (and assigning from) the `$attributes` array.

lib/Doctrine/ODM/MongoDB/Mapping/MappingException.php
```diff
@@ -276,4 +278,39 @@ public static function schemaValidationError(int $errorCode, string $errorMessag
...
+
+    public static function enumsRequirePhp81(string $className, string $fieldName): self
+    {
+        return new self(sprintf('Enum types require PHP 8.1 in %s::$%s', $className, $fieldName));
+    }
+
+    public static function nonEnumTypeMapped(string $className, string $fieldName, string $enumType): self
+    {
+        return new self(sprintf(
+            'Attempting to map non-enum type %s as enum in document class %s::$%s',
```
malarzm:
```suggestion
            'Attempting to map a non-enum type %s as an enum: %s::%s',
```

This should be the most consistent with other exceptions we have

lib/Doctrine/ODM/MongoDB/Mapping/MappingException.php
```diff
@@ -276,4 +278,39 @@ public static function schemaValidationError(int $errorCode, string $errorMessag
     {
         return new self(sprintf('The following schema validation error occurred while parsing the "%s" property of the "%s" class: "%s" (code %s).', $property, $className, $errorMessage, $errorCode));
     }
+
+    public static function enumsRequirePhp81(string $className, string $fieldName): self
+    {
+        return new self(sprintf('Enum types require PHP 8.1 in %s::$%s', $className, $fieldName));
```
malarzm:
```suggestion
        return new self(sprintf('Enum types require PHP 8.1 in %s::%s', $className, $fieldName));
```

We don't use `$` in other exceptions

lib/Doctrine/ODM/MongoDB/Mapping/MappingException.php
```diff
@@ -276,4 +278,39 @@ public static function schemaValidationError(int $errorCode, string $errorMessag
...
+     */
+    public static function invalidEnumValue(
+        string $className,
+        string $fieldName,
+        string $value,
+        string $enumType,
+        ValueError $previous
+    ): self {
+        return new self(sprintf(
+            'Trying to hydrate enum property "%s::$%s". Problem: Case "%s" is not listed in enum "%s"',
```
malarzm:
```suggestion
            'Trying to hydrate an enum property %s::%s, but "%s" is not listed as one of "%s" cases',
```

lib/Doctrine/ODM/MongoDB/Mapping/MappingException.php
```diff
@@ -276,4 +278,39 @@ public static function schemaValidationError(int $errorCode, string $errorMessag
...
+            $className,
+            $fieldName
+        ));
+    }
+
+    /**
+     * @param class-string             $className
+     * @param class-string<BackedEnum> $enumType
+     */
+    public static function invalidEnumValue(
```
malarzm: This one is definitely not a mapping exception but rather a https://github.com/doctrine/mongodb-odm/blob/2.3.x/lib/Doctrine/ODM/MongoDB/Hydrator/HydratorException.php

lib/Doctrine/ODM/MongoDB/Mapping/ReflectionEnumProperty.php
```diff
@@ -0,0 +1,82 @@
...
+use ReflectionProperty;
+use ReturnTypeWillChange;
+use ValueError;
+
+use function assert;
+use function get_class;
+use function is_int;
+use function is_string;
+
+class ReflectionEnumProperty extends ReflectionProperty
```
malarzm: It's a pity this class did not went into a common repository. If you're not willing to pursue this then it should be final and marked as `@internal`

lib/Doctrine/ODM/MongoDB/Mapping/ReflectionEnumProperty.php
```diff
@@ -0,0 +1,82 @@
...
+            return null;
+        }
+
+        $enum = $this->originalReflectionProperty->getValue($object);
+
+        if ($enum === null) {
+            return null;
+        }
+
+        return $enum->value;
```
malarzm: Given this class is usable and actually loaded only in PHP 8.1 we should be safe to call
```suggestion
        return $this->originalReflectionProperty->getValue($object)?->value;
```
but feel free to ignore, probably there's a lot more places that would benefit from the approach in the future and by then we'll have a CS-rule that'd autofix everything.

tests/Doctrine/ODM/MongoDB/Tests/Mapping/xml/Documents81.Card.dcm.xml
```diff
@@ -0,0 +1,14 @@
...
+<doctrine-mongo-mapping
+    xmlns="http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping
+    http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping.xsd"
+>
+    <document name="Documents81\Card">
+        <id />
+        <field name="suit" type="string" enum-type="Documents81\Suit" />
+        <field name="nullableSuit" type="string" enum-type="Documents81\Suit" nullable="true" />
```
malarzm, line 12: When we go down type-inferring rule we'll need to think how to correctly choose string vs int for the type conversion.

malarzm (changes requested):
Great you took care of porting adding the feature! Beside my remarks above please also add a relevant documentation.

Food for thought: in the past it was requested that we pass `$mapping` to types. While I'm still reluctant to do so it'd make the implementation easier as we'd just add new `enum` type that would read target enum from the mapping. Thoughts on this?

lib/Doctrine/ODM/MongoDB/Mapping/ReflectionEnumProperty.php
malarzm: @beberlei do you have any insights why ReflectionEnumProperty isn't a part of https://github.com/doctrine/persistence/tree/2.3.x/lib/Doctrine/Persistence/Reflection?

tests/Doctrine/ODM/MongoDB/Tests/Mapping/xml/Documents81.Card.dcm.xml
IonBazan: We can check if the enum class implements `StringBackedEnum` or `IntBackedEnum` to detect the underlying type.

lib/Doctrine/ODM/MongoDB/Mapping/ReflectionEnumProperty.php
IonBazan: This `$enum` seems to become `null` when it's `null` in DB so I reverted that. See: https://github.com/doctrine/mongodb-odm/runs/5283674934?check_suite_focus=true

malarzm: That's because the `?` in `?->` got lost :)

IonBazan: You are right @malarzm but after second thought, I decided to keep it pre-PHP8.1 compatible. This is in case someone accidentally imports this class in their code which would trow a really bad syntax error. I think we could revisit that topic once we decide to drop older PHP versions.

lib/Doctrine/ODM/MongoDB/Mapping/ReflectionEnumProperty.php
malarzm: While waiting for doctrine/common merge and release we could add this class (final and internal). Once doctrine/common is there we'll change implementation :)

malarzm:
#2411 was merged, this needs to be rebased (it's conflicting anyway) and adapted :)

lib/Doctrine/ODM/MongoDB/Mapping/ReflectionEnumProperty.php
IonBazan: @malarzm What about the exception thrown? The `doctrine/persistence` one throws native `ValueError`, while here we are converting it to a `HydratorException` which is not available in `doctrine/persistence`. I suggest simplifying this implementation to throw native error instead too for easier migration.

IonBazan:
@malarzm rebased and adjusted accordingly. I will keep it as separate commits for cleaner diffs but I'll squash them once approved.

lib/Doctrine/ODM/MongoDB/Mapping/ReflectionEnumProperty.php
malarzm: The HydratorInterface doesn't make any promises on what exceptions can be thrown, but generally it throws HydratorException when something is off with data (https://github.com/doctrine/mongodb-odm/blob/2.3.x/lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php#L333 and more, look for `throw`). If possible and relatively simple I'd wrap native ValueError with our HydratorException. If it's not simple then let's the native error be thrown, we can't save users from everything :)

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -145,6 +149,7 @@
  *      index?: bool,
  *      criteria?: array<string, string>,
  *      alsoLoadFields?: list<string>,
+ *      enumType?: class-string<BackedEnum>|null,
```
franmomu:
```suggestion
 *      enumType?: class-string<BackedEnum>,
```
I guess `enumType` cannot contain `null`

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -100,7 +103,8 @@
  *      criteria?: array<string, string>,
  *      alsoLoadFields?: list<string>,
  *      order?: int|string,
- *      background?: bool
+ *      background?: bool,
+ *      enumType?: class-string<BackedEnum>|null,
```
franmomu:
```suggestion
 *      enumType?: class-string<BackedEnum>,
```

IonBazan:
Since `doctrine/persistence` v2.4 is released, we can now use it. I decided to drop the custom `ReflectionProperty` and keep throwing the `ValueError` if something goes wrong (to be consistent with normal fields).

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -2556,6 +2575,15 @@ private function validateAndCompleteTypedFieldMapping(array $mapping): array
             return $mapping;
         }
 
+        if (PHP_VERSION_ID >= 80100 && ! $type->isBuiltin() && enum_exists($type->getName(), true)) {
```
malarzm, line 2578: Please add this feature to documentation https://github.com/doctrine/mongodb-odm/pull/2411/files#diff-239d9a128ce24a76e204d8b9821ce009ef5c7b16298d549165da9bc05d5d8f97 created

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -2556,6 +2575,15 @@ private function validateAndCompleteTypedFieldMapping(array $mapping): array
...
+            $mapping['enumType'] = $type->getName();
+
+            $reflection = new ReflectionEnum($type->getName());
+            $type       = $reflection->getBackingType();
+
+            assert($type instanceof ReflectionNamedType);
```
malarzm, line 2584: `null` case is not handled (assert is just for PHPStan). Using a non-backed enum should warrant an exception

VincentLanglet:
What's need to finish this PR ? Do you need help @IonBazan ?
Seems to be a blocking point for the 2.4 version with doctrine/persistence v3 support (cf https://github.com/doctrine/mongodb-odm/pull/2428#issuecomment-1111467384)

malarzm:
FWIW I'll be wrapping this up over the weekend.

IonBazan:
Thanks @malarzm for taking over. Sorry, I didn't have enough time to make it prod-ready.

PS. Let me know if you want me to squash anything before you add your changes.

malarzm:
Merged in #2440

Issue #2419, opened 2022-03-19
Support doctrine/persistence 3.0


malarzm:
Ideally we need to allow `lowest possible 2.x || ^3.0`. I guess we'll need to go for 2.4 as a minimum because of ReflectionEnum and annotation driver changes anyway.

Basing on https://github.com/doctrine/persistence/blob/3.0.x/UPGRADE.md:

- [x] Removed `OnClearEventArgs::clearsAllEntities()` and `OnClearEventArgs::getEntityClass()` - https://github.com/doctrine/mongodb-odm/pull/2423
- [x] removed `ObjectManagerAware` - not used directly in ODM
- [x] removed `AnnotationDriver` - https://github.com/doctrine/mongodb-odm/pull/2420 https://github.com/doctrine/mongodb-odm/pull/2443
- [x] Removed `MappingException::pathRequired()` - not used directly in ODM
- [x] removed LifecycleEventArgs::getEntity() - method was deprecated for 10 years now (see https://github.com/doctrine/persistence/pull/281), a note in UPGRADE will cut it
- [x] removed support for short namespace aliases - we'll bring no compatibility layer for this
- [x] removed ObjectManager::merge() - #2435
- [x] removed support for doctrine/cache - https://github.com/doctrine/mongodb-odm/pull/2253
- [x] changed signatures - we can't add types to method parameters while we still support 2.x
- [x] Removed `PersistentObject` - not used directly in ODM

PR allowing `|| ^3.0` should mention in the upgrade notes that users may need to lock `doctrine/persistence` themselves should they rely on functionalities that were removed, especially:
- removed LifecycleEventArgs::getEntity()
- removed support for short namespace aliases

malarzm:
We need to prepare for https://github.com/doctrine/persistence/pull/272/files as our event is reusing common one's properties: https://github.com/doctrine/mongodb-odm/blob/2.3.x/lib/Doctrine/ODM/MongoDB/Event/OnClearEventArgs.php

PR #2420, opened 2022-03-22
Use colocated mapping driver

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

This is based on https://github.com/doctrine/orm/pull/9587 to allow https://github.com/doctrine/mongodb-odm/issues/2419 and start decoupling from `doctrine/annotations`.


lib/Doctrine/ODM/MongoDB/Mapping/Driver/AnnotationDriver.php
```diff
@@ -335,6 +362,23 @@ private function setShardKey(ClassMetadata $class, ODM\ShardKey $shardKey): void
...
     }
 
+    /**
+     * Retrieve the current annotation reader
+     *
+     * @return Reader
+     */
+    public function getReader()
+    {
+        Deprecation::trigger(
```
malarzm: We're using trigger_deprecation in ODM

franmomu: 🤦 thanks I forgot about that.

malarzm (approved):

malarzm:
Thanks @franmomu!

PR #2425, opened 2022-04-11
Added --disable-validators option to odm:schema:update command

### LINKED ISSUE #2405: After upgrading  odm into 2.3.0+ faced with the issue with doctrine:mongodb:schema:update command
since 2.3.0+ doctrine:mongodb:schema:update command start execute validators on the schema also and it can't be skipped, beforehand only indexes were applied, are there any chances to add option or parameter into the command to skip this?

ksowa:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | #2405 

#### Summary

This PR adds `--disable-validators` option to the `odm:schema:update` command. Once the option is enabled, the command will not try to update database-level validation rules.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Tools/Console/Command/Schema/UpdateCommand.php
+++ b/lib/Doctrine/ODM/MongoDB/Tools/Console/Command/Schema/UpdateCommand.php
@@ -27,7 +27,8 @@ class UpdateCommand extends AbstractCommand
         $this
             ->setName('odm:schema:update')
             ->addOption('class', 'c', InputOption::VALUE_OPTIONAL, 'Document class to process (default: all classes)')
-            ->setDescription('Update indexes for your documents');
+            ->addOption('disable-validators', null, InputOption::VALUE_NONE, 'Do not update database-level validation rules')
+            ->setDescription('Update indexes and validation rules for your documents');
     }
 
     /**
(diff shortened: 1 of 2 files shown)


ksowa:
I just fixed coding standards issues. The PR is ready to be checked again.

lib/Doctrine/ODM/MongoDB/Tools/Console/Command/Schema/UpdateCommand.php
```diff
@@ -27,15 +27,17 @@ protected function configure()
         $this
             ->setName('odm:schema:update')
             ->addOption('class', 'c', InputOption::VALUE_OPTIONAL, 'Document class to process (default: all classes)')
-            ->setDescription('Update indexes for your documents');
+            ->addOption('disable-validators', 'd', InputOption::VALUE_NONE, 'Do not update database-level validation rules')
```
malarzm: Could we please provide no shortcut for the option? Let's be as verbose as possible with the opt-out. Also simple `-d` for validators will look weird if we ever add other switch

malarzm (commented):
I like the simplicity of this approach 👍 We'll think about throwing exceptions when we receive first complaint about validators not being created, let's hope it'll be never.

Also please squash commits :)

lib/Doctrine/ODM/MongoDB/Tools/Console/Command/Schema/UpdateCommand.php
ksowa: @malarzm Done.

malarzm (approved):

malarzm:
Thanks @ksowa :tada:

PR #2433, opened 2022-04-26
Specify iterable types in Aggregation namespace

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

Follow-up of https://github.com/doctrine/mongodb-odm/pull/2432 with `Aggregation` namespace. I think there will be one or two more to finish adding iterable types in the `lib` directory.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
@@ -41,7 +41,7 @@ class Expr
      */
     private $currentField;
 
-    /** @var array|null */
+    /** @var array{case: mixed|self, then?: mixed|self}|null */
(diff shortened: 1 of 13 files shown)


lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
```diff
@@ -41,7 +41,7 @@ class Expr
      */
     private $currentField;
 
-    /** @var array|null */
+    /** @var array{case: mixed|self, then?: mixed|self}|null */
```
SenseException, line 44: Expressions are really a mixed bag of types.

SenseException (approved):

PR #2434, opened 2022-04-27
Deprecate NOTIFY tracking policy

### LINKED ISSUE #2424: Deprecate NOTIFY change tracking policy
I've just found out that ORM plans to deprecate the feature (https://github.com/doctrine/orm/issues/8383):

> the notify change tracking policy has issues and no performance gains over deferred explicit.

Given we've done nothing to improve NOTIFY change tracking since I'm here I guess we're having same issues as ORM had and just don't know about it. Let this issue here serve as an announcement and we'll see if anybody complains :)

malarzm:
Closes #2424

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataFactory.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataFactory.php
@@ -27,6 +27,7 @@ use function get_class;
 use function get_class_methods;
 use function in_array;
 use function interface_exists;
+use function trigger_deprecation;
 use function ucfirst;
 
 /**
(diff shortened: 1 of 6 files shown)


lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataFactory.php
```diff
@@ -207,6 +208,15 @@ protected function doLoadMetadata($class, $parent, $rootEntityFound, array $nonS
 
         $eventArgs = new LoadClassMetadataEventArgs($class, $this->dm);
         $this->evm->dispatchEvent(Events::loadClassMetadata, $eventArgs);
+
+        if ($class->isChangeTrackingNotify()) {
```
franmomu, line 213: the CS build wants to invert the condition to use early exit, but in this case the code is more readable the way it is

malarzm: I've decided to ignore the error. Offending lines will be gone eventually and so will be the suppression. Is this OK?

franmomu (approved):

PR #2436, opened 2022-04-29
Add more iterable types

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

Follow-up of https://github.com/doctrine/mongodb-odm/pull/2433

With this finishes adding iterable types in `lib` directory, next PR will add iterable types in `tests` and will add the PHPStan check.

The ignored issues in PHPStan are expected, one is because the type is incorrect and that's what the test is checking (same for Psalm) and the other issues are because PHPStan doesn't yet understand importing custom types in traits.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Doctrine\ODM\MongoDB\Persisters;
 
 use BadMethodCallException;
+use Doctrine\ODM\MongoDB\Aggregation\Stage\Sort;
 use Doctrine\ODM\MongoDB\DocumentManager;
 use Doctrine\ODM\MongoDB\Hydrator\HydratorException;
 use Doctrine\ODM\MongoDB\Hydrator\HydratorFactory;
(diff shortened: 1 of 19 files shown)


lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -944,6 +969,11 @@ private function getSortDirection($sort)
     /**
      * Prepare a sort specification array by converting keys to MongoDB field
      * names and changing direction strings to int.
+     *
+     * @param array<string, int|string|array<string, string>> $fields
+     * @psalm-param SortShape $fields
+     *
+     * @psalm-return array<string, -1|1|array{"$meta": 'textScore'}>
```
SenseException: I don't understand how `array{"$meta": 'textScore'}` fits into the sorting, but it seems the CI is happy with it. Maybe it's worth to put the type into a class and import it?

SenseException (approved):

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
franmomu: 👍 I've created a type, the thing here is that I think this is only used for aggregations: https://www.mongodb.com/docs/manual/reference/operator/aggregation/sort/#mongodb-pipeline-pipe.-sort but these methods are used for queries and aggregations.

PR #2437, opened 2022-04-30
Add PHPStan check for iterable types

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

Last PR to add the PHPStan check for iterable types, the remaining ones where the ones in `tests`.

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/Aggregation/AggregationOperatorsProviderTrait.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Aggregation/AggregationOperatorsProviderTrait.php
@@ -606,7 +606,9 @@ trait AggregationOperatorsProviderTrait
     }
 
     /**
-     * @param mixed $args
+     * @param Closure(Expr): mixed[]|mixed[] $args
+     *
+     * @return mixed[]
      */
     protected function resolveArgs($args): array
     {
--- a/tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/ProjectTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/ProjectTest.php
@@ -62,6 +62,8 @@ class ProjectTest extends BaseTest
     }
 
     /**
+     * @param string[] $args
+     *
      * @dataProvider provideProxiedExprMethods
      */
     public function testProxiedExprMethods(string $method, array $args = []): void
--- a/tests/Doctrine/ODM/MongoDB/Tests/Events/LifecycleListenersTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Events/LifecycleListenersTest.php
@@ -7,6 +7,7 @@ namespace Doctrine\ODM\MongoDB\Tests\Events;
 use BadMethodCallException;
 use Doctrine\Common\Collections\Collection;
 use Doctrine\ODM\MongoDB\DocumentManager;
+use Doctrine\ODM\MongoDB\Event\LifecycleEventArgs;
 use Doctrine\ODM\MongoDB\Event\PostCollectionLoadEventArgs;
 use Doctrine\ODM\MongoDB\Events;
 use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
@@ -224,6 +225,9 @@ class MyEventListener
     /** @psalm-var array<string, list<class-string>> */
     public array $called = [];
 
+    /**
+     * @param array{LifecycleEventArgs} $args
+     */
     public function __call(string $method, array $args): void
     {
         $document                = $args[0]->getDocument();
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/CommitImprovementTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/CommitImprovementTest.php
@@ -6,6 +6,7 @@ namespace Doctrine\ODM\MongoDB\Tests\Functional;
 
 use Doctrine\Common\EventSubscriber;
 use Doctrine\ODM\MongoDB\APM\CommandLogger;
+use Doctrine\ODM\MongoDB\Event\LifecycleEventArgs;
 use Doctrine\ODM\MongoDB\Events;
 use Doctrine\ODM\MongoDB\LockException;
 use Doctrine\ODM\MongoDB\PersistentCollection\PersistentCollectionInterface;
@@ -160,6 +161,9 @@ class PhonenumberMachine implements EventSubscriber
         ];
     }
 
+    /**
+     * @param array{LifecycleEventArgs} $args
+     */
     public function __call(string $eventName, array $args): void
     {
         $document = $args[0]->getDocument();
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/ReadPreferenceTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/ReadPreferenceTest.php
@@ -13,6 +13,9 @@ use Documents\User;
 use MongoDB\Driver\ReadPreference;
 use MongoDB\Driver\WriteConcern;
 
+/**
+ * @psalm-type ReadPreferenceTagShape = array{dc?: string, usage?: string}
+ */
 class ReadPreferenceTest extends BaseTest
 {
     public function setUp(): void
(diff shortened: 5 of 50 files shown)


tests/Doctrine/ODM/MongoDB/Tests/Aggregation/AggregationOperatorsProviderTrait.php
```diff
@@ -615,6 +615,8 @@ protected function createExpr(): Expr
 
     /**
      * @param mixed $args
+     *
+     * @return mixed[]
```
malarzm, line 611:
```suggestion
     * @return Closure|mixed[]
```

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/ProjectTest.php
```diff
@@ -64,6 +64,8 @@ public function provideAccumulators(): array
     }
 
     /**
+     * @param mixed[] $args
```
malarzm: FWIW I'm not really a fan of adding `mixed[]` all over the place, IMO this is bringing little to no value

tests/Doctrine/ODM/MongoDB/Tests/Events/LifecycleListenersTest.php
```diff
@@ -225,6 +225,9 @@ class MyEventListener
     /** @psalm-var array<string, list<class-string>> */
     public $called = [];
 
+    /**
+     * @param mixed[] $args
```
malarzm: Can we somehow note that first element should be an instance of `LifecycleEventArgs` (and we don't care about the rest)?

tests/Doctrine/ODM/MongoDB/Tests/Functional/CommitImprovementTest.php
```diff
@@ -162,6 +162,9 @@ public function getSubscribedEvents(): array
         ];
     }
 
+    /**
+     * @param mixed[] $args
```
malarzm: Same question as before with `LifecycleEventArgs`

tests/Doctrine/ODM/MongoDB/Tests/Functional/ReadPreferenceTest.php
```diff
@@ -43,6 +43,8 @@ public function testHintIsNotSetByDefault(): void
     }
 
     /**
+     * @param array<array<string, string>> $tags
```
malarzm: Shall we introduce a `ReadPreferenceTagsShape` shape similar to `SortShape`?

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH1232Test.php
```diff
@@ -93,7 +93,7 @@ class GH1232Comment
 class GH1232CommentRepository extends DocumentRepository
 {
     /**
-     * @return array|int|object|null
+     * @return array<string, mixed>|int|object|null
```
malarzm: Isn't this returning `GH1232Comment[]`?

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH2002Test.php
```diff
@@ -14,6 +14,8 @@
 class GH2002Test extends BaseTest
 {
     /**
+     * @psalm-param array{"$ref": string, class: class-string} $expectedReference
```
malarzm: The `class` key is available only in first dataset. This makes me believe these changes are not really helpful :)

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH560Test.php
```diff
@@ -114,6 +117,9 @@ public function getSubscribedEvents(): array
         return $this->events;
     }
 
+    /**
+     * @param mixed[] $args
```
malarzm: Here goes `LifecycleEventArgs` question again

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/MODM83Test.php
```diff
@@ -63,6 +63,9 @@ class MODM83EventListener
     /** @var array<string, class-string[]> */
     public $called = [];
 
+    /**
+     * @param mixed[] $args
```
malarzm: I'm just marking all places with `LifecycleEventArgs` I see in case there'd be a place to improve (next 2 classes could use this as well)

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/MODM95Test.php
```diff
@@ -78,7 +78,7 @@ public function __construct()
      * by mapping array indexes (size URL's are required, cropMetadata is not).
      * Any invalid elements will be ignored.
      *
-     * @param array|Traversable $embeddedDocuments
+     * @param array<MODM95TestEmbeddedDocument>|Traversable<MODM95TestEmbeddedDocument> $embeddedDocuments
```
malarzm: Both `@param` and code could be improved to use `iterable<MODM95TestEmbeddedDocument>` which wasn't available at the time of writing the test

tests/Doctrine/ODM/MongoDB/Tests/Persisters/PersistenceBuilderTest.php
```diff
@@ -229,6 +231,10 @@ public function getDocumentsAndExpectedData(): array
         ];
     }
 
+    /**
+     * @param array<string, mixed>      $expectedData
+     * @param array<string, mixed>|null $preparedData
```
malarzm: I doubt that `null` should be allowed for `$preparedData` as the very first thing we do is iterate over it :)

tests/Doctrine/ODM/MongoDB/Tests/UnitOfWorkTest.php
```diff
@@ -249,6 +249,9 @@ public function testComputingChangesetForFileWithoutMetadataThrowsNoError(): voi
     }
 
     /**
+     * @param array<string, string>|null $origData
+     * @param array<string, string>|null $updateData
```
malarzm: There are currently also ints and bools in the array

malarzm (changes requested):
I'm not saying "definitely no", but given few definitions are off and tests are green I'm not sure these changes provide any real value other than satisfying SA tools. Maybe we could stay with level 5 for tests and improve just the main codebase for higher levels? To add another pro some time ago while working on sanity checks I needed to ignore phpstan/psalm errors in tests because I wanted to pass wrong values to check if exceptions are raised properly.

franmomu:
> I'm not saying "definitely no", but given few definitions are off and tests are green I'm not sure these changes provide any real value other than satisfying SA tools. Maybe we could stay with level 5 for tests and improve just the main codebase for higher levels? To add another pro some time ago while working on sanity checks I needed to ignore phpstan/psalm errors in tests because I wanted to pass wrong values to check if exceptions are raised properly.

Thanks @malarzm for the review! I've finally found some time to take a look at this again. I've just updated the code with the comments and make it green improving a bit some of the `mixed[]` types.

I'll try to add another commit removing some of the phpdoc that don't add value and adding them to the baseline.

malarzm:
Awesome! Let me know if you'd need anything :)

franmomu:
I think it's ready to review, I understand the `mixed[]` vs `array` concern, to me when I see `mixed[]` it's like the type has been reviewed, but maybe `list<mixed>` would be better to express that it is list of mixed types and the key doesn't matter.

tests/Doctrine/ODM/MongoDB/Tests/Events/LifecycleListenersTest.php
```diff
@@ -224,6 +225,9 @@ class MyEventListener
     /** @psalm-var array<string, list<class-string>> */
     public array $called = [];
 
+    /**
+     * @param array{LifecycleEventArgs} $args
```
malarzm, line 229: Cool this was possible :)

malarzm (approved):
Thank you very much @franmomu for your hard work on this 🍻

PR #2438, opened 2022-05-01
Leverage generic persistence event classes

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

Follows https://github.com/doctrine/persistence/pull/273 and imitates https://github.com/doctrine/orm/pull/9633

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Event/LifecycleEventArgs.php
+++ b/lib/Doctrine/ODM/MongoDB/Event/LifecycleEventArgs.php
@@ -7,11 +7,11 @@ namespace Doctrine\ODM\MongoDB\Event;
 use Doctrine\ODM\MongoDB\DocumentManager;
 use Doctrine\Persistence\Event\LifecycleEventArgs as BaseLifecycleEventArgs;
 
-use function assert;
-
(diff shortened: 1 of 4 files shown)


lib/Doctrine/ODM/MongoDB/Event/LifecycleEventArgs.php
```diff
@@ -12,6 +12,8 @@
 /**
  * Lifecycle Events are triggered by the UnitOfWork during lifecycle transitions
  * of documents.
+ *
+ * @template-extends BaseLifecycleEventArgs<DocumentManager>
```
malarzm, line 14: I recall we're free to remove `assert($dm instanceof DocumentManager);` here now :)

franmomu: Good point! I guessed PHPStan would complain about redundant condition or similar, thanks!

malarzm (approved):

malarzm:
Thanks @franmomu!

PR #2439, opened 2022-05-01
Add some generic information

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

To reach level 6 of PHPStan the remaining check is about generics, there are ~110 errors. This PR fixes some the "easy" ones.

The rest I guess that will involve making some classes generic (`Builder`, `Query` and `Aggregate` probably) and that would mean more errors and changes.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Configuration.php
+++ b/lib/Doctrine/ODM/MongoDB/Configuration.php
@@ -89,8 +89,8 @@ class Configuration
      *      autoGeneratePersistentCollectionClasses?: self::AUTOGENERATE_*,
      *      classMetadataFactoryName?: class-string<ClassMetadataFactory>,
      *      defaultCommitOptions?: CommitOptions,
-     *      defaultDocumentRepositoryClassName?: class-string<ObjectRepository>,
-     *      defaultGridFSRepositoryClassName?: class-string<GridFSRepository>,
+     *      defaultDocumentRepositoryClassName?: class-string<ObjectRepository<object>>,
+     *      defaultGridFSRepositoryClassName?: class-string<GridFSRepository<object>>,
      *      defaultDB?: string,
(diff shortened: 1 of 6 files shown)


IonBazan (approved):
I just wish PHPStan would allow default values for template values.

lib/Doctrine/ODM/MongoDB/PersistentCollection/AbstractPersistentCollectionFactory.php
```diff
@@ -36,8 +36,12 @@ public function create(DocumentManager $dm, array $mapping, ?BaseCollection $col
      * Creates instance of collection class to be wrapped by PersistentCollection.
      *
      * @param string $collectionClass FQCN of class to instantiate
+     * @psalm-param class-string<TDocument> $collectionClass
      *
      * @return BaseCollection
+     * @psalm-return BaseCollection<array-key, TDocument>
```
malarzm: This looks wrong: `TDocument` is taken from the parameter (which tells what class should be instantiated) and is definitely not what is later stored in the collection (we don't know what it'll be at this point). The only thing we can be kind-of-certain is that `$collectionClass` is a descendant of the `Collection` interface.

malarzm (changes requested):

lib/Doctrine/ODM/MongoDB/PersistentCollection/AbstractPersistentCollectionFactory.php
franmomu: yeah, totally right, I missed this, thanks!

malarzm (approved):

SenseException (approved):

PR #2440, opened 2022-05-02
add BackedEnum support

malarzm:
Finishing touches for #2412

Things changed contrary to original PR:
- PHPStan is no longer locked
- added exception when non-backed enum is being used
- added tests for class metadata creation
- added docs

### DIFF
--- a/docs/en/reference/basic-mapping.rst
+++ b/docs/en/reference/basic-mapping.rst
@@ -215,6 +215,10 @@ follows:
 - ``int``: ``int``
 - ``string``: ``string``
 
+Doctrine will also autoconfigure any backed ``enum`` it encounters: ``type``
+will be set to ``string`` or ``int``, depending on the enum's backing type,
+and ``enumType`` to the enum's |FQCN|.
+
 Please note that at this time, due to backward compatibility reasons, nullable type does not imply `nullable` mapping.
 
 Property Mapping
--- a/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Doctrine\ODM\MongoDB\Mapping;
 
+use BackedEnum;
 use BadMethodCallException;
 use DateTime;
 use DateTimeImmutable;
@@ -18,10 +19,12 @@ use Doctrine\ODM\MongoDB\Utility\CollectionHelper;
 use Doctrine\Persistence\Mapping\ClassMetadata as BaseClassMetadata;
 use Doctrine\Persistence\Mapping\ReflectionService;
 use Doctrine\Persistence\Mapping\RuntimeReflectionService;
+use Doctrine\Persistence\Reflection\EnumReflectionProperty;
 use InvalidArgumentException;
 use LogicException;
 use ProxyManager\Proxy\GhostObjectInterface;
 use ReflectionClass;
+use ReflectionEnum;
 use ReflectionNamedType;
 use ReflectionProperty;
 
@@ -34,6 +37,7 @@ use function assert;
 use function class_exists;
 use function constant;
 use function count;
+use function enum_exists;
 use function extension_loaded;
 use function get_class;
 use function in_array;
@@ -100,7 +104,8 @@ use const PHP_VERSION_ID;
  *      criteria?: array<string, string>,
  *      alsoLoadFields?: list<string>,
  *      order?: int|string,
- *      background?: bool
+ *      background?: bool,
+ *      enumType?: class-string<BackedEnum>,
  * }
  * @psalm-type FieldMapping = array{
  *      type: string,
@@ -145,6 +150,7 @@ use const PHP_VERSION_ID;
  *      index?: bool,
  *      criteria?: array<string, string>,
  *      alsoLoadFields?: list<string>,
+ *      enumType?: class-string<BackedEnum>,
  * }
  * @psalm-type AssociationFieldMapping = array{
  *      type?: string,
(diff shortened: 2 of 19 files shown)


tests/Doctrine/ODM/MongoDB/Tests/Functional/EnumTest.php
```diff
@@ -0,0 +1,57 @@
...
+        $this->dm->getDocumentCollection(Card::class)->insertOne($document);
+
+        $this->expectException(ValueError::class);
+        $this->expectExceptionMessage(sprintf('"ABC" is not a valid backing value for enum "%s"', Suit::class));
+        $this->dm->getRepository(Card::class)->findOneBy([]);
+    }
+
+    protected function createMetadataDriverImpl(): MappingDriver
+    {
+        return AttributeDriver::create(__DIR__ . '/../../../Documents');
```
franmomu, line 54: just saw this, should this be `Documents81`? or maybe it's not used

tests/Doctrine/ODM/MongoDB/Tests/Functional/EnumTest.php
```diff
@@ -0,0 +1,57 @@
...
+    {
+        $doc       = new Card();
+        $doc->suit = Suit::Clubs;
+
+        $this->dm->persist($doc);
+        $this->dm->flush();
+        $this->dm->clear();
+
+        $saved = $this->dm->find(Card::class, $doc->id);
+        assert($saved instanceof Card);
```
franmomu:
```suggestion
        $this->assertInstanceOf(Card::class, $saved);
```
To be consistent with other tests

tests/Doctrine/ODM/MongoDB/Tests/Mapping/AbstractMappingDriverTest.php
```diff
@@ -637,6 +640,24 @@ public function testView(): void
...
+
+        self::assertSame(Suit::class, $metadata->fieldMappings['suit']['enumType']);
+        self::assertSame('string', $metadata->fieldMappings['suit']['type']);
+        self::assertFalse($metadata->fieldMappings['suit']['nullable']);
+        self::assertInstanceOf(EnumReflectionProperty::class, $metadata->reflFields['suit']);
+
+        self::assertSame(Suit::class, $metadata->fieldMappings['nullableSuit']['enumType']);
+        self::assertSame('string', $metadata->fieldMappings['nullableSuit']['type']);
+        self::assertTrue($metadata->fieldMappings['nullableSuit']['nullable']);
+        self::assertInstanceOf(EnumReflectionProperty::class, $metadata->reflFields['nullableSuit']);
```
franmomu, line 659: Maybe we could use a rule to enforce using `static::assertX` or `self::assertX` or `$this->assertX` (to always use the same thing, so we don't have to think which one should we use)

franmomu (approved):

tests/Doctrine/ODM/MongoDB/Tests/Mapping/AbstractMappingDriverTest.php
IonBazan: I'm for `self::assert*` (alternatively `static::assert*` but tests should be `final`) as all these methods (besides of `expectException*`) are static.

tests/Doctrine/ODM/MongoDB/Tests/Functional/EnumTest.php
malarzm: The `path` is not used, it's only needed to discover all metadata

docs/en/reference/basic-mapping.rst
```diff
@@ -214,6 +214,7 @@ follows:
 - ``float``: ``float``
 - ``int``: ``int``
 - ``string``: ``string``
+- ``BackedEnum``: ``string`` or ``int`` + ``enumType``
```
IonBazan: I'd put it into a separate sentence explaining the behaviour.

tests/Doctrine/ODM/MongoDB/Tests/Mapping/AbstractMappingDriverTest.php
malarzm: I was pretty sure we've changed all `$this->assert` to `self::assert` due to the fact these are static methods :) I'll stick with `$this->` for now to be consistent but yeah, a rule would be nice.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -2389,10 +2389,15 @@ public function mapField(array $mapping): array
                 throw MappingException::enumsRequirePhp81($this->name, $mapping['fieldName']);
             }
 
-            if (! enum_exists($mapping['enumType'])) {
+            if (! enum_exists($mapping['enumType'], true)) {
```
IonBazan: `bool $autoload = true` by default. Is explicitly passing it intended?

tests/Doctrine/ODM/MongoDB/Tests/Mapping/AbstractMappingDriverTest.php
malarzm: Heh actually we're already inconsistent

tests/Doctrine/ODM/MongoDB/Tests/Mapping/ClassMetadataTest.php
```diff
@@ -187,6 +191,87 @@ public function testFieldTypeFromReflection(): void
         self::assertEquals(CustomCollection::class, $cm->getAssociationCollectionClass('referenceMany'));
     }
 
+    /**
+     * @requires PHP >= 8.1
+     */
+    public function testEnumTypeFromReflection(): void
```
IonBazan, line 197: Good idea to add these tests here 👍🏻

tests/Doctrine/ODM/MongoDB/Tests/Mapping/AbstractMappingDriverTest.php
malarzm: So anyway I think I'm gonna leave this as-is and vote in favour of `self::assert` as that's what I'm using at work :D

IonBazan: Let's enforce that in the future with static analysis in a separate PR and ignore it for now. Tracked in #2441

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
malarzm: In this case no. Seen in the other place and thought `false` is default. Will remove from both 👍

docs/en/reference/basic-mapping.rst
malarzm: Done. I'll also expand the chapter a bit once this is merged

PR #2444, opened 2022-05-04
Final touches for type autoconfiguration

malarzm:
First commit is from #2442, feel free to ignore it.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
@@ -1563,6 +1563,7 @@ use const PHP_VERSION_ID;
     {
         $mapping['embedded'] = true;
         $mapping['type']     = self::MANY;
+
         $this->mapField($mapping);
     }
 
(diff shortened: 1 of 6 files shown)


lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -2120,7 +2122,7 @@ public function getAssociationTargetClass($assocName): ?string
             throw new InvalidArgumentException("Association name expected, '" . $assocName . "' is not an association.");
         }
 
-        return $this->associationMappings[$assocName]['targetDocument'];
+        return $this->associationMappings[$assocName]['targetDocument'] ?? null;
```
malarzm, line 2125: This is a leftover from my previous findings. Now the `null` will be returned as per return type instead of an illegal offset.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -2170,6 +2172,14 @@ public function mapField(array $mapping): array
             $mapping['fieldName'] = $mapping['name'];
         }
 
+        if ($this->isTypedProperty($mapping['fieldName'])) {
```
malarzm, line 2175: Moving the autocompletion all the way up before any sanity checks. Previously autoconfiguration was able to set incorrect values such as:
- combine `targetDocument` with `discriminatorMap`
- set `collectionClass` that would cause an exception if set explicitly

PR #2447, opened 2022-05-10
Allow using enums directly in queries

IonBazan:
|      Q       |   A
|------------- | -----------
| BC Break     | no

#### Summary

This change allows using `BackedEnum`s directly in the query builder instead of using `->value`:
```php
// Before:
$this->createQueryBuilder()->field('status')->equals(Status::APPROVED->value)->getQuery();

// Now
$this->createQueryBuilder()->field('status')->equals(Status::APPROVED)->getQuery();
```

This supports arrays too. Please do note that if `enumType` is not set, the value is passed directly to the `Type` without any changes to allow developers to customize the behaviour in their custom types if needed (see `testQueryWithMappedNonEnumFieldIsPassedToTypeDirectly()` where value is passed directly to `StringType` and producing an error).

I've added the tests in `EnumTest` instead of `DocumentPersisterTest` to avoid conditional mapping loading there (this feature requires PHP 8.1+).

Not sure if this should be treated as bugfix or improvement so feel free to assign to correct milestone.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Doctrine\ODM\MongoDB\Persisters;
 
+use BackedEnum;
 use BadMethodCallException;
 use Doctrine\ODM\MongoDB\Aggregation\Stage\Sort;
 use Doctrine\ODM\MongoDB\DocumentManager;
@@ -1128,6 +1129,10 @@ final class DocumentPersister
         }
 
         if (! $this->class->hasField($fieldName)) {
+            if ($value instanceof BackedEnum) {
+                $value = $value->value;
+            }
+
             return Type::convertPHPToDatabaseValue($value);
         }
 
@@ -1142,6 +1147,10 @@ final class DocumentPersister
             throw new InvalidArgumentException(
                 sprintf('Mapping type "%s" does not exist', $typeName)
             );
+        }
+
+        if ($value instanceof BackedEnum && isset($mapping['enumType'])) {
+            $value = $value->value;
         }
 
         if (in_array($typeName, ['collection', 'hash'])) {
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/EnumTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/EnumTest.php
@@ -9,6 +9,7 @@ use Doctrine\ODM\MongoDB\Tests\BaseTest;
 use Doctrine\Persistence\Mapping\Driver\MappingDriver;
 use Documents81\Card;
 use Documents81\Suit;
+use Error;
 use MongoDB\BSON\ObjectId;
 use ValueError;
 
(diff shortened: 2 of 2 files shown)


lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -1144,6 +1149,10 @@ private function convertToDatabaseValue(string $fieldName, $value)
             );
         }
 
+        if ($value instanceof BackedEnum && $mapping['enumType']) {
```
malarzm: Given `$mapping['enumType']` can be either `null` or a backed enum's FQCN (guaranteed by metadata validation) we should be OK to do following check (maybe backed by an `\assert` call to satisfy SA) which should be cheaper

```suggestion
        if (isset($mapping['enumType'])) {
```

malarzm: Also I'm not sure about the call being here instead of after next `if` statement. Actually I started wondering if we shouldn't ensure that `enumType` can be combined with `type=int|string` (or not set which allows autoconfiguration to kick in). Right now

```
/** @ODM\Field(type="collection", enumType=Suit::class) */
private array $enums = [];
```

looks OK from user's perspective but I believe it'll blow up?

malarzm (approved):
> Please do note that if enumType is not set, the value is passed directly to the Type without any changes

Thanks to this I'm inclined to release this PR as a bugfix :instead of hacing people wait until next minor 🎉 If anybody asks let's say we thought it'll work out of the box 🙈

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
IonBazan: This `$value instanceof BackedEnum` is necessary because the value might be the enum's backed value too:
```php
$this->dm->createQueryBuilder(Card::class)
            ->field('suit')->equals(Suit::Spades->value)
            ->field('nullableSuit')->equals('C');
```

I'll add a test for that.

Regarding mixing `enumType` with arrays, it will probably blow up if you map it this way. Not sure how could we solve it here though.

jmikola:
Is this PR in any way related to the driver? We recently had a bug report come in about using enums in conjunction with the driver's [`Persistable`](https://www.php.net/manual/en/class.mongodb-bson-persistable.php) interface. While that was mainly an issue with unserializing BSON _into_ an enum, I wonder if there's also another problem with how enums are serialized _to_ BSON (ignoring anything to do with [`Persistable`](https://www.php.net/manual/en/class.mongodb-bson-persistable.php)).

I'm tracking this enum/Persistable issue in [PHPC-2083](https://jira.mongodb.org/browse/PHPC-2083) and have started working on a PR.

I haven't actually tested standard BSON serialization for enums, but based on this issue it seems like they get encoded as BSON documents with `name` field. And for BackedEnum instances, there's would be an additional `value` field. According to the docs for [BackedEnum](https://www.php.net/manual/en/language.enumerations.backed.php), that type is intended for cases where an enum does need to get stored in a database (since a pure enum wouldn't have a scalar value).

Would it be preferable for the driver to raise an exception when the user attempts to serialize a pure enum (without Persistable) and encode a BackedEnum as its value (i.e. BSON integer or string)?

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
malarzm: > Not sure how could we solve it here though.

Not necessarily here, I'll hack  a PR later today to throw a mapping exception if `enumType` is combined with something else than `int` or `string` types. Then we'll have a clear "it's not working message" for userland instead of weird blow ups under the hood.

malarzm:
> Is this PR in any way related to the driver?

No, we ended with our own (ok ok, ORM's 🙄) way of handling enums (#2412): user must either specify an `enumClass` property in the mapping or use an enum as property's type and have ODM figure out correct mapping (#2411). Under the hood we use a [special reflection]( https://github.com/doctrine/persistence/blob/3.0.x/src/Persistence/Reflection/EnumReflectionProperty.php) with which we get enum's backing value and then treat it as a normal string/int. That's also what is stored in the database

> Would it be preferable for the driver to raise an exception when the user attempts to serialize a pure enum (without Persistable)

With ODM we've decided to deal only with `BackedEnum`s, using pure one ends with [a mapping exception](https://github.com/doctrine/mongodb-odm/blob/21d650f15f07df708365f7dc3c8851bcbc806f9b/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php#L2396-L2398)

> and encode a BackedEnum as its value (i.e. BSON integer or string)?

Encoding any backed enum as its backing value would save us some headaches that we are yet to discover, maybe they'd render this PR not needed as well. But is the behaviour safe from driver's perspective? I mean we're cool saving just the value because we have a mapping to know what enum instatiate later. I guess same could be done with an explicit typeMap

malarzm:
@IonBazan please squash commits ;)

malarzm:
Thanks @IonBazan!

jmikola:
> Encoding any backed enum as its backing value would save us some headaches that we are yet to discover, maybe they'd render this PR not needed as well. But is the behaviour safe from driver's perspective?

@IonBazan: I haven't worked enough with enums to say with any certainty, but the [documentation](https://www.php.net/manual/en/language.enumerations.backed.php) does say that backed enums exist to allow for database round-tripping. I assume that is alluding to ORM logic that would re-construct an enum from mapping information given a value (using the `from()` or `tryFrom()` methods).

IMO, allowing a pure or backed enum to be serialized to BSON is preferable from a driver perspective (assuming users opt in via the Persistable interface), since it's less for the application or an ORM to worry about; however, that does invite possible issues converting BSON to PHP if the document data references a class that no longer exists in the application. Enums just introduce some additional ways that code changes could break if a name or value is no longer defined.

PR #2449, opened 2022-05-22
Specify class-string return type

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

I've targeted 2.4.x, but not sure if should go into 2.5.x


malarzm (approved):
Good to go to 2.4.x IMO 👍

malarzm:
Thanks @franmomu!

Issue #2149, opened 2020-01-10
UUID generator doesn't work with binary UUID types


alcaeus:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.0.3

#### Summary

Given an identifier using `strategy="uuid"` and `type="bin_uuid"` or `type="bin_uuid_rfc4122"`, an error occurs when trying to store data:
> MongoDB\Driver\Exception\InvalidArgumentException : Expected UUID length to be 16 bytes, 32 given

#### Expected behavior

The UUID generator should be able to generate UUIDs for consumption with the binary BSON types for UUIDs.

dmitry84:
Hi, Any news regarding this issue?

alcaeus:
I had lost track of this issue. I wanted to revisit this yesterday, but with travis-ci no longer reliably running our builds I had to prioritise moving the build environment. I'll pick this up once CI is stable again, but if you want to take a shot at fixing this, please do.

Romaxx:
The problem is the representation of string.
uuid is passed as "string of hex" and not "hex string", "ab" = 16 bits not 8 bits, so the value is 32 bytes long instead the 16 bytes long (128 bits) attended size.
using this "dirty" code in convertToDatabaseValue in BinDataType.php solve the problem
```php
$str = $value;
        if (strlen($value) == 32){
            $str = '';
            for($i=0;$i<strlen($value);$i+=2){
                $str.=chr(hexdec(substr($value,$i,2)));
            }
        }
        if (! $value instanceof Binary) {
            return new Binary($str, $this->binDataType);
        }

alcaeus:
As a general rule, you shouldn't make such changes in the type itself, but rather in the code that generates the UUID. I just haven't gotten around to making the change and probably won't have the time to do so in the foreseeable future. If you're interested in working on this, I'm happy to help out.

MVA18:
Is there any progress on this issue?

alcaeus:
Given the lack of updates here, I'd say "obviously not".

Sayma23:
As a small immediate step, could someone open a dedicated failing test case that reproduces the error with `strategy='uuid'` and `type='bin_uuid'` or `type='bin_uuid_rfc4122'`? Even without a fix, a reproducible test would give contributors a clear starting point, confirm the bug scope, and prevent regressions once resolved. That's something everyone can likely agree on right now.

alcaeus:
[...]

PR #2456, opened 2022-08-26
Expected UUID length should be 16 bytes 

simondeeley:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #2149 

#### Summary

Quick and dirty attempt to fix length of UUID returned by the auto UUID generator.


simondeeley:
This won't work...

Issue #1863, opened 2018-09-20
Add pipeline to $lookup stage


Strick3r:
|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes/no
| BC Break    | yes/no

Mongo 3.6 add a pipeline lookup functionality that helps a lot in the futures developments, do you think that this feature can be added in a future veresion?

Thanks

malarzm:
[`$lookup`](https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/1.2/reference/aggregation-builder.html#lookup) stage is already available in 1.2, what seems to be missing?

Strick3r:
This is a new nice feature

New in version 3.6.

To perform uncorrelated subqueries between two collections as well as allow other join conditions besides a single equality match, the $lookup stage has the following syntax:

copycopied
{
   $lookup:
     {
       from: <collection to join>,
       let: { <var_1>: <expression>, …, <var_n>: <expression> },
       pipeline: [ <pipeline to execute on the collection to join> ],
       as: <output array field>
     }
}

malarzm:
All right, thanks for clarification. I've put this into 2.x milestone. If you have some time it'd be great to see a PR!

bluedogmilan:
Strongly needed!

malarzm:
@bluedogmilan sending us a PR with implemented feature will greatly speed things up

alcaeus:
@bluedogmilan furthermore, we’re holding all further features back to 2.1 to avoid introducing more complexity in 2.0. 2.1 will mainly focus on adding aggregation pipeline features from previous mongodb releases.

bluedogmilan:
> @bluedogmilan sending us a PR with implemented feature will greatly speed things up

@malarzm Sorry for the newb question but I really do not know what PR stands for...

Steveb-p:
I've taken a little time on it, since I'll probably need to use this functionality myself. Those two ways of handling `$lookup` are however really different and do not really mix & match.

Using one practically should disable another.

Is it acceptable to treat them as two separate classes with common ancestor? Or simply use some kind of flag to indicate which type of lookup is being used?

Should `pipeline` accept an instance of `Stage` / `Builder` like `Facet` does, or should it also allow to pass an array?

I've started by adding:
```php
public function let(array $let) : self
{
    $this->let = $let;
    return $this;
}

public function pipeline($pipeline) : self
{
[...]

PR #2459, opened 2022-09-03
Add pipeline to $lookup stage

khaperets:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | yes
| Fixed issues | #1863 

#### Summary

Added a pipeline lookup functionality that helps a lot in the futures developments,


Ravetracer:
I discovered that "foreignField" and "localField" is not mandatory when using "pipeline" in the $lookup stage. In fact it dropped the performance in our case so we had a query running for about 4 seconds. Without foreignField and localField the query runs in 250-300ms and the result set was the same.
Also it would be great if you can add an option to set variables using "let".

khaperets:
@Ravetracer Thanks for your comment. Changes pushed to PR.

malarzm:
@khaperets thanks for the PR! It seems that the test you've added is failing: https://github.com/doctrine/mongodb-odm/actions/runs/3067090011/jobs/4971970883#step:12:1783. The rest of errors seems to be unrelated, we'll investigate them. Once the test is green please fix coding standards, this can be done by running `vendor/bin/phpcbf` command. If possible, please also squash your commits so there's just one.

khaperets:
@malarzm I've done everything, thank you for the feedback.

malarzm:
@khaperets we've managed to get checks back to shape. May I ask you to rebase your PR atop newly created `2.5.x` branch?

khaperets:
@malarzm done

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Lookup.php
```diff
@@ -98,14 +104,25 @@ public function from(string $from): self
...
             ],
         ];
+
+        if (! empty($this->pipeline)) {
+            $expression['$lookup']['pipeline'] = $this->pipeline;
+
+            if (! empty($this->let)) {
+                $expression['$lookup']['let'] = $this->let;
+            }
+        } else {
```
malarzm: According to https://www.mongodb.com/docs/manual/reference/operator/aggregation/lookup/#correlated-subqueries-using-concise-syntax MongoDB 5.0 allows using all 4 fields together so the current these-or-those approach doesn't seem like a good idea.

khaperets: fixed

khaperets:
@malarzm Added compatibility with MongoDB 5.0

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Lookup.php
```diff
@@ -98,14 +106,30 @@ public function from(string $from): self
...
             '$lookup' => [
                 'from' => $this->from,
-                'localField' => $this->localField,
-                'foreignField' => $this->foreignField,
                 'as' => $this->as,
             ],
         ];
+
+        $serverVersion                    = $this->dm->getServerVersion($this->class->getName());
+        $isSupportsPipelineWithLocalField = version_compare($serverVersion, '5.0') >= 0;
```
malarzm: Generally we're letting end users handle version-specific behaviour as they do know what MongoDB version they're running. The most important reason is to prevent additional communication with database, as introduced here. So instead I'd rather see a way to reset builder's `localField` and `foreignField` properties when adding `pipeline`/`let`. I think the cleanest way to do will be a new method `resetLocalAndForeignField` that will `unset` both properties. Additionally this will allow use cases like https://github.com/doctrine/mongodb-odm/pull/2459#issuecomment-1249154690 to control how they query will be ran.

Also I didn't manage to comment last time, but I believe I've seen the test fail due to MongoDB version. Once we get back to letting users control the behaviour, you can skip some tests using methods like `\Doctrine\ODM\MongoDB\Tests\BaseTest::requireMongoDB42`.

Also feel free to ask me anything on our Slack, that way I'll be able to help faster :)

malarzm (changes requested):

malarzm (approved):

malarzm:
Merged manually in https://github.com/doctrine/mongodb-odm/commit/7ebc038973587222c89de11cad4367c2185776f9. Thanks a lot @khaperets!

PR #2462, opened 2022-09-19
Fix failing CI checks

malarzm:
Follow up to #2452


lib/Doctrine/ODM/MongoDB/DocumentManager.php
```diff
@@ -621,13 +621,15 @@ public function getReference(string $documentName, $identifier): object
         /** @psalm-var ClassMetadata<T> $class */
         $class = $this->metadataFactory->getMetadataFor(ltrim($documentName, '\\'));
         assert($class instanceof ClassMetadata);
+        /** @psalm-var T|false $document */
```
malarzm, line 624: Some of the changes I had to do look like a bug in Psalm. I'll try to pin them down in the meantime, figured green CI is more important

PR #2465, opened 2022-09-28
Allow collections 2

greg0ire:


tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/MODM29Test.php
```diff
@@ -25,6 +27,7 @@ public function testTest(): void
         $this->dm->persist($doc);
         $this->dm->flush();
 
+        assert(isset($collection[0], $collection[1], $collection[2]));
```
greg0ire, line 30: Not sure why, but PHPStan/Psalm complained here that `$collection` was an `ArrayCollection<MODM29Embedded|null>`. This doesn't look right, and I can no longer reproduce it locally, let me revert it.

greg0ire:
I just received a Travis email about a failing build: https://app.travis-ci.com/github/greg0ire/mongodb-odm/builds/256171720

Why do we still have Travis here?

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/MODM29Test.php
greg0ire: OK so it's PHPStan: https://github.com/doctrine/mongodb-odm/actions/runs/3158035027/jobs/5139609323#step:9:13

greg0ire: I managed to reproduce it locally.

greg0ire, line 30: It's unclear why this wasn't an issue before.

alcaeus (approved):

alcaeus:
> I just received a Travis email about a failing build: https://app.travis-ci.com/github/greg0ire/mongodb-odm/builds/256171720
> 
> Why do we still have Travis here?

Do you have an old state in your fork that still contains travis.yml?

greg0ire:
The only reference to it I could find is this one: https://github.com/doctrine/mongodb-odm/pull/2471

greg0ire:
Ah there is a webhook here: https://github.com/doctrine/mongodb-odm/settings/hooks

Let me delete it.

greg0ire:
Not sure that will be enough, Travis also appears here: https://github.com/doctrine/mongodb-odm/settings/installations. I don't know how to uninstall it.

malarzm:
The last job run on Travis was over year ago. Also it tells me that we're -10 credits whatever that means now :D Anyway everything seems to be turned off on Travis' end.

malarzm:
@greg0ire will you be updating https://github.com/doctrine/mongodb-odm/blob/2.5.x/UPGRADE-2.5.md for a warning about new methods?

greg0ire:
@malarzm I was not planning to no… there is no action to take if you upgrade to 2.5. What kind of warning would you add?

malarzm:
Something along

> ODM supports `doctrine/collections` 2.x along 1.x. Please note that we added `findFirst(...)` and `reduce(...)` methods to comply with the new `Collection` interface. Make sure signatures of new methods comply with your own ones, should you have implemented such methods before.

I think it's worthwhile because `doctrine/collections` is usually a transient dependency and one may not be aware of such changes.

PR #2467, opened 2022-09-29
Address changes from doctrine/collections 1.8.0

greg0ire:
`Collection::contains()` and `Collection::indexOf()` use templating to infer if they are going to return false or null respectively, in case `$element` does not even match the type of the collection.


alcaeus (approved):

malarzm:
Thanks @greg0ire! :)

Issue #2464, opened 2022-09-27
[RFC] Bump minimal PHP version to 7.4


malarzm:
https://packagist.org/packages/doctrine/mongodb-odm/php-stats#2 shows that ~95% of 2.x users are using at least PHP 7.4 while the minimum required version in 7.2. Given 7.2 -> 7.4 is not bumping for the sake of bumping but would allow us to modernize our code this is a minimum bump we should consider for next minor version.

- [x] Bump minimal PHP version in composer.json
- [x] Remove PHP version checks for PHP 7.x
- [x] Add typed properties and default values where applicable
- [x] Migrate the code to use latest syntax features (arrow functions, etc)

IonBazan:
By all means! Bumping to 7.4 would give us many syntax improvements but some breaking changes may occur when using covariance/contravariance.

franmomu:
is anything else missing here? 🤔 

It would be nice to release 2.5 to unlock other libraries to add support for new `doctrine/*` package versions (like DoctrineMongoDBBundle).

We can finish https://github.com/doctrine/mongodb-odm/pull/2489 and leave the removal of annotation dependency for 2.6.

alcaeus:
No objections from my end. Happy to defer making use of additional 7.4 features (i.e. typed properties) in a future minor version.

IonBazan:
Closing as implemented in linked PRs.

PR #2470, opened 2022-09-30
Make use of PHP 7.4 syntax

IonBazan:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | -

#### Summary

Migrates some of the old syntax to PHP 7.4 where applicable:
- Arrow functions
- Null-coalesce assignment operator
- `array_key_first` and `is_iterable`

Migrates usage of PHPUnit API:
- Adds `@doesNotPerformAssertions` where no assertions are performed
- Uses `$this->createMock()` shorthand

Linked to #2464

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/DocumentNotFoundException.php
+++ b/lib/Doctrine/ODM/MongoDB/DocumentNotFoundException.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 namespace Doctrine\ODM\MongoDB;
 
+use JsonException;
+
 use function json_encode;
 use function sprintf;
 
@@ -20,10 +22,15 @@ final class DocumentNotFoundException extends MongoDBException
      */
     public static function documentNotFound(string $className, $identifier): self
     {
+        try {
+            $id = json_encode($identifier, JSON_THROW_ON_ERROR);
+        } catch (JsonException $e) {
+        }
+
         return new self(sprintf(
             'The "%s" document with identifier %s could not be found.',
             $className,
-            json_encode($identifier, JSON_THROW_ON_ERROR)
-        ));
+            $id ?? false,
+        ), 0, $e ?? null);
     }
 }
--- a/lib/Doctrine/ODM/MongoDB/DocumentNotFoundException.php
+++ b/lib/Doctrine/ODM/MongoDB/DocumentNotFoundException.php
@@ -7,6 +7,8 @@ namespace Doctrine\ODM\MongoDB;
 use function json_encode;
 use function sprintf;
 
+use const JSON_THROW_ON_ERROR;
+
 /**
  * Class for exception when encountering proxy object that has
  * an identifier that does not exist in the database.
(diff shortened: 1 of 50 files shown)


lib/Doctrine/ODM/MongoDB/DocumentNotFoundException.php
```diff
@@ -21,7 +23,7 @@ public static function documentNotFound(string $className, $identifier): self
         return new self(sprintf(
             'The "%s" document with identifier %s could not be found.',
             $className,
-            json_encode($identifier)
+            json_encode($identifier, JSON_THROW_ON_ERROR)
```
IonBazan: This is something I'm not really sure if we should add here, as it can be considered a breaking change.

Steveb-p: You could catch the `JsonException` first if it occurs and concatenate it to the end of the message / set it as `$previous` exception. Then it should no longer be a BC break, as it would no longer change the type of an exception that might come out of it, while providing that little bit of extra information.

malarzm: `$previous` is the way to go IMO 👍

tests/Doctrine/ODM/MongoDB/Tests/Functional/MemoryUsageTest.php
```diff
@@ -25,6 +24,8 @@ class MemoryUsageTest extends BaseTest
 {
     /**
      * Output for jwage "Memory increased by 14.09 kb"
+     *
+     * @doesNotPerformAssertions
```
malarzm, line 28: Since you've found this test: I wonder if it's needed :D I mean nobody looks at tests that did not fail explicitly

malarzm: And this one won't be failing with an exception either

tests/Doctrine/ODM/MongoDB/Tests/Performance/HydrationPerformanceTest.php
```diff
@@ -20,6 +20,8 @@ class HydrationPerformanceTest extends BaseTest
 {
     /**
      * [jwage: 10000 objects in ~6 seconds]
+     *
+     * @doesNotPerformAssertions
```
malarzm, line 24: Same here. Additionally `phpbench` is in use for Performance tests nowadays

tests/Doctrine/ODM/MongoDB/Tests/Functional/MemoryUsageTest.php
IonBazan: Those tests are excluded from the test run in `phpunit.xml.dist`:
```xml
    <groups>
        <exclude>
            <group>performance</group>
        </exclude>
    </groups>
```

I think their main purpose was to quickly test the performance and memory usage before adding phpbench (the test is dated 2014).

IonBazan: I think I we could keep these tests for now, for historical reasons and perhaps delete them in a separate PR if you don't mind. I'd like to keep the diff ass small as possible here.

malarzm: Sure 👍

malarzm (approved):

malarzm:
Thanks @IonBazan!

PR #2471, opened 2022-09-30
Retire Travis

greg0ire:


IonBazan:
I think there is still `.travis.yml` file in the repository root, on both 2.4.x and 2.5.x branches. They should be deleted too.

greg0ire:
Oh right, I `grep`-ped, but did not `find`

malarzm (approved):

malarzm:
Thanks @greg0ire!

PR #2473, opened 2022-10-04
Use typed properties

IonBazan:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes (kind of)
| Fixed issues | -

#### Summary

Makes use of PHP 7.4 typed properties where applicable. Most of it was handled by Rector 🚀  and then manually fixed up.

Additionally, I removed some invalid `@doesNotPerformAssertions` annotations on the tests that actually perform them. Adding `failOnRisky="true"` in `phpunit.xml.dist` marks them as failed so it won't happen in the future.

Linked issue #2464

While all tests are passing at the moment, it would be great if someone could help testing it in a real-life application to eliminate potential BC breaks.


lib/Doctrine/ODM/MongoDB/Query/Expr.php
```diff
@@ -40,7 +40,7 @@ class Expr
     /**
      * The query criteria array.
      *
-     * @var array<string, mixed>
+     * @var array<string, mixed>|mixed
      */
     private $query = [];
```
IonBazan, line 45: This may not be an array but the actual value (when calling `equals()` without setting the `fieldName`. Reported by `\Doctrine\ODM\MongoDB\Tests\Query\ExprTest::testOperatorWithoutCurrentFieldWrapsEqualityCriteria()`

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Bucket.php
```diff
@@ -12,7 +12,7 @@
 class Bucket extends AbstractBucket
 {
     /** @var mixed[] */
-    private $boundaries;
+    private ?array $boundaries = null;
```
malarzm: Why nullable? PHPDoc says it's an array

malarzm: Ah I guess it's because of null checks later. Maybe we should keep this typed as array and use `isset` later? On the other hand this would change how `$fields = ['boundaries' => $this->boundaries];` would be behaving when no boundaries were set so maybe `null` is a good choice 🤔

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Facet.php
```diff
@@ -17,10 +17,9 @@
 class Facet extends Stage
 {
     /** @var Builder[] */
-    private $pipelines = [];
+    private array $pipelines = [];
 
-    /** @var string */
-    private $field;
+    private ?string $field = null;
```
malarzm: This could use `isset` (not saying it's necessary, just adding to my previous comment)

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Lookup.php
```diff
@@ -17,26 +17,19 @@
...
-    private $localField;
+    private ?string $localField = null;
 
-    /** @var string */
-    private $foreignField;
+    private ?string $foreignField = null;
 
-    /** @var string */
-    private $as;
+    private ?string $as = null;
```
malarzm, line 32: Some fields below were missed

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Out.php
```diff
@@ -13,11 +13,9 @@
 
 class Out extends Stage
 {
-    /** @var DocumentManager */
-    private $dm;
+    private DocumentManager $dm;
 
-    /** @var string */
-    private $collection;
+    private ?string $collection = null;
```
malarzm: If I'm seeing correctly `collection` will always be a string or an exception will be thrown

lib/Doctrine/ODM/MongoDB/DocumentManager.php
```diff
@@ -58,107 +58,83 @@ class DocumentManager implements ObjectManager
...
-    private $closed = false;
+    private bool $closed = false;
 
     /**
      * Collection of query filters.
-     *
-     * @var FilterCollection
      */
-    private $filterCollection;
+    private ?FilterCollection $filterCollection = null;
```
malarzm, line 133: We could initialize this in ctor to get rid of nullability and lazy creation (it seems light nowadays, maybe it wasn't  in the past)

lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
```diff
@@ -44,53 +44,40 @@ final class HydratorFactory
...
-    private $dm;
+    private DocumentManager $dm;
 
     /**
      * The UnitOfWork used to coordinate object-level transactions.
-     *
-     * @var UnitOfWork
      */
-    private $unitOfWork;
+    private ?UnitOfWork $unitOfWork = null;
```
malarzm, line 53: I guess we can't get rid of nullability here as there's a chicken and egg problem?

lib/Doctrine/ODM/MongoDB/Iterator/CachingIterator.php
```diff
@@ -31,16 +31,14 @@
 final class CachingIterator implements Countable, Iterator
 {
     /** @var array<mixed, TValue> */
-    private $items = [];
+    private array $items = [];
 
     /** @var Generator<mixed, TValue>|null */
     private $iterator;
```
malarzm: Why this was omitted?

lib/Doctrine/ODM/MongoDB/Iterator/UnrewindableIterator.php
```diff
@@ -26,8 +26,7 @@ final class UnrewindableIterator implements Iterator
     /** @var Generator<mixed, TValue>|null */
     private $iterator;
```
malarzm: Same as above

malarzm (commented):
Halfway through 🙈

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Bucket.php
IonBazan: I fixed that by making the property `private array $boundaries` and using null coalesce operator below: `$this->boundaries ?? null` to keep the result same as before. Looks much cleaner now but Psalm complains about it.

IonBazan:
I fixed the issues mentioned above. I think we need to change some phpstan rules to allow these changes if we want to avoid nullable properties.

tests/Doctrine/ODM/MongoDB/Tests/Functional/CustomCollectionsTest.php
```diff
@@ -174,16 +174,14 @@ public function testModifyingCollectionInChangeTrackingNotifyDocument(): void
...
-     * @doesNotPerformAssertions
-     */
     public function testMethodWithVoidReturnType(): void
     {
         $d = new DocumentWithCustomCollection();
         $this->dm->persist($d);
         $this->dm->flush();
 
         $d = $this->dm->find(get_class($d), $d->id);
+        $this->assertInstanceOf(MyEmbedsCollection::class, $d->coll);
```
malarzm:
```suggestion
        self::assertInstanceOf(MyEmbedsCollection::class, $d->coll);
```

malarzm (approved):

malarzm:
I haven't thought of any way to fix SA so I think that nullable optional properties will be the way to go after all. Sorry for wrong advice there 🙈

tests/Doctrine/ODM/MongoDB/Tests/Functional/CommitImprovementTest.php
```diff
@@ -148,11 +147,9 @@ public function testSchedulingCollectionDeletionAfterSchedulingForUpdate(): void
 
 class PhonenumberMachine implements EventSubscriber
 {
-    /** @var string[] */
-    private $numbers = ['12345678', '87654321'];
+    private array $numbers = ['12345678', '87654321'];
```
franmomu, line 151:
```suggestion
    /** @var string[] */
    private array $numbers = ['12345678', '87654321'];
```

franmomu (approved):
I think we can ignore the `RedundantPropertyInitializationCheck` psalm issues for now

tests/Doctrine/ODM/MongoDB/Tests/Functional/CustomCollectionsTest.php
IonBazan: I replaced other occurrences in the file but seems that we will need to merge up 2.4.x into 2.5.x to apply `self::assert*` normalization changes introduced recently.

franmomu (approved):

PR #2475, opened 2022-10-06
Normalizes assert calls in tests

### LINKED ISSUE #2441: Normalize `assert*()` calls in tests
### Improvement

Currently `assert*()` calls are inconsistent between tests:
- `self::assert*()` (preferred)
- `static::assert*()` - good alternative for above, but tests should usually be `final` and overriding core methods is not expected
- `$this->assert*()` - this one is invalid as assertion methods are usually static

This should be enforced by static analysis/code style.

hunomina:
Resolves https://github.com/doctrine/mongodb-odm/issues/2441

|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #2441 

#### Summary

Normalizes assert*() calls in tests

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/Aggregation/BuilderTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Aggregation/BuilderTest.php
@@ -170,7 +170,7 @@ class BuilderTest extends BaseTest
             ->addStage(new TestStage($builder))
             ->out('collectionName');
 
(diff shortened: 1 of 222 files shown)


malarzm (approved):

malarzm:
Thank you very much @hunomina!

hunomina:
You're welcome 😉
Thanks for merging it 👌

PR #2477, opened 2022-10-10
Normalize assert calls in tests

hunomina:
Some unusual calls were forgotten in [the preview PR](https://github.com/doctrine/mongodb-odm/pull/2475)

|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #2441 

#### Summary

Normalizes assert*() calls in tests

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/Events/LifecycleListenersTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Events/LifecycleListenersTest.php
@@ -12,7 +12,7 @@ use Doctrine\ODM\MongoDB\Events;
 use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
 use Doctrine\ODM\MongoDB\PersistentCollection\PersistentCollectionInterface;
 use Doctrine\ODM\MongoDB\Tests\BaseTest;
-use PHPUnit\Framework\TestCase;
+use PHPUnit\Framework\Assert;
 
 use function get_class;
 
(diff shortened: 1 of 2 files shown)


IonBazan (approved):

.idea/.gitignore
```diff
@@ -0,0 +1,8 @@
+# Default ignored files
```
malarzm: Can you please remove the `.idea` files from the PR? Usually it's best to keep IDE-specific files excluded from git via a global `.gitignore` file

hunomina: Oh yeah absolutely, my last (forced) push didn't go through apparently 😅 
I'll fix that 😉 
I'll add the `idea` directory to the `.gitignore` in a separate commit 😉

malarzm: Deal is that we don't want to keep IDE-specific files in project's `.gitignore` either :)

hunomina: Done 😉

hunomina: Oh ok, not even in the `.gitignore` 😮 ?
That's weird but ok whatever 😉
I removed my last commit, the one changing the `.gitignore` 😉

malarzm: Yeah, not even in `.gitignore`. Reasoning is simple: we don't want to have each and every contributor's IDE-specific files listed there 😅 While a lot of folks use JetBrains' IDEs, it's not always the case. The global `.gitignore` is a file you point your local git to and it adds up to whatever `.gitignore` is found in projects. Really handy :)

malarzm (approved):

malarzm:
Thanks @hunomina!

PR #2480, opened 2022-10-30
Allow doctrine/event-manager 2

### LINKED ISSUE #2478: Support doctrine/event-manager 2.x
`doctrine/event-manager` was just released: https://github.com/doctrine/event-manager/releases/tag/2.0.0. We need to support both 1.x and 2.x at the same time.

By the way we may revisit old constructs like https://github.com/doctrine/mongodb-odm/blob/f5c16795858ab537922ead80eb122927f6d78686/lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php#L475-L477 the `if` is no longer necessary, `dispatchEvent` is a NOP when no listener is registered.

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | 

#### Summary

I haven't found any calls to `EventManager::getListeners()` method.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataFactory.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataFactory.php
@@ -202,12 +202,10 @@ final class ClassMetadataFactory extends AbstractClassMetadataFactory
 
         $class->setParentClasses($nonSuperclassParents);
 
-        if (! $this->evm->hasListeners(Events::loadClassMetadata)) {
-            return;
-        }
-
-        $eventArgs = new LoadClassMetadataEventArgs($class, $this->dm);
-        $this->evm->dispatchEvent(Events::loadClassMetadata, $eventArgs);
+        $this->evm->dispatchEvent(
+            Events::loadClassMetadata,
+            new LoadClassMetadataEventArgs($class, $this->dm)
+        );
 
         // phpcs:ignore SlevomatCodingStandard.ControlStructures.EarlyExit.EarlyExitNotUsed
         if ($class->isChangeTrackingNotify()) {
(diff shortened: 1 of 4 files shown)


malarzm:
@franmomu any opinion on hasListeners usage? I was thinking we could get rid of it https://github.com/doctrine/mongodb-odm/issues/2478

franmomu:
> hasListeners

I didn't see the issue open 🤦 yeah I'll take a look at those cases

franmomu:
There is this one left:

https://github.com/doctrine/mongodb-odm/blob/f5c16795858ab537922ead80eb122927f6d78686/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataFactory.php#L90-L101

which maybe it's easier to read with the check first, but we could change it too.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataFactory.php
```diff
@@ -202,12 +202,10 @@ protected function doLoadMetadata($class, $parent, $rootEntityFound, array $nonS
...
-
-        $eventArgs = new LoadClassMetadataEventArgs($class, $this->dm);
-        $this->evm->dispatchEvent(Events::loadClassMetadata, $eventArgs);
+        $this->evm->dispatchEvent(
+            Events::loadClassMetadata,
+            new LoadClassMetadataEventArgs($class, $this->dm)
+        );
 
         // phpcs:ignore SlevomatCodingStandard.ControlStructures.EarlyExit.EarlyExitNotUsed
         if ($class->isChangeTrackingNotify()) {
```
franmomu, line 211: This looks like it should have been called always not only when there are no event listeners for `Events::loadClassMetadata`.

malarzm: Good catch 👍

malarzm (approved):

malarzm:
Re `onNotFoundMetadata` method: yeah, this seems more readbale this way :) Let's keep it as is.

malarzm:
Thanks @franmomu!

PR #2481, opened 2022-10-30
Upgrade to doctrine/coding-standard 10

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | task
| BC Break     | no
| Fixed issues | 

#### Summary

It allows to use Doctrine CS 10 (based on https://github.com/doctrine/orm/pull/10009)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -40,12 +40,12 @@
     },
     "require-dev": {
         "ext-bcmath": "*",
-        "doctrine/coding-standard": "^9.0",
+        "doctrine/coding-standard": "^10.0",
         "jmikola/geojson": "^1.0",
         "phpbench/phpbench": "^1.0.0",
         "phpstan/phpstan": "^1.4.6",
         "phpstan/phpstan-phpunit": "^1.0",
-        "phpunit/phpunit": "^9.5",
+        "phpunit/phpunit": "^9.5.5",
         "squizlabs/php_codesniffer": "^3.5",
         "symfony/cache": "^4.4 || ^5.0 || ^6.0",
         "vimeo/psalm": "^4.20.0"
--- a/lib/Doctrine/ODM/MongoDB/Query/Query.php
+++ b/lib/Doctrine/ODM/MongoDB/Query/Query.php
@@ -72,8 +72,8 @@ final class Query implements IteratorAggregate
     public const TYPE_DISTINCT        = 9;
     public const TYPE_COUNT           = 11;
 
-    public const HINT_REFRESH = 1;
     // 2 was used for HINT_SLAVE_OKAY, which was removed in 2.0
+    public const HINT_REFRESH         = 1;
     public const HINT_READ_PREFERENCE = 3;
     public const HINT_READ_ONLY       = 5;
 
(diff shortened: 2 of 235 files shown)


SenseException (approved):

composer.json
```diff
@@ -40,7 +40,7 @@
     },
     "require-dev": {
         "ext-bcmath": "*",
-        "doctrine/coding-standard": "^9.0",
+        "doctrine/coding-standard": "^9.0 || ^10.0",
```
malarzm: What's the reason for requiring one or another?

lib/Doctrine/ODM/MongoDB/Query/Query.php
```diff
@@ -74,6 +74,7 @@ final class Query implements IteratorAggregate
     public const TYPE_COUNT           = 11;
 
     public const HINT_REFRESH = 1;
+
```
malarzm: This one's a pity but I guess there's nothing we can do?

lib/Doctrine/ODM/MongoDB/Types/Type.php
```diff
@@ -23,13 +23,16 @@
...
+
     /** @deprecated const was deprecated in doctrine/mongodb-odm 2.1 and will be removed in 3.0. Use Type::INT instead */
     public const INTID    = 'int_id';
     public const CUSTOMID = 'custom_id';
     public const BOOL     = 'bool';
+
     /** @deprecated const was deprecated in doctrine/mongodb-odm 2.1 and will be removed in 3.0. Use Type::BOOL instead */
     public const BOOLEAN = 'boolean';
     public const INT     = 'int';
+
```
malarzm: Same here

malarzm (approved):
Let's hope this won't conflict much with https://github.com/doctrine/mongodb-odm/pull/2473 👀

franmomu:
> Let's hope this won't conflict much with #2473 👀

We can wait for that PR to get merged since these changes are automatic and do it in `2.5.x` (I guess next release would be minor) and since we're requiring there `phpunit/phpunit` ^9.5 we won't need to exclude `tests` for the one line comment.

lib/Doctrine/ODM/MongoDB/Query/Query.php
franmomu: I guess we can move the comment to the top, I'll give it a try.

lib/Doctrine/ODM/MongoDB/Types/Type.php
franmomu: maybe grouping all deprecated constants?

composer.json
franmomu: none 😬 I thought it was because the php version, but it's not necessary, thanks!

lib/Doctrine/ODM/MongoDB/Types/Type.php
malarzm: That'd be an option as well :)

IonBazan (approved):

PR #2485, opened 2022-11-29
Fix deprecation notices in CommandLogger

IonBazan:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | -

#### Summary

When using `ext-mongodb` [v1.15](https://github.com/mongodb/mongo-php-driver/releases/tag/1.15.0), following deprecation notices are shown:

>Deprecated: Return type of Doctrine\ODM\MongoDB\APM\CommandLogger::commandStarted(MongoDB\Driver\Monitoring\CommandStartedEvent $event) should either be compatible with MongoDB\Driver\Monitoring\CommandSubscriber::commandStarted(MongoDB\Driver\Monitoring\CommandStartedEvent $event): void, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /.../mongodb-odm/lib/Doctrine/ODM/MongoDB/APM/CommandLogger.php on line 47

This PR fixes it by adding `ReturnTypeWillChange` attribute.

See `ext-mongodb` release notes: https://github.com/mongodb/mongo-php-driver/releases/tag/1.15.0

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/APM/CommandLogger.php
+++ b/lib/Doctrine/ODM/MongoDB/APM/CommandLogger.php
@@ -44,12 +44,12 @@ final class CommandLogger implements Countable, CommandLoggerInterface
         $this->registered = false;
     }
 
-    public function commandStarted(CommandStartedEvent $event)
+    public function commandStarted(CommandStartedEvent $event): void
     {
         $this->startedCommands[$event->getRequestId()] = $event;
     }
 
-    public function commandSucceeded(CommandSucceededEvent $event)
+    public function commandSucceeded(CommandSucceededEvent $event): void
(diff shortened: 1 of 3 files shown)


alcaeus (changes requested):
Since the class is final, we could also add the `void` return type instead of suppressing the error, WDYT?

IonBazan:
@alcaeus I think you are right. I was afraid of the side effects on PHP 7.2 and 7.3 with older extension version (contravariance) but it's not a case here as adding missing types in derived classes was always allowed - only narrowing them down was not permitted.

alcaeus (approved):
LGTM

PR #2486, opened 2022-12-07
use aggregation builder in lookup stage pipeline

### LINKED ISSUE #2474: Lookup::pipeline should allow builder as an argument
#2459 introduced option to use `pipeline` which right now accepts an array. It should also accept a builder instance. We're free to change type signature here as `pipeline` was introduced in 2.5.0 and is not released yet.

aturki:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #2474

#### Summary

This PR provides implementation for using the aggregation builder as an argument to pipeline in the $lookup stage

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Lookup.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Lookup.php
@@ -11,6 +11,7 @@ use Doctrine\ODM\MongoDB\Mapping\ClassMetadata;
 use Doctrine\ODM\MongoDB\Mapping\MappingException;
 use Doctrine\ODM\MongoDB\Persisters\DocumentPersister;
 use Doctrine\Persistence\Mapping\MappingException as BaseMappingException;
+use InvalidArgumentException;
 
 /**
  * Fluent interface for building aggregation pipelines.
@@ -34,7 +35,7 @@ class Lookup extends Stage
     /** @var array<string, string>|null */
     private ?array $let = null;
 
-    /** @var Builder|null */
+    /** @var Builder|array<array<string, mixed>>|null */
     private $pipeline = null;
 
     private bool $excludeLocalAndForeignField = false;
@@ -109,8 +110,12 @@ class Lookup extends Stage
             $expression['$lookup']['let'] = $this->let;
         }
 
-        if ($this->pipeline instanceof Builder) {
-            $expression['$lookup']['pipeline'] = $this->pipeline->getPipeline(false);
+        if ($this->pipeline !== null) {
+            if ($this->pipeline instanceof Builder) {
+                $expression['$lookup']['pipeline'] = $this->pipeline->getPipeline(false);
+            } else {
+                $expression['$lookup']['pipeline'] = $this->pipeline;
+            }
         }
 
         return $expression;
@@ -171,7 +176,7 @@ class Lookup extends Stage
      * Instead, define variables for the joined document fields using the let option
      * and then reference the variables in the pipeline stages.
      *
-     * @param Builder|Stage $pipeline
+     * @param Builder|Stage|array<array<string, mixed>> $pipeline
      */
     public function pipeline($pipeline): self
     {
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Lookup.php
```diff
@@ -177,11 +177,15 @@ public function let(array $let): self
      * Instead, define variables for the joined document fields using the let option
      * and then reference the variables in the pipeline stages.
      *
-     * @param array<array<string, mixed>> $pipeline
+     * @param Builder|Stage $pipeline
```
alcaeus: I suggest using `Builder|Stage|array<array<string, mixed>>` as argument: there's no reason why people shouldn't be allowed to pass an assembled pipeline to this method.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Lookup.php
```diff
@@ -177,11 +177,15 @@ public function let(array $let): self
...
      */
-    public function pipeline(array $pipeline): self
+    public function pipeline($pipeline): self
     {
-        $this->pipeline = $pipeline;
+        if ($pipeline instanceof Stage) {
+            $this->pipeline = $pipeline->builder;
+        } else {
+            $this->pipeline = $pipeline;
+        }
```
alcaeus, line 187: In case `$this->pipeline` ends up being a builder, we should ensure that it's not identical to `$this->builder` as that would result in some really nasty pipelines as well as potentially infinite recursion.

alcaeus (changes requested):

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Lookup.php
```diff
@@ -171,11 +176,19 @@ public function let(array $let): self
...
+
+        if ($this->pipeline instanceof Builder && $this->builder === $this->pipeline) {
```
alcaeus: Since the === operator is optimised we don't need to check the pipeline type first. Checking whether it's identical to the builder should be sufficient.

franmomu: Thanks! I started doing this check at the beginning of the method and then didn't update it

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/LookupTest.php
```diff
@@ -87,27 +88,83 @@ public function testLookupStageWithPipeline(): void
             ],
         ];
 
-        $this->assertEquals($expectedPipeline, $builder->getPipeline());
+            $this->assertEquals($expectedPipeline, $builder->getPipeline(false));
```
alcaeus: Indentation seems off here. Also, I'd expect this to work just the same with `true` being passed, as the parameter was introduced for nested pipelines only.

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/LookupTest.php
```diff
@@ -87,27 +88,83 @@ public function testLookupStageWithPipeline(): void
...
+                                'username' => 1,
+                            ],
+                        ],
                     ],
                 ],
-            ]);
+            ],
+        ];
+
+        $this->assertEquals($expectedPipeline, $builder->getPipeline(false));
```
alcaeus: This and following `getPipeline` calls should not set `$applyFilters` to false.

alcaeus (approved):
A couple of notes on indentation and not applying filters in pipeline calls, but LGTM besides those.

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/LookupTest.php
franmomu: 😕 fixed

alcaeus:
Thanks @aturki and @franmomu!

Issue #2487, opened 2022-12-11
Custom _id for GridFS documents


andrey-tech:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         |  no
| BC Break    | no
| Version | 2.4.x

#### Summary

I would like to have a way to set custom file document identifier (`_id`) for GridFS document, when I insert a new file into GridFS bucket by methods `uploadFromFile()` or `uploadFromStream()` of [GridFSRepository](https://github.com/doctrine/mongodb-odm/blob/2.4.x/lib/Doctrine/ODM/MongoDB/Repository/GridFSRepository.php) interface.

This important functionality is required when migrating from MongoDB using old `doctrine/mongodb-odm` **v1.x** to a new MongoDB using `doctrine/mongodb-odm` **v2.x**. I need to have the same document identifiers (`_id`) for GridFS documents in old and new MongoDB.

Used MongoDB library (`mongodb/mongodb:^1.2.0`) supports this functionality in method `uploadFromStream()` of class [Bucket](https://github.com/mongodb/mongo-php-library/blob/master/src/GridFS/Bucket.php).

I suggest to add the parameter `_id` as a new property of class [UploadOptions](https://github.com/doctrine/mongodb-odm/blob/2.4.x/lib/Doctrine/ODM/MongoDB/Repository/UploadOptions.php) and update method `prepareOptions()` of class [DefaultGridFSRepository](https://github.com/doctrine/mongodb-odm/blob/2.4.x/lib/Doctrine/ODM/MongoDB/Repository/DefaultGridFSRepository.php) to support new option `_id`. So, no big changes are required.

```php
final class UploadOptions
{
    /** @var object|null */
    public $metadata;

    /** @var int|null */
    public $chunkSizeBytes;

    /** @var mixed */
    public $_id;
}
```

malarzm:
@andrey-tech this sounds like a good idea! While we're tweaking `UploadOptions` we could also add this missing options

https://github.com/mongodb/mongo-php-library/blob/4c090a88821db2665bf77a0742ae81efec121380/src/GridFS/Bucket.php#L610-L611

@andrey-tech would you be up for a PR?

djkoza:
We need this feature too! A nice improvement would be converting values to a database via mapping type.
With the @andrey-tech tweak, we need to self-convert the UUID object to the Binary object.

[...]

PR #2489, opened 2022-12-12
 Custom _id for GridFS documents (#2487)

andrey-tech:
|      Q       |   A
|------------- | -----------
| Type         |  feature
| BC Break     | no
| Fixed issues | #2487
| Version        | 2.5.x

#### Summary

Added two missing options to `UploadOptions`:
- `id` (mixed): File document identifier. Defaults to a new ObjectId.
~~- `disableMD5` (boolean): When true, no MD5 sum will be generated for the stored file. Defaults to `false`.~~

~~The `mongodb/mongodb` **^1.4.0** is required for option `disableMD5`.~~


andrey-tech:
It might make sense to change the property name from `_id` to `id` in `UploadOptions` for PHP CodeSniffer with Slevomat coding standard rules.

andrey-tech:
1) Renamed `_id` to `id` in `UploadOptions` for PHP CodeSniffer with SlevomatCodingStandard rules.
2) Added tests for new options `id` and `disableMD5`.

franmomu:
Hi @andrey-tech, thanks for the PR! since this is a feature, can you please target to `2.5.x` instead?

andrey-tech:
@franmomu Done.

lib/Doctrine/ODM/MongoDB/Repository/DefaultGridFSRepository.php
```diff
@@ -108,6 +110,14 @@ private function prepareOptions(?UploadOptions $uploadOptions = null): array
         $chunkSizeBytes = $uploadOptions->chunkSizeBytes ?: $this->class->getChunkSizeBytes();
         $options        = [];
 
+        if ($uploadOptions->id !== null) {
+            $options['_id'] = $this->class->getDatabaseIdentifierValue($uploadOptions->id);
```
malarzm: Would be great to have a test that covers type transformation

malarzm (approved):
Thanks for the PR! Mind adding one more test please? :)

lib/Doctrine/ODM/MongoDB/Repository/DefaultGridFSRepository.php
andrey-tech: MongoDB documentation lists `ObjectId` as just one of the common variants for the [_id field](https://docs.mongodb.com/manual/core/document/#the-_id-field).  So I removed type transformation from method `prepareOptions()` of class `DefaultGridFSRepository` for more flexibility. Test that covers this type transformation not needed anymore.

lib/Doctrine/ODM/MongoDB/Repository/UploadOptions.php
```diff
@@ -6,9 +6,15 @@
 
 final class UploadOptions
 {
-    /** @var object|null */
-    public $metadata;
+    /** @var mixed */
+    public $id;
+
+    /** @var bool|null */
+    public $disableMD5;
```
alcaeus: Do you have a specific use-case for the `disableMD5` option? The reason I omitted this was that the [specification](https://github.com/mongodb/specifications/blob/master/source/gridfs/gridfs-spec.rst) marks the option as deprecated and it will be removed from the PHP driver in the next major release. I would prefer not adding it in now, as we'd have to do another BC break for 3.x then (as opposed to breaking it once for people migrating to ODM 2.0).

composer.json
```diff
@@ -32,7 +32,7 @@
         "doctrine/persistence": "^2.4 || ^3.0",
         "friendsofphp/proxy-manager-lts": "^1.0",
         "jean85/pretty-package-versions": "^1.3.0 || ^2.0.1",
-        "mongodb/mongodb": "^1.2.0",
+        "mongodb/mongodb": "^1.4.0",
```
alcaeus: Might as well bump this to 1.15 while you're at it. 1.4 is ancient and no longer supported, we should encourage people to upgrade :)

alcaeus (changes requested):

lib/Doctrine/ODM/MongoDB/Repository/UploadOptions.php
malarzm: This one's on me (https://github.com/doctrine/mongodb-odm/issues/2487#issuecomment-1345624957) I didn't know it's deprecated :)

composer.json
andrey-tech: I don't mind :)

lib/Doctrine/ODM/MongoDB/Repository/UploadOptions.php
alcaeus: Hehe, yes. We deprecated it since using MD5 is forbidden when being compliant with [FIPS](https://en.wikipedia.org/wiki/Federal_Information_Processing_Standards), and personally I wouldn't rely on MD5 for checksums anymore. @andrey-tech do you mind dropping it? I don't think we should introduce the option at this point, especially since we haven't had anyone ask for it.

andrey-tech: @alcaeus I don't mind.

andrey-tech: @malarzm Do you mind dropping it?

composer.json
andrey-tech: @malarzm Do you mind to set `^1.15.0`?

lib/Doctrine/ODM/MongoDB/Repository/UploadOptions.php
franmomu: @andrey-tech since it's deprecated and it was not added on purpose, we should not add it

andrey-tech: The deprecated option `disableMD5` has been removed from this PR.

composer.json
andrey-tech: I set `mongodb/mongodb` to `^1.15.0`.

alcaeus (approved):
LGTM.

We should add a test for using an identifier with a custom type, but I'm fine deferring that to a separate pull request.

IonBazan:
@alcaeus I think many people are still stuck on older versions and I wouldn't force them to update to v1.15 yet, especially that it requires updating the PHP extension via `pecl` which may not be an easy option, depending on their infrastructure.
There are still number of installs of older version. Would you consider relaxing the requirement to `^1.8.0` or `^1.10.0` instead?
image

https://packagist.org/packages/mongodb/mongodb/stats#major/1

alcaeus:
Bumping to `^1.10` seems sensible in that case. I'm aware that this will require folks to upgrade their extension, but considering that we don't support anything older than 1.15, they should've upgraded already :)

.github/workflows/continuous-integration.yml
```diff
@@ -37,7 +37,7 @@ jobs:
           - dependencies: "lowest"
             php-version: "7.2"
             mongodb-version: "3.6"
-            driver-version: "1.5.0"
+            driver-version: "1.10.0"
```
alcaeus: mongodb/mongodb `1.10` requires `ext-mongodb` `1.11` - we've corrected this with 1.15 to restore version number parity.

IonBazan: Noted, thanks! Some builds are canceled because of `ubuntu-18.04` being deprecated but 20.04 doesn't play well with the `mongodb-labs/drivers-evergreen-tools` action. Let's look into that in a separate PR.

alcaeus: I'll take a look 👍

alcaeus:
Thanks @andrey-tech!

PR #2492, opened 2022-12-20
Ignore SA issues

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | task
| BC Break     | no
| Fixed issues | 

#### Summary

These new issues appeared because Symfony `Command::getHelper()` changed the return type from `mixed` to `HelperInterface`.

We can ignore these issues and I'll see if I can make something similar to https://github.com/doctrine/orm/pull/8524 to remove the use of the helper.


malarzm (approved):

malarzm:
Thanks @franmomu!

PR #2498, opened 2022-12-27
Make `doctrine/annotations` dependency optional

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | task
| BC Break     | no
| Fixed issues | 

#### Summary

Similar to https://github.com/doctrine/orm/pull/8787, we have PHP attributes and XML mapping, there is no need to be a hard dependency.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,6 @@
     "require": {
         "php": "^8.1",
         "ext-mongodb": "^1.11",
-        "doctrine/annotations": "^1.12 || ^2.0",
         "doctrine/cache": "^1.11 || ^2.0",
         "doctrine/collections": "^1.5 || ^2.0",
         "doctrine/event-manager": "^1.0 || ^2.0",
@@ -39,6 +38,7 @@
     },
     "require-dev": {
         "ext-bcmath": "*",
+        "doctrine/annotations": "^1.12 || ^2.0",
         "doctrine/coding-standard": "^12.0",
         "jmikola/geojson": "^1.0",
         "phpbench/phpbench": "^1.0.0",
@@ -49,7 +49,11 @@
         "symfony/cache": "^5.4 || ^6.0 || ^7.0",
         "vimeo/psalm": "^5.9.0"
     },
+    "conflict": {
+        "doctrine/annotations": "<1.12 || >=3.0"
+    },
     "suggest": {
+        "doctrine/annotations": "For annotation mapping support",
         "ext-bcmath": "Decimal128 type support"
     },
     "autoload": {
--- a/UPGRADE-2.7.md
+++ b/UPGRADE-2.7.md
@@ -8,3 +8,8 @@
 ## doctrine/persistence
 
 * MongoDB ODM 2.7 requires `doctrine/persistence` 3.2 or newer.
+
+## `doctrine/annotations` is optional
+
+ODM no longer requires `doctrine/annotations` to be installed. If you're using
+annotations for mapping, you will need to install `doctrine/annotations`.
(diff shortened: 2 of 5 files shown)


composer.json
```diff
@@ -49,6 +49,9 @@
         "symfony/cache": "^4.4 || ^5.0 || ^6.0",
         "vimeo/psalm": "^4.30.0"
     },
+    "conflict": {
+        "doctrine/annotations": "<1.12 || >= 3.0"
+    },
     "suggest": {
         "ext-bcmath": "Decimal128 type support"
     },
```
malarzm, line 58: `doctrine/annotations` could be added to the `suggest` section

composer.json
```diff
@@ -49,6 +49,9 @@
         "symfony/cache": "^4.4 || ^5.0 || ^6.0",
         "vimeo/psalm": "^4.30.0"
     },
+    "conflict": {
+        "doctrine/annotations": "<1.12 || >= 3.0"
```
malarzm: nitpicking, but one operator has an extra space while the other one does not ;)

malarzm (commented):
This warrants an UPDATE file entry IMO. On the other hand I'm thinking if it's worth doing in a minor version.

composer.json
SenseException: Having them in the suggestions can help if people read them and still need annotations. Older PHP versions still need them.

SenseException (approved):
After adding annotations to composer-suggest this is :+1:

composer.json
franmomu: 👍 added

IonBazan (approved):

IonBazan:
This will currently affect the `AttributeDriver` too: https://github.com/doctrine/mongodb-odm/blob/2.4.x/lib/Doctrine/ODM/MongoDB/Mapping/Driver/AttributeDriver.php as it relies on `Doctrine\Common\Annotations\Reader`. How about we refactor that first to be completely independent from `doctrine/annotations`? We could even drop that package in v3.0.

greg0ire:
Here is how you could proceed: https://github.com/doctrine/orm/pull/9671

franmomu:
> This will currently affect the `AttributeDriver` too: https://github.com/doctrine/mongodb-odm/blob/2.4.x/lib/Doctrine/ODM/MongoDB/Mapping/Driver/AttributeDriver.php as it relies on `Doctrine\Common\Annotations\Reader`. How about we refactor that first to be completely independent from `doctrine/annotations`? We could even drop that package in v3.0.

👍 I thought they were independents

alcaeus (dismissed):
Slapping a "request changes" on this for good measure. As noted before, we first need to make the attribute driver work independently of annotations, then we can make the dependency optional.

That said, I'm very much in favour of removing the hard requirement, and I would also do so in a minor version. We can then think of dropping support for annotations entirely in 3.x.

composer.json
```diff
@@ -49,7 +49,11 @@
         "symfony/cache": "^5.4 || ^6.0 || ^7.0",
         "vimeo/psalm": "^5.9.0"
     },
+    "conflict": {
+        "doctrine/annotations": "<1.12 || >=3.0"
```
malarzm, line 53: FTR conflicting here is IMO OK just because we control the lib and said it's feature complete. Otherwise I'd be against having a conflict section as one could use the lib we're conflicting with, but not use it with ODM.

malarzm (approved):

malarzm:
I took the liberty and slapped a "BC Break" label for visibility in the release notes.

Thanks @franmomu!

franmomu:
> I took the liberty and slapped a "BC Break" label for visibility in the release notes.

👍 thanks @malarzm

Issue #2063, opened 2019-09-17
Support new aggregation features


alcaeus:
|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

This issue serves as an epic to track various aggregation features that are not supported in ODM yet.

##### New aggregation pipeline stages

* [`$currentOp`](https://docs.mongodb.com/manual/reference/operator/aggregation/currentOp/#pipe._S_currentOp)
* [`$listLocalSessions`](https://docs.mongodb.com/manual/reference/operator/aggregation/listLocalSessions/#pipe._S_listLocalSessions)
* [`$listSessions`](https://docs.mongodb.com/manual/reference/operator/aggregation/listSessions/#pipe._S_listSessions)
* [`$merge`](https://docs.mongodb.com/manual/reference/operator/aggregation/merge/#pipe._S_merge)
* [`$planCacheStats`](https://docs.mongodb.com/manual/reference/operator/aggregation/planCacheStats/#pipe._S_planCacheStats)
* [`$sortByCount`](https://docs.mongodb.com/manual/reference/operator/aggregation/sortByCount/#pipe._S_sortByCount)

In addition to these stages, the new `$replaceWith`, `$set`, and `$unset` stages are aliases for `$replaceRoot`, `$addFields`, and `$project`, respectively. We should consider adding those to the builder as well.

##### New aggregation operators

* [`$arrayToObject`](https://docs.mongodb.com/manual/reference/operator/aggregation/arrayToObject/#exp._S_arrayToObject)
* [`$dateFromParts`](https://docs.mongodb.com/manual/reference/operator/aggregation/dateFromParts/#exp._S_dateFromParts)
* [`$dateFromString`](https://docs.mongodb.com/manual/reference/operator/aggregation/dateFromString/#exp._S_dateFromString)
* [`$dateToParts`](https://docs.mongodb.com/manual/reference/operator/aggregation/dateToParts/#exp._S_dateToParts)
* [`$ltrim`](https://docs.mongodb.com/manual/reference/operator/aggregation/ltrim/#exp._S_ltrim)
* [`$mergeObjects`](https://docs.mongodb.com/manual/reference/operator/aggregation/mergeObjects/#exp._S_mergeObjects)
* [`$objectToArray`](https://docs.mongodb.com/manual/reference/operator/aggregation/objectToArray/#exp._S_objectToArray)
[...]

PR #2499, opened 2023-01-06
Support for some new aggregation features

aturki:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | #2063 (Partially)

#### Summary

Provides implementation of some aggregation operators

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Doctrine\ODM\MongoDB\Aggregation;
 
 use BadMethodCallException;
+use Doctrine\ODM\MongoDB\Aggregation\Operator\GenericOperatorsInterface;
 use Doctrine\ODM\MongoDB\DocumentManager;
 use Doctrine\ODM\MongoDB\Mapping\ClassMetadata;
 use Doctrine\ODM\MongoDB\Persisters\DocumentPersister;
@@ -23,7 +24,7 @@ use function substr;
 /**
  * Fluent interface for building aggregation pipelines.
  */
-class Expr
+class Expr implements GenericOperatorsInterface
 {
     private DocumentManager $dm;
 
@@ -1788,10 +1789,10 @@ class Expr
      *
      * @see https://docs.mongodb.com/manual/reference/operator/aggregation/round/
      *
-     * @param mixed|self $number
-     * @param mixed|self $place
+     * @param mixed|self      $number
+     * @param mixed|self|null $place
      */
-    public function round($number, $place): self
+    public function round($number, $place = null): self
     {
         return $this->operator('$round', [$number, $place]);
     }
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
@@ -299,7 +299,7 @@ class Expr
     }
 
     /**
-     * Converts an expression object into an array, recursing into nested items
+     * Converts an expression object into an array, recursing into nested items.
      *
      * For expression objects, it calls getExpression on the expression object.
      * For arrays, it recursively calls itself for each array item. Other values
@@ -452,7 +452,7 @@ class Expr
     }
 
     /**
-     * Returns a new expression object
+     * Returns a new expression object.
      */
     public function expr(): self
     {
@@ -624,8 +624,8 @@ class Expr
      *
      * @see https://docs.mongodb.com/manual/reference/operator/aggregation/indexOfArray/
      *
-     * @param mixed|self $arrayExpression  Can be any valid expression as long as it resolves to an array.
-     * @param mixed|self $searchExpression Can be any valid expression.
+     * @param mixed|self $arrayExpression  can be any valid expression as long as it resolves to an array
+     * @param mixed|self $searchExpression can be any valid expression
      * @param mixed|self $start            Optional. An integer, or a number that can be represented as integers (such as 2.0), that specifies the starting index position for the search. Can be any valid expression that resolves to a non-negative integral number.
      * @param mixed|self $end              An integer, or a number that can be represented as integers (such as 2.0), that specifies the ending index position for the search. Can be any valid expression that resolves to a non-negative integral number.
      */
@@ -650,8 +650,8 @@ class Expr
      *
      * @see https://docs.mongodb.com/manual/reference/operator/aggregation/indexOfBytes/
      *
-     * @param mixed|self      $stringExpression    Can be any valid expression as long as it resolves to a string.
-     * @param mixed|self      $substringExpression Can be any valid expression as long as it resolves to a string.
+     * @param mixed|self      $stringExpression    can be any valid expression as long as it resolves to a string
+     * @param mixed|self      $substringExpression can be any valid expression as long as it resolves to a string
      * @param string|int|null $start               An integral number that specifies the starting index position for the search. Can be any valid expression that resolves to a non-negative integral number.
      * @param string|int|null $end                 An integral number that specifies the ending index position for the search. Can be any valid expression that resolves to a non-negative integral number.
      */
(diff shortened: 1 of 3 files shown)


lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
```diff
@@ -1600,7 +1600,9 @@ private function requiresCurrentField(?string $method = null): void
         }
     }
 
-    /** @throws BadMethodCallException If there is no current switch operator. */
+    /**
+     * @throws BadMethodCallException If there is no current switch operator.
+     */
```
alcaeus: CI is complaining about this change, it needs to be reverted back:
```suggestion
    /** @throws BadMethodCallException If there is no current switch operator. */
```

lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
```diff
@@ -1655,4 +1657,387 @@ public function then($expression): self
...
+    /**
+     * Rounds a number to a whole integer or to a specified decimal place.
+     *
+     * The <number> argument can be any valid expression as long as it resolves
+     * to a number.
+     *
+     * @see https://docs.mongodb.com/manual/reference/operator/aggregation/round/
+     *
+     * @param int|float|self $number
+     * @param int            $place
```
alcaeus: Both of these annotations are incorrect: they should use `mixed|self` as they accept any valid expression. A string can also be a valid expression, as it can resolve to a field name (i.e. `'$myNumberField'`).

lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
```diff
@@ -1655,4 +1657,387 @@ public function then($expression): self
...
+    }
+
+    /**
+     * Removes whitespace characters, including null, or the specified characters from
+     * the beginning and end of a string.
+     *
+     * @see https://docs.mongodb.com/manual/reference/operator/aggregation/trim/
+     *
+     * @param string|self $input
+     * @param string|null $chars
```
alcaeus: Same here, these should be `mixed|self`, as they accept any expression that resolves to a string. For example, `['$toString' => 5]` would be a valid value for `$input` as it resolves to a string when executed.

lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
```diff
@@ -1655,4 +1657,387 @@ public function then($expression): self
...
+        return $this->operator('$trim', [$input, $chars]);
+    }
+
+    /**
+     * Removes whitespace characters, including null, or the specified characters from
+     * the beginning and end of a string.
+     *
+     * @see https://docs.mongodb.com/manual/reference/operator/aggregation/ltrim/
+     */
+    public function ltrim(string $input, ?string $chars = null): self
```
alcaeus: Parameter types are incorrect.

lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
```diff
@@ -1655,4 +1657,387 @@ public function then($expression): self
...
+    {
+        return $this->operator('$ltrim', [$input, $chars]);
+    }
+
+    /**
+     * Removes whitespace characters, including null, or the specified characters from the end of a string.
+     *
+     * @see https://docs.mongodb.com/manual/reference/operator/aggregation/rtrim/
+     */
+    public function rtrim(string $input, ?string $chars = null): self
```
alcaeus: Parameter types are incorrect.

lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
```diff
@@ -1655,4 +1657,387 @@ public function then($expression): self
...
+    }
+
+    /**
+     * Returns the inverse tangent (arc tangent) of y / x in radians, where y and x are the first and second values passed to the expression respectively.
+     *
+     * @see https://docs.mongodb.com/manual/reference/operator/aggregation/atan2/
+     *
+     * @param mixed|self $expression
+     */
+    public function atan2($expression): self
```
alcaeus: `atan2` takes two parameters according to the documentation, so the signature for this operator is wrong.

lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
```diff
@@ -1655,4 +1657,387 @@ public function then($expression): self
...
+     * Converts a value to a specified type.
+     *
+     * @see https://docs.mongodb.com/manual/reference/operator/aggregation/convert/
+     *
+     * @param mixed|self      $expression
+     * @param int|string|self $to
+     * @param mixed|self|null $expression
+     * @param mixed|self|null $expression
+     */
+    public function convert($expression, $to, ?Expr $onError = null, ?Expr $onNull = null): self
```
alcaeus: Multiple issues here:
* `$to` can be any valid expression
* `$onError` and `$onNull` can be any valid expression, including a string.

In essence, all parameters here should accept `mixed|self`, and we can include `|null` for `$onError` and `$onNull`.

lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
```diff
@@ -1655,4 +1657,387 @@ public function then($expression): self
...
+    {
+        return $this->operator('$convert', [$expression, $to, $onError, $onNull]);
```
alcaeus: This would result in the following expression:
```
{
    $convert: [ <expression>, <expression>, <expression>, <expression>]
}
```

This is incorrect, as `$convert` expects an object containing `input` and `to` keys, along with optional `onError` and `onNull` keys as per the documentation.

Instead of passing values as an array, you should pass a hash containing the right keys. I'd suggest omitting `onError` and `onNull` if they are `null` to make sure the default behaviour is not changed. I'd assume passing a `null` value for `onError` would cause the expression to return a `null` value if the conversion is not successful, as opposed to aborting the pipeline with an error.

lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
```diff
@@ -1655,4 +1657,387 @@ public function then($expression): self
...
+    /**
+     * Converts value to a boolean.
+     *
+     * @see https://docs.mongodb.com/manual/reference/operator/aggregation/toBool/
+     *
+     * @param mixed|self $expression
+     */
+    public function toBool($expression): self
+    {
+        return $this->convert($expression, 'bool');
```
alcaeus: I suggest using `$this->operator('$toBool', $expression)` here and for all other `to...` operators.

lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
```diff
@@ -1655,4 +1657,387 @@ public function then($expression): self
...
+     *
+     * @see https://docs.mongodb.com/manual/reference/operator/aggregation/toString/
+     *
+     * @param mixed|self $expression
+     */
+    public function toString($expression): self
+    {
+        return $this->convert($expression, 'string');
+    }
+ }
```
alcaeus: There's an extra space here that phpcs is complaining about:
```suggestion
}
```

alcaeus (changes requested):
Thank you for adding those operators. I've left some comments, but have not yet reviewed the tests.

I'll also note that methods in the class are currently sorted alphabetically - I'd appreciate it if you could keep the ordering and move newly added methods to the correct place.

alcaeus (approved):

alcaeus:
I've extracted an interface from the `Expr` class to ensure that new methods are added to the abstract `Operator` stage. In addition, I've removed the `__call` magic. The idea behind the egg builder is to provide an expressive interface to help people along, and the call magic does the opposite. It was a fallback for when we couldn't have methods named `switch`, `default`, etc.

lib/Doctrine/ODM/MongoDB/Aggregation/Operator/GenericOperatorsInterface.php
```diff
@@ -0,0 +1,1356 @@
...
+     * Returns the absolute value of a number.
+     *
+     * The <number> argument can be any valid expression as long as it resolves
+     * to a number.
+     *
+     * @see https://docs.mongodb.com/manual/reference/operator/aggregation/abs/
+     *
+     * @param mixed|Expr $number
+     */
+    public function abs($number): self;
```
IonBazan, line 28: Here and all methods below - shouldn't we make it something like:
```suggestion
     * @param mixed|Expr $number
     */
    public function abs($number): static;
```

Since we are targeting PHP 7.4+, we can use `static` typehint.

alcaeus: I was considering that, but this was added in 8.0, not 7.4: https://php.watch/versions/8.0/static-return-type.

Since the interface is internal, we can make this change once we drop support for 7.4. 👍

IonBazan: Makes sense, thanks!

IonBazan (approved):

IonBazan:
Thanks @aturki! 🚀

PR #2501, opened 2023-01-19
Allow `doctrine/instantiator` 2

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | 

#### Summary

There shouldn't be any BC break since the changes involved are only about modernizing the code: https://github.com/doctrine/instantiator/releases/tag/2.0.0

Update: for example, in https://github.com/doctrine/mongodb-odm/actions/runs/3959728289/jobs/6782933801#step:8:39 version 2 is installed.


SenseException (approved):
The tests are installing Instantiator 1 and 2 and are green.

IonBazan (approved):

PR #2502, opened 2023-02-05
Decouple AttributeDriver from AnnotationDriver

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | 

#### Summary

This is a start so we can discuss what to do to decouple `AttributeDriver` from annotations, there are some BC breaks involved, so we need to see if this should be done in `2.5` or wait for `3.0`.

The first one is that `AttributeDriver` no longer extends `AnnotationDriver`, I think this one is fine to do it in `2.5` (this is what this PR does at the moment).

But if we want to totally remove the dependency from `AttributeDriver` we should probably modify the constructor to remove the `?Reader $reader = null` parameter, that would mean that the `$reader` property would always be of type `AttributeReader` (which would be also modified to not implement the `Reader` interface).

So, should we go all the way in? or postpone it for `3.0`?


UPGRADE-2.5.md
```diff
@@ -1,5 +1,10 @@
 # UPGRADE FROM 2.4 to 2.5
 
+## Backward compatibility breaks
+
+* `Doctrine\ODM\MongoDB\Mapping\Driver\AttributeDriver` no longer extends from
```
IonBazan:
```suggestion
* `Doctrine\ODM\MongoDB\Mapping\Driver\AttributeDriver` no longer extends
```

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AttributeDriver.php
```diff
@@ -5,16 +5,368 @@
...
+use function MongoDB\BSON\fromJSON;
+use function MongoDB\BSON\toPHP;
+use function trigger_deprecation;
 
 /**
- * The AnnotationDriver reads the mapping metadata from docblock annotations.
+ * The AtttributeDriver reads the mapping metadata from attributes.
  */
-class AttributeDriver extends AnnotationDriver
+class AttributeDriver extends CompatibilityAnnotationDriver
```
stof: why would it extends the CompatibilityAnnotationDriver ? the attribute driver has no reason to keep extending `octrine\Persistence\Mapping\Driver\AnnotationDriver` either

franmomu:
I've created https://github.com/doctrine/mongodb-odm/pull/2596 on top of this PR to test the attribute driver

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AttributeDriver.php
```diff
@@ -5,16 +5,372 @@
...
+        if ($reader !== null) {
+            trigger_deprecation(
+                'doctrine/mongodb-odm',
+                '2.7',
+                'Passing a $reader parameters to %s is deprecated',
+                __METHOD__,
+            );
+        }
+
+        $this->reader = $reader ?? new AttributeReader();
```
malarzm, line 59: Shall we have a type check on `$reader` being an `AttributeReader`?

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AttributeDriver.php
```diff
@@ -5,16 +5,372 @@
...
+        foreach ($classAttributes as $attribute) {
+            if ($attribute instanceof ODM\AbstractDocument) {
+                return false;
+            }
+        }
+
+        return true;
+    }
+
+    public function loadMetadataForClass($className, \Doctrine\Persistence\Mapping\ClassMetadata $metadata): void
```
malarzm: I think it's rather unfortunate to have the code duplicated :/ Is there any other option?

tests/Doctrine/ODM/MongoDB/Tests/Mapping/AbstractAnnotationDriverTestCase.php
```diff
@@ -173,7 +176,9 @@ public static function provideClassCanBeMappedByOneAbstractDocument(): ?Generato
              * @ODM\Document()
              * @ODM\EmbeddedDocument
              */
-            new class () {
+            new #[ODM\Document]
+        #[ODM\EmbeddedDocument]
```
malarzm: Oh boy this is ugly 🙈 Not the point of this PR obviously, just pointing out ;)

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AttributeDriver.php
franmomu: what do you mean? the constructor is receiving a `?Reader` (which is deprecated), if no argument is passed it will be an `AttributeReader`.

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AttributeDriver.php
franmomu: Yeah I know 😞, what I was thinking is after we do this, we can make `doctrine/annotations` optional and then we can also deprecate `AnnotationDriver`.

tests/Doctrine/ODM/MongoDB/Tests/Mapping/AbstractAnnotationDriverTestCase.php
franmomu: it is! I've aligned the attributes, not sure if there is something else we can do

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AttributeDriver.php
malarzm: I was thinking about throwing an exception in case somebody would pass a `AnnotationReader` by accident :)

tests/Doctrine/ODM/MongoDB/Tests/Mapping/AbstractAnnotationDriverTestCase.php
malarzm: I'm not sure either :D but most probably it's not worth discussing in CS repo and if we can have it at least aligned it's OK :)

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AttributeDriver.php
malarzm: Maybe we could have a trait with shared code? Or would it be possible to have AnnotationDriver extend AttributeDriver? My biggest concern is that we're far from removing AnnotationDriver as we've never talked 3.0

alcaeus: Only other option I can think of is to reverse the inheritance and have `AnnotationDriver` extend `AttributeDriver`, which again would remove the code duplication.

While we've never talked about ORM 3.0, I don't think it should be a "big thing" again. Instead, it should be released at the same time as a minor release, dropping all deprecated paths from that minor release.

malarzm: > While we've never talked about ORM 3.0, I don't think it should be a "big thing" again. Instead, it should be released at the same time as a minor release, dropping all deprecated paths from that minor release.

I'd be fine with this approach, it works well for Symfony :+1:

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AttributeDriver.php
franmomu: ahm, the thing is that passing any `Reader` (including `AnnotationReader`) would work, I haven't changed that to maintain BC

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AnnotationDriver.php
```diff
@@ -6,36 +6,12 @@
...
-use function interface_exists;
-use function is_array;
-use function MongoDB\BSON\fromJSON;
-use function MongoDB\BSON\toPHP;
-use function trigger_deprecation;
 
 /**
  * The AnnotationDriver reads the mapping metadata from docblock annotations.
  */
-class AnnotationDriver extends CompatibilityAnnotationDriver
```
franmomu: This could also be BC break, that class exists to have compatibility with `doctrine/persistence` 2 and 3

We can drop support of version 2 or leave it like this (`AnnotationDriver` would stop extending `Doctrine\Persistence\Mapping\Driver\AnnotationDriver`)

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AttributeDriver.php
franmomu: Done, but something else came up https://github.com/doctrine/mongodb-odm/pull/2502#discussion_r1430464832

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AnnotationDriver.php
malarzm: We've decided in https://github.com/doctrine/DoctrineMongoDBBundle/pull/824 to drop support for `doctrine/persistence` v2 :)

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AttributeDriver.php
malarzm: But wouldn't the driver do nothing with anything else than an `AttributeReader`? Or am I missing something?

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AnnotationDriver.php
franmomu: Let's deal with that in https://github.com/doctrine/mongodb-odm/pull/2603

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AttributeDriver.php
franmomu: I added these methods: https://github.com/doctrine/mongodb-odm/pull/2502/files#diff-ac3d9c518c39d96ef76766a33e63f5115a72783ae78bc99bb4b70405a04e718eR389-R416:
```php
    /** @return object[] */
    private function getClassAttributes(ReflectionClass $class): array
    {
        if ($this->reader instanceof AttributeReader) {
            return $this->reader->getClassAttributes($class);
        }

        return $this->reader->getClassAnnotations($class);
    }

    /** @return object[] */
    private function getMethodAttributes(ReflectionMethod $method): array
    {
        if ($this->reader instanceof AttributeReader) {
            return $this->reader->getMethodAttributes($method);
        }

        return $this->reader->getMethodAnnotations($method);
    }

    /** @return object[] */
    private function getPropertyAttributes(ReflectionProperty $property): array
    {
        if ($this->reader instanceof AttributeReader) {
            return $this->reader->getPropertyAttributes($property);
        }

        return $this->reader->getPropertyAnnotations($property);
    }
```
to maintain BC in case someone is passing a `Reader` object

alcaeus (approved):

IonBazan (approved):

GromNaN (approved):

franmomu:
thanks for the reviews!

malarzm:
Sorry I didn't manage to get back in time. Anyway, happy new year folks!

Issue #2509, opened 2023-03-08
Ensure PHPUnit 10 compatibility


IonBazan:
This is to track PHPUnit 10 compatibility with the test suite:

- [x] Rename abstract test classes ending with `Test` #2507 
- [x] Make all data providers static
- [x] Replace usages of deprecated and deleted test methods
- [x] Migrate XML configuration
- [x] Get rid of deprecations in tests

PR #2512, opened 2023-03-13
Use static data providers

IonBazan:
#### Summary

Makes all test data providers `static`. This is to fix PHPUnit 10 deprecations.

See #2509

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -1,33 +1,30 @@
 <?xml version="1.0" encoding="UTF-8"?>
-
-<phpunit backupGlobals="false"
-         backupStaticAttributes="false"
-         colors="false"
-         convertErrorsToExceptions="true"
-         convertNoticesToExceptions="true"
-         convertWarningsToExceptions="true"
-         failOnRisky="true"
-         stopOnFailure="false"
-         bootstrap="tests/bootstrap.php"
+<phpunit
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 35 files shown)


alcaeus (approved):
LGTM. Not sure if the Psalm errors are related, they don't really look related.

phpunit.xml.dist
```diff
@@ -1,33 +1,30 @@
...
-         bootstrap="tests/bootstrap.php"
+<phpunit
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        backupGlobals="false"
+        backupStaticProperties="false"
+        colors="false"
+        failOnRisky="true"
+        stopOnFailure="false"
+        bootstrap="tests/bootstrap.php"
+        xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/10.0/phpunit.xsd"
```
SenseException:
```suggestion
        xsi:noNamespaceSchemaLocation="vendor/phpunit/phpunit/phpunit.xsd"
```

SenseException (changes requested):

SenseException (approved):
I don't see a connection to psalm in this PR, but neither do I see it in the current breaking build on 2.5.x

franmomu (approved):
Looks like a bug in psalm that after `assertCount`, the type is changed to `Countable` (removing `ArrayAccess`), but apparently the psalm version is the same as in `2.5.x`, so maybe something has changed in phpunit 10

PR #2513, opened 2023-03-17
Support new aggregation pipeline stages in builder

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | 

#### Summary

This pull requests adds a number of previously unsupported aggregation pipeline stages to the builder:
* `$densify`
* `$fill`
* `$merge`
* `$replaceWith`
* `$set`
* `$unset`
* `$unionWith`

It also leverages templates to allow `Builder::addStage` to indicate that it's returning the exact stage that was given, simplifying code in the Builder class.

This leaves the `$search`, `$searchMeta`, and `$setWindowFields` stages left to be implemented. I'll create separate pull requests for those due to their complexity (especially `$search`). I'll also have a separate pull request to support new pipeline operators.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
@@ -29,6 +29,8 @@ use function trigger_deprecation;
  * Fluent interface for building aggregation pipelines.
  *
  * @psalm-import-type SortShape from Sort
+ * @psalm-import-type StageExpression from Stage
+ * @psalm-type PipelineExpression = list<StageExpression>
  */
 class Builder
 {
@@ -269,6 +271,7 @@ class Builder
      * given.
      *
      * @return array<array<string, mixed>>
+     * @psalm-return PipelineExpression
      */
     // phpcs:enable Squiz.Commenting.FunctionComment.ExtraParamComment
     public function getPipeline(/* bool $applyFilters = true */): array
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
@@ -443,6 +443,20 @@ class Builder
     }
 
     /**
+     * Writes the results of the aggregation pipeline to a specified collection.
+     * The $merge operator must be the last stage in the pipeline.
+     *
+     * @see https://www.mongodb.com/docs/rapid/reference/operator/aggregation/merge/
+     */
+    public function merge(): Stage\Merge
+    {
+        $stage = new Stage\Merge($this, $this->dm);
+        $this->addStage($stage);
+
+        return $stage;
+    }
+
+    /**
      * Takes the documents returned by the aggregation pipeline and writes them
      * to a specified collection. This must be the last stage in the pipeline.
      *
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
@@ -83,9 +83,8 @@ class Builder
     public function addFields(): Stage\AddFields
     {
         $stage = new Stage\AddFields($this);
-        $this->addStage($stage);
 
-        return $stage;
+        return $this->addStage($stage);
     }
 
     /**
@@ -103,9 +102,8 @@ class Builder
     public function bucket(): Stage\Bucket
     {
         $stage = new Stage\Bucket($this, $this->dm, $this->class);
-        $this->addStage($stage);
 
-        return $stage;
+        return $this->addStage($stage);
     }
 
     /**
@@ -125,9 +123,8 @@ class Builder
     public function bucketAuto(): Stage\BucketAuto
     {
         $stage = new Stage\BucketAuto($this, $this->dm, $this->class);
-        $this->addStage($stage);
 
-        return $stage;
+        return $this->addStage($stage);
     }
 
     /**
@@ -141,9 +138,8 @@ class Builder
     public function collStats(): Stage\CollStats
     {
         $stage = new Stage\CollStats($this);
-        $this->addStage($stage);
 
-        return $stage;
+        return $this->addStage($stage);
     }
 
     /**
@@ -155,9 +151,8 @@ class Builder
     public function count(string $fieldName): Stage\Count
     {
         $stage = new Stage\Count($this, $fieldName);
-        $this->addStage($stage);
 
-        return $stage;
+        return $this->addStage($stage);
     }
 
     /**
@@ -168,9 +163,8 @@ class Builder
     public function densify(string $fieldName): Stage\Densify
     {
         $stage = new Stage\Densify($this, $fieldName);
-        $this->addStage($stage);
 
-        return $stage;
+        return $this->addStage($stage);
     }
 
     /**
@@ -208,9 +202,8 @@ class Builder
     public function facet(): Stage\Facet
     {
         $stage = new Stage\Facet($this);
-        $this->addStage($stage);
 
-        return $stage;
+        return $this->addStage($stage);
     }
 
     /**
@@ -221,9 +214,8 @@ class Builder
     public function fill(): Stage\Fill
     {
         $stage = new Stage\Fill($this);
-        $this->addStage($stage);
 
-        return $stage;
+        return $this->addStage($stage);
     }
 
     /**
@@ -244,9 +236,8 @@ class Builder
     public function geoNear($x, $y = null): Stage\GeoNear
     {
         $stage = new Stage\GeoNear($this, $x, $y);
-        $this->addStage($stage);
 
-        return $stage;
+        return $this->addStage($stage);
     }
 
     /**
@@ -347,9 +338,8 @@ class Builder
     public function graphLookup(string $from): Stage\GraphLookup
     {
         $stage = new Stage\GraphLookup($this, $from, $this->dm, $this->class);
-        $this->addStage($stage);
 
-        return $stage;
+        return $this->addStage($stage);
     }
 
     /**
@@ -361,9 +351,8 @@ class Builder
     public function group(): Stage\Group
     {
         $stage = new Stage\Group($this);
-        $this->addStage($stage);
 
-        return $stage;
+        return $this->addStage($stage);
     }
 
     /**
@@ -384,9 +373,8 @@ class Builder
     public function indexStats(): Stage\IndexStats
     {
         $stage = new Stage\IndexStats($this);
-        $this->addStage($stage);
 
-        return $stage;
+        return $this->addStage($stage);
     }
 
     /**
@@ -397,9 +385,8 @@ class Builder
     public function limit(int $limit): Stage\Limit
     {
         $stage = new Stage\Limit($this, $limit);
-        $this->addStage($stage);
 
-        return $stage;
+        return $this->addStage($stage);
     }
 
     /**
@@ -412,9 +399,8 @@ class Builder
     public function lookup(string $from): Stage\Lookup
     {
         $stage = new Stage\Lookup($this, $from, $this->dm, $this->class);
-        $this->addStage($stage);
 
-        return $stage;
+        return $this->addStage($stage);
     }
 
     /**
@@ -426,9 +412,8 @@ class Builder
     public function match(): Stage\MatchStage
     {
         $stage = new Stage\MatchStage($this);
-        $this->addStage($stage);
 
-        return $stage;
+        return $this->addStage($stage);
     }
 
     /**
@@ -451,9 +436,8 @@ class Builder
     public function merge(): Stage\Merge
     {
         $stage = new Stage\Merge($this, $this->dm);
-        $this->addStage($stage);
 
-        return $stage;
+        return $this->addStage($stage);
     }
 
     /**
@@ -465,9 +449,8 @@ class Builder
     public function out(string $from): Stage\Out
     {
         $stage = new Stage\Out($this, $from, $this->dm);
-        $this->addStage($stage);
 
-        return $stage;
+        return $this->addStage($stage);
     }
 
     /**
@@ -480,9 +463,8 @@ class Builder
     public function project(): Stage\Project
     {
         $stage = new Stage\Project($this);
-        $this->addStage($stage);
 
-        return $stage;
+        return $this->addStage($stage);
     }
 
     /**
@@ -494,9 +476,8 @@ class Builder
     public function redact(): Stage\Redact
     {
         $stage = new Stage\Redact($this);
-        $this->addStage($stage);
 
-        return $stage;
+        return $this->addStage($stage);
     }
 
     /**
@@ -513,9 +494,8 @@ class Builder
     public function replaceRoot($expression = null): Stage\ReplaceRoot
     {
         $stage = new Stage\ReplaceRoot($this, $this->dm, $this->class, $expression);
-        $this->addStage($stage);
 
-        return $stage;
+        return $this->addStage($stage);
     }
 
     /**
@@ -534,9 +514,8 @@ class Builder
     public function replaceWith($expression = null): Stage\ReplaceWith
     {
         $stage = new Stage\ReplaceWith($this, $this->dm, $this->class, $expression);
-        $this->addStage($stage);
 
-        return $stage;
+        return $this->addStage($stage);
     }
 
     /**
@@ -557,9 +536,8 @@ class Builder
     public function sample(int $size): Stage\Sample
     {
         $stage = new Stage\Sample($this, $size);
-        $this->addStage($stage);
 
-        return $stage;
+        return $this->addStage($stage);
     }
 
     /**
@@ -573,9 +551,8 @@ class Builder
     public function set(): Stage\Set
     {
         $stage = new Stage\Set($this);
-        $this->addStage($stage);
 
-        return $stage;
+        return $this->addStage($stage);
     }
 
     /**
@@ -587,9 +564,8 @@ class Builder
     public function skip(int $skip): Stage\Skip
     {
         $stage = new Stage\Skip($this, $skip);
-        $this->addStage($stage);
 
-        return $stage;
+        return $this->addStage($stage);
     }
 
     /**
@@ -610,9 +586,8 @@ class Builder
         $fields = is_array($fieldName) ? $fieldName : [$fieldName => $order];
         // fixme: move to sort stage
         $stage = new Stage\Sort($this, $this->getDocumentPersister()->prepareSort($fields));
-        $this->addStage($stage);
 
-        return $stage;
+        return $this->addStage($stage);
     }
 
     /**
@@ -624,9 +599,8 @@ class Builder
     public function sortByCount(string $expression): Stage\SortByCount
     {
         $stage = new Stage\SortByCount($this, $expression, $this->dm, $this->class);
-        $this->addStage($stage);
 
-        return $stage;
+        return $this->addStage($stage);
     }
 
     /**
@@ -639,9 +613,8 @@ class Builder
     public function unionWith(string $collection): Stage\UnionWith
     {
         $stage = new Stage\UnionWith($this, $this->dm, $collection);
-        $this->addStage($stage);
 
-        return $stage;
+        return $this->addStage($stage);
     }
 
     /**
@@ -652,9 +625,8 @@ class Builder
     public function unset(string ...$fields): Stage\UnsetStage
     {
         $stage = new Stage\UnsetStage($this, $this->getDocumentPersister(), ...$fields);
-        $this->addStage($stage);
 
-        return $stage;
+        return $this->addStage($stage);
     }
 
     /**
@@ -668,15 +640,18 @@ class Builder
     {
         // Fixme: move field name translation to stage
         $stage = new Stage\Unwind($this, $this->getDocumentPersister()->prepareFieldName($fieldName));
-        $this->addStage($stage);
 
-        return $stage;
+        return $this->addStage($stage);
     }
 
     /**
      * Allows adding an arbitrary stage to the pipeline
      *
-     * @return Stage The method returns the stage given as an argument
+     * @param T $stage
+     *
+     * @return T The method returns the stage given as an argument
+     *
+     * @template T of Stage
      */
     public function addStage(Stage $stage): Stage
     {
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
@@ -214,6 +214,19 @@ class Builder
     }
 
     /**
+     * Populates null and missing field values within documents.
+     *
+     * @see https://www.mongodb.com/docs/rapid/reference/operator/aggregation/fill/
+     */
+    public function fill(): Stage\Fill
+    {
+        $stage = new Stage\Fill($this);
+        $this->addStage($stage);
+
+        return $stage;
+    }
+
+    /**
      * Outputs documents in order of nearest to farthest from a specified point.
      *
      * A GeoJSON point may be provided as the first and only argument for
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
@@ -563,6 +563,22 @@ class Builder
     }
 
     /**
+     * Adds new fields to documents. $set outputs documents that contain all
+     * existing fields from the input documents and newly added fields.
+     *
+     * The $set stage is an alias for $addFields.
+     *
+     * @see https://www.mongodb.com/docs/rapid/reference/operator/aggregation/set/
+     */
+    public function set(): Stage\Set
+    {
+        $stage = new Stage\Set($this);
+        $this->addStage($stage);
+
+        return $stage;
+    }
+
+    /**
      * Skips over the specified number of documents that pass into the stage and
      * passes the remaining documents to the next stage in the pipeline.
      *
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
@@ -630,6 +630,19 @@ class Builder
     }
 
     /**
+     * Removes/excludes fields from documents.
+     *
+     * @see https://www.mongodb.com/docs/rapid/reference/operator/aggregation/unset/
+     */
+    public function unset(string ...$fields): Stage\UnsetStage
+    {
+        $stage = new Stage\UnsetStage($this, $this->getDocumentPersister(), ...$fields);
+        $this->addStage($stage);
+
+        return $stage;
+    }
+
+    /**
      * Deconstructs an array field from the input documents to output a document
      * for each element. Each output document is the input document with the
      * value of the array field replaced by the element.
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
@@ -630,6 +630,21 @@ class Builder
     }
 
     /**
+     * Performs a union of two collections. $unionWith combines pipeline results
+     * from two collections into a single result set. The stage outputs the
+     * combined result set (including duplicates) to the next stage.
+     *
+     * @see https://www.mongodb.com/docs/rapid/reference/operator/aggregation/unionWith/
+     */
+    public function unionWith(string $collection): Stage\UnionWith
+    {
+        $stage = new Stage\UnionWith($this, $this->dm, $collection);
+        $this->addStage($stage);
+
+        return $stage;
+    }
+
+    /**
      * Removes/excludes fields from documents.
      *
      * @see https://www.mongodb.com/docs/rapid/reference/operator/aggregation/unset/
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
@@ -519,6 +519,27 @@ class Builder
     }
 
     /**
+     * Replaces the input document with the specified document. The operation
+     * replaces all existing fields in the input document, including the _id
+     * field. With $replaceWith, you can promote an embedded document to the
+     * top-level. You can also specify a new document as the replacement.
+     *
+     * The $replaceWith stage is an alias for $replaceRoot.
+     *
+     * @see https://www.mongodb.com/docs/rapid/reference/operator/aggregation/replaceWith/
+     *
+     * @param string|mixed[]|Expr|null $expression Optional. A replacement expression that
+     * resolves to a document.
+     */
+    public function replaceWith($expression = null): Stage\ReplaceWith
+    {
+        $stage = new Stage\ReplaceWith($this, $this->dm, $this->class, $expression);
+        $this->addStage($stage);
+
+        return $stage;
+    }
+
+    /**
      * Controls if resulting iterator should be wrapped with CachingIterator.
      */
     public function rewindable(bool $rewindable = true): self
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
@@ -161,6 +161,19 @@ class Builder
     }
 
     /**
+     * Creates new documents in a sequence of documents where certain values in a field are missing.
+     *
+     * @see https://www.mongodb.com/docs/rapid/reference/operator/aggregation/densify/
+     */
+    public function densify(string $fieldName): Stage\Densify
+    {
+        $stage = new Stage\Densify($this, $fieldName);
+        $this->addStage($stage);
+
+        return $stage;
+    }
+
+    /**
      * Executes the aggregation pipeline
      *
      * @deprecated This method was deprecated in doctrine/mongodb-odm 2.2. Please use getAggregation() instead.
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/AddFields.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/AddFields.php
@@ -7,12 +7,21 @@ namespace Doctrine\ODM\MongoDB\Aggregation\Stage;
 /**
  * Fluent interface for adding a $addFields stage to an aggregation pipeline.
  */
-final class AddFields extends Operator
+class AddFields extends Operator
 {
+    private bool $isSet = false;
+
+    protected function isSet(): void
+    {
+        $this->isSet = true;
+    }
+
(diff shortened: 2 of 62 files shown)


lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
```diff
@@ -563,15 +640,18 @@ public function unwind(string $fieldName): Stage\Unwind
...
 
     /**
      * Allows adding an arbitrary stage to the pipeline
      *
-     * @return Stage The method returns the stage given as an argument
+     * @param T $stage
+     *
+     * @return T The method returns the stage given as an argument
+     *
+     * @template T of Stage
```
jmikola, line 657: What does `@template T of Stage` accomplish here? Is this necessary to hint to psalm that the very same instance is returned? Is it relevant that the [psalm docs](https://psalm.dev/docs/annotating_code/templated_annotations/#template-psalm-template) use `psalm-param` instead of `param`?

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Densify.php
```diff
@@ -0,0 +1,68 @@
...
+
+    public function getExpression(): array
+    {
+        $params = (object) ['field' => $this->field];
+
+        if ($this->partitionByFields) {
+            $params->partitionByFields = $this->partitionByFields;
+        }
+
+        if ($this->range) {
```
jmikola: If you change the type hint to `object`, would this need to be `isset()` to avoid accessing an uninitialized property?

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Densify.php
```diff
@@ -0,0 +1,68 @@
...
+ * Fluent interface for adding a $densify stage to an aggregation pipeline.
+ */
+class Densify extends Stage
+{
+    private string $field;
+
+    /** @var array<string> */
+    private array $partitionByFields = [];
+
+    private ?object $range = null;
```
jmikola: The [`range` option](https://www.mongodb.com/docs/rapid/reference/operator/aggregation/densify/#std-label-densify-range) is required. Should this really use `?object` instead of `object` and allowing it to be initially empty?

I understand building the range document is a bit complicated, so you don't want to complicate the `Builder::densify()` method.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Set.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ODM\MongoDB\Aggregation\Stage;
+
+use Doctrine\ODM\MongoDB\Aggregation\Builder;
+
+class Set extends AddFields
```
jmikola: Wouldn't it be more convenient to copy AddFields and make both classes `final`? There's essentially only one line of code in the class, so that would be simpler than implementing `isSet()` and some conditional logic in `getExpression()`.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/AddFields.php
```diff
@@ -7,12 +7,21 @@
 /**
  * Fluent interface for adding a $addFields stage to an aggregation pipeline.
  */
-final class AddFields extends Operator
+class AddFields extends Operator
```
jmikola: See my related suggestion in `Set.php`. I don't think this class needs to be changed.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Fill.php
```diff
@@ -0,0 +1,105 @@
...
+    /** @var mixed|Expr|null */
+    private $partitionBy = null;
+
+    /** @var array<string> */
+    private array $partitionByFields = [];
+
+    /** @var array<string, int> */
+    private array $sortBy = [];
+
+    private ?Output $output = null;
```
jmikola: [`output`](https://www.mongodb.com/docs/rapid/reference/operator/aggregation/fill/#std-label-fill-output) is also required by the server. Should this drop the nullable type hint and just start out uninitialized?

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Fill.php
```diff
@@ -0,0 +1,105 @@
...
+    public function partitionBy($expression): self
+    {
+        $this->partitionBy = $expression;
+
+        return $this;
+    }
+
+    public function partitionByFields(string ...$fields): self
+    {
+        $this->partitionByFields = $fields;
```
jmikola: `...` is always guaranteed to yield a packed array, correct? Just wondering if there might be some edge case with named parameters, although I'm pretty sure those are mutually exclusive with variadics.

Per my related comments in later files, should we just use `array_values()` here?

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Fill.php
```diff
@@ -0,0 +1,105 @@
...
+
+        return $this;
+    }
+
+    /**
+     * @param array<string, int|string>|string $fieldName Field name or array of field/order pairs
+     * @param int|string                       $order     Field order (if one field is specified)
```
jmikola, line 70: FYI: [`sortBy`](https://www.mongodb.com/docs/rapid/reference/operator/aggregation/fill/#std-label-fill-sortBy) uses the same syntax as [`$sort`](https://www.mongodb.com/docs/rapid/reference/operator/aggregation/sort/#mongodb-pipeline-pipe.-sort), which shows `{ $meta: "textScore" }` as a possible ordering.

I have no idea if text search is relevant here, but it seemed worth pointing out in case this was an oversight.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Fill/Output.php
```diff
@@ -0,0 +1,109 @@
...
+    {
+        return $this->output;
+    }
+
+    /**
+     * Ensure that a current field has been set.
+     *
+     * @throws LogicException if a current field has not been set.
+     */
+    private function requiresCurrentField(?string $method = null): void
```
jmikola: `requiresCurrentField()` is always called with an argument. Any reason not to just use `string $method` here?

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Fill/Output.php
```diff
@@ -0,0 +1,109 @@
...
+    {
+        if (! $this->currentField) {
+            throw new LogicException(($method ?: 'This method') . ' requires you set a current field using field().');
```
jmikola: "requires that you set a..." or "requires setting a..."

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Fill/Output.php
```diff
@@ -0,0 +1,109 @@
...
+
+    public function __construct(Builder $builder, Fill $fill)
+    {
+        parent::__construct($builder);
+
+        $this->fill = $fill;
+    }
+
+    /** @param mixed|Expr $expression */
+    public function partitionBy($expression): Fill
```
jmikola, line 38: Related to my previous question about using `@template` to possibly indicate some method returning the very same parameter, is there some templating you may want to use here to indicate that the very same instance of Fill is returned by these methods as is received in the constructor?

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Merge.php
```diff
@@ -0,0 +1,148 @@
...
+    private $into;
+
+    /** @var list<string> */
+    private array $on = [];
+
+    /** @var array<string, string> */
+    private array $let = [];
+
+    /**
+     * @var string|array|Builder
```
jmikola: Not sure if this is relevant, but WhenMatchedParam also includes Stage.

I assume you split this with `@psalm-var` to avoid using `list<>` here.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Merge.php
```diff
@@ -0,0 +1,148 @@
...
+ */
+class Merge extends Stage
+{
+    private DocumentManager $dm;
+
+    /**
+     * @var string|array
+     * @psalm-var OutputCollection
+     */
+    private $into;
```
jmikola, line 42: Related to my earlier questions about not using nullable type hints for required options, I noted that [`into`](https://www.mongodb.com/docs/rapid/reference/operator/aggregation/merge/#std-label-merge-into) is required and you didn't use a nullable type hint here. However, that might just be because there's no PHP syntax available to do so.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Merge.php
```diff
@@ -0,0 +1,148 @@
...
+        return $this;
+    }
+
+    /**
+     * Optional. Specifies variables to use in the pipeline stages.
+     *
+     * Use the variable expressions to access the fields from
+     * the joined collection's documents that are input to the pipeline.
+     *
+     * @param array<string, string> $let
```
jmikola: Are all expressions in this [`let`](https://www.mongodb.com/docs/rapid/reference/operator/aggregation/merge/#std-label-merge-let) limited to strings?

If it's anything like [`$let`](https://www.mongodb.com/docs/manual/reference/operator/aggregation/let/) then the expression could be a document. I believe that's also what you accept in `Doctrine\ODM\MongoDB\Aggregation\Expr::let()`, although the expressions aren't interpreted further.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Merge.php
```diff
@@ -0,0 +1,148 @@
...
+    public function let(array $let): self
+    {
+        $this->let = $let;
+
+        return $this;
+    }
+
+    public function on(string ...$fields): self
+    {
+        $this->on = array_values($fields);
```
jmikola, line 132: Related to my earlier question about splat in `Fill::partitionByFields()`, I'm curious why you only used `array_values()` here.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Merge.php
```diff
@@ -0,0 +1,148 @@
...
+    public function whenMatched($whenMatched): self
+    {
+        if ($whenMatched instanceof Stage) {
+            $this->whenMatched = $whenMatched->builder;
+        } else {
+            $this->whenMatched = $whenMatched;
+        }
+
+        if ($this->builder === $this->whenMatched) {
+            throw new InvalidArgumentException('Cannot use the same Builder instance for $merge whenMatched pipeline.');
```
jmikola, line 150: Ignore this is the trailing period is consistent with other exceptions. I'm assuming that single-sentence messages don't need it, though.

```suggestion
            throw new InvalidArgumentException('Cannot use the same Builder instance for $merge whenMatched pipeline');
```

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Merge.php
```diff
@@ -0,0 +1,148 @@
...
+        }
+
+        if ($this->let) {
+            $params->let = $this->let;
+        }
+
+        if ($this->whenMatched) {
+            $params->whenMatched = $this->whenMatched instanceof Builder
+                ? $this->whenMatched->getPipeline(false)
+                : $this->whenMatched;
```
jmikola, line 83: In the event an array is passed, I assume it gets used as-is?

I know psalm would enforce the `list<array<string, mixed>>` type but should there be a runtime check for well-formed pipeline (e.g. [`array_is_list()`](https://www.php.net/manual/en/function.array-is-list.php)) or is this something you'll defer to the server to report?

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/UnionWith.php
```diff
@@ -0,0 +1,75 @@
...
+use Doctrine\ODM\MongoDB\Aggregation\Builder;
+use Doctrine\ODM\MongoDB\Aggregation\Stage;
+use Doctrine\ODM\MongoDB\DocumentManager;
+use Doctrine\Persistence\Mapping\MappingException;
+use InvalidArgumentException;
+
+/**
+ * Fluent interface for adding a $unionWith stage to an aggregation pipeline.
+ *
+ * @psalm-type Pipeline = Builder|Stage|list<array<string, mixed>>
```
jmikola: This seems like a generally useful type alias. For instance, WhenMatchedParam could have used `Pipeline|string`.

Are class doc blocks the highest point we can define these, or is there a way to share them across files?

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/UnionWith.php
```diff
@@ -0,0 +1,75 @@
...
+ */
+class UnionWith extends Stage
+{
+    private DocumentManager $dm;
+
+    private string $collection;
+
+    /**
+     * @var array|Builder|null
+     * @psalm-var Pipeline|null
```
jmikola: Should this be `?Pipeline`, or does psalm prefer `|null`?

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/UnionWith.php
```diff
@@ -0,0 +1,75 @@
...
+    public function __construct(Builder $builder, DocumentManager $documentManager, string $collection)
+    {
+        parent::__construct($builder);
+
+        $this->dm = $documentManager;
+
+        try {
+            $class            = $this->dm->getClassMetadata($collection);
+            $this->collection = $class->getCollection();
+        } catch (MappingException $e) {
```
jmikola, line 46: I assume `$collection` may either be a class name or collection name, and you want to give preference to a class name. Would calling `class_exists()` be more efficient than relying on an exception, or is that less reliable (or inviting of some other edge case)? I'm not sure if there's an edge case where we can have metadata for a class that isn't actually defined in PHP (maybe through one of the metadata drivers)?

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/UnionWith.php
```diff
@@ -0,0 +1,75 @@
...
+    public function pipeline($pipeline): self
+    {
+        if ($pipeline instanceof Stage) {
+            $this->pipeline = $pipeline->builder;
+        } else {
+            $this->pipeline = $pipeline;
+        }
+
+        if ($this->builder === $this->pipeline) {
+            throw new InvalidArgumentException('Cannot use the same Builder instance for $unionWith pipeline.');
```
jmikola, line 64:
```suggestion
            throw new InvalidArgumentException('Cannot use the same Builder instance for $unionWith pipeline');
```

As with my previous suggestion, ignore if this would be inconsistent.

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/DensifyTest.php
```diff
@@ -0,0 +1,44 @@
...
+class DensifyTest extends BaseTest
+{
+    use AggregationTestTrait;
+
+    public function testDensifyStage(): void
+    {
+        $densifyStage = new Densify($this->getTestAggregationBuilder(), 'someField');
+        $densifyStage
+            ->partitionByFields('field1', 'field2')
+            ->range('full', 1);
```
jmikola, line 20: Should you test the third `$unit` arg as well?

Also, `array{0: int, 1: int}` is a permitted type for `$bounds` and `$step` can also be a `float` if you care to test those.

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/DensifyTest.php
```diff
@@ -0,0 +1,44 @@
...
+                        'bounds' => 'full',
+                        'step' => 1,
+                    ],
+                ],
+            ],
+            $densifyStage->getExpression(),
+        );
+    }
+
+    public function testCountFromBuilder(): void
```
jmikola:
```suggestion
    public function testDensifyFromBuilder(): void
```

Copypasta?

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/FillTest.php
```diff
@@ -0,0 +1,70 @@
...
+                        'bar' => ['method' => 'linear'],
+                        'fixed' => ['value' => 0],
+                    ],
+                ],
+            ],
+            $fillStage->getExpression(),
+        );
+    }
+
+    public function testCountFromBuilder(): void
```
jmikola:
```suggestion
    public function testFillFromBuilder(): void
```

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/FillTest.php
```diff
@@ -0,0 +1,70 @@
...
+    public function testFillStage(): void
+    {
+        $fillStage = new Fill($this->getTestAggregationBuilder());
+        $fillStage
+            ->partitionByFields('field1', 'field2')
+            ->sortBy('field1', 1)
+            ->output()
+                ->field('foo')->locf()
+                ->field('bar')->linear()
+                ->field('fixed')->value(0);
```
jmikola: Is it worth testing a more complex expression here?

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/FillTest.php
```diff
@@ -0,0 +1,70 @@
...
+class FillTest extends BaseTest
+{
+    use AggregationTestTrait;
+
+    public function testFillStage(): void
+    {
+        $fillStage = new Fill($this->getTestAggregationBuilder());
+        $fillStage
+            ->partitionByFields('field1', 'field2')
+            ->sortBy('field1', 1)
```
jmikola, line 21: You may also want to test string values (e.g. "asc") and `{ $meta: "textScore" }` if you decide to handle that.

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/MergeTest.php
```diff
@@ -0,0 +1,113 @@
...
+            },
+        ];
+    }
+
+    /**
+     * @param array<array<string, mixed>>|callable $pipeline
+     *
+     * @dataProvider providePipeline
+     */
+    public function testMergeStageWithPipeline($pipeline): void
```
jmikola: Is it worth testing the exception for re-using the same Builder instance as the parent stage?

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/ReplaceWithTest.php
```diff
@@ -0,0 +1,90 @@
...
+use MongoDB\BSON\UTCDateTime;
+
+class ReplaceWithTest extends BaseTest
+{
+    public function testTypeConversion(): void
+    {
+        $builder = $this->dm->createAggregationBuilder(User::class);
+
+        $dateTime  = new DateTimeImmutable('2000-01-01T00:00Z');
+        $mongoDate = new UTCDateTime((int) $dateTime->format('Uv'));
```
jmikola: Why not construct UTCDateTime from `$dateTime` directly (see: [`php_phongo_utcdatetime_init_from_date`](https://github.com/mongodb/mongo-php-driver/blob/1.15.1/src/BSON/UTCDateTime.c#L92))?

I'm not sure if there's a difference using `format('Uv')` vs. PHPC reading `DateTimeImmutable` if the DateTime was constructed from a string with sub-second precision, but it looks like [``](https://github.com/php/php-src/blob/php-8.2.3/ext/date/lib/parse_date.c#L5138) does set `us` for anything after a dot.

However, the string above has no sub-second precision so behavior should be predictable.

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/SetTest.php
```diff
@@ -0,0 +1,35 @@
...
+    {
+        $setStage = new Set($this->getTestAggregationBuilder());
+        $setStage
+            ->field('product')
+            ->multiply('$field', 5);
+
+        self::assertSame(['$set' => ['product' => ['$multiply' => ['$field', 5]]]], $setStage->getExpression());
+    }
+
+    public function testProjectFromBuilder(): void
```
jmikola:
```suggestion
    public function testSetFromBuilder(): void
```

On that note, I think there's a similar typo in AddFieldsTest, from which this was copied.

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/UnsetTest.php
```diff
@@ -0,0 +1,31 @@
...
+
+    public function testUnsetStage(): void
+    {
+        $documentPersister = $this->dm->getUnitOfWork()->getDocumentPersister(User::class);
+        $unsetStage        = new UnsetStage($this->getTestAggregationBuilder(), $documentPersister, 'id', 'foo', 'bar');
+
+        self::assertSame(['$unset' => ['_id', 'foo', 'bar']], $unsetStage->getExpression());
+    }
+
+    public function testLimitFromBuilder(): void
```
jmikola:
```suggestion
    public function testUnsetFromBuilder(): void
```

Maybe you want to just rename all of these tests to "testFromBuilder()"?

lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
alcaeus: Yes, not having this templated was causing issues as psalm couldn't know that the method returns a specific stage instance, so it would complain if we were to return the result directly. This way, psalm knows that whatever type `T` is (besides being a subtype of `Stage`), we're returning the same type in the method.

`@psalm-param` is a vendor-specific param annotation to specify a more specific param annotation for psalm but retain the original for phpdoc tools. I'm not sure if it makes any difference considering we don't generate documentation, but generally speaking a type declared with `@psalm-type` wouldn't be known to tooling that only reads `@param`, so those types should only be used in `@psalm-param`. Since we're using a generic `@template` annotation which is understood by tools like psalm, phpstan, and also PhpStorm, we can use the `T` type in `@param` as well. I hope that explains the thinking.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Densify.php
alcaeus: That's correct. It's an alternative to the current design that I honestly haven't considered. I'm open to using an uninitialised `object` property and only accessing it if it was set beforehand if that makes more sense than using a nullable property with a `null` default value.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Densify.php
alcaeus: I've considered this, but ultimately decided against it. A fatal error due to accessing a non-existent property might be more difficult to handle for users than handling a command error from the driver due to sending an incomplete `$densify` stage to the server. Alternatively, we could throw when we generate the expression in this case.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Set.php
alcaeus: That's a good point, I'll do that.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Fill.php
alcaeus: With named parameters, this could actually be a hash. Consider this:
```php
$fill->partitionByFields(foo: 'bar', bar: 'baz');
```

In this case, `$fields` would be `['foo' => 'bar', 'bar' => 'bar']`. I'll go through all stages and make sure to use `array_values` to get a packed array where field names are irrelevant (or even undesired).

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Fill.php
alcaeus: I omitted this on purpose. The [documentation for `$sort`](https://www.mongodb.com/docs/manual/reference/operator/aggregation/sort/) explicitly mentions the `$meta` sorting, whereas the [documentation for `$sortArray`](https://www.mongodb.com/docs/manual/reference/operator/aggregation/sortArray/) does not.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Fill/Output.php
alcaeus: This was copy-pasted from the `Expr` class and thus not adapted. I'll remove the nullability.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Fill/Output.php
alcaeus: Will update here and in other places where we use such a method.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Fill/Output.php
alcaeus: Ah, good point. I believe we can only tell psalm that we're returning a specific type, but not that we're returning the exact instance that `output()` was called on.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Merge.php
alcaeus: Not including `Stage` here was an oversight.

The `@psalm-var` is the case that I mentioned above with psalm-specific types not being used in generic annotations, as any tool that doesn't read psalm annotations won't know what `WhenMatchedParam` is.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Merge.php
alcaeus: That is correct. Typically I would've used `string|array|null` here, but since we can't use union types yet (ODM still supports PHP 7.4) I had to leave out the property type entirely.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Merge.php
alcaeus: Good catch! This should indeed be `array<string, mixed|Expr>` here. I'll update this.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Merge.php
alcaeus: Some use it, some don't. I usually include trailing periods, but I'm fine with either once we've decided on a format for ODM.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Merge.php
alcaeus: I decided to defer such checks to the server for brevity, as otherwise we wouldn't just have to check if it's a list, but also that it's a list of pipeline stages (which could be objects or hashes with a single key that starts with a dollar sign).

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/UnionWith.php
alcaeus: These can be shared using `@psalm-import-type`. I'll go through and use some more specific type annotations throughout the new classes 👍

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/UnionWith.php
alcaeus: Good question, I believe `?Pipeline` may work just as well. I'll double-check though.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/UnionWith.php
alcaeus: Theoretically, a metadata driver could return metadata for a non-existent class. The exception handling is for the case that somebody passes a collection name instead of a class name, in which case we wouldn't find any metadata causing `getClassMetadata` to throw.

We also can't use `ClassMetadataFactory::hasMetadataFor()` here, as it only returns `true` if the metadata for a class was loaded already. If the metadata wasn't loaded (e.g. because ODM hasn't interacted with a class for that type before), it will return `false` despite there potentially being metadata available.

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/DensifyTest.php
alcaeus: Will add more tests for this 👍

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/DensifyTest.php
alcaeus: Busted!

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/FillTest.php
alcaeus: I'll add a test that passes a nested expression 👍

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/FillTest.php
alcaeus: Ack to handling strings 👍

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/ReplaceWithTest.php
alcaeus: Mainly because I shamelessly copied this from `ReplaceRootTest`. I'll update all of these instances 👍

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/UnsetTest.php
alcaeus: Good idea, that way people won't see just how much copy/paste my job consists of 😇

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Set.php
alcaeus: Classes split and declared as final again.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Fill.php
alcaeus: Added `array_values` to instances where we expect a list of field names from a variadic parameter.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Fill/Output.php
alcaeus: Changed, and also changed the `Query\Expr` class to also print the method name instead of throwing a generic "This method ..." exception

alcaeus:
While adding types for things like `Pipeline`, I added type declarations to most of the pipeline stages that can be imported. Due to a bug in phpstan we have to silence a number of phpstan errors that arise from it tripping over object shapes in `@psalm-type` declarations. See https://github.com/phpstan/phpstan/issues/9077 for a bug report.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Densify.php
jmikola: This relates to my question about using `isset()` later. With `isset()`, I don't think there'd be any risk of fatal error.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Densify.php
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * Fluent interface for adding a $densify stage to an aggregation pipeline.
+ *
+ * @psalm-type BoundsType = 'full'|'partition'|array{0: int|float|UTCDateTime, 1: int|float|UTCDateTime}
+ * @psalm-type UnitType = 'year'|'month'|'week'|'day'|'hour'|'minute'|'second'|'millisecond'
+ * @psalm-type DensifyStageExpression = array{
+ *     '$densify': object{
+ *         field: string,
+ *         partitionByFields?: list<string>,
+ *         range?: object{
```
jmikola: Does this need to be optional because you conditionally set `range` in `getExpression()`? Even if you removed the nullable modifier on the class property, conditionally setting it in the return value (using `isset()` to avoid uninitialized property access) would still necessitate this, correct?

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Densify.php
jmikola: I understand DensifyStageExpression may need to remain as is (with the optional modifier on `range`), but I think avoiding the nullable modifier on the class property and using `isset()` here makes sense.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Fill.php
jmikola: I left some feedback for the docs team to consider clarifying this in the `$fill` docs.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Merge.php
jmikola: You noted that you're adding `array_values()` in more places, so I'll resolve this.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Densify.php
alcaeus: That is correct. If users never call the `range` method, we'll return a stage without a `range` object. While the server will reject this, it's still what we return

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Fill.php
alcaeus: `output` is now always generated in the constructor, removing any special handling.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Densify.php
alcaeus: Changed to create an empty object in the constructor. That does away with dealing with uninitialised properties.

jmikola (approved):

PR #2514, opened 2023-03-22
Support new aggregation operators in builder

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues |

#### Summary

This PR adds the following new aggregation pipeline operators to the builder: `$accumulator`, `$binarySize`, `$bottom`, `$bottomN`, `$bsonSize`, `$dateAdd`, `$dateDiff`, `$dateFromParts`, `$dateFromString`, `$dateSubtract`, `$dateToParts`, `$dateTrunc`, `$firstN`, `$function`, `$getField`, `$lastsN`, `$maxN`, `$mergeObjects`, `$minN`, `$rand`, `$regexFind`, `$regexFindAll`, `$regexMatch`, `$replaceAll`, `$replaceOne`, `$sampleRate`, `$setField`, `$sortArray`, `$top`, `$topN`, `$tsIncrement`, `$tsSecond`.

This covers all aggregation pipeline operators currently supported in MongoDB, with the exception of accumulator operators only supported in the `$setWindowFields` stage, which will be introduced in a separate pull request.

I've also taken liberty to organise the code a bit. We now have interfaces for the various operator groups instead of one generic operator interface, and I've cleaned up the logic around accumulator stages (such as `$sum`) which have two different forms depending on the stage they're being used in. In the `$group`, `$bucket`, and `$bucketAuto` stages they take a single argument, whereas they take multiple arguments in other stages such as `$project`. This is now properly accounted for and easier to handle due to a trait that handles this more gracefully.

Note: GitHub Copilot was instrumental in getting this work done as it automated away certain repetitive tasks (such as adding proxy methods to the `Operator` class. I've also marked that class as `@internal` as we may want to do away with it in the future in favour of multiple traits that add specific operator groups, similar to how we have interfaces for the groups.


lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
```diff
@@ -192,10 +278,62 @@ public function arrayElemAt($array, $index): self
...
+        return $this->operator('$binarySize', $expression);
+    }
+
+    /**
+     * Returns the bottom element within a group according to the specified sort order.
+     *
+     * @see https://docs.mongodb.com/manual/reference/operator/aggregation/bottom/
+     *
+     * @param mixed|self                $output
+     * @param array<string, int|string> $sortBy
```
jmikola: Would it make sense to use a psalm type alias here for a sort spec?

lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
```diff
@@ -325,6 +463,175 @@ public static function convertExpression($expression)
         return $expression;
     }
 
+    /**
+     * Returns the number of documents in a group.
+     *
+     * @see https://docs.mongodb.com/manual/reference/operator/aggregation/count/
+     */
+    public function countDocuments(): self
```
jmikola: Is this intentionally not named `count()` to avoid a potential conflict with Countable?

lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
```diff
@@ -325,6 +463,175 @@ public static function convertExpression($expression)
...
+     *
+     * @param mixed|self $dateString
+     * @param mixed|self $format
+     * @param mixed|self $timezone
+     * @param mixed|self $onError
+     * @param mixed|self $onNull
+     */
+    public function dateFromString($dateString, $format = null, $timezone = null, $onError = null, $onNull = null): self
+    {
+        return $this->operator(
```
jmikola, line 335: Any reason this uses different formatting than `dateFromParts()` above?

lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
```diff
@@ -325,6 +463,175 @@ public static function convertExpression($expression)
...
+                    'timezone',
+                    'onError',
+                    'onNull',
+                ],
+            ),
+        );
+    }
+
+    /**
+     * Decrements a Date object by a specified number of time units
```
jmikola: Should "Date" be capitalized in this context? The casing is inconsistent across doc blocks, but if we're only talking about an object in the pipeline (i.e. BSON document with date fields within) I think lowercase makes sense to avoid anyone confusing this with a PHP object.

lib/Doctrine/ODM/MongoDB/Aggregation/Operator/ArrayOperators.php
```diff
@@ -0,0 +1,267 @@
...
+     * Returns the first array element. Distinct from the $first accumulator.
+     *
+     * @see https://www.mongodb.com/docs/manual/reference/operator/aggregation/first-array-element/
+     *
+     * @param mixed|Expr $expression
+     */
+    public function first($expression): self;
+
+    /**
+     * Returns a specified number of elements from the beginning of an array. Distinct from the $firstN accumulator.
```
jmikola: Should this be wrapped like other doc blocks?

lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
```diff
@@ -517,6 +853,39 @@ public function first($expression): self
         return $this->operator('$first', $expression);
     }
 
+    /**
+     * Returns a specified number of elements from the beginning of an array. Distinct from the $firstN accumulator.
```
jmikola: Should this be wrapped like other doc blocks?

Also, it might be helpful to refer to "$group accumulator" for more context. I was initially confused because there are references to other accumulators in this class.

lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
```diff
@@ -517,6 +853,39 @@ public function first($expression): self
...
+
+    /**
+     * Defines a custom aggregation function or expression in JavaScript.
+     *
+     * You can use the $function operator to define custom functions to
+     * implement behavior not supported by the MongoDB Query Language.
+     *
+     * @see https://docs.mongodb.com/manual/reference/operator/aggregation/function/
+     *
+     * @param string|Javascript $body
```
jmikola: Is this actually enforced anywhere, or do you just rely on the server to do so?

lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
```diff
@@ -538,6 +907,22 @@ public function getExpression(): array
         return $this->expr;
     }
 
+    /**
+     * Returns the value of a specified field from a document. If you don't specify an object, $getField returns the value of the field from $$CURRENT.
```
jmikola: Should this be wrapped? I'll stop mentioning this, as it's either intentional or you can address these on your own.

lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
```diff
@@ -903,10 +1304,38 @@ public function map($input, $as, $in): self
      * @see https://docs.mongodb.com/manual/reference/operator/aggregation/max/
      *
      * @param mixed|self $expression
+     * @param mixed|self $expressions
+     */
+    public function max($expression, ...$expressions): self
+    {
+        return $this->accumulatorOperator('$max', $expression, ...$expressions);
```
jmikola: Alternatively, could you just splat `func_get_args()`? I assume this is slightly less overhead.

And the first argument is required because `$max` expects at least one expression, correct?

lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
```diff
@@ -1110,6 +1564,16 @@ public function reverseArray($expression): self
...
     }
 
+    /**
+     * Matches a random selection of input documents. The number of documents selected approximates the sample rate expressed as a percentage of the total number of documents.
+     *
+     * @see https://docs.mongodb.com/manual/reference/operator/aggregation/sampleRate/
+     */
+    public function sampleRate(float $rate): self
+    {
+        return $this->operator('$sampleRate', $rate);
```
jmikola, line 778: I see that `operator()` calls `ensureArray()` on the second argument, which is a `float` in that case. Internally, it appears that `ensureArray()` will leave the `float` as-is, but that begs the question: should the method have a better name?

lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
```diff
@@ -1685,6 +2239,19 @@ private function operator(string $operator, $expression): self
...
+     * Adds one or more $or clause to the current expression.
+     *
+     * @see https://docs.mongodb.com/manual/reference/operator/aggregation/or/
+     *
+     * @param array<string, mixed>|self $expression
+     * @param array<string, mixed>|self ...$expressions
+     */
+    public function or($expression, ...$expressions): self
+    {
+        return $this->operator('$or', func_get_args());
```
jmikola, line 1078: I noticed you used `func_get_args()` here but `$expression, ...$expressions` elsewhere. Should these be consistent?

lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
```diff
@@ -1781,6 +2348,98 @@ public function objectToArray($object): self
         return $this->operator('$objectToArray', $object);
     }
 
+    /**
+     * Provides regular expression (regex) pattern matching capability in
```
jmikola: These methods are missing `@see` links.

lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
```diff
@@ -2059,4 +2710,23 @@ public function isNumber($expression): self
...
+     */
+    private function filterOptionalNullArguments(array $args, array $optionalArgNames): array
+    {
+        return array_filter(
+            $args,
+            /**
+             * @param mixed $value
+             * @param array-key $key
+             */
+            static fn ($value, $key): bool => ! in_array($key, $optionalArgNames) || $value !== null,
```
jmikola: The `$value !== null` check is more efficient, so I'd lead with that to allow `in_array()` to be short-circuited.

lib/Doctrine/ODM/MongoDB/Aggregation/Operator/AccumulatorOperators.php
```diff
@@ -0,0 +1,89 @@
...
+ * Interface containing all accumulator aggregation pipeline operators.
+ *
+ * This interface can be used for type hinting, but must not be implemented by
+ * users. Methods WILL be added to the public API in future minor versions.
+ *
+ * @internal
+ */
+interface AccumulatorOperators
+{
+    /**
```
jmikola, line 19: I assume most of these doc blocks are duplicated in the interfaces and whatever implementations exist. Do you think they're prone to getting out of sync?

I assume you might need the `@param` stuff in both places for tooling.

jmikola (commented):
Not done with the review, but I'm publishing what comments I have so far.

I'm intentionally not reviewing the interfaces in detail and am instead focusing on the implementations.

lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
alcaeus: Yes, I'll re-use the sort types declared in #2513 once that's been merged.

lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
alcaeus: The `count` accumulator is available in the `$group` stage for example, but every stage needs to declare methods for all other stages to allow for the fluent interface. This leads to a conflict between the `count` method to add a `$count` stage and the `$count` accumulator. In the stages, I've named the method `countDocuments`, and I decided to re-use that name for consistency here.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Operator.php
```diff
@@ -281,6 +326,51 @@ public function atanh($expression): self
         return $this;
     }
 
+    /**
+     * Returns the average value of numeric values. Ignores non-numeric values.
+     *
+     * @see https://docs.mongodb.com/manual/reference/operator/aggregation/avg/
```
jmikola: Should this have `@see Expr::avg`? If so, you may want to check other doc blocks in this class for similar omissions.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Operator.php
```diff
@@ -281,6 +326,51 @@ public function atanh($expression): self
...
+     *
+     * @param mixed|Expr $expression
+     * @param mixed|Expr ...$expressions
+     */
+    public function avg($expression, ...$expressions): self
+    {
+        $this->expr->avg(...func_get_args());
```
jmikola, line 192: Is there a reason oyu use `...func_get_args()` here but `and()` above used `$expression, ...$expressions`? Also, is `func_get_args()`  problematic for psalm's type inference?

If `func_get_args()` is generally useful, I think it may be more beneficial in the date methods below, which have many more args.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Operator.php
```diff
@@ -1403,6 +1779,88 @@ public function reduce($input, $initialValue, $in): self
         return $this;
     }
 
+    /**
+     * Provides regular expression (regex) pattern matching capability in
+     * aggregation expressions.
+     *
+     * If a match is found, returns a document that contains information on the
```
jmikola: The regex methods are missing `@see` annotations.

phpunit.xml.dist
```diff
@@ -20,6 +20,10 @@
         <include>
             <directory suffix=".php">./lib/Doctrine/ODM/MongoDB</directory>
         </include>
+        <exclude>
+            <file>./lib/Doctrine/ODM/MongoDB/Aggregation/Stage/GraphLookup/Match.php</file>
+            <file>./lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Match.php</file>
```
jmikola, line 25: Why do these need to be excluded?

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/ExprTest.php
```diff
@@ -18,7 +18,23 @@ class ExprTest extends BaseTest
      * @param array<string, string>          $expected
      * @param Closure(Expr): mixed[]|mixed[] $args
      *
-     * @dataProvider provideAllOperators
+     * @dataProvider provideAccumulatorExpressionOperators
```
jmikola, line 21: I like the idea of traits as data providers. This could be useful in PHPLIB to avoid stuffing all of our common providers in base classes.

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/BucketAutoTest.php
```diff
@@ -4,34 +4,46 @@
...
 
 class BucketAutoTest extends BaseTest
 {
+    use AggregationOperatorsProviderTrait;
     use AggregationTestTrait;
 
-    public function testBucketAutoStage(): void
+    /**
+     * @param array<string, string>          $expected
+     * @param Closure(Expr): mixed[]|mixed[] $args
```
jmikola: Is the space after the colon here problematic for parsing the return type, or does the tooling allow for that?

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/GroupTest.php
```diff
@@ -17,57 +17,18 @@ class GroupTest extends BaseTest
...
-            'first()' => ['first', ['$field']],
-            'last()' => ['last', ['$field']],
-            'max()' => ['max', ['$field']],
-            'min()' => ['min', ['$field']],
-            'push()' => ['push', ['$field']],
-            'stdDevPop()' => ['stdDevPop', ['$field']],
-            'stdDevSamp()' => ['stdDevSamp', ['$field']],
-            'sum()' => ['sum', ['$field']],
-        ];
+        self::assertSame($groupStage, $groupStage->field('foo')->$operator(...$args));
```
jmikola, line 30: Why does `BucketTest::testGroupAccumulators()` not perform a similar assertion on `$bucketStage`?

lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
alcaeus: None other than me relying on phpcbf to format these. Since it doesn't apply consistent formatting for multi-line calls I've updated these manually to be consistent.

lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
alcaeus: While it may make sense in the MongoDB docs, there's no reason for this to be capitalised here. I've updated this throughout this file and the interfaces.

lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
alcaeus: We rely on the server to validate this - mainly because the only reason we're not enforcing this (yet) is because we can't use a union type hint.

lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
alcaeus: Renamed the method to `prepareArgument` and documented it accordingly.

lib/Doctrine/ODM/MongoDB/Aggregation/Operator/AccumulatorOperators.php
alcaeus: They are extremely prone to getting out of sync. I'll see if using `@inheritdoc` (or omitting the docblock entirely) works, or if we at least can omit the description and only keep type information.

phpunit.xml.dist
alcaeus: When PHPUnit generates code coverage information, it trips over these due to the classes using a reserved identifier as class name.

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/ExprTest.php
alcaeus: Yes, traits work great for composing functionality in PHPUnit tests to avoid duplicating tons of logic across test case classes. There are some more instances to clean up in ODM, but I figured this PR got unwieldy enough as it is.

lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
alcaeus: Added missing `@see` links to all methods in the interfaces.

lib/Doctrine/ODM/MongoDB/Aggregation/Operator/AccumulatorOperators.php
alcaeus: I played around with this and was able to remove the docblocks, keeping only the ones in the interfaces. Not only does this reduce the effort of keeping them in sync, but it also dramatically reduces the size of the `Expr` and `Operator` classes 🎉

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Operator.php
alcaeus: I've switched all methods to use `...func_get_args()`, regardless of the number of arguments. This makes it less error-prone when adding parameters or new methods (and more opportunity for copypasta)

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Operator.php
alcaeus: Added to the interfaces.

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/BucketAutoTest.php
alcaeus: I've asked in the Symfony Slack's static-analysis channel as I don't have a good answer for this. I'll report back 👍

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/GroupTest.php
alcaeus: Because once you get into the `output` of a `Bucket` stage, it no longer returns the stage itself but an output object for said stage. That makes it pretty much impossible to test.

lib/Doctrine/ODM/MongoDB/Aggregation/Operator/ArrayOperators.php
alcaeus: Wrapped description in interface docblocks at 80 chars.

lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
alcaeus: Wrapped and updated to say "group accumulators" 👍

lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
alcaeus: Yes, I wanted to show that this takes at least one expression as an argument but can take more, hence the two parameters. I've also updated this to use `func_get_args()`.

lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
alcaeus: Simplified this for operators that take an array of arguments.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Operator.php
alcaeus: Added `@see` comments to all methods in the interfaces.

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/BucketAutoTest.php
alcaeus: The space is not problematic, but the union is. In this case I flipped the order of the types in the union, but in other cases it's a good idea to disambiguate this using parentheses.

alcaeus:
Note: test failures on PHP 7.4 are due to `self` not being sufficient for what we want to do. Using a `static` return type will fix the problem but requires PHP 8. We'll most likely drop support for PHP 7.4 as a result.

phpunit.xml.dist
jmikola: Wait, isn't "match" [prohibited in PHP 8.0+](https://www.php.net/manual/en/reserved.keywords.php)? Ignoring PHPUnit, how is that allowed in the first place?

alcaeus: All logic was moved from `Match` to `MatchStage`, and `Match` changed to extend `MatchStage`. Then, all ODM code was updated to only work with `MatchStage` instances. The `Match` classes are only loaded if a user specifically references them, which will cause errors on PHP 8.

On second thought, this introduced a subtle BC break which we didn't notice until now, but considering that nobody complained about it I'm not going to elaborate or care about it.

See https://github.com/doctrine/mongodb-odm/commit/8a5bf0e03f7c9be6b3f933559a353c912e0e8dfd for the commit that included the change.

lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
```diff
@@ -584,205 +376,124 @@ public function dateFromString($dateString, $format = null, $timezone = null, $o
...
     public function dateToParts($date, $timezone = null, $iso8601 = null): self
     {
-        return $this->operator('$dateToParts', $this->filterOptionalNullArguments([
-            'date' => $date,
-            'timezone' => $timezone,
-            'iso8601' => $iso8601,
-        ], ['timezone', 'iso8601']));
+        return $this->operator(
+            '$dateToParts',
+            $this->filterOptionalNullArguments(
```
jmikola, line 370: Just a thought but there may be an opporunity to replace the `$this->filterOptionalNullArguments()` stuff with another helper that builds an associative array from `func_get_args()` and filters automatically based on their default value, assuming this is common enough. I expect that would require method reflection so certainly a bit more overhead at runtime, though.

Not actually suggesting you pursue that, but I wanted to brain dump the idea.

lib/Doctrine/ODM/MongoDB/Aggregation/Operator/CustomOperators.php
```diff
@@ -28,15 +29,17 @@ interface CustomOperators
...
+     * @param string|Javascript      $accumulate
+     * @param mixed|Expr             $accumulateArgs
+     * @param string|Javascript      $merge
+     * @param string|Javascript|null $finalize
+     * @param string                 $lang
+     *
+     * @return static
      */
-    public function accumulator($init, $initArgs, $accumulate, $accumulateArgs, $merge, $finalize, $lang = 'js'): self;
+    public function accumulator($init, $initArgs, $accumulate, $accumulateArgs, $merge, $finalize = null, $lang = 'js'): self;
```
jmikola: Good catch on `$finalize`. I realize `$initArgs` are also optional. Do you think it makes sense to default that to `null` as well? It'd be convenient for folks using named params, as they could omit it.

alcaeus: Can do if we want to change the signature to have `$initArgs` just before `finalize`. Despite named arguments all parameters with a default value have to come after those without one. Happy to make the change if we don't care about keeping the arg list together with the code.

lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php
alcaeus: Yep, that would include some reflection logic as well, as `func_get_args` doesn't give us argument names. In a perfect world we'd use code generation for this, but introducing that in ODM is a big effort that I'm not sure is worth it.

lib/Doctrine/ODM/MongoDB/Aggregation/Operator/CustomOperators.php
jmikola: Ah, here's the deprecation notice: https://3v4l.org/QMnWc#v8.2.4

I think it's fine to move `$initArgs` to the first optional param. It's probably less likely to be used in practice -- assuming anyone even makes use of this helper to begin with.

jmikola (approved):
Feel free to make the change to `accumulator()` arg order on your own.

Everything else LGTM.

lib/Doctrine/ODM/MongoDB/Aggregation/Operator/CustomOperators.php
alcaeus: Done :)

alcaeus:
Ignoring PHPStan as those failures are the same we're already seeing due to psalm types.

Issue #2460, opened 2022-09-08
Add  $search the official aggregation stage for running search queries


Alaarafeh:
Doctrine\ODM\MongoDB\Aggregation\Stage we have all possible aggregation stages excapt two Stages:

1. $search 
2. $searchMeta

Useful links:
https://www.mongodb.com/docs/atlas/atlas-search/query-syntax/

malarzm:
@Alaarafeh would you be interested in submitting a PR adding missing stages and tests for them?

Alaarafeh:
@malarzm  i could give try i have already patch in my project and works for me , i will try to create PR but i don't know what version policy you have as PR (should i create new branch from current version as bug/feature) ? and which kind of tests you used (unit, integration or functional)?

malarzm:
> should i create new branch from current version as bug/feature

Your contribution will be a feature in 2.5. Please target 2.4.x branch for now however, I'll take care of branching 2.5.x off once I'm done with bringing back CI to life.

> which kind of tests you used

Functional test is not needed, we're mostly interested whether the stage produces correct query after usage. Something like https://github.com/doctrine/mongodb-odm/blob/2.4.x/tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/LookupTest.php or https://github.com/doctrine/mongodb-odm/blob/2.4.x/tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/CountTest.php :)

malarzm:
@Alaarafeh `2.5.x` was just created, please start your work from there whenever you have a chance :)

GromNaN:
There is 2 sides for this feature:

1. [Defining Atlas Search Indexes](https://www.mongodb.com/docs/atlas/atlas-search/index-definitions/)
2. [Add `$search` operators](https://www.mongodb.com/docs/atlas/atlas-search/operators-and-collectors/) to the aggregation stages

alcaeus:
#2516 adds the `$search` stage to the aggregation builder, along with almost all operators that the stage supports.

Defining Atlas Search index currently isn't possible through the driver, but only through the UI. The documentation will contain information on how to set up these indexes, but there is a project at MongoDB to expose search index creation to the drivers. I have no further information on the timeline of that yet.

GromNaN:
Feature documentation: 
[...]

PR #2515, opened 2023-03-24
Drop support for PHP 7.4

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

This PR drops support for PHP 7.4. This is mainly required due to the aggregation pipeline builder needing to enforce `static` types which cannot be worked around (see #2514).

Contained in this PR is a commit that also enforces PHP 8.0 syntax in phpcs. All changes were made using phpcbf and should be safe, but I can drop the commit if we want to defer this to a later PR.

PR #2516, opened 2023-03-28
Add $search stage to aggregation pipeline builder

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | #2460 

#### Summary

This adds the [`$search` aggregation pipeline stage](https://www.mongodb.com/docs/atlas/atlas-search/query-syntax/#mongodb-pipeline-pipe.-search) to the aggregation builder. This stage only works on MongoDB Atlas with Search indexes.

The following is missing from this PR:
* The [`span` operator](https://www.mongodb.com/docs/atlas/atlas-search/span/) - I'm currently looking into the documentation as it is using undocumented operators
* The [`facet` collector](https://www.mongodb.com/docs/atlas/atlas-search/facet/) - this is relevant for the `$searchMeta` stage but would conflict with the `facet` stage that people could add down the line. For now, users should use the [`$$SEARCH_META` aggregation variable](https://www.mongodb.com/docs/atlas/atlas-search/facet/#std-label-fts-facet-aggregation-variable)
* Documentation on how to create indexes (which is only possible through Atlas) and how to use the `$search` stage itself.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search.php
@@ -1,0 +1,147 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ODM\MongoDB\Aggregation\Stage;
+
+use Doctrine\ODM\MongoDB\Aggregation\Builder;
+use Doctrine\ODM\MongoDB\Aggregation\Stage;
+use Doctrine\ODM\MongoDB\Aggregation\Stage\Search\SearchOperator;
+use Doctrine\ODM\MongoDB\Aggregation\Stage\Search\SupportsAllSearchOperators;
+use Doctrine\ODM\MongoDB\Aggregation\Stage\Search\SupportsAllSearchOperatorsTrait;
+
+/**
+ * @psalm-type CountType = 'lowerBound'|'total'
+ * @psalm-type SearchStageExpression = array{
+ *     '$search': object{
+ *         index?: string,
+ *         count?: object{
+ *            type: CountType,
+ *            threshold?: int,
+ *         },
+ *         highlight?: object{
+ *             path: string,
+ *             maxCharsToExamine?: int,
+ *             maxNumPassages?: int,
+ *         },
+ *         returnStoredSource?: bool,
+ *         autocomplete?: object,
+ *         compound?: object,
+ *         embeddedDocument?: object,
+ *         equals?: object,
+ *         exists?: object,
+ *         geoShape?: object,
+ *         geoWithin?: object,
+ *         moreLikeThis?: object,
+ *         near?: object,
+ *         phrase?: object,
+ *         queryString?: object,
+ *         range?: object,
+ *         regex?: object,
+ *         text?: object,
+ *         wildcard?: object,
+ *     }
+ * }
+ */
+class Search extends Stage implements SupportsAllSearchOperators
+{
+    use SupportsAllSearchOperatorsTrait;
+
+    private string $indexName         = '';
+    private ?object $count            = null;
+    private ?object $highlight        = null;
+    private ?bool $returnStoredSource = null;
+    private ?SearchOperator $operator = null;
+
+    public function __construct(Builder $builder)
+    {
+        parent::__construct($builder);
+    }
+
+    /** @psalm-return SearchStageExpression */
+    public function getExpression(): array
+    {
+        $params = (object) [];
+
+        if ($this->indexName) {
+            $params->index = $this->indexName;
+        }
+
+        if ($this->count) {
+            $params->count = $this->count;
+        }
+
+        if ($this->highlight) {
+            $params->highlight = $this->highlight;
+        }
+
+        if ($this->returnStoredSource !== null) {
+            $params->returnStoredSource = $this->returnStoredSource;
+        }
+
+        if ($this->operator !== null) {
+            $operatorName          = $this->operator->getOperatorName();
+            $params->$operatorName = $this->operator->getOperatorParams();
+        }
+
+        return ['$search' => $params];
+    }
+
+    public function index(string $name): static
+    {
+        $this->indexName = $name;
+
+        return $this;
+    }
+
+    /** @psalm-param CountType $type */
+    public function countDocuments(string $type, ?int $threshold = null): static
+    {
+        $this->count = (object) ['type' => $type];
+
+        if ($threshold !== null) {
+            $this->count->threshold = $threshold;
+        }
+
+        return $this;
+    }
+
+    public function highlight(string $path, ?int $maxCharsToExamine = null, ?int $maxNumPassages = null): static
+    {
+        $this->highlight = (object) ['path' => $path];
+
+        if ($maxCharsToExamine !== null) {
+            $this->highlight->maxCharsToExamine = $maxCharsToExamine;
+        }
+
+        if ($maxNumPassages !== null) {
+            $this->highlight->maxNumPassages = $maxNumPassages;
+        }
+
+        return $this;
+    }
+
+    public function returnStoredSource(bool $returnStoredSource = true): static
+    {
+        $this->returnStoredSource = $returnStoredSource;
+
+        return $this;
+    }
+
+    /**
+     * @param T $operator
+     *
+     * @return T
+     *
+     * @template T of SearchOperator
+     */
+    protected function addOperator(SearchOperator $operator): SearchOperator
+    {
+        return $this->operator = $operator;
+    }
+
+    protected function getSearchStage(): static
+    {
+        return $this;
+    }
+}
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/AbstractSearchOperator.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/AbstractSearchOperator.php
@@ -36,7 +36,7 @@ abstract class AbstractSearchOperator extends Stage implements SearchOperator
     }
 
     /** @return array<string, object> */
-    public function getExpression(): array
+    final public function getExpression(): array
     {
         return [$this->getOperatorName() => $this->getOperatorParams()];
     }
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/AbstractSearchOperator.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/AbstractSearchOperator.php
@@ -1,0 +1,48 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ODM\MongoDB\Aggregation\Stage\Search;
+
+use Doctrine\ODM\MongoDB\Aggregation\Stage;
+use Doctrine\ODM\MongoDB\Aggregation\Stage\Search;
+
+/** @internal */
+abstract class AbstractSearchOperator extends Stage implements SearchOperator
+{
+    public function __construct(private Search $search)
+    {
+        parent::__construct($search->builder);
+    }
+
+    public function index(string $name): Search
+    {
+        return $this->search->index($name);
+    }
+
+    public function countDocuments(string $type, ?int $threshold = null): Search
+    {
+        return $this->search->countDocuments($type, $threshold);
+    }
+
+    public function highlight(string $path, ?int $maxCharsToExamine = null, ?int $maxNumPassages = null): Search
+    {
+        return $this->search->highlight($path, $maxCharsToExamine, $maxNumPassages);
+    }
+
+    public function returnStoredSource(bool $returnStoredSource): Search
+    {
+        return $this->search->returnStoredSource($returnStoredSource);
+    }
+
+    /** @return array<string, object> */
+    public function getExpression(): array
+    {
+        return [$this->getOperatorName() => $this->getOperatorParams()];
+    }
+
+    protected function getSearchStage(): Search
+    {
+        return $this->search;
+    }
+}
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/Autocomplete.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/Autocomplete.php
@@ -1,0 +1,91 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ODM\MongoDB\Aggregation\Stage\Search;
+
+use Doctrine\ODM\MongoDB\Aggregation\Stage\Search;
+
+use function array_values;
+
+/** @internal */
+class Autocomplete extends AbstractSearchOperator implements ScoredSearchOperator
+{
+    use ScoredSearchOperatorTrait;
+
+    /** @var list<string> */
+    private array $query;
+    private string $path;
+    private string $tokenOrder = '';
+    private ?object $fuzzy     = null;
+
+    public function __construct(Search $search, string $path, string ...$query)
+    {
+        parent::__construct($search);
+
+        $this->query(...$query);
+        $this->path($path);
+    }
+
+    public function query(string ...$query): static
+    {
+        $this->query = array_values($query);
+
+        return $this;
+    }
+
+    public function path(string $path): static
+    {
+        $this->path = $path;
+
+        return $this;
+    }
+
+    public function tokenOrder(string $order): static
+    {
+        $this->tokenOrder = $order;
+
+        return $this;
+    }
+
+    public function fuzzy(?int $maxEdits = null, ?int $prefixLength = null, ?int $maxExpansions = null): static
+    {
+        $this->fuzzy = (object) [];
+        if ($maxEdits !== null) {
+            $this->fuzzy->maxEdits = $maxEdits;
+        }
+
+        if ($prefixLength !== null) {
+            $this->fuzzy->prefixLength = $prefixLength;
+        }
+
+        if ($maxExpansions !== null) {
+            $this->fuzzy->maxExpansions = $maxExpansions;
+        }
+
+        return $this;
+    }
+
+    public function getOperatorName(): string
+    {
+        return 'autocomplete';
+    }
+
+    public function getOperatorParams(): object
+    {
+        $params = (object) [
+            'query' => $this->query,
+            'path' => $this->path,
+        ];
+
+        if ($this->tokenOrder) {
+            $params->tokenOrder = $this->tokenOrder;
+        }
+
+        if ($this->fuzzy) {
+            $params->fuzzy = $this->fuzzy;
+        }
+
+        return $this->appendScore($params);
+    }
+}
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/Autocomplete.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/Autocomplete.php
@@ -8,7 +8,11 @@ use Doctrine\ODM\MongoDB\Aggregation\Stage\Search;
 
 use function array_values;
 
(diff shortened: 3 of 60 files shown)


alcaeus:
Creating as a draft for initial review until #2515 is merged.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/Compound/CompoundedAutocomplete.php
```diff
@@ -0,0 +1,16 @@
...
+declare(strict_types=1);
+
+namespace Doctrine\ODM\MongoDB\Aggregation\Stage\Search\Compound;
+
+use Doctrine\ODM\MongoDB\Aggregation\Stage\Search\Autocomplete;
+use Doctrine\ODM\MongoDB\Aggregation\Stage\Search\CompoundedSearchOperatorTrait;
+use Doctrine\ODM\MongoDB\Aggregation\Stage\Search\CompoundSearchOperatorInterface;
+use Doctrine\ODM\MongoDB\Aggregation\Stage\Search\SupportsCompoundableOperatorsTrait;
+
+class CompoundedAutocomplete extends Autocomplete implements CompoundSearchOperatorInterface
```
alcaeus, line 13: All of these compounded operator classes could be created as anonymous classes, but that trips psalm into an endless recursion: https://github.com/vimeo/psalm/issues/9573.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search.php
```diff
@@ -0,0 +1,147 @@
...
+
+use Doctrine\ODM\MongoDB\Aggregation\Builder;
+use Doctrine\ODM\MongoDB\Aggregation\Stage;
+use Doctrine\ODM\MongoDB\Aggregation\Stage\Search\SearchOperator;
+use Doctrine\ODM\MongoDB\Aggregation\Stage\Search\SupportsAllSearchOperators;
+use Doctrine\ODM\MongoDB\Aggregation\Stage\Search\SupportsAllSearchOperatorsTrait;
+
+/**
+ * @psalm-type CountType = 'lowerBound'|'total'
+ * @psalm-type SearchStageExpression = array{
```
jmikola, line 15: Is this the first time you've added detailed typing information to the return value of `getExpression()`? I don't recall seeing this in previous PRs, but perhaps that's because those weren't adding a new stage.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search.php
```diff
@@ -0,0 +1,147 @@
...
+            $params->highlight = $this->highlight;
+        }
+
+        if ($this->returnStoredSource !== null) {
+            $params->returnStoredSource = $this->returnStoredSource;
+        }
+
+        if ($this->operator !== null) {
+            $operatorName          = $this->operator->getOperatorName();
+            $params->$operatorName = $this->operator->getOperatorParams();
```
jmikola, line 84: I assume these will never conflict with existing `$search` options? Is that something you should be asserting?

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/GeoShape.php
```diff
@@ -0,0 +1,76 @@
...
+
+    /** @var LineString|Point|Polygon|MultiPolygon|array|null */
+    private $geometry = null;
+
+    /** @param LineString|Point|Polygon|MultiPolygon|array|null $geometry */
+    public function __construct(Search $search, $geometry = null, string $relation = '', string ...$path)
+    {
+        parent::__construct($search);
+
+        $this
```
jmikola, line 35: Is it typical for the builder constructors to use public setter methods instead of directly assigning properties?

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/Wildcard.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ODM\MongoDB\Aggregation\Stage\Search;
+
+/** @internal */
+class Wildcard extends Regex
```
jmikola: https://www.mongodb.com/docs/atlas/atlas-search/wildcard/ doesn't talk about a common relationship between `wildcard` and `regex`. Should this inherit the other, or does it make more sense for them to extend a common base class?

If you don't want to add an extra class for that reason, I think a comment here explaining that you're only inheriting for convenience (and not a relationship in the design of Atlas Search) would make sense.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/Autocomplete.php
```diff
@@ -0,0 +1,91 @@
...
+
+    public function getOperatorName(): string
+    {
+        return 'autocomplete';
+    }
+
+    public function getOperatorParams(): object
+    {
+        $params = (object) [
+            'query' => $this->query,
```
jmikola, line 81: According to the [Atlas docs](https://www.mongodb.com/docs/atlas/atlas-search/autocomplete/#options), this can be a string or array of strings. Is it worth conditionally flattening this to a single string if it's an array with one element?

I understand there may be no significant benefit to doing that, so I'm not arguing one way or another.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/ScoredSearchOperatorTrait.php
```diff
@@ -0,0 +1,51 @@
...
+declare(strict_types=1);
+
+namespace Doctrine\ODM\MongoDB\Aggregation\Stage\Search;
+
+/** @internal */
+trait ScoredSearchOperatorTrait
+{
+    private ?object $score = null;
+
+    protected function appendScore(object $params): object
```
jmikola: What necessitates this being `protected` instead of `private`?

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/ScoredSearchOperatorTrait.php
```diff
@@ -0,0 +1,51 @@
...
+
+        if ($path !== null) {
+            $boost->path = $path;
+        }
+
+        if ($undefined !== null) {
+            $boost->undefined = $undefined;
+        }
+
+        $this->score = (object) ['boost' => $boost];
```
jmikola, line 38: Should it be an error for this to replace an existing value for `$this->score`? I'm just curious if users might inadvertently shoot themselves in the foot by calling `boostScore()` and then `constantScore()`. Or perhaps that's the very point of a builder since they could use some method that returns a pre-built operator and then call additional methods to override things.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/Compound.php
```diff
@@ -0,0 +1,121 @@
...
+    {
+        return 'compound';
+    }
+
+    public function getOperatorParams(): object
+    {
+        $params = (object) [];
+
+        foreach ($this->operators as $clause => $operators) {
+            if (! $operators) {
```
jmikola, line 95: Should this be more explicit and check that `$operators` is empty? I would have expected a CI check to complain about using an array as-is in a conditional.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/Text.php
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    public function getOperatorParams(): object
+    {
+        $params = (object) [
+            'query' => $this->query,
+            'path' => $this->path,
+        ];
+
+        if ($this->fuzzy) {
```
jmikola, line 77: Should you be using `!== null` here to be consistent? Not sure if this would trigger code analysis warnings for using an object in a conditional.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/QueryString.php
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    public function getOperatorParams(): object
+    {
+        $params = (object) [
+            'query' => $this->query,
+            'defaultPath' => $this->defaultPath,
+        ];
+
+        return $this->appendScore($params);
```
jmikola, line 56: I assume this method should be called in each `getOperatorParams()` implementation where the class itself implements ScoredSearchOperator. Do you have any way to enforce that or does it just come down to remembering to return with this method?

I imagine it could be done if there was a parent method that took care of abstracting the call to `getOperatorParams()` (e.g. if this was instead `getBaseOperatorParams()`), but that'd probably just further complicate things.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/MoreLikeThis.php
```diff
@@ -0,0 +1,40 @@
...
+    /** @param array<string, mixed>|object $documents */
+    public function __construct(Search $search, ...$documents)
+    {
+        parent::__construct($search);
+
+        $this->like = array_values($documents);
+    }
+
+    /** @return array<string, object> */
+    public function getExpression(): array
```
jmikola: How is this different than `AbstractSearchOperator::getExpression()`?

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/MoreLikeThis.php
```diff
@@ -0,0 +1,40 @@
...
+declare(strict_types=1);
+
+namespace Doctrine\ODM\MongoDB\Aggregation\Stage\Search;
+
+use Doctrine\ODM\MongoDB\Aggregation\Stage\Search;
+
+use function array_values;
+
+/** @internal */
+class MoreLikeThis extends AbstractSearchOperator
```
jmikola, line 16: I think it'd be helpful to add `@see` links to these operators. The links you need should be found in https://www.mongodb.com/docs/atlas/atlas-search/operators-and-collectors/#operators.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/Exists.php
```diff
@@ -0,0 +1,32 @@
...
+/** @internal */
+class Exists extends AbstractSearchOperator
+{
+    public function __construct(Search $search, private string $path = '')
+    {
+        parent::__construct($search);
+    }
+
+    /** @return array<string, object> */
+    public function getExpression(): array
```
jmikola: Is this redundant? Similar to what I commented on in MoreLikeThis.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/AbstractSearchOperator.php
```diff
@@ -0,0 +1,48 @@
...
+        return $this->search->highlight($path, $maxCharsToExamine, $maxNumPassages);
+    }
+
+    public function returnStoredSource(bool $returnStoredSource): Search
+    {
+        return $this->search->returnStoredSource($returnStoredSource);
+    }
+
+    /** @return array<string, object> */
+    public function getExpression(): array
```
jmikola: Does any subclass actually need to override this? If not, should you make it `final`?

That goes for some other methods here as well if you don't think there's any need to override.

jmikola (commented):
Here's a first pass.

FYI: I had trouble following what all of the compound operator classes were used for. I understand the syntax in Atlas Search (although just barely), but this was too much to wrap my head around in a PR.

alcaeus:
> I had trouble following what all of the compound operator classes were used for.

Yeah, that's where stuff gets messy. In a nutshell, the `$search` pipeline stage supports a single operator, for example:
```
{ $search: { query: 'jmikola', path: 'content' } }
```

This is constructed using:
```php
$builder->search()
    ->text()
        ->query('jmikola')
        ->path('content')
```

However, when using compound, one can combine multiple operators in one of four conditions (must, mustNot, should, filter). For example:
```php
$builder->search()
    ->compound()
        ->must()
            ->text()->query('jmikola')->path('content')
            ->equals()->value('alcaeus')->path('author')
        ->mustNot()
            ->text()->query('bad words')->path('content')
```

In this case, the result of `text()` has to return not only an instance of the `Text` class used to configure the operator, but at the same time an instance of the surrounding `compound` operator to allow for switching to another filter. We can do this by composing the original operator class with the interface for the `compound` operator and using a trait to fill in the functionality (as every operator added will have to be wrapped as well). I had originally planned returning an anonymous class instance for this, but Psalm did not agree with me there so I had to create those separate `Compound...` operator classes.

The idea behind all this is to not rely on some `__call` magic and instead provide proper typing information so tools like PhpStorm can show correct code completion options when using the fluent interface.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search.php
alcaeus: I've added these to a number of other stages in #2513 to provide better typing for the stages. It started with a few types for parameters, then I figured I might as well start to better document the return type of `getExpression` to start moving away from just plain `array`.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search.php
alcaeus: I figured this wouldn't conflict, as they are defined by the Atlas team and added via a trait. I'm sure it's not completely idiot proof and if one really wanted to they'd be able to generate an invalid pipeline using custom operators, but I'm not overly concerned about that.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/GeoShape.php
alcaeus: I've used this in some other builders as well so I decided to apply the same pattern here. I don't think it makes any difference in most cases, so I'm happy to directly assign properties if you prefer that.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/Wildcard.php
alcaeus: I mainly reused it because it uses the same options. Happy to separate the classes to not create an inheritance chain that we may have to break up down the road if one of the operators diverges from the other.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/Autocomplete.php
alcaeus: I thought about it, then decided it's not worth it. That's an optimisation I'll leave to the Atlas team.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/GeoShape.php
jmikola: If there's prior art for using methods this is fine as-is. I wasn't sure so wanted to defer to you.

I'm not worried about the overhead of function calls :)

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/ScoredSearchOperatorTrait.php
alcaeus: Your second guess is correct: since the builder is also designed to be used programmatically, one piece of code might want to change the scoring option when it has already been set, hence the replacement.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/Compound.php
alcaeus: I usually short-circuit such checks, as `if ($someArray)` is not any different from `if ($someArray !== [])` for empty arrays.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/Text.php
alcaeus: Since the variable can only be `null` or an object, we can short-circuit this. This would be different if this were untyped and it could have a falsy value like `'0'` which can cause unwanted effects. Since `null` is the only possible value to evaluate to `false` here, I think this is safe.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/QueryString.php
alcaeus: Technically we could move this to `AbstractSearchOperator::getExpression`, which could call `getOperatorParams()` to get the parameters and then pass it on to `appendScore` if it is an instance of `ScoredSearchOperator`. This way is more error-prone, but more clearly separates the concerns as opposed to `AbstractSearchOperator` having knowledge of what to do with any `ScoredSearchOperator`.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/MoreLikeThis.php
alcaeus: It isn't, will remove 👍

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/MoreLikeThis.php
alcaeus: Will do 👍

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/Exists.php
alcaeus: Same here, will remove 👍

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/AbstractSearchOperator.php
alcaeus: That is correct - I think I'll make some of these methods `final` to prevent shenanigans.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/ScoredSearchOperatorTrait.php
alcaeus: Nothing, just a good old fashioned layer 8 problem 👍

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/QueryString.php
jmikola: Aye, I'd be more worried about maintaining separation of concerns and not further complicating the base class. If you do mess this up, it should be caught by tests anyway.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/Wildcard.php
jmikola: I think that makes sense. Alternatively, just duplicate the class and leave a note in the doc block of each that there's a potential for refactoring if a third related class were to enter the picture.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/MoreLikeThis.php
jmikola: Note that there were a few occurrences of this so make sure you get them all. Feel free to resolve this thread after reading.

jmikola (approved):

PR #2517, opened 2023-03-28
Add $setWindowFields aggregation pipeline stage and window operators

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | 

#### Summary

Last but not least, this pull request adds support for window operators and the `$setWindowFields` aggregation pipeline stage to the builder. There is some code duplication between group accumulators and window operators, as the latter includes all group accumulators except for `$mergeObjects`. To avoid polluting the interfaces and traits, I've decided to accept the duplication considering that the methods only proxy calls to the `Expr` class.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Operator/ProvidesWindowAccumulatorOperators.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Operator/ProvidesWindowAccumulatorOperators.php
@@ -1,0 +1,225 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ODM\MongoDB\Aggregation\Operator;
+
+use Doctrine\ODM\MongoDB\Aggregation\Expr;
+
+use function func_get_args;
+
+/** @internal */
+trait ProvidesWindowAccumulatorOperators
+{
+    abstract protected function getExpr(): Expr;
+
+    public function addToSet($expression): static
+    {
+        $this->getExpr()->addToSet(...func_get_args());
+
+        return $this;
+    }
+
+    public function avg($expression, ...$expressions): static
+    {
+        $this->getExpr()->avg(...func_get_args());
+
+        return $this;
+    }
+
+    public function bottom($output, $sortBy): static
+    {
+        $this->getExpr()->bottom(...func_get_args());
+
+        return $this;
+    }
+
+    public function bottomN($output, $sortBy, $n): static
+    {
+        $this->getExpr()->bottomN(...func_get_args());
+
+        return $this;
+    }
+
+    public function countDocuments(): static
+    {
+        $this->getExpr()->countDocuments();
+
+        return $this;
+    }
+
+    public function covariancePop($expression1, $expression2): static
+    {
+        $this->getExpr()->covariancePop(...func_get_args());
+
+        return $this;
+    }
+
+    public function covarianceSamp($expression1, $expression2): static
+    {
+        $this->getExpr()->covarianceSamp(...func_get_args());
+
+        return $this;
+    }
+
+    public function denseRank(): static
+    {
+        $this->getExpr()->denseRank();
+
+        return $this;
+    }
+
+    public function derivative($input, string $unit): static
+    {
+        $this->getExpr()->derivative(...func_get_args());
+
+        return $this;
+    }
+
+    public function documentNumber(): static
+    {
+        $this->getExpr()->documentNumber();
+
+        return $this;
+    }
+
+    public function expMovingAvg($input, ?int $n = null, ?float $alpha = null): static
+    {
+        $this->getExpr()->expMovingAvg(...func_get_args());
+
+        return $this;
+    }
+
+    public function first($expression): static
+    {
+        $this->getExpr()->first(...func_get_args());
+
+        return $this;
+    }
+
+    public function firstN($expression, $n): static
+    {
+        $this->getExpr()->firstN(...func_get_args());
+
+        return $this;
+    }
+
+    public function integral($input, string $unit): static
+    {
+        $this->getExpr()->integral(...func_get_args());
+
+        return $this;
+    }
+
+    public function last($expression): static
+    {
+        $this->getExpr()->last(...func_get_args());
+
+        return $this;
+    }
+
+    public function lastN($expression, $n): static
+    {
+        $this->getExpr()->lastN(...func_get_args());
+
+        return $this;
+    }
+
+    public function linearFill($expression): static
+    {
+        $this->getExpr()->linearFill(...func_get_args());
+
+        return $this;
+    }
+
+    public function locf($expression): static
+    {
+        $this->getExpr()->locf(...func_get_args());
+
+        return $this;
+    }
+
+    public function max($expression, ...$expressions): static
+    {
+        $this->getExpr()->max(...func_get_args());
+
+        return $this;
+    }
+
+    public function maxN($expression, $n): static
+    {
+        $this->getExpr()->maxN(...func_get_args());
+
+        return $this;
+    }
+
+    public function min($expression, ...$expressions): static
+    {
+        $this->getExpr()->min(...func_get_args());
+
+        return $this;
+    }
+
+    public function minN($expression, $n): static
+    {
+        $this->getExpr()->minN(...func_get_args());
+
+        return $this;
+    }
+
+    public function push($expression): static
+    {
+        $this->getExpr()->push(...func_get_args());
+
+        return $this;
+    }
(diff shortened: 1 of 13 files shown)


lib/Doctrine/ODM/MongoDB/Aggregation/Operator/WindowOperators.php
```diff
@@ -0,0 +1,318 @@
...
+interface WindowOperators
+{
+    /**
+     * Returns an array of unique expression values for each group.
+     *
+     * @see https://docs.mongodb.com/manual/reference/operator/aggregation/addToSet/
+     *
+     * @param mixed|Expr $expression
+     */
+    public function addToSet($expression): static;
```
jmikola, line 26: I assume some of these methods are defined in other interfaces as well, since you didn't have to add these methods to Expr when implementing WindowOperators.

Does anything enforce that the doc blocks and typing (outside of the actual PHP method signature) are consistent?

lib/Doctrine/ODM/MongoDB/Aggregation/Operator/ProvidesWindowAccumulatorOperators.php
```diff
@@ -0,0 +1,225 @@
...
+use function func_get_args;
+
+/** @internal */
+trait ProvidesWindowAccumulatorOperators
+{
+    abstract protected function getExpr(): Expr;
+
+    public function addToSet($expression): static
+    {
+        $this->getExpr()->addToSet(...func_get_args());
```
jmikola: In #2516 I noticed you avoided using `...func_get_args()` for general forwarding calls like this and instead hard-coded the argument list. If there's no performance concern with using this I think you should probably apply it to the other PR as well.

lib/Doctrine/ODM/MongoDB/Aggregation/Operator/ProvidesWindowAccumulatorOperators.php
```diff
@@ -0,0 +1,225 @@
...
+declare(strict_types=1);
+
+namespace Doctrine\ODM\MongoDB\Aggregation\Operator;
+
+use Doctrine\ODM\MongoDB\Aggregation\Expr;
+
+use function func_get_args;
+
+/** @internal */
+trait ProvidesWindowAccumulatorOperators
```
jmikola: Where does "Accumulator" come from? I would have expected this to be ProvideWindowOperators and match the interface name more directly.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/SetWindowFields/Output.php
```diff
@@ -0,0 +1,128 @@
...
+        $this->expr->field($fieldName);
+
+        return $this;
+    }
+
+    /**
+     * Specifies the window boundaries and parameters.
+     *
+     * @psalm-param WindowBounds|null $documents
+     * @psalm-param WindowBounds|null $range
```
jmikola, line 82: Are the PHPStan failures below the result of PHPStan failing to resolve the psalm type alias in the class doc block?

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/SetWindowFields/Output.php
```diff
@@ -0,0 +1,128 @@
...
+            ],
+            static fn ($value): bool => $value !== null,
+        );
+
+        return $this;
+    }
+
+    public function getExpression(): array
+    {
+        return array_merge_recursive(
```
jmikola, line 106: If I understand correctly, the result of `array_map()` is going to be an associative array of field names to `{ window: ... }` objects. `Expr::getExpression()` is documented as returning `array<string, mixed>`.

Isn't that problematic if the expression value isn't an array? Actually, even if it is an array how do you merge a `stdClass` into it?

lib/Doctrine/ODM/MongoDB/Aggregation/Operator/ProvidesWindowAccumulatorOperators.php
alcaeus: It comes from copypasta - fixed.

lib/Doctrine/ODM/MongoDB/Aggregation/Operator/WindowOperators.php
alcaeus: Not at this time. Since the interface methods don't declare types for parameters, methods implementing these interfaces can't add/change the types accepted. Even if we were to use union types (which is possible now that we're requiring PHP 8), we'd only be able to specify a `mixed` type, due to PHP not allowing types to show up in unions twice.

FWIW, this is one of the reasons for requiring PHP 8, as the `self` return type didn't always work out when combining interfaces that declared the same methods. With `static` being allowed as return type, we no longer have this problem :)

lib/Doctrine/ODM/MongoDB/Aggregation/Operator/ProvidesWindowAccumulatorOperators.php
alcaeus: I made a point to add these as was done in #2514. I'll go through #2516 to make sure I'm using `func_get_args` to forward arguments as it helps avoid errors.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/SetWindowFields/Output.php
alcaeus: Yes. PHPStan also reads psalm-vendored annotations, but doesn't support the full syntax (e.g. object shapes like `object{foo: string}`. Any type it can't resolves is treated as a standard PHP type, which in this case doesn't exist. I'm still thinking about how to work around this without adding a bunch of phpstan-vendored annotations to prevent it from reading psalm annotations.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/SetWindowFields/Output.php
alcaeus: Ah, this is one where I wasn't sure whether using `array_merge_recursive` would be sensible. To explain, while the `Expr` class can generate the expressions used in the `setWindowFields` output, it can't handle the additional `window` option used there. `Expr` can generate this:
```
{ foo: { $rank: {} }
// Generated using $expr->field('foo')->rank();
```

But not this as is valid in a window stage:
```
{ foo: {
    $rank: {},
    window: {...}
}}
```

To work around this, we use the `Expr` class to handle operators and store the windows for fields ourselves. In `getExpression`, we generate the output of the `Expr` object, which gives us something like this:
```
[
    'foo' => ['$rank' => []],
    'bar' => ['$documentNumber' => []],
]
```

The `windows` property contains the windows (in my example, only `foo` uses a window):
```
[
    'foo' => ['documents' => ['unbounded', 'current']],
]
```

The `array_map` call then generates the window portion, resulting in this:
```
[
    'foo' => ['window' => (object) ['documents' => ['unbounded', 'current']]],
]
```

We can then use `array_merge_recursive` to recursively merge windows and field expressions together, resulting in this:
```
[
    'foo' => [
        '$rank' => [],
        'window' => (object) [...],
    ],
    'bar' => ['$documentNumber' => []],
]
```

To make a long story short, I'll add a comment explaining what this does.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/SetWindowFields/Output.php
jmikola: Oh, I overlooked that these are already psalm-vendored annotations. In that case, this seems like yet another argument for dropping PHPStan.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/SetWindowFields/Output.php
```diff
@@ -0,0 +1,146 @@
...
+     * to be wrapped inside a "window" object for the output expression. This
+     * method prepares the window options appropriately for use in
+     * getExpression().
+     *
+     * @return array<string, Window>
+     */
+    private function getWindowOptionsForFields(): array
+    {
+        return array_map(
+            static fn ($window): object => (object) ['window' => $window],
```
jmikola, line 142: https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/issues/5449#issuecomment-764516178 was an interesting read about the relevance of `static` here.

jmikola (approved):

PR #2521, opened 2023-04-03
Restore document proxy state to uninitialized on load exception

notrix:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | 

#### Summary

This PR should fix issue when exception is thrown during proxy object load. If application catches exceptions and flushes at the end of request, unit of work finds changes for fields that have default values. I have seen network glitches of Amazon KMS service on load and successful flush at the end. This broke production environment badly.
Proxy is marked as not initialized if id is not found in database. So I think same should be if exception happens on load event.

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/Proxy/Factory/StaticProxyFactoryTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Proxy/Factory/StaticProxyFactoryTest.php
@@ -1,0 +1,100 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ODM\MongoDB\Tests\Proxy\Factory;
+
+use Closure;
+use Doctrine\ODM\MongoDB\DocumentManager;
+use Doctrine\ODM\MongoDB\Event\DocumentNotFoundEventArgs;
+use Doctrine\ODM\MongoDB\Events;
+use Doctrine\ODM\MongoDB\LockException;
+use Doctrine\ODM\MongoDB\Tests\BaseTestCase;
+use Documents\Cart;
+use MongoDB\Client;
+use MongoDB\Collection;
+use MongoDB\Database;
+use PHPUnit\Framework\MockObject\MockObject;
+use ProxyManager\Proxy\GhostObjectInterface;
+
+class StaticProxyFactoryTest extends BaseTestCase
+{
+    /** @var Client|MockObject */
+    private Client $client;
+
+    public function setUp(): void
+    {
+        parent::setUp();
+
+        $this->dm = $this->createMockedDocumentManager();
+    }
+
+    public function testProxyInitializeWithException(): void
+    {
+        $collection = $this->createMock(Collection::class);
+        $database   = $this->createMock(Database::class);
+
+        $this->client->expects($this->once())
+            ->method('selectDatabase')
+            ->willReturn($database);
+
+        $database->expects($this->once())
+            ->method('selectCollection')
+            ->willReturn($collection);
+
+        $collection->expects($this->once())
+            ->method('findOne')
+            ->willThrowException(LockException::lockFailed(null));
+
+        $uow = $this->dm->getUnitOfWork();
+
+        $proxy = $this->dm->getReference(Cart::class, '123');
+        self::assertInstanceOf(GhostObjectInterface::class, $proxy);
+
+        $closure = static function (DocumentNotFoundEventArgs $eventArgs) {
+            self::fail('DocumentNotFoundListener should not be called');
+        };
+        $this->dm->getEventManager()->addEventListener(Events::documentNotFound, new DocumentNotFoundListener($closure));
+
+        try {
+            $proxy->initializeProxy();
+            self::fail('An exception should have been thrown');
+        } catch (LockException $exception) {
+            self::assertInstanceOf(LockException::class, $exception);
+        }
+
+        $uow->computeChangeSets();
+
+        self::assertFalse($proxy->isProxyInitialized(), 'Proxy should not be initialized');
+    }
+
+    public function tearDown(): void
+    {
+        // db connection is mocked, nothing to clean up
+    }
+
+    private function createMockedDocumentManager(): DocumentManager
+    {
+        $config = $this->getConfiguration();
+
+        $this->client = $this->createMock(Client::class);
+
+        return DocumentManager::create($this->client, $config);
+    }
+}
+
+class DocumentNotFoundListener
+{
+    private Closure $closure;
+
+    public function __construct(Closure $closure)
+    {
+        $this->closure = $closure;
+    }
+
+    public function documentNotFound(DocumentNotFoundEventArgs $eventArgs): void
+    {
+        $closure = $this->closure;
+        $closure($eventArgs);
+    }
+}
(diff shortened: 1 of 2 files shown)


notrix:
@malarzm I would like some comments in here: do you agree that it is a bug to leave ghost object initialized if there was exception loading data?
Is the branch `doctrine:2.6.x` correct to point MR to?
CI issues with PHPStan and Psalm? Are they relevant to my MR and should I fix it (it seems that errors are not related to my MR)

malarzm:
> I would like some comments in here: do you agree that it is a bug to leave ghost object initialized if there was exception loading data?

I'm not sure yet :) Sounds plausible but I haven't given it a deeper thought yet

> Is the branch doctrine:2.6.x correct to point MR to?

Given your description it's better to classify as a bug, in which case `2.5.x` is a correct target

> CI issues with PHPStan and Psalm? Are they relevant to my MR and should I fix it (it seems that errors are not related to my MR)

Please rebase atop 2.5.x - there checks are green :)

tests/Doctrine/ODM/MongoDB/Tests/Proxy/Factory/StaticProxyFactoryTest.php
```diff
@@ -0,0 +1,80 @@
...
+            ->method('selectDatabase')
+            ->willReturn($database);
+
+        $database->expects($this->once())
+            ->method('selectCollection')
+            ->willReturn($collection);
+
+        $collection->expects($this->once())
+            ->method('findOne')
+            ->willThrowException(LockException::lockFailed(null));
```
malarzm, line 47: I'm a bit torn so here goes nothing: how about instead of mocking 3 levels deep we mock `DocumentPersister`, have it thrown an exception on `load`? This would require manipulating `UnitOfWork::$documentPersisters` with reflection though, but the test should be easier to maintain with real database connection. For your consideration :)

tests/Doctrine/ODM/MongoDB/Tests/Proxy/Factory/StaticProxyFactoryTest.php
```diff
@@ -0,0 +1,80 @@
...
+
+        $uow = $this->dm->getUnitOfWork();
+
+        $metadata = $this->dm->getClassMetadata(PaymentProvider::class);
+
+        $proxy = $this->proxyFactory->getProxy($metadata, '123');
+        $uow->registerManaged($proxy, '123', []);
```
malarzm: It's better to call `DocumentManager::getReference`

tests/Doctrine/ODM/MongoDB/Tests/Proxy/Factory/StaticProxyFactoryTest.php
```diff
@@ -0,0 +1,80 @@
...
+
+        try {
+            $proxy->initializeProxy();
```
malarzm, line 60:
```suggestion
            $proxy->initializeProxy();
            self::fail('An exception should have been thrown');
```

Let's save us from any future refactorings that may change expectations on how the exception is thrown.

tests/Doctrine/ODM/MongoDB/Tests/Proxy/Factory/StaticProxyFactoryTest.php
```diff
@@ -0,0 +1,80 @@
...
+        } catch (LockException $exception) {
+            $this->assertInstanceOf(LockException::class, $exception);
```
malarzm: Please use `self::assert*` wherever possible, most of code was migrated to that

tests/Doctrine/ODM/MongoDB/Tests/Proxy/Factory/StaticProxyFactoryTest.php
```diff
@@ -0,0 +1,80 @@
...
+        }
+
+        $uow->computeChangeSets();
+
+        $this->assertFalse($uow->isScheduledForUpdate($proxy), 'Proxy should not be scheduled for update');
```
malarzm: In my opinion this is not needed. The test should care whether proxy is still uninitialized after the exception, not how change set computation works with uninitialized proxies (there are other tests for that).

tests/Doctrine/ODM/MongoDB/Tests/Proxy/Factory/StaticProxyFactoryTest.php
```diff
@@ -0,0 +1,80 @@
...
+        $this->assertFalse($proxy->isProxyInitialized(), 'Proxy should not be initialized');
```
malarzm: If possible, the test should also add an event listener for `documentNotFound` and assert the listener is never called.

tests/Doctrine/ODM/MongoDB/Tests/Proxy/Factory/StaticProxyFactoryTest.php
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    public function tearDown(): void
+    {
+    }
```
malarzm, line 74:
```suggestion
        // db connection is mocked, nothing to clean up
    }
```

For clarity, I wondered why we have an empty `tearDown` :)

tests/Documents/PaymentProvider.php
```diff
@@ -0,0 +1,80 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Documents;
+
+use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
+
+/** @ODM\Document */
+class PaymentProvider
```
malarzm: Could you please use any existing document instead of adding new one?

tests/Doctrine/ODM/MongoDB/Tests/Proxy/Factory/StaticProxyFactoryTest.php
```diff
@@ -0,0 +1,80 @@
...
+
+    protected function createMockedDocumentManager(): DocumentManager
```
malarzm:
```suggestion
    private function createMockedDocumentManager(): DocumentManager
```

No need for protected visibility unless this was supposed to be `createTestDocumentManager` originally?

malarzm (commented):
I'm A-OK with changes in the StaticProxyFactory itself 👍

tests/Doctrine/ODM/MongoDB/Tests/Proxy/Factory/StaticProxyFactoryTest.php
notrix: At first I also wanted to mock DocumentPersister::load, but DocumentPersister is a final class. So my best shot was to mock Collection::findOne

tests/Documents/PaymentProvider.php
notrix: The main issue for me was that my document has default values and those default value were flushed to database. But since we do not assert change set - I have removed new document

tests/Doctrine/ODM/MongoDB/Tests/Proxy/Factory/StaticProxyFactoryTest.php
malarzm: Agreed then!

malarzm (approved):
Thank you very much for changes!

malarzm:
And enjoy [2.5.2](https://github.com/doctrine/mongodb-odm/releases/tag/2.5.2) ;)

PR #2523, opened 2023-04-03
Parameter 4 in `Query::__construct()` can not use an empty array as default value

phansys:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | n/a

#### Summary

Offset "type" is explicitly required in the passed array:

https://github.com/doctrine/mongodb-odm/blob/dacfa0e3c5840b756b58b6dd00c9c717b1725ce6/lib/Doctrine/ODM/MongoDB/Query/Query.php#L143


malarzm (approved):

malarzm:
Thanks @phansys!

PR #2524, opened 2023-04-04
Parameters 2 and 3 in `HydratorFactory::__construct()` are not nullable

phansys:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | n/a

#### Summary

A falsy loose check is performed in the method's body:

https://github.com/doctrine/mongodb-odm/blob/dacfa0e3c5840b756b58b6dd00c9c717b1725ce6/lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php#L84-L90

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
+++ b/lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
@@ -59,16 +59,22 @@ final class HydratorFactory
 
     /**
      * Which algorithm to use to automatically (re)generate hydrator classes.
+     *
+     * @psalm-var Configuration::AUTOGENERATE_*
      */
     private int $autoGenerate;
 
     /**
      * The namespace that contains all hydrator classes.
+     *
+     * @psalm-var non-empty-string
      */
-    private ?string $hydratorNamespace;
+    private string $hydratorNamespace;
 
     /**
      * The directory that contains all hydrator classes.
+     *
+     * @psalm-var non-empty-string
      */
     private string $hydratorDir;
 
@@ -79,14 +85,21 @@ final class HydratorFactory
      */
     private array $hydrators = [];
 
-    /** @throws HydratorException */
-    public function __construct(DocumentManager $dm, EventManager $evm, ?string $hydratorDir, ?string $hydratorNs, int $autoGenerate)
+    /**
+     * @psalm-param Configuration::AUTOGENERATE_* $autoGenerate
+     *
+     * @throws HydratorException
+     *
+     * @psalm-assert non-empty-string $hydratorDir
+     * @psalm-assert non-empty-string $hydratorNs
+     */
+    public function __construct(DocumentManager $dm, EventManager $evm, string $hydratorDir, string $hydratorNs, int $autoGenerate)
     {
-        if (! $hydratorDir) {
+        if ($hydratorDir === '') {
             throw HydratorException::hydratorDirectoryRequired();
         }
 
-        if (! $hydratorNs) {
+        if ($hydratorNs === '') {
             throw HydratorException::hydratorNamespaceRequired();
         }
 


lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
```diff
@@ -79,14 +85,21 @@ final class HydratorFactory
...
-    public function __construct(DocumentManager $dm, EventManager $evm, ?string $hydratorDir, ?string $hydratorNs, int $autoGenerate)
+    /**
+     * @psalm-param Configuration::AUTOGENERATE_* $autoGenerate
+     *
+     * @throws HydratorException
+     *
+     * @psalm-assert non-empty-string $hydratorDir
+     * @psalm-assert non-empty-string $hydratorNs
+     */
+    public function __construct(DocumentManager $dm, EventManager $evm, string $hydratorDir, string $hydratorNs, int $autoGenerate)
```
malarzm, line 96: The change conflicts with type definition in Configuration

https://github.com/doctrine/mongodb-odm/blob/3091159e4758a37dbe92a034cf8a26df2fb84c2d/lib/Doctrine/ODM/MongoDB/Configuration.php#L316

What's the reason for getting rid of null? Nulls are cool :)

malarzm (changes requested):

lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
phansys: How should we proceed then? Changing the return type declaration at `Configuration::getHydratorDir()` (from `?string` to `string`) should be aware about the BC promise, as the class is not final.
The fact is that, the combination of the Not operator (`!`) with `null` always will evaluate to true. So what's the point of allowing an argument type that we already know is not valid for the method?

I realized about this situation while working on a test suite for another package. I passed `null` to these parameters hoping they would perform a default behavior, just to discover later (in runtime) that they are not allowed to be `null`.

malarzm: Ah true, that's the point I've been missing.

> So what's the point of allowing an argument type that we already know is not valid for the method?

From HydratorFactory's point of view it makes no sense, as you pointed out: `null` is not a valid value. But if we look from higher up, it lets user know that the `Configuration` they provided is missing some settings. So all in all it has _some_ value.

I personally see very little value in touching this as there's no "default" behaviour to reason about. The only thing we could do is an improvement that `$hydratorDir` is not required when dealing with `AUTOGENERATE_EVAL` mode and namespace is indeed optional.

phansys: >But if we look from higher up, it lets user know that the Configuration they provided is missing some settings. So all in all it has some value.

Could you please explain why in this case `TypeError` might not be enough to tell the user the passed setting is not using the required type?

>The only thing we could do is an improvement that $hydratorDir is not required when dealing with AUTOGENERATE_EVAL mode and namespace is indeed optional.

IMO, that's completely another story. I agree with this proposal. It makes sense to me and I think it could really be a legit reason to allow `null` in these parameters. But it isn't what we have now.

malarzm: > Could you please explain why in this case `TypeError` might not be enough to tell the user the passed setting is not using the required type?

Because an exception explicitly saying what is wrong is better than `TypeError` from a class the user is not instantiating himself.

phansys: >Because an exception explicitly saying what is wrong is better than TypeError from a class the user is not instantiating himself.

IMO, using the built-in typing mechanism from the language is highly desirable. It does not prevent to add the additional check about a more specific type of value inside the method's body, where `HydratorException` can still be thrown if the passed value does not accomplish the requirements.
In other words, I think deliberately lying about the allowed types is harmful and leads to mistrust, as the typing system from the PHP version used brings the chance to be clear about that.

malarzm: I'd agree, if the `HydratorFactory` would be something users are supposed to instantiate themselves, but it's not the case. It's wired automatically by ODM and 99% will never see (or even care about) ctor's signature. And for those folks, throwing an exception saying that hydrator directory is needed is better than having them investigate why a null value was passed to a service requiring a non-null argument.

malarzm:
Closing as we won't be merging this. Thanks for the proposal @phansys!

PR #2530, opened 2023-04-24
Document new aggregation pipeline builder stages

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

This PR documents the new pipeline stages added in 2.6. To make the aggregation builder documentation more browsable, I've extracted the documentation for each stage to a separate documentation page, similar to what is done for mapping annotations.

This refactoring was done in the first commit, with the second commit adding language for new pipeline stages. I recommend reviewing the second commit individually if only reviewing new content.


docs/en/reference/aggregation-stage-reference.rst
```diff
@@ -0,0 +1,888 @@
...
+        ->count('numSingleItemOrders')
+    ;
+
+The example above returns a single document with the ``numSingleItemOrders``
+containing the number of orders found.
+
+$densify
+--------
+
+Cretaes new documents in a sequence of documents where certain values in a
```
SenseException: Please don't write docs in Latin. :-)

```suggestion
Creates new documents in a sequence of documents where certain values in a
```

docs/en/reference/aggregation-stage-reference.rst
```diff
@@ -0,0 +1,888 @@
...
+- `$replaceWith <https://docs.mongodb.com/manual/reference/operator/aggregation/replaceWith/>`_
+- `$sample <https://docs.mongodb.com/manual/reference/operator/aggregation/sample/>`_
+- `$search <https://www.mongodb.com/docs/atlas/atlas-search/query-syntax/#-search>`_
+- `$set <https://docs.mongodb.com/manual/reference/operator/aggregation/set/>`_
+- `$setWindowFields <https://docs.mongodb.com/manual/reference/operator/aggregation/setWindowFields/>`_
+- `$skip <https://docs.mongodb.com/manual/reference/operator/aggregation/skip/>`_
+- `$sort <https://docs.mongodb.com/manual/reference/operator/aggregation/project/>`_
+- `$sortByCount <https://docs.mongodb.com/manual/reference/operator/aggregation/sortByCount/>`_
+- `$unionWith <https://docs.mongodb.com/manual/reference/operator/aggregation/sortByCount/>`_
+- `$unset <https://docs.mongodb.com/manual/reference/operator/aggregation/sortByCount/>`_
```
SenseException:
```suggestion
- `$unionWith <https://docs.mongodb.com/manual/reference/operator/aggregation/unionWith/>`_
- `$unset <https://docs.mongodb.com/manual/reference/operator/aggregation/unset/>`_
```

SenseException (changes requested):
Thanks for the separation in the mentioned commit.

docs/en/reference/aggregation-stage-reference.rst
```diff
@@ -0,0 +1,888 @@
...
+add missing values between groups of data, or to populate your data with a
+specified range of values. Taking the partition example from the
+`$densify documentation <https://www.mongodb.com/docs/manual/reference/operator/aggregation/densify/#densifiction-with-partitions>`_,
+this is how you would create the pipeline from the example with the aggregation
+builder:
+
+.. code-block:: php
+
+    <?php
+    $builder = $dm->createAggregationBuilder(\Documents\Coffee::class);
```
jmikola: Other code blocks have a line break after `<?php`.

```suggestion
    <?php

    $builder = $dm->createAggregationBuilder(\Documents\Coffee::class);
```

docs/en/reference/aggregation-stage-reference.rst
```diff
@@ -196,6 +228,46 @@ where its results are stored as an array of documents.
...
+    $builder = $dm->createAggregationBuilder(\Documents\StockPrice::class);
+    $builder
+        ->fill()
+            ->sortBy('time', 1)
+            ->output()
+                ->field('interpolated')->linear()
+                ->field('lastValue')->locf()
+                ->field('fixedValue')->value('foo')
+                ->field('computedValue')->value(
+                    $builder->expr()->multiply('$value', 5),
```
jmikola: Seeing `$value` here alongside a `value()` builder method is a bit confusing.

I assume `$value` here is referencing a field path within the document. Would it help to use `$fieldPath` in the example to make this more self-documenting?

docs/en/reference/aggregation-stage-reference.rst
```diff
@@ -416,6 +488,57 @@ You can also use fields defined in previous stages:
...
+
+The ``$merge`` stage is used to write the results of an aggregation pipeline to
+a collection. Unlike the ``$out`` stage, this stage does not replace the entire
+output collection, but lets you define how to handle conflicts or missing data
+in the output collection. ``$merge`` must be the last stage in an aggregation
+pipeline.
+
+The following pipeline uses the ``$merge`` pipeline stage to aggregate orders
+that were created after the last aggregation run (tracked separately in the
+``$lastAggregateRun`` variable) and updates the ``monthlyOrderStats`` collection
```
jmikola: `$lastAggregateRunAt` or `$lastAggregatedAt` would be more clear for a date field. On my first read, I didn't realize this was a date.

Might also help to say it's a `UTCDateTime` type. I didn't check the Orders model but I assume that's what you're using here.

docs/en/reference/aggregation-stage-reference.rst
```diff
@@ -529,13 +652,87 @@ or create a new document for promotion.
...
+
+    <?php
+
+    $builder = $dm->createAggregationBuilder(\Documents\InfectionNumbers::class);
+    $builder
+        ->setWindowFields()
+            ->partitionBy($builder->expr()->year('$purchaseDate'))
+            ->sortBy('purchaseDate', 1)
+            ->output()
+                ->output()
```
jmikola: Did you mean to call `output()` twice here?

The [`$setWindowFields` syntax](https://www.mongodb.com/docs/manual/reference/operator/aggregation/setWindowFields/#syntax) suggests that properties under `output` are field names, so I'd expect the builder to call `output()->field()`.

docs/en/reference/aggregation-stage-reference.rst
```diff
@@ -578,6 +775,62 @@ The example above is equivalent to the following pipeline:
...
+Removes fields from documents. The ``$unset`` stage is an alias for the
+``$project`` stage that removes fields.
+
+.. code-block:: php
+
+    <?php
+
+    $builder = $dm->createAggregationBuilder(\Documents\Orders::class);
+    $builder
+        ->unset('customer', 'shippingAddress.street', 'billingAddress.street')
```
jmikola, line 818: I find it interesting that `unset()` takes multiple arguments while `excludeFields()` and `includeFields()` take a list of fields as a single array argument.

Is there a particular reason you avoided variadics for exclude/include, or are they just older? Is this something worth addressing in a subsequent PR? Are there other functions with a similar inconsistency?

docs/en/reference/aggregation-stage-reference.rst
alcaeus: Changed to `$someField` to indicate that it could be any field. At this point in the docs I'd assume people are familiar with the syntax, but it still doesn't hurt to make it a little clearer.

docs/en/reference/aggregation-stage-reference.rst
alcaeus: Renamed to `$lastAggregateRunAt`. The type here doesn't matter as much as the aggregation builder is aware of the mapping and thus able to convert types as necessary. It's thus possible to use either a `DateTimeInterface` or `UTCDateTime` instance.

docs/en/reference/aggregation-stage-reference.rst
alcaeus: Great catch, the second call to `output()` is indeed unnecessary.

docs/en/reference/aggregation-stage-reference.rst
alcaeus: I'm not entirely sure if there was any other reason, but back when the `$project` stage builder was added there was no way to use variadic args other than mentioning in the docs and then using `func_get_args`. However, we did that in a few other places, so this may very well have been missed by me.

Since we now have a good way to handle variadic args, I figured it would make more sense to build it this way. In my head I've even gone a step further and considered using named variadic args. Consider the following:
```php
// Builder signature: public function set(mixed ...$fields): static;

// Usage
$builder->set(
    numOrders: $builder->expr()->countDocuments(),
    totalValue: $builder->expr()->sum('$orderTotal')
);
```

The only downside with this would be setting more complex field names (e.g. nested documents or other field names containing characters that aren't allowed as PHP identifiers). There is a fallback way to allow a single array argument instead of the variadic list (think `array_keys($fields) === [0] && is_array($fields[0])`), but I'm not too sure about it yet. This might be part of the builder that we're designing for PHPLIB and Laravel, which would eventually also make its way into ODM.

docs/en/reference/aggregation-stage-reference.rst
jmikola: I see you re-requested a review, but I don't think you pushed any changes from my last review.

alcaeus: Oh, dang, I missed that. I've pushed changes now.

jmikola (approved):
Unrelated to this PR, but the CS and PHPStan errors suggest you may want to limit the language version used to the oldest supported.

alcaeus:
> Unrelated to this PR, but the CS and PHPStan errors suggest you may want to limit the language version used to the oldest supported.

Yep, I'll have to check out the various failures separately.

SenseException (approved):

Issue #2529, opened 2023-04-20
[2.0]XML schema validation seems too strict 


rupeshwankhede:
### BC Break Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.0.1
--

<br class="Apple-interchange-newline">

#### Summary

After updating version 2.0. Getting exception for below line. It is working on previous version.
XML Schema validation throw an exception about this line

` <field field-name="created" type="date" name="created">
        <gedmo:timestampable on="create"/>
  </field>
  `
 

#### Previous behavior

There wasn't any error about that line in previous versions. We just return.
` if (@$document->schemaValidate(__DIR__ . '/../../../../../../doctrine-mongo-mapping.xsd')) {
                return;
   }`

#### Current behavior

In current version throwing exception.
 `if (! $document->schemaValidate(__DIR__ . '/../../../../../../doctrine-mongo-mapping.xsd')) {                
                throw MappingException::xmlMappingFileInvalid($filename, $this->formatErrors(libxml_get_errors()));
 }`
I don't know if it's possible, but maybe the validation should be less strict same like in ORM. 

Refer: https://github.com/doctrine/mongodb-odm/issues/2095

malarzm:
Well, there's still no resolution in https://github.com/doctrine-extensions/DoctrineExtensions/issues/2055 for which we hoped. I believe the only tangible thing we can do is to let users opt-out from the XSD validation, but that's obviously a footgun.

rupeshwankhede:
Yeah. Can we do that if it is possible?

malarzm:
Sure, are you up for the implementation?

rupeshwankhede:
@malarzm  Can you check PR and let us know if we need to change anything.
https://github.com/doctrine/mongodb-odm/pull/2535/files

etshy:
I wonder if that could be fixed with a composer script that generate a custom xsd depending on the required packages ?

PR #2535, opened 2023-05-29
[Removed strict xml validation]

rupeshwankhede:
|      Q       |   A
|------------- | -----------
| Type         |improvement
| BC Break     | yes

#### Summary
Removed strict xml validation. Before due this not able to use external gedmo extension.
https://github.com/doctrine/mongodb-odm/issues/2529

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
@@ -681,7 +681,7 @@ class XmlDriver extends FileDriver
             libxml_clear_errors();
 
             if (! $document->schemaValidate(__DIR__ . '/../../../../../../doctrine-mongo-mapping.xsd')) {
-                throw MappingException::xmlMappingFileInvalid($filename, $this->formatErrors(libxml_get_errors()));
+                return;
             }
         } finally {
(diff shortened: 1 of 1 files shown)


lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
```diff
@@ -681,7 +681,7 @@ private function validateSchema(string $filename): void
             libxml_clear_errors();
 
             if (! $document->schemaValidate(__DIR__ . '/../../../../../../doctrine-mongo-mapping.xsd')) {
-                throw MappingException::xmlMappingFileInvalid($filename, $this->formatErrors(libxml_get_errors()));
+                return;
```
malarzm, line 684: This removes feature altogether instead of letting user opt-out of it. There should be a flag passed to `XmlDriver::__construct` that defaults to `true` so current behaviour is preserved.

Also version 2.2.x that you're targeting is no longer maintained. This technically is a feature so 2.6.x branch should be targeted although we may consider releasing as a bugfix version (depending on where we'll end with configurating opt-out).

malarzm (changes requested):

GromNaN:
Closing, fill free to re-open a PR using a configuration flag to disable XML validation if this is still required.

PR #2542, opened 2023-07-03
Dev timeseries

cyildirim:
|      Q       |   A
|------------- | -----------
| Type         |feature
| BC Break     | yes/no
| Fixed issues |

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
@@ -420,6 +420,34 @@ use const PHP_VERSION_ID;
      */
     public $collectionMax;
 
+        /**
+     * READ-ONLY: If the collection is fixed size, its size in bytes.
+     *
+     * @var int|null
+     */
+    public $collectionSize;
+
+    /**
+     * READ-ONLY: If the collection is fixed size, the maximum number of elements to store in the collection.
+     *
+     * @var int|null
+     */
+    public $collectionMax;
+
+    /**
+     * READ-ONLY: If the collection should be a timeseries
+     *
+     * @var array{timeField:string,metaField?:string,granularity?:string}|null
+     */
+    public $collectionTimeseries;
+
+    /**
+     * READ-ONLY: If the collection is timeseries, sets automatic removal for the collection.
+     *
+     * @var int|null
+     */
+    public $collectionExpireAfterSeconds;
+
     /**
      * READ-ONLY Describes how MongoDB clients route read operations to the members of a replica set.
      *
@@ -652,6 +680,13 @@ use const PHP_VERSION_ID;
      * @var bool
      */
     public $isMappedSuperclass = false;
+
+     /**
+     * READ-ONLY: Whether this class describes the mapping of Timeseries.
+     *
+     * @var bool
+     */
+    public $isTimeSeriesDocument = false;
 
     /**
      * READ-ONLY: Whether this class describes the mapping of a embedded document.
@@ -1387,7 +1422,7 @@ use const PHP_VERSION_ID;
      * Sets the collection this Document is mapped to.
      *
      * @param array|string $name
-     * @psalm-param array{name: string, capped?: bool, size?: int, max?: int}|string $name
+     * @psalm-param array{name: string, capped?: bool, size?: int, max?: int}|array{name: string, timeseries?: array{timeField:string,metaField?:string,granularity?:string}, expireAfterSeconds?: int}|string $name
      *
      * @throws InvalidArgumentException
      */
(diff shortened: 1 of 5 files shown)


lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -420,6 +420,34 @@
...
+     * @var int|null
+     */
+    public $collectionSize;
+
+    /**
+     * READ-ONLY: If the collection is fixed size, the maximum number of elements to store in the collection.
+     *
+     * @var int|null
+     */
+    public $collectionMax;
```
alcaeus, line 421: There already is a `$collectionMax` property a few lines above.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -65,7 +65,7 @@ public function __construct(protected DocumentManager $dm, protected ClassMetada
     public function ensureIndexes(?int $maxTimeMs = null, ?WriteConcern $writeConcern = null, bool $background = false): void
     {
         foreach ($this->metadataFactory->getAllMetadata() as $class) {
-            if ($class->isMappedSuperclass || $class->isEmbeddedDocument || $class->isQueryResultDocument || $class->isView()) {
+            if ($class->isMappedSuperclass || $class->isEmbeddedDocument || $class->isQueryResultDocument || $class->isView() || $class->isTimeSeriesDocument) {
```
alcaeus: MongoDB 6.0 adds support for secondary indexes on time series collections: https://www.mongodb.com/docs/manual/core/timeseries/timeseries-limitations/#time-series-secondary-indexes

With that in mind, I don't think this should be added.

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AnnotationDriver.php
```diff
@@ -179,6 +179,9 @@ public function loadMetadataForClass($className, \Doctrine\Persistence\Mapping\C
 
         if (isset($documentAnnot->collection)) {
             $metadata->setCollection($documentAnnot->collection);
+            if (is_array($documentAnnot->collection) && array_key_exists('timeseries', $documentAnnot->collection)) {
```
alcaeus, line 182: There should be a corresponding change in the XML metadata driver.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -420,6 +420,34 @@
      */
     public $collectionMax;
 
+        /**
+     * READ-ONLY: If the collection is fixed size, its size in bytes.
+     *
+     * @var int|null
+     */
+    public $collectionSize;
```
alcaeus: Public properties are a leftover from old PHP versions where method calls were associated with a significant performance penalty. Nowadays there is no problem to hide these values in private properties and expose them via getters if necessary.

cyildirim: I've removed `$collectionSize` here as there was already one. Should I convert to current ones to private ?

bradroc:
@cyildirim this is looking great. Thanks for your contribution 🙏

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
alcaeus: Sorry for the delay. Yes, any new properties that are introduced should be made private so we have fewer breaks when we eventually make everything private.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -339,7 +339,7 @@ public function updateValidators(?int $maxTimeMs = null, ?WriteConcern $writeCon
     public function updateDocumentValidator(string $documentName, ?int $maxTimeMs = null, ?WriteConcern $writeConcern = null): void
     {
         $class = $this->dm->getClassMetadata($documentName);
-        if ($class->isMappedSuperclass || $class->isEmbeddedDocument || $class->isQueryResultDocument || $class->isView() || $class->isFile) {
+        if ($class->isMappedSuperclass || $class->isEmbeddedDocument || $class->isQueryResultDocument || $class->isView() || $class->isFile || $class->isTimeSeriesDocument) {
             throw new InvalidArgumentException('Cannot update validators for files, views, mapped super classes, embedded documents or aggregation result documents.');
```
lstrojny, line 343: Message needs to be changed as well

lstrojny (commented):
Tests should be added for ClassMetadata and XmlDriver.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
cyildirim: Thanks for reviewing. I've fixed the tests for now and started to write some test for this bit.

alcaeus:
@cyildirim Your work was picked up by @PReimers recently and continued in #2597. He's also started writing tests, so maybe you two can get together and see what else is missing?

alcaeus:
Closing as #2687 has been merged, containing the feature.

PR #2554, opened 2023-10-18
Add sort operator to $search stage

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

This adds the missing `sort` operator to the builder for the `$search` aggregation pipeline stage.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Stage.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Stage.php
@@ -236,10 +236,8 @@ abstract class Stage
      * Limits the number of documents passed to the next stage in the pipeline.
      *
      * @see https://docs.mongodb.com/manual/reference/operator/aggregation/limit/
-     *
-     * @return Stage\Limit
      */
-    public function limit(int $limit)
+    public function limit(int $limit): self
     {
         return $this->builder->limit($limit);
     }
@@ -434,10 +432,8 @@ abstract class Stage
      *
      * @param array<string, int|string>|string $fieldName Field name or array of field/order pairs
      * @param int|string                       $order     Field order (if one field is specified)
-     *
-     * @return Stage\Sort
      */
-    public function sort($fieldName, $order = null)
+    public function sort($fieldName, $order = null): self
     {
         return $this->builder->sort($fieldName, $order);
     }
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Stage.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Stage.php
@@ -434,8 +434,10 @@ abstract class Stage
      *
      * @param array<string, int|string>|string $fieldName Field name or array of field/order pairs
      * @param int|string                       $order     Field order (if one field is specified)
+     *
+     * @return Stage\Sort
      */
-    public function sort($fieldName, $order = null): Stage\Sort
+    public function sort($fieldName, $order = null)
     {
         return $this->builder->sort($fieldName, $order);
     }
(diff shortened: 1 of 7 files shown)


lib/Doctrine/ODM/MongoDB/Aggregation/Stage.php
```diff
@@ -434,8 +434,10 @@ public function sortByCount(string $expression): Stage\SortByCount
      *
      * @param array<string, int|string>|string $fieldName Field name or array of field/order pairs
      * @param int|string                       $order     Field order (if one field is specified)
+     *
+     * @return Stage\Sort
```
malarzm: What's wrong with native return type?

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search.php
```diff
@@ -10,8 +10,15 @@
...
 use Doctrine\ODM\MongoDB\Aggregation\Stage\Search\SupportsAllSearchOperatorsTrait;
 
+use function in_array;
+use function is_array;
+use function is_string;
+use function strtolower;
+
 /**
  * @psalm-type CountType = 'lowerBound'|'total'
+ * @psalm-type SortMetaKeywords = 'searchScore'
```
malarzm, line 21: This looks odd, name suggests there can be multiple keywords yet this is a string. Is this coming from MongoDB itself?

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search.php
```diff
@@ -128,6 +143,27 @@ public function returnStoredSource(bool $returnStoredSource = true): static
         return $this;
     }
 
+    public function sort($fieldName, $order = null): static
```
malarzm, line 154: Please add PHPDoc, especially given multiple meanings of `$fieldName`. Or maybe there's a way to get rid of this overloading? Maybe we should encourage multiple calls to `sort` instead?

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/AbstractSearchOperator.php
```diff
@@ -35,6 +35,11 @@ public function returnStoredSource(bool $returnStoredSource): Search
         return $this->search->returnStoredSource($returnStoredSource);
     }
 
+    public function sort($fieldName, $order = null): Search
```
malarzm, line 52: PHPDoc is missing here too

malarzm (commented):
I don't really like the dynamic nature of `sort`'s arguments. Or maybe there's already a precedence in our API that I don't recall :)

lib/Doctrine/ODM/MongoDB/Aggregation/Stage.php
alcaeus: The `sort` method for a `Search` stage does not return a sort stage, but handles this internally. We've had this with other stages (e.g. `GeoNear::limit`), but I believe I can still use a native type of `Stage` and refine it using the `@return` annotation.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search.php
alcaeus: I used the same strategy we use in the `Sort` stage itself, where there are multiple meta sort options. In the `$search` aggregation pipeline stage, the only allowed meta sort is on the search score, hence the single keyword here. I can also change the naming to `SortMetaKeywordType` or something similar.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search.php
alcaeus: This all grew out of the query builder, which allows for precisely that:
```php
$builder
    ->sort('foo', 1)
    ->sort('bar', -1);

// Equivalent to the above
$builder
    ->sort(['foo' => 1, 'bar' => -1]);
```

I've kept this strategy when creating the egg builder, and now it continues in the `$search` stage.

In the new aggregation pipeline builder we're creating we're instead leveraging named arguments for this:
```php
Stage::sort(foo: 1, bar: -1)
```

Long story short, this is going to change once the MongoDB-supported builder is stable enough for us to use it underneath and deprecate ODM's builder :)

malarzm:
```
Stage::sort(foo: 1, bar: -1)
```

this looks nice :) looking forward to it!

malarzm (approved):

malarzm:
Thanks @alcaeus!

PR #2556, opened 2023-10-23
Require PHP 8.1

malarzm:
|      Q       |   A
|------------- | -----------
| Type         | task
| BC Break     | no
| Fixed issues | 

#### Summary
Continuing https://github.com/doctrine/mongodb-odm/pull/2546#discussion_r1353132009

> [PHP] 8.0 usage in 2.5.x is rather minimal: https://packagist.org/packages/doctrine/mongodb-odm/php-stats#2.5

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Mapping/MappingException.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/MappingException.php
@@ -57,11 +57,6 @@ final class MappingException extends BaseMappingException
         return new self(sprintf("No mapping found for field by DB name '%s' in class '%s'.", $dbFieldName, $className));
     }
 
-    public static function duplicateFieldMapping(string $document, string $fieldName): self
-    {
-        return new self(sprintf('Property "%s" in "%s" was already declared, but it must be declared only once', $fieldName, $document));
-    }
-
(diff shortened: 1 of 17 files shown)


lib/Doctrine/ODM/MongoDB/Mapping/MappingException.php
```diff
@@ -57,11 +57,6 @@ public static function mappingNotFoundByDbName(string $className, string $dbFiel
         return new self(sprintf("No mapping found for field by DB name '%s' in class '%s'.", $dbFieldName, $className));
     }
 
-    public static function duplicateFieldMapping(string $document, string $fieldName): self
```
malarzm, line 60: Technically these removals are BC breaks but come on :)

SenseException (approved):

lib/Doctrine/ODM/MongoDB/Mapping/MappingException.php
alcaeus: Are all those factories unused? o.O

malarzm: Yup. I was surprised a bit but figured they were just leftovers. Haven't investigated the reason tho

PR #2560, opened 2023-10-26
Last PHP 8.1 chores

malarzm:
I think these are the last leftovers saying we need to do something once we're rocking PHP 8.1 :)

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/PersistentCollection/CollWithPHP81Types.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/PersistentCollection/CollWithPHP81Types.php
@@ -14,16 +14,12 @@ use Doctrine\Common\Collections\Collection;
  */
 class CollWithPHP81Types extends ArrayCollection
 {
-    /**
-     * @param Collection<TKey, TElement>&ArrayCollection<TKey, TElement> $param
-     * @return Collection<TKey, TElement>&ArrayCollection<TKey, TElement>
-     */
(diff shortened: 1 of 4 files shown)


tests/Doctrine/ODM/MongoDB/Tests/PersistentCollection/CollWithPHP81Types.php
```diff
@@ -14,16 +14,12 @@
  */
 class CollWithPHP81Types extends ArrayCollection
 {
-    /**
-     * @param Collection<TKey, TElement>&ArrayCollection<TKey, TElement> $param
```
malarzm, line 18: PHPCS somehow was producing `[..]TElement> &ArrayCol[..]` (note the space) and I have no willpower to dig into why. Everyone's happy with no PHPDoc ;)

franmomu: Maybe prefixing with `@psalm-` like `@psalm-param` and `@psalm-return` would make PHPCS do not fail, anyway I don't know why static analysis is not complaining 🤔

franmomu (approved):

tests/Doctrine/ODM/MongoDB/Tests/PersistentCollection/CollWithPHP81Types.php
malarzm: I think it could be because they got better over time and now they can infer same thing from class' `@template-extends` and return types? :)

PR #2558, opened 2023-10-25
feat: Symfony 7 support

kbond:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | n/a

#### Summary

Wanted to get this started.

PR #2563, opened 2023-10-28
Support Symfony 7 by adding return types conditionally

franmomu:
Similar to https://github.com/doctrine/orm/pull/10919

In https://github.com/doctrine/mongodb-odm/pull/2558 we allowed Symfony 7, but we were not testing it.


malarzm (approved):

malarzm:
Ouch, I should have thought more with that previous PR 👀 Thanks @franmomu for catching this! 🥃

PR #2566, opened 2023-10-29
Bump workflows

franmomu:


.github/workflows/release-on-milestone-closed.yml
```diff
@@ -8,7 +8,7 @@ on:
 jobs:
   release:
     name: "Git tag, release & create merge-up PR"
-    uses: "doctrine/.github/.github/workflows/release-on-milestone-closed.yml@1.4.1"
+    uses: "doctrine/.github/.github/workflows/release-on-milestone-closed.yml@3.1.0"
```
malarzm, line 11: Anything we need to do bumping two majors?

franmomu: Apparently not:

2.0: https://github.com/doctrine/.github/releases/tag/2.0.0 was to set PHP 8.1 by default
3.0: https://github.com/doctrine/.github/releases/tag/3.0.0 was to set PHP 8.2 by default and run on ubuntu 22.04

Following ORM, [here updated to 2.0](https://github.com/doctrine/orm/pull/10020/files) and [here to 3.0](https://github.com/doctrine/orm/pull/10287)

Looking at this, we can use ubuntu 22.04 as well, I'll add those changes in this PR

malarzm: Thanks for checking!

malarzm (approved):

malarzm:
Thanks @franmomu!

alcaeus:
Thanks @franmomu! For the record, dependabot can automatically update GitHub Actions Workflows, and we might want to leverage this here as well. See https://github.com/mongodb/mongo-php-library/pull/1157 for an example :)

Issue #2561, opened 2023-10-27
Indices from sub classes always get re-created


buffcode:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.5.3

#### Summary

Having an inheritance and defining indices on a subclass (eg. `UniqueIndex`) will always re-create the subclasses' indices.

#### Current behavior

Having an inheritance and defining indices on a subclass (eg. `UniqueIndex`) will result in the subclass indices to be dropped on every run of `doctrine:mongodb:schema:update` and re-created afterwards.

Notice that the index will exist after the command, so the collection looks okay, but the indices are effectively dropped and recreated.
This is usually not such of a big problem but it put some our our databases down after dropping an index on a 0,5 TB collection and starting to recreate it for multiple hours.

#### How to reproduce

```php
<?php

use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;

/**
 * @ODM\Document()
 * @ODM\InheritanceType("SINGLE_COLLECTION")
 * @ODM\DiscriminatorField("metadataType")
 */
abstract class Metadata
{
    /**
     * @ODM\Id()
     */
    protected ?string $id = null;

    public function getId(): ?string
    {
        return $this->id;
    }
}
```

```php
<?php

use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;

/**
 * @ODM\Document()
 */
class PreAggregationMetadata extends Metadata
{
    /**
     * @ODM\UniqueIndex()
     * @ODM\Field(type="string")
     */
    private string $cacheKey;
}
```

For debugging purposes add the following code to `\Doctrine\ODM\MongoDB\SchemaManager::updateDocumentIndexes`, right before `$collection->dropIndex(...)`:
```php
dump('Dropping index ' . $mongoIndex['name'] . ' on ' . $documentName);
```

Run:
```bash
$ bin/console doctrine:mongodb:schema:update
Dropping index cacheKey on Metadata
Updated indexes for all classes
Updated validation for all classes
```

#### Expected behavior

No index drop

```bash
$ bin/console doctrine:mongodb:schema:update
Updated indexes for all classes
Updated validation for all classes
```

#### Workaround

All indices have to be defined on the root class.

malarzm:
[...]

PR #2573, opened 2023-11-05
fix: Don't process subclass indexes with SingleCollection inheritance

buffcode:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | https://github.com/doctrine/mongodb-odm/issues/2561

#### Summary

When using single collection inheritance, all sub classes are processed as well and indexes get dropped or re-created, depending on class order.

With this PR only the parent classes are processed and will ensure the indexes of all subclasses.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataFactory.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataFactory.php
@@ -139,7 +139,11 @@ final class ClassMetadataFactory extends AbstractClassMetadataFactory implements
             $class->setIdGeneratorType($parent->generatorType);
             $this->addInheritedFields($class, $parent);
             $this->addInheritedRelations($class, $parent);
-            $this->addInheritedIndexes($class, $parent);
+            if ($parent->isMappedSuperclass) {
+                // only inherit indexes if parent won't apply them itself
+                $this->addInheritedIndexes($class, $parent);
+            }
+
             $this->setInheritedShardKey($class, $parent);
             $class->setIdentifier($parent->identifier);
             $class->setVersioned($parent->isVersioned);
--- a/lib/Doctrine/ODM/MongoDB/SchemaManager.php
+++ b/lib/Doctrine/ODM/MongoDB/SchemaManager.php
@@ -92,6 +92,12 @@ final class SchemaManager
                 continue;
             }
 
+            if ($class->isInheritanceTypeSingleCollection() && count($class->parentClasses) > 0) {
+                // Skip document nodes that use the same collection as one of their parents.
+                // Indexes will be added by the parent document.
+                continue;
+            }
+
             $this->updateDocumentIndexes($class->name, $maxTimeMs, $writeConcern);
         }
     }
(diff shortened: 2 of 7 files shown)


buffcode:
@malarzm I am not too familiar with psalm and how to fix [this CI error](https://github.com/doctrine/mongodb-odm/actions/runs/6763941812/job/18381651244?pr=2573#step:6:25), as it appears to be in the baseline file already:

```
Error: tests/Doctrine/ODM/MongoDB/Tests/SchemaManagerTest.php:330:26: InternalClass: MongoDB\Model\IndexInfoIteratorIterator is internal to MongoDB but called from Doctrine\ODM\MongoDB\Tests\SchemaManagerTest (see https://psalm.dev/174)

Error: tests/Doctrine/ODM/MongoDB/Tests/SchemaManagerTest.php:330:26: InternalMethod: Constructor MongoDB\Model\IndexInfoIteratorIterator::__construct is internal to MongoDB, MongoDB, and MongoDB but called from Doctrine\ODM\MongoDB\Tests\SchemaManagerTest::testUpdateDocumentIndexesShouldCreateIndexesFromMappedSuperclass (see https://psalm.dev/175)
```

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataFactory.php
```diff
@@ -139,7 +139,11 @@ protected function doLoadMetadata($class, $parent, $rootEntityFound, array $nonS
             $class->setIdGeneratorType($parent->generatorType);
             $this->addInheritedFields($class, $parent);
             $this->addInheritedRelations($class, $parent);
-            $this->addInheritedIndexes($class, $parent);
+            if ($parent->isMappedSuperclass) {
+                // only inherit indexes if parent won't apply them itself
+                $this->addInheritedIndexes($class, $parent);
```
malarzm, line 144: We can't do such a thing as it changes produced `ClassMetadata` too much:, especially for a patch release. I'm not sure yet if the change would be OK in a minor patch anyway.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -172,56 +178,68 @@ private function doGetDocumentIndexes(string $documentName, array &$visited): ar
...
-        foreach ($class->fieldMappings as $fieldMapping) {
-            if (isset($fieldMapping['embedded'])) {
-                if (isset($fieldMapping['targetDocument'])) {
-                    $possibleEmbeds = [$fieldMapping['targetDocument']];
-                } elseif (isset($fieldMapping['discriminatorMap'])) {
-                    $possibleEmbeds = array_unique($fieldMapping['discriminatorMap']);
-                } else {
-                    continue;
-                }
+        if ($class->isInheritanceTypeSingleCollection()) {
```
malarzm, line 184: I'd add a failsafe throwing an `\InvalidArgumentException` when a document that has `$class->parentClasses` is passed here

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -172,56 +178,68 @@ private function doGetDocumentIndexes(string $documentName, array &$visited): ar
...
 
-                foreach ($possibleEmbeds as $embed) {
-                    if (isset($embeddedDocumentIndexes[$embed])) {
-                        $embeddedIndexes = $embeddedDocumentIndexes[$embed];
+        $indexes                 = [];
+        $embeddedDocumentIndexes = [];
+        foreach ($processClasses as $class) {
+            $visited[$class->name] = true;
+
+            $indexes = array_merge($indexes, $this->prepareIndexes($class));
```
malarzm, line 196: If I'm not mistaken we should be able to work around my first comment here by applying array_unique or calculating some kind of key for an index that's shared between classes

malarzm: This will be dirtier than what you're proposing but we're not changing behaviour that was established for years

malarzm (commented):
As for the psalm error please try updating baseline (`vendor/bin/psalm --threads=$(nproc) --php-version=8.1 --update-baseline --set-baseline=psalm-baseline.xml`) and we'll see what it produces. Most probably it was allowing single usages of offending code and now there will be multiple :)

PR #2575, opened 2023-11-06
fix: Do not drop/re-create 2dsphere indexes

### LINKED ISSUE #2574: 2dsphere indexes get re-created everytime
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.5.x

#### Summary

Indexes of type `2dsphere` are always re-created.

#### Current behavior

Because the server-side options contain a `2dsphereIndexVersion` a mismatch is caused when comparing index options.

#### How to reproduce

```php

/**
 * @ODM\Document
 * @ODM\Index(keys={"coordinatesWith2DIndex"="2d"})
 * @ODM\Index(keys={"coordinatesWithSphereIndex"="2dsphere"})
 */
class GeoIndexDocument
{
    /**
     * @ODM\Id
     *
     * @var string|null
     */
    public $id;

    /** @ODM\Field(type="hash") */
    public $coordinatesWith2DIndex;

    /** @ODM\Field(type="hash") */
    public $coordinatesWithSphereIndex;
}
``` 

For debugging purposes add the following code to \Doctrine\ODM\MongoDB\SchemaManager::updateDocumentIndexes, right before $collection->dropIndex(...):
```php
dump('Dropping index ' . $mongoIndex['name'] . ' on ' . $documentName);
```

Run:
```bash
$ bin/console doctrine:mongodb:schema:update
Dropping index coordinates_2dsphere on GeoIndexDocument
Updated indexes for all classes
Updated validation for all classes
```

#### Expected behavior

Do not drop/re-create 2dsphere indexes

buffcode:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | Fixes #2574

#### Summary

Indexes of tpye `2dsphere` are always re-created, because the server-side options contain a `2dsphereIndexVersion`, which causes a mismatch when comparing index options.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/SchemaManager.php
+++ b/lib/Doctrine/ODM/MongoDB/SchemaManager.php
@@ -52,6 +52,7 @@ final class SchemaManager
         'language_override',
         'textIndexVersion',
         'name',
+        '2dsphereIndexVersion',
     ];
 
     protected DocumentManager $dm;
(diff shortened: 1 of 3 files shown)


alcaeus (approved):
LGTM, thanks for fixing! 🎉

PR #2577, opened 2023-11-06
Validate that targetDocument can be resolved

### LINKED ISSUE #2576: Mapping validation does not validate targetDocument
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | maybe
| Version     | 2.5.x

#### Summary

Validity of `EmbedOne/EmbedMany/ReferenceOne/ReferenceMany` with `targetDocument` is not confirmed.

#### Current behavior

Running `doctrine:mongodb:mapping:info` or just warming up the `ClassMetadata` does not detect references within `EmbedOne/EmbedMany/ReferenceOne/ReferenceMany` to `targetDocument`s that cannot be resolved.

#### How to reproduce

```php
/** @ODM\Document */
class InvalidTargetDocumentTestDocument
{
    /**
     * @ODM\Id
     *
     * @var string|null
     */
    public $id;

    /**
     * @ODM\ReferenceOne(targetDocument=Doctrine\ODM\MongoDB\Tests\Functional\SomeInvalidClass::class)
     *
     * @var object|null
     */
    public $reference;
}
```

#### Expected behavior

`ClassMetadata` generation should fail or `doctrine:mongodb:mapping:info` should display an error.

buffcode:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | maybe
| Fixed issues | #2576 

#### Summary

Show an error when validating the `ClassMetadata`, if a `targetDocument` can neither be resolved to an existing interface or class.

```bash
$ bin/console doctrine:mongodb:mapping:info
[FAIL] InvalidTargetDocumentTestDocument
Target document class 'Doctrine\ODM\MongoDB\Tests\Functional\SomeInvalidClass' used in field 'reference' of class 'Doctrine\ODM\MongoDB\Tests\Functional\InvalidTargetDocumentTestDocument' does not exist.
```

BC breaks could occur, if current metadata contains undetected errors.

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/TargetDocumentTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/TargetDocumentTest.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Doctrine\ODM\MongoDB\Tests\Functional;
 
 use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
+use Doctrine\ODM\MongoDB\Mapping\MappingException;
 use Doctrine\ODM\MongoDB\Tests\BaseTestCase;
 use stdClass;
 
@@ -22,7 +23,13 @@ class TargetDocumentTest extends BaseTestCase
 
     public function testTargetDocumentIsResolvable(): void
     {
-        self::expectExceptionMessage("Target document class 'Doctrine\ODM\MongoDB\Tests\Functional\SomeInvalidClass' used in field 'reference' of class 'Doctrine\ODM\MongoDB\Tests\Functional\InvalidTargetDocumentTestDocument' does not exist.");
+        self::expectExceptionObject(
+            MappingException::invalidTargetDocument(
+                SomeInvalidClass::class,
+                InvalidTargetDocumentTestDocument::class,
+                'reference',
+            ),
+        );
 
         $test            = new InvalidTargetDocumentTestDocument();
         $test->reference = new stdClass();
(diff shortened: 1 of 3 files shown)


tests/Doctrine/ODM/MongoDB/Tests/Functional/TargetDocumentTest.php
```diff
@@ -18,6 +19,15 @@ public function testMappedSuperClassAsTargetDocument(): void
         $this->dm->persist($test->reference);
         $this->dm->flush();
     }
+
+    public function testTargetDocumentIsResolvable(): void
+    {
+        self::expectExceptionMessage("Target document class 'Doctrine\ODM\MongoDB\Tests\Functional\SomeInvalidClass' used in field 'reference' of class 'Doctrine\ODM\MongoDB\Tests\Functional\InvalidTargetDocumentTestDocument' does not exist.");
```
alcaeus: Pro tip: use `expectException` with exception factories to not be impacted by exception message changes:
```suggestion
        self::expectException(MappingException::invalidTargetDocument(SomeInvalidClass::class, InvalidTargetDocumentTestDocument::class, 'reference'));
```

Note that the code suggestion will most likely need a couple of `use` statements.

alcaeus (commented):
Great find. I have a small suggestion for checking the exception.

Did you check if invalid class names in a `discriminatorMap` are handled correctly?

tests/Doctrine/ODM/MongoDB/Tests/Functional/TargetDocumentTest.php
buffcode: Thanks, good suggestion!

buffcode:
> Did you check if invalid class names in a `discriminatorMap` are handled correctly?

The map is only checked on class level and throws `MappingException::invalidClassInDiscriminatorMap`. Field references aren't checked currently, I will add that as well.

alcaeus (approved):

alcaeus:
Thanks @buffcode!

PR #2578, opened 2023-11-06
Deprecate legacy commit options

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

The `fsync`, `safe`, and `w` options can be deprecated, as they are all handled by the `writeConcern` commit option.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Configuration.php
+++ b/lib/Doctrine/ODM/MongoDB/Configuration.php
@@ -32,6 +32,7 @@ use ProxyManager\GeneratorStrategy\FileWriterGeneratorStrategy;
 use Psr\Cache\CacheItemPoolInterface;
 use ReflectionClass;
 
+use function array_key_exists;
 use function interface_exists;
 use function trigger_deprecation;
 use function trim;
@@ -449,6 +450,19 @@ class Configuration
     /** @psalm-param CommitOptions $defaultCommitOptions */
     public function setDefaultCommitOptions(array $defaultCommitOptions): void
     {
+        foreach (UnitOfWork::DEPRECATED_WRITE_OPTIONS as $deprecatedOption) {
+            if (! array_key_exists($deprecatedOption, $defaultCommitOptions)) {
+                continue;
+            }
+
+            trigger_deprecation(
+                'doctrine/mongodb-odm',
+                '2.6',
+                'The "%s" commit option used in the configuration is deprecated.',
+                $deprecatedOption,
+            );
+        }
+
         $this->attributes['defaultCommitOptions'] = $defaultCommitOptions;
     }
 
--- a/lib/Doctrine/ODM/MongoDB/Configuration.php
+++ b/lib/Doctrine/ODM/MongoDB/Configuration.php
@@ -451,16 +451,14 @@ class Configuration
     public function setDefaultCommitOptions(array $defaultCommitOptions): void
     {
         foreach (UnitOfWork::DEPRECATED_WRITE_OPTIONS as $deprecatedOption) {
-            if (! array_key_exists($deprecatedOption, $defaultCommitOptions)) {
-                continue;
+            if (array_key_exists($deprecatedOption, $defaultCommitOptions)) {
+                trigger_deprecation(
+                    'doctrine/mongodb-odm',
+                    '2.6',
+                    'The "%s" commit option used in the configuration is deprecated.',
+                    $deprecatedOption,
+                );
             }
-
-            trigger_deprecation(
-                'doctrine/mongodb-odm',
-                '2.6',
-                'The "%s" commit option used in the configuration is deprecated.',
-                $deprecatedOption,
-            );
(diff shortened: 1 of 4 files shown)


alcaeus:
Note: static analysis errors are handled in #2554

lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -449,6 +450,19 @@ public function getDefaultCommitOptions(): array
     /** @psalm-param CommitOptions $defaultCommitOptions */
     public function setDefaultCommitOptions(array $defaultCommitOptions): void
     {
+        foreach (UnitOfWork::DEPRECATED_WRITE_OPTIONS as $deprecatedOption) {
+            if (! array_key_exists($deprecatedOption, $defaultCommitOptions)) {
```
jmikola: Would it make sense to flip this logic and avoid the `continue`? This isn't a matter of nesting conditionals, so I think we could go either way here.

Up to you.

lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -449,6 +450,19 @@ public function getDefaultCommitOptions(): array
...
+                continue;
+            }
+
+            trigger_deprecation(
```
jmikola: TIL [symfony/deprecation-contracts](https://github.com/symfony/deprecation-contracts) exists. Is this something we might want to use in PHPLIB down the line? Or do you think we should stick with the forthcoming logging API?

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -88,6 +90,9 @@ final class UnitOfWork implements PropertyChangedListener
      */
     public const STATE_REMOVED = 4;
 
+    /** @internal */
+    public const DEPRECATED_WRITE_OPTIONS = ['fsync', 'safe', 'w'];
```
jmikola, line 94: This makes me wonder how `w` interacts with an explicit `writeConcern` option, but that's beyond the scope of this deprecation PR.

jmikola (approved):
Optional suggestion to replace `continue` with the `trigger_deprecation` call but I defer to you. LGTM.

lib/Doctrine/ODM/MongoDB/Configuration.php
malarzm: FWIW there's also https://github.com/doctrine/deprecations but we never changed ODM (we were using Symfony's stuff before Doctrine's was introduced)

malarzm:
`writeConcern` is not allowed by Symfony's bundle (https://github.com/doctrine/DoctrineMongoDBBundle/blob/4.7.x/DependencyInjection/Configuration.php#L100-L111) so we'll need to make sure it is in next version :) And deprecate `w` there as well. No clue how Laminas' module is handling this.

lib/Doctrine/ODM/MongoDB/Configuration.php
alcaeus: @jmikola I honestly haven't considered that at all yet - but then on the other hand it's just a wrapper around `@trigger_error` with some formatting sugar.

@malarzm and I'm very much against changing it at this point ;)

lib/Doctrine/ODM/MongoDB/Configuration.php
alcaeus: It is possible with a small change to the phpcs configuration. The early exit is now no longer required if there's only a single if in the scope. Good judgment should still be applied :)

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
alcaeus: 😅

alcaeus:
> `writeConcern` is not allowed by Symfony's bundle (https://github.com/doctrine/DoctrineMongoDBBundle/blob/4.7.x/DependencyInjection/Configuration.php#L100-L111) so we'll need to make sure it is in next version :) And deprecate `w` there as well. No clue how Laminas' module is handling this.

I'll take care of the bundle separately - I need to think about how I can allow users to configure a WriteConcern object without having to create a service for it.

lib/Doctrine/ODM/MongoDB/Configuration.php
malarzm: I see no benefits either so no worries ;)

malarzm:
> I need to think about how I can allow users to configure a WriteConcern object without having to create a service for it.

I think you could try having something along:

```
parameters:
    # ...
    writeConcern:
        w: int
        # ...
```

and then in Configuration class remap `$parameters['writeConcern'] = new WriteConcern(...$parameters['writeConcern'])`. This would allow to easily add new options, should they appear in `WriteConcern` itself :)

PR #2580, opened 2023-11-08
Refactor commit logic

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

[PHPORM-112](https://jira.mongodb.org/browse/PHPORM-112)

(This PR currently targets 2.6.x, but will be held back until we work on 2.7.x)

As is, failures during a `commit` operation in `UnitOfWork` result in an inconsistent state. While updates and deletions mostly handle errors during the operation gracefully, errors during an insert or upset result in the UnitOfWork having a broken state.

This PR changes the timing of when we mark changes as done and remove changesets. All of these are now handled after all commit operations have concluded, which is a change made in preparation for transaction support (which requires us to keep changesets around until we successfully commit the transaction or give up). As a side effect, this will eventually allow fixing the inconsistencies mentioned above. A new test class keeps track of these inconsistencies so they can be addressed in the future.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
+++ b/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
@@ -162,42 +162,42 @@ final class UnitOfWork implements PropertyChangedListener
      *
      * @var array<string, object>
      */
-    private array $documentInsertions = [];
+    private array $scheduledDocumentInsertions = [];
 
     /**
      * A list of all pending document updates.
      *
      * @var array<string, object>
      */
-    private array $documentUpdates = [];
+    private array $scheduledDocumentUpdates = [];
 
     /**
      * A list of all pending document upserts.
      *
      * @var array<string, object>
      */
-    private array $documentUpserts = [];
+    private array $scheduledDocumentUpserts = [];
 
     /**
      * A list of all pending document deletions.
      *
      * @var array<string, object>
      */
-    private array $documentDeletions = [];
+    private array $scheduledDocumentDeletions = [];
 
     /**
      * All pending collection deletions.
      *
      * @psalm-var array<string, PersistentCollectionInterface<array-key, object>>
      */
-    private array $collectionDeletions = [];
+    private array $scheduledCollectionDeletions = [];
 
     /**
      * All pending collection updates.
      *
      * @psalm-var array<string, PersistentCollectionInterface<array-key, object>>
      */
-    private array $collectionUpdates = [];
+    private array $scheduledCollectionUpdates = [];
 
     /**
      * A list of documents related to collections scheduled for update or deletion
@@ -418,12 +418,12 @@ final class UnitOfWork implements PropertyChangedListener
         $this->computeChangeSets();
 
         if (
-            ! ($this->documentInsertions ||
-            $this->documentUpserts ||
-            $this->documentDeletions ||
-            $this->documentUpdates ||
-            $this->collectionUpdates ||
-            $this->collectionDeletions ||
+            ! ($this->scheduledDocumentInsertions ||
+            $this->scheduledDocumentUpserts ||
+            $this->scheduledDocumentDeletions ||
+            $this->scheduledDocumentUpdates ||
+            $this->scheduledCollectionUpdates ||
+            $this->scheduledCollectionDeletions ||
             $this->orphanRemovals)
         ) {
             return; // Nothing to do.
@@ -444,22 +444,22 @@ final class UnitOfWork implements PropertyChangedListener
             // Raise onFlush
             $this->evm->dispatchEvent(Events::onFlush, new Event\OnFlushEventArgs($this->dm));
 
-            foreach ($this->getClassesForCommitAction($this->documentUpserts) as $classAndDocuments) {
+            foreach ($this->getClassesForCommitAction($this->scheduledDocumentUpserts) as $classAndDocuments) {
                 [$class, $documents] = $classAndDocuments;
                 $this->executeUpserts($class, $documents, $options);
             }
 
-            foreach ($this->getClassesForCommitAction($this->documentInsertions) as $classAndDocuments) {
+            foreach ($this->getClassesForCommitAction($this->scheduledDocumentInsertions) as $classAndDocuments) {
                 [$class, $documents] = $classAndDocuments;
                 $this->executeInserts($class, $documents, $options);
             }
 
-            foreach ($this->getClassesForCommitAction($this->documentUpdates) as $classAndDocuments) {
+            foreach ($this->getClassesForCommitAction($this->scheduledDocumentUpdates) as $classAndDocuments) {
                 [$class, $documents] = $classAndDocuments;
                 $this->executeUpdates($class, $documents, $options);
             }
 
-            foreach ($this->getClassesForCommitAction($this->documentDeletions, true) as $classAndDocuments) {
+            foreach ($this->getClassesForCommitAction($this->scheduledDocumentDeletions, true) as $classAndDocuments) {
                 [$class, $documents] = $classAndDocuments;
                 $this->executeDeletions($class, $documents, $options);
             }
@@ -468,17 +468,17 @@ final class UnitOfWork implements PropertyChangedListener
             $this->evm->dispatchEvent(Events::postFlush, new Event\PostFlushEventArgs($this->dm));
 
             // Clear up
-            $this->documentInsertions          =
-            $this->documentUpserts             =
-            $this->documentUpdates             =
-            $this->documentDeletions           =
-            $this->documentChangeSets          =
-            $this->collectionUpdates           =
-            $this->collectionDeletions         =
-            $this->visitedCollections          =
-            $this->scheduledForSynchronization =
-            $this->orphanRemovals              =
-            $this->hasScheduledCollections     = [];
+            $this->scheduledDocumentInsertions  =
+            $this->scheduledDocumentUpserts     =
+            $this->scheduledDocumentUpdates     =
+            $this->scheduledDocumentDeletions   =
+            $this->documentChangeSets           =
+            $this->scheduledCollectionUpdates   =
+            $this->scheduledCollectionDeletions =
+            $this->visitedCollections           =
+            $this->scheduledForSynchronization  =
+            $this->orphanRemovals               =
+            $this->hasScheduledCollections      = [];
         } finally {
             $this->commitsInProgress--;
         }
@@ -537,7 +537,7 @@ final class UnitOfWork implements PropertyChangedListener
      */
     private function computeScheduleInsertsChangeSets(): void
     {
-        foreach ($this->documentInsertions as $document) {
+        foreach ($this->scheduledDocumentInsertions as $document) {
             $class = $this->dm->getClassMetadata($document::class);
             if ($class->isEmbeddedDocument || $class->isView()) {
                 continue;
@@ -554,7 +554,7 @@ final class UnitOfWork implements PropertyChangedListener
      */
     private function computeScheduleUpsertsChangeSets(): void
     {
-        foreach ($this->documentUpserts as $document) {
+        foreach ($this->scheduledDocumentUpserts as $document) {
             $class = $this->dm->getClassMetadata($document::class);
             if ($class->isEmbeddedDocument || $class->isView()) {
                 continue;
@@ -647,7 +647,7 @@ final class UnitOfWork implements PropertyChangedListener
      * entry is the new value of the property. Changesets are used by persisters
      * to INSERT/UPDATE the persistent document state.
      *
-     * {@link documentUpdates}
+     * {@link scheduledDocumentUpdates}
      * If the document is already fully MANAGED (has been fetched from the database before)
      * and any changes to its properties are detected, then a reference to the document is stored
      * there to mark it for an update.
@@ -930,9 +930,9 @@ final class UnitOfWork implements PropertyChangedListener
                 // Only MANAGED documents that are NOT SCHEDULED FOR INSERTION, UPSERT OR DELETION are processed here.
                 $oid = spl_object_hash($document);
                 if (
-                    isset($this->documentInsertions[$oid])
-                    || isset($this->documentUpserts[$oid])
-                    || isset($this->documentDeletions[$oid])
+                    isset($this->scheduledDocumentInsertions[$oid])
+                    || isset($this->scheduledDocumentUpserts[$oid])
+                    || isset($this->scheduledDocumentDeletions[$oid])
                     || ! isset($this->documentStates[$oid])
                 ) {
                     continue;
@@ -953,7 +953,7 @@ final class UnitOfWork implements PropertyChangedListener
      */
     private function computeAssociationChanges(object $parentDocument, array $assoc, $value): void
     {
-        $isNewParentDocument   = isset($this->documentInsertions[spl_object_hash($parentDocument)]);
+        $isNewParentDocument   = isset($this->scheduledDocumentInsertions[spl_object_hash($parentDocument)]);
         $class                 = $this->dm->getClassMetadata($parentDocument::class);
         $topOrExistingDocument = ( ! $isNewParentDocument || ! $class->isEmbeddedDocument);
 
@@ -1285,19 +1285,19 @@ final class UnitOfWork implements PropertyChangedListener
     {
         $oid = spl_object_hash($document);
 
-        if (isset($this->documentUpdates[$oid])) {
+        if (isset($this->scheduledDocumentUpdates[$oid])) {
             throw new InvalidArgumentException('Dirty document can not be scheduled for insertion.');
         }
 
-        if (isset($this->documentDeletions[$oid])) {
+        if (isset($this->scheduledDocumentDeletions[$oid])) {
             throw new InvalidArgumentException('Removed document can not be scheduled for insertion.');
         }
 
-        if (isset($this->documentInsertions[$oid])) {
+        if (isset($this->scheduledDocumentInsertions[$oid])) {
             throw new InvalidArgumentException('Document can not be scheduled for insertion twice.');
         }
 
-        $this->documentInsertions[$oid] = $document;
+        $this->scheduledDocumentInsertions[$oid] = $document;
 
         if (! isset($this->documentIdentifiers[$oid])) {
             return;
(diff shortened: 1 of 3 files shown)


lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -147,26 +138,6 @@ public function addInsert(object $document): void
         $this->queuedInserts[spl_object_hash($document)] = $document;
     }
 
-    /** @return array<string, object> */
-    public function getUpserts(): array
```
alcaeus: This class is marked as internal, so BC breaks shouldn't be an issue here. I don't think there's a benefit to keeping broken functionality around for the sake of backward compatibility.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -312,31 +254,16 @@ private function executeUpsert(object $document, array $options): void
          * NOOP otherwise. MongoDB 2.6+ does not allow empty modifiers, so $set
          * the identifier to the same value in our criteria.
          *
-         * This will fail for versions before MongoDB 2.6, which require an
-         * empty $set modifier. The best we can do (without attempting to check
-         * server versions in advance) is attempt the 2.6+ behavior and retry
-         * after the relevant exception.
-         *
```
alcaeus: We can safely remove this, as any supported driver version will refuse to connect to MongoDB 2.6 anyways. I don't see any benefits in keeping this around.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -312,31 +254,16 @@ private function executeUpsert(object $document, array $options): void
...
-        } catch (WriteException $e) {
-            if (empty($retry) || strpos($e->getMessage(), 'Mod on _id not allowed') === false) {
-                throw $e;
-            }
-        }
-
         assert($this->collection instanceof Collection);
-        $this->collection->updateOne($criteria, ['$set' => new stdClass()], $options);
+        $this->collection->updateOne($criteria, $data, $options);
+        $this->handleCollections($document, $options);
```
alcaeus: This was previously called in `executeUpserts`, so I moved it here. This is consistent with `update`, which also calls `handleCollections` itself.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -444,41 +481,23 @@ public function commit(array $options = []): void
...
             }
 
             // Raise postFlush
             $this->evm->dispatchEvent(Events::postFlush, new Event\PostFlushEventArgs($this->dm));
 
             // Clear up
-            $this->documentInsertions          =
-            $this->documentUpserts             =
-            $this->documentUpdates             =
-            $this->documentDeletions           =
```
alcaeus, line 474: These have been cleaned up in `cleanupExecutedChanges` and shouldn't be cleared entirely, as uncommitted changes may still be part of the UnitOfWork.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -1187,13 +1217,8 @@ private function executeUpserts(ClassMetadata $class, array $documents, array $o
...
         foreach ($documents as $oid => $document) {
-            $persister->addUpsert($document);
-            unset($this->documentUpserts[$oid]);
-        }
-
-        $persister->executeUpserts($options);
-
-        foreach ($documents as $document) {
+            $persister->upsert($document, $options);
+            $this->executedDocumentUpserts[] = $oid;
```
alcaeus: There is a subtle behavioural change here: while previously `postPersist` events were dispatched after all changes have been persisted, they are now dispatched after each individual change. This is consistent with `update` operation. I don't think it should post a big issue, but thought I'd point it out regardless.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -1247,31 +1271,7 @@ private function executeDeletions(ClassMetadata $class, array $documents, array
...
-                if (! ($value instanceof PersistentCollectionInterface)) {
-                    continue;
-                }
-
-                $value->clearSnapshot();
-            }
-
-            // Document with this $oid after deletion treated as NEW, even if the $oid
-            // is obtained by a new document because the old one went out of scope.
-            $this->documentStates[$oid] = self::STATE_NEW;
```
alcaeus, line 1273: Note: this has not been added in `cleanupExecutedChanges`, as unsetting the `documentState` works just as well. Any untracked entity is considered new unless otherwise indicated.

tests/Doctrine/ODM/MongoDB/Tests/UnitOfWorkCommitConsistencyTest.php
```diff
@@ -0,0 +1,197 @@
...
+declare(strict_types=1);
+
+namespace Doctrine\ODM\MongoDB\Tests;
+
+use Documents\ForumUser;
+use Documents\FriendUser;
+use MongoDB\BSON\ObjectId;
+use Throwable;
+
+class UnitOfWorkCommitConsistencyTest extends BaseTestCase
```
alcaeus, line 16: These are the tests I used to test original behaviour and make sure that refactoring didn't break anything. I then used the undesired behaviour (see first commit in this PR) as a guideline for fixing the persister class.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -162,42 +164,77 @@ final class UnitOfWork implements PropertyChangedListener
...
      */
-    private array $collectionDeletions = [];
+    private array $scheduledCollectionDeletions = [];
+
+    /**
+     * All pending collection deletions.
+     *
+     * @psalm-var array<string, PersistentCollectionInterface<array-key, object>>
+     */
+    private array $executedCollectionDeletions = [];
```
GromNaN: Listen to PHPStan which tells you that this property is useless.

alcaeus:
After a number of unsuccessful attempts at getting non-transactional flushes to behave somewhat normal, I realised that this requires a lot more changes to UnitOfWork, DocumentPersister, and CollectionPersister than I'm willing to make at this point. I've kept the tests around, but changed the assertions to match current broken behaviour (which changes with this PR for some of the tests, but then you can't break something that isn't broken already).

Scheduled document changes and changesets are now cleared at the end of the commit operation only if the operation itself was successful. In other cases, no changes were made and a subsequent flush may want to persist data again. I'll try to address this once transactional commit is implemented, as that will give me a rough idea of how the behaviour differs.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -1161,9 +1161,8 @@ private function executeInserts(ClassMetadata $class, array $documents, array $o
     {
         $persister = $this->getDocumentPersister($class->name);
 
-        foreach ($documents as $oid => $document) {
+        foreach ($documents as $document) {
             $persister->addInsert($document);
-            unset($this->documentInsertions[$oid]);
```
jmikola, line 1166: How does the removal of these `unset()` calls _during_ the execute methods accomplish the desired goal of clearing changes at the _end_ of a commit? Does this just rely on existing code in the commit step to wipe out all of the document HashTables (recently renamed with the "scheduled" prefix)?

tests/Doctrine/ODM/MongoDB/Tests/UnitOfWorkCommitConsistencyTest.php
```diff
@@ -98,17 +97,17 @@ public function testErrorDuringInsertKeepsFailingInsertionsForDocument(): void
...
-        // First user was inserted and changeset cleared
-        self::assertFalse($this->uow->isScheduledForInsert($firstUser));
+        // Wrong behaviour: user was saved and should no longer be scheduled for insertion
+        self::assertTrue($this->uow->isScheduledForInsert($firstUser));
         // Wrong behaviour: changeset should be empty
         self::assertNotEquals([], $this->uow->getDocumentChangeSet($firstUser));
 
         // Wrong behaviour: user should still be scheduled for insertion
-        self::assertFalse($this->uow->isScheduledForInsert($secondUser));
+        self::assertTrue($this->uow->isScheduledForInsert($secondUser));
```
jmikola, line 101: Why did the "Wrong behaviour" comment above pre-date the change from `assertFalse` to `assertTrue`?

tests/Doctrine/ODM/MongoDB/Tests/UnitOfWorkCommitConsistencyTest.php
```diff
@@ -465,8 +462,7 @@ public function testFailingDeleteWithEmbeddedDocumentKeepsChangeset(): void
         );
 
         self::assertTrue($this->uow->isScheduledForDelete($user));
-        // Wrong behaviour: embedded document should be scheduled for deletion
-        self::assertFalse($this->uow->isScheduledForDelete($address));
+        self::assertTrue($this->uow->isScheduledForDelete($address));
```
jmikola, line 410: Is this a legitimate bug fix or an unintended side effect? I assume it's related to having removed `$this->documentDeletions[$oid]` from `executeDeletes()`?

Same question applies to `isScheduledForInsert` and `isScheduledForUpsert` above, where you also removed "Wrong behaviour" comments.

tests/Doctrine/ODM/MongoDB/Tests/UnitOfWorkCommitConsistencyTest.php
```diff
@@ -0,0 +1,496 @@
...
+
+        $friendUser = new FriendUser('GromNaN');
+        $this->uow->persist($friendUser);
+
+        // Add failpoint to let the first insert command fail. This affects the ForumUser documents
+        $this->dm->getClient()->selectDatabase('admin')->command([
+            'configureFailPoint' => 'failCommand',
+            'mode' => ['times' => 1],
+            'data' => [
+                'errorCode' => 10107, // Not primary
```
jmikola: 10107 is a [retryable write error](https://github.com/mongodb/specifications/blob/master/source/retryable-writes/retryable-writes.rst#retryablewriteerror-labels). Do you actually need a second error to overcome the driver retrying by default, or are you explicitly disabling `retryWrites` when constructing the client?

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
alcaeus: > Does this just rely on existing code in the commit step to wipe out all of the document HashTables

Correct. The initial approach was to track executed changes and clear only those that were actually executed in case the commit process fails with an exception. However, while this works fine for root documents, the handling of embedded documents in UnitOfWork needs to be reworked more significantly to allow for correctly marking embedded documents as handled, and correctly dispatching commit events for embedded documents is also more tricky.

For that reason, I've decided to not try and fix this broken UnitOfWork behaviour when a commit operation fails, and instead keep document changesets around until the commit is completed so that it can work with transactions.

tests/Doctrine/ODM/MongoDB/Tests/UnitOfWorkCommitConsistencyTest.php
alcaeus: TL;DR: the comment is obsolete and was left in on error. I've removed it.

As for why the behaviour was (and still is) wrong, inserts and upserts are handled in batches: the operation is added to the persister, removed from the `scheduledDocumentInsertions` or `scheduledDocumentUpserts` hash table, then the persister executes all operations. This execution however is not all-or-nothing: for insertions, an `insertMany` is used, and for upserts each document is upserted separately using an `updateOne` operation. This meant that with the unique key violation in the second document the old behaviour was to mark all three documents as inserted (which was wrong), whereas now all three documents are still tracked as scheduled (which is still wrong).

tests/Doctrine/ODM/MongoDB/Tests/UnitOfWorkCommitConsistencyTest.php
alcaeus: That's correct - unsetting this at the end fixes this behaviour but may introduce other inconsistencies. Commits without a transaction are a mess to track in UnitOfWork.

tests/Doctrine/ODM/MongoDB/Tests/UnitOfWorkCommitConsistencyTest.php
alcaeus: Ah, the "benefit" of testing with a standalone server. I've changed the error code to [192 (FailPointEnabled)](https://github.com/mongodb/mongo/blob/26be5a539d9e02519bbc17e1df6e44f8c3abbbfe/src/mongo/base/error_codes.yml#L238) for all fail points in this test. This makes more sense as we just want the operation to fail without a specific reason or remedy.

tests/Doctrine/ODM/MongoDB/Tests/BaseTestCase.php
```diff
@@ -162,4 +169,43 @@ protected function requireMongoDB42(string $message): void
     {
         $this->requireVersion($this->getServerVersion(), '4.2.0', '<', $message);
     }
+
+    protected static function getUri(bool $useMultipleMongoses = true): string
+    {
+        $uri = getenv('DOCTRINE_MONGODB_SERVER') ?: DOCTRINE_MONGODB_SERVER;
+
+        return $useMultipleMongoses ? $uri : self::removeMultipleHosts($uri);
```
jmikola, line 177: Expect this to fail horribly if we ever test with an SRV URI pointing to a sharded cluster.

tests/Doctrine/ODM/MongoDB/Tests/UnitOfWorkCommitConsistencyTest.php
```diff
@@ -0,0 +1,501 @@
...
+        // The document still exists, the deletion is still scheduled
+        self::assertSame(
+            1,
+            $this->dm->getDocumentCollection(ForumUser::class)->countDocuments(['username' => 'alcaeus']),
+        );
+
+        self::assertTrue($this->uow->isScheduledForDelete($user));
+    }
+
+    public function testFailingDeleteWithEmbeddedDocumentKeepsChangeset(): void
```
jmikola: Since this is derivative of the previous test, I think the names could be more similar:

 * testDeleteFailureKeepsChangeset
 * testDeleteFailureWithEmbeddedDocumentKeepsChangeset
 
Something to that effect.

tests/Doctrine/ODM/MongoDB/Tests/UnitOfWorkCommitConsistencyTest.php
```diff
@@ -0,0 +1,501 @@
...
+        self::assertSame(
+            1,
+            $this->dm->getDocumentCollection(User::class)->countDocuments(['username' => 'alcaeus']),
+        );
+
+        self::assertTrue($this->uow->isScheduledForDelete($user));
+        self::assertTrue($this->uow->isScheduledForDelete($address));
+    }
+
+    public function testSuccessfulDeleteWithEmbeddedDocumentClearsChangeset(): void
```
jmikola: Related to my previous comment, this could be:

 * testDeleteWithEmbeddedDocumentClearsChangeset

I don't think "Success" adds much value (it can be assumed).

tests/Doctrine/ODM/MongoDB/Tests/UnitOfWorkCommitConsistencyTest.php
```diff
@@ -0,0 +1,501 @@
...
+
+        // Make sure delete command fails once
+        $this->dm->getClient()->selectDatabase('admin')->command([
+            'configureFailPoint' => 'failCommand',
+            'mode' => ['times' => 1],
+            'data' => [
+                'errorCode' => 192, // FailPointEnabled
+                'failCommands' => ['delete'],
+            ],
+        ]);
```
jmikola: Consider a private helper method that just takes a command name to remove all of this duplication and make the tests more readable.

tests/Doctrine/ODM/MongoDB/Tests/UnitOfWorkCommitConsistencyTest.php
```diff
@@ -0,0 +1,501 @@
...
+
+        self::assertTrue($this->uow->isScheduledForInsert($friendUser));
+        self::assertNotEquals([], $this->uow->getDocumentChangeSet($friendUser));
+    }
+
+    public function testErrorDuringInsertKeepsFailingInsertionsForDocument(): void
+    {
+        // Create a unique index on the collection to let the second insert fail
+        $collection = $this->dm->getDocumentCollection(ForumUser::class);
+        $collection->createIndex(['username' => 1], ['unique' => true]);
```
jmikola, line 70: Why bother with a unique index when `update` and `delete` use fail points?

jmikola (approved):
A few optional suggestions to improve the test file, but changes LGTM.

tests/Doctrine/ODM/MongoDB/Tests/BaseTestCase.php
alcaeus: Yep, I'm hoping it won't come to that. Please point me to this comment with an appropriate gif once the time comes.

tests/Doctrine/ODM/MongoDB/Tests/UnitOfWorkCommitConsistencyTest.php
alcaeus: The `commit` call in this test will run `insertMany` with three documents. A fail point would make the `insert` command fail before writing the first document; however I specifically wanted to test the behaviour when the first document can be inserted but the second can't; the only way we can test this is by using an index error on the second document. I've expanded the comment.

tests/Doctrine/ODM/MongoDB/Tests/UnitOfWorkCommitConsistencyTest.php
alcaeus: Good point.

tests/Doctrine/ODM/MongoDB/Tests/UnitOfWorkCommitConsistencyTest.php
alcaeus: I've renamed all tests to be of the form `test<Operation>(Error)?<Description>`.

PR #2586, opened 2023-11-24
Transaction support

### LINKED ISSUE #1888: Transaction support
### Feature Request

Introduce transaction support on database level as MongoDB starts to allow since version 4.0.

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes/no
| BC Break    | should not be a BC break

#### Summary

I'd like to ask for opinion and discuss future solutions regarding transactions in Doctrine ODM.

MongoDB documentation regarding transactions: https://docs.mongodb.com/manual/core/transactions/

There are a few relevant bits of information regarding this feature. Specifically:

> In most cases, multi-document transaction incurs a greater performance cost over single document writes, and the availability of multi-document transaction should not be a replacement for effective schema design. 

> Multi-document transactions are available for replica sets only. Transactions for sharded clusters are scheduled for MongoDB 4.2

What is Doctrine ODM desired approach on this? Introducing `beginTransaction()`, `transactional()` methods into `DocumentManager`? Since MongoDB transactions incur performance penalty, they probably will not be enabled on `flush` by default?

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | 

#### Summary

This pull request introduces support for transactions. When calling `DocumentManager::commit`, a transaction will be used depending on a configuration setting. Alternatively, transaction usage can also be changed using a commit option.

#### Usage

A new `useTransactionalFlush` configuration option can be enabled when running on the following deployments:
* Replica sets starting with MongoDB 4.0
* Sharded clusters starting with MongoDB 4.2
* Serverless deployments on MongoDB Atlas
* Loadbalanced deployments on MongoDB Atlas

To enable this feature globally, enable transactional flushes in the configuration:
```php
$config = new Configuration();
$config->setUseTransactionalFlush(true);
// Other configuration

$dm = DocumentManager::create(null, $config);
```

Alternatively, the `withTransaction` commit option can be specified to override this default value and selectively enable or disable transactions:
```php
// To explicitly enable transaction for this write
$dm->flush(['withTransaction' => true]);

// To disable transaction usage for a write, regardless of the ``useTransactionalFlush`` config:
$dm->flush(['withTransaction' => false]);
```

This pull request has been reviewed in smaller chunks in the following pull requests:
* #2580
* #2587
* #2589
* #2594
* #2604
* #2606

This PR is also accompanied by a [pull request in the Symfony Bundle](https://github.com/doctrine/DoctrineMongoDBBundle/pull/814) to expose the transactional flush setting in the bundle configuration.

### DIFF
--- a/.github/workflows/coding-standards.yml
+++ b/.github/workflows/coding-standards.yml
@@ -5,6 +5,7 @@ on:
   pull_request:
     branches:
       - "*.x"
+      - "feature/*"
   push:
 
 jobs:
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -4,6 +4,7 @@ on:
   pull_request:
     branches:
       - "*.x"
+      - "feature/*"
   push:
 
 env:
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -45,7 +45,7 @@ jobs:
           # Test with highest dependencies
           - topology: "server"
             php-version: "8.2"
-            mongodb-version: "6.0"
+            mongodb-version: "7.0"
             driver-version: "stable"
             dependencies: "highest"
             symfony-version: "7"
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -22,6 +22,7 @@ jobs:
           - "8.2"
           - "8.3"
         mongodb-version:
+          - "7.0"
           - "6.0"
           - "5.0"
           - "4.4"
@@ -34,24 +35,34 @@ jobs:
         symfony-version:
           - "stable"
         include:
+          # Test against lowest dependencies
           - dependencies: "lowest"
             php-version: "8.1"
             mongodb-version: "4.4"
             driver-version: "1.11.0"
             topology: "server"
             symfony-version: "stable"
-          - topology: "sharded_cluster"
-            php-version: "8.2"
-            mongodb-version: "4.4"
-            driver-version: "stable"
-            dependencies: "highest"
-            symfony-version: "stable"
+          # Test with highest dependencies
           - topology: "server"
             php-version: "8.2"
             mongodb-version: "6.0"
             driver-version: "stable"
             dependencies: "highest"
             symfony-version: "7"
+          # Test with a 4.4 replica set
+          - topology: "replica_set"
+            php-version: "8.2"
+            mongodb-version: "4.4"
+            driver-version: "stable"
+            dependencies: "highest"
+            symfony-version: "stable"
+          # Test with a 4.4 sharded cluster
+          - topology: "sharded_cluster"
+            php-version: "8.2"
+            mongodb-version: "4.4"
+            driver-version: "stable"
+            dependencies: "highest"
+            symfony-version: "stable"
 
     steps:
       - name: "Checkout"
--- a/lib/Doctrine/ODM/MongoDB/Configuration.php
+++ b/lib/Doctrine/ODM/MongoDB/Configuration.php
@@ -94,6 +94,7 @@ class Configuration
      *      defaultGridFSRepositoryClassName?: class-string<GridFSRepository<object>>,
      *      defaultDB?: string,
      *      documentNamespaces?: array<string, string>,
+     *      enableTransactionalFlush?: bool,
      *      filters?: array<string, array{
      *          class: class-string,
      *          parameters: array<string, mixed>
@@ -116,6 +117,8 @@ class Configuration
     private ProxyManagerConfiguration $proxyManagerConfiguration;
 
     private int $autoGenerateProxyClasses = self::AUTOGENERATE_EVAL;
+
+    private bool $useTransactionalFlush = false;
 
     public function __construct()
     {
@@ -595,6 +598,16 @@ class Configuration
     public function getProxyManagerConfiguration(): ProxyManagerConfiguration
     {
         return $this->proxyManagerConfiguration;
+    }
+
+    public function setUseTransactionalFlush(bool $useTransactionalFlush): void
+    {
+        $this->useTransactionalFlush = $useTransactionalFlush;
+    }
+
+    public function isTransactionalFlushEnabled(): bool
+    {
+        return $this->useTransactionalFlush;
     }
 }
 
--- a/lib/Doctrine/ODM/MongoDB/Configuration.php
+++ b/lib/Doctrine/ODM/MongoDB/Configuration.php
@@ -94,7 +94,6 @@ class Configuration
      *      defaultGridFSRepositoryClassName?: class-string<GridFSRepository<object>>,
      *      defaultDB?: string,
      *      documentNamespaces?: array<string, string>,
-     *      enableTransactionalFlush?: bool,
      *      filters?: array<string, array{
      *          class: class-string,
      *          parameters: array<string, mixed>
--- a/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
+++ b/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
@@ -3183,6 +3183,9 @@ final class UnitOfWork implements PropertyChangedListener
     /**
      * This following method was taken from the MongoDB Library and adapted to not use the default 120 seconds timeout.
      * The code within this method is licensed under the Apache License. Copyright belongs to MongoDB, Inc.
+     *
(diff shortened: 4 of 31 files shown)


.github/workflows/coding-standards.yml
```diff
@@ -5,6 +5,7 @@ on:
   pull_request:
     branches:
       - "*.x"
+      - "feature/*"
```
GromNaN, line 8: Do you still need to run CI on feature branches?

GromNaN (approved):

.github/workflows/coding-standards.yml
alcaeus: Not necessarily, but it also doesn't hurt having it for other larger topics (such as a potential refactoring of persisters or hydrators which may be coming).

.github/workflows/continuous-integration.yml
```diff
@@ -33,24 +35,34 @@ jobs:
...
             php-version: "8.1"
             mongodb-version: "4.4"
             driver-version: "1.11.0"
             topology: "server"
             symfony-version: "stable"
-          - topology: "sharded_cluster"
+          # Test with highest dependencies
+          - topology: "server"
+            php-version: "8.2"
+            mongodb-version: "6.0"
```
jmikola: Should this be bumped to 7.0, or does "highest dependencies" only apply to Composer deps?

lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -94,6 +94,7 @@ class Configuration
      *      defaultGridFSRepositoryClassName?: class-string<GridFSRepository<object>>,
      *      defaultDB?: string,
      *      documentNamespaces?: array<string, string>,
+     *      enableTransactionalFlush?: bool,
```
jmikola: Where is this configuration option actually referenced? I don't see any other mention of this symbol in the PR apart from a `testEnableTransactionalFlush()` method.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -3119,8 +3120,136 @@ public function isUninitializedObject(object $obj): bool
...
+                $options,
+            ),
+            self::TRANSACTION_OPTIONS,
+        );
+    }
+
+    /**
+     * This following method was taken from the MongoDB Library and adapted to not use the default 120 seconds timeout.
+     * The code within this method is licensed under the Apache License. Copyright belongs to MongoDB, Inc.
+     */
```
jmikola, line 3189:
```suggestion
     *
     * @see https://github.com/mongodb/mongo-php-library/blob/1.17.0/src/Operation/WithTransaction.php
     * @see https://github.com/mongodb/specifications/blob/master/source/transactions-convenient-api/transactions-convenient-api.rst#pseudo-code
     */
```

lib/Doctrine/ODM/MongoDB/Utility/LifecycleEventManager.php
```diff
@@ -128,15 +182,24 @@ public function preRemove(ClassMetadata $class, object $document): void
...
      * @template T of object
      */
-    public function preUpdate(ClassMetadata $class, object $document): void
+    public function preUpdate(ClassMetadata $class, object $document, ?Session $session = null): void
     {
+        if (! $this->shouldDispatchEvent($document, Events::preUpdate, $session)) {
+            return;
+        }
+
+        $eventArgs = new PreUpdateEventArgs($document, $this->dm, $this->uow->getDocumentChangeSet($document), $session);
```
jmikola: `$eventArgs` is only referenced within the subsequent conditional. Should you move this within its scope?

tests/Doctrine/ODM/MongoDB/Tests/BaseTestCase.php
```diff
@@ -162,4 +202,55 @@ protected function requireMongoDB42(string $message): void
...
+        $uri = getenv('DOCTRINE_MONGODB_SERVER') ?: DOCTRINE_MONGODB_SERVER;
+
+        return $useMultipleMongoses ? $uri : self::removeMultipleHosts($uri);
+    }
+
+    /**
+     * Removes any hosts beyond the first in a URI. This function should only be
+     * used with a sharded cluster URI, but that is not enforced.
+     */
+    protected static function removeMultipleHosts(string $uri): string
```
jmikola, line 217: This is technically also lifted from PHPLIB but I have no qualms about copying it here. :shushing_face:

tests/Doctrine/ODM/MongoDB/Tests/BaseTestCase.php
```diff
@@ -162,4 +202,55 @@ protected function requireMongoDB42(string $message): void
...
+    protected static function supportsTransactions(): bool
+    {
+        return self::$supportsTransactions ??= self::detectTransactionSupport();
+    }
+
+    private static function detectTransactionSupport(): bool
+    {
+        $manager = new Manager(self::getUri());
+
+        return $manager->selectServer()->getType() !== Server::TYPE_STANDALONE;
```
jmikola, line 254: Remind me if we discussed this before, but transactions require either a 4.0+ RS or 4.2+ sharded cluster. Since ODM only tests as low as 4.4, we don't need to concern ourselves with a version check.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
malarzm: Should this link to master? The file/anchor may be gone and we won't know

malarzm (approved):
I admit I've skimmed through as I tried to do all partial PRs, I did not catch more than @jmikola's eagle eye already did ;)

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
jmikola: I suppose the PHPLIB link can use https://github.com/mongodb/mongo-php-library/blob/1.17.0/src/Operation/WithTransaction.php (will edit the suggestion), but we don't have tags on the specs repo and tend to use master branch links for those throughout the drivers team.

tests/Doctrine/ODM/MongoDB/Tests/BaseTestCase.php
alcaeus: We didn't discuss this before, so good point bringing it up. We haven't discussed version support in ODM in a while, but I'd say we can ignore anything < 4.4.

lib/Doctrine/ODM/MongoDB/Configuration.php
alcaeus: Good catch. This isn't actually part of that array, so I've removed it.

.github/workflows/continuous-integration.yml
alcaeus: "highest" only applies on composer, but it would make sense to test with MongoDB 7.0. Updated.

tests/Doctrine/ODM/MongoDB/Tests/BaseTestCase.php
malarzm: We already do https://github.com/doctrine/mongodb-odm/blob/c9b7f6bdec37f32c2b91637adeee52949e7596e6/.github/workflows/continuous-integration.yml#L23-L26

alcaeus: I meant for the purpose of skipping other topologies. In the library, we check server versions for replica sets and sharded clusters to ensure we only test on supported platforms. This isn't necessary here.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -1268,11 +1273,7 @@ private function executeDeletions(ClassMetadata $class, array $documents, array
                 $value->clearSnapshot();
             }
 
-            // Document with this $oid after deletion treated as NEW, even if the $oid
-            // is obtained by a new document because the old one went out of scope.
-            $this->documentStates[$oid] = self::STATE_NEW;
```
jmikola, line 1273: Was just curious about this but traced the explanation back to https://github.com/doctrine/mongodb-odm/pull/2580/files#r1386613660

jmikola (approved):
LGTM, but note the Psalm failures.

alcaeus:
> note the Psalm failures.

Thanks for pointing those out. I didn't have those locally, until I realised that I was running on Psalm 5.17. 5.20 has these errors, but they're unrelated to this code so I've skipped them via the baseline.

PR #2587, opened 2023-11-24
Add configuration setting for transactional flush

alcaeus:
Adds a configuration setting to enable transactional flushes.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Configuration.php
+++ b/lib/Doctrine/ODM/MongoDB/Configuration.php
@@ -118,7 +118,7 @@ class Configuration
 
     private int $autoGenerateProxyClasses = self::AUTOGENERATE_EVAL;
 
-    private bool $enableTransactionalFlush = false;
+    private bool $useTransactionalFlush = false;
 
     public function __construct()
     {
@@ -600,19 +600,14 @@ class Configuration
         return $this->proxyManagerConfiguration;
     }
 
-    public function disableTransactionalFlush(): void
+    public function setUseTransactionalFlush(bool $useTransactionalFlush): void
     {
-        $this->enableTransactionalFlush = false;
-    }
-
-    public function enableTransactionalFlush(): void
-    {
-        $this->enableTransactionalFlush = true;
+        $this->useTransactionalFlush = $useTransactionalFlush;
     }
 
     public function isTransactionalFlushEnabled(): bool
     {
-        return $this->enableTransactionalFlush;
+        return $this->useTransactionalFlush;
     }
 }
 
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -596,6 +599,21 @@ public function getProxyManagerConfiguration(): ProxyManagerConfiguration
     {
         return $this->proxyManagerConfiguration;
     }
+
+    public function disableTransactionalFlush(): void
```
GromNaN: In DBAL, Boolean options are defined with a [regular setter](https://github.com/doctrine/dbal/blob/ef0576b9c5763bd99caf16229eb4d4b7fdc8a897/src/Configuration.php#L194-L220). Using enable/disable methods is smart but less practical when you have to set the option from an external Boolean:
```
$enableTransactionalFlush ? $configuration->enableTransactionalFlush() : $configuration->disableTransactionalFlush();
```

I'm in favor of a classic setter. Especially since there's no Boolean option in this class yet.

lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -94,6 +94,7 @@ class Configuration
      *      defaultGridFSRepositoryClassName?: class-string<GridFSRepository<object>>,
      *      defaultDB?: string,
      *      documentNamespaces?: array<string, string>,
+     *      enableTransactionalFlush?: bool,
```
GromNaN, line 97: Have you considered an option with more open values? Using constants for values.
It's always possible to come up with other flush modes in the future.

alcaeus: I'll quote the [Defy Augury](https://github.com/mongodb/specifications#defy-augury) section in the Driver's spec repository:
> While we have some idea of what the server will do in the future, don't design features with those expectations in mind. Design and implement based on what is expected in the next release.

Applied to this scenario, we shouldn't make the option more complicated because we may at some point need this.

lib/Doctrine/ODM/MongoDB/Configuration.php
alcaeus: I was contemplating whether to use semantic methods to enable or disable the value, or to use a simple setter. I'll change it to a setter to be consistent with other options.

lib/Doctrine/ODM/MongoDB/Configuration.php
GromNaN: Ok, that was to gauge whether you were considering other options in the future.

lib/Doctrine/ODM/MongoDB/Configuration.php
alcaeus: Updated 👍

GromNaN (approved):

PR #2589, opened 2023-11-24
Add logic for transactional commits

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | 

#### Summary

~This PR contains a commit from #2587, which will be removed when rebasing.~

After refactoring commit logic in #2580 to no longer clear changesets and scheduled document actions after executing them, this pull request allows for running all `UnitOfWork::commit` operations within a multi-document transaction. Note that the feature does not check whether the deployment supports transactions, but it always acts if the configuration setting is enabled.

Included in this PR is a test change that enables transactional commit for all tests if the deployment supports it. This effectively tests transactional commit logic when testing with replica sets (introduced in this PR) or sharded clusters (already present). Standalone deployments continue to test legacy logic.

The commit logic makes use of the [convenient transaction API](https://github.com/mongodb/specifications/blob/master/source/transactions-convenient-api/transactions-convenient-api.rst), which takes care of starting the transaction, committing it, and restarting the entire transaction if necessary. The convenient API retries transactions for up to 2 minutes when encountering transient errors, which may be undesirable. As the API does not allow changing this timeout by design, we have to decide whether we want to keep using the convenient API, or rather introduce our own logic for retrying transactions that uses a lower timeout. Note that the 2-minute limit has been set since the introduction of the convenient API, and has not been challenged since. This PR may be a reason to suggest changing the specification to allow for lower timeouts, as a 2-minute timeout makes no sense in a web-based environment where we'd rather show the user a quick error page instead of delaying a page response. Either way, this issue should not come up in practice unless you're dealing with highly contentious parallel workloads, which I'd argue is quite rare.

Documentation for this feature will be introduced in a separate PR.


jmikola:
> Note that the 2-minute limit has been set since the introduction of the convenient API, and has not been challenged since.

This is where client-side operations timeout comes into play. `timeoutMS` is intended to supersede the non-configurable, two-minute default for `withTransaction()`. See [CSOT: Convenient Transactions API](https://github.com/mongodb/specifications/blob/master/source/client-side-operations-timeout/client-side-operations-timeout.rst#convenient-transactions-api) for more details.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -1562,6 +1563,10 @@ private function getQueryForDocument(object $document): array
      */
     private function getWriteOptions(array $options = []): array
     {
+        if ($this->isInTransaction($options)) {
+            return $options;
```
jmikola: Realistically, is `writeConcern` the only option that would be inherited below? I know specifying per-operation RC/WC options is [not permitted](https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#writeconcern), but just want to make sure nothing else is being overlooked from `getDefaultCommitOptions()` that _may_ be permitted.

If so, this logic should be rewritten so that only the `writeConcern` gets ignored.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -3114,4 +3113,59 @@ private function objToStr(object $obj): string
...
+            [$class, $documents] = $classAndDocuments;
+            $this->executeDeletions($class, $documents, $options);
+        }
+    }
+
+    /** @psalm-param CommitOptions $options */
+    private function useTransaction(array $options): bool
+    {
+        if (isset($options['withTransaction'])) {
+            return $options['withTransaction'];
```
jmikola, line 3158: Should you cast here or is the type of `withTransaction` enforced elsewhere?

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -3114,4 +3113,59 @@ private function objToStr(object $obj): string
...
+
+    /** @psalm-param CommitOptions $options */
+    private function getTransactionOptions(array $options): array
+    {
+        return array_intersect_key(
+            array_merge(
+                $this->dm->getConfiguration()->getDefaultCommitOptions(),
+                $options,
+            ),
+            array_flip(self::TRANSACTION_OPTIONS),
```
jmikola: Does it make sense to use a list of strings in the constant if you need to flip them for each usage? Assuming the constant is internal, there should be no concern with storing it in the format you actually need.

tests/Doctrine/ODM/MongoDB/Tests/BaseTestCase.php
```diff
@@ -208,4 +243,16 @@ protected static function removeMultipleHosts(string $uri): string
...
+    protected static function supportsTransactions(): bool
+    {
+        return self::$supportsTransactions ??= self::detectTransactionSupport();
+    }
+
+    private static function detectTransactionSupport(): bool
+    {
+        $manager = new Manager(self::getUri());
+
+        return $manager->selectServer()->getType() !== Server::TYPE_STANDALONE;
```
jmikola, line 254: I assume you intentionally did not copy the logic from PHPLIB's [`skipIfTransactionsAreNotSupported()`](https://github.com/mongodb/mongo-php-library/blob/1.17.0/tests/FunctionalTestCase.php#L502) since ODM only tests with server version 4.4+.

tests/Doctrine/ODM/MongoDB/Tests/Functional/AtomicSetTest.php
```diff
@@ -25,6 +25,9 @@
  */
 class AtomicSetTest extends BaseTestCase
 {
+    // This test counts executed commands and thus doesn't work with transactions
+    protected static bool $allowsTransactions = false;
```
jmikola, line 29: Is this preferable to a method override? It's functionally equivalent and this is far more concise so no objections from my end.

tests/Doctrine/ODM/MongoDB/Tests/BaseTestCase.php
```diff
@@ -127,6 +134,34 @@ protected function getServerVersion(): string
         return $result['version'];
     }
 
+    protected function getPrimaryServer(): Server
+    {
+        return $this->dm->getClient()->getManager()->selectServer();
+    }
+
+    protected function skipTestIfNoTransactionSupport(): void
```
jmikola, line 142: Any reason not to defer to `supportsTransactions()` here?

tests/Doctrine/ODM/MongoDB/Tests/Functional/DocumentPersisterTest.php
```diff
@@ -648,13 +650,37 @@ public function testExecuteInsertsRespectsWriteConcern(string $class, $writeConc
         $this->dm->flush();
     }
 
+    /** @psalm-param class-string $class */
+    #[DataProvider('dataProviderTestWriteConcern')]
+    public function testExecuteInsertsDoesNotUseWriteConcernInTransaction(string $class): void
```
jmikola: Would "...OmitsWriteConcernOptionInTransaction" be a more fitting title?

tests/Doctrine/ODM/MongoDB/Tests/Functional/ShardKeyTest.php
```diff
@@ -48,13 +49,15 @@ public function testUpdateAfterSave(): void
...
         $o->title = 'test2';
+
+        $this->logger->clear();
         $this->dm->flush();
 
-        $queries   = $this->logger->getAll();
-        $lastQuery = end($queries);
-        self::assertSame('update', $lastQuery->getCommandName());
+        $queries     = $this->logger->getAll();
+        $updateQuery = array_shift($queries);
```
jmikola, line 57: How did the transaction implementation reveal a problem with `end()` in this test? I understand the issue would have been that the last query was never actually removed from the list.

tests/Doctrine/ODM/MongoDB/Tests/UnitOfWorkTest.php
```diff
@@ -548,6 +550,31 @@ public function testCommitsInProgressIsUpdatedOnException(): void
...
     }
+
+    public function testTransactionalCommitWithCustomWriteOptions(): void
+    {
+        $this->skipTestIfNoTransactionSupport();
+
+        // Force transaction config to be enabled
+        $this->dm->getConfiguration()->setUseTransactionalFlush(true);
+
+        $collection = $this->createMock(\MongoDB\Collection::class);
```
jmikola: Should this utilize a `use` statement instead of a FQCN here?

tests/Doctrine/ODM/MongoDB/Tests/UnitOfWorkTest.php
```diff
@@ -548,6 +550,31 @@ public function testCommitsInProgressIsUpdatedOnException(): void
 
         $this->fail('This should never be reached, an exception should have been thrown.');
     }
+
+    public function testTransactionalCommitWithCustomWriteOptions(): void
```
jmikola: The name suggests that you're testing whether a WC is handled, but the actual test ignores it -- presumably because it's the default WC. Would it make more sense to use WC "majority" here and assert that it's passed on?

tests/Doctrine/ODM/MongoDB/Tests/Functional/ShardKeyTest.php
alcaeus: The test assumed that the `update` command was the last in the list of logged commands. This of course is no longer true when using transactions, so I changed the logic to clear the list of logged commands before flushing the `DocumentManager` instance and check the first command. An alternative would be to look for the first `update`, but not sure if it's even necessary.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
alcaeus: The `CommitOptions` type declares `withTransaction` as a `bool`, so static analysis would error for people that use it. For other people, the return type practically serves as a type check, alerting them to the wrong type.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
alcaeus: Yes, that makes sense. The declaration now uses the option names as keys.

tests/Doctrine/ODM/MongoDB/Tests/BaseTestCase.php
alcaeus: That's correct. I don't believe ODM specifies a minimum MongoDB version, and we'd probably do good calling out that we support MongoDB 4.4+ from this point onwards.

tests/Doctrine/ODM/MongoDB/Tests/BaseTestCase.php
alcaeus: No good reason, other than me not thinking straight. Updated.

tests/Doctrine/ODM/MongoDB/Tests/UnitOfWorkTest.php
alcaeus: Yes - since the `Collection` interface from doctrine/collections was imported, PhpStorm reverted to using a FQCN. I've replaced it with an alias.

tests/Doctrine/ODM/MongoDB/Tests/UnitOfWorkTest.php
alcaeus: This test tests that a write concern is not passed to the individual operation. I've renamed the test and added a separate test that verifies that the `commitTransaction` command uses the write concern specified as part of the commit options.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
alcaeus: Rethinking this, `writeConcern` is not the only one. Technically speaking, `getDefaultCommitOptions` may return a host of options that we can't control. Given that, I think it makes most sense to strip transaction options in `DocumentPersister` as well, after they've been assembled from default commit options, document options, and the specific commit options. This not only ensures that we never pass forbidden options to a write operation within a transaction, but it also future-proofs this code in case other options are eventually added to default commit options or the document commit options.

tests/Doctrine/ODM/MongoDB/Tests/Functional/AtomicSetTest.php
alcaeus: I figured it was a more concise way than overriding the `getConfiguration` method in every test class that does not support transactions.

tests/Doctrine/ODM/MongoDB/Tests/UnitOfWorkTest.php
```diff
@@ -575,6 +578,32 @@ public function testTransactionalCommitWithCustomWriteOptions(): void
...
+        $this->dm->getConfiguration()->setUseTransactionalFlush(true);
+
+        $user           = new ForumUser();
+        $user->username = '12345';
+        $this->uow->persist($user);
+
+        $logger = new CommandLogger();
+        $logger->register();
+
+        $this->uow->commit(['writeConcern' => new WriteConcern(2)]);
```
jmikola: "majority" may be safer as that would allow the test to run against a single-member replica set.

jmikola (approved):
Suggestion to use "majority" over "2" for the WC test, but LGTM otherwise.

alcaeus:
phpstan errors are unrelated and are fixed separately in https://github.com/doctrine/mongodb-odm/pull/2593.

PR #2593, opened 2023-11-29
Fix phpstan errors

alcaeus:
|      Q       |   A
|------------- | -----------
| BC Break     | yes/no
| Fixed issues | 

#### Summary

This fixes one error where the `enumType` mapping field gets rewritten to the `class-string<UnitEnum>` type because we don't ensure that the enum is backed before re-assigning `enumType`. The other issue in `mapField` is less obvious and I can't figure out where it takes the `class-string<UnitEnum>` type from, so I decided to ignore that for the time being. These issues started appearing with the latest phpstan release.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
@@ -51,8 +51,6 @@ use function strtolower;
 use function strtoupper;
 use function trigger_deprecation;
 
-use const PHP_VERSION_ID;
-
 /**
  * A <tt>ClassMetadata</tt> instance holds all the object-document mapping metadata
  * of a document and it's references.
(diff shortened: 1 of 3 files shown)


franmomu:
I've changed to use `ReflectionEnum::isBacked()` call, this should eventually narrow the type of `ReflectionEnum::getType()` to `class-string<BackedEnum>` I guess, so I think we can ignore these errors for now.

[I'll try to fix it in phpstan](https://github.com/phpstan/phpstan-src/pull/2806) following https://github.com/phpstan/phpstan/issues/10192#issuecomment-1827743195

malarzm (approved):
~~I have no clue what's going on with ignored errors either :/~~

Ah, you fixed the bug in PHPStan, cool!

malarzm:
Thanks @alcaeus and @franmomu!

alcaeus:
Thanks @franmomu for finishing this! 🎉

PR #2594, opened 2023-11-29
Update events to play nice with transactions

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

With transactional commit enabled, there's a chance that lifecycle events get dispatched more than once. This can be a problem for existing code that may assume that lifecycle events only get dispatched once, especially since this is also the behaviour when using transactions in ORM. Since a failing transaction may lead to retrying the entire commit operation, we need to make sure to only invoke lifecycle events once.

This is done by adding a transactional mode to the `LifecycleEventManager` class, which then tracks which events have been dispatched for each document. The `preUpdate`, `postPersist`, `postUpdate`, and `postRemove` events are then skipped if they have previously been dispatched for this document.

In addition to the above, the `LifecycleEventArgs` is now capable of telling event handlers whether the operation is part of a transaction. This and the `session` property in the event is necessary for any lifecycle event handlers that want to access the database. Since any changes already made to the database are only visible within the transaction, event handlers that want to read or write data need to pass the session option to ensure the write happens within the same transaction.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Event/LifecycleEventArgs.php
+++ b/lib/Doctrine/ODM/MongoDB/Event/LifecycleEventArgs.php
@@ -6,6 +6,8 @@ namespace Doctrine\ODM\MongoDB\Event;
 
 use Doctrine\ODM\MongoDB\DocumentManager;
 use Doctrine\Persistence\Event\LifecycleEventArgs as BaseLifecycleEventArgs;
+use Doctrine\Persistence\ObjectManager;
+use MongoDB\Driver\Session;
 
 /**
  * Lifecycle Events are triggered by the UnitOfWork during lifecycle transitions
@@ -15,6 +17,15 @@ use Doctrine\Persistence\Event\LifecycleEventArgs as BaseLifecycleEventArgs;
  */
 class LifecycleEventArgs extends BaseLifecycleEventArgs
 {
+    public function __construct(
+        object $object,
+        ObjectManager $objectManager,
+        public readonly bool $isInTransaction = false,
+        public readonly ?Session $session = null,
+    ) {
+        parent::__construct($object, $objectManager);
+    }
+
     public function getDocument(): object
     {
         return $this->getObject();
--- a/lib/Doctrine/ODM/MongoDB/Event/LifecycleEventArgs.php
+++ b/lib/Doctrine/ODM/MongoDB/Event/LifecycleEventArgs.php
@@ -20,7 +20,6 @@ class LifecycleEventArgs extends BaseLifecycleEventArgs
     public function __construct(
         object $object,
         ObjectManager $objectManager,
-        public readonly bool $isInTransaction = false,
         public readonly ?Session $session = null,
     ) {
         parent::__construct($object, $objectManager);
@@ -34,5 +33,10 @@ class LifecycleEventArgs extends BaseLifecycleEventArgs
     public function getDocumentManager(): DocumentManager
     {
         return $this->getObjectManager();
+    }
+
+    public function isInTransaction(): bool
+    {
+        return $this->session?->isInTransaction() ?? false;
     }
 }
--- a/lib/Doctrine/ODM/MongoDB/Event/PreLoadEventArgs.php
+++ b/lib/Doctrine/ODM/MongoDB/Event/PreLoadEventArgs.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Doctrine\ODM\MongoDB\Event;
 
 use Doctrine\ODM\MongoDB\DocumentManager;
+use MongoDB\Driver\Session;
 
 /**
  * Class that holds event arguments for a preLoad event.
@@ -15,9 +16,9 @@ final class PreLoadEventArgs extends LifecycleEventArgs
     private array $data;
 
     /** @param array<string, mixed> $data */
-    public function __construct(object $document, DocumentManager $dm, array &$data)
+    public function __construct(object $document, DocumentManager $dm, array &$data, bool $isInTransaction = false, ?Session $session = null)
     {
-        parent::__construct($document, $dm);
+        parent::__construct($document, $dm, $isInTransaction, $session);
 
         $this->data =& $data;
     }
--- a/lib/Doctrine/ODM/MongoDB/Event/PreLoadEventArgs.php
+++ b/lib/Doctrine/ODM/MongoDB/Event/PreLoadEventArgs.php
@@ -12,19 +12,14 @@ use MongoDB\Driver\Session;
  */
 final class PreLoadEventArgs extends LifecycleEventArgs
 {
-    /** @var array<string, mixed> */
-    private array $data;
-
     /** @param array<string, mixed> $data */
     public function __construct(
         object $document,
         DocumentManager $dm,
-        array &$data,
+        private array &$data,
         ?Session $session = null,
     ) {
         parent::__construct($document, $dm, $session);
-
-        $this->data =& $data;
     }
 
     /**
--- a/lib/Doctrine/ODM/MongoDB/Event/PreLoadEventArgs.php
+++ b/lib/Doctrine/ODM/MongoDB/Event/PreLoadEventArgs.php
@@ -16,8 +16,12 @@ final class PreLoadEventArgs extends LifecycleEventArgs
     private array $data;
 
     /** @param array<string, mixed> $data */
-    public function __construct(object $document, DocumentManager $dm, array &$data, ?Session $session = null)
-    {
+    public function __construct(
+        object $document,
+        DocumentManager $dm,
+        array &$data,
+        ?Session $session = null,
+    ) {
         parent::__construct($document, $dm, $session);
 
         $this->data =& $data;
--- a/lib/Doctrine/ODM/MongoDB/Event/PreLoadEventArgs.php
+++ b/lib/Doctrine/ODM/MongoDB/Event/PreLoadEventArgs.php
@@ -16,9 +16,9 @@ final class PreLoadEventArgs extends LifecycleEventArgs
     private array $data;
 
     /** @param array<string, mixed> $data */
-    public function __construct(object $document, DocumentManager $dm, array &$data, bool $isInTransaction = false, ?Session $session = null)
+    public function __construct(object $document, DocumentManager $dm, array &$data, ?Session $session = null)
     {
-        parent::__construct($document, $dm, $isInTransaction, $session);
+        parent::__construct($document, $dm, $session);
 
         $this->data =& $data;
     }
--- a/lib/Doctrine/ODM/MongoDB/Event/PreUpdateEventArgs.php
+++ b/lib/Doctrine/ODM/MongoDB/Event/PreUpdateEventArgs.php
@@ -7,6 +7,7 @@ namespace Doctrine\ODM\MongoDB\Event;
 use Doctrine\ODM\MongoDB\DocumentManager;
 use Doctrine\ODM\MongoDB\UnitOfWork;
 use InvalidArgumentException;
+use MongoDB\Driver\Session;
 
 use function get_class;
 use function sprintf;
@@ -22,9 +23,9 @@ final class PreUpdateEventArgs extends LifecycleEventArgs
     private array $documentChangeSet;
 
     /** @psalm-param array<string, ChangeSet> $changeSet */
-    public function __construct(object $document, DocumentManager $dm, array $changeSet)
+    public function __construct(object $document, DocumentManager $dm, array $changeSet, bool $isInTransaction = false, ?Session $session = null)
     {
-        parent::__construct($document, $dm);
+        parent::__construct($document, $dm, $isInTransaction, $session);
 
         $this->documentChangeSet = $changeSet;
     }
--- a/lib/Doctrine/ODM/MongoDB/Event/PreUpdateEventArgs.php
+++ b/lib/Doctrine/ODM/MongoDB/Event/PreUpdateEventArgs.php
@@ -19,30 +19,27 @@ use function sprintf;
  */
 final class PreUpdateEventArgs extends LifecycleEventArgs
 {
-    /** @psalm-var array<string, ChangeSet> */
-    private array $documentChangeSet;
-
     /** @psalm-param array<string, ChangeSet> $changeSet */
     public function __construct(
         object $document,
         DocumentManager $dm,
-        array $changeSet,
+        private array $changeSet,
         ?Session $session = null,
     ) {
         parent::__construct($document, $dm, $session);
 
-        $this->documentChangeSet = $changeSet;
+        $this->changeSet = $changeSet;
     }
 
     /** @return array<string, ChangeSet> */
     public function getDocumentChangeSet(): array
     {
-        return $this->documentChangeSet;
+        return $this->changeSet;
     }
 
     public function hasChangedField(string $field): bool
     {
-        return isset($this->documentChangeSet[$field]);
+        return isset($this->changeSet[$field]);
     }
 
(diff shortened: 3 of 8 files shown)


lib/Doctrine/ODM/MongoDB/Event/LifecycleEventArgs.php
```diff
@@ -15,6 +17,15 @@
  */
 class LifecycleEventArgs extends BaseLifecycleEventArgs
 {
+    public function __construct(
+        object $object,
+        ObjectManager $objectManager,
+        public readonly bool $isInTransaction = false,
+        public readonly ?Session $session = null,
```
jmikola: Smooth.

lib/Doctrine/ODM/MongoDB/Event/PreLoadEventArgs.php
```diff
@@ -15,9 +16,9 @@ final class PreLoadEventArgs extends LifecycleEventArgs
     private array $data;
 
     /** @param array<string, mixed> $data */
-    public function __construct(object $document, DocumentManager $dm, array &$data)
+    public function __construct(object $document, DocumentManager $dm, array &$data, bool $isInTransaction = false, ?Session $session = null)
```
jmikola: Do you want to break this out across multiple lines like you did in LifecycleEventArgs? Likewise for PreUpdateEventArgs below.

Also, can `array &$data` be refactored do use inline declarations, or is the by-reference attribute an issue?

lib/Doctrine/ODM/MongoDB/Event/PreUpdateEventArgs.php
```diff
@@ -22,9 +23,9 @@ final class PreUpdateEventArgs extends LifecycleEventArgs
     private array $documentChangeSet;
 
     /** @psalm-param array<string, ChangeSet> $changeSet */
-    public function __construct(object $document, DocumentManager $dm, array $changeSet)
+    public function __construct(object $document, DocumentManager $dm, array $changeSet, bool $isInTransaction = false, ?Session $session = null)
```
jmikola: This can also be split on multiple lines. There's also a refactoring opportunity to use an inline declaration for `array $documentChangeSet` and remove the assignment below (unless renaming the parameter would be problematic).

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -1218,13 +1221,13 @@ private function executeUpdates(ClassMetadata $class, array $documents, array $o
...
         foreach ($documents as $oid => $document) {
-            $this->lifecycleEventManager->preUpdate($class, $document);
+            $this->lifecycleEventManager->preUpdate($class, $document, $options['session'] ?? null);
 
             if (! empty($this->documentChangeSets[$oid]) || $this->hasScheduledCollections($document)) {
                 $persister->update($document, $options);
             }
 
-            $this->lifecycleEventManager->postUpdate($class, $document);
+            $this->lifecycleEventManager->postUpdate($class, $document, $options['session'] ?? null);
```
jmikola, line 1232: Are the InvalidArgument errors above relevant?

lib/Doctrine/ODM/MongoDB/Utility/LifecycleEventManager.php
```diff
@@ -55,11 +74,18 @@ public function postCollectionLoad(PersistentCollectionInterface $coll): void
...
-        $this->dispatchEvent($class, Events::postPersist, new LifecycleEventArgs($document, $this->dm));
-        $this->cascadePostPersist($class, $document);
+        if (! $this->shouldDispatchEvent($document, Events::postPersist)) {
+            return;
+        }
+
+        $isInTransaction = $session ? $session->isInTransaction() : false;
+        $eventArgs       = new LifecycleEventArgs($document, $this->dm, $isInTransaction, $session);
+
+        $class->invokeLifecycleCallbacks(Events::postPersist, $document, [$eventArgs]);
```
jmikola, line 90: I realize this API pre-dates the PR, but it seems odd to me that LifecycleEventArgs needs to be wrapped in an array when calling `invokeLifecycleCallbacks`. Is there ever a case where this third array parameter takes multiple values, or is everything in LifecycleEventArgs?

lib/Doctrine/ODM/MongoDB/Utility/LifecycleEventManager.php
```diff
@@ -128,15 +176,21 @@ public function preRemove(ClassMetadata $class, object $document): void
...
     {
+        if (! $this->shouldDispatchEvent($document, Events::preUpdate)) {
+            return;
+        }
+
+        $isInTransaction = $session ? $session->isInTransaction() : false;
+
         if (! empty($class->lifecycleCallbacks[Events::preUpdate])) {
-            $class->invokeLifecycleCallbacks(Events::preUpdate, $document, [new PreUpdateEventArgs($document, $this->dm, $this->uow->getDocumentChangeSet($document))]);
+            $class->invokeLifecycleCallbacks(Events::preUpdate, $document, [new PreUpdateEventArgs($document, $this->dm, $this->uow->getDocumentChangeSet($document), $isInTransaction, $session)]);
```
jmikola: Any reason you chose to construct `$eventArgs` in advance in earlier methods but not here? It would probably make these `invokeLifecycleCallbacks` lines more readable.

I understand the PreUpdateEventArgs are different, since the one below is constructed _after_ calling `recomputeSingleDocumentChangeSet`.

lib/Doctrine/ODM/MongoDB/Utility/LifecycleEventManager.php
```diff
@@ -232,4 +295,24 @@ private function dispatchEvent(ClassMetadata $class, string $eventName, ?EventAr
...
+        // Check whether the event has already been dispatched.
+        $hasDispatched = isset($this->transactionalEvents[$this->getObjectId($document)][$eventName]);
+
+        // Mark the event as dispatched - no problem doing this if it already was dispatched
+        $this->transactionalEvents[$this->getObjectId($document)][$eventName] = true;
+
+        return ! $hasDispatched;
+    }
+
+    private function getObjectId(object $document): string
```
jmikola: Is this method worth avoiding two inline calls to `spl_object_hash` in `shouldDispatchEvent`?

tests/Doctrine/ODM/MongoDB/Tests/Events/TransactionalLifecycleEventsTest.php
```diff
@@ -0,0 +1,264 @@
...
+    public function postRemove(Event\LifecycleEventArgs $e): void
+    {
+        $this->assertTransactionState($e);
+        $this->postRemove++;
+    }
+
+    private function assertTransactionState(LifecycleEventArgs $e): void
+    {
+        $this->test->assertTrue($e->isInTransaction);
+        $this->test->assertInstanceOf(Session::class, $e->session);
```
jmikola: I don't think there's any need to inject the TestCase into these models. You can just invoke the PHPUnit assertions statically or via their function counterparts. We do this in PHPLIB's test suite (e.g. Unified Test Runner).

lib/Doctrine/ODM/MongoDB/Event/LifecycleEventArgs.php
malarzm: Quick n00b question: why do we need bool flag while there's `$session->isInTransaction()` present? At a glance it seems we could add an isser `return $session?->isInInTransaction() ?? false` instead of a property?

lib/Doctrine/ODM/MongoDB/Utility/LifecycleEventManager.php
```diff
@@ -15,14 +15,33 @@
...
+    private bool $transactionalModeEnabled = false;
+
+    /** @var array<string, array<string, true>> */
+    private array $transactionalEvents = [];
+
     public function __construct(private DocumentManager $dm, private UnitOfWork $uow, private EventManager $evm)
     {
     }
 
+    public function clearTransactionalState(): void
```
malarzm, line 37: I think I'd slap an `@internal` to both new methods here.

lib/Doctrine/ODM/MongoDB/Utility/LifecycleEventManager.php
```diff
@@ -55,11 +74,18 @@ public function postCollectionLoad(PersistentCollectionInterface $coll): void
...
-    public function postPersist(ClassMetadata $class, object $document): void
+    public function postPersist(ClassMetadata $class, object $document, ?Session $session = null): void
     {
-        $class->invokeLifecycleCallbacks(Events::postPersist, $document, [new LifecycleEventArgs($document, $this->dm)]);
-        $this->dispatchEvent($class, Events::postPersist, new LifecycleEventArgs($document, $this->dm));
-        $this->cascadePostPersist($class, $document);
+        if (! $this->shouldDispatchEvent($document, Events::postPersist)) {
+            return;
+        }
+
```
malarzm, line 87: Should there be a failsafe checking for `$session !== null` when transactional mode is enabled? Or are we separating db-transaction from this concept?

lib/Doctrine/ODM/MongoDB/Utility/LifecycleEventManager.php
malarzm: I think it's like this since forever and whatever reason there was, it holds no more :)

lib/Doctrine/ODM/MongoDB/Utility/LifecycleEventManager.php
malarzm: +1 for using `spl_object_hash` directly, we do the same in UoW

lib/Doctrine/ODM/MongoDB/Event/LifecycleEventArgs.php
alcaeus: Good point. I've added a new method and removed the `$isInTransaction` constructor parameter.

lib/Doctrine/ODM/MongoDB/Utility/LifecycleEventManager.php
alcaeus: Ah, the ~good~ old times when we didn't have variadics. Variadics would make this a little nicer, but I don't think there's a good way to refactor this without serious breakage.

lib/Doctrine/ODM/MongoDB/Event/PreLoadEventArgs.php
alcaeus: Split across multiple lines. The reference does not prevent property promotion, so I've inlined this as well.

lib/Doctrine/ODM/MongoDB/Event/PreUpdateEventArgs.php
alcaeus: I renamed the private property to keep the parameter name the same, thus avoiding a potential BC break. While many frameworks consider parameter names exempt from a BC promise, there's no need to change this name.

lib/Doctrine/ODM/MongoDB/Utility/LifecycleEventManager.php
alcaeus: I extracted them into variables where they were constructed in an array. In `preUpdate`, I've extracted the first occurrence and split the `dispatchEvent` calls into multiple lines to make it more readable.

lib/Doctrine/ODM/MongoDB/Utility/LifecycleEventManager.php
alcaeus: The entire class is internal, which is why I haven't added it to the methods.

lib/Doctrine/ODM/MongoDB/Utility/LifecycleEventManager.php
alcaeus: I was planning on separating these concepts, unless there's a good reason to check whether the session is in a transaction. WDYT @jmikola?

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
alcaeus: Yes, those were due to me reusing the `CommitOptions` type from `UnitOfWork`, which omits the `session` option on purpose. I've created a separate type in `DocumentPersister` to allow that option.

malarzm (approved):

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -74,7 +74,14 @@
  *
  * @template T of object
  *
- * @psalm-import-type CommitOptions from UnitOfWork
+ * @psalm-type CommitOptions array{
```
jmikola, line 77: Noted that this was introduced to resolve https://github.com/doctrine/mongodb-odm/pull/2594#discussion_r1431235563.

Also, TIL about [`@psalm-import-type`](https://psalm.dev/docs/annotating_code/supported_annotations/#psalm-import-type), which was originally introduced in https://github.com/doctrine/mongodb-odm/commit/e00e9e90015e432151a8869b65b648d1d2ed7345#diff-ddc55123df78abdb7a5c057a6e80364bb4a234271e18aac03392e6939ed3b26aR48. I think that answers a previous question I had about re-using type definitions in PHPLIB.

lib/Doctrine/ODM/MongoDB/Utility/LifecycleEventManager.php
jmikola: Given that LifecycleEventManager is an internal class, I imagine `$session` should always be non-null if ODM is behaving properly.

Would this be an appropriate use case for an `assert()`, or perhaps an InvalidArgumentException would be more user-friendly? Either way, it seems reasonable to have some sanity check for this even though it's only relevant to how the event args are constructed.

See also: https://github.com/doctrine/mongodb-odm/pull/2594#discussion_r1433075399

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -456,6 +456,8 @@ public function commit(array $options = []): void
             $this->evm->dispatchEvent(Events::onFlush, new Event\OnFlushEventArgs($this->dm));
 
             if ($this->useTransaction($options)) {
+                $this->lifecycleEventManager->enableTransactionalMode();
```
jmikola: Related to @malarzm's other thread asking about asserting that `$session` is non-null when LifecycleEventManager dispatches events, this appears to be the only place where transactional mode is enabled, and that is done based on the `withTransaction` _option_ being specified (no direct Session check at this point).

I realize `with_transaction()` below is going to be responsible for passing a Session, which we'll then forward to `doCommit()`, but I don't believe there will be any further checks on a non-null Session after that point.

In that case, perhaps it's undesirable to perform such an assertion in LifecycleEventManager. But if we do add a check, I'd at least suggest capturing this logic in a comment for the first assertion within the file (to explain _why_ we're asserting that the Session is set, and how we're trusting UnitOfWork to ferry it along to the dispatch method).

jmikola (approved):
I have some thoughts on the Session assertion in LifecycleEventManager, but the rest of this LGTM.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
alcaeus: Good point. I added checks for this: `enableTransactionalMode` now requires a session that will be stored. When checking if an event should be dispatched, the given session is compared with the one stored to ensure we're still running in the same transaction. This should guard against dispatching events without a session in transactional mode.

PR #2604, opened 2024-01-09
Only retry transaction once

alcaeus:
This PR adds a modified implementation of MongoDB's [Convenient Transaction API](https://github.com/mongodb/specifications/blob/master/source/transactions-convenient-api/transactions-convenient-api.rst) specification that is better suited for our use case. When encountering transient errors, the convenient API will retry the transaction for up to 120 seconds. The only way to change this time limit is by using the `timeoutMS` option specified in the [Client Side Operations Timeout (CSOT)](https://github.com/mongodb/specifications/blob/master/source/client-side-operations-timeout/client-side-operations-timeout.md) specification. However, libmongoc and thus by extension the PHP driver do not yet implement this specification, so users have no way of reducing the time `with_transaction` will potentially take. Since 120 seconds is not a good default timeout in a web context and users of the ODM will most likely want to abort sooner on transient errors, a custom implementation is needed.

The implementation in this PR takes the entire logic of the convenient transaction API, but only allows the closure to be run twice, meaning that transient command errors will lead to a single retry. For the time being, the amount of retries is not configurable, but this can be added later. I would like to avoid introducing a config setting unless absolutely needed so that we remain flexible to switch back to using the API provided by the driver once it implements the CSOT specification.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
+++ b/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
@@ -3186,13 +3186,13 @@ final class UnitOfWork implements PropertyChangedListener
      */
     private function withTransaction(Session $session, callable $callback, array $transactionOptions = []): void
     {
-        $closureAttempts = 0;
+        $numAttempts = 0;
 
         while (true) {
             $session->startTransaction($transactionOptions);
 
             try {
-                $closureAttempts++;
+                $numAttempts++;
                 call_user_func($callback, $session);
             } catch (Throwable $e) {
                 if ($session->isInTransaction()) {
@@ -3202,7 +3202,7 @@ final class UnitOfWork implements PropertyChangedListener
                 if (
                     $e instanceof RuntimeException &&
                     $e->hasErrorLabel('TransientTransactionError') &&
-                    ! $this->shouldAbortWithTransaction($closureAttempts)
+                    ! $this->shouldAbortWithTransaction($numAttempts)
                 ) {
                     continue;
                 }
@@ -3222,7 +3222,7 @@ final class UnitOfWork implements PropertyChangedListener
                     if (
                         $e->getCode() !== 50 /* MaxTimeMSExpired */ &&
                         $e->hasErrorLabel('UnknownTransactionCommitResult') &&
-                        ! $this->shouldAbortWithTransaction($closureAttempts)
+                        ! $this->shouldAbortWithTransaction($numAttempts)
                     ) {
                         // Retry committing the transaction
                         continue;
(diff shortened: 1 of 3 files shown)


lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -3177,4 +3179,77 @@ private function getTransactionOptions(array $options): array
             self::TRANSACTION_OPTIONS,
         );
     }
+
+    /**
+     * This following method was taken from the MongoDB Library and adapted to not use the default 120 seconds timeout.
+     * The code within this method is licensed under the Apache License. Copyright belongs to MongoDB, Inc.
+     */
```
alcaeus, line 3186: Not sure what the best course of action is here, but since the code in question comes from the apache-2.0-licensed MongoDB library and the ODM is MIT licensed, we do need some kind of license and copyright information. Let me know your thoughts on this.

tests/Doctrine/ODM/MongoDB/Tests/UnitOfWorkTransactionalCommitConsistencyTest.php
```diff
@@ -534,12 +578,11 @@ protected static function getConfiguration(): Configuration
         return $configuration;
     }
 
-    private function createTransientFailPoint(string $failCommand): void
+    private function createTransientFailPoint(string $failCommand, int $times = 1): void
     {
         $this->dm->getClient()->selectDatabase('admin')->command([
             'configureFailPoint' => 'failCommand',
-            // Trigger the error twice, working around retryable writes
```
alcaeus, line 541: While testing this change, I noticed that with a single retry all other tests started failing. Upon further investigation, it turns out that using an additional failure to account for retryable writes isn't necessary. Quoting from the [Retryable Writes Documentation](https://www.mongodb.com/docs/manual/core/retryable-writes/#retryable-writes-and-multi-document-transactions):
> The write operations inside the transaction are not individually retryable, regardless of value of [retryWrites.](https://www.mongodb.com/docs/manual/reference/connection-string/#mongodb-urioption-urioption.retryWrites)

With that in mind, we don't need to account for this and a single failure is sufficient for this fail point (except for the one test where we want multiple failures to specifically test that the operation is only retried once).

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -3177,4 +3179,77 @@ private function getTransactionOptions(array $options): array
...
+    private function withTransaction(Session $session, callable $callback, array $transactionOptions = []): void
+    {
+        $closureAttempts = 0;
```
jmikola: `$numAttempts` may be clearer.

jmikola (approved):
Code changes LGTM.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
alcaeus: Yes, absolutely. Changed 👍

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
malarzm: I think it's OK but I'm not a lawyer :) Maybe you could ask internally if it's sufficient?

PR #2607, opened 2024-01-16
Allow elements from other namespaces

malarzm:
This is how we should ideally allow extensions to be combined with ODM's schema. Sadly, this does not work and it seems work on extensions' maintainers is needed anyway :( 

Cross referencing https://github.com/doctrine-extensions/DoctrineExtensions/issues/2318

### DIFF
--- a/doctrine-mongo-mapping.xsd
+++ b/doctrine-mongo-mapping.xsd
@@ -103,6 +103,7 @@
       <xs:element name="shard-key" type="odm:shard-key" minOccurs="0" />
       <xs:element name="read-preference" type="odm:read-preference" minOccurs="0" />
       <xs:element name="schema-validation" type="odm:schema-validation" minOccurs="0" />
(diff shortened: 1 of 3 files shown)


greg0ire:
> Sadly, this does not work and it seems work on extensions' maintainers is needed anyway

I think it's because the extension xsd file is not provided here: https://github.com/doctrine/mongodb-odm/blob/c9b7f6bdec37f32c2b91637adeee52949e7596e6/lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php#L718-L720

You'd have to let the user provide a schema combining both xsd files as in https://github.com/doctrine/orm/issues/11117#issuecomment-1894492234

malarzm:
@greg0ire thanks for all the pointers! I'll prolly join you in https://github.com/doctrine/orm/issues/11123 soon

PR #2596, opened 2023-12-08
Use attributes in tests

franmomu:
This PR changes most of the annotations for attributes since our minimum PHP version is 8.1, the tests changed are the ones relying on `BaseTestCase::createMetadataDriverImpl()`

This builds on top of https://github.com/doctrine/mongodb-odm/pull/2502 to make sure it works fine.

[I've used rector](https://github.com/rectorphp/rector-doctrine/pull/239) with some manual adjustments.

PR #2610, opened 2024-02-11
Fix criteria mapping type

franmomu:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | 

#### Summary

While doing https://github.com/doctrine/mongodb-odm/pull/2596, the SA tool complained about the type of this property. For example in https://github.com/doctrine/mongodb-odm/blob/a51bfc01425d2c30b69f2c5ce13fe0d3c58bbd98/tests/Documents/BlogPost.php#L63
`$criteria` is holding an array that contains a boolean.

It is used in https://github.com/doctrine/mongodb-odm/blob/a51bfc01425d2c30b69f2c5ce13fe0d3c58bbd98/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php#L762-L767
and 
https://github.com/doctrine/mongodb-odm/blob/a51bfc01425d2c30b69f2c5ce13fe0d3c58bbd98/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php#L830-L834

so it looks like it could contain any type, not only strings.

Ref: https://github.com/doctrine/mongodb-odm/pull/2596/files#r1441864943


malarzm (approved):
Good catch! There's new PHPStan complaint tho :D

franmomu:
> Good catch! There's new PHPStan complaint tho :D

I've ignored the complaint since it has been reported in https://github.com/vimeo/psalm/issues/10578

SenseException (approved):
<![CDATA[Thank you]]>

PR #2617, opened 2024-03-19
Drop support for MongoDB 4.4

malarzm:
MongoDB 4.4 reached its EOL in February so let's make the CI lighter for upcoming version

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -25,7 +25,6 @@ jobs:
           - "7.0"
           - "6.0"
           - "5.0"
-          - "4.4"
         driver-version:
           - "stable"
         topology:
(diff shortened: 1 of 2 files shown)


SenseException:
If 4.4 gets dropped, does it make sense to have older versions mentioned in the docs? At this point multi-document transactions are a given.
https://github.com/doctrine/mongodb-odm/blob/2.8.x/docs/en/reference/transactions-and-concurrency.rst?plain=1#L18-L19

malarzm:
Good catch! It does not. Github did not showed me those when grepping for `4.4` with UI. Will adjust it later :)

SenseException (approved):

GromNaN:
For the record, even if the server is no longer supported in version 4.4, we continue to maintain the driver for 3 years from the server version EOL.

GromNaN (approved):

malarzm:
> For the record, even if the server is no longer supported in version 4.4, we continue to maintain the driver for 3 years from the server version EOL.

Luckily enough no code paths were made, so technically 4.4 still works with ODM. We're just not testing against it which I think is fair given it has reached EOL :)

I'll be merging this PR as CI fails are not related.

malarzm:
Dang, I've missed some 5.0 sharded cluster fail. That's what one gets from a broken CI :( I'll be looking into it

PR #2249, opened 2020-11-26
Use symfony/deprecation-contracts to announce deprecations in code

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

This replaces manual calls to `@trigger_error` with `trigger_deprecation` from the `symfony/deprecation-contracts` package. This allows users to override the `trigger_deprecation` function to implement their custom logic.

Note that this PR drops deprecation functionality for field mapping annotations that were supposed to be dropped in 2.0. Since these methods are marked `@internal`, I have no problem dropping them in a minor release.

PR #2622, opened 2024-04-17
Fix deprecation notices Indexes annotation

jmikola:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Related issues | #2235, #2249

#### Summary

The `Indexes` attribute/annotation was deprecated in [doctrine/mongodb-odm@7435e66](https://github.com/doctrine/mongodb-odm/commit/7435e66b974130931b3a4acadb492fe572ff9298) for 2.2. There were deprecation messages added for using `Indexes` at the class level, and `indexes` within the Document annotation; however, it looks like Indexes on the property level was [missed](https://github.com/doctrine/mongodb-odm/blob/7435e66b974130931b3a4acadb492fe572ff9298/lib/Doctrine/ODM/MongoDB/Mapping/Driver/AnnotationDriver.php#L189).

Also, one of the deprecation messages got screwed up in [doctrine/mongodb-odm@3bdfe97](https://github.com/doctrine/mongodb-odm/commit/3bdfe97be472160ebbc924d5882ce61bd640a515#diff-65de5973d8c41abd6a3cc14c862908909d49367a01216544630b343b500929d1R153). The message changed but the sprintf args weren't swapped to account for that.

I struggled to add tests for this because:

 * `trigger_deprecation` calls `@trigger_error()`, so output is silenced. I could theoretically use [`error_get_last()`](https://www.php.net/manual/en/function.error-get-last.php) to check for the deprecation message, but there's no prior art for doing so. In fact, no deprecation messages seem to be tested.
 * The `Indexes` attribute cannot target a property, so the property-level deprecation message could only trigger when using it as annotation.
 * I couldn't discern the syntax for using `Indexes` as a class-level attribute with `Index` nested within. I consulted doctrine/orm#9241 some some examples of nesting attributes, but all of my attempts resulted in exceptions (e.g. `$value` property of `Indexes` couldn't be set).

For manual testing of the annotation form, this is what I used:

```php
/**
 * @ODM\Document
 * @ODM\Indexes({
 *   @ODM\Index(keys={"foo"="asc"})
 * })
 */
class DeprecatedIndexesAttributeOnClass
{
    /**
     * @ODM\Field(type="string")
     * @ODM\Indexes({
     *   @ODM\Index()
     * })
     */
    public string $foo;
}
```

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/Mapping/AnnotationDriverTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Mapping/AnnotationDriverTest.php
@@ -5,8 +5,18 @@ declare(strict_types=1);
 namespace Doctrine\ODM\MongoDB\Tests\Mapping;
 
 use Doctrine\Common\Annotations\AnnotationReader;
+use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
+use Doctrine\ODM\MongoDB\Mapping\Annotations\Document;
+use Doctrine\ODM\MongoDB\Mapping\ClassMetadata;
 use Doctrine\ODM\MongoDB\Mapping\Driver\AnnotationDriver;
 use Doctrine\Persistence\Mapping\Driver\MappingDriver;
+
+use function call_user_func;
+use function restore_error_handler;
+use function set_error_handler;
+use function sprintf;
+
+use const E_USER_DEPRECATED;
 
 class AnnotationDriverTest extends AbstractAnnotationDriverTestCase
 {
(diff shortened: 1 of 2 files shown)


IonBazan (approved):
I think we are missing tests for deprecation messages. Should we add them in this PR then?

jmikola:
> I think we are missing tests for deprecation messages. Should we add them in this PR then?

I'll make an attempt to do so before wrapping this up. We have some prior art for doing this in [PHPLIB](https://github.com/mongodb/mongo-php-library/blob/1.18.0/tests/TestCase.php#L188), but [Expecting Deprecations (E_USER_DEPRECATED)](https://docs.phpunit.de/en/11.1/error-handling.html#expecting-deprecations-e-user-deprecated) may have a better approach.

jmikola:
I made some progress in asserting deprecation messages, but ran into a roadblock attempting to use Indexes as an attribute (annotation is fine). I'll hold off on this until we can talk it through in the [Slack thread](https://doctrine.slack.com/archives/CAAE6T66B/p1713538516260789).

jmikola:
Manually rebased on 2.7.x, since that's the same branch I used for previous fixes (https://github.com/doctrine/mongodb-odm/pull/2627).

jmikola:
Deprecation messages for annotations are now tested. CI failures are unrelated to this PR and due to https://github.com/doctrine/mongodb-odm/issues/2624.

alcaeus (approved):

tests/Doctrine/ODM/MongoDB/Tests/Mapping/AnnotationDriverTest.php
```diff
@@ -16,4 +26,125 @@ protected static function loadDriver(): MappingDriver
...
+        self::assertSame(sprintf('Since doctrine/mongodb-odm 2.2: The "@Indexes" attribute used in property "foo" of class "%s" is deprecated. Specify all "@Index" and "@UniqueIndex" attributes on the class.', DeprecatedIndexesPropertyAnnotation::class), $errors[0]);
+
+        $indexes = $classMetadata->indexes;
+
+        self::assertTrue(isset($indexes[0]['keys']['foo']));
+        self::assertEquals(1, $indexes[0]['keys']['foo']);
+    }
+
+    /** @param list<string>|null $errors */
+    private function captureDeprecationMessages(callable $callable, ?array &$errors): mixed
```
IonBazan, line 88: I think we can also use https://symfony.com/doc/current/components/phpunit_bridge.html but seems like a drastic change so let's keep it this way for now.

IonBazan (approved):

tests/Doctrine/ODM/MongoDB/Tests/Mapping/AnnotationDriverTest.php
jmikola: Aye. We use that on the MongoDB driver but I noted it wasn't being utilized here.

PR #2630, opened 2024-04-24
Support search indexes in mappings and SchemaManager

jmikola:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | [PHPORM-38](https://jira.mongodb.org/browse/PHPORM-38)

### DIFF
--- a/docs/en/reference/search-indexes.rst
+++ b/docs/en/reference/search-indexes.rst
@@ -1,0 +1,188 @@
+.. _search_indexes:
+
+Search Indexes
+==============
+
+In addition to standard :ref:`indexes <indexes>`, ODM allows you to define
+search indexes for use with `MongoDB Atlas Search <https://www.mongodb.com/docs/atlas/atlas-search/>`__.
+Search indexes may be queried using the `$search <https://www.mongodb.com/docs/atlas/atlas-search/aggregation-stages/search/>`__
+and `$searchMeta <https://www.mongodb.com/docs/atlas/atlas-search/aggregation-stages/searchMeta/>`__
+aggregation pipeline stages.
+
+Search indexes have some notable differences from regular
+:ref:`indexes <indexes>` in ODM. They may only be defined on document classes.
+Definitions will not be incorporated from embedded documents. Additionally, ODM
+will **NOT** translate field names in search index definitions. Database field
+names must be used instead of mapped field names (i.e. PHP property names).
+
+Search Index Options
+--------------------
+
+Search indexes are defined using a more complex syntax than regular
+:ref:`indexes <indexes>`.
+
+ODM supports the following search index options:
+
+-
+    ``name`` - Name of the search index to create, which must be unique to the
+    collection. Defaults to ``"default"``.
+-
+    ``dynamic`` - Enables or disables dynamic field mapping for this index.
+    If ``true``, the index will include all fields with
+    `supported data types <https://www.mongodb.com/docs/atlas/atlas-search/define-field-mappings/#std-label-bson-data-chart>`__.
+    If ``false``, the ``fields`` attribute must be specified. Defaults to ``false``.
+-
+    ``fields`` - Associative array of `field mappings <https://www.mongodb.com/docs/atlas/atlas-search/define-field-mappings/>`__
+    that specify the fields to index (keys). Required only if dynamic mapping is disabled.
+-
+    ``analyzer`` - Specifies the `analyzer <https://www.mongodb.com/docs/atlas/atlas-search/analyzers/>`__
+    to apply to string fields when indexing. Defaults to the
+    `standard analyzer <https://www.mongodb.com/docs/atlas/atlas-search/analyzers/standard/>`__.
+-
+    ``searchAnalyzer`` - Specifies the `analyzer <https://www.mongodb.com/docs/atlas/atlas-search/analyzers/>`__
+    to apply to query text before the text is searched. Defaults to the
+    ``analyzer`` attribute, or the `standard analyzer <https://www.mongodb.com/docs/atlas/atlas-search/analyzers/standard/>`__.
+    if both are unspecified.
+-
+    ``analyzers`` - Array of `custom analyzers <https://www.mongodb.com/docs/atlas/atlas-search/analyzers/custom/>`__
+    to use in this index.
+-
+    ``storedSource`` - Specifies document fields to store for queries performed
+    using the `returnedStoredSource <https://www.mongodb.com/docs/atlas/atlas-search/return-stored-source/>`__
+    option. Specify ``true`` to store all fields, ``false`` to store no fields,
+    or a `document <https://www.mongodb.com/docs/atlas/atlas-search/stored-source-definition/#std-label-fts-stored-source-document>`__
+    to specify individual fields to include or exclude from storage. Defaults to ``false``.
+-
+    ``synonyms`` - Array of `synonym mapping definitions <https://www.mongodb.com/docs/atlas/atlas-search/synonyms/>`__
+    to use in this index.
+
+Additional documentation for defining search indexes may be found in
+`search index definition <https://www.mongodb.com/docs/manual/reference/command/createSearchIndexes/#search-index-definition-syntax>`__
+within the MongoDB manual.
+
+Static Mapping
+--------------
+
+`Static mapping <https://www.mongodb.com/docs/atlas/atlas-search/define-field-mappings/#static-mappings>`__
+can be used to configure indexing of specific fields within a document.
+
+The following example demonstrates how to define a search index using static
+mapping.
+
+.. configuration-block::
+
+    .. code-block:: php
+
+        <?php
+
+        /**
+         * @Document
+         * @SearchIndex(
+         *   name="usernameAndAddresses",
+         *   fields={
+         *     "username"={
+         *       {"type"="string"},
+         *       {"type"="autocomplete"},
+         *     },
+         *     "addresses"={"type"="embeddedDocuments", "dynamic"=true},
+         *   },
+         * )
+         */
+        class User
+        {
+            /** @Id */
+            private $id;
+
+            /** @Field(type="string") */
+            private $username;
+
+            /** @EmbedMany(targetDocument=Address::class) */
+            private $addresses;
+
+            // ...
+        }
+
+    .. code-block:: xml
+
+        <doctrine-mongo-mapping xmlns="http://doctrine-project.org/schemas/orm/doctrine-mongo-mapping"
+              xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+              xsi:schemaLocation="http://doctrine-project.org/schemas/orm/doctrine-mongo-mapping
+                            http://doctrine-project.org/schemas/orm/doctrine-mongo-mapping.xsd">
+
+            <document name="Documents\User">
+                <search-indexes>
+                    <search-index name="usernameAndAddresses">
+                        <field name="username" type="string" />
+                        <field name="username" type="autocomplete" />
+                        <field name="addresses" type="embeddedDocuments" dynamic="true" />
+                    </search-index>
+                </search-indexes>
+
+                <!-- ... -->
+            </document>
+        </doctrine-mongo-mapping>
+
+The ``username`` field will indexed both as a string and for autocompletion.
+Since the ``addresses`` field uses an :ref:`embed-many <embed_many>`
+relationship, it must be indexed using the ``embeddedDocuments`` type; however,
+embedded documents within the array are permitted to use dynamic mapping.
+
+Dynamic Mapping
+---------------
+
+`Dynamic mapping <https://www.mongodb.com/docs/atlas/atlas-search/define-field-mappings/#dynamic-mappings>`__
+can be used to automatically index fields with
+`supported data types <https://www.mongodb.com/docs/atlas/atlas-search/define-field-mappings/#std-label-bson-data-chart>`__
+within a document. Dynamically mapped indexes occupy more disk space than
+statically mapped indexes and may be less performant; however, they may be
+useful if your schema changes  or for when experimenting with Atlas Search
+
+.. note::
+
+    Atlas Search does **NOT** dynamically index embedded documents contained
+    within arrays (e.g. :ref:`embed-many <embed_many>` relationships). You must
+    use static mappings with the `embeddedDocument <https://www.mongodb.com/docs/atlas/atlas-search/field-types/embedded-documents-type/>`__
+    field type.
+
+The following example demonstrates how to define a search index using dynamic
+mapping:
+
+.. configuration-block::
+
+    .. code-block:: php
+
+        <?php
+
+        /**
+         * @Document
+         * @SearchIndex(dynamic=true)
+         */
+        class BlogPost
+        {
+            /** @Id */
+            private $id;
+
+            /** @Field(type="string") */
+            private $title;
+
+            /** @Field(type="string") */
+            private $body;
+
+            // ...
+        }
+
+    .. code-block:: xml
+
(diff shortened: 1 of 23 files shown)


doctrine-mongo-mapping.xsd
```diff
@@ -466,6 +467,96 @@
     </xs:choice>
   </xs:complexType>
 
+  <xs:complexType name="search-indexes">
+    <xs:choice minOccurs="0" maxOccurs="unbounded">
```
alcaeus: Is there a reason to have an empty `search-indexes` element anywhere? Otherwise you could consider using `minOccurs="1"`.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -318,6 +321,140 @@ public function deleteDocumentIndexes(string $documentName, ?int $maxTimeMs = nu
         $this->dm->getDocumentCollection($documentName)->dropIndexes($this->getWriteOptions($maxTimeMs, $writeConcern));
     }
 
+    /**
+     * Create search indexes for all mapped document classes.
+     */
+    public function createSearchIndexes(): void
```
alcaeus, line 327: Leaving a note to follow up on later, but these methods aren't invoked by the schema commands yet. Since search indexes are an Atlas-only feature, I would suggest duplicating the existing logic for managing indexes instead of combining them with regular indexes.

alcaeus (commented):
Initial review looks good. There's a number of PHPStan issues due to not properly specifying all array key/value pairs. I'll let you decide whether it's worth specifying them all or whether you'd rather add the errors to the baseline.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
jmikola: I'm still working through this in the command classes and forgot to preemptively leave a comment about this last week. I was on the fence about creating entirely new methods since that would change the public interface for AbstractCommand. If you're not concerned about that being a BC break, I have no qualms in doing so.

Otherwise, I planned to incorporate this into the `--index` option and introduce a supporting option to enable/disable search indexes. That feels clunkier though, so I'm certainly in favor of a new, explicit option.

alcaeus: While `AbstractCommand` is not marked as internal, I'm not particularly concerned about people extending it, and we can work around it by providing a default implementation for the new methods instead of leaving them empty to avoid compile errors. Any objections to that plan @malarzm?

jmikola: A default implementation seems prudent, even if it'd only be useful internally for ShardCommand (where I'm presently throwing a BadMethodCallException like other unsupported methods).

malarzm: I'd say do whatever you guys need with `AbstractCommand`. I deem it highly unlikely that somebody extended those. And if they did they'll be smart enough to fix it themselves ;)

doctrine-mongo-mapping.xsd
jmikola: This was copied from the `<indexes>` definition, but I see no reason we can't use `minOccurs="1"` here. I'll leave `<indexes>` as-is, though.

The actual change here will just be removing `minOccurs="0"`, since it defaults to one.

jmikola:
Sharding failures seem unrelated: https://github.com/doctrine/mongodb-odm/actions/runs/8914210744/job/24481504241?pr=2630

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/SearchIndex.php
```diff
@@ -0,0 +1,39 @@
...
+{
+    /**
+     * @param array<string, array>|null     $fields
+     * @param list<array>|null              $analyzers
+     * @param SearchIndexStoredSource|null  $storedSource
+     * @param list<SearchIndexSynonym>|null $synonyms
+     */
+    public function __construct(
+        public ?string $name = null,
+        public ?bool $dynamic = null,
```
GromNaN, line 30: Wouldn't it be better to set actual default values rather than nullable types? That would document the default values in the code directly.

```suggestion
        public bool $dynamic = false,
```

But then, my suggestion isn't consistent with the rest of the Doctrine attributes.

tests/Documents/CmsArticle.php
```diff
@@ -6,9 +6,9 @@
 
 use Doctrine\Common\Collections\Collection;
 use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
-use Doctrine\ODM\MongoDB\Mapping\Annotations\Index;
 
-#[Index(keys: ['topic' => 'asc'])]
+#[ODM\Index(keys: ['topic' => 'asc'])]
+#[ODM\SearchIndex(dynamic: true)]
```
GromNaN, line 11: Could you use all the fields in the attribute to make the test more exhaustive?

lib/Doctrine/ODM/MongoDB/Tools/Console/Command/Schema/AbstractCommand.php
```diff
@@ -34,23 +36,81 @@ protected function configure()
...
-    abstract protected function processDocumentCollection(SchemaManager $sm, string $document, ?int $maxTimeMs, ?WriteConcern $writeConcern);
+    /**
+     * @return void
+     *
+     * @throws BadMethodCallException
+     */
+    protected function processDocumentCollection(SchemaManager $sm, string $document, ?int $maxTimeMs, ?WriteConcern $writeConcern)
+    {
+        throw new BadMethodCallException('This command does not support collections');
+    }
```
alcaeus, line 47: 👍 for adding default implementations for all methods in this class.

lib/Doctrine/ODM/MongoDB/Tools/Console/Command/Schema/CreateCommand.php
```diff
@@ -53,17 +60,23 @@ private function doExecute(InputInterface $input, OutputInterface $output): int
         $isErrored = false;
 
         foreach ($create as $option) {
+            $method = match ($option) {
+                self::COLLECTION => 'Collection',
+                self::INDEX => 'Index',
+                self::SEARCH_INDEX => 'SearchIndex',
+                default => throw new UnhandledMatchError($option),
```
alcaeus: This is standard behaviour in PHP, see https://3v4l.org/LU66Z.
```suggestion
```

lib/Doctrine/ODM/MongoDB/Tools/Console/Command/Schema/DropCommand.php
```diff
@@ -50,17 +59,24 @@ private function doExecute(InputInterface $input, OutputInterface $output): int
         $isErrored = false;
 
         foreach ($drop as $option) {
+            $method = match ($option) {
+                self::DB => 'Db',
+                self::COLLECTION => 'Collection',
+                self::INDEX => 'Index',
+                self::SEARCH_INDEX => 'SearchIndex',
+                default => throw new UnhandledMatchError($option),
```
alcaeus:
```suggestion
```

lib/Doctrine/ODM/MongoDB/Tools/Console/Command/Schema/UpdateCommand.php
```diff
@@ -28,14 +27,16 @@ protected function configure()
         $this
             ->setName('odm:schema:update')
             ->addOption('class', 'c', InputOption::VALUE_OPTIONAL, 'Document class to process (default: all classes)')
+            ->addOption('disable-search-indexes', null, InputOption::VALUE_NONE, 'Do not update search indexes')
```
alcaeus: I'm aware the naming is consistent with `disable-validators`, but wouldn't `skip-search-indexes` be more appropriate?

alcaeus (changes requested):

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/SearchIndex.php
jmikola: I did originally consider this, but thought it best to minimize what ends up in the definition shipped to `createSearchIndexes`.

tests/Documents/CmsArticle.php
jmikola: The mapping tests are exhaustive in their use of all options. This model is only used for the SchemeManager tests, in which we assert that _something_ is passed along to the mocked MongoDB\Collection methods. That is based on prior art for the standard index tests.

lib/Doctrine/ODM/MongoDB/Tools/Console/Command/Schema/CreateCommand.php
jmikola: Indeed it is; however, PHPStan doesn't approve: https://github.com/phpstan/phpstan/issues/4451

I can replace this with a baseline addition. I'm currently only relying on the baseline for the truly unreasonable issues (`fields` and `analyzers` options).

lib/Doctrine/ODM/MongoDB/Tools/Console/Command/Schema/UpdateCommand.php
jmikola: SGTM. I'll make the change.

Per our earlier discussion about default behavior for CreateCommand and DropCommand, do you think a similar option is warranted for those? Since search indexes are included in the default `$createOrder` and `$dropOrder` lists, I suppose users would need to tweak the CLI args either way (be it `--skip-search-indexes` or explicitly specifying the things to process). So I'm not convinced it's needed there, in contrast to UpdateCommand.

lib/Doctrine/ODM/MongoDB/Tools/Console/Command/Schema/CreateCommand.php
alcaeus: I'll leave this up to you and chalk this up as another point against PHPStan. Feel free to resolve after reading.

lib/Doctrine/ODM/MongoDB/Tools/Console/Command/Schema/UpdateCommand.php
alcaeus: I'd say for now we can leave create and drop as-is, but I agree that it's unfortunate to have to redefine the entire list when wanting to exclude a single item, especially as that can create issues when we add new functionality. A dedicated `skip-<feature>` or `include-<feature>` (for optional things) might be better, despite it meaning more command-line arguments. Let's create a ticket for this and put it on the map for a future refactoring of those commands.

jmikola: Opened https://github.com/doctrine/mongodb-odm/issues/2634. I only proposed `skip-<feature>` as the commands already support opt-in behavior. For example, CreateCommand's `--collection` option would accomplish the same thing as `include-collection`, assuming both would disable the `$createOrder` and expect everything to be explicitly requested (as `--collection` does today).

lib/Doctrine/ODM/MongoDB/Tools/Console/Command/Schema/CreateCommand.php
jmikola: In the interest of keeping the code cleaner, I'll remove the useless `default` cases and track this in the PHPStan baseline.

jmikola:
Psalm is failing due to [two unused baseline entries](https://github.com/doctrine/mongodb-odm/actions/runs/8972457667/job/24640499243?pr=2630#step:6:27), but I'm unable to reproduce that locally with PHP 8.2 (same as is used in CI):

```
Error: lib/Doctrine/ODM/MongoDB/UnitOfWork.php:0:0: UnusedBaselineEntry: Baseline for issue "TypeDoesNotContainType" has 1 extra entry. (see https://psalm.dev/316)
Error: tests/Doctrine/ODM/MongoDB/Tests/SchemaManagerTest.php:0:0: UnusedBaselineEntry: Baseline for issue "TypeDoesNotContainType" has 1 extra entry. (see https://psalm.dev/316)
```

UnitOfWork is also entirely unrelated to this PR, so I'm not sure what to make of that.

lib/Doctrine/ODM/MongoDB/Tools/Console/Command/Schema/UpdateCommand.php
jmikola: Renamed this to `skip-search-indexes` and added the same option to CreateCommand and DropCommand. #2634 can be used for renaming `disable-validators` and adding additional `skip-<feature>` options if desired.

alcaeus:
> Psalm is failing due to [two unused baseline entries](https://github.com/doctrine/mongodb-odm/actions/runs/8972457667/job/24640499243?pr=2630#step:6:27), but I'm unable to reproduce that locally with PHP 8.2 (same as is used in CI):

This could be related to using an outdated psalm version - I was able to reproduce it locally with psalm 5.24.0.

lib/Doctrine/ODM/MongoDB/Tools/Console/Command/Schema/UpdateCommand.php
alcaeus: Sounds good, thank you!

alcaeus (approved):
LGTM with psalm issues fixed.

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/SearchIndex.php
alcaeus: Agree - we've typically used nullable bools when we want to omit values from the command unless they were explicitly specified, and the only way to achieve this is by using a `null` default value.

jmikola:
Docs are still in progress (very not fun).

tests/Doctrine/ODM/MongoDB/Tests/Mapping/AbstractMappingDriverTestCase.php
```diff
@@ -634,6 +686,27 @@ public function testEnumType(): void
...
+ *       {
+ *         "type"="string",
+ *         "multi"={
+ *           "english"={"type"="string", "analyzer"="lucene.english"},
+ *           "french"={"type"="string", "analyzer"="lucene.french"},
+ *         },
+ *       },
+ *       {"type"="autocomplete"},
+ *     },
+ *     "embedded_phone_number"={"type"="embeddedDocuments", "dynamic"=true},
```
jmikola, line 703: Note: [embeddedDocuments](https://www.mongodb.com/docs/atlas/atlas-search/field-types/embedded-documents-type/) is used because the field is mapped as EmbedMany:

```
@ODM\EmbedMany(targetDocument=Phonenumber::class, name="embedded_phone_number")
```

If the relationship was EmbedOne then [document](https://www.mongodb.com/docs/atlas/atlas-search/field-types/document-type/) would be appropriate.

jmikola: I've ensured one of the docs examples uses `embeddedDocuments`.

alcaeus (approved):
Gave the docs a read and they look good to me. Since we can't easily test docs generation, I'd suggest we merge this as is and fix any RST issues after the fact.

docs/en/reference/search-indexes.rst
```diff
@@ -0,0 +1,188 @@
...
+
+.. configuration-block::
+
+    .. code-block:: php
+
+        <?php
+
+        /**
+         * @Document
+         * @SearchIndex(
```
stof, line 80: shouldn't this show an eample using attributes rather than annotations ?

jmikola: It looks like that was addressed in https://github.com/doctrine/mongodb-odm/commit/37013acdf1db8c10e2982e0700f72682d1fd2339

PR #2636, opened 2024-05-10
Fix of static analysis errors

SenseException:
Fixes Static analysis errors

|      Q       |   A
|------------- | -----------
| BC Break     | no

#### Summary

Static analysis does fail because there are fixes in new versions that the composer constraint allows. I reduce the constraints' range a bit with this PR and fix the psalm baseline.


SenseException:
Tests may fail, but the first step to fix them happens in #2635.

SenseException:
I'll rebase my branch to 2.7 when I'm back on my laptop.

alcaeus (approved):

PR #2638, opened 2024-05-10
Fix link target in Index annotation docs

jmikola:


malarzm (approved):

malarzm:
@jmikola mind fixing this in 2.7.x instead?

PR #2641, opened 2024-06-05
Decouple UnitOfWork from Hydrators

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

Hydrators currently get the `UnitOfWork` injected directly. This leads to a circular dependency, as the `DocumentManager` needs a `HydratorFactory`, but the `HydratorFactory` needs a `UnitOfWork` instance which can't exist without a `DocumentManager`. To resolve this, `HydratorFactory` no longer takes a `UnitOfWork` instance, and generated hydrators instead take the `UnitOfWork` instance exposed by the document manager.

This PR also moves a method from `UnitOfWork` to the `DocumentManager` as it's not tied to `UnitOfWork`.

Note that this PR is only the first of a few refactorings necessary to properly decouple things from `UnitOfWork`, with the goal of having `UnitOfWork` represent a single set of persistable document changes. Removing the circular dependency also allows us to make hydrator factories configurable, which is important to leverage the new codec infrastructure in the PHP driver and thus speed up a lot of hydration operations.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/DocumentManager.php
+++ b/lib/Doctrine/ODM/MongoDB/DocumentManager.php
@@ -882,6 +882,48 @@ class DocumentManager implements ObjectManager
         return $this->filterCollection;
     }
 
+    /**
+     * Gets the class name for an association (embed or reference) with respect
+     * to any discriminator value.
+     *
+     * @internal
+     *
+     * @param FieldMapping              $mapping
+     * @param array<string, mixed>|null $data
+     *
+     * @psalm-return class-string
+     */
+    public function getClassNameForAssociation(array $mapping, $data): string
+    {
+        $discriminatorField = $mapping['discriminatorField'] ?? null;
+
+        $discriminatorValue = null;
+        if (isset($discriminatorField, $data[$discriminatorField])) {
+            $discriminatorValue = $data[$discriminatorField];
+        } elseif (isset($mapping['defaultDiscriminatorValue'])) {
+            $discriminatorValue = $mapping['defaultDiscriminatorValue'];
+        }
+
+        if ($discriminatorValue !== null) {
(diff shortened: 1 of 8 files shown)


lib/Doctrine/ODM/MongoDB/DocumentManager.php
```diff
@@ -883,6 +882,48 @@ public function getFilterCollection(): FilterCollection
...
+     * to any discriminator value.
+     *
+     * @internal
+     *
+     * @param FieldMapping              $mapping
+     * @param array<string, mixed>|null $data
+     *
+     * @psalm-return class-string
+     */
+    public function getClassNameForAssociation(array $mapping, $data): string
```
malarzm, line 896: I'm not a fan of expanding DM's public surface but I'm not sure where else this could go... definitely not into generated hydrators as maintaining that will be PITA, ClassMetadataFactory is a no-go because of interface.

So just a food for thought, no other concerns from my end 👌

malarzm (approved):

lib/Doctrine/ODM/MongoDB/DocumentManager.php
alcaeus: Yeah, I moved it over because it doesn't need to be in there - ideally we'd have some kind of metadata construct that allows us to get this information. Either way, I'm saving that for a future refactoring of metadata (which I don't want to tackle)

franmomu (approved):

SenseException:
Are the failing shared cluster tests somehow relevant to this PR?

alcaeus:
> Are the failing shared cluster tests somehow relevant to this PR?

Nope, they have been failing for a while now and I haven't had a chance to dig into it and figure out what's causing them.

PR #2644, opened 2024-06-17
doc: Migrate from Annotations to Attributes

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | doc
| BC Break     | no
| Fixed issues | [PHPORM-193](https://jira.mongodb.org/browse/PHPORM-193)

#### Summary

Doctrine annotations are deprecated, I think we can remove any reference of them from the doc and rely on previous versions for projects that are still using them.

I translated every annotation example into attributes, but not tested yet. I don't have the knowledge to find what need to be modified specifically.


docs/en/cookbook/resolve-target-document-listener.rst
```diff
@@ -70,16 +68,11 @@ An Invoice document:
     use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
     use Acme\InvoiceModule\Model\InvoiceSubjectInterface;
 
-    /**
-     * @ODM\Document
-     */
+    #[ODM\Document]
```
GromNaN, line 71: The doc is not consistent on using the `ODM` namespace alias for attributes. Putting it everywhere could solve the problem of missing `use` statements.

docs/en/reference/indexes.rst
```diff
@@ -161,31 +159,27 @@ you can specify them on the class doc block:
             </document>
         </doctrine-mongo-mapping>
 
-To specify multiple indexes you must use the ``@Indexes``
-annotation:
```
GromNaN, line 165: Updating this example as `#[Indexes]` is deprecated.

docs/en/sidebar.rst
```diff
@@ -30,7 +30,7 @@
        reference/trees
        reference/storing-files-with-gridfs
        reference/xml-mapping
-       reference/annotations-reference
+       reference/attributes-reference
```
GromNaN: This file is renamed and the reference of annotations is removed. Relying on previous doc versions for annotations.

docs/en/reference/migrating-schemas.rst
```diff
@@ -157,25 +159,26 @@ Later on, you may want to migrate this data into an embedded Address document:
         }
     }
 
-    /** @Document @HasLifecycleCallbacks */
+    #[Document]
+    #[HasLifecycleCallbacks]
```
GromNaN, line 163: I decided not combining attributes for more readability, even if they are [allowed by PER Coding Style 2.0](https://www.php-fig.org/per/coding-style/#123-compound-attributes).

docs/en/reference/basic-mapping.rst
```diff
@@ -29,49 +28,17 @@ document mapping metadata:
...
-own set of docblock annotations for supplying object-relational mapping
-metadata.
-
-.. note::
-
-    If you're not comfortable with the concept of docblock
-    annotations, don't worry, as mentioned earlier Doctrine provides
-    XML alternative and you could easily implement your own
-    favorite mechanism for defining ODM metadata.
+@TODO intoduce PHP attributes
```
GromNaN: I need to write something that explains what are PHP attributes.

docs/en/cookbook/soft-delete-extension.rst
```diff
@@ -200,7 +198,7 @@ You just need to setup an event listener like the following:
             $sdm = $args->getSoftDeleteManager();
             $document = $args->getDocument();
             if ($document instanceof User) {
-                $sdm->deleteBy(Post:class, ['user.id' => $document->getId()]);
+                $sdm->deleteBy(Post::class, ['user.id' => $document->getId()]);
```
GromNaN, line 201: Fixed syntax error

docs/en/reference/aggregation-stage-reference.rst
```diff
@@ -377,14 +377,17 @@ pipeline stages. Take the following relationship for example:
...
-     * @ReferenceMany(
-     *     targetDocument=Documents\Item::class,
-     *     cascade="all",
-     *     storeAs="id"
-     * )
-     */
-    private $items;
+    namespace Documents;
+
+    class Orders
```
GromNaN, line 382: Adding the wrapping class so that the code snippet is valid PHP, and can be linted.

docs/en/reference/basic-mapping.rst
```diff
@@ -359,10 +339,10 @@ as an option for the ``CUSTOM`` strategy:
 
         <?php
 
-        /** Document */
+        #[Document]
         class MyPersistentClass
         {
-            /** @Id(strategy="CUSTOM", type="string", options={"class"="Vendor\Specific\Generator"}) */
+            #[Id(strategy: 'CUSTOM', type: 'string', options: ['class' => 'Vendor\\Specific\\Generator'])]
```
GromNaN: Maybe switch to the FQCN

```suggestion
            #[Id(strategy: 'CUSTOM', type: 'string', options: ['class' => \Vendor\Specific\Generator::class])]
```

docs/en/reference/indexes.rst
```diff
@@ -75,15 +76,14 @@ Unique Index
...
+        namespace Documents;#[Document */]
         class User
         {
-            /** @Id */
+            #[Id]
             public $id;
 
-            /** @Field(type="string") @Index(unique=true, order="asc") */
+            #[Field(type: 'string')]
+            #[Index(unique:true, order: 'asc')]
```
GromNaN:
```suggestion
            #[Index(unique: true, order: 'asc')]
```

docs/en/cookbook/resolve-target-document-listener.rst
SenseException: I wouldn't call it a problem but having less `use` code lines by using the ODM alias in code examples could increase the readability while still having valid code.

docs/en/reference/indexes.rst
```diff
@@ -301,26 +297,24 @@ options structures manually:
...
+            #[EmbedOne(targetDocument: Coordinates::class)]
             public $coordinates;
         }
 
-        /** @EmbeddedDocument */
+        #[EmbeddedDocument]
         class Coordinates
         {
-            /** @Field(type="float") */
+           #[Field(type: 'float')]
```
alcaeus:
```suggestion
            #[Field(type: 'float')]
```

docs/en/reference/indexes.rst
```diff
@@ -301,26 +297,24 @@ options structures manually:
...
             public $latitude;
 
-            /** @Field(type="float") */
+           #[Field(type: 'float')]
```
alcaeus:
```suggestion
            #[Field(type: 'float')]
```

docs/en/reference/storing-files-with-gridfs.rst
```diff
@@ -219,7 +219,7 @@ uploading:
     $stream = fopen('tmp/path/to/copy', 'w+');
     try {
         $repository->downloadToStream($file->getId(), $stream);
-    finally {
+    } finally {
```
alcaeus, line 222: Nice catch 👍

alcaeus (approved):
Reviewing this is no pleasure, but at first glance it looked fine. Found a couple of nits, but nothing significant 👍

GromNaN:
> Reviewing this is no pleasure, but at first glance it looked fine.

I'm sorry, I've held back and only changed the annotations and a few syntax errors.

docs/en/cookbook/resolve-target-document-listener.rst
GromNaN: The problem I'm thinking of is using attributes without importing namespace. Even though the reader should be able to find it easily because all the classes are in the same namespace, he may be in doubt. Example: https://github.com/doctrine/mongodb-odm/blob/277731cdbe1f0bdd650e9ca3651e6c2d269d721a/docs/en/cookbook/blending-orm-and-mongodb-odm.rst#define-product

docs/en/sidebar.rst
GromNaN: Is there a way to create redirections in the docs? I’d like to redirect
```
https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/latest/reference/annotations-reference.html
```
to
```
https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/latest/reference/attributes-reference.html
```

alcaeus:
> I'm sorry, I've held back and only changed the annotations and a few syntax errors.

Sorry, this shouldn't have come across as criticism. I struggle reviewing repetitive changes, so I needed to do this in multiple steps. The work itself was great and long overdue 👍

docs/en/cookbook/validation-of-documents.rst
```diff
@@ -146,8 +150,8 @@ insertions and updates through a schema associated to the collection
 (cf. `MongoDB documentation <https://docs.mongodb.com/manual/core/schema-validation/>`_).
 
 Doctrine MongoDB ODM now provides a way to take advantage of this functionality
-thanks to the new :doc:`@Validation <../reference/annotations-reference#validation>`
-annotation and its properties (also available with XML mapping):
+thanks to the new :doc:`#[Validation] <../reference/attributes-reference#validation>`
+attribute and its properties (also available with XML mapping):
```
GromNaN: I think this should be reworded to remove the "now" and "new" that are time relative.

docs/en/sidebar.rst
GromNaN: File renamed back to `reference/annotations-reference` because redirections are not possible. It's best not to create a 404 on such an important page.

Issue #2640, opened 2024-05-22
Add CI workflow to check documentation


jmikola:
While writing docs for #2630, I was looking for some way to validate RST files within the PR. I asked about this in Slack (see: [thread](https://doctrine.slack.com/archives/GERFT2W5T/p1716228434425089)) and @greg0ire pointed me to a workflow in the ORM repository that utilizes [phpDocumentor/guides-cli](https://github.com/phpDocumentor/guides-cli): https://github.com/doctrine/orm/blob/3.1.x/.github/workflows/documentation.yml

This issue can serve as a tracking ticket to look into that and see if we can adapt it for the ODM repo.

GromNaN:
Fixed by #2646

PR #2646, opened 2024-06-19
doc: Validate documentation rst with phpdocumentor

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #2640

#### Summary

Use phpdocumentator cli to validate rst files.

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -29,16 +29,11 @@ jobs:
           coverage: "none"
           php-version: "8.3"
 
-      - name: "Remove existing composer file"
-        run: "rm composer.json"
-
-      - name: "Require phpdocumentor/guides-cli"
-        run: "composer require --dev phpdocumentor/guides-cli --no-update"
-
       - name: "Install dependencies with Composer"
         uses: "ramsey/composer-install@v3"
         with:
+          working-directory: "docs"
           dependency-versions: "highest"
 
       - name: "Run guides-cli"
-        run: "vendor/bin/guides -vvv --no-progress docs/en 2>&1 | grep -v 'No template found for rendering directive' | ( ! grep WARNING )"
+        run: "docs/vendor/bin/guides -vvv --no-progress docs/en 2>&1 | grep -v 'No template found for rendering directive' | ( ! grep WARNING )"
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -35,5 +35,9 @@ jobs:
           working-directory: "docs"
           dependency-versions: "highest"
 
+      - name: "Add orphan metadata where needed"
+        run: |
+          printf '%s\n\n%s\n' ":orphan:" "$(cat docs/en/sidebar.rst)" > docs/en/sidebar.rst
+
       - name: "Run guides-cli"
         run: "docs/vendor/bin/guides -vvv --no-progress docs/en 2>&1 | grep -v 'No template found for rendering directive' | ( ! grep WARNING )"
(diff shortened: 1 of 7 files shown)


docs/en/cookbook/implementing-array-access-for-domain-objects.rst
```diff
@@ -1,7 +1,7 @@
 Implementing ArrayAccess for Domain Objects
 ===========================================
 
-.. sectionauthor:: Roman Borschel (roman@code-factory.org)
+.. sectionauthor:: Roman Borschel <roman@code-factory.org>
```
GromNaN, line 4: > WARNING: Content of `.. sectionauthor:: name <email>` must specify a name and can also specify an email

docs/en/sidebar.rst
```diff
@@ -1,3 +1,4 @@
+:orphan:
```
GromNaN: > WARNING: Document "sidebar" isn't included in any toctree. Include it in a `.. toctree::` directive or add `:orphan:` in the first line of the rst document {"rst-file":"sidebar.rst"} []

docs/en/reference/custom-mapping-types.rst
```diff
@@ -94,3 +94,6 @@ type in your mapping like this:
     .. code-block:: xml
 
         <field field-name="field" type="mytype" />
+
+.. |FQCN| raw:: html
+  <abbr title="Fully-Qualified Class Name">FQCN</abbr>
```
GromNaN, line 99: > WARNING: No replacement was found for variable |FQCN| {"rst-file":"reference/custom-mapping-types.rst"}

.github/workflows/documentation.yml
```diff
@@ -0,0 +1,44 @@
...
+        uses: "shivammathur/setup-php@v2"
+        with:
+          coverage: "none"
+          php-version: "8.3"
+
+      - name: "Remove existing composer file"
+        run: "rm composer.json"
+
+      - name: "Require phpdocumentor/guides-cli"
+        run: "composer require --dev phpdocumentor/guides-cli --no-update"
```
Steveb-p: You can use a different working directory for composer instead, which would make the step of removal composer.json from the main directory unnecessary.

GromNaN: Good idea, updated. And this is easier to run the command locally.

docs/en/sidebar.rst
jmikola: I noted that the [ORM's documentation workflow](https://github.com/doctrine/orm/blob/3.1.x/.github/workflows/documentation.yml#L43) manually added `:orphan:` before validating. Is there a particular reason that approach was preferred there instead of modifying the RST directly as you've done here?

I don't have any objections to this, but I wonder if it'd be worth suggesting upstream to the ORM team.

.github/workflows/documentation.yml
```diff
@@ -0,0 +1,39 @@
+name: "Documentation"
```
jmikola, line 1: Noted that this was adapated from the [ORM action](https://github.com/doctrine/orm/blob/3.1.x/.github/workflows/documentation.yml).

jmikola (approved):

greg0ire:
@jmikola `orphan` is not supported by `doctrine/rst-parser`, which we still use for generating our docs

alcaeus (approved):

docs/en/sidebar.rst
greg0ire: `orphan` is not supported by `doctrine/rst-parser`, which we still use for generating our docs

This means that if you don't use the trick I used, you are going to end up with `:orphan:` being printed above the sidebar of your docs :sweat_smile:

GromNaN: Thanks for the clarification, I've adopted your solution that adds `:orphan:` in the job.

PR #2647, opened 2024-06-19
doc: Review introduction

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | doc
| BC Break     | no
| Fixed issues | -

### DIFF
--- a/docs/en/reference/introduction.rst
+++ b/docs/en/reference/introduction.rst
@@ -4,10 +4,10 @@ Introduction
 What is Doctrine MongoDB ODM?
 -----------------------------
 
-Doctrine MongoDB ODM (Object Document Mapper) is a PHP 8.1+ library that provides
-an abstraction layer to work with `MongoDB`_ documents in PHP applications.
-It allows developers to map PHP objects to MongoDB documents, enabling an
-intuitive and structured approach to handling data within a MongoDB database.
+Doctrine MongoDB ODM (Object Document Mapper) is an object modeling library for
+PHP applications. It allows developers to map PHP objects to `MongoDB`_
+documents, enabling an intuitive and structured approach to handling data within
+a MongoDB database.
 
 Features Overview
 -----------------
@@ -172,13 +172,13 @@ The above would insert the following documents into MongoDB collections:
 
     // Project collection
     {
-        _id: ObjectId("..2"),
+        _id: ObjectId("6674375a140216f6730b2263"),
         name: "New Project"
     }
 
     // Manager collection
     {
-        _id: ObjectId("..3"),
+        _id: ObjectId("6674375a140216f6730b2264"),
         changes: 0,
         notes: [],
         name: "Manager",
@@ -194,7 +194,7 @@ The above would insert the following documents into MongoDB collections:
 
     // Employee collection
     {
-        _id: ObjectId("..1"),
+        _id: ObjectId("6674375a140216f6730b2262"),
         changes: 0,
         notes: [],
         name: "Employee",
@@ -224,7 +224,7 @@ efficient update query using the atomic operators:
     $dm->persist($newProject);
     $dm->flush();
 
-The above could would produce an update to Manager's collection that looks
+The above code would produce an update to Manager's collection that looks
 something like this:
 
 ::
@@ -245,7 +245,7 @@ something like this:
                 $each: [
                     {
                         $ref: "Project",
-                        $id: ObjectId("..5")
+                        $id: ObjectId("6674375a140216f6730b2265")
                     }
                 ]
             }
--- a/docs/en/reference/introduction.rst
+++ b/docs/en/reference/introduction.rst
@@ -42,126 +42,90 @@ the features.
 
     <?php
 
+    use DateTimeImmutable;
     use Doctrine\Common\Collections\ArrayCollection;
+    use Doctrine\Common\Collections\Collection;
     use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
-    use DateTime;
 
     #[ODM\MappedSuperclass]
     abstract class BaseEmployee
     {
         #[ODM\Id]
-        private $id;
+        public string $id;
 
         #[ODM\Field(type: 'int', strategy: 'increment')]
-        private $changes = 0;
+        public int $changes = 0;
 
+        /** @var string[] */
         #[ODM\Field(type: 'collection')]
-        private $notes = [];
+        public array $notes = [];
 
         #[ODM\Field(type: 'string')]
-        private $name;
+        public string $name;
 
         #[ODM\Field(type: 'int')]
-        private $salary;
+        public int $salary;
 
         #[ODM\Field(type: 'date')]
-        private $started;
+        public DateTimeImmutable $started;
 
         #[ODM\Field(type: 'date')]
-        private $left;
+        public DateTimeImmutable $left;
 
         #[ODM\EmbedOne(targetDocument: Address::class)]
-        private $address;
-
-        public function getId(): ?string { return $this->id; }
-
-        public function getChanges(): int { return $this->changes; }
-        public function incrementChanges(): void { $this->changes++; }
-
-        public function getNotes(): array { return $this->notes; }
-        public function addNote($note) { $this->notes[] = $note; }
-
-        public function getName(): ?string { return $this->name; }
-        public function setName(string $name): void { $this->name = $name; }
-
-        public function getSalary(): ?int { return $this->salary; }
-        public function setSalary(int $salary): void { $this->salary = $salary; }
-
-        public function getStarted(): ?DateTime { return $this->started; }
-        public function setStarted(DateTime $started) { $this->started = $started; }
-
-        public function getLeft(): ?DateTime { return $this->left; }
-        public function setLeft(DateTime $left) { $this->left = $left; }
-
-        public function getAddress(): ?Address { return $this->address; }
-        public function setAddress(Address $address): void { $this->address = $address; }
+        public Address $address;
     }
 
     #[ODM\Document]
     class Employee extends BaseEmployee
     {
         #[ODM\ReferenceOne(targetDocument: Manager::class)]
-        private $manager;
-
-        public function getManager(): ?Manager { return $this->manager; }
-        public function setManager(Manager $manager): void { $this->manager = $manager; }
+        public ?Manager $manager = null;
     }
 
     #[ODM\Document]
     class Manager extends BaseEmployee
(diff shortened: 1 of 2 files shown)


docs/en/reference/introduction.rst
```diff
@@ -499,7 +464,6 @@ handle the results. If you need the client elsewhere with a different typeMap,
 please create separate clients for your application and ODM.
 
 .. _MongoDB: https://www.mongodb.com/
-.. _10gen: http://www.10gen.com
```
GromNaN, line 502: The company have been renamed to "MongoDB" ([source](https://www.mongodb.com/company/newsroom/press-releases/10gen-announces-solutions-name-change-mongodb-inc)), no need to mention 10gen.

docs/en/reference/introduction.rst
```diff
@@ -431,8 +398,6 @@ more paths) and register the annotations for the driver:
 
     $config->setMetadataDriverImpl(AnnotationDriver::create(__DIR__ . '/Documents'));
 
-    require_once('path/to/vendor/autoload.php');
```
GromNaN: The autoloader is already loaded, and it makes no sense to load it after the preceding code that uses Doctrine classes.

docs/en/reference/introduction.rst
```diff
@@ -179,83 +196,55 @@ Doctrine:
...
-                        [state] => TN
-                        [zipcode] => 37209
-                    )
-
-            )
-
-    )
+    // Project collection
+    {
+        _id: ObjectId("..2"),
```
GromNaN: Reformatting using JS syntax. More concise and easier to read than `var_dump` outputs. That matches with MongoDB's documentation.

docs/en/reference/introduction.rst
```diff
@@ -266,58 +255,36 @@ efficient update query using the atomic operators:
     $newProject = new Project('Another Project');
     $manager->setSalary(200000);
     $manager->addNote('Gave user 100k a year raise');
-    $manager->incrementChanges(2);
```
GromNaN, line 269: The method is declared without argument.

docs/en/reference/introduction.rst
```diff
@@ -1,24 +1,42 @@
...
+   simplifies the creation of simple and complex queries.
+-  **Aggregation Framework**: Supports MongoDB's powerful aggregation framework
+   for advanced data processing and analysis.
+-  **Repositories**: Offers a repository pattern for encapsulating data access
+   logic and promotes code reusability and separation of concerns.
+-  **Events System**: Leverages an events system that allows you to hook into
+   various stages of the document lifecycle for custom behavior.
+-  **GridFS Support**: Stores large files and binary data in GridFS buckets.
+-  **Doctrine ORM Integration**: Seamlessly integrates with Doctrine ORM to
+   allow you to map your ORM entities to the ODM, and use different databases.
```
GromNaN: I've tried to cover all the features in a generic way.

docs/en/reference/introduction.rst
```diff
@@ -266,58 +255,36 @@ efficient update query using the atomic operators:
...
+                    {
+                        $ref: "Project",
+                        $id: ObjectId("..5")
+                    }
+                ]
+            }
+        },
+        $set: {
+            notes: [
+                "Gave user 100k a year raise"
```
GromNaN: Contrary to what was in the example previously, the note is not in `$push`, but the complete list is in `$set`.

And the `$push` on "projects" is done in a distinct command.

docs/en/reference/introduction.rst
```diff
@@ -1,24 +1,42 @@
 Introduction
 ============
 
-Doctrine MongoDB Object Document Mapper is built for PHP 5.3.0+ and
-provides transparent persistence for PHP objects to the popular `MongoDB`_ database by `10gen`_.
```
alcaeus, line 5: Oh, so that's how long we haven't touched this. Wow.

docs/en/reference/introduction.rst
```diff
@@ -1,24 +1,42 @@
...
+-  **Query Builder**: Provides a fluent and flexible query builder that
+   simplifies the creation of simple and complex queries.
+-  **Aggregation Framework**: Supports MongoDB's powerful aggregation framework
+   for advanced data processing and analysis.
+-  **Repositories**: Offers a repository pattern for encapsulating data access
+   logic and promotes code reusability and separation of concerns.
+-  **Events System**: Leverages an events system that allows you to hook into
+   various stages of the document lifecycle for custom behavior.
+-  **GridFS Support**: Stores large files and binary data in GridFS buckets.
+-  **Doctrine ORM Integration**: Seamlessly integrates with Doctrine ORM to
```
alcaeus: The word "seamless" is doing a lot of heavy lifting here. Fine to leave as-is, but I wish it was easier to integrate the two.

docs/en/reference/introduction.rst
alcaeus: `$set` is probably due to the `notes` field using the `atomicSet` strategy, which is why the change is done in the first query. `pushAll` and `addToSet` strategies leverage an additional query for each field to avoid conflicts (e.g. because you can't modify `foo.<index>.bar` when also modifying `foo`.

docs/en/reference/introduction.rst
```diff
@@ -67,7 +85,7 @@ Here is a quick example of some PHP object documents that demonstrates a few of
         public function setName(string $name): void { $this->name = $name; }
 
         public function getSalary(): ?int { return $this->salary; }
-        public function setSalary(int $salary): void { $this->salary = (int) $salary; }
+        public function setSalary(int $salary): void { $this->salary = $salary; }
```
alcaeus: Not sure how much rewriting you want to do here, but we could do away with all the needless setters and getters by using public typed properties.

The only exception is the `id` property, as the concept of ID generators is not compatible with using `readonly` properties that need initialisation in the constructor.

Writing this comment, I also noted that the `getNotes` method defines a wrong return type: after persisting the document or loading it from the DocumentManager, the `notes` field will contain an instance of `PersistentCollection`, clashing with the `getNotes` return type.

docs/en/reference/introduction.rst
GromNaN: Can I modify the class attributes to use the `addToSet` strategy for notes?

docs/en/reference/introduction.rst
GromNaN: I'll update the code after rebasing on #2644

docs/en/reference/introduction.rst
alcaeus: I think `pushAll` is more appropriate here, but I see no reason why we couldn't change it.

GromNaN: > Uncaught Doctrine\ODM\MongoDB\Mapping\MappingException: Invalid strategy pushAll used in Documents\Employee::notes with type collection

From `ClassMetadata::applyStorageStrategy`: https://github.com/doctrine/mongodb-odm/blob/517fb9a36e15df322a76e7e39f806ffc45f6b1ed/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php#L1558-L1612

alcaeus: Ah, for some reason I missed that it was a collection type and had thought it was an embedded document. In that case, it's very much expected that it uses `$set`, and also doesn't support any storage strategies. That also renders my previous comment about `getNotes` and type mismatches moot.

docs/en/reference/introduction.rst
GromNaN: Updated all the classes to use public properties and CPP when accurate, demonstrating modern PHP.
Should I note that using public properties is no longer an issue with [property accessors](https://wiki.php.net/rfc/property_accessors)?

alcaeus: We can already mention them, yes. For libraries, moving from public properties to private properties with accessors is quite the BC break, which is why many people have still used accessors. However, in an application I'd argue that there are much fewer concerns about BC breaks, and it's a case of refactoring effort. However, property accessors indeed remove that last possibility where an accessor method made sense. The only remaining case is the identifier (which can't be readonly), for which we'd need asymmetric visibility.

docs/en/reference/introduction.rst
```diff
@@ -1,149 +1,131 @@
 Introduction
 ============
 
-Doctrine MongoDB Object Document Mapper is built for PHP 5.3.0+ and
-provides transparent persistence for PHP objects to the popular `MongoDB`_ database by `10gen`_.
+What is Doctrine MongoDB ODM?
+-----------------------------
+
+Doctrine MongoDB ODM (Object Document Mapper) is a PHP 8.1+ library that provides
```
jmikola: Would it make sense to remove mention of the PHP version requirement? `composer.json` should be the canonical source for that.

docs/en/reference/introduction.rst
```diff
@@ -158,166 +140,117 @@ Doctrine:
...
+    $manager->notes[] = 'Gave user 100k a year raise';
+    $manager->changes++;
+    $manager->projects->add($newProject);
 
     $dm->persist($newProject);
     $dm->flush();
 
-The above could would produce an update that looks something like
-this:
+The above could would produce an update to Manager's collection that looks
```
jmikola:
```suggestion
The above code would produce an update to Manager's collection that looks
```

Not sure if "code" is what that word should have been, but you can also remove the word entirely.

docs/en/reference/introduction.rst
```diff
@@ -158,166 +140,117 @@ Doctrine:
...
+    }
 
-    )
+    {
+        $push: {
+            projects: {
+                $each: [
+                    {
+                        $ref: "Project",
+                        $id: ObjectId("..5")
```
jmikola: I noticed the ObjectId examples above are all using `..#` as an example string. Would it be more clear to use actual placeholder values here?

I'm also not sure if this should correspond to the same example value used in the Project object's earlier example.

GromNaN: It's a new project ID. I updated to use real object id values.

docs/en/reference/introduction.rst
GromNaN: Updated to remove the PHP version. I rewrote the paragraph.

jmikola (approved):
The Psalm errors on `XmlDriver.php` look unrelated. I suppose those keep popping up (I've seen it before on other files) because CI keeps pulling in newer releases of the tooling.

GromNaN:
Rebased. The psalm error is fixed by https://github.com/doctrine/mongodb-odm/pull/2648

PR #2650, opened 2024-06-21
doc: Review Getting Started

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | doc
| BC Break     | no
| Fixed issues | -

### DIFF
--- a/docs/en/tutorials/getting-started.rst
+++ b/docs/en/tutorials/getting-started.rst
@@ -84,6 +84,8 @@ You can provide your mapping information in Annotations or XML:
         <?php
 
         use DateTimeImmutable;
+        use Doctrine\Common\Collections\ArrayCollection;
+        use Doctrine\Common\Collections\Collection;
         use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
 
         #[ODM\Document]
@@ -100,7 +102,7 @@ You can provide your mapping information in Annotations or XML:
                 public string $email = '',
 
                 #[ODM\ReferenceMany(targetDocument: BlogPost::class, cascade: 'all')]
-                public ArrayCollection $posts = new ArrayCollection(),
+                public Collection $posts = new ArrayCollection(),
             ) {
             }
 
@@ -120,7 +122,7 @@ You can provide your mapping information in Annotations or XML:
                 #[ODM\Field(type: 'string')]
                 public string $body = '',
 
-                #[ODM\Field(type: 'date')]
+                #[ODM\Field(type: 'date_immutable')]
                 public DateTimeImmutable $createdAt = new DateTimeImmutable(),
             ) {
             }
@@ -230,9 +232,9 @@ Here is how you would use your models now:
     Note that you do not need to explicitly call persist on the ``$post`` because the operation
     will cascade on to the reference automatically.
 
-After running this code, you should have those two objects stored in MongoDB in
-the collections "User" and "BlogPost". You can use the `MongoDB Compass`_
-to inspect the contents of your database, where you will see this documents:
+After running this code, you should have those two objects stored in the
+collections "User" and "BlogPost". You can use `MongoDB Compass`_ to inspect
+the contents of your database, where you will see this documents:
 
 ::
 
@@ -283,11 +285,17 @@ If you want to iterate over the posts the user references it is as easy as the f
         echo $post->title;
     }
 
+.. note::
+
+    When retrieved from the database, ``$user->posts`` is an instance of
+    ``Doctrine\ODM\MongoDB\PersistentCollection``. This class lazy-loads the
+    referenced documents from the database when you access them. It keeps track
+    of changes to the collection and will automatically update the database when
+    you call ``flush()``.
+
 You will notice that working with objects is nothing magical and you only have
-access to the properties and methods that you have defined yourself so the
-semantics are very clear. You can continue reading about the MongoDB in the
-:doc:`Introduction to MongoDB Object Document Mapper <../reference/introduction>`.
+access to the properties and methods that you have defined yourself. You can
+continue reading :doc:`Introduction to MongoDB Object Document Mapper <../reference/introduction>`.
 
 .. _MongoDB Compass: https://www.mongodb.com/products/tools/compass
 .. _ObjectId: https://www.php.net/manual/en/class.mongodb-bson-objectid.php
-
--- a/docs/en/tutorials/getting-started.rst
+++ b/docs/en/tutorials/getting-started.rst
@@ -12,16 +12,18 @@ overview of what could be accomplished using Doctrine MongoDB ODM.
 Example Model: Simple Blog
 --------------------------
 
-To create the simplest example, let’s assume the following in a simple blog web application:
+To create the simplest example, let’s assume the following in a simple blog web
+application:
 
--  Blog has a user.
--  Blog user can make blog posts
+-  a Blog has a "user".
+-  a Blog "user" can make "blog posts"
 
 A first prototype
 -----------------
 
-For the above mentioned example, something as simple as this could be modeled with plain PHP classes.
-First define the ``User`` document:
+For the above mentioned example, we start by defining two simple PHP classes:
+``User`` and ``BlogPost``. The ``User`` class will have a collection of
+``BlogPost`` objects.
 
 .. code-block:: php
 
@@ -31,14 +33,18 @@ First define the ``User`` document:
 
     class User
     {
-        private $name;
-        private $email;
-        private $posts = [];
+        public function __construct(
+            public string $name = '',
+            public string $email = '',
+            public array $posts = [],
+        ) {
+        }
 
         // ...
     }
 
-Now define the ``BlogPost`` document:
+Now define the ``BlogPost`` document that contains the title, body and the date
+of creation:
 
 .. code-block:: php
 
@@ -46,20 +52,28 @@ Now define the ``BlogPost`` document:
 
     namespace Documents;
 
+    use DateTimeImmutable;
+
     class BlogPost
     {
-        private $title;
-        private $body;
-        private $createdAt;
+        public function __construct(
+            public string $title = '',
+            public string $body = '',
+            public DateTimeImmutable $createdAt = new DateTimeImmutable(),
+        ) {
+        }
 
         // ...
     }
 
+
+
 Persistent Models
 -----------------
 
-To make the above classes persistent, all we need to do is provide Doctrine with some mapping
-information so that it knows how to consume the objects and persist them to the database.
+To make the above classes persistent, all we need to do is provide Doctrine with
+some mapping information so that it knows how to consume the objects and persist
+them to the database.
 
 You can provide your mapping information in Annotations or XML:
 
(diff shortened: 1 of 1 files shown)


docs/en/tutorials/getting-started.rst
```diff
@@ -263,12 +279,15 @@ If you want to iterate over the posts the user references it is as easy as the f
 
     <?php
 
-    $posts = $dm->getPosts();
```
GromNaN, line 266: That was an error. `$user->getPosts()` expected.

docs/en/tutorials/getting-started.rst
```diff
@@ -199,44 +230,29 @@ Here is how you would use your models now:
     Note that you do not need to explicitly call persist on the ``$post`` because the operation
     will cascade on to the reference automatically.
 
-Now if you did everything correctly, you should have those two objects
-stored in MongoDB in correct collections and databases. You can use the
-`php-mongodb-admin project, hosted on github`_ to look at your
```
GromNaN, line 204: This was a PHP UI for MongoDB, but [archived](https://github.com/jwage/php-mongodb-admin). Replaced with Compass.

docs/en/tutorials/getting-started.rst
```diff
@@ -151,16 +180,16 @@ instance. Read more about setting up the Doctrine MongoDB ODM in the
...
+    $config->setProxyDir(__DIR__ . '/generated/proxies');
     $config->setProxyNamespace('Proxies');
-    $config->setHydratorDir('/path/to/generate/hydrators');
+    $config->setHydratorDir(__DIR__ . '/generated/hydrators');
     $config->setHydratorNamespace('Hydrators');
-    $config->setMetadataDriverImpl(AttributeDriver::create('/path/to/document/classes'));
+    $config->setMetadataDriverImpl(AttributeDriver::create(__DIR__ . '/src'));
 
-    $dm = DocumentManager::create(null, $config);
+    $dm = DocumentManager::create(config: $config);
```
GromNaN, line 194: Using named arguments here. More expressive.

docs/en/tutorials/getting-started.rst
```diff
@@ -263,12 +279,15 @@ If you want to iterate over the posts the user references it is as easy as the f
...
+        echo $post->title;
     }
 
-You will notice that working with objects is nothing magical and you only have access to the properties,
-getters and setters that you have defined yourself so the semantics are very clear. You can continue
-reading about the MongoDB in the :doc:`Introduction to MongoDB Object Document Mapper <../reference/introduction>`.
+You will notice that working with objects is nothing magical and you only have
+access to the properties and methods that you have defined yourself so the
+semantics are very clear. You can continue reading about the MongoDB in the
+:doc:`Introduction to MongoDB Object Document Mapper <../reference/introduction>`.
```
GromNaN: I find this page very similar to the introduction. It shows how to instantiate the DocumentManager and create classes using mapping.

docs/en/tutorials/getting-started.rst
```diff
@@ -176,20 +205,22 @@ Here is how you would use your models now:
...
     $dm->persist($user);
 
     // create blog post
-    $post = new BlogPost();
-    $post->setTitle('My First Blog Post');
-    $post->setBody('MongoDB + Doctrine 2 ODM = awesomeness!');
-    $post->setCreatedAt(new DateTime());
+    $post = new BlogPost(
+        title: 'My First Blog Post',
+        body: 'MongoDB + Doctrine 2 ODM = awesomeness!',
```
alcaeus:
```suggestion
        body: 'MongoDB + Doctrine ODM = awesomeness!',
```

alcaeus (approved):
Small nit in the example, but looking good other than that 👍

docs/en/tutorials/getting-started.rst
```diff
@@ -31,35 +33,47 @@ First define the ``User`` document:
 
     class User
     {
-        private $name;
-        private $email;
-        private $posts = [];
+        public function __construct(
+            public string $name = '',
+            public string $email = '',
+            public array $posts = [],
```
GromNaN, line 39: I think the modeling is done the wrong direction. It's the post that should refer to the user, not the user that contains the list of posts.

docs/en/tutorials/getting-started.rst
```diff
@@ -68,40 +82,48 @@ You can provide your mapping information in Annotations or XML:
...
 
-            #[ODM\Field(type: 'string')]
-            private $email;
+                #[ODM\Field(type: 'string')]
+                public string $email = '',
 
-            #[ODM\ReferenceMany(targetDocument: BlogPost::class, cascade: 'all')]
-            private $posts = [];
+                #[ODM\ReferenceMany(targetDocument: BlogPost::class, cascade: 'all')]
+                public ArrayCollection $posts = new ArrayCollection(),
```
jmikola: Initializing as a ArrayCollection instance is fine, but I think you'd run into trouble typing the property as ArrayCollection. When the model is hydrated from a database result, ODM will use a PersistentCollection. There should be a Collection interface available to address this, and it may be something worth calling out in a note.

docs/en/tutorials/getting-started.rst
jmikola: I agree, although this may have been intentional for a "Getting Started" document so that inverse relationships could be introduced at a later point: https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/2.7/reference/bidirectional-references.html#owning-and-inverse-sides

docs/en/tutorials/getting-started.rst
```diff
@@ -31,35 +33,47 @@ First define the ``User`` document:
...
     }
 
+
+
 Persistent Models
 -----------------
 
-To make the above classes persistent, all we need to do is provide Doctrine with some mapping
-information so that it knows how to consume the objects and persist them to the database.
+To make the above classes persistent, all we need to do is provide Doctrine with
```
jmikola:
```suggestion
To make the above classes persistent, we need to provide Doctrine with
```

MongoDB's own docs have a style guide to avoid subjective language like this. It also comes across as filler, so I have no qualms about removing it.

docs/en/tutorials/getting-started.rst
```diff
@@ -31,35 +33,47 @@ First define the ``User`` document:
...
+
     class BlogPost
     {
-        private $title;
-        private $body;
-        private $createdAt;
+        public function __construct(
+            public string $title = '',
+            public string $body = '',
+            public DateTimeImmutable $createdAt = new DateTimeImmutable(),
```
jmikola, line 62: Does ODM utilize DateTimeImmutable when hydrating models from a database result, or would we need to use DateTimeInterface here?

docs/en/tutorials/getting-started.rst
```diff
@@ -199,44 +230,29 @@ Here is how you would use your models now:
     Note that you do not need to explicitly call persist on the ``$post`` because the operation
     will cascade on to the reference automatically.
 
-Now if you did everything correctly, you should have those two objects
-stored in MongoDB in correct collections and databases. You can use the
-`php-mongodb-admin project, hosted on github`_ to look at your
-``BlogPost`` collection, where you will see only one document:
+After running this code, you should have those two objects stored in MongoDB in
+the collections "User" and "BlogPost". You can use the `MongoDB Compass`_
```
jmikola:
```suggestion
the collections "User" and "BlogPost". You can use `MongoDB Compass`_
```

Unless you were writing "the MongoDB Compass tool", you can omit "the" when referring to it as a proper noun.

docs/en/tutorials/getting-started.rst
```diff
@@ -199,44 +230,29 @@ Here is how you would use your models now:
     Note that you do not need to explicitly call persist on the ``$post`` because the operation
     will cascade on to the reference automatically.
 
-Now if you did everything correctly, you should have those two objects
-stored in MongoDB in correct collections and databases. You can use the
-`php-mongodb-admin project, hosted on github`_ to look at your
-``BlogPost`` collection, where you will see only one document:
+After running this code, you should have those two objects stored in MongoDB in
```
jmikola: > stored in the "User" and "BlogPost" collections.

I think we can omit "in MongoDB" to avoid repetition of "in".

Also, would the collection names end up being `users` and `blogposts`? I forget whether ODM uses inflection to create the namespaces for each model.

docs/en/tutorials/getting-started.rst
```diff
@@ -263,12 +279,15 @@ If you want to iterate over the posts the user references it is as easy as the f
 
     <?php
 
-    $posts = $dm->getPosts();
-    foreach ($posts as $post) {
+    foreach ($user->posts as $post) {
```
jmikola, line 284: This is where I'd expect `posts` to be a PersistentCollection. It may be too much information to mention that here (or earlier when you use a Collection interface), though, so I won't advocate for a note.

docs/en/tutorials/getting-started.rst
```diff
@@ -263,12 +279,15 @@ If you want to iterate over the posts the user references it is as easy as the f
...
+        echo $post->title;
     }
 
-You will notice that working with objects is nothing magical and you only have access to the properties,
-getters and setters that you have defined yourself so the semantics are very clear. You can continue
-reading about the MongoDB in the :doc:`Introduction to MongoDB Object Document Mapper <../reference/introduction>`.
+You will notice that working with objects is nothing magical and you only have
+access to the properties and methods that you have defined yourself so the
+semantics are very clear. You can continue reading about the MongoDB in the
```
jmikola: "so the semantics are very clear" can probably be removed, as it makes the sentence run on a bit.

I'd also advocate for removing "the MongoDB" and maybe just rewriting the second sentence entirely:

> You can continue reading [LINK]

docs/en/tutorials/getting-started.rst
jmikola: Noted. That sounds like something worth revisiting after the docs are cleared up. Ultimately, it may be something our docs team can assist with, as they did for Laravel.

docs/en/tutorials/getting-started.rst
GromNaN: You're right, I need to type with `Collection`.

docs/en/tutorials/getting-started.rst
GromNaN: Yes, the `date` type is automatically switched to `date_immutable`.

docs/en/tutorials/getting-started.rst
GromNaN: Adding a note that explains the behavior of `PersistentCollection`.

SenseException (approved):

docs/en/tutorials/getting-started.rst
jmikola: @GromNaN: I think "all we need to do" can still be removed.

docs/en/tutorials/getting-started.rst
```diff
@@ -138,6 +162,13 @@ You can provide your mapping information in Annotations or XML:
           </document>
         </doctrine-mongo-mapping>
 
+.. note::
+
+   The `$id` property is a special property that is used to store the unique
```
jmikola: Nit: technically, the property mapped with `#[ODM\Id]` is special. The property name itself can be anything. If you want to address this, I'd say something like:

> The `$id` property above is annotated with `#[ODM\Id]`, which makes it special...

Here, I'm using "annotated" as a verb. My word choice has nothing to do with attributes vs. annotations :)

jmikola (approved):
LGTM but one of my earlier suggestions was overlooked.

PR #2651, opened 2024-06-21
doc: Move annotations-reference to attributes-reference

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | doc
| BC Break     | no
| Fixed issues | -

#### Summary

Rename the page that contains the reference to attributes and add migration instructions on the old page.


.github/workflows/documentation.yml
```diff
@@ -38,6 +38,7 @@ jobs:
       - name: "Add orphan metadata where needed"
         run: |
           printf '%s\n\n%s\n' ":orphan:" "$(cat docs/en/sidebar.rst)" > docs/en/sidebar.rst
+          printf '%s\n\n%s\n' ":orphan:" "$(cat docs/en/reference/attributes-reference.rst)" > docs/en/reference/attributes-reference.rst
```
jmikola: Is `:orphan:` actually required given this is referenced in `sidebar.rst` and `validation-of-documents.rst`?

jmikola (approved):
Question about whether `:orphan:` is required but no objections to the content changes.

.github/workflows/documentation.yml
GromNaN: Oups, that's the wrong file.
```suggestion
          printf '%s\n\n%s\n' ":orphan:" "$(cat docs/en/reference/attributes-reference.rst)" > docs/en/reference/annotations-reference.rst
```

PR #2652, opened 2024-06-24
doc: Add type to properties

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | doc
| BC Break     | no
| Fixed issues | -

#### Summary

Using types properties is recommended to improve type safety.

I still have to check that everything works properly if the properties are not set with a default value. Because when there's no type, the value is `null` by default.

### DIFF
--- a/docs/en/reference/aggregation-stage-reference.rst
+++ b/docs/en/reference/aggregation-stage-reference.rst
@@ -413,13 +413,12 @@ to be considered when looking up one-to-one relationships:
 
     class Orders
     {
-        /** @var Collection<Item> */
         #[ReferenceOne(
             targetDocument: Item::class,
             cascade: 'all',
             storeAs: 'id',
         )]
-        private Collection $items;
+        private Item $items;
     }
 
 .. code-block:: php
--- a/docs/en/reference/aggregation-stage-reference.rst
+++ b/docs/en/reference/aggregation-stage-reference.rst
@@ -386,7 +386,7 @@ pipeline stages. Take the following relationship for example:
             cascade: 'all',
             storeAs: 'id',
         )]
-        private $items;
+        private Collection $items;
     }
 
 .. code-block:: php
@@ -417,7 +417,7 @@ to be considered when looking up one-to-one relationships:
             cascade: 'all',
             storeAs: 'id',
         )]
-        private $items;
+        private Collection $items;
     }
 
 .. code-block:: php
--- a/docs/en/reference/aggregation-stage-reference.rst
+++ b/docs/en/reference/aggregation-stage-reference.rst
@@ -381,6 +381,7 @@ pipeline stages. Take the following relationship for example:
 
     class Orders
     {
+        /** @var Collection<Item> */
         #[ReferenceMany(
             targetDocument: Item::class,
             cascade: 'all',
@@ -412,6 +413,7 @@ to be considered when looking up one-to-one relationships:
 
     class Orders
     {
+        /** @var Collection<Item> */
         #[ReferenceOne(
             targetDocument: Item::class,
             cascade: 'all',
--- a/docs/en/reference/attributes-reference.rst
+++ b/docs/en/reference/attributes-reference.rst
@@ -18,7 +18,7 @@ does not exist.
     {
         #[Field(type: 'string')]
         #[AlsoLoad('name')]
-        public $fullName;
+        public string $fullName;
     }
 
 The ``$fullName`` property will be loaded from ``fullName`` if it exists, but
@@ -226,7 +226,7 @@ Optional attributes:
             ],
             defaultDiscriminatorValue: 'book',
         )]
-        private $tags = [];
+        private Collection $tags;
     }
 
 Depending on the embedded document's class, a value of ``user`` or ``author``
@@ -281,7 +281,7 @@ Optional attributes:
              ],
              defaultDiscriminatorValue: 'user',
         )]
-        private $creator;
+        private User|Author $creator;
     }
 
 Depending on the embedded document's class, a value of ``user`` or ``author``
@@ -304,7 +304,7 @@ relationship.
     class Money
     {
         #[Field(type: 'float')]
-        private $amount;
+        private float $amount;
 
         public function __construct(float $amount)
         {
@@ -317,7 +317,7 @@ relationship.
     class Wallet
     {
         #[EmbedOne(targetDocument: Money::class)]
-        private $money;
+        private Money $money;
 
         public function setMoney(Money $money): void
         {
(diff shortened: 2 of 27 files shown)


docs/en/reference/attributes-reference.rst
```diff
@@ -281,7 +281,7 @@ Optional attributes:
              ],
              defaultDiscriminatorValue: 'user',
         )]
-        private $creator;
+        private User|Author $creator;
```
GromNaN, line 285: I hope I didn't make a mistake. I find that the type helps to understand the content of the variable here.

docs/en/reference/aggregation-stage-reference.rst
```diff
@@ -417,7 +417,7 @@ to be considered when looking up one-to-one relationships:
             cascade: 'all',
             storeAs: 'id',
         )]
-        private $items;
+        private Collection $items;
```
GromNaN: This is a `ReferenceOne`. The variable is plural so I supposed it's a `Collection`. But I would have expected a single `Item`. I'll check.

docs/en/reference/best-practices.rst
```diff
@@ -56,19 +56,23 @@ Example:
...
             $this->articles = new ArrayCollection();
         }
     }
+
+.. note::
+
+   The properties' type hints must be ``Collection``, and cannot be
+   ``ArrayCollection``. When the ``User`` object is retrieved from the database,
+   the properties ``$addresses`` and ``$articles`` are instances of
+   ``Doctrine\ODM\MongoDB\PersistentCollection`` to track changes.
```
GromNaN, line 78: `PersistentCollection` is not internal, so I think we can mention this class.

docs/en/reference/change-tracking-policies.rst
```diff
@@ -85,11 +85,11 @@ follows:
     {
         // ...
 
-        private $_listeners = [];
```
GromNaN, line 88: Prefixing private properties with an underscore is so PHP 4.

docs/en/reference/custom-mapping-types.rst
```diff
@@ -31,7 +31,7 @@ class:
         public function convertToPHPValue($value): \DateTime
         {
             // This is called to convert a Mongo value to a PHP representation
-            return new \DateTime('@' . $value->sec);
+            return $value->toDateTime();
```
GromNaN, line 34: It's so much simpler.

docs/en/reference/attributes-reference.rst
```diff
@@ -501,9 +502,16 @@ customize this via the :ref:`strategy <basic_mapping_identifiers>` attribute.
     class User
     {
         #[Id]
-        protected $id;
+        protected string $id;
     }
 
+.. note::
```
alcaeus, line 508: Good to have this documented 👍

docs/en/reference/attributes-reference.rst
alcaeus: This looks correct to me based on the discriminator map.

IonBazan (approved):

docs/en/reference/aggregation-stage-reference.rst
alcaeus: This should be:
```suggestion
        private Item $item;
```

docs/en/reference/change-tracking-policies.rst
```diff
@@ -85,11 +85,11 @@ follows:
...
+        private array $listeners = [];
```
alcaeus, line 89:
```suggestion
        /** @var array<PropertyChangedListener> */
        private array $listeners = [];
```

or alternatively as a list:
```suggestion
        /** @var list<PropertyChangedListener> */
        private array $listeners = [];
```

docs/en/reference/custom-mapping-types.rst
alcaeus: Unrelated, but we should try to come up with a different example, as this only reproduces the `DateTime` type. I've previously used the example of storing the time zone alongside the `UTCDateTime` instance and applying that to the date type when reading.

docs/en/reference/query-builder-api.rst
```diff
@@ -589,13 +589,13 @@ document with a text index:
...
-            public $id;
+            public string $id;
 
             #[Field(type: 'string')]
-            public $description;
+            public string $description;
 
             #[Field(notSaved: true)]
-            public $score;
+            public int $score;
```
alcaeus, line 598: The type of the field defaults to string, but I agree that the property name indicates an integer. With that in mind, we should add `type: 'int'` to the field attribute.

docs/en/reference/reference-mapping.rst
```diff
@@ -94,8 +94,9 @@ Reference many documents:
         {
             // ...
 
+            /** @var Collection<Account> */
             #[ReferenceMany(targetDocument: Account::class)]
-            private $accounts = [];
+            private Collection $accounts;
```
alcaeus, line 99: This may apply to multiple classes, but we should default to initialising this with an empty `ArrayCollection` instance so it doesn't have to happen in the constructor:

```suggestion
            private Collection $accounts = new ArrayCollection();
```

docs/en/reference/change-tracking-policies.rst
GromNaN: I've a preference for this syntax:
```suggestion
        /** @var PropertyChangedListener[] */
        private array $listeners = [];
```

docs/en/reference/reference-mapping.rst
GromNaN: That's not allowed unless CPP is used. Adding the constructor.

docs/en/reference/custom-mapping-types.rst
GromNaN: Done: https://github.com/doctrine/mongodb-odm/pull/2654

docs/en/reference/reference-mapping.rst
SenseException: Yep, this is still PHP :grin:

SenseException (approved):

docs/en/reference/introduction.rst
```diff
@@ -42,7 +42,7 @@ the features.
 
     <?php
 
-    use DateTimeImmutable;
+    use DateTime, DateTimeImmutable;
```
alcaeus:
```suggestion
    use DateTime;
    use DateTimeImmutable;
```

docs/en/reference/reference-mapping.rst
alcaeus: OMFG - I could've sworn that you can initialise properties like that in general now. Boo :(

alcaeus (approved):
One minor change to namespaces, but LGTM aside from that

PR #2654, opened 2024-06-25
doc: Update custom mapping example

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | doc
| BC Break     | no
| Fixed issues | -

#### Summary

The previous example was a simplified copy of the `date` type. In order to present something more useful, the new example is inspired by the [codecs tutorial](https://www.mongodb.com/docs/php-library/current/tutorial/codecs/).

Also initializing tests on documentation examples.

### DIFF
--- a/docs/en/reference/custom-mapping-types.rst
+++ b/docs/en/reference/custom-mapping-types.rst
@@ -7,8 +7,12 @@ to replace the existing implementation of a mapping type.
 
 In order to create a new mapping type you need to subclass
 ``Doctrine\ODM\MongoDB\Types\Type`` and implement/override
-the methods. Here is an example skeleton of such a custom type
-class:
+the methods.
+
+The following example defines a custom type that stores ``DateTimeInterface``
+instances as an embedded document containing a BSON date and accompanying
+timezone string. Those same embedded documents are then be translated back into
+a ``DateTimeImmutable`` when the data is read from the database.
 
 .. code-block:: php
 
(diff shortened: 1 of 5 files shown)


docs/en/reference/custom-mapping-types.rst
```diff
@@ -7,45 +7,54 @@ to replace the existing implementation of a mapping type.
...
+            $timeZone = new DateTimeZone($value['tz']);
+            $dateTime = $value['utc']
+                ->toDateTime()
+                ->setTimeZone($timeZone);
+
+            return DateTimeImmutable::createFromMutable($dateTime);
         }
 
-        public function convertToDatabaseValue($value): UTCDateTime
+        public function convertToDatabaseValue($value): ?array
```
SenseException:
```suggestion
        public function convertToDatabaseValue($value): array
```

docs/en/reference/custom-mapping-types.rst
```diff
@@ -7,45 +7,54 @@ to replace the existing implementation of a mapping type.
...
-     * My custom datatype.
-     */
-    class MyType extends Type
+    class DateTimeWithTimezoneType extends Type
     {
         // This trait provides default closureToPHP used during data hydration
         use ClosureToPHP;
 
-        public function convertToPHPValue($value): \DateTime
+        public function convertToPHPValue($value): ?DateTimeImmutable
```
SenseException:
```suggestion
        public function convertToPHPValue($value): DateTimeImmutable
```

tests/Documentation/CustomMapping/DateTimeWithTimezoneType.php
```diff
@@ -0,0 +1,35 @@
...
+    {
+        $timeZone = new DateTimeZone($value['tz']);
+        $dateTime = $value['utc']
+            ->toDateTime()
+            ->setTimeZone($timeZone);
+
+        return DateTimeImmutable::createFromMutable($dateTime);
+    }
+
+    public function convertToDatabaseValue($value): ?array
```
SenseException: Do the methods have to be nullable for a reason?

SenseException (changes requested):

tests/Documentation/CustomMapping/DateTimeWithTimezoneType.php
GromNaN: Thank you for spotting this error. Indeed, no need to manage `null` values.

docs/en/reference/custom-mapping-types.rst
```diff
@@ -7,45 +7,54 @@ to replace the existing implementation of a mapping type.
...
     {
         // This trait provides default closureToPHP used during data hydration
         use ClosureToPHP;
 
-        public function convertToPHPValue($value): \DateTime
+        public function convertToPHPValue($value): DateTimeImmutable
         {
-            // This is called to convert a Mongo value to a PHP representation
-            return new \DateTime('@' . $value->sec);
+            $timeZone = new DateTimeZone($value['tz']);
```
alcaeus, line 36: Just a suggestion, but this could benefit from some value checks:
```php
if (! is_array($value) || ! isset($value['utc'], $value['tz'])) {
    throw new Exception(); // TODO: throw correct exception class
}
```

The end result will be the same, but I feel like there can be more useful information in the exception message rather than letting PHP throw its type errors. Not sure if there's prior art for this in our type, so feel free to disregard and leave this as an exercise to the user.

alcaeus (approved):
LGTM with a minor suggestion for checks

docs/en/reference/custom-mapping-types.rst
GromNaN: Absolutely, it's easy to be sure of a property's value, especially with typing, but there can always be surprises with database content, especially when there's no schema (which is certainly very common).

SenseException:
@GromNaN Please update your [commit messages](https://www.doctrine-project.org/contribute/index.html#crafting-meaningful-commit-messages) to more meaningful ones. Don't repeat messages and don't use just one word.

GromNaN:
> Please update your [commit messages](https://www.doctrine-project.org/contribute/index.html#crafting-meaningful-commit-messages) to more meaningful ones. Don't repeat messages and don't use just one word.

The commit history is not very meaningful. I squash and update the unique commit message before merging. This commits are just a raw history of my changes and fixes that help when reviewing a second time the PR.

GromNaN:
@SenseException I squashed the commits in this PR.

SenseException (approved):

PR #2655, opened 2024-06-25
doc: Add cookbook about embedding referenced documents using $lookup

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | doc
| BC Break     | no
| Fixed issues | -

#### Summary

On reviewing the aggregation stages reference, and in particular `$lookup`, @alcaeus and I decided that we needed more detailed documentation on this specific use case.

### DIFF
--- a/docs/en/cookbook/lookup-reference.rst
+++ b/docs/en/cookbook/lookup-reference.rst
@@ -1,11 +1,21 @@
 Loading references with Lookup
 ==============================
 
-When you have a reference to a document, Doctrine ODM can load the referenced
-by running a distinct query. This is called "lazy loading". However, when you
-need to load a reference with a query, you can use the ``$lookup`` stage in
-MongoDB's aggregation pipeline. It's similar to a SQL join, without duplication
-of data in the result set.
+Doctrine ODM provides a way to load reference documents from other collections
+using the ``#[ReferenceOne]`` and ``#[ReferenceMany]`` annotations. This is
+perfect to keep independent document updates and avoid data duplication. But
+sometimes you need to load the referenced documents with the main document in a
+single query. This is where MongoDB's aggregation pipeline and the ``$lookup``
+stage come into play.
+
+By default, referenced documents are loaded with a separate query. When you
+access them. This is called "lazy loading". However, when you know you will
+need the referenced documents, you can use the ``$lookup`` stage in MongoDB's
+aggregation pipeline. It's similar to a SQL join, without duplication of data in
+the result set when there is many references to load.
+
+Introduction of the example
+---------------------------
 
 For this cookbook, we will use 3 collections:
 - ``users``: contains the users who can pass orders
@@ -156,6 +166,9 @@ without stage, you will get the following documents with reference ids for
         ]
     ];
 
+Embed a list of referenced documents
+------------------------------------
+
 Now, let's see how to load items with each order using an aggregation pipeline.
 MongoDB's ``$lookup`` stage requires a local field and a foreign field to match
 documents. In our case, this parameters are extracted automatically from the
@@ -216,10 +229,12 @@ documents.
         ]
     ];
 
+Embed a single referenced document
+----------------------------------
+
 To get the user, you can also use the ``$lookup`` stage. It will always return a
 list of documents. You need to add the ``$unwind`` stage to reduce to a single
 document.
-
 
 .. code-block:: php
 
@@ -270,6 +285,9 @@ document.
         ]
     ];
 
+Combine multiple lookups
+------------------------
+
 Both ``$lookup`` stages can be combined in a single pipeline to get the full
 order document, with user and items.
 
@@ -283,9 +301,6 @@ order document, with user and items.
             ->lookup('user')
                 ->alias('user')
             ->unwind('$user');
-
-.. code-block:: php
-
 
 .. code-block:: php
 
@@ -342,8 +357,12 @@ order document, with user and items.
 The result is still an array. You may be tempted to hydrate the result into the
 ``Order`` class, but this will fail because the ``items`` and ``user`` fields
 contains embedded documents instead of reference ids as expected by the
-``ReferenceMany`` and ``ReferenceOne`` mappings. You need to create a new class
-to hold the result of the aggregation.
+``ReferenceMany`` and ``ReferenceOne`` mappings.
+
+Hydrate the result into a custom class
+--------------------------------------
+
+You need to create a new class to hold the result of the aggregation.
 
 .. code-block:: php
 
@@ -438,6 +457,9 @@ as an array of ``OrderResult`` instances.
 Perfect, now you know how to load references with the ``$lookup`` and hydrate
 the result into a custom class as embedded documents.
 
+Embed relations from another collection
+---------------------------------------
+
 Let's see how to embed relations in the inverse way: load users with their
 orders. Remember, it's the "order" documents that have a reference to the user.
 We now wish to load the users first and use ``$lookup`` to load the list of
@@ -488,6 +510,9 @@ list of order documents.
             ],
         ]
     ]
+
+Embed 2 levels of references in a single query
+----------------------------------------------
 
 It becomes more complex when you want to load the items of each order. You need
 to ``$unwind`` all the orders in separate results, then ``$lookup`` the items
@@ -672,3 +697,6 @@ will return the result as an array of ``UserResult`` instances.
             ],
         ]
     ]
+
+That's it! You now know how to embed references with the ``$lookup`` stage and
+hydrate the result into custom classes.
--- a/docs/en/cookbook/lookup-reference.rst
+++ b/docs/en/cookbook/lookup-reference.rst
@@ -141,28 +141,28 @@ without stage, you will get the following documents with reference ids for
 
     [
         [
-            '_id' => new MongoDB\BSON\ObjectId('667b034c75590cbbe601061c'),
-            'date' => new MongoDB\BSON\UTCDateTime('-147398400000'),
+            '_id' => MongoDB\BSON\ObjectId('667b034c75590cbbe601061c'),
+            'date' => MongoDB\BSON\UTCDateTime('-147398400000'),
             'items' => [
-                new MongoDB\BSON\ObjectId('667b034c75590cbbe6010613')
+                MongoDB\BSON\ObjectId('667b034c75590cbbe6010613')
             ],
-            'user' => new MongoDB\BSON\ObjectId('667b034c75590cbbe6010619'),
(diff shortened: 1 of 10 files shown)


docs/en/cookbook/lookup-reference.rst
```diff
@@ -0,0 +1,674 @@
...
+        public function __construct()
+        {
+            $this->date  = new DateTimeImmutable();
+            $this->items = new ArrayCollection();
+        }
+    }
+
+In order to make tests, you can import the following documents:
+
+.. code-block:: php
```
GromNaN, line 95: I will check how to collapse this less interesting code.

docs/en/cookbook/lookup-reference.rst
```diff
@@ -0,0 +1,674 @@
...
+            'date' => new MongoDB\BSON\UTCDateTime('399686400000'),
+            'items' => [
+                new MongoDB\BSON\ObjectId('667b034c75590cbbe6010617'),
+                new MongoDB\BSON\ObjectId('667b034c75590cbbe6010613'),
+                new MongoDB\BSON\ObjectId('667b034c75590cbbe6010615'),
+            ],
+            'user' => new MongoDB\BSON\ObjectId('667b034c75590cbbe6010618'),
+        ]
+    ];
+
```
GromNaN, line 168: Need to add titles.

docs/en/cookbook/lookup-reference.rst
```diff
@@ -0,0 +1,702 @@
...
+            ]),
+            user: new User(
+                id: '667b034c75590cbbe6010619',
+                name: 'Eddy Merckx',
+            ),
+        ),
+        new OrderResult(
+            id: new MongoDB\BSON\ObjectId('667b034c75590cbbe601061b'),
+            date: new DateTimeImmutable('1974-07-01'),
+            items: new Doctrine\ODM\MongoDB\PersistentCollection([]),
```
GromNaN: I know we can't instantiate a `PersistentCollection` like this, but I find is a lot more legible than a `var_export` or `var_dump`. Maybe I can remove the `new` keyword so that it's clear that's not real PHP code.

docs/en/cookbook/lookup-reference.rst
```diff
@@ -0,0 +1,702 @@
...
+You need to create a new class to hold the result of the aggregation.
+
+.. code-block:: php
+
+    <?php
+
+    #[QueryResultDocument]
+    class OrderResult
+    {
+        public function __construct(
```
GromNaN: I used CPP so that the constructor can be user in result view. But I would prefer using public properties only. We don't need to initialize collection properties for `QueryResultDocument` (something I should note).

docs/en/cookbook/lookup-reference.rst
GromNaN: Done in https://github.com/doctrine/mongodb-odm/pull/2655/commits/2bee655d1682f07c0daead2cb7da5a0e4f833db1

docs/en/cookbook/lookup-reference.rst
GromNaN: Done in https://github.com/doctrine/mongodb-odm/pull/2655/commits/2bee655d1682f07c0daead2cb7da5a0e4f833db1

docs/en/cookbook/lookup-reference.rst
```diff
@@ -0,0 +1,707 @@
...
+without stage, you will get the following documents with reference ids for
+``user`` and ``items``.
+
+.. code-block:: php
+
+    <?php
+
+    [
+        [
+            '_id' => MongoDB\BSON\ObjectId('667b034c75590cbbe601061c'),
```
GromNaN, line 144: Not using the `new` keyboard to make is clear that's data dump and not actual instantiation of objects.

docs/en/cookbook/lookup-reference.rst
```diff
@@ -0,0 +1,707 @@
...
+.. code-block:: php
+
+    <?php
+
+    $aggregation = $this->dm->createAggregationBuilder(User::class)
+        ->sort('name', 'asc')
+        ->lookup('Order')
+            ->alias('orders')
+            ->localField('_id')
+            ->foreignField('user');
```
GromNaN, line 485: Here we have to set the collection name because the `User` class doesn't have an `orders` property mapping.

docs/en/cookbook/lookup-reference.rst
```diff
@@ -0,0 +1,707 @@
...
+By default, referenced documents are loaded with a separate query. When you
+access them. This is called "lazy loading". However, when you know you will
+need the referenced documents, you can use the ``$lookup`` stage in MongoDB's
+aggregation pipeline. It's similar to a SQL join, without duplication of data in
+the result set when there is many references to load.
+
+Introduction of the example
+---------------------------
+
+For this cookbook, we will use 3 collections:
```
alcaeus:
```suggestion
For this example, we will use 3 collections:
```

docs/en/cookbook/lookup-reference.rst
```diff
@@ -0,0 +1,707 @@
...
+stage come into play.
+
+By default, referenced documents are loaded with a separate query. When you
+access them. This is called "lazy loading". However, when you know you will
+need the referenced documents, you can use the ``$lookup`` stage in MongoDB's
+aggregation pipeline. It's similar to a SQL join, without duplication of data in
+the result set when there is many references to load.
+
+Introduction of the example
+---------------------------
```
alcaeus:
```suggestion
Example setup
-------------
```

docs/en/cookbook/lookup-reference.rst
```diff
@@ -0,0 +1,707 @@
...
+
+For this cookbook, we will use 3 collections:
+- ``users``: contains the users who can pass orders
+- ``orders`` are affected to a user and contain a list of ``items``
+- ``items``: contains the products
+
+The document classes ``User`` and ``Item`` contains only an id and a name.
```
alcaeus:
```suggestion
The document classes ``User`` and ``Item`` contain only an id and a name.
```

docs/en/cookbook/lookup-reference.rst
```diff
@@ -0,0 +1,707 @@
...
+            public string $name,
+        ) {
+        }
+    }
+
+    #[Document]
+    class Item
+    {
+        #[Id]
+        public ?string $id = null;
```
alcaeus: Nit: is there a reason why identifiers are nullable here but not in the `User` and Order documents?

docs/en/cookbook/lookup-reference.rst
```diff
@@ -0,0 +1,707 @@
...
+
+        public function __construct(
+            #[Field(type: 'string')]
+            public ?string $name = null,
+        ) {
+        }
+    }
+
+The ``Order`` class as references to one ``User`` and a list of ``Item``, an id
```
alcaeus:
```suggestion
The ``Order`` class has references to one ``User``, a list of ``Item``, an id,
```

docs/en/cookbook/lookup-reference.rst
```diff
@@ -0,0 +1,707 @@
...
+            'user' => MongoDB\BSON\ObjectId('667b034c75590cbbe6010618'),
+        ]
+    ];
+
+Embed a list of referenced documents
+------------------------------------
+
+Now, let's see how to load items with each order using an aggregation pipeline.
+MongoDB's ``$lookup`` stage requires a local field and a foreign field to match
+documents. In our case, this parameters are extracted automatically from the
```
alcaeus:
```suggestion
documents. In our case, these parameters are extracted automatically from the
```

docs/en/cookbook/lookup-reference.rst
```diff
@@ -0,0 +1,707 @@
...
+``#[ReferenceMany]`` mapping. The alias is the name of the field in the
+resulting document.
```
alcaeus:
```suggestion
resulting document. In this case, we replace the original list of references
with a list of ``Item`` documents.
```

docs/en/cookbook/lookup-reference.rst
```diff
@@ -0,0 +1,707 @@
...
+
+.. code-block:: php
+
+    <?php
+
+        $aggregation = $this->dm->createAggregationBuilder(Order::class)
+            ->lookup('items')
+                ->alias('items');
+
+The result is a list of "order" document, each one containing a list of "item"
```
alcaeus:
```suggestion
The result is a list of ``Order`` documents, each one containing a list of ``Item``
```

docs/en/cookbook/lookup-reference.rst
```diff
@@ -0,0 +1,707 @@
...
+        #[EmbedOne(targetDocument: User::class)]
+        public User $user;
+    }
+
+.. note::
+
+    You don't need to initialize the collections in the constructor, as the
+    ``QueryResultDocument`` are only used to hydrate the results from the
+    database. Doctrine will create the collections. You will never instantiate
+    this class directly.
```
alcaeus:
```suggestion
    ``QueryResultDocument`` is only used to hydrate the results from the
    database and you should never instantiate this class directly.
```

docs/en/cookbook/lookup-reference.rst
```diff
@@ -0,0 +1,707 @@
...
+                [
+                    '_id' => MongoDB\BSON\ObjectId('667b034c75590cbbe601061c'),
+                    // all other fields
+                ]
+            ],
+        ]
+    ]
+
+Embed 2 levels of references in a single query
+----------------------------------------------
```
alcaeus: Numbers up to ten are usually written out:

```suggestion
Embed two levels of references in a single query
------------------------------------------------
```

docs/en/cookbook/lookup-reference.rst
```diff
@@ -0,0 +1,707 @@
...
+                            'name' => 'Handlebars',
+                        ]
+                    ],
+                ]
+            ],
+        ]
+    ]
+
+The last challenge is to hydrate the result into a custom class. You need to
+create 2 classes: one for the user ``UserResult`` that can embed an ``order``
```
alcaeus:
```suggestion
create two classes: one for the user ``UserResult`` that can embed an ``order``
```

alcaeus (commented):
A few suggestions for improvements, but reads good so far!

docs/en/cookbook/lookup-reference.rst
GromNaN: Because I started with CPP and I don't want to set a value in the constructor. Then figured that it don't have to be a constructor argument.
```suggestion
        public string $id;
```

alcaeus (approved):

PR #2657, opened 2024-06-26
doc: Remove soft-delete-cookbook

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | doc
| BC Break     | no
| Fixed issues | -

#### Summary

The repository have been archived: https://github.com/doctrine/mongodb-odm-softdelete


alcaeus (approved):
Does this count as soft-delete since it's still available in git history? 🤔

GromNaN:
It looks like we can rewrite this cookbook with doctrine extensions: https://github.com/doctrine-extensions/DoctrineExtensions/blob/main/src/SoftDeleteable/Mapping/Event/Adapter/ODM.php

SenseException (approved):
Was this a remnant of something older? I can't find a link that is directing to the removed file. :thinking:

GromNaN:
> Was this a remnant of something older? I can't find a link that is directing to the removed file. 🤔

The sidebar has an automatic listing of all cookbooks: https://github.com/doctrine/mongodb-odm/blob/8b542f4c5b9239a3083a375894c4249e922b315d/docs/en/sidebar.rst?plain=1#L63

PR #2660, opened 2024-06-27
doc: Review and add tests on `ResolveTargetDocumentListener`

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | doc
| BC Break     | no
| Fixed issues | -

#### Summary

- Rewrote the introduction. The problem wasn't clear to me on first reading.
- Modified the order in which the classes are presented, to make the explanations more logical. Highlight the module (namespace) of each class to make the purpose more explicit.
- Added an usage example at the end, so that the result is more obvious.
- Added tests

### DIFF
--- a/docs/en/cookbook/resolve-target-document-listener.rst
+++ b/docs/en/cookbook/resolve-target-document-listener.rst
@@ -13,8 +13,8 @@ when no mapping metadata has been found for the original class name.
 Background
 ----------
 
-In the following example, we have an `InvoiceModule` that provides invoicing
-functionality, and a `CustomerModule` that contains customer management tools.
+In the following example, we have an ``InvoiceModule`` that provides invoicing
+functionality, and a ``CustomerModule`` that contains customer management tools.
 We want to keep these separated, because they can be used in other systems
 without each other; however, we'd like to use them together in our application.
 
@@ -44,7 +44,7 @@ the application:
         #[Id]
         public string $id;
 
-        #[Field(type: 'string')]
+        #[Field]
         public string $name;
     }
 
@@ -64,7 +64,7 @@ An ``Invoice`` document in the ``InvoiceModule``:
         #[Id]
         public string $id;
 
-        #[ReferenceOne(targetDocument: \Acme\InvoiceModule\Model\InvoiceSubjectInterface::class)]
+        #[ReferenceOne]
         public InvoiceSubjectInterface $subject;
     }
 
--- a/docs/en/cookbook/resolve-target-document-listener.rst
+++ b/docs/en/cookbook/resolve-target-document-listener.rst
@@ -1,20 +1,14 @@
 Keeping Your Modules Independent
 ================================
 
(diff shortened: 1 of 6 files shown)


docs/en/cookbook/resolve-target-document-listener.rst
```diff
@@ -1,26 +1,20 @@
...
-``ResolveTargetDocumentListener``, that functions by intercepting certain calls
-inside Doctrine and rewriting ``targetDocument`` parameters in your metadata
-mapping at runtime. This allows your bundle to use an interface or abstract
-class in its mappings while still allowing the mapping to resolve to a concrete
-document class at runtime. It will also rewrite class names when no mapping
-metadata has been found for the original class name.
-
-This functionality allows you to define relationships between different
-documents without creating hard dependencies.
+If you work with independent modules, you may encounter the problem of creating
```
alcaeus, line 4: In case you want to add an example, Sylius uses this concept extensively to allow for customising entity classes.

docs/en/cookbook/resolve-target-document-listener.rst
```diff
@@ -32,79 +26,94 @@ with a real class that implements that interface.
...
-        protected InvoiceSubjectInterface $subject;
+        #[Id]
+        public string $id;
+
+        #[ReferenceOne]
+        public InvoiceSubjectInterface $subject;
     }
 
-An InvoiceSubjectInterface:
+This class has de reference to the ``InvoiceSubjectInterface``. This interface
```
alcaeus, line 71:
```suggestion
This class has a reference to an ``InvoiceSubjectInterface``. This interface
```

alcaeus (approved):
The example we're using may not be the best considering how much code is necessary to explain this, but using these listeners is not necessary in most applications. This tutorial would do well with better examples, but I'm afraid that would be beyond the scope of the improvements we're planning to make. For the time being, the addition of tests and the better explanations are already a big improvement over the status quo.

I did leave a note about mentioning Sylius, as that's probably the most well-known practical usage of what we're explaining here. I'll let you decide whether you want to add that as an example of if you'd like to keep this abstract.

docs/en/cookbook/resolve-target-document-listener.rst
GromNaN: I don't know Sylius to know what to write. I haven't found any reference to `ResolveTargetDocumentListener` or `ResolveTargetEntityListener` in the code or in the doc of Sylius.

docs/en/cookbook/resolve-target-document-listener.rst
GromNaN: https://github.com/doctrine/mongodb-odm/pull/2667

Steveb-p:
> The example we're using may not be the best considering how much code is necessary to explain this, but using these listeners is not necessary in most applications.

In applications yeah. In frameworks, it is really helpful to have this. I am using it often.

PR #2663, opened 2024-06-28
doc: Remove wakeup and clone cookbook

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | doc
| BC Break     | no
| Fixed issues | -

#### Summary

This cookbook was copied [from the ORM](https://github.com/doctrine/orm/blob/a34dc0a0e3fb6352e9645fc473b9b6bac383f11e/docs/en/cookbook/implementing-wakeup-or-clone.rst) but removed because the solution was not compatible with lazy ghost objects https://github.com/doctrine/orm/pull/10187.

The examples are also far too abbreviated to show the use case and the complete solution.


IonBazan (approved):

docs/en/reference/architecture.rst
```diff
@@ -15,11 +15,8 @@ be any regular PHP class observing the following restrictions:
 -  All persistent properties/field of any document class should
    always be private or protected, otherwise lazy-loading might not
    work as expected.
--  A document class must not implement ``__clone`` or
-   :doc:`do so safely <../cookbook/implementing-wakeup-or-clone>`.
--  A document class must not implement ``__wakeup`` or
-   :doc:`do so safely <../cookbook/implementing-wakeup-or-clone>`.
-   Also consider implementing
+-  A document class must not implement ``__clone``.
```
malarzm: Is this still true given we're using `ProxyManager`?

SenseException (approved):

docs/en/reference/architecture.rst
GromNaN: Excellent question, I'll have to test that. I'd like to know in which cases it causes a problem.

GromNaN: Removed this restrictions from the doc. I didn't find any issue with clone or wakeup.

PR #2664, opened 2024-07-01
doc: Add tests for introduction

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | doc
| BC Break     | yes/no
| Fixed issues | -

#### Summary

- Add tests
- Remove explicit field types, to rely on prop types.


alcaeus (approved):
LGTM, but PhpStan doesn't like the `readonly` identifier.

PR #2666, opened 2024-07-01
Add native type to private properties and final classes

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | -

#### Summary

Some private properties were not typed, mainly because `mixed` or union types were not available in previous PHP version.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Fill.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Fill.php
@@ -32,8 +32,7 @@ use function strtolower;
  */
 class Fill extends Stage
 {
-    /** @var mixed|Expr|null */
-    private $partitionBy = null;
+    private mixed $partitionBy = null;
 
     /** @var array<string> */
     private array $partitionByFields = [];
--- a/lib/Doctrine/ODM/MongoDB/Iterator/HydratingIterator.php
+++ b/lib/Doctrine/ODM/MongoDB/Iterator/HydratingIterator.php
@@ -25,7 +25,7 @@ use Traversable;
 final class HydratingIterator implements Iterator
 {
     /** @var Generator<mixed, array<string, mixed>>|null */
-    private $iterator;
+    private ?Generator $iterator;
 
     /**
      * @param Traversable<mixed, array<string, mixed>> $traversable
--- a/lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionTrait.php
+++ b/lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionTrait.php
@@ -41,71 +41,61 @@ trait PersistentCollectionTrait
      *
      * @var array<TKey, T>
      */
-    private $snapshot = [];
+    private array $snapshot = [];
 
     /**
      * Collection's owning document
-     *
-     * @var object|null
      */
-    private $owner;
+    private ?object $owner = null;
 
     /**
-     * @var array|null
+     * @var array<string, mixed>|null
      * @psalm-var FieldMapping|null
      */
-    private $mapping;
+    private ?array $mapping = null;
 
     /**
      * Whether the collection is dirty and needs to be synchronized with the database
      * when the UnitOfWork that manages its persistent state commits.
-     *
-     * @var bool
      */
-    private $isDirty = false;
+    private bool $isDirty = false;
 
     /**
      * Whether the collection has already been initialized.
-     *
-     * @var bool
      */
-    private $initialized = true;
+    private bool $initialized = true;
 
     /**
      * The wrapped Collection instance.
      *
      * @var BaseCollection<TKey, T>
      */
-    private $coll;
+    private BaseCollection $coll;
 
     /**
      * The DocumentManager that manages the persistence of the collection.
-     *
-     * @var DocumentManager|null
      */
-    private $dm;
+    private DocumentManager $dm;
 
     /**
      * The UnitOfWork that manages the persistence of the collection.
-     *
-     * @var UnitOfWork
      */
-    private $uow;
+    private UnitOfWork $uow;
 
     /**
      * The raw mongo data that will be used to initialize this collection.
      *
      * @var mixed[]
      */
-    private $mongoData = [];
+    private array $mongoData = [];
 
     /**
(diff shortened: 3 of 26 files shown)


lib/Doctrine/ODM/MongoDB/Query/ReferencePrimer.php
```diff
@@ -64,23 +56,6 @@ public function __construct(DocumentManager $dm, UnitOfWork $uow)
     {
         $this->dm  = $dm;
         $this->uow = $uow;
-
-        $this->defaultPrimer = static function (DocumentManager $dm, ClassMetadata $class, array $ids, array $hints): void {
```
GromNaN, line 68: Moved to private class method.

lib/Doctrine/ODM/MongoDB/Query/ReferencePrimer.php
```diff
@@ -124,7 +99,7 @@ public function primeReferences(ClassMetadata $class, $documents, string $fieldN
             throw new LogicException(sprintf('Field "%s" is an identifier reference without a target document class in class "%s"', $fieldName, $class->name));
         }
 
-        $primer     = $primer ?: $this->defaultPrimer;
+        $primer     = $primer ?: $this->defaultPrimer(...);
```
GromNaN: First-class closure available since PHP 8.1

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Fill.php
```diff
@@ -32,8 +32,7 @@
  */
 class Fill extends Stage
 {
-    /** @var mixed|Expr|null */
-    private $partitionBy = null;
+    private mixed $partitionBy = null;
```
malarzm, line 35: Dunno how much value it's bringing IRL, but knowing there could be `Expr` makes one remember about it

lib/Doctrine/ODM/MongoDB/Iterator/HydratingIterator.php
```diff
@@ -24,8 +24,8 @@
  */
 final class HydratingIterator implements Iterator
 {
-    /** @var Generator<mixed, array<string, mixed>>|null */
-    private $iterator;
+    /** @var Generator<mixed, array<string, mixed>> */
```
malarzm: Shouldn't `|null` stay here?

lib/Doctrine/ODM/MongoDB/Query/ReferencePrimer.php
```diff
@@ -276,4 +251,26 @@ private function addManyReferences(PersistentCollectionInterface $persistentColl
             $groupedIds[$className][serialize($id)] = $id;
         }
     }
+
+    /**
+     * @param mixed[]           $ids
+     * @param array<int, mixed> $hints
+     */
+    private function defaultPrimer(DocumentManager $dm, ClassMetadata $class, array $ids, array $hints): void
```
malarzm: should this remain static?

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Fill.php
GromNaN: This is reported by phpcs.
```
FILE: lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Fill.php
-----------------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
-----------------------------------------------------------------------------------
 35 | ERROR | [x] Property
    |       |     \Doctrine\ODM\MongoDB\Aggregation\Stage\Fill::$partitionBy has
    |       |     useless @var annotation.
-----------------------------------------------------------------------------------
```

I could try to restrict the types, but I would have to accept `object` which also overlaps `Expr`.

This is a private property, the setter keeps the full documented types:
https://github.com/doctrine/mongodb-odm/blob/99b4183889414e593029557c356a829eaba34f8d/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Fill.php#L53-L54

lib/Doctrine/ODM/MongoDB/Iterator/HydratingIterator.php
GromNaN: Good catch. I tried to make it non-nullable and `unset` it in `__destruct`, but that didn't work as expected.

lib/Doctrine/ODM/MongoDB/Query/ReferencePrimer.php
GromNaN: We can, so we are sure we don't access the `ReferencePrimer` instance.

lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionTrait.php
```diff
@@ -41,71 +41,61 @@ trait PersistentCollectionTrait
...
-    private $coll;
+    private BaseCollection $coll;
 
     /**
      * The DocumentManager that manages the persistence of the collection.
-     *
-     * @var DocumentManager|null
      */
-    private $dm;
+    private ?DocumentManager $dm = null;
```
alcaeus: Not sure how exactly this property is used in this code, but would it make more sense to rely on the `TypeError` if it wasn't properly initialised before access?

lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionTrait.php
```diff
@@ -41,71 +41,61 @@ trait PersistentCollectionTrait
...
 
     /**
      * The UnitOfWork that manages the persistence of the collection.
-     *
-     * @var UnitOfWork
      */
-    private $uow;
+    private ?UnitOfWork $uow = null;
```
alcaeus: Same question as with the `dm` property above

alcaeus (approved):
LGTM with a minor question, but fine to merge as is.

lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionTrait.php
GromNaN: Removed the nullable, it seems to work.

PR #2669, opened 2024-07-06
Ensure compatibility with persistence 4

greg0ire:
This is mostly about adding return type declarations. In one instance though, it is more than that: `ClassMetadataFactory` redeclares `$cacheSalt`, a protected property inherited from the persistence package. Since it is not possible to widen or narrow the type, and since the redeclaration seems to be about setting a default value, let us set it in the constructor.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/DocumentManager.php
+++ b/lib/Doctrine/ODM/MongoDB/DocumentManager.php
@@ -21,6 +21,7 @@ use Doctrine\ODM\MongoDB\Repository\RepositoryFactory;
 use Doctrine\ODM\MongoDB\Repository\ViewRepository;
 use Doctrine\Persistence\Mapping\ProxyClassNameResolver;
 use Doctrine\Persistence\ObjectManager;
+use Doctrine\Persistence\ObjectRepository;
 use InvalidArgumentException;
 use Jean85\PrettyVersions;
 use MongoDB\Client;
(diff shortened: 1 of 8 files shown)


lib/Doctrine/ODM/MongoDB/DocumentManager.php
```diff
@@ -235,16 +232,20 @@ public function getMetadataFactory()
...
+    public function initializeObject($obj): void
     {
         $this->unitOfWork->initializeObject($obj);
     }
 
     /**
      * Helper method to check whether a lazy loading proxy or persistent collection has been initialized.
      */
-    public function isUninitializedObject(object $obj): bool
+    public function isUninitializedObject(mixed $obj): bool
```
malarzm, line 243: How come it's `mixed` in persistence?

alcaeus: Can't find the original discussion, but the argument was that if this wasn't mixed, every check would have to be `is_object($obj) && $dm->isUninitializedObject($obj)`. The idea is that anything that isn't an object also isn't an uninitialised object. There was some pushback to changing the interface, and I honestly didn't care enough to force the issue 🤷

alcaeus (approved):
LGTM, and I also don't see any changes that we'd want to announce as deprecations in 2.x 👍

malarzm:
> LGTM, and I also don't see any changes that we'd want to announce as deprecations in 2.x 👍

There's still `DocumentManager::merge` and `NOTIFY` tracking policy on the hook

lib/Doctrine/ODM/MongoDB/DocumentManager.php
greg0ire: The discussion happened on #proxy-integration, and I cannot invite @malarzm there.

greg0ire: Link to the original discussion: https://doctrine.slack.com/archives/C045LBBNALQ/p1716381659877529

SenseException (approved):

lib/Doctrine/ODM/MongoDB/DocumentManager.php
malarzm: Thanks for digging this up

Issue #2634, opened 2024-05-06
Implement `skip-<feature>` options in CLI commands


jmikola:
### Feature Request

Allow users to explicitly exclude features (e.g. search indexes) in CLI commands.

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

The CLI commands for managing schema entities (e.g. creating collections and indexes) attempt to process _all_ schema definitions by default. This is problematic for features such as Atlas Search indexes, which have more restrictions.

For example, CreateCommand defines an ordered list (i.e. `$createOrder`) for schema entities to process when no command line arguments are specified. If any Atlas Search indexes exist in class metadata, those will be processed by default; however, that may not be supported in a development environment.

If users want to exclude certain DDL operations, they cannot rely on the default order and must manually enumerate all schema entities to process. Consider introducing `skip-<feature>` options to allow users to opt out of specific features.

Context: https://github.com/doctrine/mongodb-odm/pull/2630#discussion_r1588893623

jmikola:
Note: #2630 introduces a `skip-search-indexes` option to UpdateCommand, which already had a `disable-validators` option with the same meaning. If we move forward with this issue, we should rename `disable-validators` to `skip-validators` and deprecate the existing name.

nikophil:
Hello,

I discovered this issue because I've just upgraded to `v2.8.0` of this lib, and had an error `Unrecognized pipeline stage name: '$listSearchIndexes'` when running `UpdateCommand`. I think I have this error because we're using an old mongo version: [`listSearchIndexes` has landed in mongo 6.0.7](https://www.mongodb.com/docs/manual/reference/operator/aggregation/listSearchIndexes/#mongodb-pipeline-pipe.-listSearchIndexes) but still, there are some older versions still maintained which do not support this instruction.

Maybe it would be a nice addition to check if the mongo version is compatible, before using it by default?

thanks!

alcaeus:
@nikophil thanks for bringing up the issue. I've created #2671 to work around the issue gracefully where possible. I'd appreciate it if you could test it and report back in the PR so I can tag a fixed version soon. Thanks!

GromNaN:
Search indexes can be used locally using [Atlas Local](https://www.mongodb.com/docs/atlas/cli/current/atlas-cli-deploy-local/). This feature flag system is not necessary.

PR #2671, opened 2024-07-31
Gracefully handle search index exceptions if possible

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | https://github.com/doctrine/mongodb-odm/issues/2634#issuecomment-2239054724

#### Summary

#2630 introduces support for managing search indexes in the schema manager. However, the logic does not correctly handle cases when the server does not support search indexes. This PR adds more graceful handling in cases where no search index changes are necessary. Here's a list of the changes:

* creating search indexes: no changes; if search indexes are declared on a class and the command to create them fails, we expect an exception to be thrown
* updating search indexes: if no search indexes are declared on a class, failure of the `listSearchIndexes` command will no longer cause an exception. If a class has search indexes, the old behaviour still applies
* deleting search indexes: failure of the `listSearchIndexes` command no longer throws an exception, since there are no search indexes to be removed.

We specifically check for error code 115 (CommandNotSupported) and check the exception message to make sure that it's actually an issue with the search index command, as there are no good ways to check for search index support.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/SchemaManager.php
+++ b/lib/Doctrine/ODM/MongoDB/SchemaManager.php
@@ -1057,6 +1057,13 @@ final class SchemaManager
 
     private function isSearchIndexCommandException(CommandException $e): bool
     {
-        return $e->getCode() === self::CODE_COMMAND_NOT_SUPPORTED && str_contains($e->getMessage(), 'Search index');
+        // MongoDB 6.0.7+ and 7.0+: "Search indexes are only available on Atlas"
+        if ($e->getCode() === self::CODE_COMMAND_NOT_SUPPORTED && str_contains($e->getMessage(), 'Search index')) {
+            return true;
+        }
+
+        // Older server versions don't support $listSearchIndexes
+        // We don't check for an error code here as the code is not documented and we can't rely on it
+        return str_contains($e->getMessage(), 'Unrecognized pipeline stage name: \'$listSearchIndexes\'');
     }
 }
--- a/lib/Doctrine/ODM/MongoDB/SchemaManager.php
+++ b/lib/Doctrine/ODM/MongoDB/SchemaManager.php
@@ -8,6 +8,7 @@ use Doctrine\ODM\MongoDB\Mapping\ClassMetadata;
 use Doctrine\ODM\MongoDB\Mapping\ClassMetadataFactoryInterface;
 use Doctrine\ODM\MongoDB\Repository\ViewRepository;
 use InvalidArgumentException;
+use MongoDB\Driver\Exception\CommandException;
 use MongoDB\Driver\Exception\RuntimeException;
 use MongoDB\Driver\Exception\ServerException;
 use MongoDB\Driver\WriteConcern;
@@ -32,6 +33,7 @@ use function iterator_count;
 use function iterator_to_array;
 use function ksort;
 use function sprintf;
+use function str_contains;
 
 /**
  * @psalm-import-type IndexMapping from ClassMetadata
@@ -44,6 +46,7 @@ final class SchemaManager
     private const GRIDFS_CHUNKS_COLLECTION_INDEX = ['filename' => 1, 'uploadDate' => 1];
 
     private const CODE_SHARDING_ALREADY_INITIALIZED = 23;
+    private const CODE_COMMAND_NOT_SUPPORTED        = 115;
 
     private const ALLOWED_MISSING_INDEX_OPTIONS = [
         'background',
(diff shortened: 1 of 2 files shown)


nikophil:
hi @alcaeus, I would be happy to help, but it seems I cannot test your PR since my app requires `doctrine/mongodb-odm-bundle` which depends on this lib. So when i add your repo as a new vcs repository in `composer.json`, composer does not accept to update to `dev-fix-search-indexes`:
```
  Problem 1
    - Root composer.json requires doctrine/mongodb-odm dev-fix-search-indexes, found doctrine/mongodb-odm[dev-fix-search-indexes] but these were not loaded, likely because it conflicts with another require.
  Problem 2
    - doctrine/mongodb-odm-bundle is locked to version 5.0.1 and an update of this package was not requested.
    - doctrine/mongodb-odm-bundle 5.0.1 requires doctrine/mongodb-odm ^2.6 -> satisfiable by doctrine/mongodb-odm[2.6.0, ..., 2.9.x-dev] from vcs repo (github https://github.com/alcaeus/mongodb-odm) has higher repository priority. The packages from the higher priority repository do not match your constraint and are therefore not installable. That repository is canonical so the lower priority repo's packages are not installable. See https://getcomposer.org/repoprio for details and assistance.
```

alcaeus:
> So when i add your repo as a new vcs repository in `composer.json`, composer does not accept to update to `dev-fix-search-indexes`:

Ah, sorry, it's been a while since I've done this. You can use an [inline alias](https://getcomposer.org/doc/articles/aliases.md#require-inline-alias) to require the package:
`doctrine/mongodb-odm@dev-fix-search-indexes as 2.8.1-dev`

That should fix the issue you encountered.

nikophil:
hey thanks, I just learnt something! :blush: 

Now I have required your PR, I still have the same error: 
```
➜ bin/console doctrine:mongodb:schema:update -n
Updated indexes for all classes
Updated validation for all classes
Unrecognized pipeline stage name: '$listSearchIndexes'
```
with exit code 255

BTW, we're using mongo 4.4

alcaeus:
> BTW, we're using mongo 4.4

Ah, that's probably the reason why you're getting a different message. Thanks for taking a look, I'll update this PR after some more testing.

alcaeus:
@nikophil I've updated the code to also handle server versions where `$listSearchIndexes` doesn't exist (tested with 4.4.29 on my machine). The error message is consistent, so this should hopefully handle all kinds of server versions.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -44,6 +46,7 @@ final class SchemaManager
     private const GRIDFS_CHUNKS_COLLECTION_INDEX = ['filename' => 1, 'uploadDate' => 1];
 
     private const CODE_SHARDING_ALREADY_INITIALIZED = 23;
+    private const CODE_COMMAND_NOT_SUPPORTED        = 115;
```
malarzm, line 49: I guess there are no consts anywhere in drivers? :)

malarzm (approved):

lib/Doctrine/ODM/MongoDB/SchemaManager.php
alcaeus: Nope :)

mustanggb:
With [2.8.1](https://github.com/doctrine/mongodb-odm/releases/tag/2.8.1) this is still causing us a problem in [TbbcMoneyBundle](https://github.com/TheBigBrainsCompany/TbbcMoneyBundle).

I think the circumstances are an index exists (not a search index), then the schema is dropped.

It tries to run a `$listSearchIndexes` pipeline, which errors (255).

We are dropping the schema with `doctrine:mongodb:schema:drop` (which I believe comes from [mongodb-odm-bundle](https://packagist.org/packages/doctrine/mongodb-odm-bundle)).

alcaeus:
@mustanggb I just ran the tests for TbbcMoneyBundle locally. With MongoDB 7.0.9, I was able to reproduce the failures using ODM 2.8.0, but using 2.8.1 fixes the tests. Removing the constraint in [#180](https://github.com/TheBigBrainsCompany/TbbcMoneyBundle/pull/180/files#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34R42) works on my machine - could you please give it another look and share the exception if one happens? I noticed that the command invocation uses a `NullOutput`, which makes debugging much more difficult than it should be.

mustanggb:
Looks like it's saying `$listSearchIndexes stage is only allowed on MongoDB Atlas`.

https://github.com/mustanggb/TbbcMoneyBundle/actions/runs/10579960710/job/29313659438#step:8:1123

HyprUsr:
For me also the error is `$listSearchIndexes stage is only allowed on MongoDB Atlas`

alcaeus:
I've released 2.8.2 with a fix for this. Sorry for the delay here.

mustanggb:
Confirming this is working for us now.

PR #2673, opened 2024-08-29
Gracefully handle search index exceptions on non MongoDB Atlas

### LINKED ISSUE #2674: Inability to list search indexes is not properly detected
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.8.1

#### Summary

Inability to list indexes is not properly detected anymore by the following lines of code:

https://github.com/doctrine/mongodb-odm/blob/2.8.1/lib/Doctrine/ODM/MongoDB/SchemaManager.php#L1060-L1063

#### Current behavior

With MongoDB `7.0.14`, the error message is `$listSearchIndexes stage is only allowed on MongoDB Atlas`.

It does not contain `Search index`, so it's not recognized as unsupported command.

As such, the `odm:schema:update` command fails, even when there are no search indexes defined.

#### How to reproduce

Run the `odm:schema:update` command.

#### Expected behavior

The `odm:schema:update` does not fail when there are no search indexes defined.

HyprUsr:
Allow users to explicitly exclude features (e.g. search indexes) in CLI commands.

|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | https://github.com/doctrine/mongodb-odm/pull/2671#issuecomment-2312782111

#### Summary

The CLI commands for managing schema entities (e.g. creating collections and indexes) attempt to process all schema definitions by default. This is problematic for features such as Atlas Search indexes, which have more restrictions.

For example, CreateCommand defines an ordered list (i.e. $createOrder) for schema entities to process when no command line arguments are specified. If any Atlas Search indexes exist in class metadata, those will be processed by default; however, that may not be supported in a development environment.

If users want to exclude certain DDL operations, they cannot rely on the default order and must manually enumerate all schema entities to process. Consider introducing skip-<feature> options to allow users to opt out of specific features.

Context: https://github.com/doctrine/mongodb-odm/pull/2671#issuecomment-2312782111

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/SchemaManager.php
+++ b/lib/Doctrine/ODM/MongoDB/SchemaManager.php
@@ -1061,6 +1061,8 @@ final class SchemaManager
         if ($e->getCode() === self::CODE_COMMAND_NOT_SUPPORTED && str_contains($e->getMessage(), 'Search index')) {
             return true;
         }
+
+        // MongoDB 6.0.7+ and 7.0+: "$listSearchIndexes stage is only allowed on MongoDB Atlas"
         if ($e->getMessage() === '$listSearchIndexes stage is only allowed on MongoDB Atlas') {
             return true;
         }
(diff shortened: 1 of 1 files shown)


alcaeus (approved):

alcaeus:
Thanks @Jafarili!

Issue #2538, opened 2023-06-15
Changeset with the field date is lost in some edge cases


olivier34000:
### BC Break Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.1

#### Summary

When we have a field with the type date if the field is updated in less than 1 second the update does not work correctly. The field is not updated

#### Previous behavior

In this previous version when the field was a date we compared 2 MongoDate  
https://github.com/doctrine/mongodb-odm/blob/1.3.x/lib/Doctrine/ODM/MongoDB/UnitOfWork.php#L816-L823

In the MongoDate object we have a timestamp and the miliseconds.

In the new version we have https://github.com/doctrine/mongodb-odm/blob/2.6.x/lib/Doctrine/ODM/MongoDB/UnitOfWork.php#L798-L809

The UTCDateTime object contains also the milliseconds but we loose this information in the changeSet.

Is-it possible to only cast the UTCDateTime in string in order to keep this information ?

The code should be 

```php
                // skip equivalent date values
                if (isset($class->fieldMappings[$propName]['type']) && $class->fieldMappings[$propName]['type'] === 'date') {
                    /** @var DateType $dateType */
                    $dateType      = Type::getType('date');
                    $dbOrgValue    = $dateType->convertToDatabaseValue($orgValue);
                    $dbActualValue = $dateType->convertToDatabaseValue($actualValue);

                    $orgMillisec    = $dbOrgValue instanceof UTCDateTime ? (string) $dbOrgValue : null;
                    $actualMillsec = $dbActualValue instanceof UTCDateTime ? (string) $dbActualValue : null;

                    if ($orgMillisec === $actualMillsec) {
                        continue;
                    }
                }
```

malarzm:
According to [docs ](https://www.php.net/manual/en/mongodb-bson-utcdatetime.tostring.php)casting to string will not be enough as it'll get us timestamp again. 

@alcaeus do you remember the reason why we're casting values to anything instead of relying on `$dbOrgValue == $dbActualValue`?

alcaeus:
> @alcaeus do you remember the reason why we're casting values to anything instead of relying on `$dbOrgValue == $dbActualValue`?

Off the top of my head, no.

> The code should be [...]

[...]

PR #2676, opened 2024-09-17
Fixes changeset being empty when datetime change is sub second

magnetik:
|      Q       |   A
|------------- | -----------
| Type         | fix
| BC Break     | no
| Fixed issues | #2538

#### Summary

The ODM was ignoring changes when they are under a second (because of the cast as timestmap).

It might be a BC break because people can rely on the old behavior ?

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
+++ b/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
@@ -835,7 +835,7 @@ final class UnitOfWork implements PropertyChangedListener
                     $dbActualValue = $dateType->convertToDatabaseValue($actualValue);
 
                     // We rely on loose comparison to compare every field (including microseconds)
-                    // phpcs:ignore Squiz.Operators.ComparisonOperatorUsage.NotAllowed
+                    // phpcs:ignore SlevomatCodingStandard.Operators.DisallowEqualOperators.DisallowedEqualOperator
                     if ($dbOrgValue == $dbActualValue) {
                         continue;
                     }
--- a/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
+++ b/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
@@ -835,10 +835,7 @@ final class UnitOfWork implements PropertyChangedListener
                     $dbOrgValue    = $dateType->convertToDatabaseValue($orgValue);
                     $dbActualValue = $dateType->convertToDatabaseValue($actualValue);
 
-                    $orgTimestamp    = $dbOrgValue instanceof UTCDateTime ? $dbOrgValue->toDateTime()->getTimestamp() : null;
-                    $actualTimestamp = $dbActualValue instanceof UTCDateTime ? $dbActualValue->toDateTime()->getTimestamp() : null;
-
-                    if ($orgTimestamp === $actualTimestamp) {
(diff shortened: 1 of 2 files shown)


magnetik:
(I do not have a local setup with mongo, so I'm sort of using the actions to test.. sorry about that)

edit : I just saw that I can enable them in my fork, i'll test there :)

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -835,9 +835,16 @@ private function computeOrRecomputeChangeSet(ClassMetadata $class, object $docum
...
+                    $orgTimestamp    = $dbOrgValue instanceof UTCDateTime ? $dbOrgValue->toDateTime()->getTimestamp() : null;
+                    $actualTimestamp = $dbActualValue instanceof UTCDateTime ? $dbActualValue->toDateTime()->getTimestamp() : null;
+
+                    if ($orgTimestamp === $actualTimestamp) {
                         continue;
                     }
+
+                    // if ($dbOrgValue == $dbActualValue) {
+                    //     continue;
+                    // }
```
alcaeus: Please remove commented code
```suggestion
```

alcaeus (changes requested):

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -835,10 +835,7 @@ private function computeOrRecomputeChangeSet(ClassMetadata $class, object $docum
                     $dbOrgValue    = $dateType->convertToDatabaseValue($orgValue);
                     $dbActualValue = $dateType->convertToDatabaseValue($actualValue);
 
-                    $orgTimestamp    = $dbOrgValue instanceof UTCDateTime ? $dbOrgValue->toDateTime()->getTimestamp() : null;
-                    $actualTimestamp = $dbActualValue instanceof UTCDateTime ? $dbActualValue->toDateTime()->getTimestamp() : null;
-
-                    if ($orgTimestamp === $actualTimestamp) {
+                    if ($dbOrgValue == $dbActualValue) {
```
alcaeus, line 839: Note that you can suppress the phpcs warning in this file, as it's appropriate to use `==` here. `===` would bypass the object comparison handler implemented for `UTCDateTime` and would thus break the behaviour again.

alcaeus (approved):
LGTM with the PHPCS fix and suppression.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
magnetik: done (I think)

magnetik: How can I find the right rule to silence this ?
Is it SlevomatCodingStandard.Operators.DisallowEqualOperators maybe ?

magnetik:
if you know how to fix the style issue, feel free to fix it directly there if you want to save some time to restart the checks later ^^

Thanks !

malarzm:
It goes like this: https://github.com/doctrine/mongodb-odm/blob/8e0deb186540a89111181c3be89924a3d791045c/lib/Doctrine/ODM/MongoDB/Tools/Console/Command/Schema/ValidateCommand.php#L44-L45

alcaeus:
Thanks @magnetik!

PR #2681, opened 2024-09-30
Fix CI job

alcaeus:
- Disable tests on sharded cluster due to errors like this one:
  > Transaction 8e609811-9b72-4d9f-a0a2-428b82dc8842:653 was aborted on statement 2 due to: an error from cluster data placement change :: caused by :: Encountered error from localhost:27217 during a transaction :: caused by :: sharding status of collection doctrine_odm_tests.Group is not currently known and needs to be recovered
- Remove upload of composer.lock - this fails on 2.10.x due to naming conflicts and its value is questionable at best.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -56,12 +56,13 @@ jobs:
             dependencies: "highest"
             symfony-version: "stable"
           # Test with a 5.0 sharded cluster
-          - topology: "sharded_cluster"
-            php-version: "8.2"
-            mongodb-version: "5.0"
-            driver-version: "stable"
-            dependencies: "highest"
-            symfony-version: "stable"
+          # Currently disabled due to a bug where MongoDB reports "sharding status unknown"
+#          - topology: "sharded_cluster"
+#            php-version: "8.2"
+#            mongodb-version: "5.0"
+#            driver-version: "stable"
+#            dependencies: "highest"
+#            symfony-version: "stable"
 
     steps:
       - name: "Checkout"
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -118,13 +118,6 @@ jobs:
           dependency-versions: "${{ matrix.dependencies }}"
           composer-options: "--prefer-dist"
 
-      - name: "Upload composer.lock as build artifact"
-        uses: actions/upload-artifact@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -56,12 +56,13 @@ jobs:
             dependencies: "highest"
             symfony-version: "stable"
           # Test with a 5.0 sharded cluster
-          - topology: "sharded_cluster"
-            php-version: "8.2"
-            mongodb-version: "5.0"
-            driver-version: "stable"
-            dependencies: "highest"
-            symfony-version: "stable"
+          # Currently disabled due to a bug where MongoDB reports "sharding status unknown"
```
GromNaN, line 59: This bug is tracked somewhere?
Do we have tests that only run on shared cluster?

.github/workflows/continuous-integration.yml
```diff
@@ -117,13 +118,6 @@ jobs:
           dependency-versions: "${{ matrix.dependencies }}"
           composer-options: "--prefer-dist"
 
-      - name: "Upload composer.lock as build artifact"
-        uses: actions/upload-artifact@v4
-        with:
-          name: "composer-lock-phpunit-${{ matrix.php-version }}-${{ matrix.dependencies }}-${{ matrix.mongodb-version }}"
-          path: composer.lock
-          overwrite: true
```
GromNaN, line 125: > Remove upload of composer.lock - this fails on 2.10.x due to naming conflicts and its value is questionable at best.

What is the name in conflict with?

I'm fine with removing this artefact since we can find the installed versions in the job logs if necessary.

.github/workflows/continuous-integration.yml
alcaeus: I found [SERVER-65939](https://jira.mongodb.org/browse/SERVER-65939) when looking around, but haven't confirmed that it's the same issue. There's also a large number of BUILDFAILURE tickets that come up in search.

.github/workflows/continuous-integration.yml
alcaeus: Since we only differentiate between files with php version, dependencies, and mongodb version, there may be multiple jobs running producing the same name (e.g. if the jobs only differ in the orchestration file being used.

GromNaN (approved):

PR #2683, opened 2024-10-01
Support version 2.0 of the MongoDB driver

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

The 2.0 release of the PHP driver will mainly remove deprecated functionality, which was already removed in a previous pull request. This PR adds a separate build job to test the ODM with version 2.0 of the extension and library to ensure everything continues to work. The goal is to support both versions of the extension at the same time to make upgrading easier.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -42,7 +42,7 @@ jobs:
           - dependencies: "lowest"
             php-version: "8.1"
             mongodb-version: "5.0"
-            driver-version: "1.20.0"
+            driver-version: "1.21.0"
             topology: "server"
             symfony-version: "stable"
             proxy: "lazy-ghost"
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -16,6 +16,7 @@ jobs:
     runs-on: "ubuntu-20.04"
 
     strategy:
+      fail-fast: false
       matrix:
         php-version:
           - "8.1"
@@ -68,6 +69,13 @@ jobs:
             dependencies: "highest"
             symfony-version: "stable"
             proxy: "proxy-manager"
+          # Test with ext-2.0
+          - topology: "server"
+            php-version: "8.2"
+            mongodb-version: "7.0"
+            driver-version: "mongodb/mongo-php-driver@v2.x"
+            dependencies: "highest"
+            symfony-version: "7"
           # Test with a 5.0 sharded cluster
           # Currently disabled due to a bug where MongoDB reports "sharding status unknown"
 #          - topology: "sharded_cluster"
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -42,7 +42,7 @@ jobs:
           - dependencies: "lowest"
             php-version: "8.1"
             mongodb-version: "5.0"
-            driver-version: "1.17.0"
+            driver-version: "1.20.0"
             topology: "server"
             symfony-version: "stable"
             proxy: "lazy-ghost"
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
     ],
     "require": {
         "php": "^8.1",
-        "ext-mongodb": "^1.17",
+        "ext-mongodb": "^1.20 || ^2.0",
         "doctrine/cache": "^1.11 || ^2.0",
         "doctrine/collections": "^1.5 || ^2.0",
         "doctrine/event-manager": "^1.0 || ^2.0",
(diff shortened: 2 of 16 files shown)


alcaeus:
Note: static analysis will be red until we figure out how to best handle multiple driver versions in Psalm and PHPStan.

.github/workflows/continuous-integration.yml
```diff
@@ -39,7 +40,7 @@ jobs:
           - dependencies: "lowest"
             php-version: "8.1"
             mongodb-version: "5.0"
-            driver-version: "1.17.0"
+            driver-version: "1.20.0"
```
GromNaN:
```suggestion
            driver-version: "1.21.1"
```

composer.json
```diff
@@ -22,15 +22,15 @@
     ],
     "require": {
         "php": "^8.1",
-        "ext-mongodb": "^1.17",
+        "ext-mongodb": "^1.20 || ^2.0",
```
GromNaN:
```suggestion
        "ext-mongodb": "^1.21 || ^2.0",
```

composer.json
```diff
@@ -22,15 +22,15 @@
...
         "doctrine/cache": "^1.11 || ^2.0",
         "doctrine/collections": "^1.5 || ^2.0",
         "doctrine/event-manager": "^1.0 || ^2.0",
         "doctrine/instantiator": "^1.1 || ^2",
         "doctrine/persistence": "^3.2",
         "friendsofphp/proxy-manager-lts": "^1.0",
         "jean85/pretty-package-versions": "^1.3.0 || ^2.0.1",
-        "mongodb/mongodb": "^1.17.0",
+        "mongodb/mongodb": "^1.20 || ^2.0@dev",
```
GromNaN:
```suggestion
        "mongodb/mongodb": "^1.21.1 || ^2.0@dev",
```

GromNaN (commented):
Calls to `selectDatabase` and `selectCollection` should be replaced with `getDatabase` and `getCollection`.

alcaeus:
> Calls to `selectDatabase` and `selectCollection` should be replaced with `getDatabase` and `getCollection`.

Done, as well as requiring 1.21.0 as minimum version.

tests/Doctrine/ODM/MongoDB/Tests/SchemaManagerTest.php
```diff
@@ -1333,15 +1338,15 @@ private function getMockCollection(?string $name = null)
     private function getMockDatabase()
     {
         $db = $this->createMock(Database::class);
-        $db->method('selectCollection')->willReturnCallback(fn (string $collection) => $this->documentCollections[$collection]);
+        $db->method('getCollection')->willReturnCallback(fn (string $collection) => $this->documentCollections[$collection]);
         $db->method('selectGridFSBucket')->willReturnCallback(fn (array $options) => $this->documentBuckets[$options['bucketName']]);
```
alcaeus, line 1342: Note that we did not add `getGridFSBucket` in 1.21. My suggestion would be to add this method in 2.0 since the deprecation of the `select` methods will only happen later in the 2.x lifetime. WDYT?

GromNaN: Tracked in [PHPLIB-1647](https://jira.mongodb.org/browse/PHPLIB-1647)

tests/Doctrine/ODM/MongoDB/Tests/QueryTest.php
```diff
@@ -552,7 +558,41 @@ public function testFindOptionInheritance(): void
...
+
+            public function current(): array|null
+            {
+                return parent::current();
+            }
+
+            public function key(): int|null
+            {
+                return parent::key();
+            }
```
GromNaN, line 594: I understand it's for specifying return types, but isn't it simpler to ignore static analysis errors?

GromNaN: Or I'm wrong and this is enforced by the [`CursorInterface`](https://github.com/mongodb/mongo-php-driver/blob/bf34d5bd895d289206d70a725cb61420c665bb2e/src/MongoDB/CursorInterface.stub.php#L20).

alcaeus: Yep, this class is just there to provide us with an easy way to create a custom cursor without a query.

GromNaN:
The bundle also needs to be updated: https://github.com/doctrine/DoctrineMongoDBBundle/blob/5.2.x/composer.json#L28

GromNaN (approved):

jorenvh:
Currently the dev release is still being used, the official mongo library released a stable 2.0.0 release 2 weeks back. Can this be updated?

https://github.com/mongodb/mongo-php-library/releases/tag/2.0.0

GromNaN:
Thanks for the reminder @jorenvh. Are you ready to start a pull request? You will need to replace `stable` with `1.21.0` and `mongodb/mongo-php-driver@v2.x` with `stable`.

GromNaN:
I'll make the PR as it seems to be blocking the GitHub Actions for the branch 2.11

PR #2685, opened 2024-10-08
Convert `@psalm-*` annotations to generic or `@phpstan-*` annotation

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | Related to https://github.com/doctrine/.github/pull/50

#### Summary

- [x] Convert any `@psalm-*` to `@phpstan-*`
- [x] `non-empty-string`, `non-empty-array` are supported by in PHPStorm.
- [x] Remove `=` in custom type definition
- [x] `$` prefixed key in array-shape need quotes
- [ ] Detect deprecation usage with phpstan/phpstan-deprecation-rules


lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
```diff
@@ -28,9 +28,9 @@
 /**
  * Fluent interface for building aggregation pipelines.
  *
- * @psalm-import-type SortShape from Sort
- * @psalm-import-type StageExpression from Stage
- * @psalm-type PipelineExpression = list<StageExpression>
+ * @phpstan-import-type SortShape from Sort
+ * @phpstan-import-type StageExpression from Stage
+ * @phpstan-type PipelineExpression list<array<string, mixed>>
```
alcaeus: Shouldn't this be `list<StageExpression>`?

lib/Doctrine/ODM/MongoDB/Aggregation/Stage.php
```diff
@@ -14,8 +14,8 @@
  *
  * @internal
  *
- * @psalm-import-type PipelineExpression from Builder
- * @psalm-type StageExpression = array<non-empty-string, mixed>
+ * @phpstan-type PipelineExpression list<array<string, mixed>>
```
alcaeus: `PipelineExpression` is already defined in `Builder`, as is `StageExpression`. Maybe this is the issue with some of the circular dependencies I was finding. I'm fine with defining `StageExpression` here and `PipelineExpression` in the builder class as we don't have a specific pipeline class.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Lookup.php
```diff
@@ -16,18 +16,18 @@
...
  * Fluent interface for building aggregation pipelines.
  *
- * @psalm-import-type PipelineExpression from Builder
- * @psalm-type PipelineParamType = Builder|Stage|PipelineExpression
- * @psalm-type LookupStageExpression = array{
- *     $lookup: array{
+ * @phpstan-import-type PipelineExpression from Builder
+ * @phpstan-type PipelineParamType Builder|Stage|PipelineExpression
+ * @phpstan-type LookupStageExpression array{
+ *     "$lookup": array{
```
alcaeus: Is there a reason for using double quotes here? I noted that other expressions use single quotes, as does the `as` field in this shape

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Lookup.php
```diff
@@ -16,18 +16,18 @@
...
  *         from: string,
  *         'as'?: string,
  *         localField?: string,
  *         foreignField?: string,
  *         pipeline?: PipelineExpression,
  *         let?: array<string, mixed>,
  *     }
- * }
+ *   }
```
alcaeus: This indentation looks off

lib/Doctrine/ODM/MongoDB/DocumentManager.php
```diff
@@ -553,11 +551,11 @@ public function unlock(object $document): void
     /**
      * Gets the repository for a document class.
      *
-     * @param string $className The name of the Document.
-     * @psalm-param class-string<T> $className
+     * @param string          $className The name of the Document.
+     * @param class-string<T> $className
```
alcaeus:
```suggestion
     * @param class-string<T> $className
```

lib/Doctrine/ODM/MongoDB/DocumentManager.php
```diff
@@ -658,21 +656,21 @@ public function getPartialReference(string $documentName, $identifier): object
...
-     * @param string $className
-     * @param mixed  $id
-     * @param int    $lockMode
-     * @param int    $lockVersion
-     * @psalm-param class-string<T> $className
+     * @param string          $className
+     * @param mixed           $id
+     * @param int             $lockMode
+     * @param int             $lockVersion
+     * @param class-string<T> $className
```
alcaeus:
```suggestion
     * @param class-string<T> $className
     * @param mixed           $id
     * @param int             $lockMode
     * @param int             $lockVersion
```

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -2029,8 +2019,8 @@ public function isInheritanceTypeCollectionPerClass(): bool
     /**
      * Sets the mapped subclasses of this class.
      *
-     * @param string[] $subclasses The names of all mapped subclasses.
-     * @psalm-param class-string[] $subclasses
+     * @param string[]       $subclasses The names of all mapped subclasses.
+     * @param class-string[] $subclasses
```
alcaeus:
```suggestion
     * @param list<class-string> $subclasses The names of all mapped subclasses.
```

alcaeus (commented):
There are a few instances of `phpstan-return` and `phpstan-param` that we could replace with the non-prefixed variants. Essentially, anytime we use a type defined using `phpstan-type` we need to use a phpstan prefix, for all other types we can skip the prefix. I'll let you decide whether you want to do this in this pull request or in a separate one.

As for the aggregation builder types, I believe the issue is that we have duplicate type declarations for `Stage` and `Pipeline`, which is responsible for the circular reference issues we've encountered before. We might want to revisit those as well.

lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
GromNaN: Yes, reverted. The issue that made me change that is gone.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage.php
GromNaN: Moving both definitions to the `Builder` class.

alcaeus (approved):
LGTM #yolo

PR #2686, opened 2024-10-09
WIP Allow usage of Symfony EventDispatcher instead of Doctrine EventManager

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | -

#### Summary

[Symfony EventDispatcher](https://symfony.com/doc/current/components/event_dispatcher.html) is a very close to [Doctrine EventManager](https://www.doctrine-project.org/projects/doctrine-event-manager/en/latest/reference/index.html) in features (add/remove listener, and dispatch an event using a name and an event object).

#### Why not sticking to [PSR-14: Event Dispatcher](https://www.php-fig.org/psr/psr-14/)?
The PSR EventDispatcher interface method `dispatch(object $event)` relies on the object class. But in Doctrine ODM, the same event class `LifecycleEventArgs` is used for multiple events (`postUpdate`, `postRemove`, `postUpdate`, `prePersist`, `preRemove`, `preUpdate`).
We can decide to create a distinct class for each event and use the PSR `EventDispatcherInterface`. With a map of class names to event names.

The event names are very generic, it would be possible to prefix them with `doctrine.mongodb.` to prevent any conflict if the same event dispatcher is used for other events outside of Doctrine.

#### Livecycle callbacks

The livecycle callbacks are not using the event manager, they are called independently in `LifecycleEventManager`.


GromNaN:
Closing as this would have too much impact on used and ecosystem.

PR #2597, opened 2023-12-09
Add support for timeseries

PReimers:
|      Q       |   A
|------------- | -----------
| Type         | bug/feature/improvement
| BC Break     | yes/no
| Fixed issues | 

#### Summary

Replaces #2542

This PR adds the missing time series tests.

PR #2687, opened 2024-10-10
Support mapping time series collections

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | 

#### Summary

Supersedes #2597. I was unhappy with the way we were mapping time series collections, so instead of adding more options to the `Document` attribute I decided to introduce a new attribute and XML mapping element to handle time series data. To make data handling easier, the `ClassMetadataInstance` will store the attribute instance entirely, as it serves as a value object for time series options. As a result, the XML driver is able to create the attribute instance from the data in the XML, needing no additional validation.

The only other change needed was to add support for time series options in the schema manager. Since time series collections need to be created as such, no handling is needed to update time series collections at all. Dropping time series collections works the same as dropping other collections, again requiring no other changes.

The tests in this PR cover the basic mapping and schema manager changes. No functional testing for time series collections was added, but I can add those tests if necessary.

### DIFF
--- a/docs/en/cookbook/time-series-data.rst
+++ b/docs/en/cookbook/time-series-data.rst
@@ -5,8 +5,9 @@ Storing Time Series Data
 
     Support for mapping time series data was added in ODM 2.10.
 
-Time series data is a sequence of data points in which insights are gained by
-analyzing changes over time.
+`time-series data <https://www.mongodb.com/docs/manual/core/timeseries-collections/>`__
+is a sequence of data points in which insights are gained by analyzing changes
+over time.
 
 Time series data is generally composed of these components:
 
--- a/docs/en/cookbook/time-series-data.rst
+++ b/docs/en/cookbook/time-series-data.rst
@@ -92,8 +92,8 @@ measurement:
         // ...
     }
 
-Once we created the schema, we can store our measurements in this time series
-collection and let MongoDB optimise the storage for faster queries:
+Once we create the schema, we can store our measurements in this time series
+collection and let MongoDB optimize the storage for faster queries:
 
 .. code-block:: php
 
@@ -109,8 +109,8 @@ collection and let MongoDB optimise the storage for faster queries:
     $documentManager->persist($measurement);
     $documentManager->flush();
 
-Note that other functionality, such as querying, aggregating data using
-aggregation pipeline, or removing data works the same as with other collections.
+Note that other functionality such as querying, using aggregation pipelines, or
+removing data works the same as with other collections.
 
 Considerations
 --------------
@@ -122,7 +122,7 @@ size. This affects storage requirements and query performance.
 
 For example, with the default ``seconds`` granularity, each bucket groups
 documents for one hour. If each sensor only reports data every few minutes, we'd
-do well to only store them with a ``minute`` granularity. This reduces the
+do well to configure ``minute`` granularity. This reduces the
 number of buckets created, reducing storage and making queries more efficient.
 However, if we were to choose ``hours`` for granularity, readings for a whole
 month would be grouped into one bucket, resulting in slower queries as more
--- a/docs/en/cookbook/time-series-data.rst
+++ b/docs/en/cookbook/time-series-data.rst
@@ -21,6 +21,11 @@ Time series data is generally composed of these components:
 -
     Measurements, which are the data points tracked at increments in time.
 
+A time series document always contains a time value, and one or more measurement
+fields. Metadata is optional, but cannot be added to a time series collection
+after creating it. When using an embedded document for metadata, fields can be
+added to this document after creating the collection.
+
 .. note::
 
     Support for time series collections was added in MongoDB 5.0. Attempting to
@@ -30,9 +35,10 @@ Time series data is generally composed of these components:
 Creating The Model
 ------------------
 
-For this example, we'll be storing data from multiple temperature sensors. Other
-examples for time series include stock data, price information, website visitors,
-and vehicle telemetry (speed, position, etc.).
+For this example, we'll be storing data from multiple sensors measuring
+temperature and humidity. Other examples for time series include stock data,
+price information, website visitors, or vehicle telemetry (speed, position,
+etc.).
 
 First, we define the model for our data:
 
@@ -57,6 +63,8 @@ First, we define the model for our data:
             public int $sensorId,
             #[ODM\Field(type: 'float')]
             public float $temperature,
+            #[ODM\Field(type: 'float')]
+            public float $humidity,
         ) {
             $this->id = (string) new ObjectId();
         }
@@ -66,11 +74,10 @@ Note that we defined the entire model as readonly. While we could theoretically
 change values in the document, in this example we'll assume that the data will
 not change.
 
-Now we can mark the document as a time series document. This is done using the
-``TimeSeries`` attribute. Since we'll be storing data from multiple sensors, we
-store the ID of each sensor as metadata. We only use the temperature as a
-measurement, but we could also add additional sensors. With that in mind, we can
-add the ``TimeSeries`` attribute:
+Now we can mark the document as a time series document. To do so, we use the
+``TimeSeries`` attribute, configuring appropriate values for the time and
+metadata field, which in our case stores the ID of the sensor reporting the
+measurement:
 
 .. code-block:: php
 
@@ -96,6 +103,7 @@ collection and let MongoDB optimise the storage for faster queries:
         time: new DateTimeImmutable(),
         sensorId: $sensorId,
         temperature: $temperature,
+        humidity: $humidity,
     );
 
     $documentManager->persist($measurement);
--- a/docs/en/cookbook/time-series-data.rst
+++ b/docs/en/cookbook/time-series-data.rst
@@ -1,0 +1,123 @@
+Storing Time Series Data
+========================
+
+.. note::
+
+    Support for mapping time series data was added in ODM 2.10.
+
+Time series data is a sequence of data points in which insights are gained by
+analyzing changes over time.
+
+Time series data is generally composed of these components:
+
+-
+    Time when the data point was recorded
+
+-
+    Metadata, which is a label, tag, or other data that identifies a data series
+    and rarely changes
+
+-
+    Measurements, which are the data points tracked at increments in time.
+
+.. note::
+
+    Support for time series collections was added in MongoDB 5.0. Attempting to
+    use this functionality on older server versions will result in an error on
+    schema creation.
+
+Creating The Model
+------------------
+
+For this example, we'll be storing data from multiple temperature sensors. Other
+examples for time series include stock data, price information, website visitors,
+and vehicle telemetry (speed, position, etc.).
+
+First, we define the model for our data:
+
+.. code-block:: php
+
+    <?php
+
+    use DateTimeImmutable;
+    use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
+    use MongoDB\BSON\ObjectId;
+
+    #[ODM\Document]
+    readonly class Measurement
+    {
+        #[ODM\Id]
+        public string $id;
+
+        public function __construct(
+            #[ODM\Field(type: 'date_immutable')]
+            public DateTimeImmutable $time,
+            #[ODM\Field(type: 'int')]
+            public int $sensorId,
+            #[ODM\Field(type: 'float')]
+            public float $temperature,
+        ) {
+            $this->id = (string) new ObjectId();
+        }
+    }
+
+Note that we defined the entire model as readonly. While we could theoretically
+change values in the document, in this example we'll assume that the data will
+not change.
+
+Now we can mark the document as a time series document. This is done using the
+``TimeSeries`` attribute. Since we'll be storing data from multiple sensors, we
+store the ID of each sensor as metadata. We only use the temperature as a
+measurement, but we could also add additional sensors. With that in mind, we can
+add the ``TimeSeries`` attribute:
+
+.. code-block:: php
+
+    <?php
+
+    // ...
+
+    #[ODM\Document]
+    #[ODM\TimeSeries(timeField: 'time', metaField: 'sensorId')]
+    readonly class Measurement
+    {
+        // ...
+    }
+
+Once we created the schema, we can store our measurements in this time series
+collection and let MongoDB optimise the storage for faster queries:
+
+.. code-block:: php
+
+    <?php
+
+    $measurement = new Measurement(
+        time: new DateTimeImmutable(),
+        sensorId: $sensorId,
+        temperature: $temperature,
+    );
+
+    $documentManager->persist($measurement);
+    $documentManager->flush();
+
+Note that other functionality, such as querying, aggregating data using
+aggregation pipeline, or removing data works the same as with other collections.
+
+Considerations
+--------------
+
+With the mapping above, data is stored with a granularity of seconds. Depending
+on how often measurements come in, we can reduce the granularity to minutes or
+hours. This changes how the data is stored internally by changing the bucket
+size. This affects storage requirements and query performance.
+
+For example, with the default ``seconds`` granularity, each bucket groups
+documents for one hour. If each sensor only reports data every few minutes, we'd
+do well to only store them with a ``minute`` granularity. This reduces the
+number of buckets created, reducing storage and making queries more efficient.
+However, if we were to choose ``hours`` for granularity, readings for a whole
+month would be grouped into one bucket, resulting in slower queries as more
+entries have to be traversed when reading data.
+
+More details on granularity and other consideration scan be found in the
+`MongoDB documentation <https://www.mongodb.com/docs/manual/core/timeseries/timeseries-considerations/>`__.
--- a/docs/en/reference/attributes-reference.rst
+++ b/docs/en/reference/attributes-reference.rst
@@ -1144,6 +1144,64 @@ for sharding the document collection.
         //...
     }
 
+#[TimeSeries]
+-------------
+
+This attribute may be used at the class level to mark a collection as containing
+`time-series data <https://www.mongodb.com/docs/manual/core/timeseries-collections/>`__.
+
+.. code-block:: php
+
+    <?php
+
+    use Doctrine\ODM\MongoDB\Mapping\TimeSeries\Granularity;
+
+    #[Document]
+    #[TimeSeries(timeField: 'time', metaField: 'metadata', granularity: Granularity::Seconds)]
+    class Measurements
+    {
+        #[Id]
+        public string $id;
+
+        #[Field]
+        public DateTimeImmutable $time;
+
+        #[EmbedOne(targetDocument: MeasurementMetadata)]
+        public MeasurementMetadata $metadata;
+
+        #[Field]
+        public int $measurement;
+    }
+
+The ``timeField`` attribute is required and denotes the field where the time of
+a time series entry is stored. The following optional attributes may be set:
+
+-
+    ``metaField`` -  The name of the field which contains metadata in each time
+    series document. The field can be of any data type.
+
+-
+    ``granularity`` - Set the granularity to the value that most closely matches
+    the time between consecutive incoming timestamps. This allows MongoDB to
+    optimize how data is stored. Note: this attribute cannot be combined with
+    ``bucketMaxSpanSeconds`` and ``bucketRoundingSeconds``.
+
+-
+    ``bucketMaxSpanSeconds`` - Used with ``bucketRoundingSeconds`` as an
+    alternative to ``granularity``. Sets the maximum time between timestamps
+    in the same bucket. Possible values are 1 - 31356000.
+
+-
+    ``bucketRoundingSeconds`` - Used with ``bucketMaxSpanSeconds``, must be set
+    to the same value as ``bucketMaxSpanSeconds``. When a document requires a
+    new bucket, MongoDB rounds down the document's timestamp value by this
+    interval to set the minimum time for the bucket.
+
+-
+    ``expireAfterSeconds`` - Enables the automatic deletion of documents in a
+    time series collection by specifying the number of seconds after which
+    documents expire. MongoDB deletes these expired documents automatically.
+
 #[UniqueIndex]
 --------------
 
--- a/docs/en/reference/attributes-reference.rst
+++ b/docs/en/reference/attributes-reference.rst
@@ -1148,7 +1148,7 @@ for sharding the document collection.
 -------------
 
 This attribute may be used at the class level to mark a collection as containing
-`time-series data <https://www.mongodb.com/docs/manual/core/timeseries-collections/>`__.
+:doc:`time-series data <../cookbook/time-series-data>`.
 
 .. code-block:: php
 
--- a/doctrine-mongo-mapping.xsd
+++ b/doctrine-mongo-mapping.xsd
@@ -640,6 +640,8 @@
     <xs:attribute name="meta-field" type="xs:NMTOKEN" />
     <xs:attribute name="granularity" type="odm:time-series-granularity" />
     <xs:attribute name="expire-after-seconds" type="xs:integer" />
+    <xs:attribute name="bucket-max-span-seconds" type="odm:time-series-group-seconds" />
+    <xs:attribute name="bucket-rounding-seconds" type="odm:time-series-group-seconds" />
   </xs:complexType>
 
   <xs:simpleType name="time-series-granularity">
@@ -647,6 +649,13 @@
       <xs:enumeration value="seconds" />
       <xs:enumeration value="minutes" />
       <xs:enumeration value="hours" />
+    </xs:restriction>
+  </xs:simpleType>
+
+  <xs:simpleType name="time-series-group-seconds">
+    <xs:restriction base="xs:integer">
+      <xs:minInclusive value="1" />
+      <xs:maxInclusive value="31536000" />
     </xs:restriction>
   </xs:simpleType>
 
--- a/doctrine-mongo-mapping.xsd
+++ b/doctrine-mongo-mapping.xsd
@@ -104,6 +104,7 @@
       <xs:element name="shard-key" type="odm:shard-key" minOccurs="0" />
       <xs:element name="read-preference" type="odm:read-preference" minOccurs="0" />
       <xs:element name="schema-validation" type="odm:schema-validation" minOccurs="0" />
+      <xs:element name="time-series" type="odm:time-series" minOccurs="0" />
     </xs:choice>
 
     <xs:attribute name="db" type="xs:NMTOKEN" />
@@ -631,6 +632,21 @@
       <xs:enumeration value="off" />
       <xs:enumeration value="strict" />
       <xs:enumeration value="moderate" />
+    </xs:restriction>
+  </xs:simpleType>
+
+  <xs:complexType name="time-series">
+    <xs:attribute name="time-field" type="xs:NMTOKEN" use="required" />
+    <xs:attribute name="meta-field" type="xs:NMTOKEN" />
+    <xs:attribute name="granularity" type="odm:time-series-granularity" />
+    <xs:attribute name="expire-after-seconds" type="xs:integer" />
+  </xs:complexType>
+
+  <xs:simpleType name="time-series-granularity">
+    <xs:restriction base="xs:token">
+      <xs:enumeration value="seconds" />
+      <xs:enumeration value="minutes" />
+      <xs:enumeration value="hours" />
     </xs:restriction>
   </xs:simpleType>
 
--- a/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
@@ -800,9 +800,6 @@ use function trigger_deprecation;
      */
     public $isReadOnly;
 
-    /** READ ONLY: indicates whether the collection is a time series collection */
-    public bool $isTimeSeries = false;
-
     /** READ ONLY: stores metadata about the time series collection */
     public ?TimeSeries $timeSeriesOptions = null;
 
@@ -2185,7 +2182,6 @@ use function trigger_deprecation;
     {
         $this->validateTimeSeriesOptions($options);
 
-        $this->isTimeSeries      = true;
         $this->timeSeriesOptions = $options;
     }
 
@@ -2542,7 +2538,6 @@ use function trigger_deprecation;
             'idGenerator',
             'indexes',
             'shardKey',
-            'isTimeSeries',
             'timeSeriesOptions',
         ];
 
--- a/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
@@ -13,6 +13,7 @@ use Doctrine\Instantiator\Instantiator;
 use Doctrine\Instantiator\InstantiatorInterface;
 use Doctrine\ODM\MongoDB\Id\IdGenerator;
 use Doctrine\ODM\MongoDB\LockException;
+use Doctrine\ODM\MongoDB\Mapping\Annotations\TimeSeries;
 use Doctrine\ODM\MongoDB\Types\Incrementable;
 use Doctrine\ODM\MongoDB\Types\Type;
 use Doctrine\ODM\MongoDB\Types\Versionable;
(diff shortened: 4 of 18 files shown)


lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -621,6 +621,20 @@ public function createDocumentCollection(string $documentName, ?int $maxTimeMs =
...
+            $options['timeseries'] = array_filter(
+                [
+                    'timeField' => $class->timeSeriesOptions->timeField,
+                    'metaField' => $class->timeSeriesOptions->metaField,
+                    'granularity' => $class->timeSeriesOptions->granularity?->value,
+                ],
+            );
+
+            if ($class->timeSeriesOptions->expireAfterSeconds) {
+                $options['expireAfterSeconds'] = $class->timeSeriesOptions->expireAfterSeconds;
```
GromNaN, line 638: Noted this option is specific to TimeSeries collections.

tests/Doctrine/ODM/MongoDB/Tests/Mapping/AbstractMappingDriverTestCase.php
```diff
@@ -1296,3 +1308,28 @@ public function createViewAggregation(Builder $builder): void
                 ->includeFields(['name']);
     }
 }
+
+/**
+ * @ODM\Document(collection="cms_users", writeConcern=1, readOnly=true)
+ * @ODM\TimeSeries(timeField="time", metadataField="metadata", granularity=Granularity::Seconds, expireAfterSeconds=86400)
+ */
+#[ODM\Document]
+#[ODM\TimeSeries('time', 'metadata', Granularity::Seconds, 86400)]
```
GromNaN: Named arguments are frequently used for attributes, you should use them to ensure there is not regression with argument name change.

GromNaN (approved):
SGTM.

The functional test would ensure we pass the correct options to create the collection and this options works with all supported server versions (past and future).

doctrine-mongo-mapping.xsd
```diff
@@ -634,4 +635,19 @@
     </xs:restriction>
   </xs:simpleType>
 
+  <xs:complexType name="time-series">
+    <xs:attribute name="time-field" type="xs:NMTOKEN" use="required" />
+    <xs:attribute name="meta-field" type="xs:NMTOKEN" />
+    <xs:attribute name="granularity" type="odm:time-series-granularity" />
```
jmikola, line 641: MongoDB 6.3 added `bucketMaxSpanSeconds` and `bucketRoundingSeconds` as an alternative to `granularity`. Both options are expected to have the same value (between `1` and `31536000`), although we can rely on the server to enforce that. See:

 * https://www.mongodb.com/docs/manual/core/timeseries/timeseries-procedures/
 * https://www.mongodb.com/docs/manual/reference/command/create/

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -621,6 +621,20 @@ public function createDocumentCollection(string $documentName, ?int $maxTimeMs =
             $options['validationLevel']  = $class->getValidationLevel();
         }
 
+        if ($class->isTimeSeries) {
+            $options['timeseries'] = array_filter(
+                [
+                    'timeField' => $class->timeSeriesOptions->timeField,
+                    'metaField' => $class->timeSeriesOptions->metaField,
+                    'granularity' => $class->timeSeriesOptions->granularity?->value,
```
jmikola: I think you can rely on PHPC to encode enums as their backing value ([PHPC-2083](https://jira.mongodb.org/browse/PHPC-2083)). If so, it'd be helpful to leave a comment explaining that.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -2769,4 +2786,15 @@ private function validateAndCompleteTypedManyAssociationMapping(array $mapping):
 
         return $mapping;
     }
+
+    private function validateTimeSeriesOptions(TimeSeries $options): void
```
jmikola, line 2774: If you do add client-side validation to make `granularity` mutually exclusive with the other two options, I assume this would be the place to do so.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -2769,4 +2786,15 @@ private function validateAndCompleteTypedManyAssociationMapping(array $mapping):
...
+    {
+        if (! $this->hasField($options->timeField)) {
+            throw MappingException::timeSeriesFieldNotFound($this->name, $options->timeField, 'time');
+        }
+
+        if ($options->metaField && ! $this->hasField($options->metaField)) {
```
jmikola: I realize field names wouldn't be empty strings, but should this compare with `!== null` to be strict?

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -806,6 +807,12 @@
      */
     public $isReadOnly;
 
+    /** READ ONLY: indicates whether the collection is a time series collection */
+    public bool $isTimeSeries = false;
```
jmikola: Is this only used in tests? If so, why not just rely on `$timeSeriesOptions` being set?

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AttributeDriver.php
```diff
@@ -288,6 +289,12 @@ public function loadMetadataForClass($className, PersistenceClassMetadata $metad
             $this->setShardKey($metadata, $classAttributes[ShardKey::class]);
         }
 
+        // Mark as time series only after mapping all fields
+        if (isset($classAttributes[TimeSeries::class])) {
+            assert($classAttributes[TimeSeries::class] instanceof TimeSeries);
```
jmikola, line 294: I realize this is consistent with how ShardKey is handled above, but is there any possibility that `$classAttributes[TimeSeries::class]` is set and _not_ a TimeSeries object? In other words, would some earlier validation or type enforcement in PHP itself handle that?

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AttributeDriver.php
```diff
@@ -288,6 +289,12 @@ public function loadMetadataForClass($className, PersistenceClassMetadata $metad
...
+            $metadata->markAsTimeSeries($classAttributes[TimeSeries::class]);
```
jmikola, line 295: Is this the first `markAs` method in ClassMetadata? Just curious why you went this route instead of a `setTimeSeries` method as we have for ShardKey.

lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
```diff
@@ -335,13 +337,34 @@ public function loadMetadataForClass($className, \Doctrine\Persistence\Mapping\C
...
-            return;
+        if (isset($xmlRoot->{'also-load-methods'})) {
+            foreach ($xmlRoot->{'also-load-methods'}->{'also-load-method'} as $alsoLoadMethod) {
+                $metadata->registerAlsoLoadMethod((string) $alsoLoadMethod['method'], (string) $alsoLoadMethod['field']);
+            }
         }
 
-        foreach ($xmlRoot->{'also-load-methods'}->{'also-load-method'} as $alsoLoadMethod) {
-            $metadata->registerAlsoLoadMethod((string) $alsoLoadMethod['method'], (string) $alsoLoadMethod['field']);
+        if (! isset($xmlRoot->{'time-series'})) {
```
jmikola: This method seems long enough to do without this guard statement. If another attribute is added down the line, you'll likely need to wrap this in a condition as you just did for `also-load-methods`, so why not just do that now and avoid ever having to do this again?

lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
```diff
@@ -335,13 +337,34 @@ public function loadMetadataForClass($className, \Doctrine\Persistence\Mapping\C
...
+            return;
         }
+
+        $attributes = $xmlRoot->{'time-series'}->attributes();
+
+        $metaField          = isset($attributes['meta-field'])
+            ? (string) $attributes['meta-field']
```
jmikola: Do I assume correctly that the casting makes `??` unusable here? Casts have higher [operator precedence](https://www.php.net/manual/en/language.operators.precedence.php), so you'd either get an error attempting to cast a nonexistent offset or end up casting `null` to a string, neither of which would work.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -621,6 +621,20 @@ public function createDocumentCollection(string $documentName, ?int $maxTimeMs =
             $options['validationLevel']  = $class->getValidationLevel();
         }
 
+        if ($class->isTimeSeries) {
+            $options['timeseries'] = array_filter(
```
jmikola, line 625: Noted this relies on `empty()` to filter out values. This could be problematic in other contexts, but I don't think any issues come up since empty strings wouldn't make sense and the `bucketMaxSpanSeconds` option's minimum value is one (not zero).

lib/Doctrine/ODM/MongoDB/SchemaManager.php
jmikola: It is a bit odd that [create](https://www.mongodb.com/docs/manual/reference/command/create/) receives this as a top-level option instead of within `timeseries`.

alcaeus: The option is also used for clustered collections, which can leverage a TTL to remove expired documents. I'm guessing the team decided to reuse the option since it was already there instead of adding an extra option in the timeseries options document.

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AttributeDriver.php
alcaeus: Based on the logic we use to fill `$classAttributes`, it can't be anything other than a `TimeSeries` instance. The `assert` is mainly for static analysis, as I don't see how we can use template annotations to specify this accordingly. Essentially, we'd need something like `array<class-string<T>, T>`, but since each item is a different instance I'm not sure how that would work. I'll see if I can annotate this accordingly.

lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
alcaeus: That is correct, and pretty much my only disappointment with the null coalesce operator.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
alcaeus: That's correct, the `null` check makes more sense. Updated accordingly 👍

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
alcaeus: That's correct, so we could probably do away with this. We have it for other flags, so I added it for consistency but it's probably not relevant for a while, in which we can always add this.

lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
alcaeus: This is our coding standard requiring an early exit, which seems pointless at this point. For now I've disabled the rule around this method, but I may want to refactor this into multiple methods at some point.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
alcaeus: I figured relying on `empty` makes sense, but with its strange behaviour around emptyish values, I added an explicit closure to only remove `null` values.

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AttributeDriver.php
alcaeus: Yes, this is the first time I used this naming. I could change it to `setTimeSeriesOptions` to be consistent with others, but I figured that technically we're marking this as a time series document. Other methods could be changed as well, e.g. `setShardKey` should be `markAsSharded` as it doesn't allow to unshard metadata at a later stage anyways.

I have started another branch with a more comprehensive metadata change (e.g. using objects instead of arrays for field mappings, along with other changes), so I can defer the naming change to then if you'd prefer consistency here.

doctrine-mongo-mapping.xsd
alcaeus: Ah, I missed those. I decided against adding validation, except for limiting the range in the XSD as we're able to type this better. I'm not sure why there are two different options that require having the same value, but something tells me that this requirement may change in the future (a great example why we have the [Defy Augury](https://github.com/mongodb/specifications/blob/master/source/driver-mantras.md#defy-augury) mantra in the driver specifications).

lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
```diff
@@ -335,15 +338,36 @@ public function loadMetadataForClass($className, \Doctrine\Persistence\Mapping\C
...
+
+            $metaField          = isset($attributes['meta-field'])
+                ? (string) $attributes['meta-field']
+                : null;
+            $granularity        = isset($attributes['granularity'])
+                ? Granularity::from((string) $attributes['granularity'])
+                : null;
+            $expireAfterSeconds = isset($attributes['expire-after-seconds'])
+                ? (int) $attributes['expire-after-seconds']
+                : null;
```
GromNaN: Is it split on multiple lines bacause of CS rule? That seems too much.

```suggestion
            $metaField          = isset($attributes['meta-field']) ? (string) $attributes['meta-field'] : null;
            $granularity        = isset($attributes['granularity']) ? Granularity::from((string) $attributes['granularity']) : null;
            $expireAfterSeconds = isset($attributes['expire-after-seconds']) ? (int) $attributes['expire-after-seconds'] : null;
```

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AttributeDriver.php
jmikola: If you intend to revise things anyway, I have no preference. Probably easier to adopt the new format earlier and migrate older stuff as needed at some later point.

lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
jmikola: I agree the rule makes little sense for this kind of method. These aren't traditional guard statements of returning early before some logical operation is performed, because the method is handling many different kinds of operations. If anything, it just makes the final operation inconsistent from those handled earlier.

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AttributeDriver.php
jmikola: I'm fine keeping the `assert()` here for static analysis. Just wanted to confirm.

Verbose template annotations can die in a fire as far as I'm concerned. :fire:

doctrine-mongo-mapping.xsd
jmikola: > I'm not sure why there are two different options that require having the same value

From the minds that brought us a required `sparsity` option for [`encrypt()`](https://www.php.net/manual/en/mongodb-driver-clientencryption.encrypt.php) methods, before it was recently made optional. I jest, but that's what this reminded me of.

lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
```diff
@@ -335,15 +338,36 @@ public function loadMetadataForClass($className, \Doctrine\Persistence\Mapping\C
...
+
+            $metadata->markAsTimeSeries(new TimeSeries(
```
jmikola, line 356: This method needs to be updated to handle the `bucket-max-span-seconds` and `bucket-rounding-seconds` attributes. You may also want to add a test for that since this was missed.

tests/Doctrine/ODM/MongoDB/Tests/Mapping/AbstractMappingDriverTestCase.php
```diff
@@ -677,6 +678,16 @@ public function testEnumType(): void
         self::assertTrue($metadata->fieldMappings['nullableSuit']['nullable']);
         self::assertInstanceOf(EnumReflectionProperty::class, $metadata->reflFields['nullableSuit']);
     }
+
+    public function testTimeSeriesDocument(): void
```
jmikola: Consider adding another test for the recently added bucket options.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -806,6 +807,9 @@
      */
     public $isReadOnly;
 
+    /** READ ONLY: stores metadata about the time series collection */
+    public ?TimeSeries $timeSeriesOptions = null;
```
jmikola, line 804: Could you use `readonly` here? AFAIK, that would just mean that `markAsTimeSeries()`, which sets this, can only be called once. Perhaps that's something to defer until later, as I expect it could be done across the board for all "READ ONLY" metadata properties.

jmikola (commented):
The API looks stable enough to add docs to this PR.

lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
alcaeus: This was my attempt at making it more readable. I've changed it to single-line.

lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
alcaeus: Good catch. Fixed this and added a test as suggested below.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
alcaeus: Yes and no. What we'd want here is asymetric visibility, where changing the property is only done through accessors (such as `markAsTimeSeries`) that can incorporate checks, while reading is always allowed. `readonly` in PHP actually means `write-once`, meaning that one wouldn't be able to override the time series configuration (e.g. to change granularity) when using inheritance mapping.

alcaeus:
@jmikola good point about the docs. I added `TimeSeries` to the attribute reference and wrote a short cookbook article based on the MongoDB documentation. I'd appreciate a thorough review of the cookbook entry to ensure it works for users 🙇‍♂️

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
Steveb-p: `readonly` = you can change the assigned object properties. You cannot change the reference to the object. And I think - I'm not sure though - that you can only make assignment in the constructor :thinking: ?

docs/en/cookbook/time-series-data.rst
```diff
@@ -0,0 +1,123 @@
...
+
+        public function __construct(
+            #[ODM\Field(type: 'date_immutable')]
+            public DateTimeImmutable $time,
+            #[ODM\Field(type: 'int')]
+            public int $sensorId,
+            #[ODM\Field(type: 'float')]
+            public float $temperature,
+        ) {
+            $this->id = (string) new ObjectId();
```
jmikola, line 69: Does ODM require a string cast here, or is it only done to satisfy the `string` type hint on the property? IIRC, ODM would have allowed an ObjectId on the initially created entity, and left is as-is for insertion.

docs/en/cookbook/time-series-data.rst
```diff
@@ -0,0 +1,123 @@
...
+    // ...
+
+    #[ODM\Document]
+    #[ODM\TimeSeries(timeField: 'time', metaField: 'sensorId')]
+    readonly class Measurement
+    {
+        // ...
+    }
+
+Once we created the schema, we can store our measurements in this time series
```
jmikola:
```suggestion
Once we create the schema, we can store our measurements in this time series
```

Alternatively: "once we have created the schema"

docs/en/cookbook/time-series-data.rst
```diff
@@ -0,0 +1,123 @@
...
+collection and let MongoDB optimise the storage for faster queries:
```
jmikola:
```suggestion
collection and let MongoDB optimize the storage for faster queries:
```

Decide if ODM docs use American or British English for spelling.

docs/en/cookbook/time-series-data.rst
```diff
@@ -0,0 +1,123 @@
...
+        time: new DateTimeImmutable(),
+        sensorId: $sensorId,
+        temperature: $temperature,
+    );
+
+    $documentManager->persist($measurement);
+    $documentManager->flush();
+
+Note that other functionality, such as querying, aggregating data using
+aggregation pipeline, or removing data works the same as with other collections.
```
jmikola:
```suggestion
Note that other functionality such as querying, using aggregation pipelines, or
removing data works the same as with other collections.
```

Removing word repetition here.

If you preserve the comma after "functionality", then there should be a second comma in "removing data, works". Both commas would segregate the "such as" list.

docs/en/cookbook/time-series-data.rst
```diff
@@ -0,0 +1,123 @@
...
+--------------
+
+With the mapping above, data is stored with a granularity of seconds. Depending
+on how often measurements come in, we can reduce the granularity to minutes or
+hours. This changes how the data is stored internally by changing the bucket
+size. This affects storage requirements and query performance.
+
+For example, with the default ``seconds`` granularity, each bucket groups
+documents for one hour. If each sensor only reports data every few minutes, we'd
+do well to only store them with a ``minute`` granularity. This reduces the
```
jmikola:
```suggestion
do well to configure ``minute`` granularity. This reduces the
```

Alternatively, replace "configure" with "opt for". Either way, we can remove "a" here.

docs/en/reference/attributes-reference.rst
```diff
@@ -1144,6 +1144,64 @@ for sharding the document collection.
         //...
     }
 
+#[TimeSeries]
+-------------
+
+This attribute may be used at the class level to mark a collection as containing
+`time-series data <https://www.mongodb.com/docs/manual/core/timeseries-collections/>`__.
```
jmikola: Would it make more sense to link to the cookbook article instead? In turn, maybe you can link to this documentation from the intro of that article (it presently isn't referenced at all, just the considerations page).

docs/en/reference/attributes-reference.rst
```diff
@@ -1144,6 +1144,64 @@ for sharding the document collection.
...
+
+        #[EmbedOne(targetDocument: MeasurementMetadata)]
+        public MeasurementMetadata $metadata;
+
+        #[Field]
+        public int $measurement;
+    }
+
+The ``timeField`` attribute is required and denotes the field where the time of
+a time series entry is stored. The following optional attributes may be set:
```
jmikola, line 1177: Isn't the annotation tag itself the "attribute"? Would something like `timeField` be considered a parameter or option?

Edit: I see I used "attribute" to refer to both in the SearchIndex docs, which was likely for consistency with existing docs. That said, it's a bit confusing to use the same term to refer to two different things.

docs/en/reference/attributes-reference.rst
```diff
@@ -1144,6 +1144,64 @@ for sharding the document collection.
...
+
+-
+    ``metaField`` -  The name of the field which contains metadata in each time
+    series document. The field can be of any data type.
```
jmikola, line 1181: This wasn't immediately clear to me from reading the [Time Series docs](https://www.mongodb.com/docs/manual/core/timeseries-collections/), but are documents in a time series collection restricted to only having `_id`, `timeField`, and `metaField` fields? The docs suggest that everything else gets stuffed into the `metaField`.

If so, this may be something to note in the cookbook article when we demonstrate model creation.

lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
```diff
@@ -335,15 +338,34 @@ public function loadMetadataForClass($className, \Doctrine\Persistence\Mapping\C
...
+            }
         }
 
-        foreach ($xmlRoot->{'also-load-methods'}->{'also-load-method'} as $alsoLoadMethod) {
-            $metadata->registerAlsoLoadMethod((string) $alsoLoadMethod['method'], (string) $alsoLoadMethod['field']);
+        if (isset($xmlRoot->{'time-series'})) {
+            $attributes = $xmlRoot->{'time-series'}->attributes();
+
+            $metaField             = isset($attributes['meta-field']) ? (string) $attributes['meta-field'] : null;
+            $granularity           = isset($attributes['granularity']) ? Granularity::from((string) $attributes['granularity']) : null;
```
jmikola, line 351: Do you want to rely on [`BackedEnum::from()`](https://www.php.net/manual/en/backedenum.from.php) throwing a ValueError for a bad string value? I don't think we can trust that the XSD will be enforced.

An explicit error is probably better than silently accepting `null` via [`tryFrom()`](https://www.php.net/manual/en/backedenum.tryfrom.php).

tests/Doctrine/ODM/MongoDB/Tests/Mapping/ClassMetadataTest.php
```diff
@@ -968,6 +970,92 @@ public function testDefaultValueForValidationLevel(): void
...
+        // We don't throw for invalid settings here, e.g. bucketMaxSpanSeconds not being equal to bucketRoundingSeconds
+        self::assertSame(10, $metadata->timeSeriesOptions->bucketRoundingSeconds);
+    }
+
+    public function testTimeSeriesMappingWithGranularityAndBucketMaxSpanSeconds(): void
+    {
+        $metadata = $this->dm->getClassMetadata(TimeSeriesTestDocument::class);
+        $metadata->markAsTimeSeries(new ODM\TimeSeries('time', granularity: Granularity::Hours, bucketMaxSpanSeconds: 15, bucketRoundingSeconds: 20));
+
+        // We don't throw for invalid settings here, e.g. bucketMaxSpanSeconds not being equal to bucketRoundingSeconds
```
jmikola: You're mainly testing that `markAsTimeSeries()` applies the options, and this method works with all three. I think all of the previous tests for individual `granularity` and bucket options could all be removed in favor of this one.

The block comment could then be updated to also note that you also ignore validation for specifying both `granularity` and any bucket options.

tests/Documents/TimeSeries/TimeSeriesDocument.php
```diff
@@ -0,0 +1,26 @@
...
+
+namespace Documents\TimeSeries;
+
+use DateTime;
+use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
+use Doctrine\ODM\MongoDB\Mapping\TimeSeries\Granularity;
+
+#[ODM\Document]
+#[ODM\TimeSeries(timeField: 'time', metaField: 'metadata', granularity: Granularity::Seconds, expireAfterSeconds: 86400)]
+class TimeSeriesDocument
```
jmikola, line 13: Where is this class actually used?

tests/Doctrine/ODM/MongoDB/Tests/Mapping/ClassMetadataTest.php
```diff
@@ -968,6 +970,92 @@ public function testDefaultValueForValidationLevel(): void
         $cm = new ClassMetadata('stdClass');
         self::assertEquals(ClassMetadata::SCHEMA_VALIDATION_LEVEL_STRICT, $cm->getValidationLevel());
     }
+
+    public function testDefaultTimeSeriesMapping(): void
+    {
+        $metadata = $this->dm->getClassMetadata(TimeSeriesTestDocument::class);
+
+        self::assertNull($metadata->timeSeriesOptions);
```
jmikola: Is this a meaningful test? I'm also a bit confused that these tests reference a TimeSeriesDocument class defined in the same file, while you also have a separate TimeSeriesDocument class defined in its own file.

docs/en/cookbook/time-series-data.rst
alcaeus: The default `id` type converts an `ObjectId` instance to a string in the entity. It's bothered me for a while as working with the `ObjectId` instance directly would make more sense, but I've shied away from making the big BC break. We could use `ObjectId|string` as type for the identifier, but this would create ambiguity in the API as users would always have to check types when using the identifier, hence the string cast.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
alcaeus: > I think - I'm not sure though - that you can only make assignment in the constructor

I thought something like that as well, but then it seems we are wrong, and the property can be initialised after the constructor, but only from within the class scope: https://3v4l.org/cV5QL

For our case, this still wouldn't work however, as we can't assign a default value to a readonly property but have to leave it uninitialised, which means that accessing the property always needs to be guarded by an `isset` check :(

docs/en/reference/attributes-reference.rst
alcaeus: Good point. I filed #2691 to replace it with `argument` across the whole file. Please add your opinion to the suggested wording and/or suggest alternatives.

docs/en/reference/attributes-reference.rst
alcaeus: Good point. I've added the following language to the cookbook entry:
> A time series document always contains a time value, and one or more measurement
fields. Metadata is optional, but cannot be added to a time series collection
after creating it. When using an embedded document for metadata, fields can be
added to this document after creating the collection.

I've also extended the example to store temperature and humidity, showing that time series documents can contain multiple measurements per entry.

lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
alcaeus: While we do use an enumeration in the schema file, I don't quite want to rely on that. Using `BackedEnum::from` is the best way to ensure validity of the value while avoiding duplication.

tests/Doctrine/ODM/MongoDB/Tests/Mapping/ClassMetadataTest.php
alcaeus: Good point, removed the other three tests and updated the comment.

tests/Documents/TimeSeries/TimeSeriesDocument.php
alcaeus: The SchemaManagerTest uses this class.

tests/Doctrine/ODM/MongoDB/Tests/Mapping/ClassMetadataTest.php
alcaeus: No, it isn't. This was left over from refactoring out the `isTimeSeries` field, where I replaced all occurrences of checking the field with the corresponding check on the `timeSeriesOptions` property. I removed the test since it's redundant.

As for the classes being used, there are lots of classes defined in this test file, I believe to avoid issues where changing the mapping of one class to test new functionality affects related tests. I've followed this for consistency, but we'd probably do well to refactor this in the near future.

jmikola: Aye. Ideally, I suppose these tests would be split into their own files or define anonymous classes within the test itself. I understand there's so much prior art that it'd be a chore to clean everything up at once, though.

tests/Doctrine/ODM/MongoDB/Tests/SchemaManagerTest.php
```diff
@@ -20,6 +21,7 @@
 use Documents\Sharded\ShardedOne;
 use Documents\Sharded\ShardedOneWithDifferentKey;
 use Documents\SimpleReferenceUser;
+use Documents\TimeSeries\TimeSeriesDocument;
```
jmikola, line 24: Noted that this is where the class is referenced. Missed that earlier.

jmikola (approved):

bradroc:
@alcaeus This had been a pain point in some of our automated testing for the past 18 months as creating timeseries collections required custom scripts. Thank you for this contribution!

alcaeus:
> @alcaeus This had been a pain point in some of our automated testing for the past 18 months as creating timeseries collections required custom scripts. Thank you for this contribution!

Thank you for taking the time to write this comment, I really appreciate it!

If you have a chance to try out the feature please do give us feedback if anything does not work as expected. I understand this only affects schema creation, but any real-world scenarios you can test help us greatly!

PR #2689, opened 2024-10-14
Fix "Cannot use object of type stdClass as array" in search index drop

### LINKED ISSUE #2684: odm:schema:drop command does not drop search indexes
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.9.0

#### Summary

odm:schema:drop does not drop  search indexes.

#### Current behavior

After running odm:schema:create while having mapped document with search index defined & mapped,  odm:schema:drop does not drop search index. While trying to drop it it outputs error: Cannot use object of type stdClass as array.

It succesfuly drops normal indexes, collections and databases after that.

#### How to reproduce
1. Add some search index to mapped document
2. run odm:schema:create
3. run odm:schema:drop

#### Expected behavior

Search index should be succesfully dropped.

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | fix #2684

#### Summary

The `typeMap` option from the `Collection` instance is not inherited (bug report [PHPLIB-1548](https://jira.mongodb.org/browse/PHPLIB-1548)).

This was not detected by unit tests as the call to `Collection::listSearchIndexes()` is mocked. https://github.com/doctrine/mongodb-odm/blob/ff90e9ab519560ec8a09794e9778dcd7d59f10ad/tests/Doctrine/ODM/MongoDB/Tests/SchemaManagerTest.php#L444-L454

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/SchemaManager.php
+++ b/lib/Doctrine/ODM/MongoDB/SchemaManager.php
@@ -413,7 +413,9 @@ final class SchemaManager
 
         $definedNames = array_column($searchIndexes, 'name');
         try {
-            $existingNames = array_column(iterator_to_array($collection->listSearchIndexes()), 'name');
+            /* The typeMap option can be removed when bug is fixed in the minimum required version.
+             * https://jira.mongodb.org/browse/PHPLIB-1548 */
+            $existingNames = array_column(iterator_to_array($collection->listSearchIndexes(['typeMap' => ['root' => 'array']])), 'name');
         } catch (CommandException $e) {
             /* If $listSearchIndexes doesn't exist, only throw if search indexes have been defined.
              * If no search indexes are defined and the server doesn't support search indexes, there's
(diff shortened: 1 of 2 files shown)


alcaeus (approved):
Change LGTM, but we should add a note that we can drop this once we bump the version requirement to the version that fixes this.

alcaeus:
Note: the PHPStan errors suggest that we need to change the `@param` annotation for `listSearchIndexes` as well - it should read `{name?: string, ...}` to allow extra elements in the options array.

GromNaN:
> Note: the PHPStan errors suggest that we need to change the `@param` annotation for `listSearchIndexes` as well - it should read `{name?: string, ...}` to allow extra elements in the options array.

Done in https://github.com/mongodb/mongo-php-library/pull/1480. I'll add this to the baseline for the time being.

Surprisingly, phpstan doesn't care.

PR #2690, opened 2024-10-14
Prevent search index creation with `dynamic: false` and empty `fields` mapping

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | -

#### Summary

When creating a document with `SearchIndex` configuration missing both `dynamic` and `fields` config, the index definition sent to MongoDB is incorrect. `definition.mappings` is required and must be an object.

By default, when  `definition.mappings` is an empty object, the server fallbacks to `dynamic: false` without fields. This mapping is useless as this means no field is mapped.

Prevent this invalid configuration by throwing a MappingException when `dynamic` is not set or `false`, and `fields` is not set or empty.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
@@ -1220,9 +1220,15 @@ use function trigger_deprecation;
      */
     public function addSearchIndex(array $definition, ?string $name = null): void
     {
+        $name ??= self::DEFAULT_SEARCH_INDEX_NAME;
+
+        if (empty($definition['mappings']['dynamic']) && empty($definition['mappings']['fields'])) {
+            throw MappingException::emptySearchIndexDefinition($this->name, $name);
+        }
+
         $this->searchIndexes[] = [
             'definition' => $definition,
-            'name' => $name ?? self::DEFAULT_SEARCH_INDEX_NAME,
+            'name' => $name,
         ];
     }
 
--- a/lib/Doctrine/ODM/MongoDB/Mapping/MappingException.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/MappingException.php
@@ -291,4 +291,9 @@ final class MappingException extends BaseMappingException
             $fieldName,
         ));
     }
+
+    public static function emptySearchIndexDefinition(string $className, string $indexName): self
+    {
+        return new self(sprintf('%s search index "%s" must be dynamic or specify a field mapping', $className, $indexName));
+    }
 }
--- a/tests/Doctrine/ODM/MongoDB/Tests/Mapping/ClassMetadataTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Mapping/ClassMetadataTest.php
@@ -971,6 +971,15 @@ class ClassMetadataTest extends BaseTestCase
         $cm = new ClassMetadata('stdClass');
         self::assertEquals(ClassMetadata::SCHEMA_VALIDATION_LEVEL_STRICT, $cm->getValidationLevel());
     }
+
+    public function testEmptySearchIndexDefinition(): void
+    {
+        $cm = new ClassMetadata('stdClass');
+
+        $this->expectException(MappingException::class);
+        $this->expectExceptionMessage('stdClass search index "default" must be dynamic or specify a field mapping');
+        $cm->addSearchIndex(['mappings' => []]);
+    }
 }
 
 /** @template-extends DocumentRepository<self> */


lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -1228,6 +1229,11 @@ public function hasIndexes(): bool
      */
     public function addSearchIndex(array $definition, ?string $name = null): void
     {
+        // "mappings" is required and must be an object
+        if (! isset($definition['mappings']) || $definition['mappings'] === []) {
+            $definition['mappings'] = new stdClass();
+        }
```
jmikola, line 1227: Similar to my question in https://github.com/mongodb/mongo-php-library/pull/1482#discussion_r1801481363, I'm not sure what use case we're supporting here. If `dynamic: false` is the default (per [`createSearchIndexes` docs](https://www.mongodb.com/docs/manual/reference/command/createSearchIndexes/)), why should we allow creating a search index with _no_ mappings at all?

The [CreateSearchIndexes](https://github.com/mongodb/mongo-php-library/blob/1.20.0/src/Operation/CreateSearchIndexes.php) operation and [SearchIndexInput](https://github.com/mongodb/mongo-php-library/blob/1.20.0/src/Model/SearchIndexInput.php), which it utilizes, do not validate this.

> By default, when `definition.mappings` is an empty object, the server fallbacks (sic) to `dynamic: true`. But we don't want to implement this behavior client-side.

On a separate note, the [`createSearchIndexes` docs](https://www.mongodb.com/docs/manual/reference/command/createSearchIndexes/) themselves say that `definition.mappings` is an optional document and `dynamic: false` is the default. If that's inaccurate I think a DOCSP ticket may be warranted.

GromNaN:
Changing the target branch, as this would be a new feature.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
GromNaN: I've completely reworked the PR to throw an exception in case none of `dynamic` or `fields` is set.

jmikola: In this case, should it go back to targeting the 2.9.x branch as a bug fix? I wouldn't consider this new functionality.

lib/Doctrine/ODM/MongoDB/Mapping/MappingException.php
```diff
@@ -291,4 +291,9 @@ public static function nonBackedEnumMapped(string $className, string $fieldName,
             $fieldName,
         ));
     }
+
+    public static function emptySearchIndexDefinition(string $className, string $indexName): self
+    {
+        return new self(sprintf('%s search index "%s" must be dynamic or specify a field mapping.', $className, $indexName));
```
jmikola:
```suggestion
        return new self(sprintf('%s search index "%s" must be dynamic or specify a field mapping', $className, $indexName));
```

I don't think periods are used for single-sentence exception messages. Disregard if I'm mistaken.

tests/Doctrine/ODM/MongoDB/Tests/Mapping/ClassMetadataTest.php
```diff
@@ -971,6 +971,15 @@ public function testDefaultValueForValidationLevel(): void
         $cm = new ClassMetadata('stdClass');
         self::assertEquals(ClassMetadata::SCHEMA_VALIDATION_LEVEL_STRICT, $cm->getValidationLevel());
     }
+
+    public function testEmptySearchIndexDefinition(): void
+    {
+        $cm = new ClassMetadata('stdClass');
+
+        $this->expectException(MappingException::class);
+        $this->expectExceptionMessage('stdClass search index "default" must be dynamic or specify a field mapping.');
```
jmikola:
```suggestion
        $this->expectExceptionMessage('stdClass search index "default" must be dynamic or specify a field mapping');
```

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -1220,9 +1220,15 @@ public function hasIndexes(): bool
      */
     public function addSearchIndex(array $definition, ?string $name = null): void
     {
+        $name ??= self::DEFAULT_SEARCH_INDEX_NAME;
+
+        if (empty($definition['mappings']['dynamic']) && empty($definition['mappings']['fields'])) {
```
jmikola, line 1225: Using `array_key_exists()` here would most directly address the original bug. Both [XmlDriver](https://github.com/doctrine/mongodb-odm/blob/2.9.0/lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php#L583) an [Attribute](https://github.com/doctrine/mongodb-odm/blob/2.9.0/lib/Doctrine/ODM/MongoDB/Mapping/Driver/AttributeDriver.php#L358) driver should only set these keys if the mappings actually specified a value, so I don't think we need to prevent users from specifying `dynamic: false` and empty fields.

jmikola (commented):
Let me know if you disagree about using `array_key_exists()` and I'll concede and LGTM.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
GromNaN: I want to check if `$definition['mappings']['dynamic']` is `false` also. 
The condition would be:

```suggestion
        if (! array_key_exists('mappings', $definition) || ((! array_key_exists('dynamics', $definition['mappings']) || ! $definition['mappings']) && (! array_key_exists('fields', $definition['mappings']) || ! is_array($definition['mappings']['fields']) || count($definition['mappings']['fields']) === 0)) {
```

It don't think this makes the code more explicit, secure or performant.

jmikola: My understanding was that the original bug only cared about both options being unspecified, so two `array_key_exists()` checks would be the minimum change to implement. ODM wasn't prohibiting `dynamic: true` with a non-empty `fields` array (I assume the server reports an error there), so I assumed the most minimal validation would be OK.

If you instead want to prohibit `dynamic: false` and an empty `fields` array, the original condition was fine (and certainly more readable). Since we know the server doesn't report an error for that, I suppose having ODM validate this will protect users from creating a useless index.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
jmikola: @GromNaN: Bumping this. If you're just adding extra validation, is 2.9.x appropriate?

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
GromNaN: I totally messed with github UI patches. I've reverted this change to validate the mappings is correct (dynamic or fields)

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
GromNaN: Since I'm adding a new validation, I think 2.10.x is better. Currently, indexes with `#[SearchIndex(dynamic: false)]` are accepted (even if they are useless).

jmikola (approved):

PR #2692, opened 2024-10-19
Switch proxies back to Doctrine Persistence and `LazyGhostTrait`

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | -

#### Summary

This PR replaces the proxy implementation from `friendsofphp/proxy-manager-lts` (which requires `laminas/laminas-code`) with `doctrine/persistence` and the `LazyGhostTrait` from `symfony/var-exporter`.

Copy the proxy implementation from Doctrine ORM: https://github.com/doctrine/orm/blob/3.3.x/src/Proxy/ProxyFactory.php

The `Doctrine\Persistence\Proxy` class was added in `doctrine/persistence: 3.1`, we already require `3.2`.

This is a breaking change, so I open this PR for 3.0. The opposite move was done between 1.x and 2.0: https://github.com/doctrine/mongodb-odm/pull/1875, but that was before Nicolas work on lazy ghost objects.


tests/Doctrine/ODM/MongoDB/Tests/Functional/ReferencesTest.php
```diff
@@ -502,8 +501,13 @@ public function testDocumentNotFoundEvent(): void
...
 
-        self::assertInstanceOf(LazyLoadingInterface::class, $profile);
-        $profile->initializeProxy();
+        self::assertInstanceOf(Proxy::class, $profile);
+        $this->expectException(DocumentNotFoundException::class);
+        $this->expectExceptionMessage(
+            'The "Documents\Profile" document with identifier ' .
+            '"abcdefabcdefabcdefabcdef" could not be found.',
+        );
+        $profile->__load();
```
GromNaN, line 510: I don't understand why there was no exception for this missing document, while there is for previous tests.

lib/Doctrine/ODM/MongoDB/Proxy/Autoloader.php
GromNaN, line 1: This class my not be required for this PR, but it enable serialization/unserialization of proxy objects. The doc need to be duplicated for the ODM: https://www.doctrine-project.org/projects/doctrine-orm/en/3.3/reference/advanced-configuration.html#autoloading-proxies

lib/Doctrine/ODM/MongoDB/Proxy/Factory/ProxyFactory.php
GromNaN, line 1: This interface could be deprecated in `2.10.x`. I don't think we need more than a single `ProxyFactory`.

GromNaN: Or maybe this will be useful again, if we can have an optimized `ProxyFactory` for PHP 8.4 with [Lazy Objects](https://wiki.php.net/rfc/lazy-objects).

malarzm:
Wouldn't it be better to wait for PHP's native proxies? Or change from Symfony's implementation to native one will be transparent for end users?

GromNaN:
My main motivation is to get rid of laminas code dependency and converge to the same implementation as the ORM. This change is compatible with every supported PHP version whereas the new native lazy objects are PHP 8.4+ only. 

I'm novice in the lazy ghost objects domain. Maybe @nicolas-grekas would have a better insight.

My guess is that we are going to have 2 ProxyFactory classes: one for PHP 8.4+ and one for older versions. But both would use var-exporter and doctrine/persistence.

nicolas-grekas:
I think it should be possible to delegate all the native proxy handling to var-exporter (+doctrine/persistence). That's where the choice of the best implement should happen.
Then, when you'll have PHP 8.4 as a minimum dependency, the lib should go native-only.
The only thing to keep in mind is making as much things internal as possible to make the switch to native-only easier.

malarzm:
FWIW we should do absolutely everything, to not release this as 3.0 and have 4.0 few months later.

alcaeus:
> FWIW we should do absolutely everything, to not release this as 3.0 and have 4.0 few months later.

I agree. I haven't had a chance to take a detailed look yet, but I think it should be possible to somehow make this happen in 2.x.

GromNaN:
Replaced by https://github.com/doctrine/mongodb-odm/pull/2700

PR #2696, opened 2024-10-28
Ensure compatibility with persistence 4

greg0ire:
This is mostly about adding return type declarations. In one instance though, it is more than that: ClassMetadataFactory redeclares $cacheSalt, a protected property inherited from the persistence package. Since it is not possible to widen or narrow the type, and since the redeclaration seems to be about setting a default value, let us set it in the constructor.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
         "doctrine/collections": "^1.5 || ^2.0",
         "doctrine/event-manager": "^1.0 || ^2.0",
         "doctrine/instantiator": "^1.1 || ^2",
-        "doctrine/persistence": "^3.2",
+        "doctrine/persistence": "^3.2 || ^4",
         "friendsofphp/proxy-manager-lts": "^1.0",
         "jean85/pretty-package-versions": "^1.3.0 || ^2.0.1",
         "mongodb/mongodb": "^1.17.0",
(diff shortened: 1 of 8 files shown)


GromNaN (approved):

GromNaN:
You can update the `doctrine/persistence` constraint in composer.json to allow version 4.

greg0ire:
@GromNaN I just did, then saw that `doctrine/orm` is a require-dev dependency, so we will have to wait for it to support v4 before being able to truly test v4 here. That's in progress at https://github.com/doctrine/orm/pull/11533

derrabus:
Are there any more blockers for this PR?

greg0ire:
The release of `doctrine/orm` 3.4.0 is the main blocker.

greg0ire:
As discussed internally, v4 compat was accidentally released as `doctrine/orm` 3.3.1, so there no longer is a blocker, it would seem.

greg0ire:
Here! There is at least [one job](https://github.com/doctrine/mongodb-odm/actions/runs/12895787166/job/35957449687?pr=2696#step:9:111) using it.

GromNaN (approved):
In fact, all jobs runs `doctrine/persistence: 4.0.0` except the one requiring "lowest" dependencies.

LGTM.

Issue #2132, opened 2019-12-04
Proxy generation and autoloading


olvlvl:
|    Q        |   A
|------------ | -----
| Version     | 2.0

### Support Question

I originally created this issue over here: https://github.com/Ocramius/ProxyManager/issues/531.

We recently upgraded  to `mongodb-odm` v2 and we're facing an issue with the proxy generation. You see, we generate the hydrators and proxies before building our docker images, and the _cache_ folder is readonly after this point. `ProxyManager\Factory\AbstractBaseFactory::generateProxy()` uses `class_exists()` to determine if a proxy needs to be generated or not, and after this point it seems there's nothing that checks that the file actually needs to be generated, and the application crashes when the proxy manager tries to move the temporary file into the readonly cache folder.

https://github.com/Ocramius/ProxyManager/blob/4d7be24018475d2627ab74078e912ca9ef5887d2/src/ProxyManager/Factory/AbstractBaseFactory.php#L70

Ocraminus recommended using `classmap` to autoload the proxies. It works, but it makes working with our application complicated because we need to run `compose dump --no-dev` before we generate the proxies (so the classmap doesn't contain outdated references) and run it again after so the classmap includes the generated files.

We don't have to do any of this for hydrators, could there be a nicer solution that doesn't involve using `classmap`?

alcaeus:
Do you use DoctrineMongoDBBundle for Symfony or do you configure ODM yourself? If you use Symfony and the bundle, you can change this by updating the proxy generation strategy to not auto-generate proxies. This will register an autoloader for proxies in the bundle: https://github.com/doctrine/DoctrineMongoDBBundle/blob/master/DoctrineMongoDBBundle.php#L63..L70.

Then, you can pair this with the `doctrine:mongodb:generate:proxies` command to generate proxies while building your docker container.

If you don't use Symfony, use the `odm:generate:proxies` command to dump proxies and register the autoloader yourself.

olvlvl:
We don't use Symfony, we configure the ODM ourselves, and we use [this package](https://github.com/olvlvl/doctrine-generators) to generate the hydrators and the proxies. I'll have a look at `odm:generate:proxies`, but are you suggesting we should register an additional autoloader for the proxies? We're not doing anything special for the hydrators, and the proxies used to work fine. Would you be so kind to point me to an example in the documentation?

alcaeus:
[...]

PR #2700, opened 2024-11-23
Switch proxies to `LazyGhostTrait`

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | Replace #2692

#### Summary

Use the `LazyGhostTrait` from `symfony/var-exporter` to generate lazy proxy classes.

- [x] Add `LazyGhostProxyFactory` that implements the `ProxyFactory` interface and use it by default
- [x] Add configuration to switch back to Proxy Manager, and run the test suite with this it
- [x] Deprecate using proxy manager, to be removed in 3.0
- [x] Add documentation (including [Autoloading Proxies](https://www.doctrine-project.org/projects/doctrine-orm/en/3.3/reference/advanced-configuration.html#autoloading-proxies))
- [x] Review implementation to ensure property hooks and native lazy ghost will not require any breaking change. Using the ORM PRs as reference implementation https://github.com/doctrine/orm/pull/11652 and https://github.com/doctrine/orm/pull/11659
- [x] Fix job names for the new matrix variable `lazy-ghost-objects`


lib/Doctrine/ODM/MongoDB/Proxy/Resolver/LazyGhostProxyClassNameResolver.php
```diff
@@ -0,0 +1,31 @@
...
+namespace Doctrine\ODM\MongoDB\Proxy\Resolver;
+
+use Doctrine\Persistence\Mapping\ProxyClassNameResolver;
+use Doctrine\Persistence\Proxy;
+
+use function strrpos;
+use function substr;
+
+/** @internal */
+class LazyGhostProxyClassNameResolver implements ClassNameResolver, ProxyClassNameResolver
```
GromNaN, line 14: This class is copied from [`Doctrine\ORM\Proxy\DefaultProxyClassNameResolver`](https://github.com/doctrine/orm/blob/3.3.x/src/Proxy/DefaultProxyClassNameResolver.php).

lib/Doctrine/ODM/MongoDB/Proxy/InternalProxy.php
```diff
@@ -0,0 +1,18 @@
...
+
+use Doctrine\Persistence\Proxy;
+
+/**
+ * @internal
+ *
+ * @template T of object
+ * @template-extends Proxy<T>
+ */
+interface InternalProxy extends Proxy
```
GromNaN, line 15: This interface is copied from [Doctrine\ORM\Proxy\InternalProxy](https://github.com/doctrine/orm/blob/3.3.x/src/Proxy/InternalProxy.php).

lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -589,14 +577,35 @@ public function getPersistentCollectionGenerator(): PersistentCollectionGenerato
...
+                break;
+            case self::AUTOGENERATE_EVAL:
+                $proxyManagerConfiguration->setGeneratorStrategy(new EvaluatingGeneratorStrategy());
+
+                break;
+            default:
+                throw new InvalidArgumentException('Invalid proxy generation strategy given - only AUTOGENERATE_FILE_NOT_EXISTS and AUTOGENERATE_EVAL are supported.');
+        }
+
+        return $proxyManagerConfiguration;
```
GromNaN: Creating `ProxyManagerConfiguration` on-demand as the class is not required when LazyGhostObjects are used.

lib/Doctrine/ODM/MongoDB/Proxy/Autoloader.php
```diff
@@ -0,0 +1,90 @@
...
+use function substr;
+
+use const DIRECTORY_SEPARATOR;
+
+/**
+ * Special Autoloader for Proxy classes, which are not PSR-0 compliant.
+ *
+ * @internal
+ */
+final class Autoloader
```
GromNaN, line 26: This class is copied from [Doctrine\ORM\Proxy\Autoloader](https://github.com/doctrine/orm/blob/3.3.x/src/Proxy/Autoloader.php).

lib/Doctrine/ODM/MongoDB/Proxy/Factory/LazyGhostProxyFactory.php
```diff
@@ -0,0 +1,385 @@
...
+
+            if (! $original && ! $factory->lifecycleEventManager->documentNotFound($proxy, $identifier)) {
+                throw DocumentNotFoundException::documentNotFound($classMetadata->getName(), $identifier);
+            }
+
+            // phpcs:ignore SlevomatCodingStandard.ControlStructures.EarlyExit.EarlyExitNotUsed
+            if ($proxy instanceof NotifyPropertyChanged) {
+                $proxy->addPropertyChangedListener($factory->uow);
+            }
+        };
```
GromNaN, line 199: Initialization code copied from [`StaticProxyFactory`](https://github.com/doctrine/mongodb-odm/blob/0b05bf2f9f629a7dc462176734b32f02b8c8ab37/lib/Doctrine/ODM/MongoDB/Proxy/Factory/StaticProxyFactory.php#L112-L137).

lib/Doctrine/ODM/MongoDB/Proxy/Factory/ProxyFactory.php
```diff
@@ -19,9 +18,9 @@ public function generateProxyClasses(array $classes): int;
      * @param mixed $identifier
      * @phpstan-param ClassMetadata<T> $metadata
      *
-     * @return T&GhostObjectInterface<T>
+     * @return T
      *
      * @template T of object
      */
-    public function getProxy(ClassMetadata $metadata, $identifier): GhostObjectInterface;
+    public function getProxy(ClassMetadata $metadata, $identifier): object;
```
GromNaN, line 25: The interface is modified, but child classes can be more restrictive. This is not a BC break for implementers of the interface, but it can be for users of the interface.

tests/Doctrine/ODM/MongoDB/Tests/BaseTestCase.php
```diff
@@ -114,6 +117,11 @@ public static function assertArraySubset(array $subset, array $array, bool $chec
         }
     }
 
+    public static function isLazyObject(object $document): bool
+    {
+        return $document instanceof InternalProxy || $document instanceof LazyLoadingInterface;
+    }
```
GromNaN, line 123: Most test changes are related to check interfaces to assert if the class is a lazy proxy object. I wonder if this method should be in UOW.

Later, it will use `ReflectionClass::getLazyInitializer()` to detect native lazy object.

lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -608,6 +617,32 @@ public function isTransactionalFlushEnabled(): bool
...
+            );
+        }
+
+        $this->useLazyGhostObject = $flag;
+    }
+
+    public function isLazyGhostObjectEnabled(): bool
+    {
+        return $this->useLazyGhostObject ?? true;
+    }
```
GromNaN: Should we use the same method names as the ORM: [`isLazyGhostObjectEnabled` and `setLazyGhostObjectEnabled`](https://github.com/doctrine/orm/blob/d6a8298652c6ec4dd89d199d9033d0b045e3f94c/src/Configuration.php#L614-L633)?

lib/Doctrine/ODM/MongoDB/Proxy/Factory/LazyGhostProxyFactory.php
```diff
@@ -0,0 +1,385 @@
...
+            // no break
+            case Configuration::AUTOGENERATE_FILE_NOT_EXISTS:
+                if (file_exists($fileName)) {
+                    break;
+                }
+            // no break
+            case Configuration::AUTOGENERATE_ALWAYS:
+                $this->generateProxyClass($class, $fileName, $proxyClassName);
+                break;
+        }
```
GromNaN: It would be more intuitive this way:
```suggestion
        if (
            match ($this->autoGenerate) {
                Configuration::AUTOGENERATE_FILE_NOT_EXISTS_OR_CHANGED => ! file_exists($fileName) || filemtime($fileName) < filemtime($class->getReflectionClass()->getFileName()),
                Configuration::AUTOGENERATE_FILE_NOT_EXISTS => ! file_exists($fileName),
                Configuration::AUTOGENERATE_ALWAYS => true,
                Configuration::AUTOGENERATE_NEVER => false,
            }
        ) {
            $this->generateProxyClass($class, $fileName, $proxyClassName);
        }
```

lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -608,6 +617,32 @@ public function isTransactionalFlushEnabled(): bool
...
+    public function setUseLazyGhostObject(bool $flag): void
+    {
+        if ($flag === false) {
+            if (! class_exists(ProxyManagerConfiguration::class)) {
+                throw new LogicException('Package "friendsofphp/proxy-manager-lts" is required to disable LazyGhostObject.');
+            }
+
+            trigger_deprecation(
+                'doctrine/mongodb-odm',
+                '2.6',
```
GromNaN:
```suggestion
                '2.10',
```

SenseException:
Considering the unchecked checkboxes in your initial PR description, is this a WIP?

GromNaN:
I've just rebased this PR.

The bundle will need an update to [register the autoloader](https://github.com/doctrine/mongodb-odm/issues/2132#issuecomment-561643472).

Regarding native lazy objects and prop hooks PR on ODM, I think the approaches are not conflicting this this PR.

lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -589,14 +577,35 @@ public function getPersistentCollectionGenerator(): PersistentCollectionGenerato
...
 
+    /** @deprecated */
     public function buildGhostObjectFactory(): LazyLoadingGhostFactory
     {
-        return new LazyLoadingGhostFactory(clone $this->getProxyManagerConfiguration());
+        return new LazyLoadingGhostFactory($this->getProxyManagerConfiguration());
     }
 
+    /** @deprecated */
     public function getProxyManagerConfiguration(): ProxyManagerConfiguration
```
alcaeus, line 593: Note that this will return a new configuration instance every time this is called. I don't recall if there were any issues with that approach (e.g. due to unique identifiers in directories), so it's worth checking that creating a separate configuration doesn't cause the proxy manager to look for proxy files in different directories.

lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -608,6 +617,32 @@ public function isTransactionalFlushEnabled(): bool
...
+                'Using "friendsofphp/proxy-manager-lts" is deprecated. Use "symfony/var-exporter" LazyGhostObjects instead.',
+            );
+        }
+
+        $this->useLazyGhostObject = $flag;
+    }
+
+    public function isLazyGhostObjectEnabled(): bool
+    {
+        return $this->useLazyGhostObject ?? true;
```
alcaeus: This property is initialised as `true`, and since it's non-nullable the null coalesce is not necessary.

```suggestion
        return $this->useLazyGhostObject;
```

lib/Doctrine/ODM/MongoDB/Proxy/Factory/LazyGhostProxyFactory.php
```diff
@@ -0,0 +1,380 @@
...
+        if (is_int($autoGenerate) && ($autoGenerate < 0 || $autoGenerate > 4)) {
+            throw new InvalidArgumentException(sprintf('Invalid auto generate mode "%s" given.', is_scalar($autoGenerate) ? (string) $autoGenerate : get_debug_type($autoGenerate)));
+        }
+
+        $this->uow                   = $dm->getUnitOfWork();
+        $this->autoGenerate          = (int) $autoGenerate;
+        $this->lifecycleEventManager = new LifecycleEventManager($dm, $this->uow, $dm->getEventManager());
+    }
+
+    /** @param array<mixed> $identifier */
```
alcaeus: According to `ProxyFactory::getProxy`, the type for `$identifier` is `mixed`, not an array, and if I recall correctly we always only pass the `_id` field to this method. This may be a leftover from ORM which needs to support compound primary keys.

phpstan-baseline.neon
```diff
@@ -545,6 +545,11 @@ parameters:
 			count: 1
 			path: lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
 
+		-
+			message: "#^Call to an undefined static method Doctrine\\\\ODM\\\\MongoDB\\\\Proxy\\\\Factory\\\\LazyGhostProxyFactory\\:\\:createLazyGhost\\(\\)\\.$#"
```
alcaeus, line 549: Do you know why this error appears? At first glance, it shouldn't happen 🤔

alcaeus (commented):
Just some minor comments on this. I think the "BC Break" in the ProxyFactory interface is fine - proxies are an implementation detail and as long as we communicate that we change the default proxy mechanism we'll be fine. People shouldn't rely on a specific proxy behaviour, other than "it behaves as if it was an instance of the original class".

lib/Doctrine/ODM/MongoDB/Configuration.php
GromNaN: I updated to keep the instance in memory so that it's still possible to customize the configuration of the proxy manager directly. But it is reset each time a configuration is set on the ODM configuration class.

lib/Doctrine/ODM/MongoDB/Proxy/Factory/LazyGhostProxyFactory.php
GromNaN: Correct, that's the type of `ClassMetadata::getIdentifierObject()` which is `mixed`.

phpstan-baseline.neon
GromNaN: The closure is created in the scope of `LazyGhostProxyFactory` but bound into a class that uses `LazyGhostTrait`. https://github.com/doctrine/mongodb-odm/pull/2700/files#diff-b97724573bc93916d65bdebb0be2e5dcff48caf916fa6ac8b1dd92785c449fb6R240-R249

alcaeus (approved):
Changes LGTM. @malarzm do you want to take another look or should we merge this for 2.10?

IonBazan (approved):

GromNaN:
I just found an incompatibility with DoctrineMongoDBODMBundle, the `ProxyManager\Proxy\GhostObjectInterface` interface [is hardcoded](https://github.com/doctrine/DoctrineMongoDBBundle/blob/783d281cca2b24943b3351cfe5c943364694e2b0/config/mongodb.php#L57). I think we have to keep lts-proxy-manager by default and add change the default configuration in a new version of `DoctrineMongoDBODMBundle`.

PR #2702, opened 2024-11-28
Improve query and aggregator classes

### LINKED ISSUE #2699: `Aggregation` class is final and can't be mocked to mock `Aggregation\Builder`
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.9.0

#### Summary

The method [`Doctrine\ODM\MongoDB\Aggregation\Builder::execute()` is deprecated](https://github.com/doctrine/mongodb-odm/blob/0b05bf2f9f629a7dc462176734b32f02b8c8ab37/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php#L175-L179) and should be replaced by `$builder->getAggregation()->getIterator()` (Note that it's more verbose for the common use case).

But [`getAggregation()` must return an instance of `Doctrine\ODM\MongoDB\Aggregation\Aggregation`](https://github.com/doctrine/mongodb-odm/blob/0b05bf2f9f629a7dc462176734b32f02b8c8ab37/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php#L250). This class is hard to instantiate in tests as it requires a lot of dependencies that would have to be mocked with the ODM internal behavior. Since the class is `final`, it cannot be mocked.

I created a draft PR to remove usage of `Aggregation\Builder::execute()` in API Platform, but the tests cannot be fixed. https://github.com/GromNaN/api-platform-core/pull/2

The solutions would be to remove the `final` keyword on `Doctrine\ODM\MongoDB\Aggregation\Aggregation`.

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | Fixes #2699 (cc @soyuka)

#### Summary

To improve handling of query and aggregation classes, a new joint `IterableResult` interface is introduced. This method includes the `execute`, `getIterator`, and `getSingleResult` methods that are common to both queries and aggregations. In addition, this allows us to keep the actual classes final, while allowing other platforms to mock what's returned from the builders for testing purposes.

Note that while the return type of `getAggregation` and `getQuery` in the builders has widened, I've added an `@return` statement to indicate that the methods themselves still return the same class instances that they returned previously. However, changing the return type is necessary to allow the mocking described above. So while this is technically a BC break, in practice it has no effect, especially as any class extending the builders would still be allowed to use the narrower return type.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
@@ -6,6 +6,7 @@ namespace Doctrine\ODM\MongoDB\Aggregation;
 
 use Doctrine\ODM\MongoDB\Aggregation\Stage\Sort;
 use Doctrine\ODM\MongoDB\DocumentManager;
+use Doctrine\ODM\MongoDB\Iterator\IterableResult;
 use Doctrine\ODM\MongoDB\Iterator\Iterator;
 use Doctrine\ODM\MongoDB\Mapping\ClassMetadata;
 use Doctrine\ODM\MongoDB\Persisters\DocumentPersister;
@@ -246,8 +247,10 @@ class Builder
      * Returns an aggregation object for the current pipeline
      *
      * @param array<string, mixed> $options
+     *
+     * @return Aggregation
      */
-    public function getAggregation(array $options = []): Aggregation
+    public function getAggregation(array $options = []): IterableResult
     {
         $class = $this->hydrationClass ? $this->dm->getClassMetadata($this->hydrationClass) : null;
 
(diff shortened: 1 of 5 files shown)


lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
```diff
@@ -246,8 +247,10 @@ public function geoNear($x, $y = null): Stage\GeoNear
      * Returns an aggregation object for the current pipeline
      *
      * @param array<string, mixed> $options
+     *
+     * @return Aggregation
```
GromNaN, line 251: Noted ~that this is necessary for static analysis and autocompletion, but~ we don't use it as return type to allow mocking more easily.

GromNaN (commented):
I think the issue it that `Aggregation\Builder::execute()` should not be deprecated. This was the perfect method for a mock.
As you can see in https://github.com/GromNaN/api-platform-core/pull/2, the alternative is more verbose, that's not good for DX.

With this solution, you have to rely on the phpdoc to give the exact return type, which is a regression. Can we revisite https://github.com/doctrine/mongodb-odm/pull/2211 instead?

GromNaN (approved):
On reviewing this PR, I finally agree with this solution, which resolves the blockage while providing a clear API.

PR #2706, opened 2024-12-13
doc: Fix simple search engine cookbook

bikusta:
|      Q       |   A
|------------- | -----------
| Type         | doc
| BC Break     | no
| Fixed issues | -

#### Summary

Use array instead of ArrayCollection. It is not possible to use an ArrayCollection with a collection field type. This would result in a runtime exception: "Collection type requires value of type array or null, Doctrine\Common\Collections\ArrayCollection given (Doctrine\ODM\MongoDB\MongoDBException)".

Conversation with @malarzm in the Slack channel #mongodm-odm on 12/12/2024:

> th docs are wrong, public Collection $keywords; has to be public array $keywords; and used as such

### DIFF
--- a/docs/en/cookbook/simple-search-engine.rst
+++ b/docs/en/cookbook/simple-search-engine.rst
@@ -27,7 +27,7 @@ setup a document like the following with a ``$keywords`` property that is mapped
 
         #[Field(type: 'collection')]
         #[Index]
-        public array $keywords;
+        public array $keywords = [];
 
         public function __construct()
         {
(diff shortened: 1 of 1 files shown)


docs/en/cookbook/simple-search-engine.rst
```diff
@@ -25,14 +25,13 @@ setup a document like the following with a ``$keywords`` property that is mapped
         #[Field(type: 'string')]
         public string $title;
 
-        /** @var Collection<string> */
         #[Field(type: 'collection')]
         #[Index]
-        public Collection $keywords;
+        public array $keywords;
```
GromNaN: We can simplify the example and remove the constructor.
```suggestion
        public array $keywords = [];
```

GromNaN (approved):
Thank you for taking the time to fix the doc.

PR #2708, opened 2024-12-16
Update BSON PHP Link

rickogden:
The page no longer exists for the original link, so updated it to the new URL.

|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues |


alcaeus (approved):
Thank you!

Issue #2105, opened 2019-11-11
Replace generators with IteratorIterator in custom iterators


alcaeus:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | improvement
| RFC         | no
| BC Break    | no

#### Summary

The `HydratingIterator` and `CachingIterator` classes use generators to wrap the given traversable. As described in #2092 and fixed in #2100, this can leak memory if the iterator class is destroyed while the generator isn't exhausted completely.

`IteratorIterator` can be used instead, but since there are some changes to generators with respect to rewind, the change shouldn't be done in a patch release.

soyuka:
What's the status on this @alcaeus ?

alcaeus:
The status is the same as five years ago, except that I have no recollection of what the issue was 🤦‍♂️ 

Did you encounter an issue around iterators that needs addressing?

soyuka:
I think we should close this, I'll open a new one if I can reproduce!

alcaeus:
No complaints from me 😊

PR #2709, opened 2024-12-18
Replace `wrapTransversable` generators to prevent memory leaks

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #2105

#### Problem

A non-static generator created with a method of a class instance holds the reference to this instance. When assigned to a property of this instance, this creates a circular reference (instance -> generator -> instance). The reference is destroyed only when the generator is exhausted.

Most of the time, when the memory limit is sufficient, this is not an issue. The garbage collector should remove this circular references, but it runs automatically only when reaching the [`GC_THRESHOLD_DEFAULT`](https://www.php.net/manual/en/features.gc.collecting-cycles.php). 

When working with a MongoDB Cursor, this is a single object which can fill up a lot of memory, that explains why `gc_collect_cycles()` is not run automatically.

#### Solution

Using `IteratorIterator` instead of `Generator`, we remove this circular reference . Also, I find the code simpler.
Same as https://github.com/mongodb/mongo-php-library/pull/694

#### Demonstration

In the following example, if `gc_collect_cycles()` is not called in the loop, we can see the memory leak. When calling `gc_collect_cycles()` explicitly, we can see 300 "collected" references. The memory leak disappears when the `Generator` is replaced by an `IteratorIterator`.

<details>
  <summary>Source code</summary>

```php
<?php

namespace Hell;

class Iterator implements \Iterator
{
    private iterable $iterator;

    public function __construct(\Traversable $data)
    {
        $this->iterator = $this->wrapTraversable($data);
        //$this->iterator = new \IteratorIterator($data);
    }

    public function current(): mixed
    {
        return $this->iterator->current();
    }

    public function next(): void
    {
        $this->iterator->next();
    }

    public function key(): mixed
    {
        return $this->iterator->key();
    }

    public function valid(): bool
    {
        return $this->iterator->valid();
    }

    public function rewind(): void
    {
        $this->iterator->rewind();
    }

    private function wrapTraversable(iterable $data): \Generator
    {
        foreach ($data as $key => $value) {
            yield $key => $value;
        }
    }
}

foreach (range(0, 100) as $i) {
    $data = new \ArrayObject(array_fill(0, 1_000_00, bin2hex(random_bytes(1_000))));
    $iterator = new Iterator($data);
    $iterator->next();
    unset($data);
    echo (memory_get_usage()).PHP_EOL;
}

gc_collect_cycles();
print_r(gc_status());
```

</details>

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Iterator/CachingIterator.php
+++ b/lib/Doctrine/ODM/MongoDB/Iterator/CachingIterator.php
@@ -5,7 +5,7 @@ declare(strict_types=1);
 namespace Doctrine\ODM\MongoDB\Iterator;
 
 use Countable;
-use Iterator;
+use Iterator as SPLIterator;
 use IteratorIterator;
 use ReturnTypeWillChange;
 use RuntimeException;
@@ -34,8 +34,8 @@ final class CachingIterator implements Countable, Iterator
     /** @var array<mixed, TValue> */
     private array $items = [];
 
-    /** @var Iterator<mixed, TValue>|null */
-    private ?Iterator $iterator;
+    /** @var SPLIterator<mixed, TValue>|null */
+    private ?SPLIterator $iterator;
 
     private bool $iteratorAdvanced = false;
 
@@ -136,8 +136,8 @@ final class CachingIterator implements Countable, Iterator
         }
     }
 
-    /** @return Iterator<mixed, TValue> */
-    private function getIterator(): Iterator
+    /** @return SPLIterator<mixed, TValue> */
+    private function getIterator(): SPLIterator
     {
         if ($this->iterator === null) {
             throw new RuntimeException('Iterator has already been destroyed');
--- a/lib/Doctrine/ODM/MongoDB/Iterator/CachingIterator.php
+++ b/lib/Doctrine/ODM/MongoDB/Iterator/CachingIterator.php
@@ -52,11 +52,7 @@ final class CachingIterator implements Countable, Iterator
     {
         $this->iterator = new IteratorIterator($iterator);
         $this->iterator->rewind();
-        if ($this->iterator->valid()) {
-            $this->storeCurrentItem();
-        } else {
-            $this->iterator = null;
-        }
+        $this->storeCurrentItem();
     }
 
     /** @see https://php.net/countable.count */
@@ -102,10 +98,6 @@ final class CachingIterator implements Countable, Iterator
             $this->iterator->next();
             $this->storeCurrentItem();
             $this->iteratorAdvanced = true;
-
-            if (! $this->iterator->valid()) {
-                $this->iterator = null;
-            }
         }
 
         next($this->items);
@@ -155,12 +147,12 @@ final class CachingIterator implements Countable, Iterator
      */
     private function storeCurrentItem(): void
     {
-        $key = $this->getIterator()->key();
+        $key = $this->iterator->key();
 
         if ($key === null) {
-            return;
+            $this->iterator = null;
+        } else {
+            $this->items[$key] = $this->getIterator()->current();
         }
-
-        $this->items[$key] = $this->getIterator()->current();
     }
 }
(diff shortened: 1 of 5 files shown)


lib/Doctrine/ODM/MongoDB/Iterator/CachingIterator.php
```diff
@@ -33,13 +34,11 @@ final class CachingIterator implements Countable, Iterator
...
     private array $items = [];
 
-    /** @var Generator<mixed, TValue>|null */
-    private ?Generator $iterator;
+    /** @var SPLIterator<mixed, TValue>|null */
+    private ?SPLIterator $iterator;
 
     private bool $iteratorAdvanced = false;
 
-    private bool $iteratorExhausted = false;
```
GromNaN, line 41: Replaced by setting `$iterator = null`, as it's already done for `UnrewindableIterator`

lib/Doctrine/ODM/MongoDB/Iterator/CachingIterator.php
```diff
@@ -51,8 +50,13 @@ final class CachingIterator implements Countable, Iterator
      */
     public function __construct(Traversable $iterator)
     {
-        $this->iterator = $this->wrapTraversable($iterator);
-        $this->storeCurrentItem();
+        $this->iterator = new IteratorIterator($iterator);
+        $this->iterator->rewind();
+        if ($this->iterator->valid()) {
```
GromNaN: This checks for a dead cursor.

lib/Doctrine/ODM/MongoDB/Iterator/CachingIterator.php
```diff
@@ -51,8 +50,13 @@ final class CachingIterator implements Countable, Iterator
...
+            $this->storeCurrentItem();
+        } else {
+            $this->iterator = null;
+        }
```
alcaeus: I believe this logic as well as the one contained could be inlined into `storeCurrentItem`. That method already contains logic to deal with an exhausted iterator (through checking the key) and does not store an item in that case - but it also doesn't mark the iterator as exhausted.

Instead of having the separate logic here, it would be nice to inline this logic into `storeCurrentItem` through a guard condition, added as a suggestion below. I've also added suggestions in other places related to this change.

```suggestion
        $this->storeCurrentItem();
```

I can't add comments on `storeCurrentItem` since it wasn't changed, but essentially the guard clause would also set `$this->iterator = null` if the iterator is no longer valid.

lib/Doctrine/ODM/MongoDB/Iterator/CachingIterator.php
```diff
@@ -94,9 +98,14 @@ public function key()
...
-        if (! $this->iteratorExhausted) {
-            $this->getIterator()->next();
+        if ($this->iterator !== null) {
+            $this->iterator->next();
             $this->storeCurrentItem();
+            $this->iteratorAdvanced = true;
+
+            if (! $this->iterator->valid()) {
+                $this->iterator = null;
+            }
```
alcaeus: Related to the `storeCurrentItem` refactoring:

```suggestion
        $this->iterator?->next();
        $this->storeCurrentItem();
        $this->iteratorAdvanced = true;
```

tests/Doctrine/ODM/MongoDB/Tests/Functional/Iterator/UnrewindableIteratorTest.php
```diff
@@ -100,6 +100,15 @@ public function testRewindAfterPartialIteration(): void
         iterator_to_array($iterator);
     }
 
+    public function testRewingAfterToArray(): void
```
alcaeus:
```suggestion
    public function testRewindingAfterToArray(): void
```

alcaeus (commented):
Great work refactoring this. I've left a suggestion to refactor `storeCurrentItems` to consolidate some logic, but LGTM aside from that.

lib/Doctrine/ODM/MongoDB/Iterator/CachingIterator.php
GromNaN: I've applied this change partially, as I think having an `if` block for the case of non-exhausted iterator is more explicit for the phase of reading the iterator.

alcaeus (approved):

PR #2711, opened 2024-12-19
Use short urls for all php.net links

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | -

#### Summary

This pages detect the user language.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Iterator/CachingIterator.php
+++ b/lib/Doctrine/ODM/MongoDB/Iterator/CachingIterator.php
@@ -55,7 +55,6 @@ final class CachingIterator implements Countable, Iterator
         $this->storeCurrentItem();
     }
 
-    /** @see https://www.php.net/countable.count */
     public function count(): int
     {
         $currentKey = key($this->items);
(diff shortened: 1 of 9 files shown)


lib/Doctrine/ODM/MongoDB/Iterator/CachingIterator.php
```diff
@@ -55,7 +55,7 @@ public function __construct(Traversable $iterator)
         $this->storeCurrentItem();
     }
 
-    /** @see https://php.net/countable.count */
+    /** @see https://www.php.net/countable.count */
```
GromNaN: Do you agree this links are useless and should be removed?

malarzm: This one definitely :)

malarzm: Yeah, all in this file

GromNaN: Done.

malarzm (approved):

PR #2715, opened 2025-01-10
Fix discriminatorValue 0

### LINKED ISSUE #2712: problem discriminatorValue with 0
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.10.0

#### Summary

Problem when having a "discriminatorValue" with value 0

#### Current behavior

When the query is being composed, it checks this condition in DocumentPersister.php:1473 

```
  if (! $key) {
      continue;
  }
```

However, this means that having a discriminatorValue with a value of 0 is not allowed. Shouldn't it be something like $key !== false instead?

#### How to reproduce

Perform any document search when child classes have a discriminator with a value of 0.
    
```
    const CONTENT_TYPE_NWS = 0;
    const CONTENT_TYPE_IMA = 1;
    const MAPPING = [
        self::CONTENT_TYPE_NWS => ContentNews::class,
        self::CONTENT_TYPE_IMA => ContentImage::class,
    ];

```

parada85:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | fix #2712

#### Summary

Fix discriminator value 0

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH2158Test.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH2158Test.php
@@ -1,0 +1,46 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ODM\MongoDB\Tests\Functional\Ticket;
+
+use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
+use Doctrine\ODM\MongoDB\Tests\BaseTestCase;
+
+class GH2158Test extends BaseTestCase
+{
+    public function testDiscriminatorMapCreationType(): void
+    {
+        $obj = new GH2158FirstType;
+        $this->dm->persist($obj);
+        $this->dm->flush();
+
+        self::assertEquals($this->dm->find(GH2158Abstract::class, $obj->getId()), $obj);
+    }
+}
+
+#[ODM\Document(collection: 'documents')]
+#[ODM\InheritanceType('SINGLE_COLLECTION')]
+#[ODM\DiscriminatorField('type')]
+#[ODM\DiscriminatorMap([0 => GH2158FirstType::class, 1 => GH2158SecondType::class])]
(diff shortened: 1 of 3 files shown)


tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH2158Test.php
```diff
@@ -0,0 +1,46 @@
...
+        $this->dm->flush();
+
+        self::assertEquals($this->dm->find(GH2158Abstract::class, $obj->getId()), $obj);
+    }
+}
+
+#[ODM\Document(collection: 'documents')]
+#[ODM\InheritanceType('SINGLE_COLLECTION')]
+#[ODM\DiscriminatorField('type')]
+#[ODM\DiscriminatorMap([0 => GH2158FirstType::class, 1 => GH2158SecondType::class])]
```
GromNaN, line 25: Using `int` as key is not permitted by [the `@param` annotation](https://github.com/doctrine/mongodb-odm/blob/bf9f88937bac27a8bac3d4a61e91c0c7b79a57ad/lib/Doctrine/ODM/MongoDB/Mapping/Annotations/DiscriminatorMap.php#L22), this rule is not enforced in the code.
I propose to relax the type.

```diff
-    /** @param array<string, class-string> $value */
+    /** @param array<class-string> $value */ 
```

GromNaN (commented):
Sounds good to me, once the type issue is solved.

GromNaN:
Also, this is a bug fix. Could you reset the PR on 2.9.x. This will also ensure that the tests pass with https://github.com/doctrine/mongodb-odm/pull/2714

GromNaN:
@parada85 you don't need to create a new PR, you can reset your branch with `git reset 2.9.x`, `git cherry-pick` your commit and `git push --force`. Then change the target of the PR.

(See https://github.com/doctrine/mongodb-odm/pull/2716)

PR #2716, opened 2025-01-10
Fix discriminatorValue key 0

### LINKED ISSUE #2712: problem discriminatorValue with 0

parada85:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | fix #2712

#### Summary

Fix discriminatorValue 0

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Mapping/Annotations/DiscriminatorMap.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/Annotations/DiscriminatorMap.php
@@ -16,7 +16,7 @@ use Doctrine\Common\Annotations\Annotation\NamedArgumentConstructor;
 #[Attribute(Attribute::TARGET_CLASS | Attribute::TARGET_PROPERTY)]
 final class DiscriminatorMap implements Annotation
 {
-    /** @var array<string, class-string> */
+    /** @var array<class-string> */
     public $value;
 
     /** @param array<string, class-string> $value */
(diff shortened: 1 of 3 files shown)


GromNaN (approved):

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/DiscriminatorMap.php
```diff
@@ -16,7 +16,7 @@
 #[Attribute(Attribute::TARGET_CLASS | Attribute::TARGET_PROPERTY)]
 final class DiscriminatorMap implements Annotation
 {
-    /** @var array<string, class-string> */
+    /** @var array<class-string> */
     public $value;
 
     /** @param array<string, class-string> $value */
```
GromNaN:
```suggestion
    /** @param array<class-string> $value */
```

GromNaN:
Thank you @parada85

PR #2725, opened 2025-02-04
Root document Entity Map

parada85:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.10.x

#### Summary

Problem class inheritance save in entitymap 

#### How to reproduce

I have a simple class inheritance like this:

```
class Section {}


class Tag extends Section {}
```

Then I do something like this:

```
/** @var DocumentManager $manager */
$manager = Kernel::stGetObj()->getContainer()->get('doctrine_mongodb')->getManager();
$manager->getRepository(Section::class)->find(68);
$manager->getRepository(Section::class)->find(68);
```
It seems like two database queries are made, but only one should be executed since it should be cached in the entity map. The issue seems to be that when it looks up the entity map, it does so using the repository class, but then stores it with the actual class (the child class) instead.

```
18:33:35 DEBUG     [doctrine] MongoDB command: {"find":"section","filter":{"_id":68,"type":{"$in":[3,4]}},"limit":1,"$db":"db","lsid":{"id":{"$binary":"R5St6qBWSAqg\/2jT16lktA==","$type":"04"}}}
18:33:35 DEBUG     [doctrine] MongoDB command: {"find":"section","filter":{"_id":68,"type":{"$in":[3,4]}},"limit":1,"$db":"db","lsid":{"id":{"$binary":"R5St6qBWSAqg\/2jT16lktA==","$type":"04"}}}
```

However, if I do this:

```
/** @var DocumentManager $manager */
$manager = Kernel::stGetObj()->getContainer()->get('doctrine_mongodb')->getManager();
$manager->getRepository(Tag::class)->find(68);
$manager->getRepository(Tag::class)->find(68);
```

It works correctly, and only one query is made.

When fetching the same document twice, only one database query should be executed if the entity is already cached in the entity map, regardless of whether the parent or child class is used.

To resolve the issue, I have overridden the find method of the DocumentRepository and do the following:
```
$parentClass = $this->getReflectionParentClass(new \ReflectionClass($documentName));        
if ($map = $this->getDocumentManager()->getClassMetadata($parentClass->getName())->discriminatorMap) {
     foreach ($map as $class) {
        if ($ret = $this->getDocumentManager()->getUnitOfWork()->tryGetById($id, $this->getDocumentManager()->getClassMetadata($class))) {
            return $ret;
          }
     }
}
```
but method tryGetById is marked internal :(

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
+++ b/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
@@ -1505,7 +1505,7 @@ final class UnitOfWork implements PropertyChangedListener
         $class = $this->dm->getClassMetadata($document::class);
         $id    = $this->getIdForIdentityMap($document);
 
-        if ($class->isEmbeddedDocument || isset($this->identityMap[$class->rootDocumentName][$id])) {
+        if (isset($this->identityMap[$class->rootDocumentName][$id])) {
             return false;
         }
 
--- a/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
+++ b/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
@@ -1505,7 +1505,7 @@ final class UnitOfWork implements PropertyChangedListener
         $class = $this->dm->getClassMetadata($document::class);
         $id    = $this->getIdForIdentityMap($document);
 
-        if (isset($this->identityMap[$class->rootDocumentName][$id])) {
+        if ($class->isEmbeddedDocument || isset($this->identityMap[$class->rootDocumentName][$id])) {
             return false;
         }
 
--- a/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
+++ b/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
@@ -1505,11 +1505,11 @@ final class UnitOfWork implements PropertyChangedListener
         $class = $this->dm->getClassMetadata($document::class);
         $id    = $this->getIdForIdentityMap($document);
 
-        if (isset($this->identityMap[$class->name][$id])) {
+        if (isset($this->identityMap[$class->rootDocumentName][$id])) {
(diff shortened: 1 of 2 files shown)


parada85:
I am having issues with embedded documents because I don't know how the entity map should behave with these types of documents.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -1505,11 +1505,11 @@ public function addToIdentityMap(object $document): bool
         $class = $this->dm->getClassMetadata($document::class);
         $id    = $this->getIdForIdentityMap($document);
 
-        if (isset($this->identityMap[$class->name][$id])) {
+        if ($class->isEmbeddedDocument || isset($this->identityMap[$class->rootDocumentName][$id])) {
             return false;
         }
 
-        $this->identityMap[$class->name][$id] = $document;
+        $this->identityMap[$class->rootDocumentName][$id] = $document;
```
malarzm, line 1512: I don't remember details about what is stored in `rootDocumentName` but at a glance now there can be a collision for `COLLECTION_PER_CLASS` inheritance type as it's perfectly possible for two separate documents to have same id

malarzm:
> I am having issues with embedded documents because I don't know how the entity map should behave with these types of documents.

Embedded documents are not reusable in any way, no can exist on their own. If they weren't included in the identity map, they still shouldn't be. If they were, we were most likely using spl_object_hash/id and the change should be pretty transparent

parada85:
> > I am having issues with embedded documents because I don't know how the entity map should behave with these types of documents.
> 
> Embedded documents are not reusable in any way, no can exist on their own. If they weren't included in the identity map, they still shouldn't be. If they were, we were most likely using spl_object_hash/id and the change should be pretty transparent

PR #2726, opened 2025-02-04
Fix root document entityMap

parada85:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.9.2

#### Summary

Problem class inheritance save in entitymap 

#### How to reproduce

I have a simple class inheritance like this:

```
class Section {}


class Tag extends Section {}
```

Then I do something like this:

```
/** @var DocumentManager $manager */
$manager = Kernel::stGetObj()->getContainer()->get('doctrine_mongodb')->getManager();
$manager->getRepository(Section::class)->find(68);
$manager->getRepository(Section::class)->find(68);
```
It seems like two database queries are made, but only one should be executed since it should be cached in the entity map. The issue seems to be that when it looks up the entity map, it does so using the repository class, but then stores it with the actual class (the child class) instead.

```
18:33:35 DEBUG     [doctrine] MongoDB command: {"find":"section","filter":{"_id":68,"type":{"$in":[3,4]}},"limit":1,"$db":"db","lsid":{"id":{"$binary":"R5St6qBWSAqg\/2jT16lktA==","$type":"04"}}}
18:33:35 DEBUG     [doctrine] MongoDB command: {"find":"section","filter":{"_id":68,"type":{"$in":[3,4]}},"limit":1,"$db":"db","lsid":{"id":{"$binary":"R5St6qBWSAqg\/2jT16lktA==","$type":"04"}}}
```

However, if I do this:

```
/** @var DocumentManager $manager */
$manager = Kernel::stGetObj()->getContainer()->get('doctrine_mongodb')->getManager();
$manager->getRepository(Tag::class)->find(68);
$manager->getRepository(Tag::class)->find(68);
```

It works correctly, and only one query is made.

When fetching the same document twice, only one database query should be executed if the entity is already cached in the entity map, regardless of whether the parent or child class is used.

To resolve the issue, I have overridden the find method of the DocumentRepository and do the following:
```
$parentClass = $this->getReflectionParentClass(new \ReflectionClass($documentName));        
if ($map = $this->getDocumentManager()->getClassMetadata($parentClass->getName())->discriminatorMap) {
     foreach ($map as $class) {
        if ($ret = $this->getDocumentManager()->getUnitOfWork()->tryGetById($id, $this->getDocumentManager()->getClassMetadata($class))) {
            return $ret;
          }
     }
}
```
but method tryGetById is marked internal :(

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
@@ -2035,7 +2035,7 @@ use function trigger_deprecation;
 
     /**
      * Sets the parent class names.
-     * Assumes that the class names in the passed array are in the order:
+     * Assumes that the class names passed array are in the order:
      * directParent -> directParentParent -> directParentParentParent ... -> root.
      *
      * @param list<class-string> $classNames
--- a/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
@@ -2044,7 +2044,7 @@ use function trigger_deprecation;
     {
         $this->parentClasses = $classNames;
 
-        if (count($classNames) <= 0) {
+        if ($this->isEmbeddedDocument || count($classNames) <= 0) {
             return;
         }
 
--- a/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
+++ b/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
@@ -1505,11 +1505,11 @@ final class UnitOfWork implements PropertyChangedListener
         $class = $this->dm->getClassMetadata($document::class);
         $id    = $this->getIdForIdentityMap($document);
 
-        if (isset($this->identityMap[$class->name][$id])) {
+        if (isset($this->identityMap[$class->rootDocumentName][$id])) {
             return false;
         }
 
-        $this->identityMap[$class->name][$id] = $document;
+        $this->identityMap[$class->rootDocumentName][$id] = $document;
 
         if ($document instanceof NotifyPropertyChanged && ! $this->isUninitializedObject($document)) {
             $document->addPropertyChangedListener($this);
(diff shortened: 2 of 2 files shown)


lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -2044,7 +2044,7 @@ public function setParentClasses(array $classNames): void
     {
         $this->parentClasses = $classNames;
 
-        if (count($classNames) <= 0) {
+        if ($this->isEmbeddedDocument || count($classNames) <= 0) {
```
GromNaN, line 2047: Why do you need to add this condition?

GromNaN (commented):
Hello @parada85, thanks for the exhaustive description and the proposed fix.
I reproduced the error you reported and confirmed that your fix works.

This is also the same approach that is used in the ORM: https://github.com/doctrine/orm/blob/36bef3f959ea9f7f3ea370faf2c803b9be40f5d2/src/UnitOfWork.php#L1753-L1754

Also, could you add some tests to cover this feature?

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
GromNaN: Found the response here: https://github.com/doctrine/mongodb-odm/pull/2725#issuecomment-2633555779

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -1505,11 +1505,11 @@ public function addToIdentityMap(object $document): bool
         $class = $this->dm->getClassMetadata($document::class);
         $id    = $this->getIdForIdentityMap($document);
 
-        if (isset($this->identityMap[$class->name][$id])) {
+        if (isset($this->identityMap[$class->rootDocumentName][$id])) {
```
GromNaN, line 1508: This need to be addressed and covered by tests: https://github.com/doctrine/mongodb-odm/pull/2725#discussion_r1940947597
> I don't remember details about what is stored in `rootDocumentName` but at a glance now there can be a collision for `COLLECTION_PER_CLASS` inheritance type as it's perfectly possible for two separate documents to have same id

We can update the value of `ClassMetadata::$rootDocumentName` to contain `$name` when the strategy is `COLLECTION_PER_CLASS`. That would solve the issue. (Or an new internal property `$identityMapName` to keep backward compatibility).

@malarzm any idea?

*I don't understand the purpose of `COLLECTION_PER_CLASS`, it looks the same as just extending the class as you cannot query the root repository. Having duplicate ids make it even more useless.*

malarzm: I think I was using COLLECTION_PER_CLASS years ago to work around some mapped superclass limitations, but it's been so long I can't be sure.

PR #2728, opened 2025-02-07
Always use insert for time series documents

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | 

#### Summary

This fixes an issue where assigning an ID to a time series document triggers an upsert, which is not allowed in time series collections. To work around this, we always use an `insert` command when inserting time series documents.


malarzm (approved):

GromNaN (approved):
The `_id` field can never be updated in MongoDB. Is this condition only when the document's id is set on an other field?

alcaeus:
Typically, if you set an ID manually then ODM runs an upsert to ensure that you're not overwriting data. However, there is logic in DocumentPersister to ensure that the ID of the document does not change.

Issue #2731, opened 2025-02-19
providing null for $config creating DocumentManager causes exception


aloudnoise:
### Bug Report

If provide null for $config in creating DocumentManager,  isLazyGhostObjectEnabled() are called on $config variable instead of $this->config, throwing exception

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.10.1

GromNaN:
Thanks for the bug report. How did you get this error?

Even with the fix #2732, calling `DocumentManager::create()` without argument (or with `$config = null`) fails because the hydrator requires the directory and namespace configuration.

PR #2732, opened 2025-02-19
Fix access to null `$config` variable

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #2731

#### Summary

Issue introduced in #2700, the variable `$config` can be null.

However, calling `DocumentManager::create()` without argument fails because the hydrator requires the directory and namespace configuration. So I could not add a test. https://github.com/doctrine/mongodb-odm/blob/a33bf090401ed1138a289c5e866fcc7e70c88379/lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php#L78-L86

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/DocumentManager.php
+++ b/lib/Doctrine/ODM/MongoDB/DocumentManager.php
@@ -159,7 +159,7 @@ class DocumentManager implements ObjectManager
             ],
         );
 
-        $this->classNameResolver = $config->isLazyGhostObjectEnabled()
+        $this->classNameResolver = $this->config->isLazyGhostObjectEnabled()
             ? new CachingClassNameResolver(new LazyGhostProxyClassNameResolver())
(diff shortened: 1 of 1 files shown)


malarzm (approved):

lib/Doctrine/ODM/MongoDB/DocumentManager.php
```diff
@@ -159,7 +159,7 @@ protected function __construct(?Client $client = null, ?Configuration $config =
             ],
         );
 
-        $this->classNameResolver = $config->isLazyGhostObjectEnabled()
+        $this->classNameResolver = $this->config->isLazyGhostObjectEnabled()
```
jmikola, line 162: This seems like the appropriate fix, since there were already several references to `$this->config` in the existing body.

jmikola (approved):

Issue #2733, opened 2025-02-21
Can't add a filter on a text search


fullbl:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 5.0.1

#### Summary
When adding a filter, it goes always in first stage. If the query already have a match two stages will be created, making mongodb throw an exception.
[Documentation](https://www.mongodb.com/docs/manual/reference/operator/query/text/#restrictions) states that $match stage that includes a $text must be the first stage in the pipeline.

#### Current behavior

The stage that comes from filters will always be the first in pipeline.

#### How to reproduce

1. create a filter 
```
final class ExcludeSoftDeletedFilter extends BsonFilter
{
    public function addFilterCriteria(ClassMetadata $targetDocument): array
    {
        return [
            'deletedAt' => null,
        ];
    }
}
```

2. run a query with texts:
```
$this->createAggregationBuilder()
            ->match()
            ->field('field')
            ->text('value')
```

3.
```
An exception occured, transforming to an Error resource. {"exception":"[object] (MongoDB\\Driver\\Exception\\CommandException(code: 17313): $match with $text is only allowed as the first pipeline stage at /var/www/html/vendor/mongodb/mongodb/src/Operation/Aggregate.php:383
```
#### Expected behavior

In my opinion, the CriteriaMerger should be used when merging the attributes, or there should be a "reorder" before submitting the query, so the text stage should go first

GromNaN:
It seems that we need to apply the same kind of logic that we already have with `$geoNear`: https://github.com/doctrine/mongodb-odm/blob/402efbd6786c3141cb5c431d1bb8d01375770707/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php#L267-L270

fullbl:
Something like this would work, don't know if it's optimal (I'm pretty new to mongodb):

```
$matchExpression = $this->applyFilters([]);
        if ($matchExpression !== []) {
            if ($this->getStage(0) instanceof Stage\MatchStage) {
                $merger = new CriteriaMerger;
                $pipeline[0]['$match'] = $merger->merge($pipeline[0]['$match'], $matchExpression);

                return $pipeline;
            }
            array_unshift($pipeline, ['$match' => $matchExpression]);
        }
```

GromNaN:
Closing as completed by https://github.com/doctrine/mongodb-odm/pull/2739

PR #2734, opened 2025-02-21
Can't add a filter on a text search

fullbl:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | yes
| Fixed issues | #2733 

#### Summary
test for the failing condition


GromNaN:
Thank you for the test, I opened https://github.com/doctrine/mongodb-odm/pull/2739 with a fix.

PR #2735, opened 2025-02-27
Improve comment in generated code

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | -

#### Summary

Minor improvement to distinct between ReferenceMany and EmbedMany in the generated code.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
+++ b/lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
@@ -294,7 +294,7 @@ EOF
                 $code .= sprintf(
                     <<<'EOF'
 
-        // ReferenceMany & EmbedMany
+        // %4$s
         $mongoData = $data['%1$s'] ?? null;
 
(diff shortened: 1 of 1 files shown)


lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
```diff
@@ -294,7 +294,7 @@ private function generateHydratorClass(ClassMetadata $class, string $hydratorCla
                 $code .= sprintf(
                     <<<'EOF'
 
-        // ReferenceMany & EmbedMany
+        // %4$s
```
SenseException, line 297: What is this commented type specifier supposed to mean?

GromNaN: It identifies the field type of the following generated code. It's not very important, but as I was reading I saw that it was possible to tell which of the both types the field is.

alcaeus (approved):

Issue #2730, opened 2025-02-17
Documents are not inserted when spl_object_hash() values collide with previously removed documents


ilminaev:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.10.1

#### Summary

When trying to insert new documents after removing other documents (not necessarily from the same collection), Doctrine skips them if there is an `spl_object_hash()` collision.

#### Current behavior

If the `spl_object_hash()` value of the new document is the same as of a removed one, Doctrine mixes them up and doesn't insert the new document.

#### How to reproduce

```php
#[ODM\Document]
class TestDocument
{
  #[ODM\Id]
  public ?string $id = null;

  #[ODM\Field(type: 'string')]
  public string $name;

  public function __construct(string $name)
  {
    $this->name = $name;
  }
}
```

At first, we populate the collection with documents that will be removed later:

```php
foreach (range(1, 3) as $i) {
  $document = new TestDocument("document_{$i}");
  $dm->persist($document);
}
$dm->flush();
```

Then we remove all the existing documents and create new ones. Object hash collisions should happen here; if they don't, increasing count of the documents might help.

```php
$documents = $dm->getRepository(TestDocument::class)->findAll();
$oldObjectHashes = array_map(spl_object_hash(...), $documents);

// Removing all old documents
foreach ($documents as $document) {
  $dm->remove($document);
}
$dm->flush();
unset($documents);

// Inserting new documents
foreach (range(1, 3) as $i) {
  $document = new TestDocument("document_{$i}");
  $dm->persist($document);
  if (in_array(spl_object_hash($document), $oldObjectHashes, true)) {
    echo "Object hash collision found.\n";
  }
}
$dm->flush();
```

After executing the script the collection is empty, i.e. no new documents have been inserted.

#### Expected behavior

All new documents are present in the database.

GromNaN:
Hello [ilminaev](https://github.com/ilminaev), thank you for this complete bug report.

This usage of `spl_object_hash()` was added in https://github.com/doctrine/mongodb-odm/pull/2586 for transactions.

Would you like to open a PR to replace the `spl_object_hash()` with a [`WeakMap`](https://php.watch/articles/practical-weakmap)? This PR must target the branch `2.10.x` and add a test (your reproducer is perfect as test).

GromNaN:
[...]

PR #2737, opened 2025-02-28
Fix duplicate object hash for removed object

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | yes
| Fixed issues | #2730

#### Summary

In think a document that has been "removed", need to be treated as a new document when persisted again. The only thing is that it's ID can be already initialized.

(I tried the WeakMap approach, but that's not going well. https://github.com/GromNaN/mongodb-odm/commit/8204cb0607c8ca1c10044efa495c6c9a2e35757a)

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
+++ b/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
@@ -1790,7 +1790,7 @@ final class UnitOfWork implements PropertyChangedListener
                 // Document becomes managed again
                 unset($this->scheduledDocumentDeletions[$oid]);
 
-                $this->documentStates[$oid] = self::STATE_MANAGED;
+                $this->persistNew($class, $document);
                 break;
 
             case self::STATE_DETACHED:
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH2730Test.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH2730Test.php
@@ -13,10 +13,11 @@ class GH2730Test extends BaseTestCase
 {
     public function testUniqueObjectIdentifier(): void
     {
-        $document = new GH2730Document();
+        $document = new GH2730();
         $oid      = spl_object_hash($document);
         $this->dm->persist($document);
         $this->dm->flush();
+        $id = $document->id;
 
         // Remove the document
         $this->dm->remove($document);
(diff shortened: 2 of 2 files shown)


tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH2730Test.php
```diff
@@ -0,0 +1,83 @@
...
+    {
+        $document = new GH2730Document();
+        $this->dm->persist($document);
+        $this->dm->flush();
+
+        // Remove the document
+        $this->dm->remove($document);
+
+        // Re-persist the same document
+        $this->dm->persist($document);
```
GromNaN, line 77: This test was already passing before the fix.

With this change, the `prePersist` event will be re-triggered when a document is removed then persisted. Before the "remove" was just cancelled.
If the ID value is already generated, it will not be regenerated.

This test may duplicate `UnitOfWorkTest::testPersistRemovedDocument`, do you want I remove it? `testPersistRemovedEmbeddedDocument` also covers it.
https://github.com/doctrine/mongodb-odm/blob/fd0ad45d618d48ecb055baa9d287224e75e2f3b3/tests/Doctrine/ODM/MongoDB/Tests/UnitOfWorkTest.php#L366-L387

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -1790,7 +1790,7 @@ private function doPersist(object $document, array &$visited): void
                 // Document becomes managed again
                 unset($this->scheduledDocumentDeletions[$oid]);
 
-                $this->documentStates[$oid] = self::STATE_MANAGED;
+                $this->persistNew($class, $document);
```
GromNaN, line 1793: The issue was with the `documentIdentifiers` that was not updated with the new ID, and the new document was not added to `scheduleForInsert`.

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH2730Test.php
jmikola: I think it's fine to leave those tests in place, since they're predating this regression test.

jmikola (approved):
Thanks for the comprehensive tests.

https://github.com/doctrine/mongodb-odm/commit/8204cb0607c8ca1c10044efa495c6c9a2e35757a is unfortunate, given that WeakMap is the proper replacement for SPL object hash and IDs. If that work isn't currently planned out, perhaps it'd warrant a PHPORM ticket to allocate proper time to investigate.

PR #2739, opened 2025-02-28
Add support for using `$text` operator in a aggregation with a filter

### LINKED ISSUE #2733: Can't add a filter on a text search

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | Fix #2733

#### Summary

The `$text` operator [must be in the first](https://www.mongodb.com/docs/manual/reference/operator/query/text/#behavior) `$match` stage. Instead of prepending the pipeline with an other `$match` stage for the filter, the builder merge them using a logical `$and`. 

We cannot detect when `$text` is used as this can be deep in the `$match` query, so the behavior is for all the pipelines.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
@@ -264,6 +264,11 @@ class Builder
      * @param bool $applyFilters Whether to apply filters on the aggregation
      * pipeline stage
      *
+     * For pipelines where the first stage is a $match stage, it will merge
+     * the document filters with the existing stage in a logical $and. This is
+     * required as $text operator can be used anywhere in the first $match stage
+     * or in the document filters.
+     *
      * For pipelines where the first stage is a $geoNear stage, it will apply
      * the document filters and discriminator queries to the query portion of
      * the geoNear operation. For all other pipelines, it prepends a $match stage
@@ -314,7 +319,17 @@ class Builder
 
         $matchExpression = $this->applyFilters([]);
         if ($matchExpression !== []) {
-            array_unshift($pipeline, ['$match' => $matchExpression]);
+            if (isset($pipeline[0]['$match'])) {
+                // Merge filters into the first $match stage with a logical $and
+                $pipeline[0]['$match'] = [
+                    '$and' => [
+                        $matchExpression,
+                        $pipeline[0]['$match'],
+                    ],
+                ];
+            } else {
+                array_unshift($pipeline, ['$match' => $matchExpression]);
+            }
         }
 
         return $pipeline;
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
@@ -14,6 +14,7 @@ use Doctrine\ODM\MongoDB\Query\Expr as QueryExpr;
 use GeoJson\Geometry\Point;
 use MongoDB\Collection;
 use OutOfRangeException;
+use stdClass;
 use TypeError;
 
 use function array_map;
@@ -311,25 +312,21 @@ class Builder
             return $pipeline;
         }
 
-        if ($this->getStage(0) instanceof Stage\GeoNear) {
+        if (isset($pipeline[0]['$geoNear'])) {
             $pipeline[0]['$geoNear']['query'] = $this->applyFilters($pipeline[0]['$geoNear']['query']);
 
             return $pipeline;
         }
 
+        if (isset($pipeline[0]['$match'])) {
+            $pipeline[0]['$match'] = $this->applyFilters($pipeline[0]['$match']);
+
+            return $pipeline;
+        }
+
         $matchExpression = $this->applyFilters([]);
-        if ($matchExpression !== []) {
-            if (isset($pipeline[0]['$match'])) {
-                // Merge filters into the first $match stage with a logical $and
-                $pipeline[0]['$match'] = [
-                    '$and' => [
-                        $matchExpression,
-                        $pipeline[0]['$match'],
-                    ],
-                ];
-            } else {
-                array_unshift($pipeline, ['$match' => $matchExpression]);
-            }
+        if ((array) $matchExpression !== []) {
+            array_unshift($pipeline, ['$match' => $matchExpression]);
         }
 
         return $pipeline;
@@ -711,18 +708,22 @@ class Builder
     /**
      * Applies filters and discriminator queries to the pipeline
      *
-     * @param array<string, mixed> $query
+     * @param array<string, mixed>|stdClass $query
      *
-     * @return array<string, mixed>
+     * @return array<string, mixed>|stdClass
      */
-    private function applyFilters(array $query): array
+    private function applyFilters(array|stdClass $query): array|stdClass
     {
+        if (! is_array($query)) {
+            $query = (array) $query;
+        }
+
         $documentPersister = $this->getDocumentPersister();
 
         $query = $documentPersister->addDiscriminatorToPreparedQuery($query);
         $query = $documentPersister->addFilterToPreparedQuery($query);
 
-        return $query;
+        return $query === [] ? (object) $query : $query;
     }
 
     private function getDocumentPersister(): DocumentPersister
--- a/tests/Doctrine/ODM/MongoDB/Tests/Aggregation/BuilderTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Aggregation/BuilderTest.php
@@ -368,6 +368,39 @@ class BuilderTest extends BaseTestCase
         self::assertEquals($expectedPipeline, $builder->getPipeline());
     }
 
+    public function testBuilderMergeFilterAndDiscriminatorWithMatchStage(): void
+    {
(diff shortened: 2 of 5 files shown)


lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
```diff
@@ -314,7 +319,16 @@ public function getPipeline(/* bool $applyFilters = true */): array
...
         if ($matchExpression !== []) {
-            array_unshift($pipeline, ['$match' => $matchExpression]);
+            if (! empty($pipeline[0]['$match'])) {
+                $pipeline[0]['$match'] = [
+                    '$and' => [
+                        $matchExpression,
+                        $pipeline[0]['$match'],
+                    ],
+                ];
+            } else {
```
jmikola: If `$pipeline[0]['$match']` is empty, is there any reason to preserve it within the pipeline? I would assume it could just be replaced with a non-empty `$matchExpression`.

GromNaN: Only if the first stage is a `$match` stage.
In fact, if the first stage in an empty query, it contains `['$match' => (object) []]` which is not empty.
https://github.com/doctrine/mongodb-odm/blob/fd0ad45d618d48ecb055baa9d287224e75e2f3b3/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/MatchStage.php#L285

I update to use `isset`. I was using `empty` with the assumption it was an empty array.

I'll open a PR to skip empty `MatchStage`.

GromNaN: https://github.com/doctrine/mongodb-odm/pull/2740

lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
```diff
@@ -314,7 +319,17 @@ public function getPipeline(/* bool $applyFilters = true */): array
 
         $matchExpression = $this->applyFilters([]);
         if ($matchExpression !== []) {
-            array_unshift($pipeline, ['$match' => $matchExpression]);
+            if (isset($pipeline[0]['$match'])) {
+                // Merge filters into the first $match stage with a logical $and
+                $pipeline[0]['$match'] = [
+                    '$and' => [
```
jmikola: Starting a separate comment thread for this. Apologies for not asking this sooner, but do you know why the `$geoNear` stage is able to use:

```php
$this->applyFilters($pipeline[0]['$geoNear']['query']);
```

Since that merges the filter and discriminator criteria into the query, I'm curious as to why we could do the following instead of `$and`:

```php
$pipeline[0]['$match'] = $this->applyFilters($pipeline[0]['$match']);
```

GromNaN: Excellent point, I probably didn't go far enough in my exploration. In fact, `CriteriaMerger` already does the job very well.

https://github.com/doctrine/mongodb-odm/blob/fd0ad45d618d48ecb055baa9d287224e75e2f3b3/lib/Doctrine/ODM/MongoDB/Query/CriteriaMerger.php#L27-L41

I modifies the implementation to match with `$geoNear`. Good news, I didn't have to touch the tests and it cleanly drops empty `$match` criteria.

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/GeoNearTest.php
```diff
@@ -72,7 +72,7 @@ public function testLimitDoesNotCreateExtraStage(): void
             ->geoNear(0, 0)
             ->limit(1);
 
-        $stage = ['near' => [0, 0], 'spherical' => false, 'distanceField' => null, 'query' => [], 'num' => 1];
-        self::assertSame([['$geoNear' => $stage]], $builder->getPipeline());
+        $stage = ['near' => [0, 0], 'spherical' => false, 'distanceField' => null, 'query' => (object) [], 'num' => 1];
+        self::assertEquals([['$geoNear' => $stage]], $builder->getPipeline());
```
GromNaN, line 76: This was an error: `$geoNear.query` does not accept an empty array
> MongoServerError[TypeMismatch]: query must be an object

lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
```diff
@@ -696,18 +708,22 @@ public function addStage(Stage $stage): Stage
...
+            $query = (array) $query;
+        }
+
         $documentPersister = $this->getDocumentPersister();
 
         $query = $documentPersister->addDiscriminatorToPreparedQuery($query);
         $query = $documentPersister->addFilterToPreparedQuery($query);
 
-        return $query;
+        return $query === [] ? (object) $query : $query;
```
GromNaN, line 726: I have to juggle between `array` and empty `stdClass` to ensure that we have a map when converting to BSON.

lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
```diff
@@ -306,14 +312,20 @@ public function getPipeline(/* bool $applyFilters = true */): array
...
 
+        if (isset($pipeline[0]['$match'])) {
+            $pipeline[0]['$match'] = $this->applyFilters($pipeline[0]['$match']);
+
+            return $pipeline;
+        }
+
         $matchExpression = $this->applyFilters([]);
-        if ($matchExpression !== []) {
+        if ((array) $matchExpression !== []) {
```
GromNaN, line 328: Or
```suggestion
        if ($matchExpression == new stdClass) {
```

lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
```diff
@@ -306,14 +312,20 @@ public function getPipeline(/* bool $applyFilters = true */): array
             return $pipeline;
         }
 
-        if ($this->getStage(0) instanceof Stage\GeoNear) {
+        if (isset($pipeline[0]['$geoNear'])) {
```
GromNaN, line 315: This is more efficient and more resilient if we introduce changes on the `$pipeline` array like #2740.

GromNaN (commented):
It's getting bigger than expected, but that's no bad thing.

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/GeoNearTest.php
jmikola: Does this suggest that the `geoNear()` builder method never worked in previous versions when query criteria was omitted? That seems like something worthy of its own ticket for changelog visibility, even if you want to keep the fix in this PR.

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/BuilderTest.php
```diff
@@ -368,6 +368,36 @@ public function testBuilderAppliesFilterAndDiscriminatorWithMatchStage(): void
...
+
+        $expectedPipeline = [
+            [
+                '$match' => [
+                    '$and' => [
+                        [
+                            'stype' => 'server_guest',
+                            '$text' => ['$search' => 'Paul'],
+                        ],
+                        ['filtered' => true],
```
jmikola, line 392: The query here seems odd, but I understand that's because `addDiscriminatorToPreparedQuery()` sets the key directly and `addFilterToPreparedQuery()` uses CriteriaMerger, which will always adds `$and` for 2+ criteria.

jmikola (approved):
You can decide if it's worth reporting a separate issue for the outstanding `geoNear()` bug.

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/GeoNearTest.php
GromNaN: Explicitly fixed in https://github.com/doctrine/mongodb-odm/pull/2743.

PR #2740, opened 2025-03-03
Skip empty `$match` from pipeline builder

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no-ish
| Fixed issues | -

#### Summary

Don't send an unnecessary stage to the server.

Might be considered as a breaking change is someone manipulate the generated pipeline.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
@@ -17,7 +17,6 @@ use OutOfRangeException;
 use stdClass;
 use TypeError;
 
-use function array_map;
 use function array_unshift;
 use function func_get_arg;
 use function func_num_args;
@@ -295,10 +294,15 @@ class Builder
             ));
         }
 
-        $pipeline = array_map(
-            static fn (Stage $stage) => $stage->getExpression(),
-            $this->stages,
-        );
+        $pipeline = [];
+        foreach ($this->stages as $stage) {
+            $stage = $stage->getExpression();
+            if (! $stage) {
+                continue;
+            }
+
+            $pipeline[] = $stage;
+        }
 
         if ($this->getStage(0) instanceof Stage\IndexStats) {
             // Don't apply any filters when using an IndexStats stage: since it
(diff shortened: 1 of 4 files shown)


lib/Doctrine/ODM/MongoDB/Aggregation/Stage.php
```diff
@@ -33,7 +33,7 @@ public function __construct(Builder $builder)
      * @return array<string, mixed>
      * @phpstan-return StageExpression
      */
-    abstract public function getExpression(): array;
+    abstract public function getExpression(): ?array;
```
GromNaN, line 36: Classes extending this can still restrict the return type.

GromNaN (commented):
We can do the same for `$addFields` and `$set`.

But also for `$unsetFields` and `$project`. This stages doesn't accept empty lists, so this can be an improvement.

lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
```diff
@@ -289,10 +288,15 @@ public function getPipeline(/* bool $applyFilters = true */): array
...
         }
 
-        $pipeline = array_map(
-            static fn (Stage $stage) => $stage->getExpression(),
-            $this->stages,
-        );
+        $pipeline = [];
+        foreach ($this->stages as $stage) {
+            $stage = $stage->getExpression();
+            if (! $stage) {
```
jmikola: Would this be better off as an explicit `null` check?

GromNaN:
```suggestion
            if ($stage === null) {
```

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/BuilderTest.php
```diff
@@ -420,6 +420,20 @@ public function testBuilderWithIndexStatsStageDoesNotApplyFilters(): void
         self::assertSame('$indexStats', array_keys($builder->getPipeline()[0])[0]);
     }
 
+    public function testEmptyMatchStage(): void
```
jmikola:
```suggestion
    public function testEmptyMatchStageIsSkipped(): void
```

jmikola (approved):
Optional test rename.

PR #2741, opened 2025-03-03
Replace "attribute" with "argument" to avoid confusion with PHP attributes

### LINKED ISSUE #2691: Fix wording in attribute reference
The attribute reference refers to parameters of attributes as "attributes", most likely as a leftover from when it was the annotation reference and wasn't as confusing. The following alternatives come to mind:
* `options`: this can be confusing and makes for awkward wording, as in "the `xyz` option is optional" and implies that everything is optional
* `arguments`: this is most likely the better choice, using the same wording used in general for parameter values in method calls
* `parameters`: used interchangeably with `arguments`, but technically incorrect as it refers to the variables listed in a declaration

I would suggest replacing all mentions of `option`, `attribute`, etc. when talking about arguments with `argument`.

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | doc
| BC Break     | no
| Fixed issues | Fix #2691

#### Summary

~I prefer the word "parameter" as it's more a configuration than a method calls. For me "argument" refers to the definition in the method signature, and "parameter" to the value in the method call.~

I kept the word "attribute" in the "xml-mappings" page as this is the correct word for XML tags.

Also improving examples and wording for "type", as the mapping from PHP property type to MongoDB Field type should be used by default for simplicity.
And a new paragraph about the date fields specifically, I think that was missing for newcomers that may make the mistake of storing dates as string.

### DIFF
--- a/docs/en/reference/basic-mapping.rst
+++ b/docs/en/reference/basic-mapping.rst
@@ -34,8 +34,8 @@ annotations by offering a standardized approach to metadata, eliminating
 the need for the separate parsing library required by annotations.
 
 In this documentation we follow the `PER Coding Style <https://www.php-fig.org/per/coding-style/#12-attributes>`_
-for attributes. We use named arguments for attributes as they argument names
-or attribute classes constructors are covered by Doctrine Backward-Compatibility promise.
+for attributes. We use named arguments for attributes as the names of their
+constructor arguments are covered by Doctrine Backward-Compatibility promise.
 
 .. note::
 
@@ -109,8 +109,8 @@ option as follows:
 Now instances of ``Documents\User`` will be persisted into a
 collection named ``users`` in the database ``my_db``.
 
-If you want to omit the db attribute you can configure the default db
-to use with the ``setDefaultDB`` method:
+If you want to omit the ``db`` argument you can configure the default database
+to use with the ``setDefaultDB()`` method:
 
 .. code-block:: php
 
@@ -195,7 +195,7 @@ _________________
     Doctrine will skip type autoconfiguration if settings are provided explicitly.
 
 Since version 2.4 Doctrine can determine usable defaults from property types
-on document classes. Doctrine will map PHP types to ``type`` attribute as
+on document classes. Doctrine will map PHP types to ``type`` arguments as
 follows:
 
 - ``DateTime``: ``date``
@@ -392,12 +392,9 @@ Then specify the ``class`` option for the ``CUSTOM`` strategy:
 Fields
 ~~~~~~
 
-To mark a property for document persistence the ``#[Field]`` docblock
-attribute can be used. This attribute usually requires at least 1
-attribute to be set, the ``type``. The ``type`` attribute specifies
-the Doctrine Mapping Type to use for the field. If the type is not
-specified, 'string' is used as the default mapping type since it is
-the most flexible.
+To mark a property to be persisted in the document, add the ``#[Field]``
+attribute. The name and the type of the field are inferred from the property
+name and type.
 
 Example:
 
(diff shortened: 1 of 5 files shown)


alcaeus (commented):
Agree with the change from "attribute" to "argument" to reduce confusion. I'll let @jmikola take a look at the wording of the "Date Fields" section, but so far: LGTM.

docs/en/reference/basic-mapping.rst
```diff
@@ -195,7 +195,7 @@ _________________
     Doctrine will skip type autoconfiguration if settings are provided explicitly.
 
 Since version 2.4 Doctrine can determine usable defaults from property types
-on document classes. Doctrine will map PHP types to ``type`` attribute as
+on document classes. Doctrine will map PHP types to ``type`` argument as
```
jmikola:
```suggestion
on document classes. Doctrine will map PHP types to ``type`` arguments as
```

Subject/verb agreement.

docs/en/reference/basic-mapping.rst
```diff
@@ -458,6 +453,46 @@ as follows:
 
         <field field-name="name" name="db_name" />
 
+Date Fields
+~~~~~~~~~~~
+
+MongoDB has a specific database type to store Date and Time. You should never
```
jmikola:
```suggestion
MongoDB has a specific database type to store date and time values. You should never
```

Not sure why these were capitalized.

docs/en/reference/basic-mapping.rst
```diff
@@ -458,6 +453,46 @@ as follows:
...
+use a string to store a date. To define a date field, use the property types
+``DateTime`` or ``DateTimeImmutable`` so that Doctrine maps it to a BSON date.
+The ``date`` and ``date_immutable`` mapping types can be used explicitly,
+they are required only if the property type is ``DateTimeInterface`` or not set.
```
jmikola, line 463:
```suggestion
but they are required only if the property type is ``DateTimeInterface`` or not set.
```

GromNaN: I'm trying to avoid the use of “but”, which is too oppositional, whereas it's a complement of information.

docs/en/reference/basic-mapping.rst
```diff
@@ -458,6 +453,46 @@ as follows:
...
+
+            #[Field(type: 'date_immutable')]
+            public \DateTimeInterface $datetime;
+        }
+
+    .. code-block:: xml
+
+        <field field-name="immutableDateTime" type="date_immutable" />
+        <field field-name="mutableDateTime" name="date" />
+        <field field-name="$datetime" name="date_immutable" />
```
GromNaN:
```suggestion
        <field field-name="datetime" name="date_immutable" />
```

docs/en/reference/basic-mapping.rst
jmikola: In that case, replacing the comma after "explicitly" with a semicolon would read better.

docs/en/reference/basic-mapping.rst
```diff
@@ -458,6 +453,46 @@ as follows:
 
         <field field-name="name" name="db_name" />
 
+Date Fields
+~~~~~~~~~~~
+
+MongoDB has a specific database type to store date and time values. You should never
+use a string to store a date. To define a date field, use the property types
+``DateTime`` or ``DateTimeImmutable`` so that Doctrine maps it to a BSON date.
+The ``date`` and ``date_immutable`` mapping types can be used explicitly,
```
GromNaN:
```suggestion
The ``date`` and ``date_immutable`` mapping types can be used explicitly;
```

jmikola (approved):

PR #2743, opened 2025-03-05
Fix `$geoNear` stage with empty query

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | https://github.com/doctrine/mongodb-odm/pull/2739#discussion_r1978342677

#### Summary

The `$geoNear` stage property `query` requires a document. When nothing is set, an empty array is currently sent; which is invalid.

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/GeoTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/GeoTest.php
@@ -1,0 +1,68 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ODM\MongoDB\Tests\Functional;
+
+use Doctrine\ODM\MongoDB\Tests\BaseTestCase;
+use Documents\Place;
+
+class GeoTest extends BaseTestCase
+{
+    public function setUp(): void
+    {
+        parent::setUp();
+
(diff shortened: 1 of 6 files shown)


tests/Doctrine/ODM/MongoDB/Tests/Functional/GeoTest.php
```diff
@@ -0,0 +1,68 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ODM\MongoDB\Tests\Functional;
+
+use Doctrine\ODM\MongoDB\Tests\BaseTestCase;
+use Documents\Place;
+
+class GeoTest extends BaseTestCase
```
jmikola, line 10: Should this be titled GeoNearTest, or would you intend for other geospatial operators to get added to this down the line?

jmikola (approved):
Decide if the test case should be renamed. Everything else LGTM.

tests/Doctrine/ODM/MongoDB/Tests/Functional/GeoTest.php
GromNaN: My idea is that we can add other tests for other operators using the fixtures already defined.

PR #2750, opened 2025-04-06
Fix: Cast numeric write-concern values from XML mapping to integers

### LINKED ISSUE #2749: Bug: write-concern="1" in XML mapping causes MongoDB error due to string cast
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.10.1

#### Summary

Using write-concern="1" in XML mapping causes Doctrine ODM to pass the value as a string ("1") instead of an integer (1), which MongoDB interprets as a named write concern mode — leading to an error.

#### Current behavior
When using XML mapping like:

`<document name="App\Document\LogEntry" collection="log_entries" write-concern="1">`

Doctrine ODM internally casts the value to a string ("1"), which results in MongoDB throwing the error:

`No write concern mode named '1' found in replica set configuration`

This is due to the XmlDriver treating all attributes as strings and not checking if the value is numeric.

#### How to reproduce
Define a document mapping in XML with write-concern="1":

`<document name="App\Document\LogEntry" collection="log_entries" write-concern="1">`
Persist and flush an object of this class.

Observe the error from MongoDB:

`No write concern mode named '1' found in replica set configuration`
Relevant code snippet from XmlDriver.php:

```
if (isset($xmlRoot['write-concern'])) {
    $metadata->setWriteConcern((string) $xmlRoot['write-concern']); // always string
}
```

#### Expected behavior
Doctrine ODM should check whether the write-concern value is numeric, and cast it to an integer if appropriate. This ensures that w: 1 is passed to MongoDB as an integer, not a named string.

#### Proposed fix:

```
if (isset($xmlRoot['write-concern'])) {
    $writeConcern = (string) $xmlRoot['write-concern'];
    if (is_numeric($writeConcern)) {
        $writeConcern = (int) $writeConcern;
    }
    $metadata->setWriteConcern($writeConcern);
}
```
This preserves support for named modes like "majority" while correctly handling numeric write concerns.

ikallali:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | Fix #2749

#### Summary

When using write-concern="1" in XML mapping, the value was being cast to string, causing MongoDB to interpret it as a named write concern mode instead of a numeric level. This led to errors like: "No write concern mode named '1' found in replica set configuration".

This commit ensures that numeric write concern values are properly cast to integers, while still allowing named modes such as "majority" to be used as strings.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
@@ -162,7 +162,7 @@ class XmlDriver extends FileDriver
             if (is_numeric($writeConcern)) {
                 $writeConcern = (int) $writeConcern;
             }
-            
+
             $metadata->setWriteConcern($writeConcern);
         }
 
--- a/lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
@@ -159,9 +159,10 @@ class XmlDriver extends FileDriver
 
         if (isset($xmlRoot['write-concern'])) {
             $writeConcern = (string) $xmlRoot['write-concern'];
-            if(is_numeric($writeConcern)) {
+            if (is_numeric($writeConcern)) {
                 $writeConcern = (int) $writeConcern;
             }
+            
             $metadata->setWriteConcern($writeConcern);
         }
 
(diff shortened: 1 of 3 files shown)


GromNaN:
In fact, the `int` value is accepted in the [attribute configuration](https://github.com/doctrine/mongodb-odm/blob/11198c2d29be2ec9ae45313fe32e7665f50bbf6e/lib/Doctrine/ODM/MongoDB/Mapping/Annotations/Document.php#L37). Since using  an `int` for this setting is [documented in MongoDB](https://www.mongodb.com/docs/manual/reference/write-concern/#w-option), let's make it possible to use it in XML.

Could you add a test to cover this feature?

GromNaN (approved):
Thank you @ikallali.
I fixed the tests to strictly check the write concern type; that confirmed the need for your patch.

Issue #2753, opened 2025-04-09
Document HydratorFactory does not support nullable date field


frenchcomp:
### Bug Report

The HydratorFactory does not properly support nullable date fields. So typed properties are not initialized and there an issue with PHP because typed property  must not be accessed before initialization.

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version   | 2.11

#### Summary

In a my Document Class, I have a nullable date field : 

```
protected ?DateTimeInterface $moderatedAt = null;
```

Its mapping (in XML) is : 
```
<field field-name="moderatedAt" name="moderated_at" type="date" nullable="true"/>
```

The generated hydrator code for this field is :

```
       // Field(type: "date")
        if (isset($data['moderated_at'])) {
            $value = $data['moderated_at'];
            if ($value === null) { $return = null; } else { $return = \Doctrine\ODM\MongoDB\Types\DateType::getDateTime($value); }
            $this->class->reflFields['moderatedAt']->setValue($document, clone $return);
            $hydratedData['moderatedAt'] = $return;
        }
```

#### Current behavior

But, $data['moderated_at'] is present and at null. So  if (isset($data['moderated_at'])) 
will be never true. And the next nested condition will be never executed

#### Expected behavior

The generated code must be like this : 

```
if (isset($data['moderated_at']) || (! empty($this->class->fieldMappings['moderatedAt']['nullable']) && array_key_exists('moderated_at', $data))) {
```

Thanks :)

frenchcomp:
The fixed sniped code is

```
// Field(type: "date")
        if (isset($data['%1$s']) || (! empty($this->class->fieldMappings['%2$s']['nullable']) && array_key_exists('%1$s', $data))) {
            $value = $data['%1$s'];
            %3$s
            if (\is_object($return)) {
              $this->class->reflFields['%2$s']->setValue($document, clone $return);
            } else {
              $this->class->reflFields['%2$s']->setValue($document, $return);
            }
            $hydratedData['%2$s'] = $return;
        }
```

It's just a proposal, I'm not satisfied by the inner if condition. There is another error with ` $this->class->reflFields['moderatedAt']->setValue($document, clone $return);` because $return can be null.

If I have the time these next day, I will try to push a PR, but if anyone wants to do it, it will have my sincere thanks.

PR #2755, opened 2025-04-11
Support nullable date field in HydratorFactory #2753

frenchcomp:
Update code generated by the HydratorFactory for date to support nullable datetime field.

|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #2753 

#### Summary

These PR update only the HydratorFactory class to update the if condition to support nullable field like for other type.
It will update also the clone $return value, because $return can be 'null' and we can not clone null value in PHP.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
+++ b/lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
@@ -188,10 +188,8 @@ EOF
         if (array_key_exists('%1$s', $data) && ($data['%1$s'] !== null || ($this->class->fieldMappings['%2$s']['nullable'] ?? false))) {
             $value = $data['%1$s'];
             %3$s
-            if (\is_object($return)) {
-              $return = clone $return;
-            }
-            $this->class->reflFields['%2$s']->setValue($document, $return);
+            $field = $this->class->reflFields['%2$s'];
+            if (null !== $return) { $field->setValue($document, clone $return); } else { $field->setValue($document, null); }
             $hydratedData['%2$s'] = $return;
         }
 
--- a/lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
+++ b/lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
@@ -188,8 +188,7 @@ EOF
         if (array_key_exists('%1$s', $data) && ($data['%1$s'] !== null || ($this->class->fieldMappings['%2$s']['nullable'] ?? false))) {
             $value = $data['%1$s'];
             %3$s
-            $field = $this->class->reflFields['%2$s'];
-            if (null !== $return) { $field->setValue($document, clone $return); } else { $field->setValue($document, null); }
+            $this->class->reflFields['%2$s']->setValue($document, $return === null ? null : clone $return);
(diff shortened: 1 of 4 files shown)


lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
```diff
@@ -185,10 +185,13 @@ private function generateHydratorClass(ClassMetadata $class, string $hydratorCla
                     <<<'EOF'
 
         // Field(type: "date")
-        if (isset($data['%1$s'])) {
+        if (isset($data['%1$s']) || (! empty($this->class->fieldMappings['%2$s']['nullable']) && array_key_exists('%1$s', $data))) {
```
GromNaN: We avoid usage of `empty` as it covers too many cases. Maybe we can refactor the condition like this;

```suggestion
        if (array_key_exists('%1$s', $data) && ($data['%1$s'] !== null || $this->class->fieldMappings['%2$s']['nullable'] ?? false)) {
```

GromNaN (changes requested):
Hello @frenchcomp, thanks for contributing this fix.
Apparently it's breaking an existing test case. Also, the feature needs to be covered by a new non-regression test. 

```
1) Doctrine\ODM\MongoDB\Tests\Functional\DateTest::testDateInstanceValueChangeDoesCauseUpdateIfValueIsTheSame
Failed asserting that an array is not empty.

/home/runner/work/mongodb-odm/mongodb-odm/tests/Doctrine/ODM/MongoDB/Tests/Functional/DateTest.php:102
```

lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
frenchcomp: Yes, I understand, I had just copied another part of the factory. In an external project I try to maintain consistency.
I forgeted the flag WIP. I just wanted to run the tests, not being able to do it on my machine

GromNaN: Indeed, I wasn't aware that `empty` is used a lot in this class.

tests/Doctrine/ODM/MongoDB/Tests/Functional/DateTest.php
```diff
@@ -84,6 +84,20 @@ public function testDateInstanceChangeWhenValueDifferenceIsSubSecond(): void
         self::assertNotEmpty($changeset);
     }
 
+    // Test Issue #2753
```
GromNaN: I think this comment is not necessary. Git Blame does the job if necessary.
```suggestion
```

tests/Documents/User.php
```diff
@@ -34,6 +34,10 @@ class User extends BaseDocument
     #[ODM\Field(type: 'date')]
     protected $createdAt;
 
+    /** @var ?DateTimeInterface */
```
GromNaN: Duplicate information.
```suggestion
```

GromNaN (commented):
LGTM after the 2 CS fixes.

GromNaN (approved):

GromNaN:
Merci Richard!

frenchcomp:
> Merci Richard!

De rien, désolé pour les A/R

PR #2761, opened 2025-04-28
Upgrade to PHPStan 2.1

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | -

#### Summary

The upgrade is necessary because phpstan 1.10 that is currently has issues with PHP 8.4.

### DIFF
--- a/docs/en/cookbook/lookup-reference.rst
+++ b/docs/en/cookbook/lookup-reference.rst
@@ -68,7 +68,7 @@ and a date.
         #[Field(type: 'date_immutable')]
         public DateTimeImmutable $date;
 
-        /** @var Collection<Item> */
+        /** @var Collection<int, Item> */
         #[ReferenceMany(
             targetDocument: Item::class,
             cascade: 'all',
@@ -378,7 +378,7 @@ You need to create a new class to hold the result of the aggregation.
         #[Field(type: 'date_immutable')]
         public DateTimeImmutable $date;
 
-        /** @var Collection<Item> */
+        /** @var Collection<int, Item> */
         #[EmbedMany(targetDocument: Item::class)]
         public Collection $items;
 
@@ -631,7 +631,7 @@ but not the user.
         #[Field(type: 'string')]
         public string $name;
 
-        /** @var Collection<UserOrderResult> */
+        /** @var Collection<int, UserOrderResult> */
         #[EmbedMany(targetDocument: UserOrderResult::class)]
         public Collection $orders;
     }
@@ -645,7 +645,7 @@ but not the user.
         #[Field(type: 'date_immutable')]
         public DateTimeImmutable $date;
 
-        /** @var Collection<Item> */
+        /** @var Collection<int, Item> */
         #[EmbedMany(targetDocument: Item::class)]
         public Collection $items;
     }
--- a/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
+++ b/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
@@ -1952,7 +1952,7 @@ final class UnitOfWork implements PropertyChangedListener
                 $prop = $this->reflectionService->getAccessibleProperty($class->name, $name);
                 assert($prop instanceof ReflectionProperty);
 
-                if (method_exists($prop, 'isInitialized') && ! $prop->isInitialized($document)) {
+                if (! $prop->isInitialized($document)) {
                     continue;
                 }
 
--- a/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
+++ b/lib/Doctrine/ODM/MongoDB/UnitOfWork.php
@@ -1749,7 +1749,7 @@ final class UnitOfWork implements PropertyChangedListener
      * NOTE: This method always considers documents that are not yet known to
      * this UnitOfWork as NEW.
      *
-     * @param array<string, object> $visited
+     * @param array<int, object> $visited
      *
      * @throws InvalidArgumentException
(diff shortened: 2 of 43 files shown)


docs/en/cookbook/lookup-reference.rst
```diff
@@ -68,7 +68,7 @@ and a date.
         #[Field(type: 'date_immutable')]
         public DateTimeImmutable $date;
 
-        /** @var Collection<Item> */
+        /** @var Collection<int, Item> */
```
GromNaN, line 71: Updating all doc examples, [the `Collection` class has 2 type templates](https://github.com/doctrine/collections/blob/2.3.0/src/Collection.php#L27-L28).

tests/Doctrine/ODM/MongoDB/Tests/Functional/LifecycleTest.php
```diff
@@ -60,7 +61,7 @@ class ParentObject
     #[ODM\Id]
     private $id;
 
-    /** @var Collection<int, ChildObject>|array<ChildObject> */
+    /** @var Collection<int, ChildObject> */
```
GromNaN, line 64: `Collection` was never instantiated; but I think that's the type used when the relation is hydrated from the database. So I changed line 100 to create an `ArrayCollection`.

tests/Doctrine/ODM/MongoDB/Tests/Mapping/Driver/AbstractDriverTestCase.php
```diff
@@ -31,7 +30,7 @@ public function setUp(): void
 
     public function tearDown(): void
     {
-        unset($this->driver);
```
GromNaN, line 34: PHPstan warns that the property may have hooks in the child classes, so `unset` is not allowed.
Marking all test classes `final` is an other option, but doesn't work when we need to extend it (`AbstractDriverTestCase`). Just replacing with with a `null` value to unset the object.

tests/Doctrine/ODM/MongoDB/Tests/PersistentCollection/CollWithPHP81Types.php
```diff
@@ -14,6 +14,7 @@
  */
 class CollWithPHP81Types extends ArrayCollection
 {
+    /** @phpstan-ignore missingType.generics, missingType.generics */
     public function intersection(Collection&ArrayCollection $param): Collection&ArrayCollection
```
GromNaN, line 18: The code is repeated because the error is present twice: for the parameter and for the return type.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -1952,7 +1953,7 @@ private function doMerge(object $document, array &$visited, ?object $prevManaged
                 $prop = $this->reflectionService->getAccessibleProperty($class->name, $name);
                 assert($prop instanceof ReflectionProperty);
 
-                if (method_exists($prop, 'isInitialized') && ! $prop->isInitialized($document)) {
+                if (! $prop->isInitialized($document)) {
```
GromNaN, line 1956: `ReflectionProperty::isInitialized` is always available since PHP 7.4, and we require PHP 8.1.

phpstan.neon.dist
```diff
@@ -11,7 +11,7 @@ parameters:
   excludePaths:
     - tests/Doctrine/ODM/MongoDB/Tests/Mapping/Driver/fixtures/User.php
     - tests/Hydrators/
-    - tests/PersistentCollections/
+    - tests/PersistentCollections/ (?)
```
GromNaN, line 14: The directory doesn't exist before tests run, this syntax allows for missing directory path.

phpstan.neon.dist
```diff
@@ -22,19 +22,26 @@ parameters:
       - message: '#^Circular definition detected in type alias#'
         path: lib/Doctrine/ODM/MongoDB/Aggregation/
 
+      # Keep type assertions even when type is declared
+      - identifier: staticMethod.alreadyNarrowedType
+        path: tests/Doctrine/
+
+      - message: '#with generic class Doctrine\\ODM\\MongoDB\\Mapping\\ClassMetadata does not specify its types: T$#'
+        identifier: missingType.generics
```
GromNaN, line 30: The `ClassMetadata` is always used in generic way. We would have to use `ClassMetadata<object>` everywhere.

phpstan.neon.dist
```diff
@@ -22,19 +22,26 @@ parameters:
...
+
+
   # To be removed when reaching phpstan level 6
   checkMissingVarTagTypehint: true
   checkMissingTypehints: true
-  checkMissingIterableValueType: true
```
GromNaN, line 28: This option was removed from PHPStan 2.

SenseException (approved):

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -1749,7 +1750,7 @@ public function persist(object $document): void
      * NOTE: This method always considers documents that are not yet known to
      * this UnitOfWork as NEW.
      *
-     * @param array<string, object> $visited
+     * @param array<int, object> $visited
```
GromNaN, line 1753: This was a bug from https://github.com/doctrine/mongodb-odm/pull/2738

GromNaN:
Thanks for the review @SenseException

PR #2764, opened 2025-05-03
Allow to paginate `$search` aggregation with `searchBefore` and `searchAfter` markers

bitgandtter:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | NA

#### Summary

feat: allow to paginate search aggregation with searchBefore and searchAfter markers

### DIFF
--- a/tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/SearchTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/SearchTest.php
@@ -1313,4 +1313,116 @@ class SearchTest extends BaseTestCase
             $searchStage->getExpression(),
         );
     }
+
+    #[DataProvider('provideAutocompleteBuilders')]
+    #[DataProvider('provideCompoundBuilders')]
+    #[DataProvider('provideEmbeddedDocumentBuilders')]
+    #[DataProvider('provideEmbeddedDocumentCompoundBuilders')]
+    #[DataProvider('provideEqualsBuilders')]
+    #[DataProvider('provideExistsBuilders')]
+    #[DataProvider('provideGeoShapeBuilders')]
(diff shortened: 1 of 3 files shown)


tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/SearchTest.php
```diff
@@ -1313,4 +1313,116 @@ public function testSearchEmbeddedDocumentOperators(array $expectedOperator, Clo
...
+    #[DataProvider('provideGeoWithinBuilders')]
+    #[DataProvider('provideMoreLikeThisBuilders')]
+    #[DataProvider('provideNearBuilders')]
+    #[DataProvider('providePhraseBuilders')]
+    #[DataProvider('provideQueryStringBuilders')]
+    #[DataProvider('provideRangeBuilders')]
+    #[DataProvider('provideRegexBuilders')]
+    #[DataProvider('provideTextBuilders')]
+    #[DataProvider('provideWildcardBuilders')]
+    public function testSearchOperatorsWithSearchBefore(array $expectedOperator, Closure $createOperator): void
```
GromNaN: I think a single case is sufficient to test this feature. As it is, you're simply repeating the same test many times over for no purpose.

GromNaN (commented):
Thanks for pushing this feature; the implementation looks good.
But I think we can reduce the testing load.

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/SearchTest.php
bitgandtter: Thanks for the advice. I leave only one data provider on each method

GromNaN:
Sorry for the delay, I rebased the branch in https://github.com/doctrine/mongodb-odm/pull/2808

PR #2765, opened 2025-05-04
Update rewindable default value in Builder.php

vanex97:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

Why not disable result caching by default? I think that the current behavior is not intuitive, as it causes additional memory usage, and this feature is likely unnecessary in most situations.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Query/Builder.php
+++ b/lib/Doctrine/ODM/MongoDB/Query/Builder.php
@@ -76,7 +76,7 @@ class Builder
      */
     private bool $readOnly = false;
 
-    private bool $rewindable = true;
+    private bool $rewindable = false;
 
     /**
      * The Collection instance.
@@ -1394,7 +1394,7 @@ class Builder
         return $this;
     }
 
-    public function setRewindable(bool $rewindable = true): self
+    public function setRewindable(bool $rewindable = false): self
     {
         $this->rewindable = $rewindable;
 


alcaeus:
Starting with the TL;DR: this would most definitely be a behavioural BC break that we can't just make in a minor version. This will require careful planning to give people a chance to update.

I definitely share your concerns around memory usage, and I'm open to making this change, provided we give people a way to update their applications to accommodate this change. This will take some careful planning, but generally it will evolve emitting a deprecation notice if a cursor is rewound after being advanced, encouraging people to make the cursor rewindable explicitly. If the cursor was explicitly made rewindable, this deprecation obviously should not be emitted. We can then flip the default in 3.0 and emit a deprecation notice that any call to `rewindable(false)` should be removed as this is the new default behaviour.

@GromNaN @malarzm what do you think of the change itself, and my suggested way to implement it?

malarzm:
The plan sounds good to me :+1: Given 3.0 is in the works, this is a good time

GromNaN:
IMO the main advantage of the `CachingInterator` is to get the `count` independently of the documents. But this is not a systematic use.

We can add a setting in the `Configuration` class to specify the default behavior. We have to balance between developer experience (every result iterator is countable and rewindable) and performance (the iterator doesn't store the results). I think DX must be prioritized with a way to optimize performance; it's already possible to `setRewindable(false)` in the `Builder` and `Query` classes for that ([docs](https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/latest/reference/query-builder-api.html#disabling-result-caching)).

```php
$blogPosts = $dm->createQueryBuilder(BlogPost::class)
    ->setRewindable(false)
    ->getQuery()
    ->execute();
```

Also, we need to audit 3rd party libraries, which may assume that the query builder is always rewindable. Is it really worth breaking backward compatibility?

malarzm:
> Also, we need to audit 3rd party libraries, which may assume that the query builder is always rewindable. Is it really worth breaking backward compatibility?

We don't really need to audit them, users will get deprecation and hopefully someone will make a PR against that library :)

alcaeus:
After some internal discussion, we don't feel like this is a BC break worth making. I'll add some performance tests to the existing benchmark suite, but IMO the impact of `CachingIterator` in most use cases is not big enough to warrant completely changing the behaviour of all iterators returned from the ODM. I understand that in some instances (e.g. batch processing) you may want to have non-rewindable cursors, but considering that all documents returned in a cursor are managed in the ODM and thus won't go out of scope when the cursor iterates past them, the benefits of non-rewindable cursors are only apparent when you also clear the document manager during iteration. Without adding that, the memory footprint of non-rewindable cursor won't change compared to rewindable cursors.

Thank you for the suggestion, but for the time being we won't be making this change.

PR #2766, opened 2025-05-05
Make XSD more strict for other Document types

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | -

#### Summary

The goal of this change is to detect earlier when an XML attribute is used while it's not supported by Doctrine ODM.

BC Break: previously, the XSD schema allowed some attributes or child elements that were not used in the metadata. Example: "collection" name for embedded documents.

| **Element**                       | **document** | **embedded-document**  | **mapped-superclass**  | **query-result-document** | **view** |
|-----------------------------------|--------------|------------------------|------------------------|---------------------------|----------|
| **id**                            | Yes          | Yes                    | Yes                    | Yes                       | Yes      |
| **field**                         | Yes          | Yes                    | Yes                    | Yes                       | Yes      |
| **embed-one**                     | Yes          | Yes                    | Yes                    | Yes                       | Yes      |
| **embed-many**                    | Yes          | Yes                    | Yes                    | Yes                       | Yes      |
| **reference-one**                 | Yes          | Yes                    | Yes                    | Yes                       | Yes      |
| **reference-many**                | Yes          | Yes                    | Yes                    | Yes                       | Yes      |
| **discriminator-field**           | Yes          | Yes                    | Yes                    | Yes                       | Yes      |
| **discriminator-map**             | Yes          | Yes                    | Yes                    | Yes                       | Yes      |
| **default-discriminator-value**   | Yes          | Yes                    | Yes                    | Yes                       | Yes      |
| **lifecycle-callbacks**           | Yes          | Yes                    | Yes                    | Yes                       | Yes      |
| **also-load-methods**             | Yes          | Yes                    | Yes                    | Yes                       | Yes      |
| **indexes**                       | Yes          | Yes                    | Yes                    |                           |          |
| **search-indexes**                | Yes          |                        |                        |                           |          |
| **shard-key**                     | Yes          |                        |                        |                           |          |
| **read-preference**               | Yes          |                        |                        |                           |          |
| **schema-validation**             | Yes          |                        |                        |                           |          |
| **time-series**                   | Yes          |                        |                        |                           |          |
| **Attibute**                      | **document** | **embedded-document**  | **mapped-superclass**  | **query-result-document** | **view** |
| **db**                            | Yes          |                        | Yes                    |                           | Yes      |
| **name**                          | Yes          | Yes                    | Yes                    | Yes                       | Yes      |
| **write-concern**                 | Yes          |                        | Yes                    |                           |          |
| **collection**                    | Yes          |                        | Yes                    |                           |          |
| **capped-collection**             | Yes          |                        | Yes                    |                           |          |
| **capped-collection-size**        | Yes          |                        | Yes                    |                           |          |
| **capped-collection-max**         | Yes          |                        | Yes                    |                           |          |
| **repository-class**              | Yes          |                        | Yes                    |                           | Yes      |
| **inheritance-type**              | Yes          | Yes                    | Yes                    | Yes                       | Yes      |
| **change-tracking-policy**        | Yes          |                        | Yes                    |                           | Yes      |
| **read-only**                     | Yes          |                        | Yes                    |                           |          |
| **view**                          |              |                        |                        |                           | Yes      |
| **root-class**                    |              |                        |                        |                           | Yes      |

### DIFF
--- a/doctrine-mongo-mapping.xsd
+++ b/doctrine-mongo-mapping.xsd
@@ -16,9 +16,9 @@
     <xs:complexType>
       <xs:choice minOccurs="0" maxOccurs="unbounded">
         <xs:element name="document" type="odm:document" minOccurs="0" maxOccurs="unbounded" />
-        <xs:element name="embedded-document" type="odm:document" minOccurs="0" maxOccurs="unbounded" />
-        <xs:element name="mapped-superclass" type="odm:document" minOccurs="0" maxOccurs="unbounded" />
-        <xs:element name="query-result-document" type="odm:document" minOccurs="0" maxOccurs="unbounded" />
+        <xs:element name="embedded-document" type="odm:embedded-document" minOccurs="0" maxOccurs="unbounded" />
+        <xs:element name="mapped-superclass" type="odm:mapped-superclass" minOccurs="0" maxOccurs="unbounded" />
+        <xs:element name="query-result-document" type="odm:query-result-document" minOccurs="0" maxOccurs="unbounded" />
         <xs:element name="view" type="odm:view" minOccurs="0" maxOccurs="unbounded" />
         <xs:element name="gridfs-file" type="odm:gridfs-file" minOccurs="0" maxOccurs="unbounded" />
       </xs:choice>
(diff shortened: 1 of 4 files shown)


doctrine-mongo-mapping.xsd
```diff
@@ -16,25 +16,61 @@
...
-    <xs:attribute name="name" type="xs:NMTOKEN" />
+    <xs:choice minOccurs="0" maxOccurs="unbounded">
+      <xs:element name="id" type="odm:id" minOccurs="0" />
+      <xs:element name="field" type="odm:field" minOccurs="0" maxOccurs="unbounded" />
+      <xs:element name="embed-one" type="odm:embed-one" minOccurs="0" maxOccurs="unbounded" />
+      <xs:element name="embed-many" type="odm:embed-many" minOccurs="0" maxOccurs="unbounded" />
+      <xs:element name="reference-one" type="odm:reference-one" minOccurs="0" maxOccurs="unbounded" />
+      <xs:element name="reference-many" type="odm:reference-many" minOccurs="0" maxOccurs="unbounded" />
+      <xs:element name="also-load-methods" type="odm:also-load-methods" minOccurs="0" />
+      <xs:element name="indexes" type="odm:indexes" minOccurs="0" />
```
alcaeus: Query result documents don't have indexes. Looking at the elements possible in a document, I think we should add lifecycle callbacks and the three discriminator elements (field, map, default value) here as well.

doctrine-mongo-mapping.xsd
```diff
@@ -16,25 +16,61 @@
...
```
alcaeus, line 41: I'd have to confirm in tests, but IIRC embedded documents can be discriminated and also have lifecycle callbacks, so those elements should be added.

doctrine-mongo-mapping.xsd
```diff
@@ -16,25 +16,61 @@
...
+    </xs:choice>
+
+    <xs:attribute name="name" type="xs:string" />
+    <xs:attribute name="inheritance-type" type="odm:inheritance-type" />
   </xs:complexType>
 
   <xs:complexType name="mapped-superclass">
-    <xs:attribute name="name" type="xs:NMTOKEN" />
+    <xs:choice minOccurs="0" maxOccurs="unbounded">
```
alcaeus, line 49: For mapped superclasses, we should check the inheritance logic in the mapping driver to confirm what elements are allowed. IIRC, attributes like the write concern, capped collection information, etc. is all inherited to documents, so we may want to add it here as well.

alcaeus (approved):
LGTM. Thank you for taking the time to dig into what elements and attributes are valid for which document types.

SenseException (approved):

PR #2768, opened 2025-05-15
Fix updates to embedded documents after removal

### LINKED ISSUE #2767: Embedded documents not updated after changing a remove operation to an update
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.11.0

#### Summary

After removing a parent document with embedded documents and then changing this removal to an update in an `onFlush` listener, the embedded document updates are not applied.

#### Current behavior

After all operations have completed the embedded documents are not updated in the database, although the parent document is.

#### How to reproduce

```php
#[ODM\Document]
class TestDocument
{
    #[ODM\Id]
    public ?string $id = null;

    #[ODM\Field(type: 'int')]
    public int $removedCount = 0;

    #[ODM\EmbedMany(targetDocument: TestEmbeddedDocument::class)]
    public $embeddedDocuments;

    public function __construct(array $embeddedDocuments)
    {
        $this->embeddedDocuments = new ArrayCollection($embeddedDocuments);
    }

    public function incrementRemovedCount(): void
    {
        $this->removedCount++;
        foreach ($this->embeddedDocuments as $embeddedDocument) {
            $embeddedDocument->removedCount++;
        }
    }
}

#[ODM\EmbeddedDocument]
class TestEmbeddedDocument
{
    #[ODM\Field(type: 'int')]
    public int $removedCount = 0;
}
```

Populate and remove the documents

```php
// Save the parent document with 1 embedded document, both initialized
// with a `removedCount` of 0.
$document = new TestDocument([new TestEmbeddedDocument()]);
$dm->persist($document);
$dm->flush();

// Sometime later, not necessarily in the same script that initially saves
// the document, remove the document and trigger the event listener.
$dm->remove($document);
$dm->flush();
```

The event listener that changes the removal operation to an update operation

```php
class MyEventListener
{
    public function onFlush(OnFlushEventArgs $eventArgs): void
    {
        $dm = $eventArgs->getDocumentManager();
        $uow = $dm->getUnitOfWork();

        foreach ($uow->getScheduledDocumentDeletions() as $document) {
            if ($document instanceof TestDocument) {
                // Increment the `removedCount` on both the parent document and
                // the embedded document
                $document->incrementRemovedCount();
                // Re-persist the same document & recompute changeset to change the operation from a
                // removal to an update.
                $dm->persist($document);
                $uow->recomputeSingleDocumentChangeSet($dm->getClassMetadata(TestDocument::class), $document);
            }
        }
    }
}

```

#### Expected behavior

After all operations have completed, both the embedded documents as well as the parent document should be updated in the database to have a `removedCount` of 1. This was the case before the 2.11.x release. What actually happens is that the parent document will have a `removedCount` of 1 but the embedded document will have a `removedCount` of 0.

jfortunato:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #2767 

#### Summary

After a parent document and its embeds are removed and then subsequently persisted, the embedded documents didn't get updated. This happened because the parent document goes through an upsert operation, but `PersistenceBuilder->prepareUpsertData()` did not handle embedded associations in the same way that
`PersistenceBuilder->prepareUpdateData()` does. This fixes that issue so that an upserted parent document handles embed-many's the same was as an updated parent document.

Fixes doctrine#2767

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
@@ -287,12 +287,23 @@ final class PersistenceBuilder
                 $updateData['$set'][$mapping['name']] = $this->prepareReferencedDocumentValue($mapping, $new);
 
             // @ReferenceMany, @EmbedMany
-            } elseif (
-                $mapping['type'] === ClassMetadata::MANY && ! $mapping['isInverseSide']
-                    && $new instanceof PersistentCollectionInterface && $new->isDirty()
-                    && CollectionHelper::isAtomic($mapping['strategy'])
-            ) {
-                $updateData['$set'][$mapping['name']] = $this->prepareAssociatedCollectionValue($new, true);
+            } elseif ($mapping['type'] === ClassMetadata::MANY) {
+                if (! $mapping['isInverseSide'] && $new instanceof PersistentCollectionInterface && $new->isDirty() && CollectionHelper::isAtomic($mapping['strategy'])) {
+                    $updateData['$set'][$mapping['name']] = $this->prepareAssociatedCollectionValue($new, true);
+                } elseif ($mapping['association'] === ClassMetadata::EMBED_MANY) {
+                    foreach ($new as $key => $embeddedDoc) {
+                        if ($this->uow->isScheduledForInsert($embeddedDoc)) {
+                            continue;
+                        }
+
+                        $update = $this->prepareUpsertData($embeddedDoc);
+                        foreach ($update as $cmd => $values) {
+                            foreach ($values as $name => $value) {
+                                $updateData[$cmd][$mapping['name'] . '.' . $key . '.' . $name] = $value;
+                            }
+                        }
+                    }
+                }
             }
             // @EmbedMany and @ReferenceMany are handled by CollectionPersister
         }
--- a/lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
@@ -283,7 +283,7 @@ final class PersistenceBuilder
                 }
 
             // @EmbedMany
-            } elseif ($mapping['association'] === ClassMetadata::EMBED_MANY) {
+            } elseif ($mapping['association'] === ClassMetadata::EMBED_MANY && ! CollectionHelper::isAtomic($mapping['strategy'])) {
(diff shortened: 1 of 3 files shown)


lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
```diff
@@ -282,19 +282,35 @@ public function prepareUpsertData($document)
...
+            } elseif ($mapping['association'] === ClassMetadata::EMBED_MANY) {
+                foreach ($new as $key => $embeddedDoc) {
+                    if ($this->uow->isScheduledForInsert($embeddedDoc)) {
+                        continue;
+                    }
+
+                    $update = $this->prepareUpsertData($embeddedDoc);
+                    foreach ($update as $cmd => $values) {
+                        foreach ($values as $name => $value) {
+                            $updateData[$cmd][$mapping['name'] . '.' . $key . '.' . $name] = $value;
```
GromNaN, line 295: This modifies the object at a given position. I'm not sure about the compatibility of deleting documents from the embedded collection.

GromNaN (commented):
Thanks @jfortunato for opening this PR with a perfect test case.

I added a prefix to the test classes to prevent name conflict, and refactored the change on `PersistenceBuilder` to not have nested `if` blocks.

Could you add a test that remove an embedded document? I'm no expert on this code, but from the comments I get the impression that the change should be made in `CollectionPersister`.

GromNaN:
Oups, the last commit broke `AtomicSetTest::testAtomicUpsert`. Looking at it.

lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
```diff
@@ -282,19 +282,35 @@ public function prepareUpsertData($document)
...
-                    && $new instanceof PersistentCollectionInterface && $new->isDirty()
-                    && CollectionHelper::isAtomic($mapping['strategy'])
+                && $new instanceof PersistentCollectionInterface && $new->isDirty()
+                && CollectionHelper::isAtomic($mapping['strategy'])
             ) {
                 $updateData['$set'][$mapping['name']] = $this->prepareAssociatedCollectionValue($new, true);
             }
-            // @EmbedMany and @ReferenceMany are handled by CollectionPersister
+
+            // @ReferenceMany is handled by CollectionPersister
```
GromNaN, line 313: This comment was referring to `@ReferenceMany` which is already handled previously. I wonder if it's really relevant.

PR #2770, opened 2025-05-22
Catch `BulkWriteException`, since `WriteException` is removed from extension v2

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | -

#### Summary

Probably forgotten in https://github.com/doctrine/mongodb-odm/pull/2683

The `WriteException` was removed in MongoDB Extension v2.0. It was never thrown itself; and only its child class `BulkWriteException` is thrown.

Spotted by PHPStan.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -33,8 +33,8 @@ use Iterator as SplIterator;
 use MongoDB\BSON\ObjectId;
 use MongoDB\Collection;
 use MongoDB\Driver\CursorInterface;
+use MongoDB\Driver\Exception\BulkWriteException;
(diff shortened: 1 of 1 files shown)


lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -266,7 +266,7 @@ public function executeUpserts(array $options = []): void
                 $this->executeUpsert($document, $options);
                 $this->handleCollections($document, $options);
                 unset($this->queuedUpserts[$oid]);
-            } catch (WriteException $e) {
+            } catch (BulkWriteException $e) {
```
GromNaN, line 269: The tests were not failing = this behavior is not tested.

alcaeus (approved):

PR #2774, opened 2025-05-23
`DocumentPersister::$collection` is never `null`

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | -

#### Summary

`DocumentPersister::$collection` is set in the constructor for root document types that use it. It doesn't have to be nullable, so that all the `assert` can be removed.

https://github.com/doctrine/mongodb-odm/blob/9cc829197581df02c227f00cfa789b077949dffa/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php#L115-L130

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -89,7 +89,7 @@ use function trigger_deprecation;
  */
 final class DocumentPersister
 {
-    private ?Collection $collection = null;
+    private Collection $collection;
 
     private ?Bucket $bucket = null;
 
@@ -226,7 +226,6 @@ final class DocumentPersister
         }
 
         try {
-            assert($this->collection instanceof Collection);
             $this->collection->insertMany($inserts, $options);
         } catch (DriverException $e) {
             $this->queuedInserts = [];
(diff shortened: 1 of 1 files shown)


lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -89,7 +89,7 @@
  */
 final class DocumentPersister
 {
-    private ?Collection $collection = null;
+    private Collection $collection;
```
malarzm, line 92: This may not be set due to an early return in ctor when we're dealing with embedded documents or query results. Same goes for bucket which still is nullable

GromNaN: Indeed, good point. I'll have another look at why phpstan is sending me up an alert then.

GromNaN: @malarzm I reopened the PR. This property should never be `null`. Either it's a `Document`, then the value is set. Otherwise it's not set and this property can't be used.

GromNaN: Well, actually there's only one duplicate assert, I'll open another PR.

malarzm: Yeah but without nullable the developer will not know that property may not be set and needs to proactively check whether it's set. The best course of action would be to split DocumentPersister so it's more specialized and there's no `collection` if it can't be used. Also I'm not sure if it's worthwhile ;)

PR #2778, opened 2025-06-10
Update "Custom Mapping Types" documentation for a fully working example

fnash:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

Update "Custom Mapping Types" documentation for a fully working example of `DateTimeWithTimezoneType`

### DIFF
--- a/docs/en/reference/custom-mapping-types.rst
+++ b/docs/en/reference/custom-mapping-types.rst
@@ -25,6 +25,7 @@ a ``DateTimeImmutable`` when the data is read from the database.
     use Doctrine\ODM\MongoDB\Types\ClosureToPHP;
     use Doctrine\ODM\MongoDB\Types\Type;
     use MongoDB\BSON\UTCDateTime;
+    use RuntimeException;
 
     class DateTimeWithTimezoneType extends Type
     {
@@ -33,6 +34,10 @@ a ``DateTimeImmutable`` when the data is read from the database.
 
         public function convertToPHPValue($value): DateTimeImmutable
         {
+            if (!isset($value['utc'], $value['tz'])) {
+                throw new RuntimeException('Database value cannot be converted to date with timezone. Expected array with "utc" and "tz" keys.');
+            }
+
             $timeZone = new DateTimeZone($value['tz']);
             $dateTime = $value['utc']
                 ->toDateTime()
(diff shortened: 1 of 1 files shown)


docs/en/reference/custom-mapping-types.rst
```diff
@@ -20,8 +20,6 @@ a ``DateTimeImmutable`` when the data is read from the database.
 
     namespace My\Project\Types;
 
-    use DateTimeImmutable;
-    use DateTimeZone;
```
GromNaN: Having `use` statements for all the classes, even root-namespaced classes is in the Doctrine CS rules. That's why we have this in the documentation.

docs/en/reference/custom-mapping-types.rst
```diff
@@ -31,20 +29,25 @@ a ``DateTimeImmutable`` when the data is read from the database.
...
                 ->setTimeZone($timeZone);
 
-            return DateTimeImmutable::createFromMutable($dateTime);
+            return \DateTimeImmutable::createFromMutable($dateTime);
         }
 
         public function convertToDatabaseValue($value): array
         {
-            if (! isset($value['utc'], $value['tz'])) {
-                throw new RuntimeException('Database value cannot be converted to date with timezone. Expected array with "utc" and "tz" keys.');
```
GromNaN, line 47: Good catch, this check should be in `convertToPHPValue`.

docs/en/reference/custom-mapping-types.rst
```diff
@@ -31,20 +29,25 @@ a ``DateTimeImmutable`` when the data is read from the database.
...
+            if (!$value instanceof \DateTimeImmutable) {
```
GromNaN: Any `DateTimeInterface` would be accepted here.

```suggestion
            if (!$value instanceof \DateTimeInterface) {
```

GromNaN (changes requested):
Good catch. Could you revert the namespace changes please?

docs/en/reference/custom-mapping-types.rst
fnash: Indeed but it will not work with a \DateTime Document field for example because ::convertToPHPValue() returns a \DateTimeImmutable anyway.
Enforcing only one type in the two conversion is better IMO.

GromNaN: OK, I would have been more tolerant in the library code, but we can be stricter for the tutorial and application code.

GromNaN (approved):
Thank you @fnash

PR #2759, opened 2025-04-25
[Encryption] Add support for Field-Level Automatic and Queryable Encryption

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | [PHPORM-317](https://jira.mongodb.org/browse/PHPORM-317)

#### Summary

⚠️ This feature requires a MongoDB Atlas or Enterprise license.

> [Client-Side Field Level Encryption (CSFLE)](https://www.mongodb.com/docs/manual/core/csfle/) is a feature that enables you to encrypt data in your application before you send it over the network to MongoDB. With CSFLE enabled, no MongoDB product has access to your data in an unencrypted form.

> [Queryable Encryption](https://www.mongodb.com/docs/manual/core/queryable-encryption/) gives you the ability to perform the following tasks:
> - Encrypt sensitive data fields from the client-side.
> - Store sensitive data fields as fully randomized encrypted data on the database server-side.
> - Run expressive queries on the encrypted data.

Implementation:
- Add `#[Encrypt]` attribute with [field encryption options](https://www.mongodb.com/docs/manual/core/queryable-encryption/fundamentals/encrypt-and-query/#configure-encrypted-fields-for-optimal-search-and-storage).
- Add `autoEncryption` configuration with [options to create an encryption client](https://www.php.net/mongodb-driver-clientencryption.construct) (MongoDB collection that stores encryption keys for each field)

The encrypted collection must be created using the schema manager (`odm:schema:create` command)

PR #2779, opened 2025-06-12
PHPORM-13 Feature Queryable Encryption

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | [PHPORM-13](https://jira.mongodb.org/browse/PHPORM-13)

Add [Automatic Queryable Encryption](https://www.mongodb.com/docs/manual/core/queryable-encryption/) to the MongoDB Documents. This feature requires a MongoDB Enterprise license or a MongoDB Atlas cluster.

- [x] https://github.com/doctrine/mongodb-odm/pull/2759
- [x] https://github.com/doctrine/mongodb-odm/pull/2781
- [x] https://github.com/doctrine/mongodb-odm/pull/2780
- [x] https://github.com/doctrine/mongodb-odm/pull/2783
- [x] https://github.com/doctrine/mongodb-odm/pull/2784
- [x] https://github.com/doctrine/mongodb-odm/pull/2785
- [x] https://github.com/doctrine/mongodb-odm/pull/2788
- [x] https://github.com/doctrine/mongodb-odm/pull/2790
- [x] https://github.com/doctrine/mongodb-odm/pull/2793
- [x] https://github.com/doctrine/mongodb-odm/pull/2801

Bundle integration: https://github.com/doctrine/DoctrineMongoDBBundle/pull/897


lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -2176,6 +2183,15 @@ public function isView(): bool
         return $this->isView;
     }
 
+    public function isDocument(): bool
```
GromNaN, line 2201: Note to leverage this new method in the SchemaManager. https://github.com/doctrine/mongodb-odm/blob/b0909cadb4f49c3bb646da367124ca0abd7d4aa7/lib/Doctrine/ODM/MongoDB/SchemaManager.php#L75

alcaeus (commented):
Requesting a Copilot for additional review since I already reviewed a number of these PRs and authored some of them, but LGTM so far.

alcaeus (approved):
LGTM with phpstan errors fixed.

docs/en/cookbook/queryable-encryption.rst
```diff
@@ -0,0 +1,265 @@
...
+ensures that sensitive data is never exposed in an unencrypted state on the
+server, in system logs, or in backups.
+
+This tutorial will guide you through the process of securing a document's
+fields using queryable encryption, from defining the document and configuring
+the connection to storing and querying the encrypted data.
+
+.. note::
+
+    Queryable Encryption is only available on MongoDB Enterprise 8.0+ or
```
jmikola:
```suggestion
    Queryable Encryption is only available on MongoDB Enterprise 7.0+ or
```

Per [Queryable Encryption General Availability](https://www.mongodb.com/docs/v7.0/release-notes/7.0-compatibility/#queryable-encryption-general-availability) in the 7.0 release notes.

docs/en/cookbook/queryable-encryption.rst
```diff
@@ -0,0 +1,265 @@
...
+  ``DocumentManager`` through ``Configuration::setKmsProvider()``. If you need
+  to work with multiple KMS providers, you must manually configure the
+  ``kmsProviders`` array and pass it as a driver option, bypassing the ODM's
+  helper method.
+- Automatic generation of the ``encryptedFieldsMap`` is not compatible with
+  ``SINGLE_COLLECTION`` inheritance. Because all classes in the hierarchy
+  share a single collection, they must also share a single encryption schema.
+  To use QE with inheritance, you must manually define the complete
+  ``encryptedFieldsMap`` for the entire hierarchy and provide it directly in
+  the client options, bypassing the ODM's automatic generation.
```
jmikola: Is it likely that users would run into conflicts attempting to do this? I suppose that depends on whether inheriting classes might have conflicting mappings (e.g. a field should be considered encrypted for one class but not another). If so, I wonder if it's worth warning users that this may be a bad idea.

docs/en/reference/attributes-reference.rst
```diff
@@ -336,6 +336,93 @@ Unlike normal documents, embedded documents cannot specify their own database or
...
 multiple document classes, and even other embedded documents!
 
+.. _encrypt_attribute:
+
+#[Encrypt]
+----------
+
+The ``#[Encrypt]`` attribute is used to define an encrypted field mapping for a
+document property. It allows you to configure fields for encryption and queryable
+encryption in MongoDB.
```
jmikola, line 346: Should this be "automatic and queryable encryption"?

docs/en/reference/attributes-reference.rst
```diff
@@ -336,6 +336,93 @@ Unlike normal documents, embedded documents cannot specify their own database or
...
+- ``min``, ``max`` - Specify minimum and maximum (inclusive) queryable values
+  for a field when possible, as smaller bounds improve query efficiency. If
+  querying values outside of these bounds, MongoDB returns an error.
+- ``sparsity``, ``precision``, ``trimFactor``, ``contention`` - For advanced
+  users only. The default values for these options are suitable for the majority
+  of use cases, and should only be modified if your use case requires it.
+
+.. note::
+
+    Queryable encryption is only supported in MongoDB version 8.0 and later.
```
jmikola:
```suggestion
    Queryable encryption is only supported in MongoDB version 7.0 and later.
```

docs/en/reference/attributes-reference.rst
```diff
@@ -336,6 +336,93 @@ Unlike normal documents, embedded documents cannot specify their own database or
...
+    #[Document]
+    class Client
+    {
+        #[Field]
+        #[Encrypt(queryType: EncryptQuery::Equality)]
+        public string $name;
+    }
+
+The ``#[Encrypt]`` attribute can be added to a class with `#[EmbeddedDocument]`_.
+This will encrypt the entire embedded document, in the field that contains it.
```
jmikola, line 383: Assuming an embedded document is mapped with `#[Encrypt]` in the parent document but the embedded document also has some field(s) that were mapped with `#[Encrypt]` (possibly queryable)? I assume the higher-level mapping would take precedence, but is this a scenario that should yield an error when ODM first analyzes the mappings?

docs/en/reference/migrating-schemas.rst
```diff
@@ -15,6 +15,44 @@ problem!
...
+    $schemaManager->createCollections();
+
+For a specific document class, you can use the `createDocumentCollection()`
+method with the class name as an argument:
+
+    <?php
+
+    $schemaManager->createDocumentCollection(Person::class);
+
+Once the collection is created, you can also set up indexes with ``ensureIndexes``,
```
jmikola, line 48: Unrelated to this PR, but were there ever plans to add `createIndexes()` as an alias of `ensureIndexes()` and deprecate the old method? The "ensure" name is a relic of earlier MongoDB APIs.

doctrine-mongo-mapping.xsd
```diff
@@ -183,7 +184,31 @@
...
+    <xs:attribute name="queryType">
+      <xs:simpleType>
+        <xs:restriction base="xs:string">
+          <xs:enumeration value="equality" />
+          <xs:enumeration value="range" />
+        </xs:restriction>
+      </xs:simpleType>
+    </xs:attribute>
+    <xs:attribute name="min" type="xs:string" />
+    <xs:attribute name="max" type="xs:string" />
```
jmikola, line 197: Are these mapped as `xs:string` because they may be either integer or floats? [`ClientEncryption::encrypt()`](https://www.php.net/manual/en/mongodb-driver-clientencryption.encrypt.php) notes they're "mixed", but I think they're still numeric. If so, a comment above might be helpful.

doctrine-mongo-mapping.xsd
```diff
@@ -254,6 +279,7 @@
       <xs:element name="discriminator-field" type="odm:discriminator-field" minOccurs="0" />
       <xs:element name="discriminator-map" type="odm:discriminator-map" minOccurs="0" />
       <xs:element name="default-discriminator-value" type="odm:default-discriminator-value" minOccurs="0" />
+      <xs:element name="encrypt" type="odm:encrypt-field" minOccurs="0" maxOccurs="1" />
```
jmikola: Why do none of the other `xs:element` tags here specifying "maxOccurs"? Is that an oversight? I realize this is outside the scope of the current PR.

Same question below for `embed-many`.

doctrine-mongo-mapping.xsd
```diff
@@ -183,7 +184,31 @@
...
+    <xs:attribute name="contention" type="xs:integer" />
+  </xs:complexType>
+
+  <xs:complexType name="encrypt-embedded-document">
+  </xs:complexType>
+
   <xs:complexType name="field">
+    <xs:choice minOccurs="0" maxOccurs="unbounded">
+      <xs:element name="encrypt" type="odm:encrypt-field" minOccurs="0" maxOccurs="1" />
+    </xs:choice>
```
jmikola, line 210: What does including one `xs:element` within `xs:choice` achieve? Is this necessary to allow a single `<encrypt>` element to be specified within `<field>`? I'm a bit rusty on how `xs:choice` is employed, so an explanation would be genuinely helpful to refresh my understanding.

I noted that everything else here is an attribute, so this would be the first tag that can be nested within `<field>`.

lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -135,6 +146,50 @@ class Configuration
...
+     *
+     * @return array{keyVaultClient?: Client|Manager, keyVaultNamespace: string, kmsProviders: array<string, mixed>, tlsOptions?: array<string, mixed>}
+     */
+    public function getClientEncryptionOptions(): array
+    {
+        if (! isset($this->attributes['kmsProvider'])) {
+            throw ConfigurationException::clientEncryptionOptionsNotSet();
+        }
+
+        return array_intersect_key($this->getAutoEncryptionOptions(), [
```
jmikola, line 186: Noted that this corresponds to supported options in [`Manager::createClientEncryption()`](https://www.php.net/manual/en/mongodb-driver-manager.createclientencryption.php).

lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -135,6 +146,50 @@ class Configuration
...
+    public function getDriverOptions(): array
+    {
+        $driverOptions = [
+            'driver' => [
+                'name' => 'doctrine-odm',
+                'version' => self::getVersion(),
+            ],
+        ];
+
+        if (isset($this->attributes['kmsProvider'])) {
```
jmikola, line 166: Is there a particular reason that `kmsProvider` is the deciding factor to assemble `autoEncryption` options? Is this because `keyVaultNamespace` is set by default and at least one KMS provider must be specified?

lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -651,6 +706,99 @@ public function isLazyGhostObjectEnabled(): bool
...
+    {
+        $this->attributes['defaultMasterKey'] = $masterKey;
+    }
+
+    /**
+     * Set the options for auto-encryption.
+     *
+     * @see https://www.php.net/manual/en/mongodb-driver-manager.construct.php
+     *
+     * @param array{ keyVaultClient?: Client|Manager, keyVaultNamespace?: string, tlsOptions?: array<string, mixed>, schemaMap?: array<string, mixed>, encryptedFieldsMap?: array<string, mixed>, extraOptions?: array<string, mixed>} $options
```
jmikola: I think a few additional options are missing if you compare with the [autoEncryption table](https://www.php.net/manual/en/mongodb-driver-manager.construct.php#mongodb-driver-manager.construct-autoencryption) in the PHPC docs. The `bypass` booleans at least.

lib/Doctrine/ODM/MongoDB/ConfigurationException.php
```diff
@@ -27,4 +29,29 @@ public static function proxyDirMissing(): self
...
     }
+
+    public static function clientEncryptionOptionsNotSet(): self
+    {
+        return new self('MongoDB client encryption options are not set in configuration');
+    }
+
+    public static function kmsProviderTypeRequired(): self
+    {
+        return new self('The KMS provider "type" is required.');
```
jmikola: Should exception messages include a trailing period if they're a single sentence? The message for `clientEncryptionOptionsNotSet()` does not.

tests/Doctrine/ODM/MongoDB/Tests/BaseTestCase.php
```diff
@@ -192,6 +193,24 @@ protected function skipTestIfSharded(string $className): void
...
+        $buildInfo = $this->getPrimaryServer()->executeCommand(
+            DOCTRINE_MONGODB_DATABASE,
+            new Command(['buildInfo' => 1]),
+        )->toArray()[0];
+
+        if (! in_array('enterprise', $buildInfo->modules ?? [])) {
+            $this->markTestSkipped('Queryable Encryption test requires MongoDB Atlas or Enterprise');
+        }
+
+        $this->requireVersion($buildInfo->version, '8.0', '<', 'Queryable Encryption test requires MongoDB 8.0 or higher');
```
GromNaN:
```suggestion
        $this->requireVersion($buildInfo->version, '7.0', '<', 'Queryable Encryption test requires MongoDB 7.0 or higher');
```

docs/en/reference/migrating-schemas.rst
GromNaN: It's used in the [data fixtures package](https://github.com/doctrine/data-fixtures/blob/2.1.x/src/Purger/MongoDBPurger.php#L89), but we can introduce a new method and deprecate the old one. [PHPORM-378](https://jira.mongodb.org/browse/PHPORM-378)

doctrine-mongo-mapping.xsd
GromNaN: They can also be date. In `XmlDriver::addEncryptionMapping()` the value is converted into BSON using `Type::convertToDatabaseValue()` on the field type.

doctrine-mongo-mapping.xsd
GromNaN: If took inspiration with `<xs:complexType name="document">`. It can be replaced with `<xs:sequence>` with the same effect.

lib/Doctrine/ODM/MongoDB/Configuration.php
GromNaN: You guessed right. There are 2 required parameters: `kmsProvider` and `keyVaultNamespace`, but the ODM defines a default value for `keyVaultNamespace`.

lib/Doctrine/ODM/MongoDB/Configuration.php
GromNaN: In the comment, I mentioned the `ClientEncryption` class, but it's the same thing, right?

lib/Doctrine/ODM/MongoDB/Configuration.php
GromNaN: Exact, I added `bypassAutoEncryption` and `bypassQueryAnalysis`. I think I initially omitted them because I don't understand what would be the purpose of using them with the ODM.

jmikola: I won't claim they're relevant for ODM users, but they seem like trivial things to add.

lib/Doctrine/ODM/MongoDB/Configuration.php
jmikola: Yes. The Manager method is just a shortcut for the ClientEncryption constructor.

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/Encrypt.php
```diff
@@ -0,0 +1,47 @@
...
+ */
+#[Attribute(Attribute::TARGET_CLASS | Attribute::TARGET_PROPERTY)]
+final class Encrypt implements Annotation
+{
+    public int|float|Int64|Decimal128|UTCDateTime|null $min;
+    public int|float|Int64|Decimal128|UTCDateTime|null $max;
+
+    /**
+     * @param EncryptQuery|null $queryType  Set the query type for the field, null if not queryable.
+     * @param int<1, 4>|null    $sparsity
```
jmikola, line 30: Noted that the manual docs above specify a range of 1-4, but the [CSFLE spec](https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.md) and our own PHPC docs report this as a positive 64-bit integer. May be worth a ticket to get this clarified in the CSFLE spec.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
jmikola: Do you intend to make this change in the current PR, or create a ticket to do so later? No objections to doing it now.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -67,6 +71,15 @@
  *    get the whole class name, namespace inclusive, prepended to every property in
  *    the serialized representation).
  *
+ * @phpstan-type EncryptConfig array{
```
jmikola, line 76: This looks very similar to the signature of the Encrypt annotation's constructor. I assume that's intentional, as the annotation would be used to create this mapping.

I assume there's no way to avoid the duplication, but would it make sense to include a comment in one of the two places to note that the structures should remain consistent? Disregard if this comes up with many annotations and should be assumed knowledge.

lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
```diff
@@ -919,6 +933,21 @@ private function addGridFSMappings(ClassMetadata $class, SimpleXMLElement $xmlRo
         $xmlRoot->metadata->addAttribute('field', 'metadata');
         $this->addEmbedMapping($class, $xmlRoot->metadata, ClassMetadata::ONE);
     }
+
+    /** @psalm-return EncryptConfig */
+    private function addEncryptionMapping(SimpleXMLElement $encrypt, string $type): array
+    {
+        $encryptMapping = [];
+        foreach ($encrypt->attributes() as $encryptKey => $encryptValue) {
```
jmikola: The earlier `foreach` for field attributes uses `as $key => $value`. I think that'd be fine here to make the code below a bit more concise with no loss in readability.

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AttributeDriver.php
```diff
@@ -264,6 +266,8 @@ public function loadMetadataForClass($className, PersistenceClassMetadata $metad
                     $mapping['version'] = true;
                 } elseif ($propertyAttribute instanceof ODM\Lock) {
                     $mapping['lock'] = true;
+                } elseif ($propertyAttribute instanceof ODM\Encrypt) {
+                    $mapping['encrypt'] = (array) $propertyAttribute;
```
jmikola, line 270: Is the structure of `$mapping['encrypt']` defined anywhere? Disregard if there is no need to do so.

I noted that `ClassMetadata::mapField()`, which is passed the `$mapping` array, doesn't inspect `encrypt` at all. I assume this is only read in EncryptedFieldsMapGenerator.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -644,10 +645,29 @@ public function createDocumentCollection(string $documentName, ?int $maxTimeMs =
...
-        $this->dm->getDocumentDatabase($documentName)->createCollection(
-            $class->getCollection(),
-            $this->getWriteOptions($maxTimeMs, $writeConcern, $options),
-        );
+        // Encryption is enabled only if the KMS provider is set and at least one field is encrypted
+        if ($this->dm->getConfiguration()->getDefaultKmsProvider()) {
+            $encryptedFields = (new EncryptedFieldsMapGenerator($this->dm->getMetadataFactory()))->getEncryptedFieldsMapForClass($class->name);
+
+            if ($encryptedFields) {
+                $options['encryptedFields'] = ['fields' => $encryptedFields];
```
jmikola: I might be mixing this up with the previous bundle PR, but did we talk about not hiding the extra top-level `fields` key from the user? That might not be relevant here, as `$encryptedFields` is just being produced by the generator. I can also see how waiting to apply the top-level field here makes the conditional easier to write since it can check for an empty array.

lib/Doctrine/ODM/MongoDB/Utility/EncryptedFieldsMapGenerator.php
```diff
@@ -0,0 +1,140 @@
...
+    /**
+     * Generate the encryption field map from the class metadata.
+     *
+     * @param class-string $className
+     *
+     * @return array<int, array{path: string, bsonType: string, keyId: ?string}>
+     */
+    public function getEncryptedFieldsMapForClass(string $className): array
+    {
+        $classMetadata = $this->classMetadataFactory->getMetadataFor($className);
```
jmikola, line 60: Does this warrant a `$classMetadata->isDocument()` check like you have in `getEncryptedFieldsMap()`, or would you want to yield an exception here if the user specified a non-document class name?

Disregard if one of the thrown exceptions in `createEncryptedFieldsMapForClass()` covers this edge case.

lib/Doctrine/ODM/MongoDB/Utility/EncryptedFieldsMapGenerator.php
```diff
@@ -0,0 +1,140 @@
...
+            throw MappingException::rootDocumentCannotBeEncrypted($classMetadata->getName());
+        }
+
+        if (isset($visitedClasses[$classMetadata->getName()])) {
+            // Prevent infinite recursion due to circular references in the metadata
+            return;
+        }
+
+        foreach ($classMetadata->fieldMappings as $mapping) {
+            // @todo support polymorphic types and inheritance?
```
jmikola: Does this relate to the documentation comment about not combining CSFLE with single-collection inheritance? This comment suggests that any and all inheritance may be unsupported.

Should there be a separate PHPORM ticket created and referenced in this todo?

lib/Doctrine/ODM/MongoDB/Utility/EncryptedFieldsMapGenerator.php
```diff
@@ -0,0 +1,140 @@
...
+            // Add fields recursively
+            if ($mapping['embedded'] ?? false) {
+                $embedMetadata = $this->classMetadataFactory->getMetadataFor($mapping['targetDocument']);
+
+                // When the embedded document class is encrypted, the field is encrypted,
+                // but none of the embedded fields are encrypted separately.
+                if ($embedMetadata->isEncrypted) {
+                    $mapping['encrypt'] ??= []; // @todo get the keyId
```
jmikola: Not sure why @alcaeus marked the copilot comment as resolved, but this seem outstanding. Perhaps improve the comment and create a PHPORM ticket?

lib/Doctrine/ODM/MongoDB/Utility/EncryptedFieldsMapGenerator.php
```diff
@@ -0,0 +1,140 @@
...
+     * @param array<class-string, true> $visitedClasses
+     * @phpstan-param ClassMetadata<T> $classMetadata
+     *
+     * @return Generator<int, array{path: string, bsonType: string, keyId: ?string}>
+     *
+     * @template T of object
+     */
+    private function createEncryptedFieldsMapForClass(
+        ClassMetadata $classMetadata,
+        string $path = '',
```
jmikola: Would `$pathPrefix` be a more descriptive name for this variable. It looks like it's used to _create_ the `path` value of the field mapping, and is always concatenated.

lib/Doctrine/ODM/MongoDB/Utility/EncryptedFieldsMapGenerator.php
```diff
@@ -0,0 +1,140 @@
...
+                        $visitedClasses + [$classMetadata->getName() => true],
+                    );
+                }
+            }
+
+            if (! isset($mapping['encrypt'])) {
+                continue;
+            }
+
+            $field = [
```
jmikola, line 119: [Specify Fields to Encrypt](https://www.mongodb.com/docs/manual/core/queryable-encryption/qe-create-encryption-schema/#specify-fields-to-encrypt.) mentions that any field other than `_id` can be specified. Does that warrant adding a check and corresponding exception, or would you rather defer to the server to enforce that?

lib/Doctrine/ODM/MongoDB/Utility/EncryptedFieldsMapGenerator.php
```diff
@@ -0,0 +1,140 @@
...
+                'path' => $path . $mapping['name'],
+                'bsonType' => match ($mapping['type']) {
+                    ClassMetadata::ONE, Type::HASH => 'object',
+                    ClassMetadata::MANY, Type::COLLECTION => 'array',
+                    Type::INT, Type::INTEGER => 'int',
```
jmikola, line 124: The [BSON type](https://www.mongodb.com/docs/manual/reference/bson-types/#std-label-bson-types) mappings for encryption differentiates between 32-bit and 64-bit integers (`int` and `long`, respectively). This could be problematic, as the match here is going to limit all integer-mapped fields to 32-bit. Defaulting to 64-bit isn't necessarily a workaround, as the server is going to expect matching types for any equality queries.

I believe this is being tracked by [SERVER-75897](https://jira.mongodb.org/browse/SERVER-75897).

lib/Doctrine/ODM/MongoDB/Utility/EncryptedFieldsMapGenerator.php
```diff
@@ -0,0 +1,140 @@
...
+
+            if (! isset($mapping['encrypt'])) {
+                continue;
+            }
+
+            $field = [
+                'path' => $path . $mapping['name'],
+                'bsonType' => match ($mapping['type']) {
+                    ClassMetadata::ONE, Type::HASH => 'object',
+                    ClassMetadata::MANY, Type::COLLECTION => 'array',
```
jmikola, line 123: I see `ONE` and `MANY` are documented as mapping types in ClassMetadata. Are these the only values that get mixed in with the other constants in `Doctrine\ODM\MongoDB\Types\Type`? I'm curious as to why everything isn't all defined in the same place.

Also, this entire `match` seems like something that might be better defined in a shared utility method. It seems like it could also be used to generate schema validations (AFAIK that feature does not yet exist in ODM). I won't push for extraction in this PR, but it seemed worth mentioning.

lib/Doctrine/ODM/MongoDB/Utility/EncryptedFieldsMapGenerator.php
```diff
@@ -0,0 +1,140 @@
...
+                    Type::BOOL, Type::BOOLEAN => 'bool',
+                    default => throw new LogicException(sprintf('Type "%s" is not supported in encrypted fields map.', $mapping['type'])),
+                },
+                'keyId' => null, // Generate the key automatically
+            ];
+
+            // When queryType is null, the field is not queryable
+            if (isset($mapping['encrypt']['queryType'])) {
+                $field['queries'] = array_filter($mapping['encrypt'], static fn ($v) => $v !== null);
+                assert($field['queries']['queryType'] instanceof EncryptQuery);
```
jmikola: Is this a valid place to `assert()`? AFAIK, they may still be disabled at runtime. Perhaps there's no issue as we'd get a fatal error on the subsequent line if this actually wasn't a BackedEnum instance, and the `assert()` here is actually needed to satisfy static analysis?

lib/Doctrine/ODM/MongoDB/Utility/EncryptedFieldsMapGenerator.php
```diff
@@ -0,0 +1,140 @@
...
+                    Type::BINDATA, Type::BINDATABYTEARRAY, Type::BINDATAFUNC, Type::BINDATACUSTOM, Type::BINDATAUUID, Type::BINDATAMD5, Type::BINDATAUUIDRFC4122 => 'binData',
+                    Type::BOOL, Type::BOOLEAN => 'bool',
+                    default => throw new LogicException(sprintf('Type "%s" is not supported in encrypted fields map.', $mapping['type'])),
+                },
+                'keyId' => null, // Generate the key automatically
+            ];
+
+            // When queryType is null, the field is not queryable
+            if (isset($mapping['encrypt']['queryType'])) {
+                $field['queries'] = array_filter($mapping['encrypt'], static fn ($v) => $v !== null);
```
jmikola: Since EncryptQuery in the Encrypt attribute is nullable, is there anything to enforce that no other options are specified if the query type is `null`? Apologies if I missed some validation elsewhere in the PR.

composer.json
```diff
@@ -30,7 +30,7 @@
         "doctrine/persistence": "^3.2 || ^4",
         "friendsofphp/proxy-manager-lts": "^1.0",
         "jean85/pretty-package-versions": "^1.3.0 || ^2.0.1",
-        "mongodb/mongodb": "^1.21 || ^2.0@dev",
+        "mongodb/mongodb": "~1.21.2 || ^2.1.1@dev",
```
GromNaN: This can be reverted if we remove the [v1.x branch](https://github.com/mongodb/mongo-php-library/tree/v1.x).

lib/Doctrine/ODM/MongoDB/Utility/EncryptedFieldsMapGenerator.php
GromNaN: Removed this todo. The `keyId` can change when DEK are rotated, so it doesn't make sense to set the `keyId` in the document metadata.
This might be reworked in [PHPORM-374](https://jira.mongodb.org/browse/PHPORM-374) with a `keyAltName` option.

doctrine-mongo-mapping.xsd
GromNaN: `minOccurs="1"` and `maxOccurs="1"` are the default values, they can be omitted.

docs/en/reference/attributes-reference.rst
GromNaN: I added the sentence "Fields within an encrypted embedded document cannot be individually encrypted."

docs/en/reference/attributes-reference.rst
GromNaN: Yes, updated.

lib/Doctrine/ODM/MongoDB/Mapping/Annotations/Encrypt.php
GromNaN: I created [DRIVERS-3267](https://jira.mongodb.org/browse/DRIVERS-3267)

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
GromNaN: I added a comment with a `{@see ... }` mark, so that PHPStorm has a reference of the `Encrypt` class.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
GromNaN: I don't think it's that simple. There are details to work out with File-type collections, which are sometimes Documents and sometimes not. I'll keep that for later.

lib/Doctrine/ODM/MongoDB/Mapping/Driver/AttributeDriver.php
GromNaN: `mapField` first argument has `@psalm-type FieldMappingConfig` which contains `encrypt` property. I think all types are good.

lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
GromNaN: Indeed, updated.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
GromNaN: I added the `fields` level to the generated encrypted fields map.
This is distinct from the bundle PR https://github.com/doctrine/DoctrineMongoDBBundle/pull/912 that in which we manipulate the field map returned from the collection metadata on the server.

lib/Doctrine/ODM/MongoDB/Utility/EncryptedFieldsMapGenerator.php
GromNaN: We rely on the server error for this.

lib/Doctrine/ODM/MongoDB/Utility/EncryptedFieldsMapGenerator.php
GromNaN: I added the check just after reading the class metadata.

lib/Doctrine/ODM/MongoDB/Utility/EncryptedFieldsMapGenerator.php
GromNaN: I rename to `$parentPath`, which is more descriptive.

lib/Doctrine/ODM/MongoDB/Utility/EncryptedFieldsMapGenerator.php
GromNaN: So, do you suggest we should change it to `long`?
```suggestion
                    Type::INT, Type::INTEGER => 'long',
```

lib/Doctrine/ODM/MongoDB/Utility/EncryptedFieldsMapGenerator.php
GromNaN: We can wait to have a 2nd use-case before adding complexity.

lib/Doctrine/ODM/MongoDB/Utility/EncryptedFieldsMapGenerator.php
GromNaN: I removed this assert, it's covered by the phpstan type.

lib/Doctrine/ODM/MongoDB/Utility/EncryptedFieldsMapGenerator.php
jmikola: Ah, and now I understand why `keyAltName` exists! That only took five years 😛

lib/Doctrine/ODM/MongoDB/Utility/EncryptedFieldsMapGenerator.php
jmikola: According to the [CSFLE Encryption Schemas: `encrypt` docs](https://www.mongodb.com/docs/manual/core/csfle/reference/encryption-schemas/#mongodb-autoencryptkeyword-autoencryptkeyword.encrypt), we can specify an array of types for `bsonType`. Can we use `['int', 'long']` here to remain flexible?

I'll also share this question in the CSFLE Slack for a sanity check.

jmikola: @kevinalbs replied in Slack:

> Notably: though CSFLE schema (JSON Schema) supports multiple BSON types, a QE schema (`encryptedFields`) only supports a single valued bsonType.

So I don't think we have a user-friendly solution here unless we prefer "long" (since it can contain _any_ integer value) and have ODM _always_ encodes the encrypted field as a `MongoDB\BSON\Int64` to ensure even values in the 32-bit range are stored as int64 and thus satisfy the QE `bsonType` restriction.

GromNaN: When sending a int64 to a field encoded with the int type, we get this error:
```
mongocryptd error: Cannot encrypt element of type long because schema requires that type is one of: [ int ]
```

Also, I cannot set multiple types:

```
mongocryptd error: BSON field 'create.encryptedFields.fields.bsonType' is the wrong type 'array', expected type 'string'                                                            
```

If I change the bson type to `long`, then it doesn't accept an `int`.

It seems that its a limitation until we add the `Int64` type in the ODM.

GromNaN: New Int64 type: https://github.com/doctrine/mongodb-odm/pull/2805

doctrine-mongo-mapping.xsd
```diff
@@ -535,7 +562,7 @@
       <xs:element name="field" type="odm:search-index-field" minOccurs="0" maxOccurs="unbounded" />
       <!-- Note: custom analyzers are intentionally unsupported in XML -->
       <xs:element name="synonym" type="odm:search-index-synonym" minOccurs="0" maxOccurs="unbounded" />
-      <xs:element name="stored-source" type="odm:search-index-stored-source" minOccurs="0" maxOccurs="1" />
+      <xs:element name="stored-source" type="odm:search-index-stored-source" minOccurs="0" />
```
jmikola, line 565: Noted that `1` is the default value for `maxOccurs`.

lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -651,6 +707,107 @@ public function isLazyGhostObjectEnabled(): bool
...
+        // To use "Automatic Credentials", the provider options must be an empty document.
+        // Fix the empty array to an empty stdClass object, as the driver expects it.
+        if ($kmsProviderOpts === []) {
+            $kmsProviderOpts = new stdClass();
+        }
+
+        return [
+            // Each kmsProvider must be an object, it can be empty
+            'kmsProviders' => [$kmsProviderName => $kmsProviderOpts],
+            'keyVaultNamespace' => $this->getDefaultDB() . '.datakeys',
```
jmikola, line 807: Is `$this->getDefaultDB() . '.datakeys'` documented as the default `keyValueNamespace` value in any of the ODM documentation? I think whatever value we use here should be mentioned somewhere.

I'm aware that the [spec](https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.md#keyvaultnamespace) uses `db.datakeys` as an example value. Meanwhile, the [server manual](https://www.mongodb.com/docs/v7.0/core/queryable-encryption/fundamentals/keys-key-vaults/#key-vault-collection-name) uses `encryption.__keyVault`. One thing I'd confirm is whether it's sensible to use a collection in the application database. This might be a question to run by other ODMs that have integrated CSFLE, or simply ask in `#dbx-encryption` or one of the other CSFLE channels.

One concern here is that there might be a conflict in an app with a DataKey model class if ODM attempts to use `db.datakeys` as its name. In that case, using an underscore prefix (as the server manual does) might be beneficial if we're going to default to the same database.

lib/Doctrine/ODM/MongoDB/Utility/EncryptedFieldsMapGenerator.php
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    /**
+     * Generate the encryption field map from the class metadata.
+     *
+     * @param class-string $className
+     *
+     * @return array{fields: array<int, array{path: string, bsonType: string, keyId: null}>}|null
+     */
+    public function getEncryptedFieldsMapForClass(string $className): ?array
```
jmikola: "EncryptedFieldsMap" is the term for a map of class names to EncryptedFields objects, which are mentioned in the [server manual](https://www.mongodb.com/docs/manual/core/queryable-encryption/fundamentals/enable-qe/). I think this method can be renamed to `getEncryptedFieldsForClass` since there is no map at this level. We're going to return an object with a `fields` array.

lib/Doctrine/ODM/MongoDB/Utility/EncryptedFieldsMapGenerator.php
```diff
@@ -0,0 +1,147 @@
...
+
+    /**
+     * @param array<class-string, true> $visitedClasses
+     * @phpstan-param ClassMetadata<T> $classMetadata
+     *
+     * @return Generator<int, array{path: string, bsonType: string, keyId: null}>
+     *
+     * @template T of object
+     */
+    private function createEncryptedFieldsMapForClass(
```
jmikola: Likewise, this method would be `createEncryptedFieldsForClass` since we're not returning a map but rather an iterable of field definitions.

tests/Doctrine/ODM/MongoDB/Tests/ConfigurationTest.php
```diff
@@ -40,4 +44,133 @@ public function testEnableTransactionalFlush(): void
         $c->setUseTransactionalFlush(false);
         self::assertFalse($c->isTransactionalFlushEnabled(), 'Transactional flush is disabled after setTransactionalFlush(false)');
     }
+
+    public function testLocalKmsProvider(): void
+    {
+        $c = new Configuration();
+        $c->setKmsProvider(['type' => 'local', 'key' => '1234567890123456789012345678901234567890123456789012345678901234']);
```
jmikola: This need only be a 96-byte base64-encoded string. Would you prefer to use something like `str_repeat('a', 96)` or a class constant to make these tests more concise?

tests/Doctrine/ODM/MongoDB/Tests/ConfigurationTest.php
```diff
@@ -40,4 +44,133 @@ public function testEnableTransactionalFlush(): void
...
+            ],
+            // Key vault namespace from the configuration
+            'keyVaultNamespace' => 'keyvault.datakeys',
+        ], $c->getDriverOptions()['autoEncryption']);
+    }
+
+    public function testEmptyKmsProviderOptions(): void
+    {
+        $c = new Configuration();
+        $c->setKmsProvider(['type' => 'aws']);
```
jmikola, line 91: Noted that this exercises the logic that casts an empty array to an object, since `array_diff_keys()` is used to strip the `type` key before preparing the KMS provider object.

tests/Doctrine/ODM/MongoDB/Tests/Tools/EncryptedFieldsMapGeneratorTest.php
```diff
@@ -0,0 +1,261 @@
...
+                'bsonType' => 'int',
+                'keyId' => null,
+                'queries' => ['queryType' => 'range', 'min' => 100, 'max' => 2000, 'sparsity' => 1, 'trimFactor' => 4],
+            ],
+        ];
+
+        self::assertEquals(['fields' => $expected], $encryptedFieldsMap);
+    }
+
+    public function testGetEncryptionFieldsMapForClassForEmbeddedDocument(): void
```
jmikola:
```suggestion
    public function testGetEncryptionFieldsMapForClassWithEmbeddedDocument(): void
```

I think this name is more descriptive, as I originally expecting this test to query ClientCard directly.

tests/Doctrine/ODM/MongoDB/Tests/Tools/EncryptedFieldsMapGeneratorTest.php
```diff
@@ -0,0 +1,261 @@
...
+
+        $expected = [
+            [
+                'path' => 'name',
+                'bsonType' => 'string',
+                'keyId' => null,
+            ],
+            [
+                'path' => 'clientCards',
+                'bsonType' => 'array',
```
jmikola, line 76: I noted that the EmbedMany relation on Client isn't mapped as encrypted. The objects _within_ the array are encrypted. Would ODM produce the same EncryptedFields object if the relationship on Client was annotated with `#[Encrypt]`?

I'll note that it doesn't seem possible to map all array elements as encrypted (based on [server manual docs](https://www.mongodb.com/docs/manual/core/queryable-encryption/qe-create-encryption-schema/#std-label-qe-create-encryption-schema)). I suppose one could specify each array index as a separate path, but that wouldn't be flexible for dynamic arrays.

Therefore, I'm not suggesting the current behavior to encrypt the entire array is incorrect. But I am curious about whether users have two ways to define a mapping for the same effect.

lib/Doctrine/ODM/MongoDB/Configuration.php
GromNaN: Added a note in the cookbook.

lib/Doctrine/ODM/MongoDB/Utility/EncryptedFieldsMapGenerator.php
GromNaN: Indeed, I renamed the method.

tests/Doctrine/ODM/MongoDB/Tests/Tools/EncryptedFieldsMapGeneratorTest.php
GromNaN: The `clientCards` field is encrypted as a whole, into a single `Binary` value. 
image

QE doesn't allow querying objects, so having an array of documents would not be useful for querying. It it could for partial updates, like pushing a new element.

Meanwhile, we could have a list of int/long/double/date that could be queryable. But this results in this exception:
```
MongoDB\Driver\Exception\RuntimeException: mongocryptd error: Cannot encrypt element of type: array
```

jmikola: > QE doesn't allow querying objects, so having an array of documents would not be useful for querying. It it could for partial updates, like pushing a new element.

Would we expect this to conflict with ODM's collection strategies? I expect they could end up pushing or pulling elements from the array.

GromNaN: Yes, since arrays are stored as a single binary value, they need to be encrypted and saved globally client-side.
```
mongocryptd error: $addToSet not allowed on encrypted values
mongocryptd error: $push not allowed on encrypted values
```

I added the restriction to the documentation.

docs/en/cookbook/queryable-encryption.rst
GromNaN: I changed to: 
- Automatic generation of the ``encryptedFieldsMap`` is not compatible with
  ``SINGLE_COLLECTION`` inheritance. Because all classes in the hierarchy
  share a single collection, the ``SchemaManager`` cannot merge their encrypted
  fields before creating the collection.

lib/Doctrine/ODM/MongoDB/Utility/EncryptedFieldsMapGenerator.php
GromNaN: This is not possible, due to the way the `SchemaManager` works (class per class).
I removed this comment and will wait for an actual use-case before opening a ticket.

docs/en/cookbook/queryable-encryption.rst
jmikola: Very clear and concise explanation. Thank you!

lib/Doctrine/ODM/MongoDB/Utility/EncryptedFieldsMapGenerator.php
jmikola: Well, that was a fun little detour ⚠️ 

Consider this resolved but I'd like to leave the thread unresolved for posterity (and future accessibility).

docs/en/cookbook/queryable-encryption.rst
```diff
@@ -0,0 +1,271 @@
...
+  to work with multiple KMS providers, you must manually configure the
+  ``kmsProviders`` array and pass it as a driver option, bypassing the ODM's
+  helper method.
+- Automatic generation of the ``encryptedFieldsMap`` is not compatible with
+  ``SINGLE_COLLECTION`` inheritance. Because all classes in the hierarchy
+  share a single collection, the ``SchemaManager`` cannot merge their encrypted
+  fields before creating the collection.
+- Embedded documents and collections are encrypted as a whole. You cannot
+  be updated partially. Only the ``set*`` and ``atomicSet*`` collection
+  strategies can be used.
```
jmikola:
```suggestion
- Embedded documents and collections are encrypted as a whole. As such,
  they cannot be partially updated. Only the ``set*`` and ``atomicSet*``
  collection strategies can be used.
```

docs/en/cookbook/queryable-encryption.rst
```diff
@@ -0,0 +1,271 @@
...
+- For a complete list of hard limitations, please refer to the official
```
jmikola:
```suggestion
- For a complete list of limitations, please refer to the official
```

Unless the server docs also refer to "hard limitations", I think the adjective can be dropped.

lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -651,6 +707,107 @@ public function isLazyGhostObjectEnabled(): bool
...
+    }
+
+    /**
+     * Set the options for auto-encryption.
+     *
+     * @see https://www.php.net/manual/en/mongodb-driver-manager.construct.php#mongodb-driver-manager.construct-autoencryption
+     *
+     * @param array{ keyVaultClient?: Client|Manager, keyVaultNamespace?: string, tlsOptions?: array<string, mixed>, schemaMap?: array<string, mixed>, bypassAutoEncryption?: bool, bypassQueryAnalysis?: bool, encryptedFieldsMap?: array<string,mixed>, extraOptions?: array<string, mixed>} $options
+     */
+    public function setAutoEncryption(array $options): void
```
jmikola, line 749: Based on the logic in `getAutoEncryptionOptions()` below, which merged this in using the spread operator (`...`), keys here may overwrite the `kmsProviders` and `keyVaultNamespace` options created by ODM. On the off chance that could lead to unexpected behavior (or a hypothetical security issue down the line), it may be prudent to document that these options will take precedence.

jmikola (approved):
A few comments on docs but I think you can address those yourself.

lib/Doctrine/ODM/MongoDB/Configuration.php
GromNaN: I don't think there could be a problem:
- In `setAutoEncryption`, `kmsProvider` is not allowed to be set (exception). This option is not in the array shape of the param.
- In `getAutoEncryptionOptions`, `keyVaultNamespace` uses the default `db.datakeys` but can be overridden by the value set with `setAutoEncryption`.

PR #2780, opened 2025-06-12
[Encryption] Set master key when creating an encrypted collection

GromNaN:
The master key is required for all KMS exception "local" ([doc](https://www.mongodb.com/docs/php-library/current/reference/method/MongoDBDatabase-createEncryptedCollection/#parameters)).

The `masterKey` and `kmsProvider` configurations as set independently from the other `autoEncryption` options so that we can validate and transform them.
A single `kmsProvider` is allowed, and used as default when an encrypted collection is created.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Configuration.php
+++ b/lib/Doctrine/ODM/MongoDB/Configuration.php
@@ -58,7 +58,8 @@ use function trim;
  * @phpstan-import-type CommitOptions from UnitOfWork
  * @phpstan-type AutoEncryptionOptions array{
  *     keyVaultNamespace: string,
- *     kmsProviders: array<string, array<string, string>>,
+ *     kmsProviders: array<string, array<string, mixed>>,
+ *     kmsProvider?: string,
  *     masterKey?: array<string, mixed>|null,
  *     tlsOptions?: array{kmip: array{tlsCAFile: string, tlsCertificateKeyFile: string}},
  * }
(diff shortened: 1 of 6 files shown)


lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -699,12 +707,25 @@ public function isLazyGhostObjectEnabled(): bool
...
+
+        if (! isset($options['kmsProvider']) && count($options['kmsProviders']) > 1) {
+            throw new InvalidArgumentException('The "kmsProvider" encryption option is required when multiple KMS providers are specified.');
+        }
+
+        $options['kmsProvider'] ??= array_key_first($options['kmsProviders']);
+
+        if (! array_key_exists($options['kmsProvider'], $options['kmsProviders'])) {
+            throw new InvalidArgumentException(sprintf('The "kmsProvider" encryption option "%s" is not defined in the "kmsProviders" option.', $options['kmsProvider']));
         }
```
alcaeus: My suggestion is to only expose a `kmsProvider` option, which we then normalise into a single `kmsProviders` element when passing the provider list to the client. This single provider would only have a `type` (e.g. `local`), which we can use as the name for the provider wherever needed. This would make the configuration simpler and still allow us to add support for multiple KMS providers in future.

lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -699,12 +707,25 @@ public function isLazyGhostObjectEnabled(): bool
...
 
-        // @todo Throw en exception if multiple KMS providers are defined. This is not supported yet and would require a setting for the KMS provider to use when creating a new collection
-        if (! isset($options['kmsProviders']) || ! is_array($options['kmsProviders']) || count($options['kmsProviders']) < 1) {
-            throw new InvalidArgumentException('The "kmsProviders" option is required.');
+        if ($options['kmsProvider'] !== 'local' && ! isset($options['masterKey'])) {
```
alcaeus: Suggestion: rename the option `defaultMasterKey`, as we'll eventually allow setting a different master key in encryption config.

lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -133,7 +130,9 @@ class Configuration
      *      proxyDir?: string,
      *      proxyNamespace?: string,
      *      repositoryFactory?: RepositoryFactory,
-     *      autoEncryption?: AutoEncryptionOptions,
+     *      kmsProvider?: KmsProvider,
+     *      defaultMasterKey?: array<string, mixed>|null,
+     *      autoEncryption?: array<string, mixed>,
```
GromNaN, line 135: I extracted the `kmsProvider` and `defaultMasterKey` from the `autoEncryption` as they are not part of the driver options.

tests/Doctrine/ODM/MongoDB/Tests/ConfigurationTest.php
```diff
@@ -5,10 +5,13 @@
 namespace Doctrine\ODM\MongoDB\Tests;
 
 use Doctrine\ODM\MongoDB\Configuration;
+use Doctrine\ODM\MongoDB\ConfigurationException;
 use Doctrine\ODM\MongoDB\PersistentCollection\PersistentCollectionFactory;
 use Doctrine\ODM\MongoDB\PersistentCollection\PersistentCollectionGenerator;
+use MongoDB\Driver\Manager;
+use PHPUnit\Framework\TestCase;
 
-class ConfigurationTest extends BaseTestCase
```
GromNaN, line 11: The parent class was opening a connection to the mongodb server, which is not necessary for validating the configuration.

lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -163,13 +162,34 @@ public function getDriverOptions(): array
...
     }
 
+    /**
+     * Get options to create a ClientEncryption instance.
+     *
+     * @see https://www.php.net/manual/en/mongodb-driver-clientencryption.construct.php
+     *
+     * @return array{keyVaultClient?: Client|Manager, keyVaultNamespace: string, kmsProviders: array<string, mixed>, tlsOptions?: array<string, mixed>}
+     */
+    public function getClientEncryptionOptions(): array
```
GromNaN, line 179: The `Configuration` class is now responsible for creating the option arrays.

alcaeus (approved):

PR #2781, opened 2025-06-12
[Encryption] Drop metadata collection when dropping encrypted collection

GromNaN:
The option `encryptedFields` is required to trigger the deletion of metadata collections using the [`DropEncryptedCollection` operation](https://github.com/mongodb/mongo-php-library/blob/37440db813ad41e060a166a80088684d443b3403/src/Collection.php#L519-L521).

We don't support [custom `esc` and `ecoc` collection names](https://github.com/mongodb/mongo-php-library/blob/37440db813ad41e060a166a80088684d443b3403/src/Operation/DropEncryptedCollection.php#L76-L79). Therefore the array can be empty.

I chose to systematically try to delete metadata collections when autoencryption is enabled; not knowing if the collection was created encrypted and then the encrypted fields deleted from the metadata.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/SchemaManager.php
+++ b/lib/Doctrine/ODM/MongoDB/SchemaManager.php
@@ -699,6 +699,14 @@ final class SchemaManager
 
         $options = $this->getWriteOptions($maxTimeMs, $writeConcern);
 
+        // When automatic encryption is enabled, we need to drop the metadata collections,
+        // we don't check if the class metadata has encrypted fields, because
+        // that does not mean that the existing collection is encrypted or not.
+        // "esc" and "ecoc" collections cannot be configured
+        if ($this->dm->getConfiguration()->getKmsProvider()) {
+            $options['encryptedFields'] = [];
+        }
+
         $this->dm->getDocumentCollection($documentName)->drop($options);
 
         if (! $class->isFile) {
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1687,6 +1687,12 @@ parameters:
 			path: tests/Doctrine/ODM/MongoDB/Tests/Functional/QueryableEncryptionTest.php
 
 		-
+			message: '#^Parameter \#1 \$value of function count expects array\|Countable, Iterator\<int, string\> given\.$#'
+			identifier: argument.type
+			count: 1
+			path: tests/Doctrine/ODM/MongoDB/Tests/Functional/QueryableEncryptionTest.php
+
+		-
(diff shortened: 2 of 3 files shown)


lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -699,6 +699,14 @@ public function dropDocumentCollection(string $documentName, ?int $maxTimeMs = n
 
         $options = $this->getWriteOptions($maxTimeMs, $writeConcern);
 
+        // When automatic encryption is enabled, we need to drop the metadata collections,
+        // we don't check if the class metadata has encrypted fields, because
+        // that does not mean that the existing collection is encrypted or not.
+        // "esc" and "ecoc" collections cannot be configured
```
alcaeus, line 705: > We don't support [custom `esc` and `ecoc` collection names](https://github.com/mongodb/mongo-php-library/blob/37440db813ad41e060a166a80088684d443b3403/src/Operation/DropEncryptedCollection.php#L76-L79). Therefore the array can be empty.

I don't think we should make that assumption. As evidenced by the linked code, the driver respects a custom name, so we should too - with this piece of code we're however overriding the config if encrypted fields are present. When coming up with a solution, I think we should consider the following use cases:
* Initially setting up a production database, in which case no collections should be deleted
* Setting up a local development environment
* Dropping a collection in a local development environment in order to recreate it

Aside from those use cases, I wouldn't really recommend people use the schema tool in production, especially not to delete any obsolete documents (which would then no longer be present in the codebase). So, for all intents and purposes, I believe that we can get away with making a best-guess effort to removing metadata collections, first by attempting to use the `encryptedFields` still contained in the document, and using an empty `encryptedFields` option if there are no (more) encrypted fields configured.

alcaeus (changes requested):

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -699,6 +699,14 @@ public function dropDocumentCollection(string $documentName, ?int $maxTimeMs = n
...
+        if ($this->dm->getConfiguration()->getKmsProvider()) {
+            $options['encryptedFields'] = [];
```
GromNaN, line 707:
```suggestion
            $options['encryptedFields'] ??= [];
```

lib/Doctrine/ODM/MongoDB/SchemaManager.php
GromNaN: The custom metadata collection names are not supported **by doctrine ODM**.
There is no way to set the option when calling the schema manager or from the document class metadata (that could be added later). The  `escCollection` and `ecocCollection` options are not part of `encryptedFields`.

I'll leave the code as this.

alcaeus (approved):

PR #2782, opened 2025-06-13
[3.x] Require Composer v2

IonBazan:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | 

#### Summary

Replacing `jean85/pretty-package-versions` with `composer-runtime-api` v2 and slightly updating the version fetch logic - it will now return `error` when an exception or error occurs during fetching and `unknown` when version returns `null` for some reason.


GromNaN (approved):

GromNaN:
Thank you @IonBazan

PR #2783, opened 2025-06-13
[Encryption] Refactor encrypted fields map generator

alcaeus:
Refactors the encrypted field map generator and adds a method to generate field maps for all metadata. Note that testing this is difficult, as loading all metadata in tests causes errors due to some documents being invalid on purpose.


lib/Doctrine/ODM/MongoDB/Utility/EncryptedFieldsMapGenerator.php
```diff
@@ -49,7 +86,7 @@ private function createEncryptionFieldMap(ClassMetadata $classMetadata, string $
                 if ($embedMetadata->isEncrypted) {
                     $mapping['encrypt'] ??= []; // @todo get the keyId
                 } elseif (! isset($mapping['encrypt'])) {
-                    yield from $this->createEncryptionFieldMap(
+                    yield from $this->createEncryptedFieldsMapForClass(
                         $embedMetadata,
                         $path . $mapping['name'] . '.',
                     );
```
GromNaN: Instead of the class property `$generatorStack`, I would use a method parameter so that the scope is clear and you don't have to unset the class once it's handled.

```suggestion
                    yield from $this->createEncryptedFieldsMapForClass(
                        $embedMetadata,
                        $path . $mapping['name'] . '.',
                        $visitedClasses + [$classMetadata->getName() => true],
                    );
```

As an improvement: using an `int` level counter instead of `true`, we can set a limit in the recursion deep (if necessary).

lib/Doctrine/ODM/MongoDB/Utility/EncryptedFieldsMapGenerator.php
```diff
@@ -14,30 +14,67 @@
...
+            if (! $classMetadata->isDocument()) {
+                continue;
+            }
+
+            $classMap = iterator_to_array($this->createEncryptedFieldsMapForClass($classMetadata));
+            if ($classMap === []) {
+                continue;
+            }
+
+            $encryptedFieldsMap[$classMetadata->getName()] = $classMap;
```
GromNaN, line 43: Since this is a database configuration, it would make more sense to use the collection name as key.

```suggestion
            $encryptedFieldsMap[$classMetadata->getCollection()] = $classMap;
```

alcaeus: When dumping the encryptedFieldsMap, we need the full namespace. However, that namespace depends on the `db` property in metadata which may be empty, in which case it falls back to the default database from the configuration. I figured it was best to return the class name as that provides useful information both for printing the map and allows fetching the information necessary to assemble the full namespace.

GromNaN (approved):

lib/Doctrine/ODM/MongoDB/Utility/EncryptedFieldsMapGenerator.php
alcaeus: Changed the recursion prevention to a method parameter. Not sure how many recursion levels would be ok, but for the time being I'll prevent any recursion.

GromNaN (approved):

PR #2785, opened 2025-06-23
[Encryption] Fix BSON types and query option name

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | -

Part of https://github.com/doctrine/mongodb-odm/pull/2779

The Doctrine ODM type names doesn't match the BSON type names. We have to map them to create the `encryptedFieldsMap`.

Also fix the name of the property `precision` in QE field definition.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Utility/EncryptedFieldsMapGenerator.php
+++ b/lib/Doctrine/ODM/MongoDB/Utility/EncryptedFieldsMapGenerator.php
@@ -121,7 +121,7 @@ final class EncryptedFieldsMapGenerator
                     Type::OBJECTID => 'objectId',
                     Type::STRING => 'string',
                     Type::BINDATA, Type::BINDATABYTEARRAY, Type::BINDATAFUNC, Type::BINDATACUSTOM, Type::BINDATAUUID, Type::BINDATAMD5, Type::BINDATAUUIDRFC4122 => 'binData',
-                    // Type BOOL is not supported in encrypted fields map
+                    Type::BOOL, Type::BOOLEAN => 'bool',
                     default => throw new LogicException(sprintf('Type "%s" is not supported in encrypted fields map.', $mapping['type'])),
                 },
                 'keyId' => null, // Generate the key automatically
--- a/lib/Doctrine/ODM/MongoDB/Utility/EncryptedFieldsMapGenerator.php
+++ b/lib/Doctrine/ODM/MongoDB/Utility/EncryptedFieldsMapGenerator.php
@@ -111,13 +111,17 @@ final class EncryptedFieldsMapGenerator
             $field = [
                 'path' => $path . $mapping['name'],
                 'bsonType' => match ($mapping['type']) {
-                    ClassMetadata::ONE => 'object',
-                    ClassMetadata::MANY => 'array',
-                    Type::STRING => 'string',
+                    ClassMetadata::ONE, Type::HASH => 'object',
+                    ClassMetadata::MANY, Type::COLLECTION => 'array',
                     Type::INT, Type::INTEGER => 'int',
                     Type::FLOAT => 'double',
                     Type::DECIMAL128 => 'decimal',
                     Type::DATE, Type::DATE_IMMUTABLE => 'date',
+                    Type::TIMESTAMP => 'timestamp',
+                    Type::OBJECTID => 'objectId',
+                    Type::STRING => 'string',
+                    Type::BINDATA, Type::BINDATABYTEARRAY, Type::BINDATAFUNC, Type::BINDATACUSTOM, Type::BINDATAUUID, Type::BINDATAMD5, Type::BINDATAUUIDRFC4122 => 'binData',
+                    // Type BOOL is not supported in encrypted fields map
                     default => throw new LogicException(sprintf('Type "%s" is not supported in encrypted fields map.', $mapping['type'])),
                 },
                 'keyId' => null, // Generate the key automatically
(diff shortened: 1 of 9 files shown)


phpstan-baseline.neon
```diff
@@ -2005,43 +1999,43 @@ parameters:
 			path: tests/Doctrine/ODM/MongoDB/Tests/Tools/EncryptedFieldsMapGeneratorTest.php
 
 		-
-			message: '#^Method Doctrine\\Persistence\\Mapping\\AbstractClassMetadataFactory@anonymous/tests/Doctrine/ODM/MongoDB/Tests/Tools/EncryptedFieldsMapGeneratorTest\.php\:165\:\:__construct\(\) has parameter \$classNames with no value type specified in iterable type array\.$#'
+			message: '#^Method Doctrine\\Persistence\\Mapping\\AbstractClassMetadataFactory@anonymous/tests/Doctrine/ODM/MongoDB/Tests/Tools/EncryptedFieldsMapGeneratorTest\.php\:168\:\:__construct\(\) has parameter \$classNames with no value type specified in iterable type array\.$#'
```
GromNaN, line 2002: The anonymous class name contains the line number it's defined. Each time the previous test is updated, this baseline entries need to be updated.

lib/Doctrine/ODM/MongoDB/Utility/EncryptedFieldsMapGenerator.php
```diff
@@ -108,11 +111,15 @@ private function createEncryptedFieldsMapForClass(
...
-                    'one' => 'object',
-                    'many' => 'array',
-                    default => $mapping['type'],
+                    ClassMetadata::ONE => 'object',
+                    ClassMetadata::MANY => 'array',
+                    Type::STRING => 'string',
+                    Type::INT, Type::INTEGER => 'int',
+                    Type::FLOAT => 'double',
+                    Type::DECIMAL128 => 'decimal',
+                    Type::DATE, Type::DATE_IMMUTABLE => 'date',
```
GromNaN: I don't know if binary types are supported.

lib/Doctrine/ODM/MongoDB/Utility/EncryptedFieldsMapGenerator.php
```diff
@@ -108,11 +111,15 @@ private function createEncryptedFieldsMapForClass(
...
+                    default => throw new LogicException(sprintf('Type "%s" is not supported in encrypted fields map.', $mapping['type'])),
                 },
-                // @todo allow setting a keyId in #[Encrypt] attribute
```
GromNaN, line 115: `@todo` removed, we don't need to support custom `keyId` with queryable encryption auto-encryption. They are generated automatically.

lib/Doctrine/ODM/MongoDB/Utility/EncryptedFieldsMapGenerator.php
GromNaN: They are supported.

lib/Doctrine/ODM/MongoDB/Utility/EncryptedFieldsMapGenerator.php
```diff
@@ -108,11 +111,19 @@ private function createEncryptedFieldsMapForClass(
...
+                    ClassMetadata::ONE, Type::HASH => 'object',
+                    ClassMetadata::MANY, Type::COLLECTION => 'array',
+                    Type::INT, Type::INTEGER => 'int',
+                    Type::FLOAT => 'double',
+                    Type::DECIMAL128 => 'decimal',
+                    Type::DATE, Type::DATE_IMMUTABLE => 'date',
+                    Type::TIMESTAMP => 'timestamp',
+                    Type::OBJECTID => 'objectId',
+                    Type::STRING => 'string',
+                    Type::BINDATA, Type::BINDATABYTEARRAY, Type::BINDATAFUNC, Type::BINDATACUSTOM, Type::BINDATAUUID, Type::BINDATAMD5, Type::BINDATAUUIDRFC4122 => 'binData',
```
alcaeus, line 123: Would it make sense to add a `getBSONType` method to the `Type` class and override it in the individual types? Happy to defer this work until later, but I think it may be more efficient especially if people use custom types that map to a given BSON type.

lib/Doctrine/ODM/MongoDB/Utility/EncryptedFieldsMapGenerator.php
```diff
@@ -108,11 +111,19 @@ private function createEncryptedFieldsMapForClass(
...
+                    // Type BOOL is not supported in encrypted fields map
```
alcaeus: Is that limitation called out in the server docs somewhere? We should definitely call it out in our documentation so people can adapt their schema accordingly.

alcaeus (approved):
LGTM. We can create a separate ticket for the ODM type => BSON type mapping and handle that later, as it's not strictly necessary.

lib/Doctrine/ODM/MongoDB/Utility/EncryptedFieldsMapGenerator.php
GromNaN: I was thinking to add a `bsonType` parameter to the `#[Encrypt]` attribute, for custom needs. Adding `Type::getBSONType()` is a better option. [PHPORM-358](https://jira.mongodb.org/browse/PHPORM-358)

lib/Doctrine/ODM/MongoDB/Utility/EncryptedFieldsMapGenerator.php
GromNaN: I tested again, bool seems to be supported. I don't know why it was failing previously.

PR #2786, opened 2025-06-26
Add `Type::getBSONType()`

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | [PHPORM-358](https://jira.mongodb.org/browse/PHPORM-358)

#### Summary

While working on `encryptedFieldsMap` (https://github.com/doctrine/mongodb-odm/pull/2785#discussion_r2163749423), we found that it would be useful if every ODM Type could indicate its [BSON type](https://www.mongodb.com/docs/manual/reference/bson-types/).

It's possible to assign 1 BSON type for most of the ODM types, but there are some exception with some ODM types that can return multiple BSON types (see comments).

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Types/CustomIdType.php
+++ b/lib/Doctrine/ODM/MongoDB/Types/CustomIdType.php
@@ -4,11 +4,20 @@ declare(strict_types=1);
 
 namespace Doctrine\ODM\MongoDB\Types;
 
+use LogicException;
+
+use function sprintf;
+
 /**
  * The Id type.
  */
 class CustomIdType extends Type
 {
+    public function getBSONType(): BsonType
+    {
+        throw new LogicException(sprintf('Cannot determine BSON type for "%s".', self::class));
+    }
+
     public function convertToDatabaseValue($value)
     {
         return $value;
--- a/lib/Doctrine/ODM/MongoDB/Types/IntType.php
+++ b/lib/Doctrine/ODM/MongoDB/Types/IntType.php
@@ -11,6 +11,11 @@ use function max;
  */
 class IntType extends Type implements Incrementable, Versionable
 {
+    public function getBSONType(): BsonType
+    {
+        return BsonType::Int32;
+    }
+
     public function convertToDatabaseValue($value)
     {
         return $value !== null ? (int) $value : null;
--- a/lib/Doctrine/ODM/MongoDB/Types/KeyType.php
+++ b/lib/Doctrine/ODM/MongoDB/Types/KeyType.php
@@ -4,14 +4,22 @@ declare(strict_types=1);
 
 namespace Doctrine\ODM\MongoDB\Types;
 
+use LogicException;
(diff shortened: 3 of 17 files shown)


lib/Doctrine/ODM/MongoDB/Types/Type.php
```diff
@@ -91,6 +92,17 @@ final private function __construct()
...
+     * Returns the alias name of the BSON type.
+     *
+     * @link https://www.mongodb.com/docs/manual/reference/bson-types/
+     */
+    public function getBSONType(): BsonType
+    {
+        // This method will be abstract in the next major version.
+        throw new BadMethodCallException(sprintf('The method "%s::getBSONType" is not implemented. You must implement this method in the concrete type class.', static::class));
+    }
+
```
GromNaN, line 105: We cannot make this method abstract in a minor version, that would be a breaking change for custom type classes that extend this class.

lib/Doctrine/ODM/MongoDB/Types/RawType.php
```diff
@@ -4,11 +4,20 @@
...
+use function sprintf;
+
 /**
  * Raw data type.
  */
 class RawType extends Type
 {
+    public function getBSONType(): BsonType
+    {
+        throw new LogicException(sprintf('Cannot determine BSON type for "%s".', self::class));
```
GromNaN, line 18: The field type can be anything.

lib/Doctrine/ODM/MongoDB/Types/KeyType.php
```diff
@@ -4,14 +4,22 @@
...
+use function sprintf;
+
 /**
  * The Key type.
  */
 class KeyType extends Type
 {
+    public function getBSONType(): BsonType
+    {
+        throw new LogicException(sprintf('Cannot determine BSON type for "%s".', self::class));
```
GromNaN, line 20: The field type can be `maxKey` or `minKey`.

lib/Doctrine/ODM/MongoDB/Types/IntType.php
```diff
@@ -11,6 +11,11 @@
  */
 class IntType extends Type implements Incrementable, Versionable
 {
+    public function getBSONType(): BsonType
+    {
+        return BsonType::Int32;
```
GromNaN, line 16: Depending on the system and the value, I think the type can be an int 32 bits or 64 bits.

lib/Doctrine/ODM/MongoDB/Types/CustomIdType.php
```diff
@@ -4,11 +4,20 @@
...
+use function sprintf;
+
 /**
  * The Id type.
  */
 class CustomIdType extends Type
 {
+    public function getBSONType(): BsonType
+    {
+        throw new LogicException(sprintf('Cannot determine BSON type for "%s".', self::class));
```
GromNaN, line 18: The field type can be anything.

PR #2787, opened 2025-07-08
PHPORM-356: Skip search indexes not enabled exception

paulinevos:
If search indexes are not enabled on the server, we should simply skip any search index operation. This commit adds an error code to the search index error check already in place.

Closes PHPORM-356

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/SchemaManager.php
+++ b/lib/Doctrine/ODM/MongoDB/SchemaManager.php
@@ -47,6 +47,7 @@ final class SchemaManager
 
     private const CODE_SHARDING_ALREADY_INITIALIZED = 23;
     private const CODE_COMMAND_NOT_SUPPORTED        = 115;
+    private const CODE_SEARCH_NOT_ENABLED           = 31082;
 
     private const ALLOWED_MISSING_INDEX_OPTIONS = [
         'background',
@@ -1079,6 +1080,11 @@ final class SchemaManager
 
     private function isSearchIndexCommandException(CommandException $e): bool
     {
+        // MongoDB 8.0+: "Using Atlas Search Database Commands and the $listSearchIndexes aggregation stage requires additional configuration."
+        if ($e->getCode() === self::CODE_SEARCH_NOT_ENABLED) {
+            return true;
+        }
+
         // MongoDB 6.0.7+ and 7.0+: "Search indexes are only available on Atlas"
         if ($e->getCode() === self::CODE_COMMAND_NOT_SUPPORTED && str_contains($e->getMessage(), 'Search index')) {
             return true;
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -47,6 +47,7 @@ final class SchemaManager
 
     private const CODE_SHARDING_ALREADY_INITIALIZED = 23;
     private const CODE_COMMAND_NOT_SUPPORTED        = 115;
+    private const CODE_SEARCH_NOT_ENABLED           = 31082;
```
malarzm, line 50: It'd be great to have this exposed as a const from one of underlying libs. Or is it still not exposed?

malarzm (commented):
Please add a test like other exceptions have: https://github.com/doctrine/mongodb-odm/blob/6f4f0fbc51e918144839de0ca5377ed5dcec3b3f/tests/Doctrine/ODM/MongoDB/Tests/SchemaManagerTest.php#L432-L451

lib/Doctrine/ODM/MongoDB/SchemaManager.php
GromNaN: I agree, we have the same issue in the Laravel package: https://github.com/mongodb/laravel-mongodb/blob/a2b4ab86dfc9248050b2592d9830773ac335774e/src/Schema/Builder.php#L386-L395

But if we rely on a new function in the `mongodb/mongodb` library, we would have to increase the minimum version for low benefits.
I'm tracking this request: [PHPLIB-1689](https://jira.mongodb.org/browse/PHPLIB-1689)

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -1079,6 +1080,10 @@ private function getWriteOptions(?int $maxTimeMs = null, ?WriteConcern $writeCon
 
     private function isSearchIndexCommandException(CommandException $e): bool
     {
+        if ($e->getCode() === self::CODE_SEARCH_NOT_ENABLED) {
```
GromNaN, line 1084: We can add a comment as other cases.
```suggestion
        // MongoDB 8.0+: "Using Atlas Search Database Commands and the $listSearchIndexes aggregation stage requires additional configuration."
        if ($e->getCode() === self::CODE_SEARCH_NOT_ENABLED) {
```

GromNaN (approved):
LGTM once the comment is added.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
paulinevos: I was thinking that, but I wasn't sure if it was version related or simply overlooked. Wasn't sure which repo to check to compare, maybe you can enlighten me?

GromNaN: I tested all minor versions and community/enterprise/atlas when I wrote the code for Laravel. Trust me 😏

PR #2790, opened 2025-08-04
[Encryption] PHPORM-360 Document limitations of encryption with collection inheritance

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | doc
| BC Break     | no
| Fixed issues | [PHPORM-360](https://jira.mongodb.org/browse/PHPORM-360)

#### Summary

Improve documentation on queryable encryption.
Add a section about schema manager to create the encrypted collection.

### DIFF
--- a/docs/en/reference/attributes-reference.rst
+++ b/docs/en/reference/attributes-reference.rst
@@ -356,6 +356,10 @@ Optional arguments:
   users only. The default values for these options are suitable for the majority
   of use cases, and should only be modified if your use case requires it.
 
+.. note::
+
+    Queryable encryption is only supported in MongoDB version 8.0 and later.
+
 Example:
 
 .. code-block:: php
@@ -373,8 +377,49 @@ Example:
         public string $name;
     }
 
+The ``#[Encrypt]`` attribute is can be added to an class with `#[EmbeddedDocument]`_.
+This will encrypt the entire embedded document, in the field that contains it.
+Queryable encryption is not supported for embedded documents, so the ``queryType``
+argument is not applicable in this case. In this case, the embedded document is
+stored as a binary value in the parent document.
+
+.. code-block:: php
+
+    <?php
+
+    use Doctrine\ODM\MongoDB\Mapping\Annotations\Encrypt;
+
+    #[Encrypt]
+    #[EmbeddedDocument]
+    class CreditCard
+    {
+        #[Field]
+        public string $number;
+
+        #[Field]
+        public string $expiryDate;
+    }
+
+    #[Document]
+    class User
+    {
+        #[EmbedOne(targetDocument: CreditCard::class)]
+        public CreditCard $creditCard;
+    }
+
 For more details, refer to the MongoDB documentation on
 `Queryable Encryption <https://www.mongodb.com/docs/manual/core/queryable-encryption/fundamentals/encrypt-and-query/>`_.
+
+
+.. note::
+
+    The encrypted collection must be created with the `Schema Manager`_ before
+    before inserting documents.
+
+.. note::
+
+    Due to the way the encrypted fields map is generated, the queryable encryption
+    is not compatible with ``SINGLE_COLLECTION`` inheritance.
 
 #[Field]
 --------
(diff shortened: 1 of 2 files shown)


docs/en/reference/attributes-reference.rst
```diff
@@ -373,9 +377,50 @@ Example:
         public string $name;
     }
 
+The ``#[Encrypt]`` attribute is can be added to an class with `#[EmbeddedDocument]`_.
```
paulinevos:
```suggestion
The ``#[Encrypt]`` attribute can be added to a class with `#[EmbeddedDocument]`_.
```

docs/en/reference/attributes-reference.rst
```diff
@@ -373,9 +377,50 @@ Example:
...
+This will encrypt the entire embedded document, in the field that contains it.
+Queryable encryption is not supported for embedded documents, so the ``queryType``
+argument is not applicable in this case. In this case, the embedded document is
```
paulinevos:
```suggestion
argument is not applicable. Encrypted embedded documents are 
```

This phrasing feels clearer, as "in this case" implies a different way of using the encryption feature (to me at least).

docs/en/reference/migrating-schemas.rst
```diff
@@ -15,6 +15,45 @@ problem!
     for the Google App Engine datastore. Additional information may be found in
     the `Objectify schema migration`_ documentation.
 
+Creating a collection
+--------------------
+
+MongoDB does not require you to create collections before using them, but
+this operation is recommended for setting up indexes or validation rules. It's
+also required to create encrypted collections, which are not supported by
+Doctrine MongoDB ODM.
```
paulinevos: Could you try to rephrase this? It's not clear to me what this means.

docs/en/reference/migrating-schemas.rst
```diff
@@ -15,6 +15,45 @@ problem!
...
+
+.. code-block:: php
+
+    <?php
+
+    // Assuming $dm is your DocumentManager instance
+    $schemaManager = $dm->getSchemaManager();
+
+To create the collections for all the documents classes, you can use the
```
paulinevos:
```suggestion
To create the collections for all the document classes, you can use the
```

paulinevos (changes requested):

docs/en/reference/migrating-schemas.rst
GromNaN: What do you think of the new wording?

> MongoDB will create collections automatically upon first insertion, there are
situation you must create them explicitly. This allows you to configure specific
options, such as indexes and validation rules. Explicit creation is also required
features like encrypted collections.

paulinevos: Ahhh okay, I get it now. How about this:

> Explicitly creating collections is not required in MongoDB, unless they require options such as indexes or validation rules. Encrypted collections must also be created explicitly.

The only thing still throwing me off is the mention of the ODM. Is that relevant to the explicit creation portion? It feels like a separate thing. Now it reads as if encrypted collections have to be explicitly created _because_ they're not supported by the ODM.

GromNaN: This sentence was an error.
> which are not supported by Doctrine MongoDB ODM.

I'm not a big fan of the passive form. "You" is regularly used in this docs.

I reworded again: 
> Collections automatically created by the MongoDB server upon first insertion.
You must explicitly create the collections if you need specific options, such as
validation rules. In particular, encrypted collections must be created explicitly.

alcaeus (commented):
Changes LGTM, deferring to @paulinevos for a final review.

docs/en/reference/migrating-schemas.rst
```diff
@@ -15,6 +15,44 @@ problem!
     for the Google App Engine datastore. Additional information may be found in
     the `Objectify schema migration`_ documentation.
 
+Creating a collection
+--------------------
+
+Collections automatically created by the MongoDB server upon first insertion.
```
paulinevos:
```suggestion
Collections are automatically created by the MongoDB server upon first insertion.
```

paulinevos: Wording looks good, agree with keeping the usage of "you" consistent (reads better too imo). Only forgot a word in there but other than that looks good to me

paulinevos (approved):

Issue #2762, opened 2025-04-29
[RFC] Plan for MongoDB ODM v3


GromNaN:
MongoDB ODM v2 was released in 2018 and PHP and MongoDB have come a long way since then.

I'd like to propose a plan for things that involve breaking changes and that can be achieved for a 3.0 release. The goal is to make the version 2.last version 100% compatible with 3.0, as soon as all deprecations are resolved (following the same process as Symfony of deprecating before removing).

- [x] Drop lazy proxy libraries and use native lazy objects
  - [x] Add support for native lazy objects in v2.x
  - [x] Require PHP 8.4+ #2895
- [x] Rename `lib/Doctrine/ODM/MongoDB/` to `src/` https://github.com/doctrine/mongodb-odm/pull/2866
- [x] Drop support for annotations. They are replaced by PHP Attributes #2931
- [x] Add native types to all properties
- [x] Add native return types
- [x] Add native parameter types (this is backwards compatible, should be already the case)
- [x] Make soft-final classes, hard-final
- [x] Require ext-mongodb 2 #2897
- [x] Remove support for `CollectionInfoIterator`, `CollectionInfoCommandIterator`, `IndexInfoIterator`, `IndexInfoIteratorIterator` from ext-mongodb 1.x
- [ ] Remove inheritance strategies `COLLECTION_PER_CLASS` https://github.com/doctrine/mongodb-odm/issues/2742
- [x] Make classMetadata fields `private(set)`
- [x] https://github.com/doctrine/mongodb-odm/issues/2448
- [x] Require `composer-runtime-api: ^2.0` #2782 (see https://github.com/mongodb/mongo-php-library/pull/1262)
- [x] Use `#[AsCommand]` attribute to enable lazy-loading #2934
- [x] Drop `doctrine/cache` dependency

aturki:
A very nice addition to the v3 would be the result cache 🙂

greg0ire:
Item 2 (renaming to `src`) should IMO be done in 2.x, as it is no breaking change and will make merging up easier.

alcaeus:
> A very nice addition to the v3 would be the result cache 🙂

You would typically design your schema in a way that avoids long-running queries, which is the entire point of a non-relational document database. That said, I'd be interested to hear your use case for a result cache.

mickverm:
I noticed that this library still requires the no longer maintained and abandoned `doctrine/cache` dependency, is this something that will also be tackled in version 3?

[...]

PR #2791, opened 2025-08-05
Symfony 8.0 support

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | [PHPORM-375](https://jira.mongodb.org/browse/PHPORM-375)

#### Summary

Dependencies blocking Symfony upgrade:

- [x] `doctrine/orm` https://github.com/doctrine/orm/pull/12102
- [x] `friendsofphp/proxy-manager-lts`
- [x] `phpbench/phpbench`

Support for Native Lazy objects is required, as [Symfony 8.0 drops support for lazy ghost objects](https://github.com/symfony/symfony/blob/8.0/UPGRADE-8.0.md#varexporter). This version will never be compatible with `symfony/var-exporter:8`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,10 +35,10 @@
         "jean85/pretty-package-versions": "^1.3.0 || ^2.0.1",
         "mongodb/mongodb": "^1.21.2 || ^2.1.1",
         "psr/cache": "^1.0 || ^2.0 || ^3.0",
-        "symfony/console": "^5.4 || ^6.0 || ^7.0",
+        "symfony/console": "^5.4 || ^6.4 || ^7.0",
         "symfony/deprecation-contracts": "^2.2 || ^3.0",
-        "symfony/var-dumper": "^5.4 || ^6.0 || ^7.0",
-        "symfony/var-exporter": "^6.2 || ^7.0"
+        "symfony/var-dumper": "^5.4 || ^6.4 || ^7.0",
+        "symfony/var-exporter": "^6.4 || ^7.0"
     },
     "require-dev": {
         "ext-bcmath": "*",
@@ -52,8 +52,8 @@
         "phpstan/phpstan-phpunit": "^2.0",
         "phpunit/phpunit": "^10.5.58",
         "squizlabs/php_codesniffer": "^4",
-        "symfony/cache": "^5.4 || ^6.0 || ^7.0",
-        "symfony/uid": "^5.4 || ^6.0 || ^7.0"
+        "symfony/cache": "^5.4 || ^6.0 || ^7.0 || ^8.0",
+        "symfony/uid": "^5.4 || ^6.0 || ^7.0 || ^8.0"
     },
     "conflict": {
         "doctrine/annotations": "<1.12 || >=3.0"
--- a/composer.json
+++ b/composer.json
@@ -11,6 +11,8 @@
         "mapping",
         "object"
     ],
+    "minimum-stability": "dev",
+    "prefer-stable": true,
     "homepage": "https://www.doctrine-project.org/projects/mongodb-odm.html",
     "license": "MIT",
     "authors": [
@@ -35,9 +37,9 @@
         "jean85/pretty-package-versions": "^1.3.0 || ^2.0.1",
         "mongodb/mongodb": "^1.21.2 || ^2.1.1",
         "psr/cache": "^1.0 || ^2.0 || ^3.0",
-        "symfony/console": "^5.4 || ^6.4 || ^7.0",
+        "symfony/console": "^5.4 || ^6.4 || ^7.0 || ^8.0",
         "symfony/deprecation-contracts": "^2.2 || ^3.0",
-        "symfony/var-dumper": "^5.4 || ^6.4 || ^7.0",
+        "symfony/var-dumper": "^5.4 || ^6.4 || ^7.0 || ^8.0",
         "symfony/var-exporter": "^6.4 || ^7.0"
     },
     "require-dev": {
--- a/src/Tools/Console/Command/CommandCompatibility.php
+++ b/src/Tools/Console/Command/CommandCompatibility.php
@@ -9,10 +9,32 @@ use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 
-if ((new ReflectionMethod(Command::class, 'execute'))->hasReturnType()) {
+// Symfony 8
+if ((new ReflectionMethod(Command::class, 'configure'))->hasReturnType()) {
     /** @internal */
     trait CommandCompatibility
     {
+        protected function configure(): void
+        {
+            $this->doConfigure();
+        }
+
+        protected function execute(InputInterface $input, OutputInterface $output): int
+        {
+            return $this->doExecute($input, $output);
+        }
+    }
+// Symfony 7
+} elseif ((new ReflectionMethod(Command::class, 'execute'))->hasReturnType()) {
+    /** @internal */
+    trait CommandCompatibility
+    {
+        /** @return void */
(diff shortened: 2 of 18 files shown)


lib/Doctrine/ODM/MongoDB/Tools/Console/Command/Schema/AbstractCommand.php
```diff
@@ -24,8 +24,7 @@ abstract class AbstractCommand extends Command
     public const INDEX        = 'index';
     public const SEARCH_INDEX = 'search-index';
 
-    /** @return void */
-    protected function configure()
+    protected function configure(): void
```
GromNaN: Technically a breaking change.

derrabus: In DBAL, we solved this with a conditional return type on the lower branches.

https://github.com/doctrine/dbal/blob/1332b9ee51d3f2a685f007806c4c21991b7ec3f9/src/Tools/Console/Command/CommandCompatibility.php

GromNaN: Thanks for the pointer, the ODM also has this trait. I created a new one for the `AbstractCommand` to use a different name for the private method, as there are 2 levels of inheritance.

composer.json
```diff
@@ -32,10 +32,10 @@
         "jean85/pretty-package-versions": "^1.3.0 || ^2.0.1",
         "mongodb/mongodb": "^1.21 || ^2.0@dev",
         "psr/cache": "^1.0 || ^2.0 || ^3.0",
-        "symfony/console": "^5.4 || ^6.0 || ^7.0",
+        "symfony/console": "^5.4 || ^6.4 || ^7.0 || ^8.0",
         "symfony/deprecation-contracts": "^2.2 || ^3.0",
-        "symfony/var-dumper": "^5.4 || ^6.0 || ^7.0",
-        "symfony/var-exporter": "^6.2 || ^7.0"
+        "symfony/var-dumper": "^5.4 || ^6.4 || ^7.0 || ^8.0",
+        "symfony/var-exporter": "^6.4 || ^7.0"
```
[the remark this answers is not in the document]
GromNaN: Where should it be documented?

src/Tools/Console/Command/Schema/AbstractCommand.php
```diff
@@ -19,16 +19,15 @@
...
     public const DB           = 'db';
     public const COLLECTION   = 'collection';
     public const INDEX        = 'index';
     public const SEARCH_INDEX = 'search-index';
 
-    /** @return void */
-    protected function configure()
+    private function configureCommonOptions(): void
     {
-        parent::configure();
```
jmikola, line 30: You remove the call to `parent::configure()` here but it's not included in any of the new implementations in AbstractCommandCompatibility. Does that mean this parent call was never required?

jmikola (approved):
One question, but LGTM if no changes are required.

src/Tools/Console/Command/CommandCompatibility.php
```diff
@@ -9,10 +9,32 @@
...
-if ((new ReflectionMethod(Command::class, 'execute'))->hasReturnType()) {
+// Symfony 8
+if ((new ReflectionMethod(Command::class, 'configure'))->hasReturnType()) {
     /** @internal */
     trait CommandCompatibility
     {
+        protected function configure(): void
+        {
+            $this->doConfigure();
+        }
```
IonBazan, line 20: Perhaps I'm missing something but I don't think we need this trait here.
Since we already require PHP 8.1+, we should be able to use `void` in child classes as per: https://3v4l.org/Mv3oq

I did something similar in https://github.com/IonBazan/composer-diff/blob/8907711f0bf74677041f7b356da2367afea44a53/src/Command/BaseTypedCommand.php#L9-L14 but that's only needed to support different PHP versions.

> If the parent does not declare a return type then the child is allowed to declare one.

See: https://wiki.php.net/rfc/return_types

IonBazan (approved):

src/Tools/Console/Command/CommandCompatibility.php
IonBazan: We can remove it in 2.13.x just to be sure

GromNaN: The issue is with older versions of the bundle that don't have the `void` return type. https://github.com/doctrine/DoctrineMongoDBBundle/blob/4.7.x/Command/CreateSchemaDoctrineODMCommand.php#L21

src/Tools/Console/Command/Schema/AbstractCommand.php
GromNaN: The parent method is empty.

SenseException (approved):

composer.json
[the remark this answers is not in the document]
jmikola: Release notes perhaps?

src/Tools/Console/Command/Schema/AbstractCommand.php
jmikola: Is that something we can rely on in perpetuity? I'm just wondering if that changes down the line, whether we'd catch it and remember to add this bac -- as admittedly unlikely as that is.

GromNaN: > Is that something we can rely on in perpetuity?

Yes!

It [is documented](https://symfony.com/doc/6.4/console.html#console-input) without calling `parent::configure()`. The purpose of this method is to be overridden. It could have been abstract but we don't want to force this method to be declared in child classes when no input definition is necessary.

composer.json
```diff
@@ -35,10 +35,10 @@
         "jean85/pretty-package-versions": "^1.3.0 || ^2.0.1",
         "mongodb/mongodb": "^1.21.2 || ^2.1.1",
         "psr/cache": "^1.0 || ^2.0 || ^3.0",
-        "symfony/console": "^5.4 || ^6.0 || ^7.0",
+        "symfony/console": "^5.4 || ^6.4 || ^7.0",
         "symfony/deprecation-contracts": "^2.2 || ^3.0",
-        "symfony/var-dumper": "^5.4 || ^6.0 || ^7.0",
-        "symfony/var-exporter": "^6.2 || ^7.0"
+        "symfony/var-dumper": "^5.4 || ^6.4 || ^7.0",
+        "symfony/var-exporter": "^6.4 || ^7.0"
```
GromNaN, line 43: @doctrine/mongodb-odm-maintainers I think we need a meta package for the proxy implementation.
Currently, we have 3 lazy object implementations:
- `friendsofphp/proxy-manager-lts` (conflict with Symfony 8)
- `symfony/var-exporter` (not available in Symfony 8)
- Native (require PHP 8.4)

When using `symfony/flex` with `SYMFONY_REQUIRE=^8`, the installation fails because `symfony/var-exporter` and `friendsofphp/proxy-manager-lts` cannot be installed.

I propose to create a meta package with 2 versions:
- v1 require `symfony/var-exporter` and `friendsofphp/proxy-manager-lts`
- v2 require PHP 8.4+

So that, depending on the PHP and Symfony version, the correct dependencies are installed.

GromNaN: Alternatively, Symfony 8 would be only supported by version 3. But we have a lot to do: https://github.com/doctrine/mongodb-odm/issues/2762

GromNaN: - https://github.com/FriendsOfPHP/proxy-manager-lts/pull/43

GromNaN: Since `symfony/var-exporter` is not listed in `symfony/symfony: >=8`, Flex doesn't apply the constraint to it.

PR #2793, opened 2025-08-06
[Encryption] Add cookbook for queryable encryption

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | doc
| BC Break     | no
| Fixed issues | [PHPORM-360](https://jira.mongodb.org/browse/PHPORM-360) [PHPORM-345](https://jira.mongodb.org/browse/PHPORM-345)

### DIFF
--- a/docs/en/cookbook/queryable-encryption.rst
+++ b/docs/en/cookbook/queryable-encryption.rst
@@ -133,8 +133,13 @@ use a non-local key management service (KMS).
 
     $config = new Configuration();
     // Enable auto encryption and set the KMS provider.
-    $config->setAutoEncryption(true);
-    $config->setKmsProvider('local', ['key' => $masterKey]);
+    $config->setAutoEncryption([
+        'keyVaultNamespace' => 'encryption.datakeys'
+    ]);
+    $config->setKmsProvider([
+        'type' => 'local',
+        'key' => new Binary($masterKey),
+    ]);
 
     // Other configuration
     $config->setProxyDir(__DIR__ . '/Proxies');
--- a/docs/en/cookbook/queryable-encryption.rst
+++ b/docs/en/cookbook/queryable-encryption.rst
@@ -1,0 +1,261 @@
+Queryable Encryption
+====================
+
+This cookbook provides a tutorial on setting up and using Queryable Encryption
+(QE) with Doctrine MongoDB ODM to protect sensitive data in your documents.
+
+Introduction
+------------
+
+In many applications, you need to store sensitive information like social
+security numbers, financial data, or personal details. MongoDB's Queryable
+Encryption allows you to encrypt this data on the client-side, store it as
+fully randomized encrypted data, and still run expressive queries on it. This
+ensures that sensitive data is never exposed in an unencrypted state on the
+server, in system logs, or in backups.
+
+This tutorial will guide you through the process of securing a document's
+fields using queryable encryption, from defining the document and configuring
+the connection to storing and querying the encrypted data.
+
+.. note::
+
+    Queryable Encryption is only available on MongoDB Enterprise 8.0+ or
+    MongoDB Atlas.
+
+The Scenario
+------------
+
+We will model a ``Patient`` document that has an embedded ``PatientRecord``.
+This record contains sensitive information:
+
+- A Social Security Number (``ssn``), which we need to query for exact
+  matches.
+- A ``billingAmount``, which should support range queries.
+- A ``billing`` object, which should be encrypted but not directly queryable.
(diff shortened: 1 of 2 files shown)


docs/en/cookbook/queryable-encryption.rst
```diff
@@ -0,0 +1,261 @@
...
+         * By not specifying a queryType, we make it non-queryable.
+         */
+        #[ODM\EmbedOne(targetDocument: Billing::class)]
+        #[Encrypt]
+        public Billing $billing;
+
+        /**
+         * Encrypted with range queries.
+         * This allows us to query for billing amounts within a certain range.
+         * min, max, and sparsity are required for range queries.
```
jmikola: Contrast with [EncryptOpts](https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.md#encryptopts) in the spec and [ClientEncryption::encrypt()](https://www.php.net/manual/en/mongodb-driver-clientencryption.encrypt.php) in the extension docs.

Some options are required for floating point values, but I don't think there are any requirements for integers. And `sparsity` is always optional.

docs/en/cookbook/queryable-encryption.rst
```diff
@@ -0,0 +1,261 @@
...
+-----------------------
+
+The following example demonstrates how to configure the ``DocumentManager`` for
+encryption and how to work with encrypted documents.
+
+Step 1: Configure the DocumentManager
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+First, we configure the ``DocumentManager`` with ``autoEncryption`` options.
+For more details on the available options, see the `MongoDB\Driver\Manager`_
```
jmikola, line 109: Do the backslashes here require escaping? They disappear entirely in the GitHub preview of this RST file, but I'm not sure how that differs from the ODM docs tooling.

docs/en/cookbook/queryable-encryption.rst
```diff
@@ -0,0 +1,261 @@
...
+documentation. We'll use the ``local`` KMS provider for simplicity. For this
+provider, you need a 96-byte master key.
+The following code will look for the key in a local file (``master-key.bin``)
+and generate it if it doesn't exist. In a production environment, you should
+use an other key management service (KMS).
```
jmikola:
```suggestion
use another key management service (KMS).
```

Alternatively, "use a non-local".

docs/en/cookbook/queryable-encryption.rst
```diff
@@ -0,0 +1,261 @@
...
+    }
+
+Limitations
+-----------
+
+- The ODM simplifies configuration by supporting a single KMS provider per
+  ``DocumentManager`` through ``Configuration::setKmsProvider()``. If you need
+  to work with multiple KMS providers, you must manually configure the
+  ``kmsProviders`` array and pass it as a driver option, bypassing the ODM's
+  helper method.
```
jmikola, line 252: Was there a particular reason you opened for a singular `setKmsProvider()` method instead of allowing multiple calls and building up a `autoEncryption.kmsProviders` driver option?

docs/en/cookbook/queryable-encryption.rst
GromNaN: I've just tested and these parameters are not required, even for `float`.

docs/en/cookbook/queryable-encryption.rst
GromNaN: You found the only 2% of this document that I wrote by myself 🤦🏻

docs/en/cookbook/queryable-encryption.rst
GromNaN: They are not escaped in other parts of the doc:
https://github.com/doctrine/mongodb-odm/blob/dcfc75b74e3a8b720c5c90fa89d0b5a711834d1a/docs/en/reference/attributes-reference.rst?plain=1#L494

They are correctly rendered: https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/2.11/reference/attributes-reference.html#id

docs/en/cookbook/queryable-encryption.rst
GromNaN: This was a design decision to simplify the configuration. But this can be modified later.

docs/en/cookbook/queryable-encryption.rst
jmikola: That's why I'm here 😎

jmikola (approved):

PR #2795, opened 2025-08-08
[Encryption] Support encrypted mappings of embedded documents in XML mappings

alcaeus:
This fixes a bug exposed by Copilot review. It complained that a field was mapped as `string` when it was mapped as an embedded document in the attribute mapping. After changing it to an embedded document mapping, I noticed there was no way to mark an embedded document as encrypted through the XML mapping, where it's possible to do so using attribute.

This PR updates the schema to support an `encrypt` element within `embed-one` and `embed-many`. A couple of notes:
- `encrypt` retains its attributes, even though none of them make sense for embedded documents at the moment. Future versions of QE may change this
- There's an `encrypt-embedded-document` type in the XSD, which is never used. Was that intended for that purpose?
- Should we opt for a simpler `encrypt` attribute to `embed-one` and `embed-many` and only add support for an `encrypt` element once QE supports queries on encrypted embedded documents?

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Doctrine\ODM\MongoDB\Mapping\Driver;
 
+use Doctrine\ODM\MongoDB\Mapping\Annotations\EncryptQuery;
 use Doctrine\ODM\MongoDB\Mapping\Annotations\TimeSeries;
 use Doctrine\ODM\MongoDB\Mapping\ClassMetadata;
 use Doctrine\ODM\MongoDB\Mapping\MappingException;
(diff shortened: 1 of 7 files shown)


GromNaN:
We need to add an integration test that ensures the embedded document is actually encrypted.

lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
```diff
@@ -934,6 +932,20 @@ private function addGridFSMappings(ClassMetadata $class, SimpleXMLElement $xmlRo
         $xmlRoot->metadata->addAttribute('field', 'metadata');
         $this->addEmbedMapping($class, $xmlRoot->metadata, ClassMetadata::ONE);
     }
+
+    private function addEncryptionMapping(?SimpleXMLElement $encrypt, $type): array
+    {
+        $encryptMapping = [];
+        foreach ($encrypt->attributes() as $encryptKey => $encryptValue) {
+            $encryptMapping[$encryptKey] = match ($encryptKey) {
+                'queryType' => EncryptQuery::from((string) $encryptValue),
```
alcaeus, line 941: In the original, this was never converted to an `EncryptQuery` enum value, so the tests initially failed. Bug squashed :)

tests/Doctrine/ODM/MongoDB/Tests/Mapping/Driver/fixtures/xml/Documents.Encryption.PatientRecord.dcm.xml
```diff
@@ -9,7 +9,7 @@
         <field name="ssn" type="string">
             <encrypt queryType="equality"/>
         </field>
-        <embed-one target-document="Documents\Encryption\Billing" field="billing">
+        <embed-one target-document="Documents\Encryption\PatientBilling" field="billing">
```
alcaeus, line 12: Good thing we added testing

GromNaN (approved):
Thank you!

PR #2797, opened 2025-08-08
Drop metadata collection when an encrypted collection is dropped but encryptedFieldsMap is not configured

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | -

#### Summary

Patch for https://jira.mongodb.org/browse/PHPLIB-1702


GromNaN:
Fixed in the library: https://github.com/mongodb/mongo-php-library/pull/1742

PR #2799, opened 2025-08-11
Update maintained versions file

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | minor
| BC Break     | no
| Fixed issues | -

- 2.12 is upcoming
- 2.11 is the current stable
- 2.10 is unmaintained

### DIFF
--- a/.doctrine-project.json
+++ b/.doctrine-project.json
@@ -6,10 +6,28 @@
     "docsSlug": "doctrine-mongodb-odm",
     "versions": [
         {
-            "name": "2.10",
-            "branchName": "2.10.x",
+            "name": "2.12",
+            "branchName": "2.12.x",
             "slug": "latest",
             "upcoming": true,
+            "aliases": [
+                "2.12.x"
+            ]
+        },
+        {
+            "name": "2.11",
+            "branchName": "2.11.x",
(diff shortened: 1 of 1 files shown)


.doctrine-project.json
```diff
@@ -5,11 +5,29 @@
...
+            ]
+        },
+        {
+            "name": "2.11",
+            "branchName": "2.11.x",
+            "slug": "latest",
+            "maintained": true,
+            "aliases": [
+                "2.11.x"
+            ]
```
greg0ire: I think you should remove the aliases `current` and `stable` from 2.9, and use `current: true` on 2.11 (it will have the same effect). Also, I'm wondering what all these `.x` aliases achieve exactly.

greg0ire: Also, I don't understand why 2.10.2 is currently marked as stable here: https://www.doctrine-project.org/projects/mongodb-odm.html

GromNaN: Updated.

GromNaN: I removed all the aliases, taking example from https://github.com/doctrine/dbal/blob/4.3.x/.doctrine-project.json

greg0ire (approved):

PR #2801, opened 2025-08-12
[Encryption] Cast KMS provider to object to support AWS empty config

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | [DoctrineMongoDBBundle#897](https://github.com/doctrine/DoctrineMongoDBBundle/pull/897/#discussion_r2270005305)

#### Summary

For some KMS providers, all options can be omitted. The authentication is done using the env var or the system.
https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.md#credentialproviders

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Configuration.php
+++ b/lib/Doctrine/ODM/MongoDB/Configuration.php
@@ -793,7 +793,8 @@ class Configuration
     private function getAutoEncryptionOptions(): array
     {
         return [
-            'kmsProviders' => [$this->attributes['kmsProvider']['type'] => array_diff_key($this->attributes['kmsProvider'], ['type' => 0])],
+            // Each kmsProvider must be an object, it can be empty
+            'kmsProviders' => [$this->attributes['kmsProvider']['type'] => (object) array_diff_key($this->attributes['kmsProvider'], ['type' => 0])],
             'keyVaultNamespace' => $this->getDefaultDB() . '.datakeys',
             ...$this->attributes['autoEncryption'] ?? [],
         ];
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -793,7 +793,8 @@ private static function getVersion(): string
     private function getAutoEncryptionOptions(): array
     {
         return [
-            'kmsProviders' => [$this->attributes['kmsProvider']['type'] => array_diff_key($this->attributes['kmsProvider'], ['type' => 0])],
+            // Each kmsProvider must be an object, it can be empty
+            'kmsProviders' => [$this->attributes['kmsProvider']['type'] => (object) array_diff_key($this->attributes['kmsProvider'], ['type' => 0])],
```
jmikola, line 797: Slight correction to the PR description, since you said only mentioned AWS there.

[credentialProviders](https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.md#credentialproviders) talks about only supporting AWS, but [Automatic Credentials](https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.md#automatic-credentials) mentions that AWS, Azure, and GCP are all supported. The struct definition in [kmsProviders](https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.md#kmsproviders) also demonstrates that those three provider types might be expressed as an empty object.

I'll also note that libmongocrypt supports named KMS providers (e.g. `aws:foo`), which in turn do _not_ support fetching automatic credentials from the environment. 

I don't think this changes the code you have here, but I wanted to clarify.

jmikola (approved):

lib/Doctrine/ODM/MongoDB/Configuration.php
GromNaN: Thanks, I updated the PR description.

PR #2805, opened 2025-08-21
[Encryption] Add Int64 type

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | https://github.com/doctrine/mongodb-odm/pull/2779#discussion_r2278161437

#### Summary

Queryable Encryption is more strict. While the PHP driver automatically switch the type of the BSON value between `int` (32 bits) or `long` (64 bits) depending on the actual value, the encrypted fields configuration requires a unique type.

This new `Int64` type is only useful for very large int encrypted fields.

### DIFF
--- a/docs/en/reference/basic-mapping.rst
+++ b/docs/en/reference/basic-mapping.rst
@@ -145,6 +145,7 @@ Here is a quick overview of the built-in mapping types:
 -  ``hash``
 -  ``id``
 -  ``int``
+-  ``int64``
 -  ``key``
 -  ``object_id``
 -  ``raw``
--- a/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
@@ -2386,6 +2386,16 @@ use function trigger_deprecation;
             $mapping['nullable'] = false;
         }
 
+        if (isset($mapping['encrypt']['queryType'])) {
+            // The encrypted range query options min and max must be converted to the database type
+            $type = Type::getType($mapping['type']);
+            foreach (['min', 'max'] as $option) {
+                if (isset($mapping['encrypt'][$option])) {
+                    $mapping['encrypt'][$option] = $type->convertToDatabaseValue($mapping['encrypt'][$option]);
+                }
+            }
+        }
+
         if (
             isset($mapping['reference'])
             && isset($mapping['storeAs'])
--- a/lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
@@ -940,7 +940,7 @@ class XmlDriver extends FileDriver
         foreach ($encrypt->attributes() as $key => $value) {
             $encryptMapping[$key] = match ($key) {
                 'queryType' => EncryptQuery::from((string) $value),
-                'min', 'max' => $value,
+                'min', 'max' => (string) $value,
                 'sparsity', 'precision', 'trimFactor', 'contention' => (int) $value,
             };
         }
--- a/lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
+++ b/lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
@@ -9,7 +9,6 @@ use Doctrine\ODM\MongoDB\Mapping\Annotations\TimeSeries;
 use Doctrine\ODM\MongoDB\Mapping\ClassMetadata;
 use Doctrine\ODM\MongoDB\Mapping\MappingException;
 use Doctrine\ODM\MongoDB\Mapping\TimeSeries\Granularity;
-use Doctrine\ODM\MongoDB\Types\Type;
 use Doctrine\ODM\MongoDB\Utility\CollectionHelper;
 use Doctrine\Persistence\Mapping\Driver\FileDriver;
 use DOMDocument;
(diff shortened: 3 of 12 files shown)


docs/en/reference/basic-mapping.rst
```diff
@@ -145,6 +145,7 @@ Here is a quick overview of the built-in mapping types:
 -  ``hash``
 -  ``id``
 -  ``int``
+-  ``int64``
```
jmikola, line 148: Contrary to the PR description, `int64` might have uses beyond encryption such as satisfying a general schema or improving compatibility in cross-language apps (other drivers might not play nice with interchangeable integer types).

lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
```diff
@@ -941,7 +940,7 @@ private function addEncryptionMapping(SimpleXMLElement $encrypt, string $type):
         foreach ($encrypt->attributes() as $key => $value) {
             $encryptMapping[$key] = match ($key) {
                 'queryType' => EncryptQuery::from((string) $value),
-                'min', 'max' => Type::getType($type)->convertToDatabaseValue((string) $value),
+                'min', 'max' => (string) $value,
```
jmikola, line 943: I noted the change to `ClassMetadata::mapField()`, where the `min` and `max` options are always filtered through `Type::convertToDatabaseValue()`. Why was a change only needed in the XmlDriver, though?

tests/Doctrine/ODM/MongoDB/Tests/Mapping/Driver/fixtures/xml/Documents.Encryption.RangeTypes.dcm.xml
```diff
@@ -10,6 +10,9 @@
         <field name="intField" type="int">
             <encrypt queryType="range" min="5" max="10"/>
         </field>
+        <field name="int64Field" type="int64">
+            <encrypt queryType="range" min="5" max="9223372036854775802"/>
```
jmikola, line 14: Is it even necessary to use `PHP_INT_MAX` for this example? Since Int64 types allow comparisons with integers, the assertion in AbstractDriverTestCase doesn't actually check that the value is converted to an Int64 object. It may make more sense to use relaxed numeric comparisons for the `assertEquals` and then also assert that the value is an instance of Int64. In that case you can do without the large values here.

tests/Doctrine/ODM/MongoDB/Tests/Types/TypeTest.php
```diff
@@ -24,40 +27,46 @@
...
+        $bsonValue ??= $phpValue;
+        $type        = Type::getType($typeName);
+
+        self::assertEquals($phpValue, $type->convertToPHPValue($bsonValue));
+        self::assertEquals($bsonValue, $type->convertToDatabaseValue($phpValue));
     }
 
     public static function provideTypes(): array
     {
+        $array = ['foo' => 'bar'];
```
jmikola: This appears to be unused.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -2386,6 +2386,16 @@ public function mapField(array $mapping): array
             $mapping['nullable'] = false;
         }
 
+        if (isset($mapping['encrypt']['queryType'])) {
+            // The encrypted range query options min and max must be converted to the database type
+            $type = Type::getType($mapping['type']);
+            foreach (['min', 'max'] as $option) {
+                if (isset($mapping['encrypt'][$option])) {
+                    $mapping['encrypt'][$option] = $type->convertToDatabaseValue($mapping['encrypt'][$option]);
```
jmikola, line 2394: Is this change only relevant to the int64 mapping, or might it be relevant to handling dates and decimal128 values as well?

lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
GromNaN: With XML you can only get an XML attribute instance converted to a string; while the PHP attribute allows all PHP types to be used. That's why I relied on the user to use the correct value type for min and max.

But I figured that it was better to move the conversation logic to the metadata class.

docs/en/reference/basic-mapping.rst
GromNaN: Ok, the new type can be used independently from encryption.

Note that the `int` type will encode to Int64 when the value exceeds 32 bits. That would be a programmer issue. I could add an Int32 type to throw an error when the value does not fit in 32 bits.

tests/Doctrine/ODM/MongoDB/Tests/Mapping/Driver/fixtures/xml/Documents.Encryption.RangeTypes.dcm.xml
GromNaN: I added a type assertion in `TypeTest`.

My idea for using a low and a high value for the min and max was to test that they are both converted to `Int64`. I could add a functional test for QE.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
GromNaN: It's relevant for Int64, Decimal128 and UTCDateTime.

But also for `int` and `float`, since the value is sent as `string` by the XML driver.

lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
jmikola: Ah, I suppose the only trade off is allowing PHP attributes to be a bit sloppy with typing. But that's probably more convenient for users. No objection.

docs/en/reference/basic-mapping.rst
jmikola: > Note that the int type will encode to Int64 when the value exceeds 32 bits.

That's legacy behavior for ODM and the expected behavior for the server (e.g. incrementing an int32 BSON type server-side could yield an int64), so I'm not worried about it. And so long as ODM continues to accept int64 BSON types for `int` mappings I think we're OK.

> I could add an Int32 type to throw an error when the value does not fit in 32 bits.

This could certainly be added down the line. I'd probably wait until it was requested, though.

PR #2813, opened 2025-09-01
Convert the value of the "version" field using its known type

### LINKED ISSUE #2789: Unable to use `#[Version]` with custom type
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.10.2 (and latest)

#### Summary

I try to use [#[Version] ](https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/2.9/reference/attributes-reference.html#version) to have optimistic locking, however it does not work with custom type.

#### Current behavior

My new type supports Ramsey/Uuid and it's register like this in the bundle

```yaml
doctrine_mongodb:
    types:
        Ramsey\Uuid\Uuid: App\Doctrine\ODM\Types\UuidType
```

When I try to update the entity, I have the following error

```
Doctrine\ODM\MongoDB\LockException: A lock failed on a document.
vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/LockException.php:24
vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php:410
vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/UnitOfWork.php:1239
vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/UnitOfWork.php:3125
vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/UnitOfWork.php:472
vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/DocumentManager.php:585
tests/Integration/Document/FooTest.php:30
```

#### How to reproduce

I created a reproducer: https://github.com/alamirault/reproducer-version-custom-type
- `composer install`
- update `mongodb_url` and `mongodb_database` in config/services.yaml
- `php bin/phpunit`

#### Expected behavior

`#[Version]` attribute must work with custom types

##### Proposal

I dig a bit and saw we call `Type::convertPHPToDatabaseValue($value)` with current version (Uuid instance) and next version (Uuid instance) which call `Type::getTypeFromPHPVariable($variable)`.

https://github.com/doctrine/mongodb-odm/blob/aba7b3050b3fd13d51122e789a2b53a32d66312a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php#L376-L384

This static method `getTypeFromPHPVariable` doesn't try to find type because variable is an object.

https://github.com/doctrine/mongodb-odm/blob/aba7b3050b3fd13d51122e789a2b53a32d66312a/lib/Doctrine/ODM/MongoDB/Types/Type.php#L157-L183

Possible solutions:

<details>

<summary>Update `Type::getTypeFromPHPVariable` to try `self::getType($variable::class)` when is an object</summary>

Something like this
```php
    public static function getTypeFromPHPVariable($variable): ?Type
    {
        if (is_object($variable)) {
            if ($variable instanceof DateTimeInterface) {
                return self::getType('date');
            }

            if ($variable instanceof ObjectId) {
                return self::getType('id');
            }

           // Try to found type from object variable as described in phpdoc "Get a Type instance based on the type of the passed php variable."
            try {
                return self::getType($variable::class);
            } catch (InvalidArgumentException $exception) {
                // no-op
            }
        } else {
            $type = gettype($variable);
            switch ($type) {
                case 'integer':
                    return self::getType('int');
            }
        }

        return null;
    }
```
</details>

<details>

<summary>Update DocumentPersister</summary>

At the point we already know version field type. So instead guess type from $currentVersion and $nextVersion, we can convert with retrieved type.
It's maybe be a BC Break if `$type->getNextVersion` doesn't return a value supported by itself
```php
        if ($this->class->isVersioned) {
            $versionMapping = $this->class->fieldMappings[$this->class->versionField];
            $currentVersion = $this->class->reflFields[$this->class->versionField]->getValue($document);
            $type           = Type::getType($versionMapping['type']);
            assert($type instanceof Versionable);
            $nextVersion                             = $type->getNextVersion($currentVersion);
            
            // Actual value is:
            // $update['$set'][$versionMapping['name']] = Type::convertPHPToDatabaseValue($nextVersion);
            // $query[$versionMapping['name']]          = Type::convertPHPToDatabaseValue($currentVersion);

            // Proposal
            $update['$set'][$versionMapping['name']] = $type->convertToDatabaseValue($nextVersion);
            $query[$versionMapping['name']]          = $type->convertToDatabaseValue($currentVersion);
        }
```
</details>

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | Fix #2789

#### Summary

Instead of guessing the type of the version field value in order to convert the value into a MongoDB compatible value, we can use the type that is configured for this field. Otherwise the detection of the field type from the value type fails for custom types.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -218,7 +218,7 @@ final class DocumentPersister
                     $this->class->reflFields[$this->class->versionField]->setValue($document, $nextVersion);
                 }
 
-                $data[$versionMapping['name']] = $type->convertPHPToDatabaseValue($nextVersion);
+                $data[$versionMapping['name']] = $type->convertToDatabaseValue($nextVersion);
             }
 
             $inserts[] = $data;
@@ -296,7 +296,7 @@ final class DocumentPersister
                 $this->class->reflFields[$this->class->versionField]->setValue($document, $nextVersion);
             }
 
-            $data['$set'][$versionMapping['name']] = $type->convertPHPToDatabaseValue($nextVersion);
+            $data['$set'][$versionMapping['name']] = $type->convertToDatabaseValue($nextVersion);
         }
 
         foreach (array_keys($criteria) as $field) {
@@ -377,8 +377,8 @@ final class DocumentPersister
             $type           = Type::getType($versionMapping['type']);
             assert($type instanceof Versionable);
             $nextVersion                             = $type->getNextVersion($currentVersion);
-            $update['$set'][$versionMapping['name']] = Type::convertPHPToDatabaseValue($nextVersion);
-            $query[$versionMapping['name']]          = Type::convertPHPToDatabaseValue($currentVersion);
+            $update['$set'][$versionMapping['name']] = $type->convertToDatabaseValue($nextVersion);
+            $query[$versionMapping['name']]          = $type->convertToDatabaseValue($currentVersion);
         }
 
         if (! empty($update)) {
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH2789Test.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH2789Test.php
@@ -1,0 +1,88 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ODM\MongoDB\Tests\Functional\Ticket;
+
+use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
+use Doctrine\ODM\MongoDB\Tests\BaseTestCase;
+use Doctrine\ODM\MongoDB\Types\Type;
+use Doctrine\ODM\MongoDB\Types\Versionable;
+use MongoDB\BSON\Binary;
+use PHPUnit\Framework\Attributes\BackupGlobals;
+
+use function assert;
+use function is_int;
+
+#[BackupGlobals(true)]
+class GH2789Test extends BaseTestCase
+{
+    public function testVersionWithCustomType(): void
+    {
+        Type::addType(GH2789CustomType::class, GH2789CustomType::class);
+        $doc = new GH2789VersionedUuid('original message');
+
+        $this->dm->persist($doc);
+        $this->dm->flush();
+
+        $documents = $this->dm->getDocumentCollection(GH2789VersionedUuid::class)->find()->toArray();
+        self::assertCount(1, $documents);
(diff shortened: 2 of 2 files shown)


lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -218,7 +218,7 @@ public function executeInserts(array $options = []): void
                     $this->class->reflFields[$this->class->versionField]->setValue($document, $nextVersion);
                 }
 
-                $data[$versionMapping['name']] = $type->convertPHPToDatabaseValue($nextVersion);
```
GromNaN, line 221: This is a static method called on the instance. So it was possible to change the behavior for a custom type by overriding the method, but that's not how it's supposed to be used.
In fact, this method should be `final`.

paulinevos (approved):

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
jmikola: Are you suggesting that the `convertPHPToDatabaseValue()` should be `final`? If so, I agree and we should probably track that with a separate ticket for the next major version.

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH2789Test.php
```diff
@@ -0,0 +1,88 @@
...
+        assert($value instanceof Binary);
+
+        return (int) $value->getData();
+    }
+
+    public function convertToDatabaseValue($value): Binary
+    {
+        assert(is_int($value));
+
+        return new Binary((string) $value, 42);
```
jmikola: I'd suggest using something in the range of `[128,255]` to stick to the range for custom values (in case users decide to reference this for their own apps).

jmikola (approved):
LGTM, but two small suggestions.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
GromNaN: Yes, that's the idea. But a better approach would be to extract the static methods into a `TypeRegistry`, similar to DBAL. https://github.com/doctrine/mongodb-odm/issues/2818

tests/Doctrine/ODM/MongoDB/Tests/Functional/Ticket/GH2789Test.php
GromNaN: Good point, I changed it to `142`.

PR #2814, opened 2025-09-01
Improve detection of the type from a PHP variable

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | Related to https://github.com/doctrine/mongodb-odm/issues/2789

#### Summary

The method `getTypeFromPHPVariable` is used only by `convertPHPToDatabaseValue`. Which is used to process expressions in the aggregation builder. 

With this new feature, we try to find the type for object class name; assuming custom types are registered using the FQCN of the objects they store.

### DIFF
--- a/docs/en/reference/custom-mapping-types.rst
+++ b/docs/en/reference/custom-mapping-types.rst
@@ -9,6 +9,9 @@ In order to create a new mapping type you need to subclass
 ``Doctrine\ODM\MongoDB\Types\Type`` and implement/override
 the methods.
 
+Date Example: Mapping DateTimeImmutable with Timezone
+-----------------------------------------------------
+
 The following example defines a custom type that stores ``DateTimeInterface``
 instances as an embedded document containing a BSON date and accompanying
 timezone string. Those same embedded documents are then be translated back into
@@ -32,6 +35,7 @@ a ``DateTimeImmutable`` when the data is read from the database.
         // This trait provides default closureToPHP used during data hydration
         use ClosureToPHP;
 
+        /** @param array{utc: UTCDateTime, tz: string} $value */
         public function convertToPHPValue($value): DateTimeImmutable
         {
             if (!isset($value['utc'], $value['tz'])) {
@@ -46,6 +50,7 @@ a ``DateTimeImmutable`` when the data is read from the database.
             return DateTimeImmutable::createFromMutable($dateTime);
         }
 
+        /** @return array{utc: UTCDateTime, tz: string} */
         public function convertToDatabaseValue($value): array
         {
             if (!$value instanceof DateTimeImmutable) {
@@ -115,5 +120,114 @@ specify a unique name for the mapping type and map that to the corresponding
 
         <field field-name="field" type="date_with_timezone" />
 
+Custom Type Example: Mapping a UUID Class
+-----------------------------------------
+
+You can create a custom mapping type for your own value objects or classes. For
+example, to map a UUID value object using the `ramsey/uuid library`_, you can
+implement a type that converts between your class and the BSON Binary UUID format.
+
+This approach works for any custom class by adapting the conversion logic to your needs.
+
+Example Implementation (using ``Ramsey\Uuid\Uuid``)::
+
+.. code-block:: php
+
+    <?php
+
+    namespace My\Project\Types;
+
+    use Doctrine\ODM\MongoDB\Types\ClosureToPHP;
+    use Doctrine\ODM\MongoDB\Types\Type;
+    use InvalidArgumentException;
+    use MongoDB\BSON\Binary;
+    use Ramsey\Uuid\Uuid;
+    use Ramsey\Uuid\UuidInterface;
+
+    final class UuidType extends Type
+    {
+        // This trait provides default closureToPHP used during data hydration
+        use ClosureToPHP;
+
+        public function convertToPHPValue(mixed $value): ?Uuid
+        {
+            if (null === $value) {
+                return null;
+            }
+
+            if ($value instanceof Uuid) {
+                return $value;
+            }
+
+            if ($value instanceof Binary) {
+                return Uuid::fromBytes($value->getData());
+            }
+
+            if (is_string($value) && Uuid::isValid($value)) {
+                return Uuid::fromString($value);
+            }
+
+            throw new InvalidArgumentException(
+                sprintf(
+                    'Could not convert database value "%s" from "%s" to %s',
+                    $value,
+                    get_debug_type($value),
+                    UuidInterface::class
+                )
+            );
+        }
+
+        public function convertToDatabaseValue(mixed $value): ?Binary
+        {
+            if (null === $value || [] === $value) {
+                return null;
+            }
+
+            if ($value instanceof Binary) {
+                return $value;
+            }
+
+            if (is_string($value) && Uuid::isValid($value)) {
+                $value = Uuid::fromString($value)->getBytes();
+            }
+
+            if ($value instanceof Uuid) {
+                return new Binary($value->getBytes(), Binary::TYPE_UUID);
+            }
+
+            throw new InvalidArgumentException(
+                sprintf(
+                    'Could not convert database value "%s" from "%s" to %s',
+                    $value,
+                    get_debug_type($value),
+                    Binary::class
+                )
+            );
+        }
+    }
+
+Register the type in your bootstrap code::
+
+.. code-block:: php
+
+    Type::addType(Ramsey\Uuid\Uuid::class, My\Project\Types\UuidType::class);
+
+Usage Example::
+
+.. code-block:: php
+
+    #[Field(type: Ramsey\Uuid\Uuid::class)]
+    public Ramsey\Uuid\Uuid $id;
+
+By using the |FQCN| of the value object class as the type name, the type is
+automatically used when encountering a property of that class. This means you
+can omit the ``type`` option when defining the field mapping::
+
+.. code-block:: php
+
+    #[Field]
+    public Ramsey\Uuid\Uuid $id;
+
+.. _`ramsey/uuid library`: https://github.com/ramsey/uuid
 .. |FQCN| raw:: html
   <abbr title="Fully-Qualified Class Name">FQCN</abbr>
--- a/docs/en/reference/custom-mapping-types.rst
+++ b/docs/en/reference/custom-mapping-types.rst
@@ -135,74 +135,43 @@ Example Implementation (using ``Ramsey\Uuid\Uuid``)::
 
     <?php
 
-    namespace My\Project\Types;
+    namespace App\MongoDB\Types;
 
     use Doctrine\ODM\MongoDB\Types\ClosureToPHP;
     use Doctrine\ODM\MongoDB\Types\Type;
     use InvalidArgumentException;
-    use MongoDB\BSON\Binary;
-    use Ramsey\Uuid\Uuid;
-    use Ramsey\Uuid\UuidInterface;
+    use MongoDB\BSON\Binary as BsonBinary;
+    use Ramsey\Uuid\Uuid as RamseyUuid;
 
-    final class UuidType extends Type
+    final class RamseyUuidType extends Type
     {
         // This trait provides default closureToPHP used during data hydration
         use ClosureToPHP;
 
-        public function convertToPHPValue(mixed $value): ?Uuid
+        public function convertToPHPValue(mixed $value): ?RamseyUuid
         {
             if (null === $value) {
                 return null;
             }
 
-            if ($value instanceof Uuid) {
-                return $value;
+            if ($value instanceof BsonBinary && $value->getType() === BsonBinary::TYPE_UUID) {
+                return RamseyUuid::fromBytes($value->getData());
             }
 
-            if ($value instanceof Binary) {
-                return Uuid::fromBytes($value->getData());
-            }
-
-            if (is_string($value) && Uuid::isValid($value)) {
-                return Uuid::fromString($value);
-            }
-
-            throw new InvalidArgumentException(
-                sprintf(
-                    'Could not convert database value "%s" from "%s" to %s',
-                    $value,
-                    get_debug_type($value),
-                    UuidInterface::class
-                )
-            );
+            throw new \InvalidArgumentException(\sprintf('Could not convert database value from "%s" to %s',get_debug_type($value),RamseyUuid::class));
         }
 
-        public function convertToDatabaseValue(mixed $value): ?Binary
+        public function convertToDatabaseValue(mixed $value): ?BsonBinary
         {
-            if (null === $value || [] === $value) {
+            if (null === $value) {
                 return null;
             }
 
-            if ($value instanceof Binary) {
-                return $value;
+            if ($value instanceof RamseyUuid) {
+                return new BsonBinary($value->getBytes(), BsonBinary::TYPE_UUID);
             }
 
-            if (is_string($value) && Uuid::isValid($value)) {
-                $value = Uuid::fromString($value)->getBytes();
-            }
-
-            if ($value instanceof Uuid) {
-                return new Binary($value->getBytes(), Binary::TYPE_UUID);
-            }
-
-            throw new InvalidArgumentException(
-                sprintf(
-                    'Could not convert database value "%s" from "%s" to %s',
-                    $value,
-                    get_debug_type($value),
-                    Binary::class
-                )
-            );
+            throw new \InvalidArgumentException(\sprintf('Could not convert database value from "%s" to %s', get_debug_type($value), Binary::class));
         }
     }
 
@@ -210,14 +179,14 @@ Register the type in your bootstrap code::
 
 .. code-block:: php
 
-    Type::addType(Ramsey\Uuid\Uuid::class, My\Project\Types\UuidType::class);
+    Type::addType(Ramsey\Uuid\Uuid::class, App\MongoDB\Types\RamseyUuidType::class);
 
 Usage Example::
 
 .. code-block:: php
 
-    #[Field(type: Ramsey\Uuid\Uuid::class)]
-    public Ramsey\Uuid\Uuid $id;
+    #[Field(type: \Ramsey\Uuid\Uuid::class)]
+    public ?\Ramsey\Uuid\Uuid $id;
 
 By using the |FQCN| of the value object class as the type name, the type is
 automatically used when encountering a property of that class. This means you
@@ -226,7 +195,13 @@ can omit the ``type`` option when defining the field mapping::
 .. code-block:: php
 
     #[Field]
-    public Ramsey\Uuid\Uuid $id;
+    public ?\Ramsey\Uuid\Uuid $id;
+
+.. note::
+
+    This implementation of ``RamseyUuidType`` is volontary simple and does not
+    handle all edge cases, but it should give you a good starting point for
+    implementing your own custom types.
 
 .. _`ramsey/uuid library`: https://github.com/ramsey/uuid
 .. |FQCN| raw:: html
--- a/docs/en/reference/custom-mapping-types.rst
+++ b/docs/en/reference/custom-mapping-types.rst
@@ -120,16 +120,16 @@ specify a unique name for the mapping type and map that to the corresponding
 
         <field field-name="field" type="date_with_timezone" />
 
-Custom Type Example: Mapping a UUID Class
------------------------------------------
+Custom Type Example: Mapping a Money Value Object
+-------------------------------------------------
 
 You can create a custom mapping type for your own value objects or classes. For
-example, to map a UUID value object using the `ramsey/uuid library`_, you can
-implement a type that converts between your class and the BSON Binary UUID format.
+example, to map a ``Money`` value object using the `moneyphp/money library`_, you can
+implement a type that converts between this class and a BSON embedded document format.
 
 This approach works for any custom class by adapting the conversion logic to your needs.
 
-Example Implementation (using ``Ramsey\Uuid\Uuid``)::
+Example Implementation (using ``Money\Money``)::
 
 .. code-block:: php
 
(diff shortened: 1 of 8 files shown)


docs/en/reference/custom-mapping-types.rst
```diff
@@ -115,5 +118,108 @@ specify a unique name for the mapping type and map that to the corresponding
...
+                )
+            );
+        }
+    }
+
+Usage Example::
+
+.. code-block:: php
+
+    #[Field(type: Uuid::class)]
```
GromNaN: Setting the custom type here should not be necessary. As soon as you register a type with the value object type name, it should be registered automatically https://github.com/doctrine/mongodb-odm/blob/afb565e6246d2dcdaa81636e5a0020d21c1d6fe5/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php#L2771-L2793

docs/en/reference/custom-mapping-types.rst
```diff
@@ -115,5 +118,108 @@ specify a unique name for the mapping type and map that to the corresponding
 
         <field field-name="field" type="date_with_timezone" />
 
-.. |FQCN| raw:: html
-  <abbr title="Fully-Qualified Class Name">FQCN</abbr>
+Custom Type Example: Mapping a UUID Class
+========================================
+
+You can create a custom mapping type for your own value objects or classes. For
+example, to map a UUID value object using the `ramsey/uuid library`_, you can
```
paulinevos: Nice to see the original use case from the issue as an example here

docs/en/reference/custom-mapping-types.rst
```diff
@@ -115,5 +118,108 @@ specify a unique name for the mapping type and map that to the corresponding
...
+implement a type that converts between your class and the MongoDB binary format.
+
+This approach works for any custom class—just adapt the conversion logic to your needs.
```
paulinevos:
```suggestion
This approach works for any custom class — just adapt the conversion logic to your needs.
```

docs/en/reference/custom-mapping-types.rst
```diff
@@ -115,5 +118,108 @@ specify a unique name for the mapping type and map that to the corresponding
...
+    public Uuid $id;
+
+Register the type in your bootstrap code::
+
+.. code-block:: php
+
+    Type::addType(Ramsey\Uuid\Uuid::class, My\Project\Types\UuidType::class);
+
+.. note::
+   The type name should match the support class name (e.g., ``uuid`` for ``UuidType``) to enable automatic association with the field type.
```
paulinevos: For the automatic association, how does that work? Does it take everything before the `Type` suffix? How does it handle casing, e.g. in `FooBarType`?

docs/en/reference/custom-mapping-types.rst
```diff
@@ -115,5 +118,108 @@ specify a unique name for the mapping type and map that to the corresponding
...
         <field field-name="field" type="date_with_timezone" />
 
-.. |FQCN| raw:: html
-  <abbr title="Fully-Qualified Class Name">FQCN</abbr>
+Custom Type Example: Mapping a UUID Class
+========================================
+
+You can create a custom mapping type for your own value objects or classes. For
+example, to map a UUID value object using the `ramsey/uuid library`_, you can
+implement a type that converts between your class and the MongoDB binary format.
```
jmikola: Should this say "BSON binary format", or do you think that'd just be more confusing for users? OK to leave as-is.

docs/en/reference/custom-mapping-types.rst
jmikola: See also: https://www.merriam-webster.com/grammar/em-dash-en-dash-how-to-use

Typically wouldn't have spaces on either side of the em dash. Alternatively, you can rephrase this as:

> This approach works for any custom class by adapting the conversion logic to your needs.

"Just" may be considered subjective language, the same way we wouldn't want to say "easily" or "simply".

docs/en/reference/custom-mapping-types.rst
```diff
@@ -115,5 +118,108 @@ specify a unique name for the mapping type and map that to the corresponding
...
+
+            if (is_string($value) && Uuid::isValid($value)) {
+                return Uuid::fromString($value);
+            }
+
+            throw new InvalidArgumentException(
+                sprintf(
+                    'Could not convert database value "%s" from "%s" to %s',
+                    $value,
+                    get_debug_type($value),
```
jmikola: Are `use function` lines necessary? I'm OK to omit them, but just want to ensure you've considered the question.

docs/en/reference/custom-mapping-types.rst
jmikola: Is it clear from this example that you're referencing `Ramsey\Uuid\Uuid`? There's no `use` statement and this is a separate code block from the custom type class above.

docs/en/reference/custom-mapping-types.rst
jmikola: I'm not certain, but I think this might utilize [doctrine/inflector](https://github.com/doctrine/inflector) internally.

docs/en/reference/custom-mapping-types.rst
```diff
@@ -115,5 +118,108 @@ specify a unique name for the mapping type and map that to the corresponding
...
+.. code-block:: php
+
+    Type::addType(Ramsey\Uuid\Uuid::class, My\Project\Types\UuidType::class);
+
+.. note::
+   The type name should match the support class name (e.g., ``uuid`` for ``UuidType``) to enable automatic association with the field type.
+
+This pattern can be adapted for any custom class—just implement the conversion logic for your value object.
+
+.. |FQCN| raw:: html <abbr title="Fully-Qualified Class Name">FQCN</abbr>
```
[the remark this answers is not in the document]
jmikola: @GromNaN: this looks like it may be a false positive, since this is some very niche RST syntax

docs/en/reference/custom-mapping-types.rst
```diff
@@ -115,5 +118,108 @@ specify a unique name for the mapping type and map that to the corresponding
...
+Register the type in your bootstrap code::
+
+.. code-block:: php
+
+    Type::addType(Ramsey\Uuid\Uuid::class, My\Project\Types\UuidType::class);
+
+.. note::
+   The type name should match the support class name (e.g., ``uuid`` for ``UuidType``) to enable automatic association with the field type.
+
+This pattern can be adapted for any custom class—just implement the conversion logic for your value object.
```
jmikola: Same feedback here as @paulinevos' thread above re: "class-just"

lib/Doctrine/ODM/MongoDB/Types/InvalidTypeException.php
```diff
@@ -0,0 +1,17 @@
...
+
+use InvalidArgumentException;
+
+use function sprintf;
+
+final class InvalidTypeException extends InvalidArgumentException
+{
+    public static function invalidTypeName(string $name): self
+    {
+        return new self(sprintf('Invalid type specified "%s".', $name));
```
jmikola: Should this use a colon?

```suggestion
        return new self(sprintf('Invalid type specified: "%s".', $name));
```

src/Types/Type.php
```diff
@@ -140,48 +139,52 @@ public static function registerType(string $name, string $class): void
     /**
      * Get a Type instance.
      *
-     * @throws InvalidArgumentException
+     * @throws InvalidTypeException
```
jmikola, line 158: Noted that InvalidTypeException extends InvalidArgumentException so there's no BC concern here.

lib/Doctrine/ODM/MongoDB/Types/Type.php
```diff
@@ -140,48 +139,52 @@ public static function registerType(string $name, string $class): void
...
+                return null;
             }
         }
 
-        return null;
+        return match (gettype($variable)) {
+            'integer' => self::getType('int'),
+            'boolean' => self::getType('bool'),
+            'double' => self::getType('float'),
+            'string' => self::getType('string'),
```
jmikola: Any reason you're not referencing the constants here, as you did above?

docs/en/reference/custom-mapping-types.rst
paulinevos: Oh my bad, I always write it like that. Maybe it's the GH formatting of this em dash as well. It read to me like a hyphen cause it looks so short

docs/en/reference/custom-mapping-types.rst
paulinevos: I think it would be helpful to add here what format would constitute a match.

docs/en/reference/custom-mapping-types.rst
GromNaN: In French we would add space around the dash in that context too. I recognize that it was copilot who wrote this text.

docs/en/reference/custom-mapping-types.rst
GromNaN: I was tempted to modify the example to use `symfony/uid` instead. I've kept this example so as not to be too partisan.

paulinevos: I'm a ramsey/uuid traditionalist anyway 😎

docs/en/reference/custom-mapping-types.rst
GromNaN: In fact, there is no such magic. The name should be the class name of the value object.

docs/en/reference/custom-mapping-types.rst
GromNaN: This sentence was duplicate if the previous one. I removed it.

docs/en/reference/custom-mapping-types.rst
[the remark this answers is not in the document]
GromNaN: I revert changes on this line. I'm not sure how this will be formatted.

docs/en/reference/custom-mapping-types.rst
GromNaN: Updated to use the FQCN.

lib/Doctrine/ODM/MongoDB/Types/Type.php
GromNaN: No reason. Updated.

docs/en/reference/custom-mapping-types.rst
```diff
@@ -115,5 +118,114 @@ specify a unique name for the mapping type and map that to the corresponding
...
+    public Ramsey\Uuid\Uuid $id;
+
+By using the |FQCN| of the value object class as the type name, the type is
+automatically used when encountering a property of that class. This means you
+can omit the ``type`` option when defining the field mapping::
+
+.. code-block:: php
+
+    #[Field]
+    public Ramsey\Uuid\Uuid $id;
```
GromNaN: This is a new feature that can compete with embedded documents: adding custom logic to convert a value object into BSON.

GromNaN (commented):
Thanks for the review. I added the type detection for DTO class name in property type.

src/Types/InvalidTypeException.php
GromNaN, line 1: We don't have a global namespace for exception like it is generally done in libraries. Existing exception names are:
- `Doctrine\ODM\MongoDB\MongoDBException`
- `Doctrine\ODM\MongoDB\ConfigurationException`
- `Doctrine\ODM\MongoDB\DocumentNotFoundException`
- `Doctrine\ODM\MongoDB\Mapping\MappingException`
- `Doctrine\ODM\MongoDB\Hydrator\HydratorException`
- `Doctrine\ODM\MongoDB\LockException`

docs/en/reference/custom-mapping-types.rst
```diff
@@ -115,5 +118,114 @@ specify a unique name for the mapping type and map that to the corresponding
 
         <field field-name="field" type="date_with_timezone" />
 
+Custom Type Example: Mapping a UUID Class
+=========================================
```
GromNaN:
```suggestion
-----------------------------------------
```

docs/en/reference/custom-mapping-types.rst
```diff
@@ -115,5 +118,114 @@ specify a unique name for the mapping type and map that to the corresponding
...
+        {
+            if (null === $value) {
+                return null;
+            }
+
+            if ($value instanceof Uuid) {
+                return $value;
+            }
+
+            if ($value instanceof Binary) {
```
jmikola: Would it make sense to restrict this to a [MongoDB\BSON\Binary](https://www.php.net/manual/en/class.mongodb-bson-binary.php) where `getType()` returns either `TYPE_OLD_UUID` or `TYPE_UUID`? As-is, it's not clear what would happen if you encountered binary data from the server with a different sub-type.

I'll note that `convertToDatabaseValue()` yields a Binary object with `TYPE_UUID`, so perhaps there's no reason to accept `TYPE_OLD_UUID` here.

If you are intentionally being liberal in what you accept and conservative in what you emit ([Robustness Principle](https://en.wikipedia.org/wiki/Robustness_principle)) and would like to leave this as-is, that's fine by me. You can optionally add a comment here explaining that design decision for the benefit of readers.

docs/en/reference/custom-mapping-types.rst
```diff
@@ -115,5 +118,114 @@ specify a unique name for the mapping type and map that to the corresponding
...
+            );
+        }
+
+        public function convertToDatabaseValue(mixed $value): ?Binary
+        {
+            if (null === $value || [] === $value) {
+                return null;
+            }
+
+            if ($value instanceof Binary) {
```
jmikola: Would you want to ensure you always yield a Binary with sub-type `TYPE_UUID`? If so, you might need to split this into two separate conditionals.

```php
if ($value instanceof Binary && $value->getType() !== Binary::TYPE_UUID) {
    // Note: this may throw if the data is not exactly 16 bytes
    return new Binary($value->getBytes(), Binary::TYPE_UUID);
}

if ($value instanceof Binary) {
    return $value;
}
```

src/Mapping/ClassMetadata.php
```diff
@@ -2750,9 +2750,19 @@ private function isTypedProperty(string $name): bool
      */
     private function validateAndCompleteTypedFieldMapping(array $mapping): array
     {
+        if (isset($mapping['type'])) {
+            return $mapping;
+        }
```
jmikola, line 2828: Noted that you split this off to avoid an unnecessary call to `ReflectionClass::getProperty()`.

docs/en/reference/custom-mapping-types.rst
```diff
@@ -115,5 +118,114 @@ specify a unique name for the mapping type and map that to the corresponding
 
         <field field-name="field" type="date_with_timezone" />
 
+Custom Type Example: Mapping a UUID Class
```
alcaeus: Note: I'm currently working on adding a type for Symfony UUIDs as I needed them for a project. That would make this entire chapter obsolete.

GromNaN: The chapter is about mapping a DTO with a custom type. UUID is only an example, we can find an other example.

docs/en/reference/custom-mapping-types.rst
GromNaN: The `use function` are not necessary for global functions. That's a CS preference to use them. I omit them for brevity in the example.

docs/en/reference/custom-mapping-types.rst
```diff
@@ -115,5 +118,114 @@ specify a unique name for the mapping type and map that to the corresponding
...
+
+        public function convertToDatabaseValue(mixed $value): ?Binary
+        {
+            if (null === $value || [] === $value) {
+                return null;
+            }
+
+            if ($value instanceof Binary) {
+                return $value;
+            }
```
GromNaN: I can simplify the example and remove this edge cases.
```suggestion
            if (null === $value) {
                return null;
            }
```

docs/en/reference/custom-mapping-types.rst
GromNaN: I'd rather simplify the example.
https://github.com/doctrine/mongodb-odm/pull/2814/files#r2393846289

docs/en/reference/custom-mapping-types.rst
GromNaN: Maybe the example should be with a custom class instead, since UUID are new supported using `symfony/uid` https://github.com/doctrine/mongodb-odm/pull/2826

docs/en/reference/custom-mapping-types.rst
jmikola: Without the `$value instanceof Binary` check, wouldn't the method then throw InvalidArgumentException for a Binary value? Or are you comfortable assuming that a Binary would never be passed in?

I didn't follow why `[]` was originally being handled alongside `null`, so I don't have much thought on that change.

GromNaN: In the example of document class, the property has the `?\Ramsey\Uuid\Uuid` type, which ensure the `RamseyUuidType::convertToDatabaseValue()` method never get anything else.
As soon as the field value is the database is not corrupted, the simple implementation will work perfectly. No need to add support for edge cases that are never encountered.

docs/en/reference/custom-mapping-types.rst
```diff
@@ -115,5 +120,89 @@ specify a unique name for the mapping type and map that to the corresponding
...
+        {
+            if (null === $value) {
+                return null;
+            }
+
+            if ($value instanceof BsonBinary && $value->getType() === BsonBinary::TYPE_UUID) {
+                return RamseyUuid::fromBytes($value->getData());
+            }
+
+            throw new \InvalidArgumentException(\sprintf('Could not convert database value from "%s" to %s',get_debug_type($value),RamseyUuid::class));
```
jmikola: Should you add spaces after the argument-delimiting commas? You did that below.

docs/en/reference/custom-mapping-types.rst
```diff
@@ -115,5 +120,89 @@ specify a unique name for the mapping type and map that to the corresponding
...
+can omit the ``type`` option when defining the field mapping::
+
+.. code-block:: php
+
+    #[Field]
+    public ?\Ramsey\Uuid\Uuid $id;
+
+.. note::
+
+    This implementation of ``RamseyUuidType`` is volontary simple and does not
```
jmikola: I assume this is a typo for "voluntary" but I don't think that makes sense. Maybe say "intentionally simple"?

jmikola (commented):
I'll defer to @alcaeus to sign off on this.

docs/en/reference/custom-mapping-types.rst
GromNaN:
```suggestion
    This implementation of ``RamseyUuidType`` is intentionally simple and does not
```

docs/en/reference/custom-mapping-types.rst
GromNaN: Good catch!
```suggestion
            throw new \InvalidArgumentException(\sprintf('Could not convert database value from "%s" to %s', get_debug_type($value), RamseyUuid::class));
```

src/Types/InvalidTypeException.php
```diff
@@ -0,0 +1,17 @@
...
+
+namespace Doctrine\ODM\MongoDB\Types;
+
+use InvalidArgumentException;
+
+use function sprintf;
+
+final class InvalidTypeException extends InvalidArgumentException
+{
+    public static function invalidTypeName(string $name): self
```
GromNaN, line 13: The factory methods of exceptions are internal.
```suggestion
    /** @internal */
    public static function invalidTypeName(string $name): self
```

src/Types/Type.php
```diff
@@ -143,52 +141,52 @@ public static function registerType(string $name, string $class): void
...
     {
         if (is_object($variable)) {
-            if ($variable instanceof DateTimeInterface) {
-                return self::getType(self::DATE);
+            if ($variable instanceof DateTimeImmutable) {
+                return self::getType(self::DATE_IMMUTABLE);
             }
 
-            if ($variable instanceof ObjectId) {
-                return self::getType(self::ID);
```
GromNaN, line 191: I must check why this could be removed.

GromNaN: In fact, this type detection is invalid when the property type is `ObjectId`:
- the `IdType` class does nothing as the target type is also `ObjectId`
- the `IdType` class convert the `ObjectId` received from the Database into `string`, which is invalid when assigning the value to the `ObjectId` property.

docs/en/reference/custom-mapping-types.rst
```diff
@@ -115,5 +120,89 @@ specify a unique name for the mapping type and map that to the corresponding
 
         <field field-name="field" type="date_with_timezone" />
 
+Custom Type Example: Mapping a UUID Class
```
alcaeus: With UUID merged, WDYT of using [moneyphp](https://github.com/moneyphp/money) as an example?

src/Mapping/ClassMetadata.php
```diff
@@ -2786,9 +2786,19 @@ private function isTypedProperty(string $name): bool
...
+
         $type = $this->reflClass->getProperty($mapping['fieldName'])->getType();
 
-        if (! $type instanceof ReflectionNamedType || isset($mapping['type'])) {
+        if (! $type instanceof ReflectionNamedType) {
+            return $mapping;
+        }
+
+        if (! $type->isBuiltin() && Type::hasType($type->getName())) {
```
alcaeus, line 2836: If I'm not mistaken, this would allow us to use a FQCN as identifier for a type name to automatically link values of that class with a given type? I haven't even thought of that before, but it makes so much sense!

src/Types/InvalidTypeException.php
alcaeus: We could introduce a marker interface similar to how we have `MongoDB\Driver\Exception\Exception` in the extension.

alcaeus (commented):
Changes LGTM, but I would prefer changing out the UUID example.

docs/en/reference/custom-mapping-types.rst
GromNaN: I updated the tutorial.

docs/en/reference/custom-mapping-types.rst
```diff
@@ -115,5 +120,92 @@ specify a unique name for the mapping type and map that to the corresponding
...
+Usage Example::
+
+.. code-block:: php
+
+    #[Field(type: \Money\Money::class)]
+    public ?\Money\Money $price;
+
+By using the |FQCN| of the value object class as the type name, the type is
+automatically used when encountering a property of that class. This means you
+can omit the ``type`` option when defining the field mapping::
```
alcaeus: I would remove the first example and explain that the mapping logic automatically detects the property type and uses it as field type.

docs/en/reference/custom-mapping-types.rst
```diff
@@ -115,5 +120,92 @@ specify a unique name for the mapping type and map that to the corresponding
...
+
+.. code-block:: php
+
+    #[Field]
+    public ?\Money\Money $price;
+
+.. note::
+
+    This implementation of ``MoneyType`` is volontairely simple and does not
```
alcaeus:
```suggestion
    This implementation of ``MoneyType`` is kept simple for illustration purposes and does not
```

alcaeus (approved):
LGTM with two suggestions for the docs

PR #2815, opened 2025-09-01
Enable optimistic locking using an `ObjectId` as version field

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | Related to https://github.com/doctrine/mongodb-odm/issues/2789

#### Summary

Optimistic Locking use a `version` field that is incremented on each version.
- Using a `int` field is safe but doesn't provide any information about the last update
- Using a `date` field provides the last update date, but is not safe if updates occurs during the same millisecond.
- Using an a MongoDB `ObjectId` is perfect as it is both safe, sortable and provide the update date using [`ObjectId::getTimestamp()`](https://www.php.net/manual/en/mongodb-bson-objectid.gettimestamp.php)

### DIFF
--- a/docs/en/reference/transactions-and-concurrency.rst
+++ b/docs/en/reference/transactions-and-concurrency.rst
@@ -191,6 +191,7 @@ Choosing the Field Type
 
 When using the date-based type in a high-concurrency environment, it is still possible to create multiple documents
 with the same version and cause a conflict. This can be avoided by using the ``int``, ``decimal128``, or ``object_id`` type.
+The ``object_id`` type contains the timestamp of its creation, but also a random value to ensure uniqueness.
 
 Usage
 """""
--- a/docs/en/reference/transactions-and-concurrency.rst
+++ b/docs/en/reference/transactions-and-concurrency.rst
@@ -105,7 +105,7 @@ a ``LockException`` is thrown, which indicates that the document was already mod
 .. note::
 
     Only types implementing the ``\Doctrine\ODM\MongoDB\Types\Versionable`` interface can be used for versioning.
-    Following ODM types can be used for versioning: ``int``, ``decimal128``, ``date``, and ``date_immutable``.
+    Following ODM types can be used for versioning: ``int``, ``decimal128``, ``date``, ``date_immutable``, and ``object_id``.
 
 Document Configuration
 ^^^^^^^^^^^^^^^^^^^^^^
@@ -190,7 +190,7 @@ Choosing the Field Type
 """""""""""""""""""""""
 
 When using the date-based type in a high-concurrency environment, it is still possible to create multiple documents
-with the same version and cause a conflict. This can be avoided by using the ``int`` or ``decimal128`` type.
+with the same version and cause a conflict. This can be avoided by using the ``int``, ``decimal128``, or ``object_id`` type.
 
 Usage
 """""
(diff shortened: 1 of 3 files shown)


tests/Doctrine/ODM/MongoDB/Tests/Types/VersionableTest.php
```diff
@@ -0,0 +1,73 @@
...
+        $type        = Type::getType(Type::OBJECTID);
+        $nextVersion = $type->getNextVersion(null);
+        self::assertInstanceOf(ObjectId::class, $nextVersion);
+        self::assertGreaterThan($now, $nextVersion);
+        self::assertLessThan(new ObjectId(), $nextVersion);
+
+        $nextVersion = $type->getNextVersion(null);
+        self::assertInstanceOf(ObjectId::class, $now);
+        self::assertGreaterThan($now, $nextVersion);
+        self::assertLessThan(new ObjectId(), $nextVersion);
```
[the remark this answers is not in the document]
GromNaN: ObjectIds are unique.

docs/en/reference/transactions-and-concurrency.rst
```diff
@@ -190,7 +190,7 @@ Choosing the Field Type
 """""""""""""""""""""""
 
 When using the date-based type in a high-concurrency environment, it is still possible to create multiple documents
-with the same version and cause a conflict. This can be avoided by using the ``int`` or ``decimal128`` type.
```
paulinevos, line 193: Honestly think it would be worth explicitly mentioning somewhere the possibility of deriving the timestamp from the object ID. It's quite a good idea and will probably not occur to many users unless it's mentioned somewhere.

paulinevos (approved):

docs/en/reference/transactions-and-concurrency.rst
jmikola: IIRC, ODM converts ObjectId values to strings, so users won't be able to easily utilize [`ObjectId::getTimestamp()`](https://www.php.net/manual/en/mongodb-bson-objectid.gettimestamp.php) for that purpose. Perhaps this would warrant a separate tutorial or cookbook entry, as there's likely a bit more code involved to parse the timestamp from the string (could be done with PHP code separate from `ext-mongodb`).

On a separate note, ObjectIds might also encounter issues in high-concurrency environments (see: [SERVER-6054](https://jira.mongodb.org/browse/SERVER-6054)). The 4-byte timestamp and 3-byte counter (see: [ObjectId spec](https://github.com/mongodb/specifications/blob/master/source/bson-objectid/objectid.md)) likely provides more uniqueness than a UTCDateTime with millisecond precision, though.

jmikola (approved):

docs/en/reference/transactions-and-concurrency.rst
GromNaN: Exact, the `ObjectIdType` convert the `ObjectId` into a `string`. You have to convert it back to `ObjectId` to get the timestamp.
https://github.com/doctrine/mongodb-odm/blob/650b14de3b1afb3280b92b519bdaccd30cc98426/lib/Doctrine/ODM/MongoDB/Types/ObjectIdType.php#L29

> ObjectIds might also encounter issues in high-concurrency environments
A max of 16 millions unique ObjectId per second should be more than what a single PHP process can handle.

PR #2816, opened 2025-09-01
Fix duplicate slug in .doctrine-project.json

GromNaN:
Currently https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/latest/index.html shows the version 2.13 of the documentation. There is no way to access 2.12.


SenseException (approved):

SenseException (changes requested):
You need to target `2.13.x` because it is the current default branch

PR #2820, opened 2025-09-05
PHPORM-381 Add class metadata for vector search indexes

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | [PHPORM-381](https://jira.mongodb.org/browse/PHPORM-381)

#### Summary

Add a new `#[ODM\VectorSearchIndex]` attribute to create [Atlas Vector Search Indexes](https://www.mongodb.com/docs/atlas/atlas-vector-search/vector-search-type/?deployment-type=atlas&interface=driver&language=php).

In XML, attribute and annotations, this is a new type of object. But they create Search Indexes in the class metadata with the type `vectorSearch`. The index management is exactly the same for search and vector search indexes.

### DIFF
--- a/docs/en/reference/attributes-reference.rst
+++ b/docs/en/reference/attributes-reference.rst
@@ -1462,7 +1462,7 @@ Example:
         #[Field(type: Type::COLLECTION)]
         public array $plotEmbeddingVoyage3Large = [];
 
-        #[Field)]
+        #[Field]
         public string $category;
     }
 
--- a/docs/en/reference/attributes-reference.rst
+++ b/docs/en/reference/attributes-reference.rst
@@ -1167,6 +1167,10 @@ Optional arguments:
 This attribute is used to specify :ref:`search indexes <search_indexes>` for
 `MongoDB Atlas Search <https://www.mongodb.com/docs/atlas/atlas-search/>`__.
 
+.. note::
+
+    For vector search indexes, see :ref:`vector_search_index` below.
+
 The arguments correspond to arguments for
 `MongoDB\Collection::createSearchIndex() <https://www.mongodb.com/docs/php-library/current/reference/method/MongoDBCollection-createSearchIndex/>`__.
 Excluding ``name``, arguments are used to create the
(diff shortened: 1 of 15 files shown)


lib/Doctrine/ODM/MongoDB/Mapping/Annotations/VectorSearchIndex.php
```diff
@@ -0,0 +1,27 @@
...
+ * @Annotation
+ * @NamedArgumentConstructor
+ * @phpstan-import-type VectorSearchIndexField from ClassMetadata
+ */
+#[Attribute(Attribute::TARGET_CLASS | Attribute::IS_REPEATABLE)]
+class VectorSearchIndex implements Annotation
+{
+    /** @param list<VectorSearchIndexField> $fields */
+    public function __construct(
+        public array $fields,
```
GromNaN, line 25: Since the fields are required, it has to be the 1st parameter of the constructor.

lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
```diff
@@ -748,6 +754,45 @@ private function getSearchIndexFieldDefinition(SimpleXMLElement $field): array
...
+            if (isset($vectorField['quantization'])) {
+                $field['quantization'] = (string) $vectorField['quantization'];
+            }
+
+            if (isset($vectorField['hnswMaxEdges'])) {
+                $field['hnswOptions']['maxEdges'] = (int) $vectorField['hnswMaxEdges'];
+            }
+
+            if (isset($vectorField['hnswNumEdgeCandidates'])) {
+                $field['hnswOptions']['numEdgeCandidates'] = (int) $vectorField['hnswNumEdgeCandidates'];
```
paulinevos, line 778: Just for my information, when/how do the types specified in the XML schema get checked? I see that the schema defines the types, but does that get validated somewhere before we get here and cast them?

paulinevos (approved):

paulinevos:
Please hold, I accepted and realized I'd only reviewed one commit. Can't remove the review 💀

paulinevos (approved):
All good now as far as I'm concerned

lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php
GromNaN: The type is only used when validating the XML file with the XSD. It's never used to cast the node to the correct type.
https://github.com/doctrine/mongodb-odm/blob/e3352c0783886e525d2b8accb8bc7984148017bf/lib/Doctrine/ODM/MongoDB/Mapping/Driver/XmlDriver.php#L867

paulinevos: Sorry, I mean it's cast here on line 778 so I was wondering if we're making sure somewhere else that it actually is a number. But ok cool, so it is validated!

docs/en/reference/attributes-reference.rst
```diff
@@ -1397,6 +1401,73 @@ for the related collection.
...
+on a document class. This enables efficient similarity search on vector fields,
+such as those used for machine learning embeddings.
+
+Optional arguments:
+
+- ``name``: (optional) The name of the vector search index. If omitted, a default name is used.
+- ``fields`` (required): A list of field definitions. Each field definition is an associative array describing a vector or filter field. For vector fields, the following keys are supported:
+
+  - ``type``: Must be set to ``'vector'`` for vector fields or ``'filter'`` for filter fields.
+  - ``path``: The name of the field in the document to index.
```
GromNaN, line 1419: Should it use the class metadata to map the doctrine field name to the mongodb field name?

docs/en/reference/attributes-reference.rst
```diff
@@ -1397,6 +1401,73 @@ for the related collection.
...
+  - ``numDimensions``: (vector fields only) The number of dimensions in the vector.
+  - ``similarity``: (vector fields only) The vector similarity function to use. Supported values include ``'euclidean'``, ``'cosine'``, and ``'dotProduct'``. Use the constants from ``Doctrine\ODM\MongoDB\Mapping\ClassMetadata::VECTOR_SIMILARITY_*`` for best compatibility.
```
jmikola, line 1421: Any reason you chose not to use an enum here? I suppose keeping this open makes it easier to be forward-compatible should the server introduce more similarity types.

docs/en/reference/attributes-reference.rst
jmikola: IIRC, we do not do that for the [`fields` option on SearchIndex](https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/latest/reference/search-indexes.html#search-index-options). The values here should correspond to the exact names in the database schema.

This may be something worth clarifying in the documentation files for both, though.

docs/en/reference/attributes-reference.rst
GromNaN: Yes, I want to keep it open for new values, and let people use strings from the documentation.

docs/en/reference/attributes-reference.rst
jmikola: Tracked in [PHPORM-391](https://jira.mongodb.org/browse/PHPORM-391).

PR #2822, opened 2025-09-08
PHPORM-382 Add $vectorSearch stage to the aggregation builder

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | [PHPORM-382](https://jira.mongodb.org/browse/PHPORM-382)

#### Summary

Add the [`$vectorSearch`](https://www.mongodb.com/docs/atlas/atlas-vector-search/vector-search-stage/) stage builder.

```php
$dm->getRepository(Face::class)
   ->createAggregationBuilder()
   ->hydrate(VectorSearchResult::class)
   ->vectorSearch()
       ->index('descriptions')
       ->path('descriptionEmbeddings')
       ->filter(
           $builder->matchExpr()
               ->field('id')
               ->notEqual($face->id)
       )
       ->numCandidates($limit * 20)
       ->queryVector($face->descriptionEmbeddings)
       ->limit($limit)
   ->project()
       ->field('_id')->expression(0)
       ->field('face')->expression('$$ROOT')
       ->field('score')->meta('vectorSearchScore');
```

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
@@ -653,6 +653,19 @@ class Builder
     }
 
     /**
+     * The $vectorSearch stage performs a vector similarity search on the specified
+     * field which must be covered by an Atlas Vector Search index.
+     *
+     * @see https://www.mongodb.com/docs/atlas/atlas-vector-search/vector-search-stage/#mongodb-pipeline-pipe.-vectorSearch
+     */
+    public function vectorSearch(): Stage\VectorSearch
+    {
+        $stage = new Stage\VectorSearch($this);
+
+        return $this->addStage($stage);
+    }
+
+    /**
      * Performs a union of two collections. $unionWith combines pipeline results
      * from two collections into a single result set. The stage outputs the
      * combined result set (including duplicates) to the next stage.
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/VectorSearch.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/VectorSearch.php
@@ -7,11 +7,12 @@ namespace Doctrine\ODM\MongoDB\Aggregation\Stage;
 use Doctrine\ODM\MongoDB\Aggregation\Builder;
 use Doctrine\ODM\MongoDB\Aggregation\Stage;
 use Doctrine\ODM\MongoDB\Query\Expr;
+use MongoDB\BSON\Binary;
 use MongoDB\BSON\Decimal128;
 use MongoDB\BSON\Int64;
 
 /**
- * @phpstan-type Vector list<int|Int64>|list<float|Decimal128>|list<bool|0|1>
+ * @phpstan-type Vector list<int|Int64>|list<float|Decimal128>|list<bool|0|1>|Binary
  * @phpstan-type VectorSearchStageExpression array{
  *     '$vectorSearch': object{
  *         exact?: bool,
@@ -33,7 +34,7 @@ class VectorSearch extends Stage
     private ?int $numCandidates = null;
     private ?string $path       = null;
     /** @phpstan-var Vector|null */
-    private ?array $queryVector = null;
+    private array|Binary|null $queryVector = null;
 
     public function __construct(Builder $builder)
     {
@@ -118,7 +119,7 @@ class VectorSearch extends Stage
     }
 
     /** @phpstan-param Vector $queryVector */
-    public function queryVector(array $queryVector): static
+    public function queryVector(array|Binary $queryVector): static
     {
         $this->queryVector = $queryVector;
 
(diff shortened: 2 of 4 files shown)


lib/Doctrine/ODM/MongoDB/Aggregation/Stage/VectorSearch.php
```diff
@@ -0,0 +1,133 @@
...
+
+use Doctrine\ODM\MongoDB\Aggregation\Builder;
+use Doctrine\ODM\MongoDB\Aggregation\Stage;
+use Doctrine\ODM\MongoDB\Query\Expr;
+use MongoDB\BSON\Binary;
+use MongoDB\BSON\Decimal128;
+use MongoDB\BSON\Int64;
+
+/**
+ * @phpstan-type Vector list<int|Int64>|list<float|Decimal128>|list<bool|0|1>
```
GromNaN: Should I accept `MongoDB\BSON\Binary` with ["vector" type](https://www.mongodb.com/docs/manual/reference/bson-types/#binary-data)? Or wait for the release of 
- https://github.com/mongodb/mongo-php-driver/pull/1853.

For the method `queryVector`, adding a type to the argument is breaking change. From `array` to `array|Binary`. 

[PHPORM-387](https://jira.mongodb.org/browse/PHPORM-387)

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/VectorSearch.php
```diff
@@ -0,0 +1,133 @@
...
+            $params['numCandidates'] = $this->numCandidates;
+        }
+
+        if ($this->path !== null) {
+            $params['path'] = $this->path;
+        }
+
+        if ($this->queryVector !== null) {
+            $params['queryVector'] = $this->queryVector;
+        }
```
GromNaN, line 74: `index`, `limit`, `path` and `queryVector` are required. The server will return an error if they are not used.
Looking at the other stages, it seems that we don't check this before.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/VectorSearch.php
```diff
@@ -0,0 +1,133 @@
...
+            $params['limit'] = $this->limit;
+        }
+
+        if ($this->numCandidates !== null) {
+            $params['numCandidates'] = $this->numCandidates;
+        }
+
+        if ($this->path !== null) {
+            $params['path'] = $this->path;
+        }
```
GromNaN, line 70: The `path` should be mapped to the field name using the class metadata.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/VectorSearch.php
paulinevos: Can you explain why it would be a breaking change? I think I don't understand

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/VectorSearch.php
paulinevos: Makes sense to me. Maybe it's nice to document this as a standard in some sort of development decision documentation in the repo, as I think we've had conversations about this before.

paulinevos (approved):

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/VectorSearch.php
GromNaN: This class is not final (should it be?). So if a subclass override this method, it cannot restrict the accepted type of the parameter of the `queryVector(array $queryVector)` method. So we cannot add a type to the union type of the parameter.
Changing an argument type is not allowed by [Symfony BC policy](https://symfony.com/doc/current/contributing/code/bc.html#changing-classes).

GromNaN: I added `Binary` to the union, so we don't have the break things.

lib/Doctrine/ODM/MongoDB/Aggregation/Builder.php
```diff
@@ -652,6 +652,19 @@ public function sortByCount(string $expression): Stage\SortByCount
         return $this->addStage($stage);
     }
 
+    /**
+     * The $vectorSearch stage performs a vector similarity search on the specified
+     * field which must be covered by an Atlas Vector Search index.
+     *
+     * @see https://www.mongodb.com/docs/atlas/atlas-vector-search/vector-search-stage/#mongodb-pipeline-pipe.-vectorSearch
+     */
+    public function vectorSearch(): Stage\VectorSearch
```
jmikola, line 661: IIUC, this PR _does not_ add a `Stage::vectorSearch()` method for the same reason that you deprecated `Stage::search()` in https://github.com/doctrine/mongodb-odm/pull/2823, correct?

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/VectorSearch.php
jmikola: Is this a TODO item? If so, is there another ticket to track this?

This seems related to https://github.com/doctrine/mongodb-odm/pull/2820#discussion_r2337830958 from the PR that introduced a `VectorSearchIndex` mapping.

PR #2823, opened 2025-09-09
Deprecate `Stage::search()` because the `$search` stage must be the first of the pipeline

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | yes
| Fixed issues | -

#### Summary

> [$search](https://www.mongodb.com/docs/atlas/atlas-search/aggregation-stages/search/#mongodb-pipeline-pipe.-search) must be the first stage of any pipeline it appears in.

And mentioned in the ODM doc: https://github.com/doctrine/mongodb-odm/blob/e3352c0783886e525d2b8accb8bc7984148017bf/docs/en/reference/aggregation-stage-reference.rst?plain=1#L679-L685

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Stage.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Stage.php
@@ -368,10 +368,14 @@ abstract class Stage
      * The $search stage performs a full-text search on the specified field or
      * fields which must be covered by an Atlas Search index.
      *
+     * @deprecated Since doctrine/mongodb-odm 2.13. This $search stage must be the first of the pipeline, use Builder::search() instead.
+     *
      * @see https://www.mongodb.com/docs/atlas/atlas-search/query-syntax/#mongodb-pipeline-pipe.-search
      */
     public function search(): Stage\Search
     {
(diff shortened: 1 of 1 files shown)


lib/Doctrine/ODM/MongoDB/Aggregation/Stage.php
```diff
@@ -368,10 +368,14 @@ public function sample(int $size): Stage\Sample
      * The $search stage performs a full-text search on the specified field or
      * fields which must be covered by an Atlas Search index.
      *
+     * @deprecated Since doctrine/mongodb-odm 2.13. This $search stage must be the first of the pipeline, use Builder::search() instead.
```
jmikola, line 371: I realize this is ODM and not PHPLIB's builder, but is there not a use case for creating your own pipeline with a list of `Stage` objects where it might still make sense to use a `$search` Stage?

jmikola (approved):
One question just for my understanding, but I assume you know what you're doing here 😄

lib/Doctrine/ODM/MongoDB/Aggregation/Stage.php
GromNaN: The `Builder` is a required dependency of the `Stage` classes. You cannot create stages without builder. Calling `->search` from an existing stage means that you append the search stage after the current stage, which is never allowed.

PR #2826, opened 2025-09-16
Support symfony UUID

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | yes
| Fixed issues | 

#### Summary

This PR adds a new type to map Symfony UUID instances to a BSON Binary with subtype UUID. In addition, the existing `AutoGenerator` is replaced by separate `ObjectIdGenerator` and `UUIDGenerator` to support auto-generating new object IDs.

TODO:
- [x] Add new type for UUIDs
- [x] Refactor ID generation to allow generating UUIDs using the `AUTO` strategy.
- [x] Document new type and generator options

### DIFF
--- a/docs/en/reference/basic-mapping.rst
+++ b/docs/en/reference/basic-mapping.rst
@@ -273,7 +273,7 @@ object ID. The available strategies are:
 - ``ALNUM`` - Generates an alpha-numeric string (based on an incrementing value).
 - ``CUSTOM`` - Defers generation to an implementation of ``IdGenerator`` specified in the ``class`` option.
 - ``INCREMENT`` - Uses another collection to auto increment an integer identifier.
-- ``UUID`` - Generates a UUID identifier.
+- ``UUID`` - Generates a UUID identifier (deprecated).
 - ``NONE`` - Do not generate any identifier. ID must be manually set.
 
 Here is an example how to manually set a string identifier for your documents:
--- a/docs/en/reference/basic-mapping.rst
+++ b/docs/en/reference/basic-mapping.rst
@@ -179,7 +179,7 @@ This list explains some of the less obvious mapping types:
 -  ``id``: string to ObjectId by default, but other formats are possible
 -  ``timestamp``: string to ``MongoDB\BSON\Timestamp``
 -  ``raw``: any type
--  ``uuid``: Symfony UUID to MongoDB\BSON\Binary instance with a "uuid" type
+-  ``uuid``: `Symfony UID <https://symfony.com/doc/current/components/uid.html>`_ to ``MongoDB\BSON\Binary`` instance with a "uuid" type
 
 .. note::
 
--- a/docs/en/reference/basic-mapping.rst
+++ b/docs/en/reference/basic-mapping.rst
@@ -151,6 +151,7 @@ Here is a quick overview of the built-in mapping types:
 -  ``raw``
 -  ``string``
 -  ``timestamp``
+-  ``uuid``
 
 You can read more about the available MongoDB types on `php.net <https://www.php.net/mongodb.bson>`_.
 
@@ -178,6 +179,7 @@ This list explains some of the less obvious mapping types:
 -  ``id``: string to ObjectId by default, but other formats are possible
 -  ``timestamp``: string to ``MongoDB\BSON\Timestamp``
 -  ``raw``: any type
+-  ``uuid``: Symfony UUID to MongoDB\BSON\Binary instance with a "uuid" type
 
 .. note::
 
@@ -206,6 +208,7 @@ follows:
 - ``float``: ``float``
 - ``int``: ``int``
 - ``string``: ``string``
+- ``Symfony\Component\Uid\Uuid``: ``uuid``
 
 Doctrine can also autoconfigure any backed ``enum`` it encounters: ``type``
 will be set to ``string`` or ``int``, depending on the enum's backing type,
@@ -269,12 +272,16 @@ Here is an example:
 You can configure custom ID strategies if you don't want to use the default
 object ID. The available strategies are:
 
-- ``AUTO`` - Uses the native generated ObjectId.
+- ``AUTO`` - Automatically generates an ObjectId or Symfony UUID depending on the identifier type.
 - ``ALNUM`` - Generates an alpha-numeric string (based on an incrementing value).
 - ``CUSTOM`` - Defers generation to an implementation of ``IdGenerator`` specified in the ``class`` option.
 - ``INCREMENT`` - Uses another collection to auto increment an integer identifier.
 - ``UUID`` - Generates a UUID identifier (deprecated).
 - ``NONE`` - Do not generate any identifier. ID must be manually set.
+
+When using the ``AUTO`` strategy in combination with a UUID identifier, the generator can create UUIDs of type 1, type 4,
+and type 7 automatically. For all other UUID types, assign the identifier manually in combination with the ``NONE``
+strategy.
 
 Here is an example how to manually set a string identifier for your documents:
 
--- a/docs/en/reference/basic-mapping.rst
+++ b/docs/en/reference/basic-mapping.rst
@@ -283,6 +283,12 @@ When using the ``AUTO`` strategy in combination with a UUID identifier, the gene
 and type 7 automatically. For all other UUID types, assign the identifier manually in combination with the ``NONE``
 strategy.
 
+.. note::
+
+    The ``UUID`` generator is deprecated, as it stores UUIDs as strings. It is recommended to use the ``AUTO`` strategy
+    with a ``uuid`` type identifier field instead. If you need to keep generating string UUIDs, you can use the
+    ``CUSTOM`` strategy with your own generator.
+
 Here is an example how to manually set a string identifier for your documents:
 
 .. configuration-block::
--- a/lib/Doctrine/ODM/MongoDB/Id/AutoGenerator.php
+++ b/lib/Doctrine/ODM/MongoDB/Id/AutoGenerator.php
@@ -10,7 +10,7 @@ use MongoDB\BSON\ObjectId;
 /**
  * AutoGenerator generates a native ObjectId
  *
- * @deprecated
+ * @deprecated use ObjectIdGenerator instead
  */
 final class AutoGenerator extends AbstractIdGenerator
 {
--- a/lib/Doctrine/ODM/MongoDB/Id/ObjectIdGenerator.php
+++ b/lib/Doctrine/ODM/MongoDB/Id/ObjectIdGenerator.php
@@ -7,11 +7,7 @@ namespace Doctrine\ODM\MongoDB\Id;
 use Doctrine\ODM\MongoDB\DocumentManager;
 use MongoDB\BSON\ObjectId;
 
-/**
- * AutoGenerator generates a native ObjectId
- *
- * @internal
- */
+/** @internal */
 final class ObjectIdGenerator extends AbstractIdGenerator
 {
     public function generate(DocumentManager $dm, object $document): ObjectId
--- a/lib/Doctrine/ODM/MongoDB/Id/ObjectIdGenerator.php
+++ b/lib/Doctrine/ODM/MongoDB/Id/ObjectIdGenerator.php
@@ -10,11 +10,11 @@ use MongoDB\BSON\ObjectId;
 /**
  * AutoGenerator generates a native ObjectId
  *
- * @deprecated
+ * @internal
  */
-final class AutoGenerator extends AbstractIdGenerator
+final class ObjectIdGenerator extends AbstractIdGenerator
 {
-    public function generate(DocumentManager $dm, object $document)
+    public function generate(DocumentManager $dm, object $document): ObjectId
     {
         return new ObjectId();
     }
--- a/lib/Doctrine/ODM/MongoDB/Id/SymfonyUuidGenerator.php
+++ b/lib/Doctrine/ODM/MongoDB/Id/SymfonyUuidGenerator.php
@@ -1,0 +1,39 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ODM\MongoDB\Id;
+
+use Doctrine\ODM\MongoDB\DocumentManager;
+use InvalidArgumentException;
+use Symfony\Component\Uid\Uuid;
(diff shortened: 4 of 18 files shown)


lib/Doctrine/ODM/MongoDB/Id/AutoGenerator.php
```diff
@@ -9,6 +9,8 @@
 
 /**
  * AutoGenerator generates a native ObjectId
+ *
+ * @deprecated
```
GromNaN: We try to add helpful details.
```suggestion
 * @deprecated use ObjectIdGenerator instead.
```

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -2391,22 +2406,18 @@ public function mapField(array $mapping): array
...
-                    $mapping['type'] = 'id';
-                    break;
-                default:
-                    if (! empty($this->generatorOptions['type'])) {
-                        $mapping['type'] = (string) $this->generatorOptions['type'];
-                    } elseif (empty($mapping['type'])) {
-                        $mapping['type'] = $this->generatorType === self::GENERATOR_TYPE_INCREMENT ? Type::INT : Type::CUSTOMID;
-                    }
+            if ($this->generatorType !== self::GENERATOR_TYPE_AUTO) {
+                if (! empty($this->generatorOptions['type'])) {
```
GromNaN, line 2410:
```suggestion
                if (! $this->generatorOptions['type'] ?? null) {
```

lib/Doctrine/ODM/MongoDB/Types/BinaryUuidType.php
```diff
@@ -0,0 +1,67 @@
...
+    {
+        return <<<'PHP'
+$return = match (true) {
+    $value === null => null,
+    $value instanceof \MongoDB\BSON\Binary => $value,
+    $value instanceof \Symfony\Component\Uid\Uuid => new \MongoDB\BSON\Binary($value->toBinary(), \MongoDB\BSON\Binary::TYPE_UUID),
+    is_string($value) => new \MongoDB\BSON\Binary(\Symfony\Component\Uid\Uuid::fromString($value)->toBinary(), \MongoDB\BSON\Binary::TYPE_UUID),
+    default => throw new InvalidArgumentException(sprintf('Invalid data type %s received for UUID', get_debug_type($value))),
+};
+PHP;
```
GromNaN, line 55: Why not using the same code as in `convertToDatabaseValue`?

lib/Doctrine/ODM/MongoDB/Id/UuidGenerator.php
```diff
@@ -20,6 +20,8 @@
 
 /**
  * Generates UUIDs.
+ *
+ * @deprecated
  */
 final class UuidGenerator extends AbstractIdGenerator
```
GromNaN: This deprecates generating a UUID value in a string field, as the new `SymfonyUuidGenerator` only works with property with a `Symfony\Component\Uid\Uuid` type. This can be a hard breaking change as it impacts the schema of the documents in the database.

lib/Doctrine/ODM/MongoDB/Id/AutoGenerator.php
alcaeus: Updated

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
alcaeus: Updated

lib/Doctrine/ODM/MongoDB/Types/BinaryUuidType.php
alcaeus: Updated logic to match

lib/Doctrine/ODM/MongoDB/Id/UuidGenerator.php
alcaeus: Discussed in person: we'll deprecate this, but we'll amend the deprecation notices and documentation to explain the reasoning and show alternatives.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
alcaeus: Reverted because of test failures.

lib/Doctrine/ODM/MongoDB/Id/UuidGenerator.php
alcaeus: Updated the deprecation messages and added an entry to the 3.0 upgrade document about the deprecation.

docs/en/reference/basic-mapping.rst
```diff
@@ -178,6 +179,7 @@ This list explains some of the less obvious mapping types:
 -  ``id``: string to ObjectId by default, but other formats are possible
 -  ``timestamp``: string to ``MongoDB\BSON\Timestamp``
 -  ``raw``: any type
+-  ``uuid``: Symfony UUID to MongoDB\BSON\Binary instance with a "uuid" type
```
GromNaN:
```suggestion
-  ``uuid``: Symfony UUID to ``MongoDB\BSON\Binary`` instance with a "uuid" type
```

Should you add link to https://symfony.com/doc/current/components/uid.html?

docs/en/reference/basic-mapping.rst
```diff
@@ -269,13 +272,17 @@ Here is an example:
 You can configure custom ID strategies if you don't want to use the default
 object ID. The available strategies are:
 
-- ``AUTO`` - Uses the native generated ObjectId.
+- ``AUTO`` - Automatically generates an ObjectId or Symfony UUID depending on the identifier type.
 - ``ALNUM`` - Generates an alpha-numeric string (based on an incrementing value).
 - ``CUSTOM`` - Defers generation to an implementation of ``IdGenerator`` specified in the ``class`` option.
 - ``INCREMENT`` - Uses another collection to auto increment an integer identifier.
-- ``UUID`` - Generates a UUID identifier.
+- ``UUID`` - Generates a UUID identifier (deprecated).
```
GromNaN, line 279: A precision would highlight why it's deprecated.
```suggestion
- ``UUID`` - Generates a UUID identifier as string (deprecated).
```

lib/Doctrine/ODM/MongoDB/Id/ObjectIdGenerator.php
```diff
@@ -0,0 +1,21 @@
...
+namespace Doctrine\ODM\MongoDB\Id;
+
+use Doctrine\ODM\MongoDB\DocumentManager;
+use MongoDB\BSON\ObjectId;
+
+/**
+ * AutoGenerator generates a native ObjectId
+ *
+ * @internal
+ */
```
GromNaN: This comment isn't necessary, and the class is internal. Consistent with other classes.
```suggestion
/** @internal */
```

lib/Doctrine/ODM/MongoDB/Id/SymfonyUuidGenerator.php
```diff
@@ -0,0 +1,39 @@
...
+        4 => UuidV4::class,
+        7 => UuidV7::class,
+    ];
+
+    public function __construct(private readonly string $class)
+    {
+        if (! in_array($this->class, self::SUPPORTED_TYPES, true)) {
+            throw new InvalidArgumentException(sprintf('Invalid UUID type "%s". Expected one of: %s.', $this->class, implode(', ', array_values(self::SUPPORTED_TYPES))));
+        }
+    }
```
GromNaN, line 33: The Symfony UuidFactory accepts a class name or a version number, you can do the same: https://github.com/symfony/symfony/blob/a30f387e11325eb4a6db35765ed8b22effc59730/src/Symfony/Component/Uid/Factory/UuidFactory.php#L30-L43

GromNaN: At the same time, the class is internal, we are not looking to improve its DX.

tests/Doctrine/ODM/MongoDB/Tests/Functional/IdTest.php
```diff
@@ -345,18 +345,6 @@ public function testStrategyNoneAndNoIdThrowsException(): void
...
-        $user     = new TestIdTypesIdAutoUser();
-        $user->id = 1;
-        $this->expectException(InvalidArgumentException::class);
-        $this->expectExceptionMessage(
-            'Doctrine\ODM\MongoDB\Tests\Functional\TestIdTypesIdAutoUser uses AUTO identifier generation strategy ' .
-            'but provided identifier is not a valid ObjectId.',
-        );
-        $this->dm->persist($user);
-    }
-
```
GromNaN, line 359: Is it still possible to get this error? Or is it dead code?

docs/en/reference/basic-mapping.rst
alcaeus: Good point, added.

docs/en/reference/basic-mapping.rst
alcaeus: Added a note to explain the deprecation

lib/Doctrine/ODM/MongoDB/Id/ObjectIdGenerator.php
alcaeus: Done

tests/Doctrine/ODM/MongoDB/Tests/Functional/IdTest.php
alcaeus: I had initially removed the error but eventually put it back in, except with a tiny mistake. I've re-added the test and fixed the issue.

lib/Doctrine/ODM/MongoDB/Types/BinaryUuidType.php
```diff
@@ -0,0 +1,78 @@
...
+
+        if (! $value instanceof Binary) {
+            throw new Exception(sprintf('Invalid data of type "%s" received for Uuid', get_debug_type($value)));
+        }
+
+        if ($value->getType() !== Binary::TYPE_UUID) {
+            throw new Exception(sprintf('Invalid binary data of type %d received for Uuid', $value->getType()));
+        }
+
+        return Uuid::fromBinary($value->getData());
```
GromNaN, line 42: I checked that this returns instances of `UuidVx`.
If there is a miss-match between the UUID version of the value in the database, and the document property type, the user will get a type error, which is perfect.

lib/Doctrine/ODM/MongoDB/Types/BinaryUuidType.php
```diff
@@ -0,0 +1,78 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ODM\MongoDB\Types;
+
+use Exception;
```
GromNaN:
```suggestion
use Doctrine\ODM\MongoDB\MongoDBException;
```

lib/Doctrine/ODM/MongoDB/Types/BinaryUuidType.php
```diff
@@ -0,0 +1,78 @@
...
+        if ($value instanceof Uuid) {
+            return $value;
+        }
+
+        if (! $value instanceof Binary) {
+            throw new Exception(sprintf('Invalid data of type "%s" received for Uuid', get_debug_type($value)));
+        }
+
+        if ($value->getType() !== Binary::TYPE_UUID) {
+            throw new Exception(sprintf('Invalid binary data of type %d received for Uuid', $value->getType()));
```
GromNaN: Having a more specific exception make it catchable.
```suggestion
            throw MongoDBException::invalidValueForType(sprintf('Invalid binary data of type %d received for Uuid', $value->getType()));
```

phpstan-baseline.neon
```diff
@@ -2106,6 +2112,18 @@ parameters:
...
+			identifier: staticMethod.impossibleType
+			count: 1
+			path: tests/Doctrine/ODM/MongoDB/Tests/Types/BinaryUuidTypeTest.php
+
+		-
+			message: '#^Call to static method PHPUnit\\Framework\\Assert\:\:assertSame\(\) with arguments Symfony\\Component\\Uid\\UuidV4, null and ''Uuid objects are…'' will always evaluate to false\.$#'
+			identifier: staticMethod.impossibleType
+			count: 1
+			path: tests/Doctrine/ODM/MongoDB/Tests/Types/BinaryUuidTypeTest.php
+
```
GromNaN, line 2126: Which lines are this errors. Is it a false-positive due to the way the PHP extension can by-pass userland rules?

alcaeus: This is from the `testClosureToMongo` tests where PHPStan gets confused due to using `eval` in the test to evaluate the code returned from the method. It reads the return value of the closure as `null` as it doesn't know what's being run in `eval` and generates these false positives on the following assertions.

lib/Doctrine/ODM/MongoDB/Types/BinaryUuidType.php
alcaeus: Replaced this with `InvalidArgumentException` which is used in all other places and types.

GromNaN: I still believe we should have an exception in the `Doctrine\ODM\MongoDB\` namespace for this kind of error, which can occur at runtime when the data in the database is incorrect.

GromNaN (approved):

lib/Doctrine/ODM/MongoDB/Types/BinaryUuidType.php
alcaeus: Agree. Not sure if it should be a `MappingException` or if there should be a different class, but we should revisit this for a future version.

GromNaN: Not a mapping error. Maybe `Doctrine\ODM\MongoDB\Types\ValueNotConvertible`. That would match `Doctrine\DBAL\Types\Exception\ValueNotConvertible`: https://github.com/doctrine/dbal/blob/2ce1218ef35cad74e488f7213df6dda701316569/src/Types/BinaryType.php#L39-L40

PR #2827, opened 2025-09-18
Support dot syntax when preparing nested query values

### LINKED ISSUE #2212: Custom types are not correctly transformered for embedded documents
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.1.1

#### Summary

Let's say we have `User` document with embedded document `Identities`. Some of fields in `Identities` has to be encrypted in the database. To achieve it, the new type was registered and assigned to the appropriate fields.
In the repository class we need to filter by the encrypted field while in code we have it decrypted. 
So query builder is used this way:

```
$passportNumber = 'FB555555';
$qb = $this->dm->createQueryBuilder(User::class);
$qb->field('identifiers.passportNumber')->equals(new PassportNumber($passportNumber));
$qb->getQuery()->toArray();
```
But the query doesn't find anything because `PassportNumberType::convertToDatabaseValue` is never called. 

#### Current behavior

PassportNumber is not encrypted because `PassportNumberType::convertToDatabaseValue` is never called.
It is never called, because `\Doctrine\ODM\MongoDB\Persisters\DocumentPersister::convertToDatabaseValue` can't see the field in the class as this line of code `$this->class->hasField($fieldName)` returns `false`.

#### How to reproduce

Here is a small code samples (namespaces, use statement and other unnecessary code is omitted):

```
/**
 * @MongoDB\Document()
 */
class User
{
    /**
     * @MongoDB\Id
     */
    private $id;

    /**
     * @var Identifiers|null
     * @MongoDB\EmbedOne(targetDocument="App\Document\User\Identities")
     */
    private $identities;
}

/**
 * @MongoDB\EmbeddedDocument
 */
class Identifiers
{
    /**
     * @MongoDB\Field(type="passport_number", name="passport_number")
     */
    private PassportNumber $passportNumber;
}



class PassportNumberType extends Type
{
    public function convertToPHPValue($value): PassportNumber
    {
        return new PassportNumber(str_replace('encrypted_', 'decrypted_', $value));
    }

    public function convertToDatabaseValue($value): string
    {
        /** @var $value PassportNumber */
        return 'encrypted_' . $value->value();
    }
}

class PassportNumber
{
    private ?string $id;

    public function __construct(?string $id)
    {
        $this->id = $id;
    }

    public function value(): ?string
    {
        return $this->id;
    }
}
```

#### Expected behavior

The value for the custom type in the embedded documents should be processed in the way it is defined in the custom type definition.

### LINKED ISSUE #2531: Query built incorrectly when having embedMany with discriminatorMap and non BSON id field
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.5.2

#### Summary

I created a combination of 5 classes Tree, Branch, Fruit, Apple and Mango. Apple and Mango extend Frui.
A Tree embeds many branches and a branch embeds many fruits using a discriminator field and discriminator map.
Fruit also has an id property which is not a BSON id but some other string (in this example UUID V4) which was generated from outside.

#### Current behavior

Writing a query to find a branch by a fruit id fails, as the query builder converts the value for the id field into a BSON Object id.
```
$this->createQueryBuilder()
            ->field('branches.fruits.id')->equals($fruitId)
            ->getQuery();
```
image

This is what it looks like when also specifying the targetDocument for `Branch.fruits` to be `Fruit::class` which has been deprecated.
image

#### How to reproduce

I provided an example here:  https://github.com/xammmue/embedManyWithNonBSONId
But these are the essentials:

The bug only occurs when there is a "chain" of at least two embedded documents.
```
#[Document]
class Tree
{
    #[Id]
    private $id;
    #[EmbedMany(targetDocument: Branch::class)]
    private Collection $branches;

    public function __construct(array $branches)
    {
        $this->branches = new ArrayCollection($branches);
    }

    public function getId()
    {
        return $this->id;
    }

    public function getBranches(): array
    {
        return $this->branches->toArray();
    }
}
```
```
#[EmbeddedDocument]
class Branch
{
    #[
        EmbedMany(
            //targetDocument: Fruit::class,
            discriminatorField: Fruit::DISCRIMINATOR_FIELD,
            discriminatorMap: Fruit::DISCRIMINATOR_MAP
        )
    ]
    private Collection $fruits;

    public function __construct(array $fruits)
    {
        $this->fruits = new ArrayCollection($fruits);
    }

    public function getFruits(): Collection
    {
        return $this->fruits;
    }
}
```
```
#[EmbeddedDocument]
class Fruit
{
    public const DISCRIMINATOR_FIELD = 'type';
    public const DISCRIMINATOR_MAP = [
        'Apple' => Apple::class,
        'Mango' => Mango::class
    ];

    #[Field]
    private string $id;

    #[Field]
    private bool $ripe;

    public function __construct(string $id, bool $ripe)
    {
        $this->id = $id;
        $this->ripe = $ripe;
    }

    public function getId(): string
    {
        return $this->id;
    }

    public function isRipe(): bool
    {
        return $this->ripe;
    }
}
```

```
class TreeRepository extends ServiceDocumentRepository
{
    public function __construct(ManagerRegistry $registry)
    {
        parent::__construct($registry, Tree::class);
    }

    public function findByFruitId(string $fruitId): ?Tree
    {
        return $this->createQueryBuilder()
            ->field('branches.fruits.id')->equals($fruitId)
            ->getQuery()
            ->getSingleResult();
    }
}
```

#### Expected behavior

I expected the correct Tree to be returned / the correct query to be built.

### LINKED ISSUE #2550: QueryBuilder is replacing "id" by "_id" when filtering on a hash field
|    Q        |   A
|------------ | -----
| Version     | 2.5.2

### Support Question

I am using this lib with [DoctrineMongoDBBundle](https://github.com/doctrine/DoctrineMongoDBBundle).

I store Stripe subscription into my User document. Users has a `subscription` embed document which has a `stripeSubscription` hash, which is the json provided by Stripe

Here is my mapping:

```
#[MongoDB\Document(collection: 'users', repositoryClass: UserRepository::class)]
class User
{
    #[MongoDB\EmbedOne(targetDocument: Subscription::class)]
    protected Subscription $subscription;
```

```
#[MongoDB\EmbeddedDocument]
class Subscription
{
    #[MongoDB\Field(type: Type::HASH, nullable: true)]
    protected null|array $stripeSubscription;
```

Here is a sample of my User document:

```
{
    _id: '1e4dc99cccd03b0898029756790625a4',
   subscription: {
        stripeSubscription: {
            id: 'sub_123',
            object: 'subscription',
            ...
        }
   }
}
```

I want to fetch users by `subscription.stripeSubscription.id` so I did:

```
        $subscriptionId="sub_123";
        $queryBuilder = $this->createQueryBuilder();
        $queryBuilder->field("subscription.stripeSubscription.id")->equals($subscriptionId);

        return $queryBuilder
            ->getQuery()
            ->getSingleResult();
```

But in the profiler, the formatted query is:
```
{
    "find": "users",
    "filter": {
        "subscription.stripeSubscription._id": "sub_123"
    },
    "limit": {
        "$numberInt": "1"
    },
    "$db": "myDb",
    "lsid": {
        "id": {
            "$binary": {
                "base64": "HWzvsCWeTlad1CBKpMSAuw==",
                "subType": "04"
            }
        }
    }
}
```

So my `->field("subscription.stripeSubscription.id")` is converted to `subscription.stripeSubscription._id` with `_id` instead of `id` and it does not find my user.

Am I missing something here ? Is there an option to filter with a field containing `.id` without rewriting to `._id` ?

Edit: I found a workaround, it's not perfect as it does 2 queries (one to find it and the other one to return an hydrated Document) but it does work:

```
$dbName = $this->getDocumentManager()->getConfiguration()->getDefaultDB();
$mongoClient = $this->getDocumentManager()->getClient();
$collection = $mongoClient->selectDatabase($dbName)->selectCollection('users');
$userArray = $collection->findOne(['subscription.stripeSubscription.id' => $subscriptionId]);

$user = $this->find($userArray['_id']);
```

### LINKED ISSUE #2825: updateMany() does not serialize EmbeddedDocument documents in embedded documents fields.
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.12.1

#### Current behavior

`updateMany()` accepts to set `EmbeddedDocument` in fields (field name is `field.target`) but it stores an empty document `{}` in the target field.
It works for root fields (field name is `target`).

I believe this worked in version 2.7.0.

#### How to reproduce

```php

#[ODM\EmbeddedDocument()]
class Embedded {
 public function __construct(
        #[ODM\Field(type: 'string')]
        #[ODM\Index()]
        private string $name,
}
}

#[ODM\EmbeddedDocument()]
class RootEmbedded {
 public function __construct(
        #[ODM\EmbedOne(targetDocument: Embedded::class)]
        private Embedded $embedded;
}
}

#[ODM\Document()]
class MainDocument {
    #[ODM\EmbedOne(targetDocument: RootEmbedded::class)]
        private RootEmbedded $field;

        #[ODM\EmbedOne(targetDocument: Embedded::class)]
        private Embedded $embedded;
}
```

In `MainDocumentRepository`:

```php

        $qb = $this->createQueryBuilder();
        $qb->updateMany()
            ->field('root.embedded.name')->equals($embedded->getName())
            ->field('embedded')->set($embedded)  // works
            ->field('root.embedded')->set($embedded) // doesn't work, store `{}`
            ->getQuery()
            ->execute();
```

#### Expected behavior

`updateMany()` accepts to set `EmbeddedDocument` in fields of embedded documents (when the field name is `field.target`), and stores the full document `{ "name": "value" }` in the target field.

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | Fixes #2825, #2550, #2531, #2212

#### Summary

This fixes an issue when preparing elements for the query builder. The issue was that when `$prepareValue` in `DocumentPersister::prepareQueryElement` was `true`, the method would prepare identifier values and such, but not convert actual field values. Instead, this was done later in `prepareQueryOrNewObj`. However, that class then didn't contain the same dot logic contained in `prepareQueryElement`, leading to incomplete preparation of values.

This fix

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
+++ b/lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
@@ -62,6 +62,7 @@ use function is_scalar;
 use function is_string;
 use function spl_object_id;
 use function sprintf;
+use function str_contains;
 use function strpos;
 use function strtolower;
 use function trigger_deprecation;
@@ -1041,7 +1042,7 @@ final class DocumentPersister
      *
      * PHP field names and types will be converted to those used by MongoDB.
      *
-     * @param array<string, mixed> $query
+     * @param array<string|int, mixed> $query
      *
      * @return array<string, mixed>
      */
@@ -1049,24 +1050,28 @@ final class DocumentPersister
     {
         $preparedQuery = [];
 
-        foreach ($query as $key => $value) {
-            $key = (string) $key;
+        foreach ($query as $field => $value) {
+            $field = (string) $field;
 
-            // Recursively prepare logical query clauses
-            if (in_array($key, ['$and', '$or', '$nor'], true) && is_array($value)) {
-                foreach ($value as $k2 => $v2) {
-                    $preparedQuery[$key][$k2] = $this->prepareQueryOrNewObj($v2, $isNewObj);
-                }
+            // Recursively prepare logical query clauses, treating each value as a separate query element
+            if (in_array($field, ['$and', '$or', '$nor'], true) && is_array($value)) {
+                $preparedQuery[$field] = array_map(
+                    fn ($v) => $this->prepareQueryOrNewObj($v, $isNewObj),
+                    $value,
+                );
 
                 continue;
             }
 
-            if (isset($key[0]) && $key[0] === '$' && is_array($value)) {
-                $preparedQuery[$key] = $this->prepareQueryOrNewObj($value, $isNewObj);
+            // Recursively prepare nested operators, treating the value as a single query element
+            if (isset($field[0]) && $field[0] === '$' && is_array($value)) {
+                $preparedQuery[$field] = $this->prepareQueryOrNewObj($value, $isNewObj);
+
                 continue;
             }
 
-            $preparedQueryElements = $this->prepareQueryElement($key, $value, null, true, $isNewObj);
+            // Prepare a single query element. This may produce multiple queries (e.g. for references)
+            $preparedQueryElements = $this->prepareQueryElement($field, $value, null, true, $isNewObj);
             foreach ($preparedQueryElements as [$preparedKey, $preparedValue]) {
                 $preparedQuery[$preparedKey] = $preparedValue;
             }
@@ -1084,8 +1089,6 @@ final class DocumentPersister
      */
     private function convertToDatabaseValue(string $fieldName, $value, ?ClassMetadata $class = null)
     {
-        $class ??= $this->class;
-
         if (is_array($value)) {
             foreach ($value as $k => $v) {
                 if ($k === '$exists' || $k === '$type' || $k === '$currentDate') {
@@ -1098,7 +1101,7 @@ final class DocumentPersister
             return $value;
         }
 
-        if (! $class->hasField($fieldName)) {
+        if (! $class || ! $class->hasField($fieldName)) {
             if ($value instanceof BackedEnum) {
                 $value = $value->value;
             }
@@ -1133,6 +1136,20 @@ final class DocumentPersister
         return $value;
     }
 
+    private function prepareQueryReference(mixed $value, ClassMetadata $class): mixed
+    {
+        if (! is_array($value)) {
+            return $class->getDatabaseIdentifierValue($value);
+        }
+
+        // Objects without operators or with DBRef fields can be converted immediately
+        if (! $this->hasQueryOperators($value) || $this->hasDBRefFields($value)) {
+            return $class->getDatabaseIdentifierValue($value);
+        }
+
+        return $this->prepareQueryExpression($value, $class);
+    }
+
     /**
      * Prepares a query value and converts the PHP value to the database value
      * if it is an identifier.
@@ -1144,16 +1161,22 @@ final class DocumentPersister
      *
      * @return array<array{string, mixed}>
      */
-    private function prepareQueryElement(string $originalFieldName, $value = null, ?ClassMetadata $class = null, bool $prepareValue = true, bool $inNewObj = false): array
+    private function prepareQueryElement(string $originalFieldName, $value = null, ?ClassMetadata $class = null, bool $prepareValue = true, bool $inNewObj = false, string $fieldNamePrefix = ''): array
     {
-        $class ??= $this->class;
+        $class   ??= $this->class;
+        $fieldName = $fieldNamePrefix . $originalFieldName;
 
-        // @todo Consider inlining calls to ClassMetadata methods
+        // Process identifier fields
+        if (($class->hasField($originalFieldName) && $class->isIdentifier($originalFieldName)) || $originalFieldName === '_id') {
+            $fieldName = $fieldNamePrefix . '_id';
+
+            return [[$fieldName, $prepareValue ? $this->prepareQueryReference($value, $class) : $value]];
+        }
 
         // Process all non-identifier fields by translating field names
-        if ($class->hasField($originalFieldName) && ! $class->isIdentifier($originalFieldName)) {
+        if ($class->hasField($originalFieldName)) {
             $mapping   = $class->fieldMappings[$originalFieldName];
-            $fieldName = $mapping['name'];
+            $fieldName = $fieldNamePrefix . $mapping['name'];
 
             if (! $prepareValue) {
                 return [[$fieldName, $value]];
@@ -1195,29 +1218,9 @@ final class DocumentPersister
             return [[$fieldName, $this->prepareQueryExpression($value, $targetClass)]];
         }
 
-        // Process identifier fields
-        if (($class->hasField($originalFieldName) && $class->isIdentifier($originalFieldName)) || $originalFieldName === '_id') {
-            $fieldName = '_id';
-
-            if (! $prepareValue) {
-                return [[$fieldName, $value]];
-            }
-
-            if (! is_array($value)) {
-                return [[$fieldName, $class->getDatabaseIdentifierValue($value)]];
-            }
-
-            // Objects without operators or with DBRef fields can be converted immediately
-            if (! $this->hasQueryOperators($value) || $this->hasDBRefFields($value)) {
-                return [[$fieldName, $class->getDatabaseIdentifierValue($value)]];
-            }
-
-            return [[$fieldName, $this->prepareQueryExpression($value, $class)]];
-        }
-
         // No processing for unmapped, non-identifier, non-dotted field names
-        if (strpos($originalFieldName, '.') === false) {
-            return [[$originalFieldName, $prepareValue ? $this->convertToDatabaseValue($originalFieldName, $value, $class) : $value]];
+        if (! str_contains($originalFieldName, '.')) {
+            return [[$fieldName, $prepareValue ? $this->convertToDatabaseValue($originalFieldName, $value, $class) : $value]];
         }
 
         /* Process "fieldName.objectProperty" queries (on arrays or objects).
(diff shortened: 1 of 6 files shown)


lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -1143,15 +1144,15 @@ private function convertToDatabaseValue(string $fieldName, $value)
      *
      * @return array<array{string, mixed}>
      */
-    private function prepareQueryElement(string $fieldName, $value = null, ?ClassMetadata $class = null, bool $prepareValue = true, bool $inNewObj = false): array
+    private function prepareQueryElement(string $originalFieldName, $value = null, ?ClassMetadata $class = null, bool $prepareValue = true, bool $inNewObj = false): array
```
GromNaN: Noted that the name was conflicting with the variable `$fieldName` line 1156.

GromNaN (approved):
Perfect!

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
alcaeus: Yes, I refactored the code to ensure I keep the original field name around instead of replacing it.

alcaeus:
Note: there are some cases that still aren't covered by this pull requests. I'm currently producing test cases for these cases to fix them properly.

tests/Doctrine/ODM/MongoDB/Tests/Functional/DocumentPersisterTest.php
```diff
@@ -134,10 +134,10 @@ public static function getTestPrepareFieldNameData(): array
             ['association.nested', 'associationName.nestedName'],
             ['association.nested.$id', 'associationName.nestedName.$id'],
             ['association.nested._id', 'associationName.nestedName._id'],
-            ['association.nested.id', 'associationName.nestedName._id'],
```
alcaeus, line 137: These tests needed changing, as the field name translation only worked by accident. The `nested` property in the association document has no `targetDocument`, so we can't reliably know how to handle field names. The old logic reused the current ClassMetadata instance in these cases, which just happened to be the correct one in the tests, but might as well be incorrect. For example, this is the reason why an `id` property in an embedded document mapped without a `targetDocument` was renamed to `_id` in queries, even when it shouldn't have to.

Instead of changing the test documents to have a `targetDocument`, I figured it would be more prudent to highlight this change here. If reviewers feel that this change is a bit too risky for a patch release or even a minor version, please speak now or forever hold your peace.

GromNaN: What is a upgrade path if someone hits this breaking change?

alcaeus: That's a good question, and the answer isn't straightforward. First, I should clarify that the instances where the ODM did field name translation, it likely didn't use the correct mappings, unless somebody had an A -> X -> A relationship. In all other cases, mappings wouldn't line up.

The problem only appeared in embeds without a target document, so the obvious solution would be to set a target document. If there is no common ancestor to the possible embedded documents, users would never be able to rely on mappings for queries anyways and would have to use field name translation themselves, i.e. write the field names as they appear in the database.

GromNaN: OK. I propose to target the minor version 2.13 instead of bugfix 2.12.2 and call it a feature.

GromNaN (approved):

tests/Doctrine/ODM/MongoDB/Tests/Functional/DocumentPersisterTest.php
jmikola: I concur with targeting 2.13 instead of 2.12.2.

> The problem only appeared in embeds without a target document

Is there a legitimate use case for using EmbedMany or EmbedOne without a targetDocument mapping? When persisting to BSON, I assume ODM can use the class name to lookup metadata, but what happens when hydrating back from BSON? There'd be no hint for class metadata.

Perhaps this is a case where we should emit a warning/notice advising users to always specify `targetDocument` (even if it's a [mapped superclass](https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/latest/reference/inheritance-mapping.html#inheritance-mapping)).

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -1083,29 +1087,29 @@ public function prepareQueryOrNewObj(array $query, bool $isNewObj = false): arra
...
 
-                $value[$k] = $this->convertToDatabaseValue($fieldName, $v);
+                $value[$k] = $this->convertToDatabaseValue($fieldName, $v, $class);
             }
 
             return $value;
         }
 
-        if (! $this->class->hasField($fieldName)) {
+        if (! $class || ! $class->hasField($fieldName)) {
```
jmikola, line 1104: Confirming this is the fallback behavior when there's no `targetDocument` mapping. I assume if we decide to warn about that, we'd do so when initially parsing the mapping info (not here).

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -1083,29 +1087,29 @@ public function prepareQueryOrNewObj(array $query, bool $isNewObj = false): arra
...
             if ($value instanceof BackedEnum) {
                 $value = $value->value;
             }
 
             return Type::convertPHPToDatabaseValue($value);
         }
 
-        $mapping  = $this->class->fieldMappings[$fieldName];
+        $mapping  = $class->fieldMappings[$fieldName];
```
jmikola, line 1112: And this is the actual bug fix where we utilize the specific class mapping instead of the "current" class. I think this lines up with your "A -> X -> A" mapping example in https://github.com/doctrine/mongodb-odm/pull/2827/#discussion_r2374937004.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -1132,6 +1136,20 @@ private function convertToDatabaseValue(string $fieldName, $value)
         return $value;
     }
 
+    private function prepareQueryReference(mixed $value, ClassMetadata $class): mixed
+    {
+        if (! is_array($value)) {
```
jmikola: Is this worth a comment, like you have below for handling objects w/o operators and DBRefs? Maybe combine this into the next conditional to reduce the size of this method (and use a 2-line comment).

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -1143,16 +1161,22 @@ private function convertToDatabaseValue(string $fieldName, $value)
...
-        $class ??= $this->class;
+        $class   ??= $this->class;
+        $fieldName = $fieldNamePrefix . $originalFieldName;
+
+        // Process identifier fields
+        if (($class->hasField($originalFieldName) && $class->isIdentifier($originalFieldName)) || $originalFieldName === '_id') {
+            $fieldName = $fieldNamePrefix . '_id';
 
-        // @todo Consider inlining calls to ClassMetadata methods
+            return [[$fieldName, $prepareValue ? $this->prepareQueryReference($value, $class) : $value]];
```
jmikola, line 1175: I realize the method's return is documented as `array<array{string, mixed}>`, but can we describe the structure of this in the doc block comment? looks like an array of tuples, each of which contains field names and values (both prepared).

tests/Doctrine/ODM/MongoDB/Tests/Functional/DocumentPersisterTest.php
alcaeus: No use case that I know of, but I'd have to check if specifying a `discriminatorMap` in the mapping without a `targetDocument` will work.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -1225,121 +1229,99 @@ private function prepareQueryElement(string $fieldName, $value = null, ?ClassMet
...
+        }
+
+        // Don't recurse for references. Instead, prepare the reference directly
+        if (! empty($mapping['reference'])) {
+            // First part is the name of the reference
+            // Second part is either a positional operator, index/key, or the name of a field
+            // Third part (if any) is the name of a field
+            // That means, we can implode all field parts except the first as the next field name
+            if ($fieldNameParts[1] === '$') {
+                $objectProperty  = $fieldNameParts[2];
```
jmikola, line 1270: Consider `assert($partCount >= 3)`.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -1225,121 +1229,99 @@ private function prepareQueryElement(string $fieldName, $value = null, ?ClassMet
          * significant: "fieldName.objectProperty" with an optional index or key
          * for collections stored as either BSON arrays or objects.
          */
-        $e = explode('.', $fieldName, 4);
+        $fieldNameParts = explode('.', $originalFieldName, 4);
```
jmikola, line 1234: Exploding with a limit of four means we'll get the following for `a.b.c.d.e`:

```php
[
    'a',
    'b',
    'c',
    'd.e',
]
```

Below, you assert that `explode()` produces _at least_ 2 parts. But later, you also access `$fieldNameParts[2]` on the condition that `$fieldNameParts[1] === '$')`. Should we have an assertion going into that as well? Maybe `assert($partCount >= 3)`? Left a comment on this below.

I noted that `$fieldNameParts[3]` (and beyond) is never accessed, but the `implode()` a few lines below will combine everything from array offset 1 onwards (i.e. `b.c.d.e` given the example above).

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -1225,121 +1229,99 @@ private function prepareQueryElement(string $fieldName, $value = null, ?ClassMet
...
+            // Without a target document, no further processing is possible
+            return [[$fieldName, $prepareValue ? $this->convertToDatabaseValue($fieldNameParts[0], $value) : $value]];
+        }
+
+        // Don't recurse for references. Instead, prepare the reference directly
+        if (! empty($mapping['reference'])) {
+            // First part is the name of the reference
+            // Second part is either a positional operator, index/key, or the name of a field
+            // Third part (if any) is the name of a field
+            // That means, we can implode all field parts except the first as the next field name
```
jmikola, line 1267: Is this new code path tested? I don't see any queries utilizing a positional operator in the accompanying tests.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -1225,121 +1229,99 @@ private function prepareQueryElement(string $fieldName, $value = null, ?ClassMet
...
+            return [[$fieldName, $prepareValue ? $this->convertToDatabaseValue($objectProperty, $value, $targetClass) : $value]];
         }
 
-        /* The property path may include a third field segment, excluding the
-         * collection item pointer. If present, this next object property must
-         * be processed recursively.
+        /*
+         * 1 element: impossible (because of the dot)
+         * 2 elements: fieldName.objectProperty, fieldName.<index>, or fieldName.$. For EmbedMany and ReferenceMany, treat the second element as index if $inNewObj is true and convert the value. Otherwise, recurse.
+         * 3+ elements: fieldname.foo.bar, fieldName.<index>.foo, or fieldName.$foo. For EmbedMany and ReferenceMany, treat the second element as index, and recurse into the third element. Otherwise, recurse with the second element as field name.
```
jmikola: Why is `fieldName.$foo` here included in the "3+ elements" line? That would be split into two components, no?

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -1225,121 +1229,99 @@ private function prepareQueryElement(string $fieldName, $value = null, ?ClassMet
...
+
+                    return [[$fieldName, $prepareValue ? $this->convertToDatabaseValue($fieldNameParts[0], $value, $targetClass) : $value]];
                 }
 
-                $fieldNames = [[$nextObjectProperty, $value]];
+                // When there are more than two segments, treat the second segment (index/key/positional operator) as part of the field name and recurse into the rest
+                $newPrefix    = $fieldNamePrefix . $mapping['name'] . '.' . $fieldNameParts[1] . '.';
+                $newFieldName = implode('.', array_slice($fieldNameParts, 2));
+            } else {
+                // When serialising a query, the second segment is a positional operator ($), a numeric index for collections, or anything else for a hash.
```
jmikola:
```suggestion
                // When serializing a query, the second segment is a positional operator ($), a numeric index for collections, or anything else for a hash.
```

tests/Doctrine/ODM/MongoDB/Tests/Query/BuilderTest.php
```diff
@@ -837,6 +837,35 @@ public function testNonRewindable(): void
...
+        $filter = $query['query'];
+        $newObj = $query['newObj'];
+
+        self::assertArrayHasKey('embedded.embedded.test', $filter);
+        self::assertInstanceOf(ObjectId::class, $filter['embedded.embedded.test']);
+
+        self::assertArrayHasKey('embedded.embedded.embedded', $newObj['$set']);
+        self::assertIsArray($newObj['$set']['embedded.embedded.embedded']);
+        self::assertArrayHasKey('test', $newObj['$set']['embedded.embedded.embedded']);
+        self::assertInstanceOf(ObjectId::class, $newObj['$set']['embedded.embedded.embedded']['test']);
```
GromNaN, line 866: I find the test easier to read with array comparison, but that we need an intermediate comparator to assert the same type and value. As `$objectId == $oidString`

```php
        self::assertSameValueAndtype(['embedded.embedded.test' => $objectId], $filter);
        self::assertSameValueAndtype(['$set' => ['embedded.embedded.embedded' => ['test' => $objectId]]], $newObj);
```

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
alcaeus: That was a typo -- it should be `fieldName.$.foo`. Fixed now.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
alcaeus: Added.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
alcaeus: Added a comment.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
alcaeus: Combined with the conditional below.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
alcaeus: Changed to Simplified English.

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
alcaeus: Added some and fixed issues I found with the many relationships.

lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
```diff
@@ -507,4 +514,23 @@ public function prepareAssociatedCollectionValue(PersistentCollectionInterface $
...
+
+    private function preparePersistentCollection(array $mapping, object $owner, mixed $rawValue): PersistentCollectionInterface
+    {
+        if ($rawValue instanceof PersistentCollectionInterface) {
+            return $rawValue;
+        }
+
+        // If $actualData[$name] is not a Collection then use an ArrayCollection.
+        if (! $rawValue instanceof Collection) {
+            $rawValue = new ArrayCollection($rawValue);
```
jmikola, line 527: Rather than assert the array type here, would it make sense to change the type hint for `$rawValue` from `mixed` to `array|Collection`?

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
```diff
@@ -1041,34 +1042,37 @@ public function addFilterToPreparedQuery(array $preparedQuery): array
      *
      * PHP field names and types will be converted to those used by MongoDB.
      *
-     * @param array<string, mixed> $query
+     * @param array<string|int, mixed> $query
```
jmikola, line 1045: In `foreach`, you cast `$query` keys to a string. Is this typing just necessary to account for possible numeric field names?

GromNaN (approved):

lib/Doctrine/ODM/MongoDB/Persisters/DocumentPersister.php
alcaeus: Yes, without this the string cast is consider superfluous and would be removed.

lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php
alcaeus: Good point. Fixed!

PR #2829, opened 2025-09-23
Add test on Atlas Search features using an MongoDB Atlas Local Docker

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | [PHPORM-390](https://jira.mongodb.org/browse/PHPORM-390)

#### Summary

- Run Atlas tests with MongoDB Atlas Local
- Add integration test for Vector Search
- Add integration test for Search

The `configureFailPoint` command is not available in Atlas Local. I mark the tests as skipped instead of adding a group to exclude them from atlas.

### DIFF
--- a/.github/workflows/atlas-ci.yml
+++ b/.github/workflows/atlas-ci.yml
@@ -9,26 +9,20 @@ on:
 
 jobs:
   atlas-local:
-    runs-on: "${{ matrix.os }}"
-    name: "PHP/${{ matrix.php-version }} Symfony/${{ matrix.symfony }} Proxy/${{ matrix.proxy }}"
+    runs-on: "ubuntu-latest"
     strategy:
       fail-fast: false
       matrix:
-        os:
-          - "ubuntu-latest"
         php-version:
-          - "8.2"
-          - "8.3"
           - "8.4"
         symfony:
           - "stable"
-          - "6.4"
         proxy:
           - "lazy-ghost"
         include:
           # Test with ProxyManager
-          - php-version: "8.2"
-            symfony: "stable"
+          - php-version: "8.1"
+            symfony: "6.4"
             proxy: "proxy-manager"
             os: "ubuntu-latest"
 
@@ -99,7 +93,7 @@ jobs:
           composer-options: "--prefer-dist"
 
       - name: "Run PHPUnit with Atlas Local"
-        run: "vendor/bin/phpunit"
+        run: "vendor/bin/phpunit --group atlas"
         env:
           DOCTRINE_MONGODB_SERVER: "mongodb://127.0.0.1:27017/?directConnection=true"
           USE_LAZY_GHOST_OBJECTS: ${{ matrix.proxy == 'lazy-ghost' && '1' || '0' }}
--- a/.github/workflows/atlas-ci.yml
+++ b/.github/workflows/atlas-ci.yml
@@ -1,0 +1,115 @@
+name: "Atlas CI"
+
+on:
+  push:
+    branches:
+      - "*.x"
+      - "feature/*"
+  pull_request:
+    branches:
+      - "*.x"
+      - "feature/*"
+
+env:
+  MONGODB_EXT_V1: mongodb-1.21.0
+  MONGODB_EXT_V2: stable
+
+jobs:
+  atlas-local:
+    runs-on: "${{ matrix.os }}"
+
+    name: "PHP/${{ matrix.php }} Symfony/${{ matrix.symfony }} Driver/${{ matrix.driver }} Proxy/${{ matrix.proxy }}"
+
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - "ubuntu-latest"
+        php:
+          - "8.2"
+          - "8.3"
+          - "8.4"
+        symfony:
+          - "stable"
+          - "6.4"
+        driver:
+          - 1
+          - 2
+        proxy:
+          - "lazy-ghost"
+        include:
+          # Test with ProxyManager
+          - php: "8.2"
+            symfony: "stable"
+            driver: 2
+            proxy: "proxy-manager"
+            os: "ubuntu-latest"
+
+    steps:
+      - uses: "actions/checkout@v5"
+
+      - name: "Create MongoDB Atlas Local"
+        run: |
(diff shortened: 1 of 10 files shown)


tests/Doctrine/ODM/MongoDB/Tests/Functional/AtlasSearchTest.php
```diff
@@ -0,0 +1,116 @@
...
+        $article3->text  = 'The MongoDB ODM for PHP provides an easy way to work with MongoDB documents using object-oriented programming.';
+        $article3->setAuthor($user);
+
+        $this->dm->persist($article1);
+        $this->dm->persist($article2);
+        $this->dm->persist($article3);
+        $this->dm->flush();
+
+        // Wait for the search index to be ready (Atlas Local needs time to build the index)
+        sleep(2);
```
GromNaN, line 51: This could be smarter with a method that detect when the indexes are ready: https://github.com/doctrine/mongodb-odm/pull/2830

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -473,7 +473,7 @@
 
     public const VECTOR_SIMILARITY_EUCLIDEAN   = 'euclidean';
     public const VECTOR_SIMILARITY_COSINE      = 'cosine';
-    public const VECTOR_SIMILARITY_DOT_PRODUCT = 'dot_product';
+    public const VECTOR_SIMILARITY_DOT_PRODUCT = 'dotProduct';
```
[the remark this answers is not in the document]
jmikola: @GromNaN was this not actually a BC break because the constant was introduced in https://github.com/doctrine/mongodb-odm/commit/9a0e994714bd72299d3742676513ddd21c33946d. Here, it looks like you're making the value consistent with what you already had in the documentation from the earlier commit.

Feel free to respond and resolve. I just wanted to confirm.

.github/workflows/continuous-integration.yml
```diff
@@ -159,7 +159,7 @@ jobs:
           topology: ${{ matrix.topology }}
 
       - name: "Run PHPUnit"
-        run: "vendor/bin/phpunit"
+        run: "vendor/bin/phpunit --exclude-group=atlas"
```
jmikola, line 162: Noted that this will skip the tests with `sleep(2)`.

.github/workflows/atlas-ci.yml
```diff
@@ -0,0 +1,105 @@
...
+          composer require --no-update --dev symfony/cache:^${{ matrix.symfony }}
+
+      - name: "Install dependencies with Composer"
+        uses: "ramsey/composer-install@v3"
+        with:
+          dependency-versions: "highest"
+          composer-options: "--prefer-dist"
+
+      - name: "Run PHPUnit with Atlas Local"
+        run: "vendor/bin/phpunit"
```
jmikola: Noted that this will run the entire test suite with Atlas Local. How do you intend to trigger the `atlas-ci.yml` jobs? Presumably we don't need this entire matrix run on each PR. Would it make more sense to initially restrict test execution to the "atlas" group to compensate for the tests that will be exluded by `continuous-integration.yml`?

tests/Doctrine/ODM/MongoDB/Tests/BaseTestCase.php
```diff
@@ -294,4 +306,28 @@ private static function detectTransactionSupport(): bool
 
         return $manager->selectServer()->getType() !== Server::TYPE_STANDALONE;
     }
+
+    protected function createFailPoint(string $failCommand, bool $transient = false, int $times = 1): void
```
jmikola, line 310: This looks like it was adapted from what we have in PHPLIB, although this method is _only_ useful for the `failCommand` fail point. That assumption also exists in the `tearDown()` method.

Just wanted to confirm that this is perfectly suitable for ODM's present needs. In the future, you may want to make this more consistent with PHPLIB's implementation, though.

tests/Doctrine/ODM/MongoDB/Tests/UnitOfWorkTransactionalCommitConsistencyTest.php
```diff
@@ -48,7 +38,7 @@ public function testFatalInsertError(): void
         $friendUser = new FriendUser('GromNaN');
         $this->uow->persist($friendUser);
 
-        $this->createFatalFailPoint('insert');
+        $this->createFailPoint('insert');
```
jmikola, line 41: Noted that `createFatalFailPoint` and `createTransientFailPoint` only differ by way of the "TransientTransactionError" error label.

jmikola (approved):
LGTM. I just have some questions but I don't they will warrant further changes.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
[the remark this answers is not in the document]
GromNaN: This code is not released yet. This is a bug fix as `dotProduct` is the value expected by MongoDB: https://www.mongodb.com/docs/php-library/current/indexes/atlas-search-index/#create-a-search-index

tests/Doctrine/ODM/MongoDB/Tests/BaseTestCase.php
GromNaN: I forgot about the [PHPLIB implementation](https://github.com/mongodb/mongo-php-library/blob/6bfa41151827a790f962a7a7511ddb38ac9a5e02/tests/FunctionalTestCase.php#L205-L239), this is definitely a simpler approach as we only need to add 1 fail point at a time.

tests/Doctrine/ODM/MongoDB/Tests/UnitOfWorkTransactionalCommitConsistencyTest.php
GromNaN: Yes, that's why I added the `$transient` parameter.

.github/workflows/atlas-ci.yml
GromNaN: My goal was to run all the tests with Atlas, but that's not possible due to the lack of support of failPoints. So I simplified the test matrix for Atlas.

PR #2830, opened 2025-09-24
Add `SchemaManager::waitForSearchIndexes()`

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | [PHPORM-399](https://jira.mongodb.org/browse/PHPORM-399)

#### Summary

This method waits until all search indexes for specified (or all) document classes are both queryable and have indexed all documents in their corresponding collections.

- It takes up to three arguments: a list of document class names (`$classNames`, optional), a maximum wait time in milliseconds (`$maxTimeMs`, default 10,000 ms), and a delay between checks in milliseconds (`$waitTimeMs`, default 100 ms).
- The method validates input arguments, then iterates over the document classes to identify the search indexes that need to be checked.
- In a loop, it checks whether all required search indexes exist and are marked as queryable.
- ~For queryable indexes, it further verifies if all documents have been indexed by comparing the count of indexed documents with the total collection count, using aggregation pipelines appropriate for either `'search'` or `'vectorSearch'` index types.~ (removed because unreliable)
- If not all indexes are ready, it waits for `$waitTimeMs` before checking again, until either all indexes are ready or the maximum wait time (`$maxTimeMs`) is exceeded (in which case an exception is thrown).

This addition provides a mechanism for applications to block until search indexes are fully prepared for querying, ensuring consistency and readiness before proceeding with operations that depend on those indexes.

Similar implementation: https://github.com/mongodb-labs/pymongo-search-utils/blob/dddd51f6e1784caa43e0dc66ec8d9487e0762d58/pymongo_search_utils/index.py#L259-L298

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/SchemaException.php
+++ b/lib/Doctrine/ODM/MongoDB/SchemaException.php
@@ -1,0 +1,26 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ODM\MongoDB;
+
+use RuntimeException;
+
+use function implode;
+use function sprintf;
+
+/**
+ * Exception for schema-related issues.
+ */
+final class SchemaException extends RuntimeException
+{
+    /**
+     * @internal
+     *
+     * @param string[] $missingIndexes
+     */
+    public static function missingSearchIndex(string $documentClass, array $missingIndexes): self
+    {
+        return new self(sprintf('The document class "%s" is missing the following search index(es): "%s"', $documentClass, implode('", "', $missingIndexes)));
+    }
+}
--- a/lib/Doctrine/ODM/MongoDB/SchemaManager.php
+++ b/lib/Doctrine/ODM/MongoDB/SchemaManager.php
@@ -20,12 +20,14 @@ use function array_diff;
 use function array_diff_key;
 use function array_filter;
 use function array_keys;
+use function array_map;
 use function array_merge;
 use function array_search;
 use function array_unique;
 use function array_values;
 use function assert;
 use function count;
+use function hrtime;
 use function implode;
 use function in_array;
 use function is_array;
@@ -35,6 +37,7 @@ use function iterator_to_array;
 use function ksort;
 use function sprintf;
 use function str_contains;
+use function usleep;
 
 /**
  * @phpstan-import-type IndexMapping from ClassMetadata
@@ -338,6 +341,65 @@ final class SchemaManager
     }
 
     /**
+     * Wait until all search indexes are queryable for the given document classes.
+     *
+     * @param list<class-string>|null $classNames List of class names to check, or null to check all mapped classes
+     * @param positive-int            $maxTimeMs  Maximum time to wait in milliseconds (default: 10,000 ms)
+     * @param positive-int            $waitTimeMs Time to wait between checks in milliseconds (default: 100 ms)
+     */
+    public function waitForSearchIndexes(?array $classNames = null, int $maxTimeMs = 10_000, int $waitTimeMs = 100): void
+    {
+        if ($maxTimeMs < 1) {
+            throw new InvalidArgumentException('$maxTimeMs must be a positive number of milliseconds.');
+        }
+
+        if ($waitTimeMs < 1) {
+            throw new InvalidArgumentException('$waitTimeMs must be a positive number of milliseconds.');
+        }
+
+        $classes = $classNames === null ? $this->metadataFactory->getAllMetadata() : array_map($this->metadataFactory->getMetadataFor(...), $classNames);
+
+        /** @var array<class-string, string[]> $indexesToCheck Search indexes for each class */
+        $indexesToCheck = [];
+        foreach ($classes as $class) {
+            if (! $class->hasSearchIndexes()) {
+                continue;
+            }
+
+            $indexesToCheck[$class->getName()] = array_column($class->getSearchIndexes(), 'name');
+        }
+
+        $start = hrtime(true);
+        while ($indexesToCheck) {
+            if (hrtime(true) > $start + $maxTimeMs * 1_000_000) {
+                throw new MongoDBException(sprintf('Timed out waiting for search indexes to become queryable after %d ms. Search indexes are not ready for the following class(es): %s', $maxTimeMs, implode(', ', array_keys($indexesToCheck))));
+            }
+
+            foreach ($indexesToCheck as $className => $indexNames) {
+                $collection = $this->dm->getDocumentCollection($className);
+
+                /** @var array<string, bool> $indexStatus Queryable status for each index name */
+                $indexStatus = array_column(iterator_to_array($collection->listSearchIndexes([
+                    'filter' => ['name' => ['$in' => array_keys($indexNames)]],
+                    'typeMap' => ['root' => 'array'],
+                ])), 'queryable', 'name');
+
+                // Check that all indexes exist
+                $missingIndexes = array_diff_key($indexNames, array_keys($indexStatus));
+                if ($missingIndexes) {
+                    throw SchemaException::missingSearchIndex($className, $missingIndexes);
+                }
+
+                // Remove the indexes that are ready from the list of indexes to check
+                $indexesToCheck[$className] = array_keys(array_filter($indexStatus, static fn ($queryable) => ! $queryable));
+            }
+
+            // Remove empty arrays and wait before checking again
+            ($indexesToCheck = array_filter($indexesToCheck)) && usleep($waitTimeMs * 1_000);
+        }
+    }
+
+    /**
      * Create search indexes for the given document class.
      *
      * @param class-string $documentName
@@ -368,7 +430,7 @@ final class SchemaManager
         $unprocessedNames = array_diff($definedNames, $createdNames);
 
         if (! empty($unprocessedNames)) {
-            throw new InvalidArgumentException(sprintf('The following search indexes for %s were not created: %s', $class->name, implode(', ', $unprocessedNames)));
+            throw SchemaException::missingSearchIndex($class->name, $unprocessedNames);
         }
     }
 
--- a/tests/Doctrine/ODM/MongoDB/Tests/Functional/AtlasSearchTest.php
+++ b/tests/Doctrine/ODM/MongoDB/Tests/Functional/AtlasSearchTest.php
@@ -7,9 +7,8 @@ namespace Doctrine\ODM\MongoDB\Tests\Functional;
 use Doctrine\ODM\MongoDB\Tests\BaseTestCase;
 use Documents\CmsArticle;
 use Documents\CmsUser;
+use MongoDB\Driver\WriteConcern;
 use PHPUnit\Framework\Attributes\Group;
-
-use function sleep;
 
 #[Group('atlas')]
 class AtlasSearchTest extends BaseTestCase
@@ -18,7 +17,6 @@ class AtlasSearchTest extends BaseTestCase
     {
         $schemaManager = $this->dm->getSchemaManager();
         $schemaManager->createDocumentCollection(CmsArticle::class);
-        $schemaManager->createDocumentSearchIndexes(CmsArticle::class);
 
         $user         = new CmsUser();
         $user->status = 'active';
@@ -45,10 +43,15 @@ class AtlasSearchTest extends BaseTestCase
         $this->dm->persist($article1);
         $this->dm->persist($article2);
         $this->dm->persist($article3);
-        $this->dm->flush();
+
+        // Write with majority concern to ensure data is visible for search
+        $this->dm->flush(['writeConcern' => new WriteConcern(WriteConcern::MAJORITY)]);
+
+        // Index must be created after data insertion, so the index status is not immediately "READY"
+        $schemaManager->createDocumentSearchIndexes(CmsArticle::class);
 
         // Wait for the search index to be ready (Atlas Local needs time to build the index)
-        sleep(2);
+        $schemaManager->waitForSearchIndexes([CmsArticle::class, CmsUser::class]);
 
         $results = $this->dm->createAggregationBuilder(CmsArticle::class)
             ->search()
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -385,18 +385,6 @@ parameters:
 			path: lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
 
 		-
-			message: '#^Class Doctrine\\ODM\\MongoDB\\Mapping\\ClassMetadata has type alias SearchIndexDefinition with no value type specified in iterable type array\.$#'
-			identifier: missingType.iterableValue
-			count: 2
-			path: lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
-
-		-
-			message: '#^Class Doctrine\\ODM\\MongoDB\\Mapping\\ClassMetadata has type alias SearchIndexMapping with no value type specified in iterable type array\.$#'
-			identifier: missingType.iterableValue
-			count: 2
-			path: lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
-
-		-
 			message: '#^Instanceof between Doctrine\\Persistence\\Reflection\\RuntimeReflectionProperty and ReflectionProperty will always evaluate to true\.$#'
 			identifier: instanceof.alwaysTrue
 			count: 1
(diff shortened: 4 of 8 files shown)


lib/Doctrine/ODM/MongoDB/SchemaException.php
```diff
@@ -0,0 +1,26 @@
...
+final class SchemaException extends RuntimeException
+{
+    /**
+     * @internal
+     *
+     * @param string[] $missingIndexes
+     */
+    public static function missingSearchIndex(string $documentClass, array $missingIndexes): self
+    {
+        return new self(sprintf('The document class "%s" is missing the following search indexes: %s', $documentClass, implode(', ', $missingIndexes)));
```
jmikola:
```suggestion
        return new self(sprintf('The document class "%s" is missing the following search index(es): %s', $documentClass, implode(', ', $missingIndexes)));
```

Since `$missingIndexes` may have one element.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -340,6 +347,124 @@ public function createSearchIndexes(): void
...
+        }
+
+        $classes = $classNames === null ? $this->metadataFactory->getAllMetadata() : array_map($this->metadataFactory->getMetadataFor(...), $classNames);
+
+        // List all the collections from the Metadata with the list of their search index names
+        /** @var array<class-string, array<string, bool>> $indexesToCheck */
+        $indexesToCheck = [];
+        foreach ($classes as $class) {
+            $searchIndexes = $class->getSearchIndexes();
+            if (! $searchIndexes) {
```
jmikola: I suppose you could also do:

```php
if (! $class->hasSearchIndexes()) {
    continue;
}

$indexesToCheck[$class->getName()] = array_fill_keys(array_column($class->getSearchIndexes(), 'name'), null);
```

Extra method call, but maybe more human-readable. I'll defer to you.

I'm not really sure why `hasSearchIndexes()` exists if internal code could always just fetch the array and check if it's empty.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -340,6 +347,124 @@ public function createSearchIndexes(): void
...
+                continue;
+            }
+
+            $indexesToCheck[$class->getName()] = array_fill_keys(array_column($searchIndexes, 'name'), null);
+        }
+
+        $start = hrtime(true);
+        while ($indexesToCheck) {
+            if (hrtime(true) > $start + $maxTimeMs * 1_000_000) {
```
jmikola, line 374: Thank goodness we've got 64-bit integers, eh?

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -340,6 +347,124 @@ public function createSearchIndexes(): void
...
+                    $indexes[$indexName]['queryable'] = false;
+                }
+
+                // Remove the indexes that are ready from the list of indexes to check
+                $indexes = array_filter($indexes, static fn ($index) => $index['queryable'] === false);
+            }
+
+            // Remove empty arrays
+            $indexesToCheck = array_filter($indexesToCheck);
+            $indexesToCheck && usleep($waitTimeMs * 1_000);
```
jmikola: I like using `and` for this purpose, but defer to you. It also has the [lowest precedence](https://www.php.net/manual/en/language.operators.precedence.php).

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -340,6 +347,124 @@ public function createSearchIndexes(): void
...
+                continue;
+            }
+
+            $indexesToCheck[$class->getName()] = array_fill_keys(array_column($searchIndexes, 'name'), null);
+        }
+
+        $start = hrtime(true);
+        while ($indexesToCheck) {
+            if (hrtime(true) > $start + $maxTimeMs * 1_000_000) {
+                throw new MongoDBException(sprintf('Timed out waiting for search indexes to become queryable after %d ms. Search indexes are not ready for this classes: %s', $maxTimeMs, implode(', ', array_keys($indexesToCheck))));
```
jmikola:
```suggestion
                throw new MongoDBException(sprintf('Timed out waiting for search indexes to become queryable after %d ms. Search indexes are not ready for the following class(es): %s', $maxTimeMs, implode(', ', array_keys($indexesToCheck))));
```

Optionally remove "the following" if you want this to be more concise.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -340,6 +347,124 @@ public function createSearchIndexes(): void
...
+            }
+
+            foreach ($indexesToCheck as $className => &$indexes) {
+                $collection = $this->dm->getDocumentCollection($className);
+
+                /** @var list<array{name: string, type: 'search'|'vectorSearch', queryable: bool, latestDefinition: array<string, mixed>}> $listSearchIndexes */
+                $listSearchIndexes = iterator_to_array($collection->listSearchIndexes(['typeMap' => ['root' => 'array', 'document' => 'array', 'array' => 'array']]));
+                $listSearchIndexes = array_column($listSearchIndexes, null, 'name');
```
jmikola: Noted per [`array_column()`](https://www.php.net/array_column) that this essentially re-indexes the array by the `name` field of each document.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -340,6 +347,124 @@ public function createSearchIndexes(): void
...
+                $indexes = array_intersect_key($listSearchIndexes, $indexes);
+
+                // Check that all indexes marked as queryable have finished indexing all the documents
+                $collectionCount = null;
+                foreach ($indexes as $indexName => $index) {
+                    if (! $index['queryable']) {
+                        continue;
+                    }
+
+                    $collectionCount ??= $collection->estimatedDocumentCount();
```
GromNaN: > I don't think this solution is safe. `estimatedDocumentCount` is only an estimate so the documents in the index might never reach that value. Another case is when documents are being deleted from the collection, the previous count will be no longer be accurate.
https://jira.mongodb.org/browse/DRIVERS-2995?focusedCommentId=7717617&focusedId=7717617&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-7717617

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -340,6 +347,124 @@ public function createSearchIndexes(): void
...
+                                    ],
+                                ],
+                                ['$project' => ['total' => '$count.total']],
+                            ],
+                            'vectorSearch' => [
+                                [
+                                    '$vectorSearch' => array_reduce(
+                                        $index['latestDefinition']['fields'],
+                                        // There is at least one vector field in the index definition
+                                        static fn ($stage, $field) => ! $stage && $field['type'] === 'vector' ? [
```
jmikola: How does this work with multiple fields? [`array_reduce()`](https://www.php.net/array_reduce) is going to iteratively reduce everything in `$index['latestDefinition']['fields']` to a single value, but I don't see how things get combined. It looks as if the reduce callable will end up using the last field to produce its `queryVector` and `path` options.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -340,6 +347,124 @@ public function createSearchIndexes(): void
...
+
+                    // Count indexed documents
+                    $indexCount = $collection->aggregate(
+                        match ($index['type']) {
+                            'search' => [
+                                [
+                                    '$searchMeta' => [
+                                        'index' => $indexName,
+                                        'exists' => ['path' => '_id'],
+                                        'count' => ['type' => 'total', 'threshold' => $collectionCount],
```
jmikola: Does `threshold` apply to a type `total` count? [Counting: Options](https://www.mongodb.com/docs/atlas/atlas-search/counting/#options) suggests it's only applicable to `lowerBound`.

Also, you're using type `total` here but later comparing with an estimated document count from the collection. Is that problematic?

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -340,6 +347,124 @@ public function createSearchIndexes(): void
...
+                    throw SchemaException::missingSearchIndex($className, array_keys($missingIndexes));
+                }
+
+                // Update the status of the indexes we are checking, others are ignored.
+                $indexes = array_intersect_key($listSearchIndexes, $indexes);
+
+                // Check that all indexes marked as queryable have finished indexing all the documents
+                $collectionCount = null;
+                foreach ($indexes as $indexName => $index) {
+                    if (! $index['queryable']) {
```
jmikola: I understand that this method count documents for both search and vector indexes. Is there any reason you couldn't rely on [Search Index Statuses](https://www.mongodb.com/docs/manual/reference/operator/aggregation/listSearchIndexes/#mongodb-search-index-statuses)?

lib/Doctrine/ODM/MongoDB/SchemaManager.php
GromNaN: > the nanoseconds are returned as `int` (64bit platforms) or `float` (32bit platforms)

I think we are safe even if we lose some bits of precision.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
GromNaN: phpcs disagrees:
> Logical operator "and" is prohibited; use "&&" instead

lib/Doctrine/ODM/MongoDB/SchemaManager.php
GromNaN: It gets the first `type=vector` field. Others are skipped thanks to the `! $stage`.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
GromNaN: We don't have a status for "all documents are synchronized".

GromNaN:
I reviewed the full implementation to remove the counting of collection vs search index. In fact when documents are inserted with `wc:majority` before creating the search index, then the status of the search index is `queryable: true` only when all documents are synchronized. (at least from my tests with atlas local container).

lib/Doctrine/ODM/MongoDB/SchemaManager.php
GromNaN: Updated, so I could inline `$class->getSearchIndexes()`.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
jmikola: Looks like [ValidLogicalOperatorsSniff](https://github.com/squizlabs/PHP_CodeSniffer/blob/3.7.2/src/Standards/Squiz/Sniffs/Operators/ValidLogicalOperatorsSniff.php) is to blame for that. Boo.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -340,6 +345,67 @@ public function createSearchIndexes(): void
...
+        }
+
+        if ($waitTimeMs < 1) {
+            throw new InvalidArgumentException('$waitTimeMs must be a positive number of milliseconds.');
+        }
+
+        $classes = $classNames === null ? $this->metadataFactory->getAllMetadata() : array_map($this->metadataFactory->getMetadataFor(...), $classNames);
+
+        // List all the collections from the Metadata with the list of their search index names
+        /** @var array<class-string, ?bool> $indexesToCheck */
```
jmikola: Is this type accurate given how `$indexesToCheck` is assembled below? Reading the subsequent lines here, it looks like you're creating a map of class names to array values (not `?bool` as this might suggest).

I'll refer to that array value as `$indexes` since the `foreach` assigns it by reference.

The `$indexes` array for each class starts as a map of index names (from `array_column`) to `null`. I understand that only the keys are significant there since you use it for `array_diff_key()` and `array_intersect_key()`.

But later, you end up changing the type of `$indexes` to the result of `$listSearchIndexes` intersected with the original keys from `$indexes`. At that point, it's values are no longer `null`.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -340,6 +345,67 @@ public function createSearchIndexes(): void
...
+        while ($indexesToCheck) {
+            if (hrtime(true) > $start + $maxTimeMs * 1_000_000) {
+                throw new MongoDBException(sprintf('Timed out waiting for search indexes to become queryable after %d ms. Search indexes are not ready for the following class(es): %s', $maxTimeMs, implode(', ', array_keys($indexesToCheck))));
+            }
+
+            foreach ($indexesToCheck as $className => &$indexes) {
+                $collection = $this->dm->getDocumentCollection($className);
+
+                /** @var list<array{name: string, queryable: bool}> $listSearchIndexes */
+                $listSearchIndexes = iterator_to_array($collection->listSearchIndexes(['typeMap' => ['root' => 'array', 'document' => 'array', 'array' => 'array']]));
```
jmikola: Doesn't this briefly violate the `@var` declaration above? Or is that excluded from Psalm?

I would think you'd need to assign `$listSearchIndexes` directly from calling `array_column()` on the result of `iterator_to_array($collection->listSearchIndexes(...))` to ensure the variable always complies with its declared type.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -340,6 +345,67 @@ public function createSearchIndexes(): void
...
+                $listSearchIndexes = array_column($listSearchIndexes, 'queryable', 'name');
+
+                // Check that all required indexes exist
+                $missingIndexes = array_diff_key($indexes, $listSearchIndexes);
+                if ($missingIndexes) {
+                    throw SchemaException::missingSearchIndex($className, array_keys($missingIndexes));
+                }
+
+                // Update the status of the indexes we are checking, others are ignored.
+                $indexes = array_intersect_key($listSearchIndexes, $indexes);
```
jmikola: Here is where the type of each value within `$indexes` changes significantly from `null` to a result from `$listSearchIndexes`. I think this presents two concerns:

* The orignal `?bool` type for `$indexesToCheck` values, which I don't believe ever held true (per my earlier comment) now changes to `array{name: string, queryable: bool}` (from the `$listSearchIndexes` type declaration above).
* Using a reference assignment on the iteration value in such a large `foreach` body makes this harder to follow even if it is more performant.

I think this would be improved by doing a read-only iteration of `$indexesToCheck` and instead building up a separate output `$waitForIndexes` array. That might also allow you to remove the `array_filter` call below, since empty arrays would never be added in the first place.

tests/Doctrine/ODM/MongoDB/Tests/Functional/AtlasSearchTest.php
```diff
@@ -45,10 +43,15 @@ public function testAtlasSearch(): void
         $this->dm->persist($article1);
         $this->dm->persist($article2);
         $this->dm->persist($article3);
-        $this->dm->flush();
+
+        // Write with majority concern to ensure data is visible for search
+        $this->dm->flush(['writeConcern' => new WriteConcern(WriteConcern::MAJORITY)]);
```
jmikola, line 48: FWIW, I asked about this in [#atlas-search](https://mongodb.slack.com/archives/CQHG10SCD/p1759156925765859) as I couldn't find this discussed in the [Atlas Search docs](https://www.mongodb.com/docs/atlas/atlas-search/). In any event, no action needed on your part here.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
GromNaN: Good catch, I'll merge the 2 lines.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
GromNaN: That's right, so I found that I could simplify this type to `array<class-string, string[]>` since the queryable status is not necessary in this array, it only contains the list of index names.

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -337,6 +340,65 @@ public function createSearchIndexes(): void
...
+        while ($indexesToCheck) {
+            if (hrtime(true) > $start + $maxTimeMs * 1_000_000) {
+                throw new MongoDBException(sprintf('Timed out waiting for search indexes to become queryable after %d ms. Search indexes are not ready for the following class(es): %s', $maxTimeMs, implode(', ', array_keys($indexesToCheck))));
+            }
+
+            foreach ($indexesToCheck as $className => $indexNames) {
+                $collection = $this->dm->getDocumentCollection($className);
+
+                /** @var array<string, bool> $indexStatus Queryable status for each index name */
+                $indexStatus = array_column(iterator_to_array($collection->listSearchIndexes([
```
jmikola, line 382: What a marvelous function [`array_column()`](https://ben.ramsey.dev/projects/array-column/) is. Thank you, @ramsey.

jmikola (approved):

PR #2831, opened 2025-09-24
Prepare field names for search indexes and vector search indexes

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | maybe
| Fixed issues | [PHPORM-391](https://jira.mongodb.org/browse/PHPORM-391) [PHPORM-392](https://jira.mongodb.org/browse/PHPORM-392)

#### Summary

Map field names in search and vector search index definitions. Since the `$search` and `$vectorSearch` stages must always be first, we know there is no field name mapped from a previous stage.

- [ ] Support field path with dot notation, from https://github.com/doctrine/mongodb-odm/pull/2827
- [x] Add unit tests
- [x] Update functional tests, from https://github.com/doctrine/mongodb-odm/pull/2829

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search.php
@@ -9,6 +9,7 @@ use Doctrine\ODM\MongoDB\Aggregation\Stage;
 use Doctrine\ODM\MongoDB\Aggregation\Stage\Search\SearchOperator;
 use Doctrine\ODM\MongoDB\Aggregation\Stage\Search\SupportsAllSearchOperators;
 use Doctrine\ODM\MongoDB\Aggregation\Stage\Search\SupportsAllSearchOperatorsTrait;
+use Doctrine\ODM\MongoDB\Persisters\DocumentPersister;
 
 use function in_array;
 use function is_array;
@@ -70,7 +71,7 @@ class Search extends Stage implements SupportsAllSearchOperators
     /** @var array<string, -1|1|SortMeta> */
     private array $sort = [];
 
-    public function __construct(Builder $builder)
+    public function __construct(Builder $builder, private DocumentPersister $persister)
     {
         parent::__construct($builder);
     }
@@ -105,7 +106,7 @@ class Search extends Stage implements SupportsAllSearchOperators
         }
 
         if ($this->sort) {
-            $params->sort = (object) $this->sort;
+            $params->sort = (object) $this->persister->prepareSort($this->sort);
         }
 
         if ($this->operator !== null) {
@@ -213,5 +214,10 @@ class Search extends Stage implements SupportsAllSearchOperators
     protected function getSearchStage(): static
     {
         return $this;
+    }
+
+    protected function getDocumentPersister(): DocumentPersister
+    {
+        return $this->persister;
     }
 }
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/Compound.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/Compound.php
@@ -5,6 +5,8 @@ declare(strict_types=1);
 namespace Doctrine\ODM\MongoDB\Aggregation\Stage\Search;
 
 use Closure;
+use Doctrine\ODM\MongoDB\Aggregation\Stage\Search;
+use Doctrine\ODM\MongoDB\Persisters\DocumentPersister;
 
 use function array_map;
 
@@ -28,6 +30,11 @@ class Compound extends AbstractSearchOperator implements CompoundSearchOperatorI
 
     private string $currentClause    = 'must';
     private ?int $minimumShouldMatch = null;
+
+    public function __construct(Search $search, private DocumentPersister $persister)
+    {
+        parent::__construct($search, $this->persister);
+    }
 
     /**
      * @param T $operator
@@ -121,5 +128,10 @@ class Compound extends AbstractSearchOperator implements CompoundSearchOperatorI
     protected function getCompoundStage(): Compound
     {
         return $this;
+    }
+
+    protected function getDocumentPersister(): DocumentPersister
+    {
+        return $this->persister;
     }
 }
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/Compound.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/Compound.php
@@ -31,11 +31,6 @@ class Compound extends AbstractSearchOperator implements CompoundSearchOperatorI
     private string $currentClause    = 'must';
     private ?int $minimumShouldMatch = null;
 
-    public function __construct(Search $search, private DocumentPersister $persister)
-    {
-        parent::__construct($search, $this->persister);
-    }
-
     /**
      * @param T $operator
      *
@@ -128,10 +123,5 @@ class Compound extends AbstractSearchOperator implements CompoundSearchOperatorI
     protected function getCompoundStage(): Compound
     {
         return $this;
-    }
-
-    protected function getDocumentPersister(): DocumentPersister
-    {
-        return $this->persister;
     }
 }
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/Compound.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/Compound.php
@@ -5,8 +5,6 @@ declare(strict_types=1);
 namespace Doctrine\ODM\MongoDB\Aggregation\Stage\Search;
 
 use Closure;
-use Doctrine\ODM\MongoDB\Aggregation\Stage\Search;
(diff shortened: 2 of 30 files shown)


tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/SearchTest.php
```diff
@@ -270,8 +263,7 @@ public static function provideEmbeddedDocumentCompoundBuilders(): Generator
                     ],
                 ],
             ],
-            /** @param Search|CompoundSearchOperatorInterface $stage */
-            'createOperator' => static function ($stage) {
+            'createOperator' => static function (Search $stage) {
```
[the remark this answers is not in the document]
GromNaN: The `->compound()` method is only on the `Search` stage or `EmbeddedDocument` search operator.

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/SearchTest.php
```diff
@@ -539,16 +520,29 @@ public static function provideMoreLikeThisBuilders(): Generator
...
+                'moreLikeThis' => (object) [
+                    'like' => [
+                        ['disable-at' => new UTCDateTime(new DateTime('2020-01-01T00:00:00Z'))],
+                    ],
+                ],
+            ],
+            'createOperator' => static function (Search|CompoundSearchOperatorInterface|SupportsEmbeddableSearchOperators $stage) {
+                return $stage->moreLikeThis(['disabledAt' => new DateTime('2020-01-01T00:00:00Z')]);
+            },
+            'className' => User::class,
```
GromNaN, line 544: Switching to an other document class to validate transformation of the date value.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/Compound.php
```diff
@@ -29,6 +31,11 @@ class Compound extends AbstractSearchOperator implements CompoundSearchOperatorI
     private string $currentClause    = 'must';
     private ?int $minimumShouldMatch = null;
 
+    public function __construct(Search $search, private DocumentPersister $persister)
```
jmikola: Why does this constructor assign `$persister` to a `private` instead of merely passing it to `parent::__construct()` like the Autocomplete stage above?

I'm also curious why `getDocumentPersister()` is redefined.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/MoreLikeThis.php
```diff
@@ -33,6 +35,8 @@ public function getOperatorName(): string
 
     public function getOperatorParams(): object
     {
-        return (object) ['like' => $this->like];
+        return (object) [
+            'like' => array_map($this->getDocumentPersister()->prepareQueryOrNewObj(...), $this->like),
```
jmikola: Any reason you defined `prepareFieldPath()` in the abstract base class but chose not to do so for `prepareQueryOrNewObj()`? I presume with doing so you could also give it the `string|array` behavior.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/SupportsCompoundableOperatorsTrait.php
```diff
@@ -47,40 +50,40 @@ abstract protected function addOperator(SearchOperator $operator): SearchOperato
 
     public function autocomplete(string $path = '', string ...$query): Autocomplete&CompoundSearchOperatorInterface
     {
-        return $this->addOperator(new CompoundedAutocomplete($this->getCompoundStage(), $this->getAddOperatorClosure(), $this->getSearchStage(), $path, ...$query));
+        return $this->addOperator(new CompoundedAutocomplete($this->getCompoundStage(), $this->getAddOperatorClosure(), $this->getSearchStage(), $this->getDocumentPersister(), $path, ...$query));
```
jmikola, line 53: Noted that this class extends Autocomplete, which had its constructor modified to accept a DocumentPersister. I assume the same is true for other classes in this file, which are themselves absent from this PR.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search.php
jmikola, line 59: I noted that this doesn't extended the abstract class earlier in the PR, which means you had to modify the constructor and introduce a `getDocumentPersister()` method.

Out of curiosity, why can't it inherit the base class like other stages?

tests/Doctrine/ODM/MongoDB/Tests/Functional/AtlasSearchTest.php
```diff
@@ -53,7 +53,7 @@ public function testAtlasSearch(): void
         $results = $this->dm->createAggregationBuilder(CmsArticle::class)
             ->search()
                 ->index('search_articles')
-                ->text()
+                ->autocomplete()
```
jmikola, line 56: What was the purpose of this change?

tests/Doctrine/ODM/MongoDB/Tests/SchemaManagerTest.php
```diff
@@ -410,22 +410,49 @@ public function testCreateSearchIndexes(): void
...
-                    ->with([
-                        [
-                            'definition' => ['mappings' => ['dynamic' => true]],
-                            'name' => 'search_articles',
-                            'type' => 'search',
-                        ],
-                    ])
-                    ->willReturn(['search_articles']);
+                    ->with($this->anything())
+                    ->willReturnCallback(function (array $indexes) {
```
jmikola, line 436: Just to confirm: you're mocking the response here to avoid requiring an Atlas cluster, correct?

lib/Doctrine/ODM/MongoDB/SchemaManager.php
```diff
@@ -486,6 +487,59 @@ public function deleteDocumentSearchIndexes(string $documentName): void
         }
     }
 
+    /**
+     * @param ClassMetadata<object> $class
+     *
+     * @phpstan-return list<SearchIndexMapping>
+     */
+    private function prepareSearchIndexes(ClassMetadata $class): array
```
jmikola, line 495: Noted that this is shared by the create and update methods above.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/Compound.php
GromNaN: An oversight: initially, I had not created the parent method `AbstractSearchOperator::getDocumentPersister()`. Removed the overrides.

tests/Doctrine/ODM/MongoDB/Tests/SchemaManagerTest.php
GromNaN: Yes, that's a unit test that doesn't depend on a MongoDB server.

tests/Doctrine/ODM/MongoDB/Tests/Functional/AtlasSearchTest.php
GromNaN: I added `autocomplete` type to the `search_articles` index to validate the field name was correctly translated. Only the `autocomplete` operator works with this field type.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/MoreLikeThis.php
GromNaN: Initially, I didn't want to expose the document persister to the child classes, but that was necessary for chaining operators. I moved this to a new `AbstractSearchOperator::prepareDocuments` so that operator classes don't have to hold such logic.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/Compound.php
jmikola: I don't think this file was updated, as I still see `private DocumentPersister` here with a reference to `$this->persister` in the parent constructor invocation. And the overridden `getDocumentPersister()` method is still below.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search.php
jmikola: cc: @GromNaN

jmikola (approved):
Outstanding question in `Compound.php`, which may apply to any other classes that had a similar behavior of assigning their own private DocumentPersister instance.

Also, a question in `Search.php` wasn't addressed. No changes required there but an answer would help my understanding.

I'll LGTM as the "private DocumentPersister" cleanup is straightforward.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search.php
GromNaN: I think because we don't want that the `Search` class implement `SearchOperator`, and the methods implemented by `AbstractSearchOperator` delegate to the  `Search` class, while the `Search` class modifies its own state.

PR #2832, opened 2025-09-25
PHPORM-395: Fix autocompletion for compound search operators

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | 

#### Summary

For some reason, PhpStorm didn't like the `@return` annotation on the method, which should've told it that the result of a `text()` operator is a special `Text` instance that allows calling other operators. To work around this, using the compounded classes is necessary, which is what's actually being returned. Since this effectively narrows the return type (as that class extends the original operator class but implements an extra interface) we can do this in the trait without running into issues.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/SupportsCompoundableOperatorsTrait.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/SupportsCompoundableOperatorsTrait.php
@@ -45,56 +45,40 @@ trait SupportsCompoundableOperatorsTrait
      */
     abstract protected function addOperator(SearchOperator $operator): SearchOperator;
 
-    /** return Autocomplete&CompoundSearchOperatorInterface */
-    public function autocomplete(string $path = '', string ...$query): Autocomplete
+    public function autocomplete(string $path = '', string ...$query): CompoundedAutocomplete
     {
         return $this->addOperator(new CompoundedAutocomplete($this->getCompoundStage(), $this->getAddOperatorClosure(), $this->getSearchStage(), $path, ...$query));
     }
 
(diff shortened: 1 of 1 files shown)


GromNaN:
I just verified that this fixes the autocompletion issue in PHPStorm.

I understand why you didn't do this initially, as you shouldn't depend on an internal class in the API of a public class. Therefore, classes like `CompoundedAutocomplete` should no longer be marked internal.

NB: It seems we have a new range of errors to exclude from the phpstan analysis.

alcaeus:
> Therefore, classes like `CompoundedAutocomplete` should no longer be marked internal.

Users shouldn't ever need to depend on those classes, unless they store the return value of somewhere within the chain, which I generally wouldn't recommend. No objection, I just don't think it's necessary.

> NB: It seems we have a new range of errors to exclude from the phpstan analysis.

Created #2833 to fix those separately.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/SupportsCompoundableOperatorsTrait.php
```diff
@@ -45,103 +45,79 @@ abstract protected function getAddOperatorClosure(): Closure;
      */
     abstract protected function addOperator(SearchOperator $operator): SearchOperator;
 
-    /** return Autocomplete&CompoundSearchOperatorInterface */
```
GromNaN, line 48: Just seen the missing `@` before `return`.

GromNaN (changes requested):
Sorry, I found a better solution: https://github.com/doctrine/mongodb-odm/pull/2836

alcaeus:
Closing in favour of #2836.

PR #2836, opened 2025-09-25
Fix return type of compound search operators

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | [PHPORM-395](https://jira.mongodb.org/browse/PHPORM-395)

#### Summary

When `Compound::autocomplete()` returns `Autocomplete&CompoundSearchOperatorInterface` it inherits from the methods of `CompoundSearchOperatorInterface`. So the methods for search operator must specify that a the returned operator is also an instance of `CompoundSearchOperatorInterface`.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/CompoundSearchOperatorInterface.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/CompoundSearchOperatorInterface.php
@@ -30,7 +30,7 @@ interface CompoundSearchOperatorInterface extends SupportsCompoundableSearchOper
 
     public function exists(string $path): Exists&CompoundSearchOperatorInterface;
 
-    /** @param LineString|Point|Polygon|MultiPolygon|array|null $geometry */
+    /** @param LineString|Point|Polygon|MultiPolygon|array<string, mixed>|null $geometry */
     public function geoShape($geometry = null, string $relation = '', string ...$path): GeoShape&CompoundSearchOperatorInterface;
 
     public function geoWithin(string ...$path): GeoWithin&CompoundSearchOperatorInterface;
@@ -39,8 +39,8 @@ interface CompoundSearchOperatorInterface extends SupportsCompoundableSearchOper
     public function moreLikeThis(...$documents): MoreLikeThis&CompoundSearchOperatorInterface;
 
     /**
-     * @param int|float|UTCDateTime|array|Point|null $origin
-     * @param int|float|null                         $pivot
+     * @param int|float|UTCDateTime|array<string, mixed>|Point|null $origin
+     * @param int|float|null                                        $pivot
      */
     public function near($origin = null, $pivot = null, string ...$path): Near&CompoundSearchOperatorInterface;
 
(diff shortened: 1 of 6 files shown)


lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Search/CompoundSearchOperatorInterface.php
```diff
@@ -4,6 +4,13 @@
...
 namespace Doctrine\ODM\MongoDB\Aggregation\Stage\Search;
 
+use GeoJson\Geometry\LineString;
+use GeoJson\Geometry\MultiPolygon;
+use GeoJson\Geometry\Point;
+use GeoJson\Geometry\Polygon;
+use MongoDB\BSON\ObjectId;
+use MongoDB\BSON\UTCDateTime;
+
 interface CompoundSearchOperatorInterface extends SupportsCompoundableSearchOperators
```
GromNaN, line 14: If we don't want to repeat all the methods, we could make `SupportsCompoundableSearchOperators` generic and have a return type declared as `Autocomplete&T`. But that would not enforce the type as we can do with concrete return types.

GromNaN: That would mean an `@extends` tag for each interface implemented by `SupportsAutocompleteOperator`, which is worse.

```php
/**
 * @template T of object
 * @extends SupportsAutocompleteOperator<T>
 * @extends SupportsEmbeddedDocumentOperator<T>
 * @extends SupportsEqualsOperator<T>
 * @extends SupportsExistsOperator<T>
 * @extends SupportsGeoShapeOperator<T>
 * @extends SupportsGeoWithinOperator<T>
 * @extends SupportsMoreLikeThisOperator<T>
 * @extends SupportsNearOperator<T>
 * @extends SupportsPhraseOperator<T>
 * @extends SupportsQueryStringOperator<T>
 * @extends SupportsRangeOperator<T>
 * @extends SupportsRegexOperator<T>
 * @extends SupportsTextOperator<T>
 * @extends SupportsWildcardOperator<T>
 */
interface SupportsCompoundableSearchOperators extends SupportsAutocompleteOperator, SupportsEmbeddedDocumentOperator, SupportsEqualsOperator, SupportsExistsOperator, SupportsGeoShapeOperator, SupportsGeoWithinOperator, SupportsMoreLikeThisOperator, SupportsNearOperator, SupportsPhraseOperator, SupportsQueryStringOperator, SupportsRangeOperator, SupportsRegexOperator, SupportsTextOperator, SupportsWildcardOperator
{
}
```

```php
/** @template T of object */
interface SupportsAutocompleteOperator
{
    /** @return Autocomplete&T */
    public function autocomplete(string $path = '', string ...$query): Autocomplete;
}
```

alcaeus: While I would typically go for a templated interface, I agree that the list of `@extends` annotations looks ugly. In that case, I'd rather update the method declarations in this interface, especially since we now have the ability to define the return types natively.

alcaeus (approved):

PR #2840, opened 2025-09-26
Add support for PHP 8.4 Native Lazy Objects with configuration flag

### LINKED ISSUE #2828: Migrate to Native Lazy Objects
|    Q        |   A
|------------ | -----
| Version     | 2.3.x

### Support Question

Are there any plans to migrate this bundle to native lazy objects which were introduced with php 8.4? Besides that being a pure technical "support" question it also has a real-world impact:  symfony/var-exporter has deprecated the use of LazyGhostTrait  in 7.3 and when running the combination of doctrine/mongodb-odm, var-exporter >= 7.3.0,  and php >= 8.4.0 logs are spammed with deprecation warnings. It seems doctrine/orm already did that change.

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | [PHPORM-321](https://jira.mongodb.org/browse/PHPORM-321) Fix https://github.com/doctrine/mongodb-odm/issues/2828

#### Summary

Shamelessly copied from the brilliant work of Benjamin Eberlei on ORM:
- PropertyAccessor: https://github.com/doctrine/orm/pull/11659
- Native Lazy Objects: https://github.com/doctrine/orm/pull/11853


lib/Doctrine/ODM/MongoDB/Mapping/PropertyAccessors/ObjectCastPropertyAccessor.php
```diff
@@ -0,0 +1,63 @@
...
+
+    public function setValue(object $object, mixed $value): void
+    {
+        if ($object instanceof InternalProxy && ! $object->__isInitialized()) {
+            $object->__setInitialized(true);
+            $this->reflectionProperty->setValue($object, $value);
+            $object->__setInitialized(false);
+        } elseif ($object instanceof GhostObjectInterface && ! $object->isProxyInitialized()) {
+            $initializer = $object->getProxyInitializer();
+            $object->setProxyInitializer();
```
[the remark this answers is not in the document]
GromNaN: `null` is the default value. https://github.com/FriendsOfPHP/proxy-manager-lts/blob/2c8a6cffc3220e99352ad958fe7cf06bf6f7690f/src/ProxyManager/Proxy/GhostObjectInterface.php#L32

lib/Doctrine/ODM/MongoDB/Hydrator/HydratorFactory.php
```diff
@@ -450,6 +451,10 @@ public function hydrate(object $document, array $data, array $hints = []): array
             }
         }
 
+        if (PHP_VERSION_ID >= 80400) {
+            $metadata->reflClass->markLazyObjectAsInitialized($document);
```
alcaeus, line 455: Noted that this is a no-op if `$document` is not a lazy object.

lib/Doctrine/ODM/MongoDB/Mapping/PropertyAccessors/PropertyAccessorFactory.php
```diff
@@ -0,0 +1,32 @@
...
+
+declare(strict_types=1);
+
+namespace Doctrine\ODM\MongoDB\Mapping\PropertyAccessors;
+
+use ReflectionProperty;
+
+use const PHP_VERSION_ID;
+
+class PropertyAccessorFactory
```
alcaeus, line 12: This is not annotated as an internal class.

```suggestion
/** @internal */
class PropertyAccessorFactory
```

lib/Doctrine/ODM/MongoDB/Mapping/PropertyAccessors/RawValuePropertyAccessor.php
```diff
@@ -0,0 +1,65 @@
...
+use const PHP_VERSION_ID;
+
+/**
+ * This is a PHP 8.4 and up only class and replaces {@see ObjectCastPropertyAccessor}.
+ *
+ * It works based on the raw values of a property, which for a case of property hooks
+ * is the backed value. If we kept using setValue/getValue, this would go through the hooks,
+ * which potentially change the data.
+ */
+class RawValuePropertyAccessor implements PropertyAccessor
```
alcaeus, line 25: This is not annotated as an internal class.

```suggestion
 *  
 * @internal
 */
class RawValuePropertyAccessor implements PropertyAccessor
```

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -638,10 +642,15 @@
     /**
      * The ReflectionProperty instances of the mapped class.
      *
-     * @var ReflectionProperty[]
+     * @deprecated Since 2.13, use $propertyAccessors instead.
+     *
+     * @var LegacyReflectionFields|array<ReflectionProperty>
```
alcaeus, line 647: This is a nice way to emit deprecation notices, but it is a minor BC break. Previously this was type-hinted as an array, whereas we now widen this to include `LegacyReflectionFields`. While this new class can be used like an array, anything that expects an `array` type and received these fields will fail going forward. It may be irrelevant, but I figured I'd point it out.

lib/Doctrine/ODM/MongoDB/Mapping/LegacyReflectionFields.php
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    /**
+     * @param string $field
+     *
+     * @psalm-suppress LessSpecificImplementedReturnType
+     */
+    public function offsetGet($field): mixed // phpcs:ignore
+    {
+        if (isset($this->reflFields[$field])) {
```
alcaeus, line 53: The early return here means that not all code instances where `reflFields` is accessed will be exposed through deprecations. Is that intentional?

lib/Doctrine/ODM/MongoDB/Mapping/LegacyReflectionFields.php
```diff
@@ -0,0 +1,136 @@
...
+        trigger_deprecation('doctrine/mongodb-odm', '2.14', 'Access to ClassMetadata::$reflFields is deprecated and will be removed in Doctrine ODM 3.0.');
+
+        if (! isset($this->classMetadata->propertyAccessors[$field])) {
+            throw new OutOfBoundsException('Unknown field: ' . $this->classMetadata->name . ' ::$' . $field);
+        }
+
+        $className = $this->classMetadata->fieldMappings[$field]['inherited']
+            ?? $this->classMetadata->fieldMappings[$field]['declared']
+            ?? $this->classMetadata->associationMappings[$field]['declared']
+            ?? $this->classMetadata->name;
```
alcaeus, line 66: Looking at this just made me realise what size the metadata mess actually is. WTF.

lib/Doctrine/ODM/MongoDB/Proxy/Factory/NativeLazyObjectFactory.php
```diff
@@ -0,0 +1,97 @@
...
+        $metadata->propertyAccessors[$metadata->identifier]->setValue($proxy, $identifier);
+
+        if (isset(self::$lazyObjects)) {
+            self::$lazyObjects[$proxy] = true;
+        }
+
+        return $proxy;
+    }
+
+    /** Only for internal tests */
```
alcaeus:
```suggestion
    /** @internal Only for internal tests */
```

lib/Doctrine/ODM/MongoDB/Proxy/Factory/NativeLazyObjectFactory.php
```diff
@@ -0,0 +1,97 @@
...
+    public static function enableTracking(bool $enabled = true): void
+    {
+        if ($enabled) {
+            self::$lazyObjects ??= new WeakMap();
+        } else {
+            self::$lazyObjects = null;
+        }
+    }
+
+    /** Only for internal tests */
```
alcaeus:
```suggestion
    /** @internal Only for internal tests */
```

lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -686,26 +689,49 @@ public function isTransactionalFlushEnabled(): bool
      * Generate proxy classes using Symfony VarExporter's LazyGhostTrait if true.
      * Otherwise, use ProxyManager's LazyLoadingGhostFactory (deprecated)
      */
-    public function setUseLazyGhostObject(bool $flag): void
+    public function setLazyGhostObject(bool $flag): void
```
alcaeus: `setUseLazyGhostObject` needs to remain as it's part of the public API.

lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -686,26 +689,49 @@ public function isTransactionalFlushEnabled(): bool
...
+    }
+
+    public function enableNativeLazyObjects(bool $nativeLazyObjects): void
+    {
+        if (PHP_VERSION_ID >= 80400 && ! $nativeLazyObjects) {
+            trigger_deprecation('doctrine/mongodb-odm', '2.14', 'Disabling native lazy objects is deprecated and will be impossible in Doctrine MongoDB ODM 3.0.');
+        }
+
+        if (PHP_VERSION_ID < 80400 && $nativeLazyObjects) {
+            throw new LogicException('Lazy loading proxies require PHP 8.4 or higher.');
```
alcaeus:
```suggestion
            throw new LogicException('Native lazy objects require PHP 8.4 or higher.');
```

lib/Doctrine/ODM/MongoDB/DocumentManager.php
```diff
@@ -607,7 +609,7 @@ public function flush(array $options = []): void
      * @param mixed           $identifier
      * @param class-string<T> $documentName
      *
-     * @return T|(T&GhostObjectInterface<T>)
+     * @return T
```
alcaeus, line 612: Technically the old return type is still accurate until we remove other proxy mechanisms.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
```diff
@@ -1990,7 +1992,7 @@ private function doMerge(object $document, array &$visited, ?object $prevManaged
                                         ->dm
                                         ->getProxyFactory()
                                         ->getProxy($targetClass, $relatedId);
-                                    $this->registerManaged($other, $relatedId, []);
+                                    $this->registerManaged($other, $relatedId, [$targetClass->identifier => $relatedId]);
```
alcaeus, line 1995: I'll admit I'm too lazy to look into the logic, but what necessitated this change?

tests/Doctrine/ODM/MongoDB/Tests/Functional/PropertyHooksTest.php
```diff
@@ -0,0 +1,100 @@
...
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        if ($this->dm->getConfiguration()->isNativeLazyObjectsEnabled()) {
+            return;
+        }
+
+        $this->markTestSkipped('Property hooks require native lazy objects to be enabled.');
```
alcaeus: Did phpcs ask for the early return? I find the following easier to read...

```suggestion
        if ($this->dm->getConfiguration()->isNativeLazyObjectsEnabled()) {
            $this->markTestSkipped('Property hooks require native lazy objects to be enabled.');
        }
```

tests/Doctrine/ODM/MongoDB/Tests/Functional/ReferencePrimerTest.php
```diff
@@ -94,7 +94,6 @@ public function testPrimeReferencesWithDBRefObjects(): void
             ->field('groups')->prime(true);
 
         foreach ($qb->getQuery() as $user) {
-            self::assertTrue(self::isLazyObject($user->getAccount()));
```
alcaeus, line 97: Noted that this change is required because we cannot distinguish an initialised lazy object from a non-lazy object

tests/Doctrine/ODM/MongoDB/Tests/Mapping/AbstractAnnotationDriverTestCase.php
```diff
@@ -96,8 +97,6 @@ public function testGetAllClassNamesIsIdempotent(): void
     {
         $annotationDriver = $this->loadDriverForCMSDocuments();
         $original         = $annotationDriver->getAllClassNames();
-
-        $annotationDriver = $this->loadDriverForCMSDocuments();
```
alcaeus, line 100: Why the change here? Looking at the test, it feels like this test was actually supposed to test the result with a fresh driver, but I may be mistaken.

lib/Doctrine/ODM/MongoDB/Mapping/PropertyAccessors/PropertyAccessorFactory.php
GromNaN: It might be necessary for external package: https://github.com/doctrine-extensions/DoctrineExtensions/blob/eb53dfcb2b592327b76ac5226fbb003d32aea37e/src/Tree/Strategy/ORM/Closure.php#L138-L142

lib/Doctrine/ODM/MongoDB/Proxy/Factory/NativeLazyObjectFactory.php
GromNaN: The class is already internal.

lib/Doctrine/ODM/MongoDB/UnitOfWork.php
GromNaN: When all the properties of a lazy object are initialized using reflection, the object is no longer flagged as uninitialized and the initialization callback is not run.

This change ensures we support documents with a single ID property.

tests/Doctrine/ODM/MongoDB/Tests/Functional/PropertyHooksTest.php
GromNaN: Yes, PHPCS. I hate this rule. We need a way to disable it when the condition only contains 1 statement.

tests/Doctrine/ODM/MongoDB/Tests/Functional/ReferencePrimerTest.php
GromNaN: The subsequent assertion is sufficient. This date from the time we add to check the interface `GhostObjectInterface` calling the method `isProxyInitialized`.
https://github.com/doctrine/mongodb-odm/blame/0aa0e7b9ee31148a0e522aa1fb42fb7f4459c2c8/tests/Doctrine/ODM/MongoDB/Tests/Functional/ReferencePrimerTest.php#L98-L99C73

Reverted.

lib/Doctrine/ODM/MongoDB/Mapping/PropertyAccessors/PropertyAccessorFactory.php
alcaeus: I'd mark it as internal so we can then remove it before the next major version. I'd still like to see this logic either in doctrine/persistence, doctrine/reflection, or a new doctrine/property-accessor package.

tests/Doctrine/ODM/MongoDB/Tests/Functional/PropertyHooksTest.php
alcaeus: The [EarlyExit sniff](https://github.com/slevomat/coding-standard/blob/master/doc/control-structures.md#slevomatcodingstandardcontrolstructuresearlyexit-) supports the `ignoreStandaloneIfInScope`, `ignoreOneLineTrailingIf`, and `ignoreTrailingIfWithOneInstruction` options that allow us to customise this. I believe the `ignoreTrailingIfWithOneInstruction` option allows this kind of usage, but I also don't mind enabling all of them.

lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php
```diff
@@ -1963,9 +1992,11 @@ public function setFieldValue(object $document, string $field, $value): void
             $document->__load();
         } elseif ($document instanceof GhostObjectInterface && ! $document->isProxyInitialized()) {
             $document->initializeProxy();
+        } elseif (PHP_VERSION_ID >= 80400 && $this->reflClass->isUninitializedLazyObject($document)) {
```
GromNaN:
```suggestion
        } elseif (PHP_VERSION_ID >= 80400) {
```

lib/Doctrine/ODM/MongoDB/Configuration.php
GromNaN: I rename the env var for tests to remote the S

lib/Doctrine/ODM/MongoDB/DocumentManager.php
GromNaN: We don't want people to rely on the internal implementation of the proxy.

lib/Doctrine/ODM/MongoDB/Mapping/LegacyReflectionFields.php
```diff
@@ -0,0 +1,136 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ODM\MongoDB\Mapping;
+
+use ArrayAccess;
+use Countable;
+use Doctrine\ORM\Mapping\ReflectionReadonlyProperty;
```
alcaeus: This class is missing. There's a `ReadonlyAccessor`, but in my project with readonly properties I'm getting an error because of this missing class.

```suggestion
use Doctrine\ODM\MongoDB\Mapping\ReflectionReadonlyProperty;
```

alcaeus (changes requested):

lib/Doctrine/ODM/MongoDB/Mapping/LegacyReflectionFields.php
GromNaN: Good catch, I imported the missing class from ORM.

alcaeus (approved):
Tested in one of my projects, works fine with readonly properties and such. Great work!

PR #2841, opened 2025-09-30
Add an assertion on missing search index when a `$search` or `$vectorSearch` pipeline returns an empty result

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | maybe
| Fixed issues | [PHPORM-388](https://jira.mongodb.org/browse/PHPORM-388)

#### Summary

When performing a search query on a search index that doesn't exist, the aggregation returns no results. This is frustrating for users that will try to optimize the query while the issue is the missing search index.

This PR adds a configuration setting (enabled by default) to check if the search index exists only when performing a search aggregation with no returned results.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Aggregation.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Aggregation.php
@@ -11,9 +11,11 @@ use Doctrine\ODM\MongoDB\Iterator\IterableResult;
 use Doctrine\ODM\MongoDB\Iterator\Iterator;
 use Doctrine\ODM\MongoDB\Iterator\UnrewindableIterator;
 use Doctrine\ODM\MongoDB\Mapping\ClassMetadata;
+use Doctrine\ODM\MongoDB\SchemaException;
 use MongoDB\Collection;
 use MongoDB\Driver\CursorInterface;
 
+use function array_key_first;
 use function array_merge;
 
 /** @phpstan-import-type PipelineExpression from Builder */
@@ -64,6 +66,48 @@ final class Aggregation implements IterableResult
             $cursor = new HydratingIterator($cursor, $this->dm->getUnitOfWork(), $this->classMetadata);
         }
 
-        return $this->rewindable ? new CachingIterator($cursor) : new UnrewindableIterator($cursor);
+        $iterator = $this->rewindable ? new CachingIterator($cursor) : new UnrewindableIterator($cursor);
+
+        $this->assertSearchIndexExistsWhenAggregationResultsIsEmpty($iterator);
+
+        return $iterator;
+    }
+
+    /**
+     * If the server implements a server-side error for missing search indexes,
+     * this assertion can be removed.
+     *
+     * @see https://jira.mongodb.org/browse/SERVER-110974
+     *
+     * @param Iterator<object> $iterator
+     */
+    private function assertSearchIndexExistsWhenAggregationResultsIsEmpty(Iterator $iterator): void
+    {
+        if ($iterator->current() !== false) {
+            return; // Results not empty
+        }
+
+        if (! $this->dm->getConfiguration()->assertSearchIndexExistsWhenAggregationResultsIsEmpty()) {
+            return; // Feature disabled
+        }
+
+        // Search stages must be the first stage in the pipeline
+        $indexName = match (array_key_first($this->pipeline[0])) {
+            '$search' => $this->pipeline[0]['$search']->index ?? null,
+            '$searchMeta' => $this->pipeline[0]['$searchMeta']->index ?? null,
+            '$vectorSearch' => $this->pipeline[0]['$vectorSearch']->index ?? null,
+            default => null,
+        };
+
+        if ($indexName === null) {
+            return; // Not a search aggregation or index not specified
+        }
+
+        $index = $this->collection->listSearchIndexes(['filter' => ['name' => $indexName]])->current();
+        if ($index) {
+            return; // Index exists
+        }
+
+        throw SchemaException::searchIndexNotFound($this->collection->getNamespace(), $indexName);
     }
 }
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Aggregation.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Aggregation.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Doctrine\ODM\MongoDB\Aggregation;
 
+use Doctrine\ODM\MongoDB\Configuration;
 use Doctrine\ODM\MongoDB\DocumentManager;
 use Doctrine\ODM\MongoDB\Iterator\CachingIterator;
 use Doctrine\ODM\MongoDB\Iterator\HydratingIterator;
@@ -15,8 +16,10 @@ use Doctrine\ODM\MongoDB\SchemaException;
 use MongoDB\Collection;
 use MongoDB\Driver\CursorInterface;
 
-use function array_key_first;
 use function array_merge;
+use function current;
+use function in_array;
+use function key;
 
 /** @phpstan-import-type PipelineExpression from Builder */
 final class Aggregation implements IterableResult
@@ -68,7 +71,7 @@ final class Aggregation implements IterableResult
 
         $iterator = $this->rewindable ? new CachingIterator($cursor) : new UnrewindableIterator($cursor);
 
-        $this->assertSearchIndexExistsWhenAggregationResultsIsEmpty($iterator);
+        $this->assertSearchIndexExistsForEmptyResult($iterator);
 
         return $iterator;
     }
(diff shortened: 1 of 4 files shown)


lib/Doctrine/ODM/MongoDB/Configuration.php
```diff
@@ -808,6 +808,23 @@ private function getAutoEncryptionOptions(): array
...
     }
+
+    /**
+     * Pipelines using a search index that does not exist or is not queryable
+     * will return zero documents. By enabling this feature, an additional query
+     * is made when the pipeline doesn't return any results to check if the
+     * search index exists and is queryable. If the index does not exist or is
+     * not queryable, an exception is thrown. This feature is enabled by default.
+     */
+    public function setAssertSearchIndexExistsWhenAggregationResultsIsEmpty(bool $enabled): void
```
GromNaN: Any better name idea?

lib/Doctrine/ODM/MongoDB/Aggregation/Aggregation.php
```diff
@@ -64,6 +66,48 @@ private function prepareIterator(CursorInterface $cursor): Iterator
...
+        if ($indexName === null) {
+            return; // Not a search aggregation or index not specified
+        }
+
+        $index = $this->collection->listSearchIndexes(['filter' => ['name' => $indexName]])->current();
+        if ($index) {
+            return; // Index exists
+        }
+
+        throw SchemaException::searchIndexNotFound($this->collection->getNamespace(), $indexName);
```
GromNaN: I don't have to check `$index['queryable']`: the service returns an error if the index exists but is not queryable.

lib/Doctrine/ODM/MongoDB/Aggregation/Aggregation.php
```diff
@@ -64,6 +66,48 @@ private function prepareIterator(CursorInterface $cursor): Iterator
...
+
+        if (! $this->dm->getConfiguration()->assertSearchIndexExistsWhenAggregationResultsIsEmpty()) {
+            return; // Feature disabled
+        }
+
+        // Search stages must be the first stage in the pipeline
+        $indexName = match (array_key_first($this->pipeline[0])) {
+            '$search' => $this->pipeline[0]['$search']->index ?? null,
+            '$searchMeta' => $this->pipeline[0]['$searchMeta']->index ?? null,
+            '$vectorSearch' => $this->pipeline[0]['$vectorSearch']->index ?? null,
```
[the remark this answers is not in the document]
jmikola: I assume Copilot is overlooking how `??` works in this context, yes? I wouldn't expect an error for any missing keys/properties.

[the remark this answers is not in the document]
jmikola: This seems like it could be a legitimate concern. This class assigns `private array $pipeline` in its constructor.

Is there any earlier assertion to guarantee that the value of each pipeline stage will be an object?

lib/Doctrine/ODM/MongoDB/Aggregation/Aggregation.php
jmikola: What is the "service"?

The docs for [`$listSearchIndexes`](https://www.mongodb.com/docs/manual/reference/operator/aggregation/listSearchIndexes/) suggest that a `queryable` boolean will be reported in the results. Are you suggesting that `queryable: false` never gets reported?

lib/Doctrine/ODM/MongoDB/Configuration.php
jmikola: Should the comment clarify that this applies to `$search`, `$searchMeta`, and `$vectorSearch` stages?

"WhenAggregationResultsIsEmpty" could be shortened to "ForEmptyResult" (can also replace "for" with "after" or "on") if we're OK assuming that any "search" implies the use of aggregation. So:

 * `setAssertSearchIndexExistsForEmptyResult`

lib/Doctrine/ODM/MongoDB/Aggregation/Aggregation.php
```diff
@@ -64,6 +66,48 @@ private function prepareIterator(CursorInterface $cursor): Iterator
             $cursor = new HydratingIterator($cursor, $this->dm->getUnitOfWork(), $this->classMetadata);
         }
 
-        return $this->rewindable ? new CachingIterator($cursor) : new UnrewindableIterator($cursor);
+        $iterator = $this->rewindable ? new CachingIterator($cursor) : new UnrewindableIterator($cursor);
```
jmikola, line 72: Noted that both [CachingIterator](https://github.com/doctrine/mongodb-odm/blob/2.12.2/lib/Doctrine/ODM/MongoDB/Iterator/CachingIterator.php#L54) and [UnrewindableIterator](https://github.com/doctrine/mongodb-odm/blob/2.12.2/lib/Doctrine/ODM/MongoDB/Iterator/UnrewindableIterator.php#L42) rewind the inner iterator upon construction, so it's kosher to immediately call `current()` in the assertion function; however, you should probably make a note of this.

I think the best way to capture this would be to use union types on the assertion method's signature and add a docblock there that notes the point above.

Otherwise, we could not be certain that an arbitrary Iterator instance would be rewound already and capable of accessing `current()`.

lib/Doctrine/ODM/MongoDB/Aggregation/Aggregation.php
```diff
@@ -64,6 +66,48 @@ private function prepareIterator(CursorInterface $cursor): Iterator
...
+     * If the server implements a server-side error for missing search indexes,
+     * this assertion can be removed.
+     *
+     * @see https://jira.mongodb.org/browse/SERVER-110974
+     *
+     * @param Iterator<object> $iterator
+     */
+    private function assertSearchIndexExistsWhenAggregationResultsIsEmpty(Iterator $iterator): void
+    {
+        if ($iterator->current() !== false) {
```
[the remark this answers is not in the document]
jmikola: I think Copilot is correct here. Assuming we end up propagating the return value from [`MongoDB\Driver\Cursor::current()`](https://www.php.net/manual/en/mongodb-driver-cursor.current.php), it will either be an array/object or `null`.

Another random thought: if the pipeline happened to yield a tailable cursor (e.g. `$changeStream`), the result might also be empty until the first `getMore`; however, that also wouldn't have anything to do with a search index so we should be fine.

lib/Doctrine/ODM/MongoDB/Aggregation/Aggregation.php
```diff
@@ -64,6 +66,48 @@ private function prepareIterator(CursorInterface $cursor): Iterator
...
+            return; // Results not empty
+        }
+
+        if (! $this->dm->getConfiguration()->assertSearchIndexExistsWhenAggregationResultsIsEmpty()) {
```
jmikola: Consider doing this check before you even invoke the assert function. Then we can avoid the `current()` call above. If you do want to leave this check within this method, I think we can at least make it the first line.

GromNaN: Since this setting is enabled by default, I don't think it's an optimization to move it earlier.

lib/Doctrine/ODM/MongoDB/Configuration.php
GromNaN: I like this shorter name — Updated.


> Should the comment clarify that this applies to `$search`, `$searchMeta`, and `$vectorSearch` stages?

I don't think that needs clarification. It's the stages that use a search index.

lib/Doctrine/ODM/MongoDB/Aggregation/Aggregation.php
GromNaN: > What is the "service"?

Sorry, the "server".


> The docs for [`$listSearchIndexes`](https://www.mongodb.com/docs/manual/reference/operator/aggregation/listSearchIndexes/) suggest that a `queryable` boolean will be reported in the results. Are you suggesting that `queryable: false` never gets reported?

`queryable: false` is reported by `$listSearchIndexes`, but the server will return an error before we reach this code when performing a search aggregation on a search index that is not queryable.

lib/Doctrine/ODM/MongoDB/Aggregation/Aggregation.php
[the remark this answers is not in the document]
GromNaN: This function uses `current()` which returns `false` when there is nothing (confirmed by tests)

Replaced by `! $iterator->current()`, so that we support both `null|false`. When there is a value, it is `array|object`.

lib/Doctrine/ODM/MongoDB/Aggregation/Aggregation.php
```diff
@@ -64,6 +68,45 @@ private function prepareIterator(CursorInterface $cursor): Iterator
...
+            return; // Feature disabled
+        }
+
+        // Search stages must be the first stage in the pipeline
+        $stage = $this->pipeline[0] ?? null;
+        if (! $stage || ! in_array(array_key_first($stage), ['$search', '$searchMeta', '$vectorSearch'], true)) {
+            return; // Not a search aggregation
+        }
+
+        $indexName = ((object) $stage[array_key_first($stage)])->index ?? 'default';
```
GromNaN, line 104: In case of custom stage class that return an array instead of an object, we cast it.

GromNaN, line 104: We can use `array_first`, that would require the [PHP 8.5 ployfill](https://github.com/symfony/polyfill-php85/blob/65345afb52bf45a3d996295b3504ace5db1be853/Php85.php#L37-L44). Not worth it.
```suggestion
        $indexName = ((object) array_first($stage))->index ?? 'default';
```

lib/Doctrine/ODM/MongoDB/Aggregation/Aggregation.php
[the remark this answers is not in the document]
jmikola: I'm not sure where a `false` return value comes from, but that seems incorrect. If this isn't coming from the driver, I wonder if there's a non-compliant Iterator implementation in ODM responsible for returning `false`.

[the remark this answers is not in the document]
GromNaN: That's an [inconsistency](https://3v4l.org/JbXVc) between the `current()` function and the `Iterator::current()` function. I'll change to use the `Iterator::key()` method which is more reliable to detect if there is an object, as said in the [documentation note](https://www.php.net/function.current#refsect1-function.current-notes).

lib/Doctrine/ODM/MongoDB/Aggregation/Aggregation.php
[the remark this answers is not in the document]
GromNaN: That was the call to `array_key_first($this->pipeline[0])`. Refactored.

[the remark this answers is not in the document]
GromNaN: Refactored.

lib/Doctrine/ODM/MongoDB/Aggregation/Aggregation.php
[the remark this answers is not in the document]
jmikola: Oh, I overlooked that you were referring to the `current()` array function. I thought we were just discussing the iterator method here.

jmikola (approved):

PR #2842, opened 2025-09-30
Add early return to `DocumentManager` getters of MongoDB client objects

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | -

#### Summary

Skip unnecessary calls to `$this->metadataFactory->getMetadataFor()` when the Database/Collection/GridFS Bucket object is already instantiated.

This should improve performance when calling this methods multiple times.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/DocumentManager.php
+++ b/lib/Doctrine/ODM/MongoDB/DocumentManager.php
@@ -322,20 +322,17 @@ class DocumentManager implements ObjectManager
      */
     public function getDocumentDatabase(string $className): Database
     {
-        $metadata = $this->metadataFactory->getMetadataFor($className);
-
-        $className = $metadata->getName();
-
         if (isset($this->documentDatabases[$className])) {
             return $this->documentDatabases[$className];
         }
 
-        $db                                  = $metadata->getDatabase();
-        $db                                  = $db ?: $this->config->getDefaultDB();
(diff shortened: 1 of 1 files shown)


lib/Doctrine/ODM/MongoDB/DocumentManager.php
```diff
@@ -322,20 +322,17 @@ public function getClassMetadata($className): ClassMetadata
      */
     public function getDocumentDatabase(string $className): Database
     {
-        $metadata = $this->metadataFactory->getMetadataFor($className);
```
GromNaN, line 325: I must be wrong, since the `getMetadataFor` method normalizes the class name. 

https://github.com/doctrine/persistence/blob/68ea6fbe6080a7dff0e21d832324b18e97c264f9/src/Persistence/Mapping/AbstractClassMetadataFactory.php#L155-L161

alcaeus: Yep, that's so you can use `$document::class` and have it work with proxy objects. We also used to have aliases (e.g. `App:User`) that we no longer support, but that also needed normalisation.

GromNaN: Now, we should normalize the other way, as using `::class` it the norm. And have a shortcut.

PR #2849, opened 2025-10-08
Use docs-builder package for building docs

paulinevos:
|      Q       |   A
|------------- | -----------
| Type         | task
| BC Break     | no
| Fixed issues |

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -15,30 +15,6 @@ on:
       - docs/**
 
 jobs:
-  validate-with-guides:
-    name: "Validate documentation with phpDocumentor/guides"
-    runs-on: "ubuntu-24.04"
-
-    steps:
-      - name: "Checkout code"
-        uses: "actions/checkout@v5"
-
-      - name: "Install PHP"
-        uses: "shivammathur/setup-php@v2"
-        with:
-          coverage: "none"
-          php-version: "8.3"
-
-      - name: "Install dependencies with Composer"
-        uses: "ramsey/composer-install@v3"
-        with:
-          working-directory: "docs"
-          dependency-versions: "highest"
-
-      - name: "Add orphan metadata where needed"
-        run: |
-          printf '%s\n\n%s\n' ":orphan:" "$(cat docs/en/sidebar.rst)" > docs/en/sidebar.rst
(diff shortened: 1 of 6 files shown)


GromNaN (approved):

composer.json
```diff
@@ -20,6 +20,9 @@
         { "name": "Ion Bazan", "email": "ion.bazan@gmail.com" },
         { "name": "Fran Moreno", "email": "franmomu@gmail.com" }
     ],
+    "scripts": {
+        "docs": "./docs/vendor/bin/build-docs.sh @additional_args"
```
greg0ire: This misses the `composer update`

.github/workflows/documentation.yml
```diff
@@ -15,30 +15,32 @@ on:
...
 
-      - name: "Add orphan metadata where needed"
+      - name: "Run docs builder"
         run: |
-          printf '%s\n\n%s\n' ":orphan:" "$(cat docs/en/sidebar.rst)" > docs/en/sidebar.rst
-          printf '%s\n\n%s\n' ":orphan:" "$(cat docs/en/reference/annotations-reference.rst)" > docs/en/reference/annotations-reference.rst
-
-      - name: "Run guides-cli"
-        run: "docs/vendor/bin/guides -vvv --no-progress docs/en 2>&1 | grep -v 'No template found for rendering directive' | ( ! grep WARNING )"
+          ./docs/vendor/bin/build-docs.sh 2>&1 | grep -v 'No template found for rendering directive' | ( ! grep WARNING )
```
greg0ire: Why not reuse the global workflow here?

paulinevos: Because those warnings need to be suppressed

GromNaN: This is probably something we need to fix in our rst files? ORM docs [uses `toctree`](https://github.com/doctrine/orm/blob/7602a5341cd5ede7b1bb2a56c24a4b2f59644e2f/docs/en/sidebar.rst#L3)

paulinevos: Ok, I can see how that's in the scope of this PR.

paulinevos: Done

GromNaN (approved):
Thank you @paulinevos

PR #2854, opened 2025-10-13
Add field type for Binary vectors

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | [PHPORM-384](https://jira.mongodb.org/browse/PHPORM-384)

#### Summary

Vector can be compressed into a binary field thanks to the new `MongoDB\BSON\Binary::TYPE_VECTOR` introduced in MongoDB PHP Extension v2.2.0

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -50,7 +50,7 @@
         "phpstan/phpstan": "^2.1",
         "phpstan/phpstan-deprecation-rules": "^2.0",
         "phpstan/phpstan-phpunit": "^2.0",
-        "phpunit/phpunit": "^10.4",
+        "phpunit/phpunit": "^10.5.58",
         "squizlabs/php_codesniffer": "^4",
         "symfony/cache": "^5.4 || ^6.0 || ^7.0",
         "symfony/uid": "^5.4 || ^6.0 || ^7.0"
--- a/lib/Doctrine/ODM/MongoDB/Types/AbstractVectorType.php
+++ b/lib/Doctrine/ODM/MongoDB/Types/AbstractVectorType.php
@@ -1,0 +1,134 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ODM\MongoDB\Types;
+
+use InvalidArgumentException;
+use MongoDB\BSON\Binary;
+use MongoDB\BSON\VectorType;
+
+use function enum_exists;
+use function get_debug_type;
+use function is_array;
+use function sprintf;
+use function str_replace;
+
+/** @internal */
+abstract class AbstractVectorType extends Type
+{
+    public function convertToDatabaseValue(mixed $value): ?Binary
+    {
+        if (! enum_exists(VectorType::class)) {
+            throw new InvalidArgumentException('MongoDB\BSON\VectorType enum does not exist. Install the MongoDB Extension version 2.2.0 or higher in order to use a vector field type.');
+        }
+
+        if ($value === null) {
+            return null;
+        }
+
+        if (is_array($value)) {
+            return Binary::fromVector($value, $this->getVectorType());
+        }
+
+        if (! $value instanceof Binary) {
+            throw new InvalidArgumentException(sprintf('Invalid data type %s received for vector field, expected null, array or MongoDB\BSON\Binary', get_debug_type($value)));
+        }
+
+        if ($value->getType() !== Binary::TYPE_VECTOR) {
+            throw new InvalidArgumentException(sprintf('Invalid binary data of type %d received for vector field, expected binary type %d', $value->getType(), Binary::TYPE_VECTOR));
+        }
+
+        if ($value->getVectorType() !== $this->getVectorType()) {
+            throw new InvalidArgumentException(sprintf('Invalid binary vector data of vector type %s received for vector field, expected vector type %s', $value->getVectorType()->name, $this->getVectorType()->name));
+        }
+
+        return $value;
+    }
+
+    /** @return list<float>|list<int>|list<bool>|null */
+    public function convertToPHPValue(mixed $value): ?array
+    {
+        if ($value === null) {
+            return null;
+        }
+
+        if (is_array($value)) {
+            return $value;
+        }
+
+        if (! $value instanceof Binary) {
+            throw new InvalidArgumentException(sprintf('Invalid data of type "%s" received for vector field', get_debug_type($value)));
+        }
+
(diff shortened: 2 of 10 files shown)


lib/Doctrine/ODM/MongoDB/Types/AbstractVectorType.php
```diff
@@ -0,0 +1,134 @@
...
+        }
+
+        if ($value->getVectorType() !== $this->getVectorType()) {
+            throw new InvalidArgumentException(sprintf('Invalid binary vector data of vector type %s received for vector field, expected vector type %s', $value->getVectorType()->name, $this->getVectorType()->name));
+        }
+
+        return $value->toArray();
+    }
+
+    public function closureToMongo(): string
```
paulinevos, line 75: What are these methods for? Might be helpful to document. At least to me it's unclear what they are

GromNaN: Ok, addressed your request in https://github.com/doctrine/mongodb-odm/pull/2855

lib/Doctrine/ODM/MongoDB/Types/AbstractVectorType.php
```diff
@@ -0,0 +1,134 @@
...
+
+        if ($value->getType() !== Binary::TYPE_VECTOR) {
+            throw new InvalidArgumentException(sprintf('Invalid binary data of type %d received for vector field', $value->getType()));
+        }
+
+        if ($value->getVectorType() !== $this->getVectorType()) {
+            throw new InvalidArgumentException(sprintf('Invalid binary vector data of vector type %s received for vector field, expected vector type %s', $value->getVectorType()->name, $this->getVectorType()->name));
+        }
+
+        return $value->toArray();
```
GromNaN, line 72: In the design, the property type is considered to be an `array`, so an `array` is always returned. However, it would be much more useful to know the type of the property in order to return a `Binary` if the property allows it, thus avoiding the need to decode the vector when it's not actually used.

lib/Doctrine/ODM/MongoDB/Types/AbstractVectorType.php
jmikola: For the record, in all my years of working with ODM (on and off) I don't think I've ever fully understood why we needed all of these methods. The answer probably rests in what PHP 5.3 was capable of back in the day.

lib/Doctrine/ODM/MongoDB/Types/AbstractVectorType.php
jmikola: Is this something not possible with the current design of the type, since we don't have any context for the class property to which we are associated?

tests/Doctrine/ODM/MongoDB/Tests/Types/VectorTypeTest.php
```diff
@@ -0,0 +1,153 @@
...
+use Doctrine\ODM\MongoDB\Types\Type;
+use InvalidArgumentException;
+use MongoDB\BSON\Binary;
+use MongoDB\BSON\VectorType;
+use PHPUnit\Framework\Attributes\DataProvider;
+use PHPUnit\Framework\Attributes\RequiresMethod;
+use PHPUnit\Framework\TestCase;
+
+/** Binary Vector type added in ext-mongodb 2.2+ */
+#[RequiresMethod(Binary::class, 'fromVector')]
```
jmikola: TIL https://docs.phpunit.de/en/11.5/writing-tests-for-phpunit.html#skipping-tests-using-attributes

GromNaN: Very handy. I don't get why Sebastian rejected `RequiresClass`: https://github.com/sebastianbergmann/phpunit/pull/4792

GromNaN: I could have used `RequiresPhpExtension('mongodb', '>= 2.2')`

jmikola: Could folks just use `RequiresMethod(MyClass::class, '__construct()')` since all classes have a constructor?

Edit: No they cannot. Not all classes have a constructor (even implicit): https://3v4l.org/Y5OdE#v8.4.13

lib/Doctrine/ODM/MongoDB/Types/AbstractVectorType.php
```diff
@@ -0,0 +1,134 @@
...
+
+            if ($value->getVectorType() !== \MongoDB\BSON\VectorType::%%vectorType%%) {
+                throw new \InvalidArgumentException(sprintf('Invalid binary vector data of vector type %s received for vector field, expected vector type %%vectorType%%', $value->getVectorType()->name));
+            }
+
+            $return = $value->toArray();
+PHP);
+    }
+
+    abstract protected function getVectorType(): VectorType;
```
jmikola, line 133: Is the PHPStan error here just due to the class not being available?

jmikola (approved):
Do you want to block this on an ext-mongodb 2.2.0 release? I'll defer to you.

I assume no changes to `composer.json` or the CI workflow are required.

lib/Doctrine/ODM/MongoDB/Types/AbstractVectorType.php
GromNaN: Exactly.

lib/Doctrine/ODM/MongoDB/Types/AbstractVectorType.php
GromNaN: I've skipped in PHPStan all errors related to binary vectors.

composer.json
```diff
@@ -50,7 +50,7 @@
         "phpstan/phpstan": "^2.1",
         "phpstan/phpstan-deprecation-rules": "^2.0",
         "phpstan/phpstan-phpunit": "^2.0",
-        "phpunit/phpunit": "^10.4",
+        "phpunit/phpunit": "^10.5.58",
```
GromNaN, line 53: Updating the minimum version of PHPUnit as the version 10.4.0 load the data providers even if `#[RequiresPhpExtension('mongodb', '>= 2.2')]` is not matched.

PR #2855, opened 2025-10-14
Document methods `Type::closureToMongo()` and `closureToPHP()`

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | https://github.com/doctrine/mongodb-odm/pull/2854#discussion_r2428622880

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Types/Type.php
+++ b/lib/Doctrine/ODM/MongoDB/Types/Type.php
@@ -122,11 +122,19 @@ abstract class Type
         return $value;
     }
 
+    /**
+     * Get the PHP code equivalent to {@see convertToDatabaseValue()}, used in code generator.
+     * Use variables $value for input and $return for output.
+     */
     public function closureToMongo(): string
     {
         return '$return = $value;';
     }
 
+    /**
+     * Get the PHP code equivalent to {@see convertToPHPValue()}, used in code generator.
+     * Use variables $value for input and $return for output.
+     */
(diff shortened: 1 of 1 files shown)


lib/Doctrine/ODM/MongoDB/Types/Type.php
```diff
@@ -122,11 +122,19 @@ public function convertToPHPValue($value)
         return $value;
     }
 
+    /**
+     * Get the PHP code equivalent to {@see convertToDatabaseValue()}, used in code generator.
+     * Use variables $value for input and $return for output.
+     */
     public function closureToMongo(): string
```
GromNaN, line 129: This method is not actually used, it could be deprecated but we could decide to use it later in the persistence layer.

lib/Doctrine/ODM/MongoDB/Types/Type.php
```diff
@@ -122,11 +122,19 @@ public function convertToPHPValue($value)
...
     {
         return '$return = $value;';
     }
 
+    /**
+     * Get the PHP code equivalent to {@see convertToPHPValue()}, used in code generator.
+     * Use variables $value for input and $return for output.
+     */
     public function closureToPHP(): string
```
GromNaN, line 138: This method is used in `HydratorFactory`. I don't specify it here, as that can change. The IDE/SA can easily find usages of this methods.

paulinevos (approved):

PR #2857, opened 2025-10-15
Validate `filter` and `queryVector` parameter in `$vectorSearch` stage builder

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | -

#### Summary

- Give early feedback when an invalid value is set as `queryVector`.
- Allow using an array for the [filter](https://www.mongodb.com/docs/atlas/atlas-vector-search/vector-search-stage/#fields) option.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/VectorSearch.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/VectorSearch.php
@@ -8,9 +8,14 @@ use Doctrine\ODM\MongoDB\Aggregation\Builder;
 use Doctrine\ODM\MongoDB\Aggregation\Stage;
 use Doctrine\ODM\MongoDB\Persisters\DocumentPersister;
 use Doctrine\ODM\MongoDB\Query\Expr;
+use InvalidArgumentException;
 use MongoDB\BSON\Binary;
 use MongoDB\BSON\Decimal128;
 use MongoDB\BSON\Int64;
+
+use function array_is_list;
+use function is_array;
+use function sprintf;
 
 /**
  * @phpstan-type Vector list<int|Int64>|list<float|Decimal128>|list<bool|0|1>|Binary
@@ -28,12 +33,15 @@ use MongoDB\BSON\Int64;
  */
 class VectorSearch extends Stage
 {
-    private ?bool $exact        = null;
-    private ?Expr $filter       = null;
-    private ?string $index      = null;
-    private ?int $limit         = null;
-    private ?int $numCandidates = null;
-    private ?string $path       = null;
+    /** @see Binary::TYPE_VECTOR introduced in ext-mongodb 2.2 */
+    private const BINARY_TYPE_VECTOR = 9;
+
+    private ?bool $exact            = null;
+    private array|Expr|null $filter = null;
+    private ?string $index          = null;
+    private ?int $limit             = null;
+    private ?int $numCandidates     = null;
+    private ?string $path           = null;
     /** @phpstan-var Vector|null */
     private array|Binary|null $queryVector = null;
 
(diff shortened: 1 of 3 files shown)


tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/VectorSearchTest.php
```diff
@@ -97,6 +106,17 @@ public function testQueryVectorAcceptsBinary(): void
         self::assertSame(['$vectorSearch' => ['queryVector' => $binaryVector]], $stage->getExpression());
     }
 
+    #[TestWith([new Binary("\x03\x00\x01\x02\x03", Binary::TYPE_GENERIC), 'Binary query vector must be of type 9 (Vector), got 0.'])]
+    #[TestWith([[1 => 1, 2 => 3], 'Query vector must be a list of numbers, got an associative array.'])]
+    #[TestWith([[], 'Query vector cannot be an empty array.'])]
```
paulinevos, line 111: Oh cool, I had not seen these attributes before.

paulinevos (approved):

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/VectorSearch.php
```diff
@@ -28,12 +33,15 @@
...
 class VectorSearch extends Stage
 {
-    private ?bool $exact        = null;
-    private ?Expr $filter       = null;
-    private ?string $index      = null;
-    private ?int $limit         = null;
-    private ?int $numCandidates = null;
-    private ?string $path       = null;
+    /** @see Binary::TYPE_VECTOR introduced in ext-mongodb 2.2 */
+    private const BINARY_TYPE_VECTOR = 9;
```
jmikola, line 37: Just to confirm, will this stick around even after the ext-mongodb 2.2.0 release? I assume you don't want to raise the driver dependency just for this feature.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/VectorSearch.php
```diff
@@ -122,6 +133,18 @@ public function path(string $path): static
     /** @phpstan-param Vector $queryVector */
     public function queryVector(array|Binary $queryVector): static
     {
+        if ($queryVector === []) {
+            throw new InvalidArgumentException('Query vector cannot be an empty array.');
+        }
+
+        if (is_array($queryVector) && ! array_is_list($queryVector)) {
+            throw new InvalidArgumentException('Query vector must be a list of numbers, got an associative array.');
```
jmikola, line 141: Do you actually intend to validate the contents of the list, or will you rely on the server to report that error?

jmikola (approved):
A couple of questions which might lead to some small changes, but defer to you.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/VectorSearch.php
GromNaN: Exactly, we don't want to update the required version for this feature. But someone could receive a binary vector from the server and use in a query. That's why we have this constant duplicated here.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/VectorSearch.php
GromNaN: Yes, this is a partial validation. I don't know what vector type is expected so iterating over the values would just cost time.

tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/VectorSearchTest.php
GromNaN: Before attributes, it was the [`@testWith` annotation](https://docs.phpunit.de/en/10.5/annotations.html#testwith) but the syntax was awful.

PR #2859, opened 2025-10-17
Randomize the collection name to avoid collisions when search indexes are created or dropped asynchronously

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | 

#### Summary

Search indexes are created and dropped asynchronously. Sometimes the search index from a previous test still exists when the new test is run, even if the collection has been dropped.

By setting a random collection name, we prevent side effects.

I think the issue arises from the new version of MongoDB Atlas Local.


jmikola (approved):
The WC addition seems unrelated to randomized collection names, but I do remember you adding a WC to data fixtures in a previous test. Was this just overlooked?

GromNaN:
Yes, that was missing. The test was randomly failing with the new version of MongoDB Atlas Local without the WC.

PR #2866, opened 2025-10-22
Move source files into `src/` directory

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | task
| BC Break     | no
| Fixed issues | -

#### Summary

Apply common convention of location source files into `src/`, possible thanks to PSR-4, while the deep directory structure was required by PSR-1.


greg0ire:
I've done this change on other Doctrine libs. Here is how I tried to proceed:

- write a script that does the migration (sadly I didn't save the script).
- do a merge up so that all branches are up to date before starting any work
- do the change on 2.13.x so as to minimise differences between branches, or do the change right before releasing 2.14.x
- do the change on all maintained branches with the script
- use `git merge -s ours` to merge up the change from lower branches

GromNaN:
2.13.x is the only maintained branch. I plan to release 2.14.0 with native lazy objects support as soon as this is merged. 2.14.x will be the only maintained branch.

greg0ire:
Okay, but what about 3.0.x? It's maintained, isn't it?

GromNaN:
3.0.x is not released yet. The upmerge will be easy: https://github.com/doctrine/mongodb-odm/compare/2.14.x...3.0.x

greg0ire (approved):

alcaeus (approved):
Glad we're finally able to do this. Thank you!

PR #2867, opened 2025-10-23
Don't set env vars in PHPUnit config

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | -

#### Summary

Variables defined in `phpunit.xml` override the real env vars. 
This fixes the CI job to actually use "proxy-manager".


GromNaN:
It seems that I'm wrong. https://docs.phpunit.de/en/10.5/configuration.html#the-env-element 
> By default, environment variables are not overwritten if they exist already.

GromNaN:
The issue is in 2.14.x, trailing `"`. https://github.com/doctrine/mongodb-odm/blob/6f5fc0d7101abc38e93c531d5e35ceff39c17859/.github/workflows/continuous-integration.yml#L172-L173

PR #2868, opened 2025-10-23
Fix binary UUID hydrator

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | 

#### Summary

Hydrator closures may not return, as that would terminate hydration for all fields.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Types/BinaryUuidType.php
+++ b/lib/Doctrine/ODM/MongoDB/Types/BinaryUuidType.php
@@ -60,18 +60,13 @@ PHP;
         return <<<'PHP'
         if ($value instanceof \Symfony\Component\Uid\Uuid) {
             $return = $value;
-            return;
-        }
-
-        if (! $value instanceof \MongoDB\BSON\Binary) {
+        } elseif (! $value instanceof \MongoDB\BSON\Binary) {
             throw new \InvalidArgumentException(sprintf('Invalid data of type "%s" received for Uuid', get_debug_type($value)));
-        }
-
-        if ($value->getType() !== \MongoDB\BSON\Binary::TYPE_UUID) {
+        } elseif ($value->getType() !== \MongoDB\BSON\Binary::TYPE_UUID) {
             throw new \InvalidArgumentException(sprintf('Invalid binary data of type %d received for Uuid', $value->getType()));
+        } else {
+            $return = \Symfony\Component\Uid\Uuid::fromBinary($value->getData());
         }
-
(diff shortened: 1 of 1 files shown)


lib/Doctrine/ODM/MongoDB/Types/BinaryUuidType.php
```diff
@@ -60,18 +60,13 @@ public function closureToPHP(): string
         return <<<'PHP'
         if ($value instanceof \Symfony\Component\Uid\Uuid) {
             $return = $value;
-            return;
-        }
-
-        if (! $value instanceof \MongoDB\BSON\Binary) {
+        } else if (! $value instanceof \MongoDB\BSON\Binary) {
```
GromNaN:
```suggestion
        } elseif (! $value instanceof \MongoDB\BSON\Binary) {
```

lib/Doctrine/ODM/MongoDB/Types/BinaryUuidType.php
```diff
@@ -60,18 +60,13 @@ public function closureToPHP(): string
...
             throw new \InvalidArgumentException(sprintf('Invalid data of type "%s" received for Uuid', get_debug_type($value)));
-        }
-
-        if ($value->getType() !== \MongoDB\BSON\Binary::TYPE_UUID) {
+        } else if ($value->getType() !== \MongoDB\BSON\Binary::TYPE_UUID) {
```
GromNaN:
```suggestion
        } elseif ($value->getType() !== \MongoDB\BSON\Binary::TYPE_UUID) {
```

GromNaN (changes requested):
Could you add tests, please?

alcaeus:
> Could you add tests, please?

There are tests for a documents with UUIDs that test the hydration logic of the type. I'm not sure why that error didn't pop up, but after seeing it in my test application it made sense. The code in the `closureTo` methods is written to the hydrator classes and run in a loop, so the `return` is very much an invalid operation. Not sure how to best test for this TBH :(

GromNaN:
The new vector types have the same issue: https://github.com/doctrine/mongodb-odm/blob/bbf953119ae5fa7769c1db1a2335ceddd43d85b7/src/Types/AbstractVectorType.php#L75-L102

We should remove this method? https://github.com/doctrine/mongodb-odm/issues/2870

GromNaN (approved):

GromNaN:
Thank you Andreas.

PR #2871, opened 2025-10-24
Fix binary vector types and conversion in `$vectorSearch` stage

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | -

#### Summary

- Fix the vector types for generated code. Calling `return` is not permitted.
- In the `$vectorSearch` stage, convert the vector to Binary using the field type.

### DIFF
--- a/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/VectorSearch.php
+++ b/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/VectorSearch.php
@@ -8,6 +8,7 @@ use Doctrine\ODM\MongoDB\Aggregation\Builder;
 use Doctrine\ODM\MongoDB\Aggregation\Stage;
 use Doctrine\ODM\MongoDB\Persisters\DocumentPersister;
 use Doctrine\ODM\MongoDB\Query\Expr;
+use Doctrine\ODM\MongoDB\Types\Type;
 use InvalidArgumentException;
 use MongoDB\BSON\Binary;
 use MongoDB\BSON\Decimal128;
@@ -81,7 +82,7 @@ class VectorSearch extends Stage
         }
 
         if ($this->queryVector !== null) {
-            $params['queryVector'] = $this->queryVector;
+            $params['queryVector'] = Type::getType($this->persister->getClassMetadata()->fieldMappings[$this->path]['type'] ?? Type::RAW)->convertToDatabaseValue($this->queryVector);
         }
 
         return [$this->getStageName() => $params];
--- a/lib/Doctrine/ODM/MongoDB/Types/AbstractVectorType.php
+++ b/lib/Doctrine/ODM/MongoDB/Types/AbstractVectorType.php
@@ -77,27 +77,17 @@ abstract class AbstractVectorType extends Type
         return str_replace('%%vectorType%%', $this->getVectorType()->name, <<<'PHP'
             if ($value === null) {
                 $return = null;
-                return;
-            }
-
-            if (\is_array($value)) {
+            } elseif (\is_array($value)) {
                 $return = \MongoDB\BSON\Binary::fromVector($value, \MongoDB\BSON\VectorType::%%vectorType%%);
-                return;
-            }
-
-            if (! $value instanceof \MongoDB\BSON\Binary) {
+            } elseif (! $value instanceof \MongoDB\BSON\Binary) {
                 throw new InvalidArgumentException(sprintf('Invalid data type %s received for vector field, expected null, array or MongoDB\BSON\Binary', get_debug_type($value)));
-            }
-
-            if ($value->getType() !== \MongoDB\BSON\Binary::TYPE_VECTOR) {
+            } elseif ($value->getType() !== \MongoDB\BSON\Binary::TYPE_VECTOR) {
                 throw new InvalidArgumentException(sprintf('Invalid binary data of type %d received for vector field, expected binary type %d', $value->getType(), \MongoDB\BSON\Binary::TYPE_VECTOR));
-            }
-
-            if ($value->getVectorType() !== \MongoDB\BSON\VectorType::%%vectorType%%) {
+            } elseif ($value->getVectorType() !== \MongoDB\BSON\VectorType::%%vectorType%%) {
                 throw new \InvalidArgumentException(sprintf('Invalid binary vector data of vector type %s received for vector field, expected vector type %%vectorType%%', $value->getVectorType()->name));
+            } else {
+                $return = $value;
             }
-
-            $return = $value;
 PHP);
     }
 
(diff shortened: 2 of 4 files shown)


tests/Doctrine/ODM/MongoDB/Tests/Functional/VectorSearchTest.php
GromNaN, line 90: Without automatic conversion of the vector to binary, this pipeline doesn't return anything.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/VectorSearch.php
```diff
@@ -82,6 +83,12 @@ public function getExpression(): array
 
         if ($this->queryVector !== null) {
             $params['queryVector'] = $this->queryVector;
+
+            // Apply type conversion for Binary vectors
+            $type = $this->persister->getClassMetadata()->fieldMappings[$this->path]['type'] ?? null;
+            if ($type) {
+                $params['queryVector'] = Type::getType($type)->convertToDatabaseValue($params['queryVector']);
+            }
```
jmikola: If `$type` isn't used elsewhere, would this be a candidate for the [null-safe call operator](https://php.watch/versions/8.0/null-safe-operator)? 😃 

Might be too much for a one-liner, though. No objections to keeping this as-is.

lib/Doctrine/ODM/MongoDB/Types/AbstractVectorType.php
```diff
@@ -77,27 +77,17 @@ public function closureToMongo(): string
...
-            if ($value->getType() !== \MongoDB\BSON\Binary::TYPE_VECTOR) {
+            } elseif ($value->getType() !== \MongoDB\BSON\Binary::TYPE_VECTOR) {
                 throw new InvalidArgumentException(sprintf('Invalid binary data of type %d received for vector field, expected binary type %d', $value->getType(), \MongoDB\BSON\Binary::TYPE_VECTOR));
-            }
-
-            if ($value->getVectorType() !== \MongoDB\BSON\VectorType::%%vectorType%%) {
+            } elseif ($value->getVectorType() !== \MongoDB\BSON\VectorType::%%vectorType%%) {
                 throw new \InvalidArgumentException(sprintf('Invalid binary vector data of vector type %s received for vector field, expected vector type %%vectorType%%', $value->getVectorType()->name));
+            } else {
+                $return = $value;
```
jmikola, line 89: This seems functionally equivalent to the previous sequence of `if` conditions. Am I missing something?

tests/Doctrine/ODM/MongoDB/Tests/Functional/VectorSearchTest.php
jmikola: Was the lack of a `hydrate()` call the root cause of this issue?

tests/Doctrine/ODM/MongoDB/Tests/Types/VectorTypeTest.php
```diff
@@ -149,4 +151,10 @@ public static function providePHPValueException(): iterable
         yield [new Binary("\x03\x00\x01\x02\x03", Binary::TYPE_GENERIC), 'Invalid binary data of type 0 received for vector field'];
         yield [Binary::fromVector([1, 2, 3], VectorType::Int8), 'Invalid binary vector data of vector type Int8 received for vector field, expected vector type Float32'];
     }
+
+    private function assertSameTypeAndValue(mixed $expected, mixed $actual): void
```
jmikola, line 155: Is this something that would ideally be done with a PHPUnit Comparator? I realize those are cumbersome to write, so no objections here.

tests/Doctrine/ODM/MongoDB/Tests/Functional/VectorSearchTest.php
```diff
@@ -36,18 +39,39 @@ public function testAtlasVectorSearch(): void
         $doc3->vectorInt   = [2, 3, 4];
         $doc3->filterField = 'active';
 
+        $doc4              = new VectorEmbedding();
+        $doc4->filterField = 'inactive';
```
alcaeus: I suppose this new document triggers the failing behaviour in the type?

lib/Doctrine/ODM/MongoDB/Types/ClosureToPHP.php
```diff
@@ -12,7 +12,7 @@ trait ClosureToPHP
     final public function closureToPHP(): string
     {
         return sprintf('
-            $type = \%s::getType($typeIdentifier);
-            $return = $type->convertToPHPValue($value);', Type::class);
+                $type = \%s::getType($typeIdentifier);
+                $return = $type->convertToPHPValue($value);', Type::class);
```
alcaeus: This change seems unrelated, why the change?

alcaeus (approved):
Changes LGTM with a minor question

lib/Doctrine/ODM/MongoDB/Types/ClosureToPHP.php
GromNaN: This indendation is more correct once this code is dumped in the hydrator class. I can revert. We will rework code generation in a dedicated ticket.

alcaeus: Fine to leave as you have it now. I was just wondering how that came to be.

tests/Doctrine/ODM/MongoDB/Tests/Functional/VectorSearchTest.php
GromNaN: Actually, no, types are never called when the value is `null`. So I revert this change.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/VectorSearch.php
GromNaN: I wrote the one-line as an exercice. But I will not keep it.

```php
$params['queryVector'] = Type::getType($this->persister->getClassMetadata()->fieldMappings[$this->path]['type'] ?? Type::RAW)->convertToDatabaseValue($this->queryVector);
```

lib/Doctrine/ODM/MongoDB/Types/AbstractVectorType.php
GromNaN: We can't use `return` inside the generated code, as it's not in its own function.

lib/Doctrine/ODM/MongoDB/Aggregation/Stage/VectorSearch.php
GromNaN: I've decided to take the one-line.

tests/Doctrine/ODM/MongoDB/Tests/Types/VectorTypeTest.php
GromNaN: Ideally, it should be added to PHPUnit.

tests/Doctrine/ODM/MongoDB/Tests/Functional/VectorSearchTest.php
GromNaN: I wonder if it's something to do with the automatic vector conversion, which would use float by default.

PR #2881, opened 2025-10-29
Support update pipelines in query builder

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | 

#### Summary

This PR adds a new `pipeline()` method to the query builder which can be used to run aggregation pipelines for `updateOne`, `updateMany`, and `findAndUpdate` operations. The method accepts either an aggregation builder instance, a `Pipeline` instance from the MongoDB library, or an array of pipeline stages.

### DIFF
--- a/tests/Tests/Query/PipelineUpdateTest.php
+++ b/tests/Tests/Query/PipelineUpdateTest.php
@@ -1,0 +1,227 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ODM\MongoDB\Tests\Query;
+
+use BadMethodCallException;
+use Doctrine\ODM\MongoDB\Tests\BaseTestCase;
+use Documents\User;
+use MongoDB\Builder\Expression;
+use MongoDB\Builder\Pipeline;
+use MongoDB\Builder\Stage;
+
+class PipelineUpdateTest extends BaseTestCase
+{
+    private User $user1;
+    private User $user2;
+
+    public function setUp(): void
+    {
+        parent::setUp();
+
+        $this->user1 = new User();
+        $this->user1->setUsername('foo');
+        $this->user1->setHits(1);
+
+        $this->user2 = new User();
+        $this->user2->setUsername('bar');
+        $this->user2->setHits(2);
+
+        $this->dm->persist($this->user1);
+        $this->dm->persist($this->user2);
+
+        $this->dm->flush();
+    }
+
+    public function testUpdateManyWithWrongType(): void
+    {
+        $this->expectException(BadMethodCallException::class);
+        $this->expectExceptionMessage('The pipeline() method can only be used with update or findAndUpdate queries.');
+
+        $this->dm->createQueryBuilder(User::class)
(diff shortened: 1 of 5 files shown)


tests/Tests/Query/PipelineUpdateTest.php
```diff
@@ -0,0 +1,230 @@
...
+        $this->user2->setUsername('bar');
+        $this->user2->setHits(2);
+
+        $this->dm->persist($this->user1);
+        $this->dm->persist($this->user2);
+
+        $this->dm->flush();
+    }
+
+    public function testUpdateManyWithWrongType(): void
```
alcaeus, line 37: Note: I didn't deem it necessary to test with all possible invalid query types, and instead only test the default query type (find).

tests/Tests/Query/PipelineUpdateTest.php
```diff
@@ -0,0 +1,230 @@
...
+            ->getQuery()
+            ->execute();
+
+        self::assertInstanceOf(User::class, $user);
+        self::assertSame(2, $user->getHits());
+    }
+
+    public function testFindOneAndUpdateWithDriverPipeline(): void
+    {
+        $this->markTestIncomplete('Collection::findAndUpdate does not support pipeline updates (PHPLIB-1699)');
```
GromNaN: Incomplete test make the test suite fail
```suggestion
        $this->markTestSkipped('Collection::findAndUpdate does not support pipeline updates (PHPLIB-1699)');
```

alcaeus: Huh, TIL that incomplete tests fail the build 🤷

GromNaN: We ask for it: https://github.com/doctrine/mongodb-odm/blob/257a0a2ab0ba140b36b7be7e3f3eecb39a7c5059/phpunit.xml.dist#L13

GromNaN (approved):

tests/Tests/Query/PipelineUpdateTest.php
```diff
@@ -0,0 +1,230 @@
...
+            ->getQuery()
+            ->execute();
+
+        self::assertInstanceOf(User::class, $user);
+        self::assertSame(2, $user->getHits());
+    }
+
+    public function testFindOneAndUpdateWithDriverPipeline(): void
+    {
+        $this->markTestSkipped('Collection::findAndUpdate does not support pipeline updates (PHPLIB-1699)');
```
GromNaN, line 197: Thinking of that, you can still convert the pipeline to an array using `iterator_to_array($pipeline)`.

alcaeus: How so? The issue is that `findAndModify` doesn't run the `update` parameter through the pipeline encoder, so it won't ever convert the stages in the resulting array either. Not to worry, I just figured I'll add the test so that it can easily be enabled once we implement this in the driver.

PR #2883, opened 2025-10-30
Deprecate `ClassMetadataFactoryInterface::setProxyClassNameResolver()`

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | -

#### Summary

Native lazy objects remove the need for a `ClassNameResolver`.
Also deprecating in https://github.com/doctrine/persistence/pull/460

### DIFF
--- a/src/DocumentManager.php
+++ b/src/DocumentManager.php
@@ -176,7 +176,9 @@ class DocumentManager implements ObjectManager
         $this->metadataFactory    = new $metadataFactoryClassName();
         $this->metadataFactory->setDocumentManager($this);
         $this->metadataFactory->setConfiguration($this->config);
-        $this->metadataFactory->setProxyClassNameResolver($this->classNameResolver);
+        if (! $this->config->isNativeLazyObjectEnabled()) {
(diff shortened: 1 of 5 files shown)


alcaeus (approved):

SenseException (approved):

src/DocumentManager.php
```diff
@@ -300,6 +316,8 @@ public function getSchemaManager(): SchemaManager
      */
     public function getClassNameResolver(): ClassNameResolver
     {
+        trigger_deprecation('doctrine/mongodb-odm', '2.15', 'The %s() method is deprecated and will be removed in Doctrine ODM 3.0. There is no proxy classes when using native lazy objects', __METHOD__);
```
GromNaN: This should be deprecated only when PHP 8.4+ is used, WDYT?

greg0ire:
```suggestion
        trigger_deprecation('doctrine/mongodb-odm', '2.15', 'The %s() method is deprecated and will be removed in Doctrine ODM 3.0. There are no proxy classes when using native lazy objects', __METHOD__);
```

PR #2884, opened 2025-10-31
Ensure proxy-manager is not used when native lazy objects are enabled

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | -

#### Summary

- Fix initialization of proxies when hydrating collection with ProxyManager (bug fix).
- Fix unsetting native lazy object config to not enable symfony lazy ghost objects (bug fix)
- Add a CI job without optional dependencies to ensure they are not required.
- Create test with read-only properties with lazy-objects. Not supported by ProxyManager https://github.com/FriendsOfPHP/proxy-manager-lts/issues/26

### DIFF
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -708,7 +708,8 @@ class Configuration
 
     public function isLazyGhostObjectEnabled(): bool
     {
-        return $this->lazyGhostObject;
+        // Always false if native lazy objects are enabled
+        return $this->lazyGhostObject && ! $this->nativeLazyObject;
     }
 
     public function setUseNativeLazyObject(bool $nativeLazyObject): void
@@ -718,7 +719,6 @@ class Configuration
         }
 
         $this->nativeLazyObject = $nativeLazyObject;
-        $this->lazyGhostObject  = ! $nativeLazyObject || $this->lazyGhostObject;
     }
 
     public function isNativeLazyObjectEnabled(): bool
--- a/src/Hydrator/HydratorFactory.php
+++ b/src/Hydrator/HydratorFactory.php
@@ -451,29 +451,18 @@ EOF
             }
         }
 
+        // Skip initialization to not load any object data
         if (PHP_VERSION_ID >= 80400) {
             $metadata->reflClass->markLazyObjectAsInitialized($document);
         }
 
         if ($document instanceof InternalProxy) {
-            // Skip initialization to not load any object data
             $document->__setInitialized(true);
         }
 
         // Support for legacy proxy-manager-lts
-        if ($document instanceof GhostObjectInterface && $document->getProxyInitializer() !== null) {
-            // Inject an empty initialiser to not load any object data
-            $document->setProxyInitializer(static function (
-                GhostObjectInterface $ghostObject,
-                string $method, // we don't care
-                array $parameters, // we don't care
-                &$initializer,
-                array $properties, // we currently do not use this
-            ): bool {
-                $initializer = null;
-
-                return true;
-            });
+        if ($document instanceof GhostObjectInterface) {
+            $document->setProxyInitializer(null);
         }
 
         $data = $this->getHydratorFor($metadata->name)->hydrate($document, $data, $hints);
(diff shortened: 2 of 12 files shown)


src/Proxy/Factory/StaticProxyFactory.php
```diff
@@ -146,7 +146,7 @@ private function skippedFieldsFqns(ClassMetadata $metadata): array
         $skippedFieldsFqns = [];
 
         foreach ($metadata->getIdentifierFieldNames() as $idField) {
-            $skippedFieldsFqns[] = $this->propertyFqcn($metadata->getReflectionProperty($idField));
```
GromNaN, line 149: Remove usage of the deprecated `ClassMetadata::getReflectionProperty()`

src/Hydrator/HydratorFactory.php
```diff
@@ -451,29 +451,18 @@ public function hydrate(object $document, array $data, array $hints = []): array
...
-                GhostObjectInterface $ghostObject,
-                string $method, // we don't care
-                array $parameters, // we don't care
-                &$initializer,
-                array $properties, // we currently do not use this
-            ): bool {
-                $initializer = null;
-
-                return true;
-            });
```
GromNaN, line 476: We need to set the initializer to `null`, so that the proxy is marked as initialized. Otherwise, updating a value that is already set doesn't initialize the proxy, and the UOW ignore the document when computing changeset.

src/Configuration.php
```diff
@@ -718,7 +719,6 @@ public function setUseNativeLazyObject(bool $nativeLazyObject): void
         }
 
         $this->nativeLazyObject = $nativeLazyObject;
-        $this->lazyGhostObject  = ! $nativeLazyObject || $this->lazyGhostObject;
```
GromNaN, line 721: This was setting `lazyGhostObject = true` when calling `setUseNativeLazyObject(false)`, which is not expected.

tests/Documents/Tag.php
```diff
@@ -14,7 +14,7 @@ class Tag
     public ?string $id;
 
     #[ODM\Field]
-    public readonly string $name;
```
GromNaN, line 17: Remove the `reaonly` property that was added in https://github.com/doctrine/mongodb-odm/pull/2840. That makes `Doctrine\ODM\MongoDB\Tests\Aggregation\BuilderTest::testAggregationBuilder` fail with proxy manager.

tests/Tests/Functional/ReadOnlyPropertiesTest.php
```diff
@@ -0,0 +1,51 @@
...
+use Doctrine\ODM\MongoDB\Mapping\Annotations\ReferenceOne;
+use Doctrine\ODM\MongoDB\Tests\BaseTestCase;
+
+class ReadOnlyPropertiesTest extends BaseTestCase
+{
+    public function testReadOnlyDocument(): void
+    {
+        $configuration = $this->dm->getConfiguration();
+        if (! $configuration->isNativeLazyObjectEnabled() && ! $configuration->isLazyGhostObjectEnabled()) {
+            $this->markTestSkipped('Read-only properties are not supported by the legacy Proxy Manager. https://github.com/FriendsOfPHP/proxy-manager-lts/issues/26');
```
GromNaN, line 19: https://github.com/FriendsOfPHP/proxy-manager-lts/issues/26

greg0ire:
Should 7cc80ec80bdec4661b2227c69982f107fe8b36d1 get fixed up into e67c780f55f188229c951fc06b93c4ca44f5c2b8?

tests/Tests/Mapping/LegacyReflectionFieldsTest.php
```diff
@@ -56,20 +60,32 @@ public function testGetSet(): void
...
 }
+
+#[Document]
+class ReadOnlyProperty
+{
+    #[Id]
+    public string $id;
+
+    public function __construct(#[Field]
+    public readonly string $name,)
```
greg0ire: Something weird happened here 🤨 

```suggestion
    public function __construct(
        #[Field]
        public readonly string $name
    )
```

GromNaN: The way this happened is a mystery.

greg0ire (approved):

PR #2886, opened 2025-10-31
Fix `useLazyGhostObject` option is `false` when `useNativeLazyObject` is `true`

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | 

#### Summary

We can't have both symfony lazy-ghost and native lazy objects enabled at the same time.
Also disable auto generate proxy classes.

### DIFF
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -337,6 +337,10 @@ class Configuration
      */
     public function getAutoGenerateProxyClasses(): int
     {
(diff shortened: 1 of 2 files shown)


GromNaN:
Closing as already included in https://github.com/doctrine/mongodb-odm/pull/2884

PR #2892, opened 2025-11-03
Remove UUID string generator strategy

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | yes
| Fixed issues | -

#### Summary

Use Symfony UUID instead.


src/Mapping/ClassMetadata.php
```diff
@@ -2168,14 +2161,6 @@ public function isIdGeneratorIncrement(): bool
         return $this->generatorType === self::GENERATOR_TYPE_INCREMENT;
     }
 
-    /**
-     * Checks whether the class will generate a uuid id.
-     */
-    public function isIdGeneratorUuid(): bool
-    {
-        return $this->generatorType === self::GENERATOR_TYPE_UUID;
-    }
```
GromNaN, line 2177: Adding a deprecation before removing this method: https://github.com/doctrine/mongodb-odm/pull/2896

docs/en/reference/basic-mapping.rst
malarzm, line 285: Note you've removed mentioned the `uuid` type, it'd be nice to have it here so people know what to use

malarzm (approved):

docs/en/reference/basic-mapping.rst
GromNaN: We have a note in previous version about this type being deprecated. I don't think it's appropriate to mention here a type that no longer exists in this version.

malarzm: Sorry I didn't make myself clear: I was suggesting to keep

> with a ``uuid`` type identifier field

somewhere here

GromNaN: The description of the AUTO strategy mentions the Symfony UUID type. Is that what you expected?

malarzm: I expected to let user know they need to combine `#[ODM\Id(strategy: 'AUTO')]` with `type: 'uuid'`, like the removed note did.

PR #2895, opened 2025-11-03
Require PHP 8.4

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | -
| BC Break     | no
| Fixed issues | -

#### Summary

Also remove tests on Symfony lazy-ghost and proxy manager.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -19,9 +19,6 @@ jobs:
       fail-fast: false
       matrix:
         php-version:
-          - "8.1"
-          - "8.2"
-          - "8.3"
           - "8.4"
         mongodb-version:
           - "8.0"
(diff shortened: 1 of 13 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -35,72 +32,43 @@ jobs:
           - "highest"
         symfony-version:
           - "stable"
-        proxy:
-          - "lazy-ghost"
         include:
```
SenseException, line 35: Do we still need the include section?

SenseException (approved):

.github/workflows/continuous-integration.yml
GromNaN: Yes for testing specific Symfony versions and server configurations.
We will add PHP 8.5 later. And we check the low deps job.

alcaeus (approved):

PR #2896, opened 2025-11-04
Deprecate `ClassMetadata::isIdGeneratorUuid()`

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | -

#### Summary

Since https://github.com/doctrine/mongodb-odm/pull/2826, the string UUID generator is deprecated, and the constant `ClassMetadata::GENERATOR_TYPE_UUID` is also deprecated.
The method that uses this method need to be deprecated too.


malarzm (dismissed):
You're missing `trigger_deprecation` in the method

jmikola (approved):

PR #2897, opened 2025-11-04
Remove Support for ext-mongodb 1.x

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | -

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -42,7 +42,7 @@ jobs:
           - dependencies: "lowest"
             php-version: "8.1"
             mongodb-version: "6.0"
-            driver-version: "1.21.0"
+            driver-version: "2.1.0"
             topology: "server"
             symfony-version: "stable"
             proxy: "lazy-ghost"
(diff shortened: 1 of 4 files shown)


SenseException (approved):

GromNaN:
Given the low amount of change, I wonder if it's worth removing support now. Upgrading PHP extension version is more complex that a library.
Note that we don't have a release date for ODM 3.0, so it might be in 6 month or more.

alcaeus:
> I wonder if it's worth removing support now

"Now" as in "merge this to 3.0.x now", or "now" as in "drop support in 2.x"? As much as I'd love to drop the old extension version, I think it's safer to do this in 3.0, even if that means we're dragging it along for a few more months.

GromNaN:
I was thinking of the opposite: not removing support for ext-mongodb 1.x in 3.0, but waiting until there was a [more significant shift in usage](https://packagist.org/packages/mongodb/mongodb-extension/stats). Composer allows us to modify the constraint in a minor 3.x version.

But looking at [pecl stats](https://pecl.php.net/package-stats.php?cid=7&pid=981), there is already a big shift
image

alcaeus:
> But looking at [pecl stats](https://pecl.php.net/package-stats.php?cid=7&pid=981), there is already a big shift

Yes. Since `pecl install mongodb` will install the latest published version, publishing the 2.x releases after 1.x patch releases ensures that people get the new version by default unless they specify the exact version they want to install.

Issue #2448, opened 2022-05-19
Switching "symfony/var-dumper" in require-dev


simonoche:
### Feature Request

The package symfony/var-dumper is currently required in the main "require" channel.
From my research, there is no valid reason for it to be loaded outside of a dev environment (for mongodb-odm).

I think it is widely accepted that putting this package in a production environment can be dangereous. 

The verbosity of the var-dumper can reveal very compromising information. For example, if a dump() or a dd() is inadvertently left in the code, and pushed to production.

Since this package requires the var-dumper in the main channel, then the var-dumper is always installed & loaded on all projects containing mongodb-odm.

Thank you for your attention, hope this can be addressed.

malarzm:
> From my research, there is no valid reason for it to be loaded outside of a dev environment (for mongodb-odm).

The reason for `symfony/var-dumper` in the `require` section is this command: https://github.com/doctrine/mongodb-odm/blob/2.4.x/lib/Doctrine/ODM/MongoDB/Tools/Console/Command/QueryCommand.php. I have never used it myself but I reckon it's useful since we're shipping it since the very beginning :)

Theoretically we could move the requirement to `require-dev`, put it in `suggest`, and throw an exception in the command if package is not present. But technically that's a BC break. Given dangling `var_dump` or `print_r` are as "dangerous" as Symfony's function I'm reluctant to do so right away.

> The verbosity of the var-dumper can reveal very compromising information. For example, if a dump() or a dd() is inadvertently left in the code, and pushed to production.

In the meantime while we're giving this a second thought, I'd advise to set simple pre-commit git hooks or even employ CI pipeline for such detections. In our project we're using https://github.com/phpro/grumphp with following task:

```
        git_blacklist:
            keywords:
                - "die("
                - "var_dump("
                - "print_r("
                - "dump("
                - " dd("
                - "exit;"
            triggered_by: ["php"]
```

simonoche:
You're right, _var_dump_ or _print_r_ may be "dangerous" as well.

However, Symfony's var-dumper package  is intended (I guess) for debugging (only?), and is quite popular those days. 

If you are not too careful, you easily end up with var-dumper loaded in production ("because of" doctrine-mongodb). Eventually, there may be unpleasant surprises then.

I totally understand your point of view, and the fact that this package is required by a Command. This was a suggestion, but I believe it's a relevant one.

Thanks for the tip, I'll take a look at grumphp

malarzm:
[...]

PR #2898, opened 2025-11-04
[3.x] Move symfony/var-dumper to dev dependencies

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | Fix https://github.com/doctrine/mongodb-odm/issues/2448

#### Summary

Enable the `odm:query` command only when the VarDumper component is installed.
Removed it from required dependencies.

Use `Command::isEnabled()` to conditionally enable the `odm:query` command when the dependency is available.


malarzm (approved):
Cool, didn't know about `Command::isEnabled` 👍

PR #2901, opened 2025-11-10
Fix documentation sidebar

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | -

#### Summary

The links need to be added explicitly, the `/*` notation does not work anymore.
I think the issue was introduced by https://github.com/doctrine/mongodb-odm/pull/2849

https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/latest/cookbook/queryable-encryption.html

### DIFF
--- a/docs/en/sidebar.rst
+++ b/docs/en/sidebar.rst
@@ -5,7 +5,7 @@
    :depth: 3
    :glob:
 
-   tutorials/*
+   tutorials/getting-started
 
 .. toctree::
    :caption: Reference
@@ -53,4 +53,14 @@
    :depth: 3
    :glob:
 
-   cookbook/*
+   cookbook/blending-orm-and-mongodb-odm
+   cookbook/implementing-array-access-for-domain-objects
+   cookbook/implementing-the-notify-changetracking-policy
+   cookbook/lookup-reference
+   cookbook/mapping-classes-to-orm-and-odm
+   cookbook/queryable-encryption
+   cookbook/resolve-target-document-listener
+   cookbook/simple-search-engine
+   cookbook/time-series-data
+   cookbook/validation-of-documents
+   cookbook/vector-search


paulinevos (approved):
Approving but I think it would be good to specify in the commit message why the asterisk notation no longer works.

greg0ire:
What's extra weird is I've tried running `composer docs` from 2.14.1, and it works locally:

```
$ ls -l docs/output/cookbook
total 120
-rw-r--r-- 1 greg greg  6481 Nov 10 20:28 blending-orm-and-mongodb-odm.html
-rw-r--r-- 1 greg greg  4135 Nov 10 20:28 implementing-array-access-for-domain-objects.html
-rw-r--r-- 1 greg greg  3286 Nov 10 20:28 implementing-the-notify-changetracking-policy.html
-rw-r--r-- 1 greg greg 26110 Nov 10 20:28 lookup-reference.html
-rw-r--r-- 1 greg greg 10677 Nov 10 20:28 mapping-classes-to-orm-and-odm.html
-rw-r--r-- 1 greg greg 11594 Nov 10 20:28 queryable-encryption.html
-rw-r--r-- 1 greg greg  6321 Nov 10 20:28 resolve-target-document-listener.html
-rw-r--r-- 1 greg greg  5587 Nov 10 20:28 simple-search-engine.html
-rw-r--r-- 1 greg greg  5200 Nov 10 20:28 time-series-data.html
-rw-r--r-- 1 greg greg 12855 Nov 10 20:28 validation-of-documents.html
-rw-r--r-- 1 greg greg  7528 Nov 10 20:28 vector-search.html
```

paulinevos:
@greg0ire I can only imagine it has something to do with the recommendation not to `composer require` PHP documentor as apparently it causes some weirdness with competing dependencies?

greg0ire:
Oh sorry, I think I'm tired, in my message I'm not checking the right thing… I should be checking the generated `sidebar.html`, not the the html pages of each page :facepalm:

greg0ire:
It appears to be working as well though. So maybe it's related to what you are saying indeed and we use different versions as `doctrine-website`?

Screenshot 2025-11-10 at 20-38-52

greg0ire:
When I run `composer outdated` on `doctrine-website`, I get this:

```
$ composer outdated
Color legend:
- patch or minor release available - update recommended
- major release available - update possible

Direct dependencies required in composer.json:
algolia/algoliasearch-client-php 4.33.0 4.34.0 API powering the features of Algolia.
doctrine/orm                     3.5.2  3.5.3  Object-Relational-Mapper for PHP
league/flysystem                 1.1.10 3.30.2 Filesystem abstraction: Many filesystems, one API.
phpdocumentor/flyfinder          1.1.0  2.0.0  Flysystem plugin to add file finding capabilities to the Filesystem entity
phpunit/phpunit                  12.4.1 12.4.2 The PHP Unit Testing framework.
symfony/cache                    7.3.5  7.3.6  Provides extended PSR-6, PSR-16 (and tags) implementations
symfony/config                   7.3.4  7.3.6  Helps you find, load, combine, autofill and validate configuration values...
symfony/console                  7.2.9  7.3.6  Eases the creation of beautiful and testable command line interfaces
symfony/css-selector             7.3.0  7.3.6  Converts CSS selectors to XPath expressions
symfony/dependency-injection     7.3.4  7.3.6  Allows you to standardize and centralize the way objects are constructed ...
symfony/error-handler            6.4.26 7.3.6  Provides tools to manage errors and ease debugging PHP code
symfony/filesystem               7.3.2  7.3.6  Provides basic utilities for the filesystem
symfony/http-kernel              7.3.4  7.3.6  Provides a structured process for converting a Request into a Response
symfony/routing                  7.3.4  7.3.6  Maps an HTTP request to a set of configuration variables
symfony/var-dumper               6.4.26 7.3.5  Provides mechanisms for walking through any arbitrary PHP variable

Transitive dependencies not required in composer.json:
doctrine/persistence             4.0.0  4.1.1  The Doctrine Persistence project is a set of shared interfaces and functi...
guzzlehttp/guzzle                7.9.3  7.10.0 Guzzle is a PHP HTTP client library
guzzlehttp/promises              2.2.0  2.3.0  Guzzle promises library
league/csv                       9.27.0 9.27.1 CSV data manipulation made easy in PHP
nette/schema                     1.3.2  1.3.3  📐 Nette Schema: validating data structures against a given Schema.
nikic/php-parser                 5.6.1  5.6.2  A PHP parser written in PHP
squizlabs/php_codesniffer        4.0.0  4.0.1  PHP_CodeSniffer tokenizes PHP files and detects violations of a defined s...
symfony/clock                    6.4.24 7.3.0  Decouples applications from the system clock
symfony/html-sanitizer           6.4.25 7.3.6  Provides an object-oriented API to sanitize untrusted HTML input for safe...
symfony/http-client              6.4.26 7.3.6  Provides powerful methods to fetch HTTP resources synchronously or asynch...
symfony/http-foundation          7.3.4  7.3.6  Defines an object-oriented layer for the HTTP specification
symfony/options-resolver         7.2.0  7.3.3  Provides an improved replacement for the array_replace PHP function
symfony/service-contracts        3.6.0  3.6.1  Generic abstractions related to writing services
symfony/string                   6.4.26 7.3.4  Provides an object-oriented API to strings and deals with bytes, UTF-8 co...
symfony/translation-contracts    3.6.0  3.6.1  Generic abstractions related to translation
webmozart/assert                 1.12.0 1.12.1 Assertions to validate method input/output with nice error messages.
```

I notice that `phpdocumentor/flyfinder` is outdated. I wonder if that package is used when wildcards are involved :thinking: 

EDIT: When I downgrade `phpdocument/flyfinder` on this repository, I do not reproduce the issue, so it's probably unrelated :disappointed:

greg0ire:
@jaapio @SenseException hi :wave: 

Maybe have some idea about what could be causing this?

greg0ire (approved):
I guess we can merge this workaround in the meantime

jaapio:
I didn't get any indications from other users that this behaviour is broken. It could be possible that there are some incompatible versions of symfony involved as we also use a number of symfony dependencies. We do test with lowest, locked and highest dependency versions, but not all in between.  

>@greg0ire I can only imagine it has something to do with the recommendation not to composer require PHP documentor as apparently it causes some weirdness with competing dependencies?

This is just about phpDocumentor itself, our libraries can be installed via composer.

I will have a look to see what I can find on our side.

jaapio:
I just checked by rendering this repositories documentation local, using the docs-builder tool it works, but we have an ugly hack in the doctrine website that might break this sidebar:

https://github.com/doctrine/doctrine-website/blob/master/lib/Docs/RST/RSTCopier.php

This class does copy all documentation into a directory when rendering the docs for the website. The sidebar is added to all pages and it's not part of the actual documentation page. Which will result in wrong paths in the actual files that are rendered, which might lead to broken menu's in some cases. 

To explain in a bit more detail. The RSTCopier is trying to find a file `sidebar.rst` in the documentation folder. Takes its content and does some regular expressions on it to fix errors in the RST format. Then it will combine the actual file content and sidebar in the target directory of the website. Those files are rendered when the website is build. 

The tricky part here is that the doctrine website project does something different than what the projects itself can do. 
I'm afraid this PR will not fix the problem.

SenseException:
> but we have an ugly hack in the doctrine website that might break this sidebar

I agree that this might be a good candidate for the culprit. I don't know if the issue, that it tries to fix, was already fixed everywhere

PR #2902, opened 2025-11-10
Upgrade to PHPUnit 11 for PHP >= 8.2 and check deprecation messages

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | -

#### Summary

In order to test deprecation messages, we must use PHPUnit 9 with Symfony bridge or PHPUnit 11+.

This PR enable PHPUnit 11 with PHP 8.2+ (and [keep 10 for PHP 8.1](https://phpunit.de/supported-versions.html)).

We have to ignore deprecation when Symfony lazy ghost are enabled, because the trait is deprecated.

### DIFF
--- a/tests/Tests/Functional/ShardKeyTest.php
+++ b/tests/Tests/Functional/ShardKeyTest.php
@@ -35,7 +35,9 @@ class ShardKeyTest extends BaseTestCase
 
     public function tearDown(): void
     {
-        $this->logger->unregister();
+        if (isset($this->logger)) {
+            $this->logger->unregister();
+        }
 
         parent::tearDown();
     }
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -190,7 +190,10 @@ jobs:
           topology: ${{ matrix.topology }}
 
       - name: "Run PHPUnit"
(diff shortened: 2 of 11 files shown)


tests/Tests/Functional/ShardKeyTest.php
```diff
@@ -35,7 +35,9 @@ public function setUp(): void
 
     public function tearDown(): void
     {
-        $this->logger->unregister();
+        if (isset($this->logger)) {
+            $this->logger->unregister();
+        }
```
alcaeus, line 40:
```suggestion
        $this->logger?->unregister();
```

alcaeus (commented):
Minor suggestion to avoid a conditional, but LGTM otherwise.

tests/Tests/Functional/ShardKeyTest.php
GromNaN: Your suggested code is not equivalent. `?->` works with `null` but not uninitialized properties. Not the same behavior as `??`.
https://3v4l.org/POp20X

GromNaN: This would work, but it would be so much less readable.
```suggestion
        ($this->logger ?? null)?->unregister();
```

alcaeus: Oh, yeah, kinda forgot those are a thing. Let's stick with your original solution then :)

alcaeus (approved):

PR #2860, opened 2025-10-17
Update .doctrine-project.json for 1.13.0

GromNaN:

PR #2906, opened 2025-11-13
Update .doctrine-project.json for 2.15

GromNaN:
- Update the current branch to `2.15`
- Add version `3.0` as work as started.
- Remove all versions older than 2 years

Reason:
> - Algolia has no unlimitied space for indices and creating indices
> - Some newer versions got rst fixes that are missing in older versions (because unmaintained and upmerge PITA) that cause at least warnings/notices during the build
> - We can remove quick-fixes or centralized solutions in the website code at some point that were done because of older project version docs

### DIFF
--- a/.doctrine-project.json
+++ b/.doctrine-project.json
@@ -6,16 +6,21 @@
     "docsSlug": "doctrine-mongodb-odm",
     "versions": [
         {
-            "name": "2.15",
-            "branchName": "2.15.x",
+            "name": "2.16",
+            "branchName": "2.16.x",
             "slug": "latest",
             "upcoming": true
         },
         {
-            "name": "2.14",
-            "branchName": "2.14.x",
-            "slug": "2.14",
+            "name": "2.15",
+            "branchName": "2.15.x",
+            "slug": "2.15",
             "current": true
+        },
+        {
+            "name": "2.14",
+            "slug": "2.14",
+            "maintained": false
         },
         {
             "name": "2.13",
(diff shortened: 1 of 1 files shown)


.doctrine-project.json
```diff
@@ -6,16 +6,27 @@
...
-            "branchName": "2.15.x",
+            "name": "3.0",
+            "branchName": "3.0.x",
+            "slug": "3.0",
+            "upcoming": true
+        },
+        {
+            "name": "2.16",
+            "branchName": "2.16.x",
             "slug": "latest",
```
alcaeus, line 17: Shouldn't `latest` refer to a stable release?

.doctrine-project.json
```diff
@@ -57,56 +68,11 @@
...
-            "name": "2.1",
-            "slug": "2.1",
-            "maintained": false
-        },
-        {
-            "name": "2.0",
-            "slug": "2.0",
-            "maintained": false
-        },
         {
```
alcaeus: 1.3 gets ~500 installs per month according to packagist. I believe we can remove this version as well.

.doctrine-project.json
GromNaN: Not in other Doctrine projects: https://github.com/doctrine/mongodb-odm/pull/2860#discussion_r2441353000

alcaeus: Got it, makes sense!

alcaeus (approved):

PR #2908, opened 2025-11-14
Replace annotations with attributes in migrating-schemas.rst

mickverm:
|      Q       |   A
|------------- | -----------
| BC Break     | no
| Fixed issues | https://github.com/doctrine/mongodb-odm/issues/2909

#### Summary

Replaced the leftover annotations with attributes.

### DIFF
--- a/docs/en/reference/migrating-schemas.rst
+++ b/docs/en/reference/migrating-schemas.rst
@@ -137,7 +137,7 @@ before normal hydration.
         }
     }
 
-The annotation is defined with one or a list of field names. During hydration,
+The attribute is defined with one or a list of field names. During hydration,
 these fields will be checked in order and, for each field present, the annotated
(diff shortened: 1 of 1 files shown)


docs/en/reference/migrating-schemas.rst
```diff
@@ -137,7 +137,7 @@ before normal hydration.
         }
     }
 
-The annotation is defined with one or a list of field names. During hydration,
+The attribute is defined with one or a list of field names. During hydration,
 these fields will be checked in order and, for each field present, the annotated
```
mickverm, line 141: Not sure if `annotated` should be replaced with something else?

GromNaN: "annotated" is the correct word.

GromNaN (approved):
Thank you @mickverm.

PR #2911, opened 2025-11-14
Deprecate `Configuration::getProxyDir` and `setProxyDir`

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | -

#### Summary

The proxy dir is not used when native lazy objects are used. Trigger the deprecation only when it's enabled.

### DIFF
--- a/UPGRADE-2.16.md
+++ b/UPGRADE-2.16.md
@@ -1,0 +1,10 @@
+# UPGRADE FROM 2.15 to 2.16
+
+## Lazy Proxy Directory
+
+Using proxy classes with PHP 8.4+ is deprecated, only native lazy objects will
+be supported in MongoDB ODM 3.0.
+
+Calling `Doctrine\ODM\MongoDB\Configuration::setProxyDir()` or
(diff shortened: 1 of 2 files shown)


greg0ire (commented):
This should be mentioned in the upgrade guide, and I think the annotations are not OK since the deprecation is conditional

alcaeus:
> This should be mentioned in the upgrade guide, and I think the annotations are not OK since the deprecation is conditional

I think the annotations are fine, since we'll be removing the methods along with all non-native proxies in 3.0.

GromNaN:
> This should be mentioned in the upgrade guide,

I added the upgrade instructions.

> and I think the annotations are not OK since the deprecation is conditional

Calling this methods is actually deprecated. I could also mark them as `@internal` because we still need to use them internally. But we are going to remove them in 3.0. https://github.com/doctrine/mongodb-odm/pull/2914

I don't want to have to keep this methods in 3.0 because we can't deprecate them in 2.x.

greg0ire:
> Calling this methods is actually deprecated.

Oh OK great then 👍

greg0ire (approved):

PR #2912, opened 2025-11-14
[BC break] Move `doctrine/cache` to optional dependencies

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | https://github.com/doctrine/mongodb-odm/issues/2762#issuecomment-3385826776

#### Summary

`doctrine/cache` is deprecated and archived.

Motivation for introducing a breaking change:
- When installing `doctrine/mongodb-odm` with composer, you get a message even if `doctrine/cache` is not actively used.
> Package doctrine/cache is abandoned, you should avoid using it. No replacement was suggested.
- The next major version will require PHP 8.4.

This is a breaking change only when calling `Configuration::getMetadataCacheImpl()`. The workaround is easy: add `doctrine/cache` as project dependency.

I consider projects that use `doctrine/cache` directly without explicitly requiring the package to be in error.

That's why I'm adding a conflict with `doctrine/mongodb-odm-bundle < 5`, which did not have an explicit dependency but used the `Doctrine\Common\Cache` classes:
https://github.com/doctrine/DoctrineMongoDBBundle/blob/48c29009c24ff74c47782c0c9c2736f1f5fd9b0e/DependencyInjection/DoctrineMongoDBExtension.php#L14-L15

### DIFF
--- a/UPGRADE-2.16.md
+++ b/UPGRADE-2.16.md
@@ -1,5 +1,11 @@
 # UPGRADE FROM 2.15 to 2.16
 
+## Package `doctrine/cache` no longer required
+
+If you use `Doctrine\ODM\MongoDB\Configuration::getMetadataCacheImpl()`,
+then you need to require `doctrine/cache` explicitly in `composer.json`;
+or use `Doctrine\ODM\MongoDB\Configuration::getMetadataCache()` instead.
+
 ## Lazy Proxy Directory
 
 Using proxy classes with PHP 8.4+ is deprecated, only native lazy objects will
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,6 @@
     "require": {
         "php": "^8.1",
         "ext-mongodb": "^1.21 || ^2.0",
-        "doctrine/cache": "^1.11 || ^2.0",
         "doctrine/collections": "^1.5 || ^2.0",
         "doctrine/event-manager": "^1.0 || ^2.0",
         "doctrine/instantiator": "^1.1 || ^2",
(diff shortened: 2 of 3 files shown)


greg0ire (commented):
The quick fix and the proper fix should probably be mentioned in an upgrade guide.

malarzm (approved):

UPGRADE-2.16.md
```diff
@@ -0,0 +1,10 @@
+# UPGRADE FROM 2.15 to 2.16
+
+## Package `doctrine/cache`
+
+The package `doctrine/cache` no longer required.
```
greg0ire:
```suggestion
The package `doctrine/cache` is no longer required.
```

greg0ire (approved):

UPGRADE-2.16.md
```diff
@@ -0,0 +1,10 @@
+# UPGRADE FROM 2.15 to 2.16
+
+## Package `doctrine/cache`
+
+The package `doctrine/cache` is no longer required.
+
+If you use `Doctrine\ODM\MongoDB\Configuration::getMetadataCacheImpl()`,
+then you need to require `doctrine/cache` explicitly in your `composer.json`;
+or use `Doctrine\ODM\MongoDB\Configuration::getMetadataCache()` instead.
```
jmikola, line 7:
```suggestion
If you use `Doctrine\ODM\MongoDB\Configuration::getMetadataCacheImpl()`,
then you need to require `doctrine/cache` explicitly in `composer.json`;
or use `Doctrine\ODM\MongoDB\Configuration::getMetadataCache()` instead.
```

Alternatively, you can say "your project's", but I think less words makes this more readable.

UPGRADE-2.16.md
```diff
@@ -0,0 +1,10 @@
+# UPGRADE FROM 2.15 to 2.16
+
+## Package `doctrine/cache`
+
+The package `doctrine/cache` is no longer required.
```
jmikola:
```suggestion
## Package `doctrine/cache` no longer required
```

alcaeus (approved):

paulinevos (approved):

PR #2917, opened 2025-11-20
Run tests with PHP 8.5 and fix deprecations

GromNaN:


SenseException (approved):

src/Persisters/DocumentPersister.php
```diff
@@ -994,7 +994,7 @@ public function prepareFieldName(string $fieldName): string
      */
     public function addDiscriminatorToPreparedQuery(array $preparedQuery): array
     {
-        if (isset($preparedQuery[$this->class->discriminatorField]) || $this->class->discriminatorField === null) {
+        if (isset($preparedQuery[$this->class->discriminatorField ?? '']) || $this->class->discriminatorField === null) {
```
IonBazan:
```suggestion
        if ($this->class->discriminatorField === null || isset($preparedQuery[$this->class->discriminatorField])) {
```

Reversing the condition order is simpler.

GromNaN: Smart, thank you.

Issue #2916, opened 2025-11-19
Regression with handling of unmapped properties in proxy objects


GromNaN:
_Originally posted by @melkamar in https://github.com/doctrine/mongodb-odm/issues/2698#issuecomment-3552176325_

Sorry about necroposting, but after upgrading to 2.11 recently, I think we have stumbled onto this and it's causing a "bug" in our code. I used quotes because I am not sure if our observed behavior is a bug or works-as-intended.

Our simplified scenario:

- `Document` has `mappedProperty` and `unmappedProperty`
- When the document is fetched from the database in its lazy state, `unmappedProperty` is `null`
- We have a service that inserts stuff into the `unmappedProperty` when the document is loaded: `#[AsDocumentListener(event: Events::postLoad)] (...)`
- Our assumption (which seems to have broken with this PR) is that when we access _any_ property, we can rely on the document to get fully loaded.
- In order to work around this, we had to resort to make a "dummy fetch" to force the loading.

```php
class Document
{
    private int $mappedProperty;
    private ServiceInterface $unmappedProperty;

    public function getUnmappedProperty()
    {
        $this->mappedProperty; // Dummy call to get proxy to initialize
        return $this->unmappedProperty;
    }

    /**
     * Called from a listener on Events::postLoad
     */
    public function initialize(ServiceInterface $service)
    {
        $this->unmappedProperty = $service;
    }
}
```

Is this expected? Is our (admittedly awkward) scenario unsupported? Or should I open an issue with this?

GromNaN:
Indeed #2698 introduced a behavior change; but it's actually a bug fix as we don't want to trigger a request to load a document from the database when an unmapped property is accessed.

I wrote a test that confirms this behavior. It fails with PHP 8.4 native lazy objects because I did not think about this behavior. I need to call `ReflectionProperty::skipLazyInitialization` for the unmapped properties. (fixed by https://github.com/doctrine/mongodb-odm/pull/2922)

<details>
<summary>Functional test</summary>

```php
<?php

declare(strict_types=1);

namespace Doctrine\ODM\MongoDB\Tests\Functional\Ticket;

use Doctrine\ODM\MongoDB\Events;
use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
use Doctrine\ODM\MongoDB\Tests\BaseTestCase;

class GH2916Test extends BaseTestCase
{
    public function testUmappedPropertyWithLazyDocument(): void
    {
[...]

PR #2922, opened 2025-11-21
Skip native lazy object initialization for unmapped properties

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | yes-ish (see https://github.com/doctrine/mongodb-odm/issues/2916)
| Fixed issues | -

#### Summary

Re-implement https://github.com/doctrine/mongodb-odm/pull/2698 for native lazy objects.

Using the method [`ReflectionProperty::skipLazyInitialization`](https://www.php.net/manual/en/reflectionproperty.skiplazyinitialization.php), accessing properties that are not mapped in Doctrine will not initialize the document object.

### DIFF
--- a/src/Proxy/Factory/NativeLazyObjectFactory.php
+++ b/src/Proxy/Factory/NativeLazyObjectFactory.php
@@ -12,8 +12,10 @@ use Doctrine\ODM\MongoDB\Utility\LifecycleEventManager;
 use Doctrine\Persistence\NotifyPropertyChanged;
 use LogicException;
 use ReflectionClass;
+use ReflectionProperty;
 use WeakMap;
 
+use function array_key_exists;
 use function count;
 
 use const PHP_VERSION_ID;
@@ -26,6 +28,9 @@ class NativeLazyObjectFactory implements ProxyFactory
 
     private readonly UnitOfWork $unitOfWork;
     private readonly LifecycleEventManager $lifecycleEventManager;
+
+    /** @var array<class-string, ReflectionProperty[]> */
+    private array $skippedProperties = [];
 
     public function __construct(
         DocumentManager $documentManager,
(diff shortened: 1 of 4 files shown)


src/Proxy/Factory/NativeLazyObjectFactory.php
```diff
@@ -68,13 +73,40 @@ public function getProxy(ClassMetadata $metadata, $identifier): object
...
 
         return $proxy;
     }
 
+    /** @return ReflectionProperty[] */
+    private function getSkippedProperties(ClassMetadata $metadata): array
+    {
+        if (isset($this->skippedProperties[$metadata->name])) {
+            return $this->skippedProperties[$metadata->name];
+        }
```
jmikola, line 92: Memoization!

src/Proxy/Factory/NativeLazyObjectFactory.php
```diff
@@ -68,13 +73,40 @@ public function getProxy(ClassMetadata $metadata, $identifier): object
...
+
+        $skippedProperties = [];
+        foreach ($metadata->reflClass->getProperties() as $property) {
+            if (array_key_exists($property->name, $metadata->propertyAccessors)) {
+                continue;
+            }
+
+            if ($property->isVirtual()) {
```
jmikola, line 100: I'm pleasantly surprised that the [ReflectionProperty docs](https://www.php.net/manual/en/reflectionproperty.isvirtual.php#refsect1-reflectionproperty.isvirtual-examples) explain what virtual properties are.

src/Proxy/Factory/NativeLazyObjectFactory.php
```diff
@@ -68,13 +73,40 @@ public function getProxy(ClassMetadata $metadata, $identifier): object
...
+    private function getSkippedProperties(ClassMetadata $metadata): array
+    {
+        if (isset($this->skippedProperties[$metadata->name])) {
+            return $this->skippedProperties[$metadata->name];
+        }
+
+        $skippedProperties = [];
+        foreach ($metadata->reflClass->getProperties() as $property) {
+            if (array_key_exists($property->name, $metadata->propertyAccessors)) {
+                continue;
```
jmikola, line 97: Just to confirm, every mapped field should have its own property accessor according to [`ClassMetadata::mapField()`](https://github.com/doctrine/mongodb-odm/blob/2.15.1/src/Mapping/ClassMetadata.php#L2626), correct?

So this `foreach` look is really just checking for unmapped, non-virtual properties.

Will this logic be useful in any other contexts? If so, it may be worth capturing this logic in a ClassMetadata method.

GromNaN: It **might** be an optimization to store the list of unmapped properties in the ClassMetadata, and leverage the cache instead of iterating over all the class properties. Also, there **might** be an issue if the ClassMetadata is modified after the first lazy object is created, as this list is not updated.

I'd rather keep this simple design for now.

tests/Tests/Functional/ReferencesTest.php
```diff
@@ -28,6 +29,15 @@
 
 class ReferencesTest extends BaseTestCase
 {
+    public function testSkipInitializationForUnmappedProperties(): void
+    {
+        $loadedDocument = $this->dm->getReference(DocumentWithUnmappedProperties::class, '123');
+        $this->assertInstanceOf(DocumentWithUnmappedProperties::class, $loadedDocument);
+
+        // Accessing unmapped property should not initialize the document
```
alcaeus: Should we add a check using `ReflectionClass` to ensure this is actually still not initialised?

GromNaN: Done. This was necessary because the class had only 2 properties, and it was immediately considered as initialized.

I will dig because there might be a bug similar to https://github.com/doctrine/mongodb-odm/pull/2840#discussion_r2448634563

GromNaN: I'm happy to say that there's no issue.

alcaeus (approved):

PR #2923, opened 2025-11-21
Add `onReference` event when a new lazy object reference is created

### LINKED ISSUE #2916: Regression with handling of unmapped properties in proxy objects

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | Fix https://github.com/doctrine/mongodb-odm/issues/2916

#### Summary

This feature enable the injection of dependencies into lazy documents, even if they are not loaded.

I am willing to discuss whether this feature is a good idea.

### DIFF
--- a/docs/en/reference/events.rst
+++ b/docs/en/reference/events.rst
@@ -159,6 +159,13 @@ the life-time of their registered documents.
    document has been loaded into the current DocumentManager from the
    database or after the refresh operation has been applied to it.
 -
+   ``onReference`` - The onReference event occurs when a lazy document reference
+   is created. The data of the document are not loaded yet at this point; if
+   you access any mapped field of the document, a database query will be
+   triggered to load the document data. Which defeats the purpose of using
+   lazy references in the first place.
+   This can be used to inject dependencies into unmapped properties.
+-
    ``loadClassMetadata`` - The loadClassMetadata event occurs after the
    mapping metadata for a class has been loaded from a mapping source
    (attributes/xml).
--- a/src/Events.php
+++ b/src/Events.php
@@ -87,6 +87,18 @@ final class Events
     public const postLoad = 'postLoad';
 
     /**
+     * The onReference event occurs when a lazy document reference is created.
+     *
+     * Note that the data of the document are not loaded yet at this point; if
+     * you access any mapped field of the document, a database query will be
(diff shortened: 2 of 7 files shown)


alcaeus:
> This feature enable the injection of dependencies into lazy documents, even if they are not loaded.
> 
> I am willing to discuss whether this feature is a good idea.

I'm not opposed to the idea of such an event, but I think that injecting dependencies into lazy documents isn't it. Lazy documents are supposed to be transparent, which we're now achieving better than before by using native PHP functionality. Note that this wouldn't work if the document isn't lazy-loaded, so this would always have to be used alongside a `PostLoad` event for it to work properly in all scenarios.

GromNaN:
Closing this PR.

This new event is just a clumsy solution that addresses a non-existent need. To trigger the `postLoad` event, you need to initialize the document with `$documentManager->initializeObject($document);` or by accessing any lazy-loaded property.

PR #2930, opened 2025-11-24
Remove every mention to PHP < 8.4, as this version is required

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | -


tests/Documents84/PropertyHooks/User.php
```diff
@@ -1,6 +1,7 @@
-<?php // phpcs:ignoreFile
```
IonBazan, line 1: Can we keep this line instead of touching `phpcs.xml.dist`?

GromNaN: Good point, I forgot that you could be specific in the section of code to ignore. I used `// phpcs:disable`.

IonBazan (approved):

PR #2931, opened 2025-11-24
[3.x] Remove support for Doctrine Annotations

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | -

#### Summary

- Remove AnnotationDriver and its configuration
- Remove `@Annotation` from attribute classes
- Remove annotations from test documents, already replaced by native attributes


IonBazan (approved):
- We should also update the `docs/annotations-reference.md` accordingly.
- PHP attributes are still in `Annotations` namespace - this is something we won't be able to change easily and might be confusing to new users as docs and other places still use annotations/attributes interchangeably.

IonBazan (approved):

mickverm:
> * We should also update the `docs/annotations-reference.md` accordingly.
> * PHP attributes are still in `Annotations` namespace - this is something we won't be able to change easily and might be confusing to new users as docs and other places still use annotations/attributes interchangeably.

What about using the Annotation namespace as an alias of the Attribute namespace, like Symfony did? And deprecating the annotation namespace later?

https://github.com/symfony/symfony/blob/a518bc2c1a6cf59b65169dadcbeb8f3b91e2c06c/src/Symfony/Component/Routing/Annotation/Route.php

IonBazan:
@mickverm this is possible, even without aliasing - using inheritance but we would need to keep a copy of all Attribute classes. That should target 2.16.x or 2.17.x then once we decide whether we should pursue that option for v3.
We also need a Rector set to automate the renaming.

GromNaN:
We have to check if that's possible with attributes, because they are read without actually loading the classes.
I would do the same as the namespace migration of Twig: add aliases in the attributes namespace in 2.x and move the concrete classes to the Attribute namespace in 3.0, keeping aliases in the Annotations namespace as deprecated.

GromNaN:
See https://github.com/doctrine/mongodb-odm/pull/2933

PR #2933, opened 2025-11-25
Change namespace of attribute classes from `Annotations` to `Attribute`

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no (deprecation)
| Fixed issues | Related https://github.com/doctrine/mongodb-odm/pull/2931

#### Summary

- Move the classes from `Doctrine\ODM\MongoDB\Mapping\Annotations` to `Doctrine\ODM\MongoDB\Mapping\Attribute`. Removing the `@Annotation` so that they can't be used for legacy annotations.
- For the classes in the `Annotations` namespace, they extend those from the `Attribute` namespace. Adding the legacy `@Annotation` for BC.
- The `EncryptQuery` enum is moved to the parent `Doctrine\ODM\MongoDB\Mapping` namespace
- The abstract annotation classes are aliased to the new abstract attribute classes, so that it's not breaking type comparison
- Documentation updated

Related initiative in Symfony: https://github.com/symfony/symfony/issues/51381


GromNaN:
This will force users to migrate without offering anything in return, except for consistency in the namespace with other projects (and not yet ORM). ~We can only do this if ORM also decides to do so.~

GromNaN:
In fact, ORM don't have this subnamespace (example: [`Doctrine\ORM\Mapping\Entity`](https://github.com/doctrine/orm/blob/3.5.x/src/Mapping/Entity.php)). So, we should consider whether we want to keep the namespace or move all the attribute classes to `Doctrine\ODM\MongoDB\Mapping`. 
@alcaeus you had the idea that we should use the attribute classes as model classes for all the class metadata. In that case, that would make sense to remove the subnamespace `Attribute`.

malarzm (commented):
I can't quite grasp it, but there's a "class_alias was problematic in the past" dangling in my head. Saying it out loud in case I'd jog somebody's memory with it. Alernative would be to:

1. Move classes to Attributes (and drop final, leave it as a phpdoc)
2. Classes in Annotations would extend their counterpart, emit a deprecation in ctor and forward everything
3. Remove Annotations/* in 3.0 and bring back `final` in Attribute/*

Also I'd like to see docs updated for 2.16 - folks copying documentation should not be greeted by a deprecation :)

greg0ire:
> I can't quite grasp it, but there's a "class_alias was problematic in the past" dangling in my head. Saying it out loud in case I'd jog somebody's memory with it.

@malarzm I wrote [an article about that](https://dev.to/greg0ire/how-to-deprecate-a-type-in-php-48cf), and  @alcaeus wrote [a medium article](https://alcaeus.medium.com/how-to-break-an-entire-ecosystem-by-publishing-a-release-b6aaab2b8aaa) about the aftermath of doing this for `doctrine/persistence` as well.

malarzm:
@greg0ire ha! Thanks for the links, I tried to look in our blog and google but prolly I was too specific with "orm namespace change" :)

src/Mapping/Annotations/File/ChunkSize.php
```diff
@@ -4,19 +4,11 @@
...
-    {
-        parent::__construct('chunkSize', 'int', false, [], null, true);
-    }
-}
+use function class_exists;
+use function trigger_deprecation;
+
+trigger_deprecation('doctrine/mongodb-odm', '2.16', 'Namespace "Doctrine\ODM\MongoDB\Mapping\Annotations" is deprecated, use "Doctrine\ODM\MongoDB\Mapping\Attribute" instead.');
+
+class_exists(ChunkSize::class);
```
IonBazan: Because we are only moving `Annotations`, can we opt to a simplified version of deprecation by extending the new classes instead of creating aliases?

```php
/**
 * @deprecated
 */
class ChunkSize extends Doctrine\ODM\MongoDB\Mapping\Attribute\ChunkSize
{
}
```

I know this approach is a BC break when we return such object but I don't think we ever do that. This also works better with IDE autocompletion and composer classmap.

Otherwise, I don't think current approach is aligned with articles linked above either. @malarzm @greg0ire perhaps you have more context.

Something similar was applied to `Route` annotation in Symfony: https://github.com/sensiolabs/SensioFrameworkExtraBundle/blob/2f886f4b31f23c76496901acaedfedb6936ba61f/src/Configuration/Route.php#L14-L24

UPGRADE-2.16.md
```diff
@@ -1,5 +1,27 @@
...
+## Attribute namespaces
+
+Doctrine annotations are already deprecated in favor of PHP attributes. As of MongoDB ODM 2.16, the
+attribute namespaces have been changed from `Doctrine\ODM\MongoDB\Mapping\Annotations` to
+`Doctrine\ODM\MongoDB\Attribute`. The old annotation namespaces will continue to work for the time being, but
+they are deprecated and will be removed in 3.0.
+
+```diff
+- use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
++ use Doctrine\ODM\MongoDB\Attribute as ODM;
```
GromNaN:
```suggestion
+ use Doctrine\ODM\MongoDB\Mapping\Attribute as ODM;
```

src/Mapping/Annotations/File/ChunkSize.php
GromNaN: Using `class_alias` is a lot simpler than removing the final, extending the class and duplicate all the `#[Attribute]` and `/** @Annotation */` annotations.

I've re-added the annotation classes for static analysis, they are never loaded.

The current approches is inspired by the [Twig](https://github.com/twigphp/Twig/blob/2.x/lib/Twig/BaseNodeVisitor.php) and [Symfony Serializer](https://github.com/symfony/symfony/blob/7.4/src/Symfony/Component/Serializer/Annotation/Context.php) migration.

I don't see any issue with the `class_alias` approach. PHPStorm improved its static analysis and actually interprets the class alias as if it was a class declaration (it seems it was not the case few years ago):
image

GromNaN:
I updated the PR, including the docs to use inheritance. That way the new classes in the `Attribute` namespace cannot be used as annotation.

New PR description:
> - Move the classes from `Doctrine\ODM\MongoDB\Mapping\Annotations` to `Doctrine\ODM\MongoDB\Mapping\Attribute`. Removing the `@Annotation` so that they can't be used for legacy annotations.
> - For the classes in the `Annotations` namespace, they extend those from the `Attribute` namespace. Adding the legacy `@Annotation` for BC.
> - The `EncryptQuery` enum is moved to the parent `Doctrine\ODM\MongoDB\Mapping` namespace
> - The abstract annotation classes are aliased to the new abstract attribute classes, so that it's not breaking type comparison
> - Documentation updated

I've kept the tests unchanged, ensuring full BC of existing code.

src/Mapping/Attribute/Document.php
```diff
@@ -0,0 +1,52 @@
...
+
+use Attribute;
+
+/**
+ * Identifies a class as a document that can be stored in the database
+ *
+ * @final
+ */
+#[Attribute(Attribute::TARGET_CLASS)]
+class Document extends AbstractDocument implements MappingAttribute
```
IonBazan: Let's move this to `AbstractDocument`, `AbstractField`, `AbstractIndex`:
```suggestion
class Document extends AbstractDocument
```
Less things to remember about in case we add another attribute which extends that class.

src/Mapping/Annotations/File/ChunkSize.php
```diff
@@ -6,17 +6,15 @@
...
 
 /**
+ * @deprecated Use \Doctrine\ODM\MongoDB\Mapping\Attribute\File\ChunkSize instead
+ *
  * @Annotation
  * @NamedArgumentConstructor
  */
 #[Attribute(Attribute::TARGET_PROPERTY)]
-final class ChunkSize extends AbstractField
+final class ChunkSize extends \Doctrine\ODM\MongoDB\Mapping\Attribute\File\ChunkSize implements Annotation
```
alcaeus: Nit: adding a use statement with alias could make this a little nicer to read:
```php
use Doctrine\ODM\MongoDB\Mapping\Attribute\File\ChunkSize as ChunkSizeAttribute;

final class ChunkSize extends ChunkSizeAttribute implements Annotation
```

Feel free to leave as is if you don't want to go through all annotation classes.

src/Mapping/Attribute/AbstractDocument.php
```diff
@@ -0,0 +1,14 @@
...
+namespace Doctrine\ODM\MongoDB\Mapping\Attribute;
+
+use function class_alias;
+
+abstract class AbstractDocument implements MappingAttribute
+{
+}
+
+// @phpstan-ignore class.notFound
+class_alias(AbstractDocument::class, \Doctrine\ODM\MongoDB\Mapping\Annotations\AbstractDocument::class);
```
alcaeus, line 15: Noted that this works in conjunction with the `class_exists` call in the `AbstractDocument` annotation file.

src/Mapping/Annotations/AbstractDocument.php
```diff
@@ -4,6 +4,14 @@
...
-abstract class AbstractDocument implements Annotation
-{
+use function class_exists;
+
+class_exists(\Doctrine\ODM\MongoDB\Mapping\Attribute\AbstractDocument::class);
+
+// @phpstan-ignore-next-line if.alwaysFalse
+if (false) {
+    /** @deprecated Use \Doctrine\ODM\MongoDB\Mapping\Attribute\AbstractDocument instead */
+    abstract class AbstractDocument extends \Doctrine\ODM\MongoDB\Mapping\Attribute\AbstractDocument
```
alcaeus, line 14: The new class no longer implements the `Annotation` class. We can fix this for the static analysis class, but it won't work for the `class_alias` declaration in the attribute.

```suggestion
    abstract class AbstractDocument extends \Doctrine\ODM\MongoDB\Mapping\Attribute\AbstractDocument implements Annotation
```

src/Mapping/Driver/AttributeDriver.php
```diff
@@ -6,10 +6,8 @@
 
 use Doctrine\Common\Annotations\Reader;
 use Doctrine\ODM\MongoDB\Events;
-use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
-use Doctrine\ODM\MongoDB\Mapping\Annotations\AbstractIndex;
-use Doctrine\ODM\MongoDB\Mapping\Annotations\ShardKey;
-use Doctrine\ODM\MongoDB\Mapping\Annotations\TimeSeries;
+use Doctrine\ODM\MongoDB\Mapping\Annotations\Indexes;
```
alcaeus, line 9: This import should be removed. It's used in line 112 and 256 where it replaced an `ODM\Indexes` use.

src/Mapping/Driver/AttributeDriver.php
```diff
@@ -22,6 +20,8 @@
 use ReflectionMethod;
 use ReflectionProperty;
 
+use function array_any;
+use function array_find;
```
alcaeus, line 24: I like the use of `array_any` and `array_find` in this class 👏

src/Mapping/Driver/AttributeReader.php
```diff
@@ -37,21 +37,21 @@ public function getPropertyAttributes(ReflectionProperty $property): array
...
     private function convertToAttributeInstances(array $attributes): array
     {
         $instances = [];
 
         foreach ($attributes as $attribute) {
             $attributeName = $attribute->getName();
-            // Make sure we only get Doctrine Annotations
-            if (! is_subclass_of($attributeName, Annotation::class)) {
+            // Make sure we only get MongoDB ODM attribute classes
+            if (! is_subclass_of($attributeName, MappingAttribute::class)) {
```
alcaeus, line 49: Noted that this doesn't change existing behaviour, as all annotations extend the attribute classes.

alcaeus (changes requested):
Not sure how to proceed with the missing `Annotation` interface on the attribute class aliases. Found one issue where annotation classes were used in the attribute reader.

src/Mapping/Attribute/AbstractDocument.php
GromNaN: Exact, having the `class_alias` in the same file as the class declaration ensure we always have both declared, even if the attribute is used first.
And the `class_exists` only triggers the autoloader.

src/Mapping/Driver/AttributeDriver.php
GromNaN: This class is only in the `Annotations` namespace. I did not create the `Attribute` class because it's already deprecated.

src/Mapping/Annotations/File/ChunkSize.php
GromNaN: Thanks Copilot, I'm doing this change with minimal efforts.

src/Mapping/Annotations/AbstractDocument.php
GromNaN: The abstract class cannot be used as an annotation. Did you find an issue?

GromNaN (commented):
> Not sure how to proceed with the missing Annotation interface on the attribute class aliases. Found one issue where annotation classes were used in the attribute reader.

Do you have an example, I don't see the issue.

src/Mapping/Driver/AttributeDriver.php
alcaeus: Ah, I see. No worries then.

src/Mapping/Annotations/AbstractDocument.php
alcaeus: This might create a BC break for people with a custom document annotation, but I'd argue that it's very unlikely someone would do that. If you're fine with the potential BC break, so am I :)

GromNaN: I added a note to the backlog and marked the new abstract attribute classes as internal.

alcaeus (approved):

GromNaN:
Before merging, we need to update the bundle: 
https://github.com/doctrine/DoctrineMongoDBBundle/blob/9c2ccf08a08fcf739d346810adbbd322ed9c0781/src/DependencyInjection/DoctrineMongoDBExtension.php#L535-L552

PR #2935, opened 2025-11-25
[3.x] Add typed properties

IonBazan:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | -

#### Summary

Adds typed properties and missing typehints (especially `mixed`, `callable` and union types which were not available earlier).

There are still many places to fix but let's consider this as first round.
Attributes were untouched to avoid conflicts with other changes.


src/Repository/DocumentRepository.php
```diff
@@ -57,11 +51,9 @@ class DocumentRepository implements ObjectRepository, Selectable
      * @param ClassMetadata   $classMetadata The class metadata.
      * @phpstan-param ClassMetadata<T>   $classMetadata The class metadata.
      */
-    public function __construct(DocumentManager $dm, UnitOfWork $uow, ClassMetadata $classMetadata)
+    public function __construct(protected DocumentManager $dm, protected UnitOfWork $uow, ClassMetadata $classMetadata)
```
GromNaN, line 54: You can use CPP for `$classMetadata`.
```suggestion
    public function __construct(protected DocumentManager $dm, protected UnitOfWork $uow, protected ClassMetadata $classMetadata)
```

src/Repository/DocumentRepository.php
```diff
@@ -92,7 +84,7 @@ public function createAggregationBuilder(): AggregationBuilder
      * @throws MappingException
      * @throws LockException
      */
-    public function find($id, int $lockMode = LockMode::NONE, ?int $lockVersion = null): ?object
+    public function find(mixed $id, int $lockMode = LockMode::NONE, ?int $lockVersion = null): ?object
```
GromNaN, line 87: I trying to find if `$id` type can be more specific, but no. Any object or scalar can be converted to MongoDB value by a type.

tests/Tests/DocumentManagerTest.php
```diff
@@ -136,7 +136,7 @@ public static function dataMethodsAffectedByNoObjectArguments(): array
     #[DataProvider('dataMethodsAffectedByNoObjectArguments')]
     public function testThrowsExceptionOnNonObjectValues(string $methodName): void
     {
-        $this->expectException(InvalidArgumentException::class);
+        $this->expectException(TypeError::class);
         $this->dm->$methodName(null);
     }
```
GromNaN, line 141: Usually we don't test PHP native behavior of rejecting invalid value. But that's good to have it now to ensure we reject the same values as before.

src/Mapping/Driver/XmlDriver.php
GromNaN, line 1: Unrelated, I wonder if we should keep 2 XML driver classes?

src/Mapping/ClassMetadata.php
```diff
@@ -718,10 +694,8 @@
      * where a discriminator field is used.</b>
      *
      * @see discriminatorField
-     *
-     * @var class-string|null
```
GromNaN, line 721: You removed the `class-string` type. It was incorrect, right? We should fix this type in the current 2.x.

src/Repository/DocumentRepository.php
IonBazan: Yes, this is actually consistent with `ObjectRepository` types too.

src/Mapping/ClassMetadata.php
IonBazan: Yes, I think this is wrong. The `discriminatorValue` holds the value that is used as key in `discriminatorMap` to get Document FQCN. I thought it can only hold `string` but it appears that `int` was a valid use case in #2712.
Let me create a fix for v2 as it's considered a bug.

src/Repository/DocumentRepository.php
IonBazan: Unfortunately, the constructor argument name does not match the property name. To reduce the BC impact, I decided to keep it this way.

src/Mapping/ClassMetadata.php
IonBazan: Looks like it's a bigger problem, reported in #2936. For now, I will revert this particular change.

tests/Tests/DocumentManagerTest.php
IonBazan: Let's leave it here for now, in case we have to relax the types here in the future.

src/Aggregation/Stage/Group.php
```diff
@@ -18,16 +17,6 @@ class Group extends Operator implements GroupAccumulatorOperators
...
-    /** @var Expr */
-    protected $expr;
-
-    public function __construct(Builder $builder)
-    {
-        parent::__construct($builder);
-
-        $this->expr = $builder->expr();
-    }
-
```
GromNaN, line 30: Good catch, this duplicated the code from the parent.

src/Iterator/CachingIterator.php
```diff
@@ -77,15 +76,12 @@ public function toArray(): array
...
     /** @return TValue|false */
-    #[ReturnTypeWillChange]
-    public function current()
+    public function current(): mixed
     {
         return current($this->items);
     }
 
-    /** @return mixed */
-    #[ReturnTypeWillChange]
```
GromNaN, line 87: The return type is modified, but thats fine since this class is internal.

src/Mapping/Driver/XmlDriver.php
```diff
@@ -76,8 +76,7 @@ class XmlDriver extends FileDriver
         ],
     ];
 
-    /** @param string|null $fileExtension */
-    public function __construct($locator, $fileExtension = self::DEFAULT_FILE_EXTENSION)
+    public function __construct($locator, ?string $fileExtension = self::DEFAULT_FILE_EXTENSION)
```
GromNaN: You can use the `$locator` type from [`doctrine/collection`](https://github.com/doctrine/persistence/blob/4fd46d131a1cb642930e9307ff36de6930fe6248/src/Mapping/Driver/FileDriver.php#L43).


```suggestion
    public function __construct(string|array|FileLocator $locator, ?string $fileExtension = self::DEFAULT_FILE_EXTENSION)
```

GromNaN (approved):
Thank you Ion!

Issue #2850, opened 2025-10-09
Add return types to `PersistentCollectionTrait`


GromNaN:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | no
| RFC         | no
| BC Break    | yes

#### Summary

The next major version of [`doctrine/collection`](https://github.com/doctrine/collections/) will add return types to the interface methods. We need to add this return types in the trait and the generated collection code in order to support it.

- [PersistentCollectionTrait](https://github.com/doctrine/mongodb-odm/blob/2.13.x/lib/Doctrine/ODM/MongoDB/PersistentCollection/PersistentCollectionTrait.php)
- [Generated code](https://github.com/doctrine/mongodb-odm/blob/2.13.x/lib/Doctrine/ODM/MongoDB/PersistentCollection/DefaultPersistentCollectionGenerator.php)

Also, the trait should be `@internal` and the generated collection classes `@final`. For now, this need to be done in a BC compatible way.

GromNaN:
Done in https://github.com/doctrine/mongodb-odm/pull/2939

PR #2939, opened 2025-11-29
[3.x] Added parameter and return types to the PersistentCollectionTrait 

krzysztof-ciszewski:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | #2850 

#### Summary

Added parameter and return types to the PersistentCollectionTrait in preparation for doctrine/collection update

### DIFF
--- a/src/PersistentCollection/PersistentCollectionTrait.php
+++ b/src/PersistentCollection/PersistentCollectionTrait.php
@@ -334,11 +334,7 @@ trait PersistentCollectionTrait
         return $this->coll->last();
     }
 
-    /**
-     * @param mixed $key
-     *
-     * @return bool|T|null
-     */
+    /** @return bool|T|null */
     public function remove(mixed $key): bool|object|null
     {
         return $this->doRemove($key, false);
@@ -588,11 +584,7 @@ trait PersistentCollectionTrait
         $this->doSet($offset, $value, true);
     }
 
-    /**
-     * @param mixed $offset
-     *
-     * @return void
-     */
+    /** @return void */
     #[ReturnTypeWillChange]
     public function offsetUnset(mixed $offset)
     {
@@ -656,7 +648,6 @@ trait PersistentCollectionTrait
      * Actual logic for adding an element to the collection.
      *
      * @param mixed $value
-     * @param bool  $arrayAccess
      *
      * @return true
      */
@@ -682,8 +673,6 @@ trait PersistentCollectionTrait
 
     /**
      * Actual logic for removing element by its key.
-     *
-     * @param mixed $offset
      *
      * @return bool|T|null
      * @phpstan-return (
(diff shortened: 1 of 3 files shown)


src/PersistentCollection/PersistentCollectionTrait.php
```diff
@@ -626,11 +648,10 @@ public function __clone()
      * Actual logic for adding an element to the collection.
      *
      * @param mixed $value
-     * @param bool  $arrayAccess
      *
      * @return true
      */
-    private function doAdd($value, $arrayAccess)
+    private function doAdd($value, bool $arrayAccess): true
```
IonBazan: You can use native `mixed` type

src/PersistentCollection/PersistentCollectionTrait.php
```diff
@@ -405,17 +423,17 @@ public function count()
...
     }
 
-    public function set($key, $value)
+    public function set($key, $value): void
     {
         $this->doSet($key, $value, false);
     }
 
-    public function add($element)
+    public function add($element): bool
```
IonBazan:
```suggestion
    public function add(mixed $element): bool
```

src/PersistentCollection/PersistentCollectionTrait.php
```diff
@@ -405,17 +423,17 @@ public function count()
         return $this->coll->count();
     }
 
-    public function set($key, $value)
+    public function set($key, $value): void
```
IonBazan: Let's see if we can add types to the key and value here too

src/PersistentCollection/PersistentCollectionTrait.php
```diff
@@ -563,36 +584,37 @@ public function offsetSet($offset, $value)
...
 
-    /**
-     * @param mixed $offset
-     *
-     * @return void
-     */
+    /** @return void */
     #[ReturnTypeWillChange]
-    public function offsetUnset($offset)
+    public function offsetUnset(mixed $offset)
```
IonBazan:
```suggestion
    public function offsetUnset(mixed $offset): void
```

This also allows to remove the PHP doc and `#[ReturnTypeWillChange]`

IonBazan (changes requested):
Please also check remaining methods in that file and add the types there

src/PersistentCollection/PersistentCollectionTrait.php
```diff
@@ -340,55 +354,59 @@ public function removeElement($element)
...
-    public function exists(Closure $p)
+    public function exists(Closure $p): bool
     {
         $this->initialize();
 
         return $this->coll->exists($p);
     }
 
-    /**
-     * @phpstan-return (TMaybeContained is T ? TKey|false : false)
```
IonBazan, line 366: I think should stay

GromNaN (approved):

IonBazan (approved):

IonBazan:
Thank you @krzysztof-ciszewski!

PR #2945, opened 2025-12-01
[3.x] Make Types strongly typed

IonBazan:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | -

#### Summary

Add types to `Types` namespace and some leftovers in other places.

### DIFF
--- a/benchmark/BaseBench.php
+++ b/benchmark/BaseBench.php
@@ -41,6 +41,7 @@ abstract class BaseBench
         $config->setDefaultDB(self::DATABASE_NAME);
         $config->setMetadataDriverImpl(self::createMetadataDriverImpl());
         $config->setMetadataCache(new ArrayAdapter());
+        $config->setUseNativeLazyObject(true);
 
         $client = new Client(
             getenv('DOCTRINE_MONGODB_SERVER') ?: self::DEFAULT_MONGODB_SERVER,
--- a/benchmark/BaseBench.php
+++ b/benchmark/BaseBench.php
@@ -9,7 +9,6 @@ use Doctrine\ODM\MongoDB\DocumentManager;
 use Doctrine\ODM\MongoDB\Mapping\Driver\AttributeDriver;
 use MongoDB\Client;
 use MongoDB\Model\DatabaseInfo;
-use PhpBench\Benchmark\Metadata\Annotations\BeforeMethods;
 use Symfony\Component\Cache\Adapter\ArrayAdapter;
 
 use function array_map;
@@ -17,14 +16,12 @@ use function getenv;
 use function in_array;
 use function iterator_to_array;
 
-/** @BeforeMethods({"initDocumentManager", "clearDatabase"}) */
 abstract class BaseBench
 {
-    public const DATABASE_NAME           = 'doctrine_odm_performance';
-    private const DEFAULT_MONGODB_SERVER = 'mongodb://localhost:27017';
+    public const string DATABASE_NAME           = 'doctrine_odm_performance';
+    private const string DEFAULT_MONGODB_SERVER = 'mongodb://localhost:27017';
 
-    /** @var DocumentManager */
-    protected static $documentManager;
+    protected static DocumentManager $documentManager;
 
     protected function getDocumentManager(): DocumentManager
     {
--- a/benchmark/Document/HydrateDocumentBench.php
+++ b/benchmark/Document/HydrateDocumentBench.php
@@ -9,29 +9,28 @@ use Doctrine\ODM\MongoDB\Hydrator\HydratorInterface;
 use Documents\User;
 use MongoDB\BSON\ObjectId;
 use MongoDB\BSON\UTCDateTime;
-use PhpBench\Benchmark\Metadata\Annotations\BeforeMethods;
-use PhpBench\Benchmark\Metadata\Annotations\Warmup;
+use PhpBench\Attributes\BeforeMethods;
+use PhpBench\Attributes\Warmup;
 
-/** @BeforeMethods({"init"}, extend=true) */
+#[BeforeMethods(['initDocumentManager', 'clearDatabase', 'init'])]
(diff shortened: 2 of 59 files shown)


src/Types/IntType.php
```diff
@@ -11,12 +11,12 @@
...
 class IntType extends Type implements Incrementable, Versionable
 {
-    public function convertToDatabaseValue($value)
+    public function convertToDatabaseValue(mixed $value): mixed
     {
         return $value !== null ? (int) $value : null;
     }
 
-    public function convertToPHPValue($value)
+    public function convertToPHPValue(mixed $value): mixed
```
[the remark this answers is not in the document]
IonBazan: I decided to keep this `mixed` as someone might extend this type and use `string` for larger numbers. I'm open to suggestions here.

src/Types/FloatType.php
```diff
@@ -29,7 +29,7 @@ public function closureToPHP(): string
         return '$return = (float) $value;';
     }
 
-    public function diff($old, $new)
+    public function diff(mixed $old, mixed $new): mixed
```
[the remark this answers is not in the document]
IonBazan: For this and above, I believe it might be necessary to cast them to `float` or `int` accordingly first.

src/Types/BooleanType.php
```diff
@@ -9,12 +9,12 @@
  */
 class BooleanType extends Type
 {
-    public function convertToDatabaseValue($value)
+    public function convertToDatabaseValue(mixed $value): ?bool
```
GromNaN, line 12: All the types must be declared using phpdoc in 2.x before being set.

benchmark/BaseBench.php
```diff
@@ -9,22 +9,19 @@
...
 abstract class BaseBench
 {
-    public const DATABASE_NAME           = 'doctrine_odm_performance';
-    private const DEFAULT_MONGODB_SERVER = 'mongodb://localhost:27017';
+    public const string DATABASE_NAME           = 'doctrine_odm_performance';
+    private const string DEFAULT_MONGODB_SERVER = 'mongodb://localhost:27017';
 
-    /** @var DocumentManager */
-    protected static $documentManager;
+    protected static DocumentManager $documentManager;
```
GromNaN, line 24: The benchmark files are not part of the package. This changes should be done in 2.16.

benchmark/Document/HydrateDocumentBench.php
```diff
@@ -9,29 +9,28 @@
 use Documents\User;
 use MongoDB\BSON\ObjectId;
 use MongoDB\BSON\UTCDateTime;
-use PhpBench\Benchmark\Metadata\Annotations\BeforeMethods;
-use PhpBench\Benchmark\Metadata\Annotations\Warmup;
+use PhpBench\Attributes\BeforeMethods;
+use PhpBench\Attributes\Warmup;
 
-/** @BeforeMethods({"init"}, extend=true) */
+#[BeforeMethods(['initDocumentManager', 'clearDatabase', 'init'])]
```
GromNaN, line 15: I did this change in https://github.com/doctrine/mongodb-odm/pull/2921, waiting for upmerge.

src/Types/IntType.php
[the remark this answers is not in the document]
GromNaN: What would be the point of extending this class, especially to make such a profound change that would involve rewriting all the methods?
I don't think we have `instanceof IntType` anywhere.

Eventually, the could use composition, but inheritance is not useful for that.

Please add the `?int` type.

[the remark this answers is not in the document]
IonBazan: Currently only `Int64Type` extends `IntType` and overrides the `convertToDatabaseValue()` return type to `Int64|null`. I could imagine someone could do something similar but indeed, creating a new type instead should be the right way to do. For this particular case we would need to stop extending this type and copy all missing methods to keep things consistent.

src/Types/BooleanType.php
IonBazan: Done in #2950

src/Types/IntType.php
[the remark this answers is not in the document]
IonBazan: I decided to separate these types - it resolves all above comments.

src/Types/Int64Type.php
```diff
@@ -21,8 +22,28 @@ public function convertToDatabaseValue($value)
...
     }
+
+    public function closureToPHP(): string
+    {
+        return '$return = (int) $value;';
+    }
+
+    public function diff(mixed $old, mixed $new): int
+    {
+        return (int) $new - $old;
```
GromNaN: The `(int)` cast has priority over the `-` operation. This should be unit tested using floats as input values.
```suggestion
        return (int) ($new - $old);
```

src/Types/IntType.php
```diff
@@ -33,14 +31,12 @@ public function closureToPHP(): string
         return '$return = (int) $value;';
     }
 
-    /** @return int */
-    public function diff($old, $new)
+    public function diff(mixed $old, mixed $new): int
     {
-        return $new - $old;
+        return (int) $new - $old;
```
GromNaN:
```suggestion
        return (int) ($new - $old);
```

tests/Tests/Aggregation/Stage/GeoNearTest.php
```diff
@@ -35,7 +35,7 @@ public function testFromBuilder(): void
             ->field('someField')
             ->equals('someValue');
 
-        $stage = ['near' => [0, 0], 'spherical' => false, 'distanceField' => 'distance', 'query' => ['someField' => 'someValue']];
+        $stage = ['near' => [0.0, 0.0], 'spherical' => false, 'distanceField' => 'distance', 'query' => ['someField' => 'someValue']];
```
GromNaN, line 38: Noted we convert the value to `float` with the arg type.

GromNaN (approved):

PR #2948, opened 2025-12-01
Add types for benchmarks

IonBazan:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no (internal classes only)
| Fixed issues | https://github.com/doctrine/mongodb-odm/pull/2945#discussion_r2575994038

#### Summary

Add missing types in Benchmarks and use `#[BeforeMethod]` attribute consistently.

>Important note here: I had to move #[BeforeMethods] only to the child classes as it appears that using PHP Attributes changes the order of execution of these methods when tagging both base and child classes - it previously executed parent first, now executes child first which breaks these benchmarks.

### DIFF
--- a/.github/workflows/performance.yml
+++ b/.github/workflows/performance.yml
@@ -16,6 +16,7 @@ jobs:
       matrix:
         php-version:
           - "8.1"
+          - "8.4"
 
     services:
       mongodb:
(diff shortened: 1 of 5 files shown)


GromNaN (approved):
Thank you!

IonBazan:
@GromNaN important note here - looks like we are only running Performance Tests against PHP 8.1 so they are not using native lazy objects.

GromNaN:
> looks like we are only running Performance Tests against PHP 8.1 so they are not using native lazy objects.

Good point, we can switch to CI to 7.4.

PR #2949, opened 2025-12-01
Fix performance tests artifact upload

IonBazan:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | yes/no
| Fixed issues | -

#### Summary

Fixes failing build after #2948 has been merged due to conflicting artifact name.

### DIFF
--- a/.github/workflows/performance.yml
+++ b/.github/workflows/performance.yml
@@ -59,7 +59,7 @@ jobs:
       - name: "Upload composer.lock as build artifact"
         uses: actions/upload-artifact@v5
         with:
-          name: composer.lock
+          name: composer.lock-${{ matrix.php-version }}
           path: composer.lock
 
       - name: "Run PHPBench"


.github/workflows/performance.yml
```diff
@@ -59,7 +59,7 @@ jobs:
       - name: "Upload composer.lock as build artifact"
         uses: actions/upload-artifact@v5
         with:
-          name: composer.lock
+          name: composer.lock-${{ matrix.php-version }}
           path: composer.lock
```
GromNaN, line 63: This cache block is not necessary, [ramsey/composer-install](https://github.com/ramsey/composer-install) already enables the cache for us.

IonBazan: I think this is not used as cache but as a reference in case you want to download the `composer.lock` and try it locally with exact same dependencies.

GromNaN (approved):

PR #2956, opened 2025-12-05
[3.x] Drop deprecated features

IonBazan:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | -

#### Summary

Removes some features marked as deprecated:
- `fsync`, `safe` and `w` write concerns
- `indexes` attribute property
- `LegacyReflectionFields` support
- `NOTIFY` change tracking policy support
- Combining `targetDocument` with `discriminatorMap` will now trigger exception.
- Passing an array to `setDiscriminatorField`


src/Configuration.php
```diff
@@ -156,55 +152,6 @@ public function getClientEncryptionOptions(): array
         ]);
     }
 
-    /**
-     * Adds a namespace under a certain alias.
-     */
-    public function addDocumentNamespace(string $alias, string $namespace): void
```
GromNaN, line 162: This methods must be deprecated in 2.x.

IonBazan: Fixing it in #2957

tests/Tests/Mapping/AbstractMappingDriverTestCase.php
```diff
@@ -869,7 +869,6 @@ public static function loadMetadata(ClassMetadata $metadata): void
         $metadata->addLifecycleCallback('doStuffOnPrePersist', 'prePersist');
         $metadata->addLifecycleCallback('doOtherStuffOnPrePersistToo', 'prePersist');
         $metadata->addLifecycleCallback('doStuffOnPostPersist', 'postPersist');
-        $metadata->setDiscriminatorField(['fieldName' => 'discr']);
```
IonBazan, line 872: I found that this `loadMetadata()` method is not even called - all tests were passing without `TypeError` thrown, when I left it like that.

Perhaps we can remove this method completely.

GromNaN: Yes, it can be removed here and from the 2.x branch.

GromNaN: It looks like a aborted attempt to use StaticPHPDriver https://github.com/doctrine/persistence/blob/f7deb2207f44cd2cd30a59bec3d2451738848af6/docs/en/reference/index.rst#staticphpdriver

IonBazan: I think we can leave it for now and try to cover it with a different test.

src/Mapping/Annotations/EmbeddedDocument.php
```diff
@@ -12,12 +12,4 @@
 #[Attribute(Attribute::TARGET_CLASS)]
 final class EmbeddedDocument extends AbstractDocument
 {
-    /** @var Index[] */
-    public $indexes;
```
GromNaN, line 16: This was already removed for the `Document` class in https://github.com/doctrine/mongodb-odm/commit/a3327808d2aa5c17b76f62f63a635a7c4dbc10a2; and the `AttributeDriver` was triggering a deprecation that you also remove in this PR.

GromNaN (approved):

GromNaN:
Thank you @IonBazan

PR #2957, opened 2025-12-05
Deprecate remaining namespace alias methods

IonBazan:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

Adds deprecation notices for `addDocumentNamespace()` and `getDocumentNamespaces()` as this functionality is deprecated since 2.3. See https://github.com/doctrine/mongodb-odm/pull/2956#discussion_r2592707528

### DIFF
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -198,6 +198,8 @@ class Configuration
 
     /**
      * Adds a namespace under a certain alias.
+     *
+     * @deprecated Document short aliases are deprecated - use ::class constant instead.
      */
     public function addDocumentNamespace(string $alias, string $namespace): void
     {
@@ -227,6 +229,8 @@ class Configuration
 
     /**
      * Retrieves the list of registered document namespace aliases.
+     *
+     * @deprecated Document short aliases are deprecated - use ::class constant instead.
      *
      * @return array<string, string>
      */
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -203,11 +203,20 @@ class Configuration
      */
     public function addDocumentNamespace(string $alias, string $namespace): void
     {
+        trigger_deprecation(
+            'doctrine/mongodb-odm',
+            '2.3',
+            'Document short namespace aliases such as "%s" are deprecated, use ::class constant instead.',
+            $alias,
+        );
+
         $this->attributes['documentNamespaces'][$alias] = $namespace;
     }
 
     /**
      * Resolves a registered namespace alias to the full namespace.
+     *
+     * @deprecated Document short aliases are deprecated - use ::class constant instead.
      *
      * @throws MongoDBException
      */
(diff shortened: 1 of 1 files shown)


src/Configuration.php
alcaeus, line 223: This method should be deprecated as well.

src/Configuration.php
```diff
@@ -198,6 +198,8 @@ public function getClientEncryptionOptions(): array
 
     /**
      * Adds a namespace under a certain alias.
+     *
+     * @deprecated Document short aliases are deprecated - use ::class constant instead.
      */
     public function addDocumentNamespace(string $alias, string $namespace): void
```
alcaeus, line 204: This method and `getDocumentNamespaces` should emit a deprecation notice using `trigger_deprecation`

alcaeus (dismissed):

src/Configuration.php
GromNaN, line 259:
```suggestion
     *
     * @deprecated Document short aliases are deprecated - use ::class constant instead.
     *
```

GromNaN, line 218:
```suggestion
     *
     * @deprecated Document short aliases are deprecated - use ::class constant instead.
     *
```

GromNaN (commented):
I think the are 2 other methods to mark as deprecated.

src/Configuration.php
IonBazan: Should I use `2.3` as intended deprecation version?

GromNaN (approved):

src/Configuration.php
GromNaN: We don't add deprecations in bugfix versions. 2.16 is the correct branch.

IonBazan: This functionality was deprecated in 2.3 but these methods were omitted which was a bug. If one was using it they would already have deprecations visible when calling other methods related to aliases.

khaperets:
@GromNaN After this pull request, I keep getting deprecation warnings in the project all the time. How can this be fixed?

```yaml
doctrine_mongodb:
    auto_generate_proxy_classes: true
    auto_generate_hydrator_classes: true
    connections:
        default:
            server: '%env(resolve:MONGODB_URI)%'
            options: { }
    default_database: '%env(resolve:MONGODB_DB)%'
    document_managers:
        default:
            auto_mapping: true
            mappings:
                App:
                    dir: '%kernel.project_dir%/src/Document'
                    prefix: 'App\Document'
```

IonBazan:
Thanks for reporting @khaperets! It seems that it's caused by the bundle code here: https://github.com/doctrine/DoctrineMongoDBBundle/blob/9c2ccf08a08fcf739d346810adbbd322ed9c0781/src/DependencyInjection/DoctrineMongoDBExtension.php#L975-L997. @GromNaN Do you think we need to revert this change?

GromNaN:
We need to add an option in the bundle configuration to disable namespace aliases. Enable this option by default in the Symfony recipe. And deprecate not not enabling the option. In the next major version we change the default behavior and throw an error if the option is set and disabled. And we deprecate setting the option.

PR #2960, opened 2025-12-05
Deprecate not overriding `Type::closureToPHP()`

### LINKED ISSUE #2870: Evaluate the performance benefit of `Type::closureToPHP` and `closureToMongo`
Both methods exist since the beginning of the project. Only `closureToPHP` is actually used to generate code in `HydratorFactory` (documented by https://github.com/doctrine/mongodb-odm/pull/2855).

They add complexity when implementing a new type (see https://github.com/doctrine/mongodb-odm/pull/2868), so types use the [`ClosureToPHP` trait](https://github.com/doctrine/mongodb-odm/blob/2.14.x/src/Types/ClosureToPHP.php).

Task:
- benchmark performance of removing this methods and always call the type class
- find refactoring optimization
- maybe deprecate this methods

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | Fix #2870

#### Summary

The default implementation of `Type::closureToPhp()` is incorrect as it forwards the unmodified value.
https://github.com/doctrine/mongodb-odm/blob/bf2e243204634fe0892794347554fc4b240c9148/src/Types/Type.php#L142-L145

In 3.0, I want to change the default implementation to use the one from the `ClosureToPHP` trait that calls the type class.
https://github.com/doctrine/mongodb-odm/blob/bf2e243204634fe0892794347554fc4b240c9148/src/Types/ClosureToPHP.php#L14-L16

Also adding the missing implementation for `collection` and `hash`.

Any type that does not override the `closureToPhp` will get a deprecation notice.

### DIFF
--- a/src/Types/CollectionType.php
+++ b/src/Types/CollectionType.php
@@ -29,4 +29,9 @@ class CollectionType extends Type
     {
         return $value !== null ? array_values($value) : null;
     }
+
+    public function closureToPHP(): string
+    {
+        return '$return = array_values($value);';
+    }
 }
--- a/src/Types/HashType.php
+++ b/src/Types/HashType.php
@@ -28,4 +28,9 @@ class HashType extends Type
     {
         return $value !== null ? (array) $value : null;
     }
+
+    public function closureToPHP(): string
+    {
+        return '$return = (array) $value;';
+    }
 }
--- a/src/Types/Type.php
+++ b/src/Types/Type.php
@@ -15,6 +15,7 @@ use function explode;
 use function gettype;
 use function is_object;
 use function str_replace;
+use function trigger_deprecation;
 
 /**
  * The Type interface.
@@ -129,18 +130,26 @@ abstract class Type
     /**
      * Get the PHP code equivalent to {@see convertToDatabaseValue()}, used in code generator.
      * Use variables $value for input and $return for output.
+     *
+     * @deprecated Since 2.16, will be removed in 3.0.
      */
     public function closureToMongo(): string
     {
+        trigger_deprecation('doctrine/mongodb-odm', '2.16', 'Type::closureToMongo() is deprecated and will be removed in 3.0.');
+
         return '$return = $value;';
     }
 
     /**
      * Get the PHP code equivalent to {@see convertToPHPValue()}, used in code generator.
      * Use variables $value for input and $return for output.
+     *
+     * @abstract The default implementation will change in 3.0.
      */
     public function closureToPHP(): string
     {
+        trigger_deprecation('doctrine/mongodb-odm', '2.16', 'The method Type::closureToPHP() will change its default implementation in 3.0 to use convertToPHPValue(). Override this method if you need custom behavior before upgrading to 3.0 or use the trait ClosureToPHP to get the upcoming behavior now.');
+
         return '$return = $value;';
     }
 
--- a/tests/Tests/Functional/CustomTypeTest.php
+++ b/tests/Tests/Functional/CustomTypeTest.php
@@ -7,11 +7,11 @@ namespace Doctrine\ODM\MongoDB\Tests\Functional;
 use DateTime;
 use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
 use Doctrine\ODM\MongoDB\Tests\BaseTestCase;
+use Doctrine\ODM\MongoDB\Tests\CaptureDeprecationMessages;
 use Doctrine\ODM\MongoDB\Types\ClosureToPHP;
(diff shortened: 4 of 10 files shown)


src/Types/CollectionType.php
```diff
@@ -29,4 +29,9 @@ public function convertToPHPValue($value)
     {
         return $value !== null ? array_values($value) : null;
     }
+
+    public function closureToPHP(): string
+    {
+        return '$return = array_values($value);';
```
GromNaN, line 35: `null` is already handled by the generated hydrator code.

src/Types/HashType.php
```diff
@@ -28,4 +28,9 @@ public function convertToPHPValue($value)
     {
         return $value !== null ? (array) $value : null;
     }
+
+    public function closureToPHP(): string
+    {
+        return '$return = (array) $value;';
```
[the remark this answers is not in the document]
GromNaN: Null value is already handled in the generated hydration code.

src/Types/Type.php
```diff
@@ -129,18 +130,27 @@ public function convertToPHPValue($value)
     /**
      * Get the PHP code equivalent to {@see convertToDatabaseValue()}, used in code generator.
      * Use variables $value for input and $return for output.
+     *
+     * @deprecated Since 2.16, will be removed in 3.0.
      */
     public function closureToMongo(): string
     {
+        trigger_deprecation('doctrine/mongodb-odm', '2.16', 'Type::closureToMongo() is deprecated and will be removed in 3.0.');
```
alcaeus, line 138: Is this method currently used in the any of our code? If so, we shouldn't deprecate it just yet as this will lead to a number of deprecation notices that people can't fix.

src/Types/Type.php
```diff
@@ -129,18 +130,27 @@ public function convertToPHPValue($value)
...
 
     /**
      * Get the PHP code equivalent to {@see convertToPHPValue()}, used in code generator.
      * Use variables $value for input and $return for output.
+     *
+     * @abstract The default implementation will change in 3.0.
      */
     public function closureToPHP(): string
     {
+        trigger_deprecation('doctrine/mongodb-odm', '2.16', 'The method Type::closureToPHP() will change its default implementation in 3.0 to use convertToPHPValue(). Override this method if you need custom behavior before upgrading to 3.0 or use the trait ClosureToPHP to get the upcoming behavior now.');
```
alcaeus, line 151: Same question as before regarding how often this message will come up.

src/Types/Type.php
```diff
@@ -129,18 +130,27 @@ public function convertToPHPValue($value)
...
+        // return sprintf('$return = \%s::getType($typeIdentifier)->convertToPHPValue($value);', Type::class);
```
alcaeus: Commented code can be removed, right?

src/Types/Type.php
GromNaN: No, it's not used anywhere.

src/Types/Type.php
GromNaN: I've updated the child classes so that this method is always overridden. This deprecation is triggered only on custom type classes that would not override this method. They have to use the `ClosureToPHP` trait to fix the deprecation.

src/Types/Type.php
GromNaN: That's the code for 3.0.

alcaeus: Then we should add it to 3.0 then, no point adding commented code to the repo.

GromNaN: Ok, removed.

tests/Tests/Functional/CustomTypeTest.php
```diff
@@ -89,6 +91,14 @@ public function testTypeFromPHPVariable(): void
...
         self::assertSame(['name' => 'French', 'code' => 'fr'], $databaseValue);
     }
+
+    #[IgnoreDeprecations]
+    public function testNotOverridingClosureToPHPIsDeprecated(): void
+    {
+        $type = Type::getType('custom_type_without_closure_to_php');
+
+        self::assertSame('$return = $value;', $type->closureToPHP());
+    }
```
SenseException: Is this test covering that the behavior is still given even without using the new Trait? Because the name of the method tells something different and that it should check for the deprecation, that get ignored.

GromNaN: Indeed, the assertion on deprecation was missing because it doesn't work correctly with PHPUnit 11, but thanks to the `CaptureDeprecationMessages` trait, I was able to add this assertion.

alcaeus (approved):

PR #2961, opened 2025-12-06
[3.x] Use PHPUnit 12

IonBazan:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | -

#### Summary

Updates tests to use PHPUnit 12 features and avoids deprecations:
- Creating a Mock with no expectations (Stub should be used)
- `isType('array')` -> `isArray()`
- Providing extra arguments to a test via data provider

### DIFF
--- a/tests/Tests/Query/BuilderTest.php
+++ b/tests/Tests/Query/BuilderTest.php
@@ -497,8 +497,8 @@ class BuilderTest extends BaseTestCase
             'type()' => ['type', [7]],
             'all()' => ['all', [['value1', 'value2']]],
             'mod()' => ['mod', [2, 0]],
-            'near()' => ['near', [1, 2], null, 5, 10],
-            'nearSphere()' => ['nearSphere', [1, 2], null, 5, 10],
+            'near()' => ['near', [[1, 2], null, 5, 10]],
+            'nearSphere()' => ['nearSphere', [[1, 2], null, 5, 10]],
             'geoIntersects()' => ['geoIntersects', [self::createGeometry()]],
             'geoWithin()' => ['geoWithin', [self::createGeometry()]],
             'geoWithinBox()' => ['geoWithinBox', [1, 2, 3, 4]],
@@ -864,8 +864,7 @@ class BuilderTest extends BaseTestCase
         return new Builder($this->dm, User::class);
     }
 
-    /** @return MockObject&Expr */
-    private function getMockExpr()
+    private function getMockExpr(): Expr&MockObject
     {
         return $this->createMock(Expr::class);
     }
--- a/tests/Tests/QueryTest.php
+++ b/tests/Tests/QueryTest.php
@@ -22,16 +22,15 @@ use LogicException;
 use MongoDB\BSON\Int64;
 use MongoDB\BSON\ObjectId;
 use MongoDB\Collection;
-use MongoDB\Driver\CursorId;
 use MongoDB\Driver\CursorInterface;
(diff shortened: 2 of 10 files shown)


tests/Tests/Query/BuilderTest.php
```diff
@@ -497,8 +497,8 @@ public static function provideProxiedExprMethods(): array
             'type()' => ['type', [7]],
             'all()' => ['all', [['value1', 'value2']]],
             'mod()' => ['mod', [2, 0]],
-            'near()' => ['near', [1, 2], null, 5, 10],
-            'nearSphere()' => ['nearSphere', [1, 2], null, 5, 10],
+            'near()' => ['near', [[1, 2], null, 5, 10]],
+            'nearSphere()' => ['nearSphere', [[1, 2], null, 5, 10]],
```
IonBazan, line 501: This caused extra arguments to be provided to the test. It only accepts 2 arguments and the second one must be array. I believe it was causing unexpected results.

tests/Tests/QueryTest.php
```diff
@@ -620,20 +619,14 @@ public function key(): int
...
+    private function getMockCollection(): Collection&MockObject
     {
         return $this->createMock(Collection::class);
     }
 
-    private function createCursorMock(): CursorInterface|Traversable
+    private function createCursorStub(): CursorInterface&Stub
     {
-        return $this->createMock(
-            // Use the cursorID class to differentiate between 1.x and 2.x
```
IonBazan, line 632: Redundant v1 check has been removed.

tests/Tests/SchemaManagerTest.php
```diff
@@ -1407,30 +1408,27 @@ private function getDatabaseName(ClassMetadata $cm): string
...
         return $mock;
     }
 
-    /** @return Collection&MockObject */
-    private function getMockCollection(?string $name = null)
+    private function getMockCollection(?string $name = null): Collection&MockObject
     {
         $collection = $this->createMock(Collection::class);
-        $collection->method('getCollectionName')->willReturnCallback(static fn () => $name);
+        $collection->expects($this->atLeast(0))->method('getCollectionName')->willReturnCallback(static fn () => $name);
```
IonBazan, line 1423: This might look weird but since we reuse this method in many tests and only in some cases we are adding expectations - therefore we can't use Stub here and adding `$this->any()` will still cause a deprecation error. Using `atLeast(0)` gets rid of that warning.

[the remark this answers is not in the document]
IonBazan: Explained below, thanks!

tests/Tests/Query/BuilderTest.php
GromNaN: Can you backport this fix?

tests/Tests/QueryTest.php
GromNaN: We can drop support for v1 in 3.0

GromNaN (approved):

tests/Tests/QueryTest.php
IonBazan: It's already dropped 😃

GromNaN: Oh yes https://github.com/doctrine/mongodb-odm/commit/b102f85b2374d06254d3f279c19d39fa22b083ca

GromNaN:
Thank you @IonBazan

PR #2962, opened 2025-12-07
Replace __sleep/__wakeup with __serialize/__unserialize

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | -

#### Summary

The magic methods `__sleep` and `__wakeup` are being soft-deprecated ([RFC](https://wiki.php.net/rfc/soft-deprecate-sleep-wakeup)). Replaced by `__serialize` and `__unserialize`. This new methods are more verbose.

Similar to https://github.com/symfony/symfony/pull/61727

### DIFF
--- a/src/Mapping/ClassMetadata.php
+++ b/src/Mapping/ClassMetadata.php
@@ -30,7 +30,6 @@ use MongoDB\BSON\Int64;
 use MongoDB\BSON\UTCDateTime;
 use ReflectionClass;
 use ReflectionEnum;
-use ReflectionMethod;
 use ReflectionNamedType;
 use Symfony\Component\Uid\UuidV1;
 use Symfony\Component\Uid\UuidV4;
@@ -57,7 +56,6 @@ use function ltrim;
 use function sprintf;
 use function strtolower;
 use function strtoupper;
-use function trigger_deprecation;
 
 /**
  * A <tt>ClassMetadata</tt> instance holds all the object-document mapping metadata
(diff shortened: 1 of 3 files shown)


src/Mapping/ClassMetadata.php
```diff
@@ -2538,111 +2538,115 @@ public function mapField(array $mapping): array
      *      - reflFields (ReflectionProperty array)
      *      - propertyAccessors (ReflectionProperty array)
      *
-     * @return array The names of all the fields that should be serialized.
-     */
-    public function __sleep(): array
```
GromNaN, line 2491: This methods should be marked as `@internal` or `@deprecated` in 2.x? They should never be called directly or overridden.

IonBazan (approved):

src/Mapping/ClassMetadata.php
IonBazan: `ClassMetadata` is already marked as `@final` but we should also mark these methods as  `@deprecated` to avoid direct calling them.

alcaeus: Generally, I'd argue that the magic methods are not meant to be called directly, and thus shouldn't need `@internal` or `@deprecated`. Instead, these methods define behaviour of the class which in this case doesn't change -- we're just changing the points where we hook in.
As for a potential compatibility layer in v2, we may want to rely on deprecation notices in certain places to highlight upcoming changes.

src/Tools/Console/Command/GeneratePersistentCollectionsCommand.php
```diff
@@ -103,6 +105,14 @@ protected function execute(InputInterface $input, OutputInterface $output): int
                         continue;
                     }
 
+                    if (method_exists($mapping['collectionClass'], '__sleep')) {
```
alcaeus: IMO, this should be in the generator class, as this command is typically not called in development environments. We also need to consider a compatibility layer in v2 to allow people to be notified of the upcoming change in v3 and implement `__serialize` instead.

GromNaN: I realize that this trait is added on generated classes that extends user defined classes. The `__sleep` method is alreadyoverridden by the trait one. So, not functioning. I'll remove this error.

src/Mapping/ClassMetadata.php
GromNaN: Adding a BC layer in https://github.com/doctrine/mongodb-odm/pull/2965

src/Mapping/ClassMetadata.php
```diff
@@ -2466,144 +2464,112 @@ public function mapField(array $mapping): array
...
+            'generatorOptions' => $this->generatorOptions,
+            'idGenerator' => $this->idGenerator,
+            'indexes' => $this->indexes,
+            'shardKey' => $this->shardKey,
+            'timeSeriesOptions' => $this->timeSeriesOptions,
+            'isEncrypted' => $this->isEncrypted,
         ];
 
         // The rest of the metadata is only serialized if necessary.
         if ($this->changeTrackingPolicy !== self::CHANGETRACKING_DEFERRED_IMPLICIT) {
```
alcaeus, line 2487: I'm wondering if it's worth keeping the conditionals around. Keeping a map of fields to serialise would also make the unserialisation more secure (see comment below).

src/Mapping/ClassMetadata.php
```diff
@@ -2466,144 +2464,112 @@ public function mapField(array $mapping): array
...
      *
-     * @param array<string, mixed> $data The serialized data.
+     * @param array<string, mixed> $data
      */
     public function __unserialize(array $data): void
     {
-        foreach ($data as $field => $value) {
-            $this->$field = $value;
+        foreach ($data as $property => $value) {
+            $this->$property = $value;
```
alcaeus, line 2577: Not sure how big an issue this still is, but this would technically allow people to override properties that we don't actually serialise or intend to read from serialised data. Paired with the comment above, having a map of all properties we want to serialise would allow us to re-use this map and only use data from fields that we know should be serialised.

src/Tools/Console/Command/GeneratePersistentCollectionsCommand.php
```diff
@@ -6,6 +6,7 @@
 
 use Doctrine\ODM\MongoDB\Tools\Console\MetadataFilter;
 use InvalidArgumentException;
+use LogicException;
```
alcaeus: This is left over from previous changes I assume...

alcaeus (changes requested):

src/Mapping/ClassMetadata.php
GromNaN: Unserialization is in the hotpath. I think it's an optimization to only serialize the data that is different from the default value.

GromNaN: As for overriding private properties that are not part of the metadata, that's a stretch. We can also use reflection.

alcaeus (approved):

src/Mapping/ClassMetadata.php
alcaeus: Sounds good -- since we're not making it any worse I'm fine leaving this as-is.

PR #2965, opened 2025-12-09
Deprecate overriding the `__sleep` method of `ClassMetadata` or `PersistentCollection`

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | 

#### Summary

BC layer for #2962

### DIFF
--- a/src/Mapping/ClassMetadata.php
+++ b/src/Mapping/ClassMetadata.php
@@ -34,6 +34,7 @@ use MongoDB\BSON\UTCDateTime;
 use ProxyManager\Proxy\GhostObjectInterface;
 use ReflectionClass;
 use ReflectionEnum;
+use ReflectionMethod;
 use ReflectionNamedType;
 use ReflectionProperty;
 use Symfony\Component\Uid\UuidV1;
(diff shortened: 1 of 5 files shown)


src/Mapping/ClassMetadata.php
```diff
@@ -2645,7 +2646,31 @@ public function mapField(array $mapping): array
      *      - reflFields (ReflectionProperty array)
      *      - propertyAccessors (ReflectionProperty array)
      *
-     * @return array The names of all the fields that should be serialized.
+     * @return array<string, mixed> The serialized data.
+     */
+    public function __serialize(): array
+    {
+        if (new ReflectionMethod($this, '__sleep')->getDeclaringClass() !== self::class) {
```
IonBazan: Can we add a test for this deprecation?

IonBazan (approved):
Do note that `ClassMetadata` is already soft-final.

src/Mapping/ClassMetadata.php
```diff
@@ -923,7 +924,7 @@ public static function getReferenceFieldName(string $storeAs, string $pathPrefix
             return $pathPrefix;
         }
 
-        return ($pathPrefix ? $pathPrefix . '.' : '') . static::getReferencePrefix($storeAs) . 'id';
+        return ($pathPrefix ? $pathPrefix . '.' : '') . self::getReferencePrefix($storeAs) . 'id';
```
GromNaN, line 927: PHPStan noticed that we can't call a private static method with `static::` unless the class is `final`.

src/Mapping/ClassMetadata.php
```diff
@@ -2645,7 +2646,31 @@ public function mapField(array $mapping): array
      *      - reflFields (ReflectionProperty array)
      *      - propertyAccessors (ReflectionProperty array)
      *
-     * @return array The names of all the fields that should be serialized.
+     * @return array<string, mixed> The serialized data.
+     */
+    public function __serialize(): array
+    {
+        if (static::class !== self::class && (new ReflectionMethod($this, '__sleep'))->getDeclaringClass() !== self::class) {
```
GromNaN, line 2653: Optimizing the happy path, since the class should not be extended.

src/Mapping/ClassMetadata.php
```diff
@@ -2665,8 +2690,10 @@ public function __sleep()
             'generatorOptions',
             'idGenerator',
             'indexes',
+            'searchIndexes',
             'shardKey',
             'timeSeriesOptions',
+            'isEncrypted',
```
GromNaN, line 2696: This properties have been added later and we I forgot to add them to the serialization. I'm not sure if that creates any bug, at least they would be fixed.

GromNaN (commented):
Thanks for the review @IonBazan. I added the test and fixed small unrelated issues.

PR #2966, opened 2025-12-10
Create a `TypeRegistry` and enable injection of `Type` instances

### LINKED ISSUE #2818: Extract the static methods of the `Type` class into a `TypeRegistry` service
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes
| BC Break    | yes

#### Summary

In order to have distinct types per DocumentManager instance, the static methods of the `Type` class must move to a `TypeRegistry` service class. This design is inspired by the [`Doctrine\DBAL\Types\TypeRegistry`](https://github.com/doctrine/dbal/blob/4.3.x/src/Types/TypeRegistry.php) class. 

- [ ] Introduce the `TypeRegistry`
- [ ] Inject the `TypeRegistry` instance everywhere the static method of the `Type` class are used
- [ ] Make the constructor of `Type` public and non-final to allow injecting dependencies or configuration into the type instances

A similar change has been discussed in https://github.com/doctrine/dbal/pull/6705#pullrequestreview-2599239196

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no (deprecation)
| Fixed issues | Closes #2818

Custom types are currently registered on a process-wide singleton through the static methods of `Type`. That makes them impossible to scope to a `DocumentManager`, and it makes types with their own dependencies awkward to build, since nothing can be injected into them.

This PR moves type resolution to an instance held by the `Configuration`, following the shape settled in [doctrine/dbal#7342](https://github.com/doctrine/dbal/pull/7342) so that users who know DBAL find the same concepts and the same names here.

The Bundle PR: https://github.com/doctrine/DoctrineMongoDBBundle/pull/963

## `TypeProvider`

The interface that `Configuration` accepts and returns. `TypeRegistry` is one implementation of it, not the only option.

```php
/** @extends Traversable<string, Type> */
interface TypeProvider extends Traversable
{
    public function get(string $name): Type;

    public function has(string $name): bool;
}
```

It extends `Traversable`, since callers may also want to enumerate the available types. `register()` is deliberately absent, so a provider stays read-only from the outside.

## `TypeRegistry`

`final`, implementing `TypeProvider` and `IteratorAggregate`. Same name and same methods as [DBAL's `TypeRegistry`](https://github.com/doctrine/dbal/blob/4.4.x/src/Types/TypeRegistry.php), except `override`, because `register` creates or replaces the type, "upsert" being a MongoDB specificity.

- Built-in types are pre-populated from a class constant, so construction is cheap. Types passed to the constructor, by instance or by class name, are layered on top and may override a built-in by name.
- Types can be lazy-loaded from a `ServiceProviderInterface` such as a Symfony `ServiceLocator`: each provided service ID acts as the type name, and a type is resolved only on first access.
  ```php
  new TypeRegistry($serviceLocator);
  ```
  This is what lets the bundle wire a service locator instead of building every custom type at boot.
- `getMap()` is replaced by `getIterator()`, a generator. Stopping early no longer instantiates every remaining type.
- `register()` no longer instantiates a class just to validate it. The constructor is inspected by reflection instead, so registering by class name stays lazy.
- Service-backed types are not memoized by the registry: each access delegates to the provider, so instance identity and reset behavior are governed by the container, not by the registry.

Unlike DBAL, there is no reverse index and no `lookupName()`, so the same instance may be registered under several names.

## `TypeGuesser`

`guessTypeFromValue()` and `convertToDatabaseValue()` are ODM specifics with no DBAL equivalent. They move out of the registry into a separate final class, which keeps `TypeProvider` minimal and implementable by third parties. The instance is created lazily and reused through `DocumentManager::getTypeGuesser()`:

```php
final class TypeGuesser
{
    public function __construct(private readonly TypeProvider $types) {}

    public function guessTypeFromValue(mixed $value): ?Type;
    public function convertToDatabaseValue(mixed $value): mixed;
}
```

## `Configuration`

```php
$config->setTypeProvider(new TypeRegistry([
    'date_with_timezone' => new DateTimeWithTimezoneType(),
]));
```

`getTypeProvider()` returns a `TypeProvider`, defaulting to the shared registry. The registry can only be set once: swapping it after `ClassMetadataFactory` has injected it into already loaded metadata would leave those instances pointing at the old one, so a second call throws a `LogicException`. I intend to propose the same guard for DBAL.

Types registered this way are scoped to the `DocumentManager` instances built from that `Configuration`, so they cannot affect the rest of the application.

## The static `Type::*` methods are deprecated

They all operate on the process-wide registry, which behaves unexpectedly as soon as a `Configuration` has its own provider: a type registered with `Type::addType()` is invisible to that `DocumentManager`. Nothing signals that today, so the mistake surfaces later as an apparently unrelated `InvalidTypeException`.

`getType()`, `hasType()`, `addType()`, `registerType()`, `overrideType()`, `getTypesMap()`, `getTypeFromPHPVariable()` and `convertPHPToDatabaseValue()` therefore carry an `@deprecated` docblock and a runtime deprecation. They keep working and still delegate to the shared registry, so nothing breaks.

## Trade-offs

**Iterating a provider resolves every type.** Laziness holds for the query paths, which resolve types by name.

**Custom types registered via `Type::addType()` are invisible to a `Configuration` that has its own registry.** This is intentional, it is the isolation the feature provides. Users who set a registry are responsible for registering all the types they need in it.

**The shared registry is preserved.** `Configuration` instances that never call `setTypeProvider()` continue to behave exactly as before.

## Notes

The `TypeRegistry` constructor accepts either an array of types or a `ServiceProviderInterface` (for example a Symfony `ServiceLocator`). `psr/container` is a `require-dev` plus a `suggest`, not a hard requirement: the interface used here is Symfony's `ServiceProviderInterface`, already available through `symfony/service-contracts`.

Along the way this fixes three tests that were already failing on the branch, left stale by earlier commits: the constructor rule relaxed in fc4d6e5 was never reflected in `CustomTypeTest`, and two deprecation messages still said 2.16.


src/DocumentManager.php
```diff
@@ -205,6 +210,21 @@ public function resolveClassName(string $className): string
...
+    private function initTypeRegistry(): void
+    {
+        $prop     = new ReflectionProperty(Type::class, 'registry');
+        $registry = $prop->getValue();
+
+        $this->typeRegistry ??= $registry ?? new TypeRegistry();
+        $registry           ??= $this->typeRegistry;
+
+        if ($registry !== $this->typeRegistry) {
+            throw new LogicException('Type registry is already set and cannot be modified. A single TypeRegistry instance is allowed, this will change in MongoDB ODM 3.0.');
```
GromNaN: There is a design flow with Symfony `cache:clear` because the container is created twice in the same process. So we have 2 distinct `TypeRegisty` that are created.

This static variable is set to maintain BC for the static methods of the `Type` class.

docs/en/reference/custom-mapping-types.rst
```diff
@@ -79,24 +79,28 @@ Restrictions to keep in mind:
...
+    $types->register('date_with_timezone', \My\Project\Types\DateTimeWithTimezoneType::class);
 
     // Overrides a type. This results in an exception if type with given name is not registered
-    Type::overrideType('date_immutable', \My\Project\Types\DateTimeWithTimezoneType::class);
+    $types->register('date_immutable', \My\Project\Types\DateTimeWithTimezoneType::class);
 
-    // Registers a type without checking whether it was already registered
-    Type::registerType('date_immutable', \My\Project\Types\DateTimeWithTimezoneType::class);
+    // Initialize DocumentManager with the TypeRegistry
+    $dm = DocumentManager::create($client, $config, null, $types);
```
alcaeus: Nit: I'd suggest a different variable name (`$typeRegistry`)

```suggestion
    $typeRegistry = new TypeRegistry();

   // Adds a type. This results in an exception if type with given name is already registered
    $typeRegistry->register('date_with_timezone', \My\Project\Types\DateTimeWithTimezoneType::class);

   // Overrides a type. This results in an exception if type with given name is not registered
    $typeRegistry->register('date_immutable', \My\Project\Types\DateTimeWithTimezoneType::class);

    // Initialize DocumentManager with the TypeRegistry
    $dm = DocumentManager::create($client, $config, typeRegistry: $typeRegistry);
```

docs/en/reference/custom-mapping-types.rst
```diff
@@ -182,7 +186,8 @@ Register the type in your bootstrap code::
 
 .. code-block:: php
 
-    Type::addType(Money::class, App\MongoDB\Types\MoneyType::class);
+    $types = new TypeRegistry();
+    $types->register(Money::class, App\MongoDB\Types\MoneyType::class);
```
alcaeus:
```suggestion
    $typeRegistry = new TypeRegistry();
    $typeRegistry->register(Money::class, App\MongoDB\Types\MoneyType::class);
```

src/DocumentManager.php
```diff
@@ -144,7 +145,7 @@ class DocumentManager implements ObjectManager
      * Creates a new Document that operates on the given Mongo connection
      * and uses the given Configuration.
      */
-    protected function __construct(?Client $client = null, ?Configuration $config = null, ?EventManager $eventManager = null)
+    protected function __construct(?Client $client = null, ?Configuration $config = null, ?EventManager $eventManager = null, private ?TypeRegistry $typeRegistry = null)
```
alcaeus: Is there a reason why you didn't add this to the configuration class similar to metadata factories etc. I'm not sure what I prefer, so I'm totally leaving this as is and starting to extract such services from the configuration class.

src/DocumentManager.php
```diff
@@ -955,4 +957,9 @@ public function getClassNameForAssociation(array $mapping, $data): string
 
         return $mapping['targetDocument'];
     }
+
+    public function getTypes(): TypeRegistry
```
alcaeus:
```suggestion
    public function getTypeRegistry(): TypeRegistry
```

src/Types/TypeRegistry.php
```diff
@@ -0,0 +1,169 @@
...
+     *
+     * @param non-empty-string        $name
+     * @param class-string<Type>|Type $type
+     */
+    public function register(string $name, string|Type $type): void
+    {
+        if ($type instanceof Type) {
+            $this->typesMap[$name]    = $type::class;
+            $this->typeObjects[$name] = $type;
+        } else {
```
alcaeus, line 74: I suppose this case handles somebody registering a type by its FQCN. We should still ensure that the type implements the `Type` interface.

src/Types/TypeRegistry.php
```diff
@@ -0,0 +1,169 @@
...
+     *
+     * @throws InvalidTypeException
+     */
+    public function get(string $name): Type
+    {
+        if (! isset($this->typesMap[$name])) {
+            throw InvalidTypeException::invalidTypeName($name);
+        }
+
+        return $this->typeObjects[$name] ??= (new ReflectionClass($this->typesMap[$name]))->newInstanceWithoutConstructor();
```
alcaeus: We make the assumption that a type has no dependencies. This is fine for the time being, but we should consider changing this pattern and removing the option to provide just a FQCN for a type, instead requiring people to add types using a type instance.

src/Types/TypeRegistry.php
```diff
@@ -0,0 +1,169 @@
...
+
+        return $type->convertToDatabaseValue($value);
+    }
+
+    /**
+     * Get a Type instance based on the type of the passed PHP variable.
+     *
+     * @internal
+     */
+    public function fromVariable(mixed $variable): ?Type
```
alcaeus: I would rename this to better indicate that we're guessing a type, and that we're expecting a native value and not a database value.

```suggestion
    public function guessTypeFromValue(mixed $variable): ?Type
```

In future, we could consider adding a separate interface that allows types to indicate whether they can handle a certain value.

alcaeus (commented):
Initial review done. I like the idea of a type registry, but it once again shows that we really need to clean up the dependency mess that we currently have.
The only concern I have is exposing the `convertToDatabase` method from DocumentPersister. I think we should mark it as `@internal` to ensure people don't treat it as part of the public API (unless the whole persister is already marked as such).

docs/en/reference/custom-mapping-types.rst
GromNaN: Even better if I inject an instance?
```suggestion
    $types = new TypeRegistry();
    $types->register(Money::class, new App\MongoDB\Types\MoneyType());
```

src/DocumentManager.php
GromNaN: Indeed, this is also consistent with the configuration class. It's all very mixed up. The configuration class can also have methods for adding a type... which would allow `TypeRegistry` to be made internal.

GromNaN: Also, I don't like adding a method to the `DocumentManager`. So the `Configuration` class makes sense.

src/Persisters/DocumentPersister.php
```diff
@@ -1087,7 +1086,7 @@ public function prepareQueryOrNewObj(array $query, bool $isNewObj = false): arra
      *
      * @return mixed
      */
-    private function convertToDatabaseValue(string $fieldName, $value, ?ClassMetadata $class = null)
+    public function convertToDatabaseValue(string $fieldName, $value, ?ClassMetadata $class = null)
```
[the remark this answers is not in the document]
GromNaN: > The only concern I have is exposing the `convertToDatabase` method from `DocumentPersister`. I think we should mark it as `@internal` to ensure people don't treat it as part of the public API (unless the whole persister is already marked as such).

@alcaeus the class `DocumentPersister` is already all `@internal`.

src/Types/TypeRegistry.php
GromNaN: I've added a check in the `register` method to ensure the class doesn't have a constructor.

docs/en/reference/custom-mapping-types.rst
alcaeus: Absolutely!

src/Types/TypeRegistry.php
GromNaN: Method renamed with your good suggestion.

> In future, we could consider adding a separate interface that allows types to indicate whether they can handle a certain value.

That's delicate as we need to have the same detection from property type and from values.

src/Types/TypeRegistry.php
```diff
@@ -0,0 +1,179 @@
...
+use function gettype;
+use function is_object;
+use function is_subclass_of;
+use function method_exists;
+use function sprintf;
+
+/**
+ * The TypeRegistry is responsible for managing the mapping types supported.
+ */
+class TypeRegistry
```
greg0ire, line 21: Should it be final?

greg0ire:
Some commits look like they should be squashed together.

GromNaN:
> Some commits look like they should be squashed together.

Yes, my plan was to have 2 separate commits to ease with rebase and merges, but the implementation changed a lot and that will not help anymore. I'll squash-merge the PR as we usually do on the MongoDB-ODM project.

src/Types/TypeRegistry.php
GromNaN: Then I would have to create an interface since it's not internal. Which is more strict for BC than a regular class.

greg0ire: > Then I would have to create an interface since it's not internal.

Do you? Is every type supposed to be replacable/mockable? I think it's OK to marks something as `final` and postpone the decision of making it extensible vs having an interface to when a need for extensibility actually emerges. For something that looks like an map with extra methods, I'm not convinced extensibility is needed.

Feel free to ignore this if you have a hard rule about this in this project.

greg0ire (approved):

SenseException (approved):

src/Types/TypeRegistry.php
alcaeus: Agree with making it final. We don't need to mock it for internal tests (yet), and people should definitely be mocking it for their own tests considering it's a registry they can create and inject as they see fit. We can always make the class non-final in a future minor release if necessary.

alcaeus (commented):
Code LGTM. Do you want to add the deprecations to the UPGRADE document now or do we do that in the 3.0.x branch when we remove deprecated functionality?

src/Types/TypeRegistry.php
GromNaN: I've already a use-case for extending the TypeRegistry: in the bundle I add lazy-loading of type instances from the container: https://github.com/doctrine/DoctrineMongoDBBundle/pull/963/changes#diff-b090a1e775d0625ed9280a9140328e118d6ad52fafb71464bf7246ac937e8442

GromNaN: I'll rework this TypeRegistry class to also support `ContainerInterface`, and add a `TypeProvider` interface. As done in https://github.com/doctrine/dbal/pull/7342

GromNaN:
Rebased on 2.17.x and reworked to follow doctrine/dbal#7342, so both projects expose the same concepts under the same names. The description above has been rewritten accordingly.

What changed since the previous state of this PR:

- Added the `TypeProvider` interface (`get()`, `has()`, `Traversable<string, Type>`). It is now what `Configuration` accepts and returns, so `TypeRegistry` is one implementation rather than the only option.
- `TypeRegistry` is `final` and implements `TypeProvider` and `IteratorAggregate`. Built-in types moved to a class constant, and the constructor takes a map of instances or class names layered on top.
- Types can be lazy-loaded from a PSR-11 container, given a map of type names to service IDs. The container is never queried during construction, nor by `has()`. This is what lets the bundle wire a service locator instead of building every custom type at boot.
- `getMap()` is replaced by `getIterator()`, so stopping the iteration early no longer instantiates the remaining types.
- `guessTypeFromValue()` and `convertToDatabaseValue()` moved out of the registry into a separate `@internal` `TypeGuesser`. These two are ODM specifics with no DBAL equivalent, and keeping them off the interface makes `TypeProvider` implementable by third parties.
- `register()` no longer instantiates a class just to validate it. The constructor is inspected by reflection instead, so registering by class name stays lazy.
- Tests inject a registry scoped to the test `Configuration`, which removes the reflection reset of the shared instance in `tearDown()`.

I kept the immutability guard on `Configuration::setTypeRegistry()`, which DBAL does not have. Swapping the registry after `ClassMetadataFactory` has injected it into already loaded metadata would leave those instances pointing at the old one. I intend to propose the same guard for DBAL rather than drop it here.

`psr/container` is a `require-dev` plus a `suggest`, not a hard requirement, since the union type and the `catch` clause are only resolved when a container is actually passed.

This also fixes three tests that were already failing on the branch: the constructor rule relaxed in fc4d6e5 was never reflected in `CustomTypeTest`, and two deprecation messages still said 2.16.

composer.json
```diff
@@ -55,6 +55,7 @@
         "phpstan/phpstan-deprecation-rules": "^2.0",
         "phpstan/phpstan-phpunit": "^2.0",
         "phpunit/phpunit": "^10.5.58|^11.5.43",
+        "psr/container": "^1.1 || ^2.0",
         "squizlabs/php_codesniffer": "^4",
         "symfony/cache": "^5.4 || ^6.0 || ^7.0 || ^8.0",
         "symfony/uid": "^5.4 || ^6.0 || ^7.0 || ^8.0"
```
alcaeus, line 61: CI is red: the `lowest`-deps job resolves `symfony/uid` to a version predating `Uuid::v7()`, which the new `TypeGuesserTest` uses.

<details><summary>Details</summary>

`PHPUnit (lowest, 8.1, 6.0, 1.21.0, server, lazy-ghost)` fails with `Call to undefined method Symfony\Component\Uid\Uuid::v7()`. Either bump the lower bound here to the release that added `v7()`, or have the test build the UUID a way that works across the whole supported range (e.g. `Uuid::fromString()`/`v4()`).
</details>

tests/Tests/Types/TypeGuesserTest.php
```diff
@@ -0,0 +1,80 @@
...
+    public static function provideTypeToGuessFromValue(): Generator
+    {
+        yield 'null' => [null, null];
+        yield 'bool' => [Type::BOOL, true];
+        yield 'int' => [Type::INT, 1];
+        yield 'float' => [Type::FLOAT, 3.14];
+        yield 'string' => [Type::STRING, 'ohai'];
+        yield 'DateTime' => [Type::DATE, new DateTime()];
+        yield 'DateTimeImmutable' => [Type::DATE_IMMUTABLE, new DateTimeImmutable()];
+        yield 'Uuid' => [Type::UUID, Uuid::v7()];
```
alcaeus, line 46: Failing line for the `lowest` CI job — see the `composer.json` comment. Blocking until CI is green.

src/Types/TypeRegistry.php
```diff
@@ -0,0 +1,248 @@
...
+            $this->typeObjects[$name] = $type;
+
+            return;
+        }
+
+        if (! is_subclass_of($type, Type::class)) {
+            throw InvalidTypeException::invalidTypeClass($name, $type);
+        }
+
+        $constructor = (new ReflectionClass($type))->getConstructor();
```
alcaeus, line 140: `register()` only checks constructor arity, not instantiability — an abstract class or private-constructor class sails through and only blows up later with a bare `Error` at `get()` time, not `InvalidTypeException`.

<details><summary>Repro + suggested fix</summary>

```php
abstract class MyAbstractType extends RawType {}
$registry->register('x', MyAbstractType::class); // no exception here
$registry->get('x'); // Error: Cannot instantiate abstract class MyAbstractType
```

`ReflectionClass::isInstantiable()` covers abstract/interface/private-constructor in one call:

```suggestion
        $reflectionClass = new ReflectionClass($type);
        if (! $reflectionClass->isInstantiable()) {
            throw InvalidTypeException::classNotInstantiable($type);
        }

        $constructor = $reflectionClass->getConstructor();
        if ($constructor !== null && $constructor->getNumberOfRequiredParameters() > 0) {
            throw InvalidTypeException::constructorHasRequiredParameters($type);
        }
```

(plus a new `InvalidTypeException::classNotInstantiable()` factory, mirroring the others in that class.)
</details>

tests/Tests/Types/TypeRegistryTest.php
```diff
@@ -0,0 +1,263 @@
...
+        $registry->register('invalid_type', stdClass::class);
+    }
+
+    public function testRegisterRejectsClassWithRequiredConstructorParameters(): void
+    {
+        $registry = new TypeRegistry();
+        self::expectException(InvalidArgumentException::class);
+        self::expectExceptionMessage('must not have a constructor with required parameters');
+        $registry->register('my_type', TypeWithRequiredConstructor::class);
+    }
```
alcaeus, line 100: Minor: worth a case here for an abstract/private-constructor class registered by FQCN, to pin down the desired behavior — see the `TypeRegistry::register()` comment.

src/UnitOfWork.php
```diff
@@ -832,12 +831,11 @@ private function computeOrRecomputeChangeSet(ClassMetadata $class, object $docum
 
                 // skip equivalent date values
                 if (isset($class->fieldMappings[$propName]['type']) && $class->fieldMappings[$propName]['type'] === 'date') {
-                    $dateType = Type::getType('date');
-                    assert($dateType instanceof DateType);
+                    $dateType      = $class->getFieldType($propName);
```
alcaeus, line 834: Minor: the old `assert($dateType instanceof DateType)` documented and debug-checked that this is the built-in date converter. Now implicit only.

<details><summary>Why this is low-risk</summary>

Not a behavior change — `convertToDatabaseValue()` is declared on `Type` itself, and the shared-registry lookup could already be overridden the same way before this PR. Just flagging that the explicit check quietly disappeared along with the `DateType` import.
</details>

src/Configuration.php
```diff
@@ -154,6 +156,22 @@ class Configuration
 
     private static string $version;
 
+    private ?TypeProvider $typeRegistry = null;
+
+    public function setTypeRegistry(TypeProvider $typeRegistry): void
+    {
+        if ($this->typeRegistry !== null) {
+            throw new LogicException('TypeRegistry is already set and cannot be changed.');
```
alcaeus, line 164: Minor: the UPGRADE note says this is settable "once, before any metadata is loaded," but the guard is actually simpler/stricter — it rejects any second `setTypeRegistry()` call outright, not specifically tied to metadata-loaded state. Safe, just worth knowing if a caller expects to reconfigure a still-fresh `Configuration`.

alcaeus (commented):
Re-reviewed after the 2026-08-25 rework (TypeProvider extraction, `final`/`IteratorAggregate` TypeRegistry, PSR-11 lazy loading, TypeGuesser split, vector-search conversion fix). Lines up nicely with the DBAL shape and the design is sound overall.

All 5 points from the earlier review round are resolved except a residual gap in point 5 — see below.

<details><summary>Status of earlier review comments</summary>

1. `convertToDatabaseValue()` internal — resolved via the class-level `@internal` on `DocumentPersister` (line 71).
2. `register()` checks `Type` implementation before reflecting — resolved (`TypeRegistry.php:136`).
3. `TypeRegistry` is `final` — resolved.
4. Registry lives on `Configuration`, not `DocumentManager` — resolved.
5. `register()` no longer instantiates to validate — resolved for the happy path, but see the inline comment on `TypeRegistry.php`: the reflection check only covers constructor arity, not instantiability.
</details>

New findings inline. Recommend fixing the CI failure and the `register()` gap before merge; the rest can go either way.

GromNaN:
Thanks for the review, Andreas. The two blocking points are now addressed in the rebased commit 5323a517 (single squashed commit on top of 2.18.x):

## CI (lowest) failure on `symfony/uid`

Bumped the lower bound in `composer.json` from `^5.4` to `^6.4` so `Uuid::v7()` used by `TypeGuesserTest` is available across the whole supported range. The lowest-deps job should now resolve a version that has `v7()`.

## `register()` instantiability

I decided not to add the `ReflectionClass::isInstantiable()` check on `register()`. It would load the class and run reflection on a path that is called for every registered type, at the expense of runtime performance. The existing constructor-arity check is kept. As a consequence, the abstract/private-constructor test case is skipped: it only makes sense if the instantiability guard is enforced.

## Minor comments

- `UnitOfWork.php`: restored the `assert($dateType instanceof DateType)` and the `DateType` import on the equivalent-date-value shortcut.
- `UPGRADE-2.17.md`: reworded the note so it matches the actual guard, which rejects any second `setTypeRegistry()` call rather than being tied to metadata-loaded state.

src/Aggregation/Stage/AbstractBucket.php
```diff
@@ -86,7 +86,7 @@ private function convertExpression($expression)
             return '$' . $this->getDocumentPersister()->prepareFieldName(substr($expression, 1));
         }
 
-        return Type::convertPHPToDatabaseValue(Expr::convertExpression($expression));
+        return (new TypeGuesser($this->dm->getConfiguration()->getTypeRegistry()))->convertToDatabaseValue($expression);
```
IonBazan:
```suggestion
        return (new TypeGuesser($this->dm->getConfiguration()->getTypeRegistry()))->convertToDatabaseValue(Expr::convertExpression($expression));
```

Is this expected behaviour change?

src/Mapping/ClassMetadata.php
```diff
@@ -2888,7 +2920,7 @@ private function validateAndCompleteTypedFieldMapping(array $mapping): array
             return $mapping;
         }
 
-        if (! $type->isBuiltin() && Type::hasType($type->getName())) {
+        if (! $type->isBuiltin() && $this->getTypeRegistry()->has($type->getName())) {
```
IonBazan: This could cause very unlikely cache issues if two document managers with different registries are using the same cache.

UPGRADE-2.17.md
IonBazan, line 1: Move to 2.18

src/Mapping/ClassMetadata.php
```diff
@@ -932,6 +937,24 @@ public static function getReferenceFieldName(string $storeAs, string $pathPrefix
...
+    public function setTypeRegistry(TypeProvider $types): void
+    {
+        $this->typeRegistry = $types;
+    }
+
+    private function getTypeRegistry(): TypeProvider
+    {
+        if (! isset($this->typeRegistry)) {
+            $this->typeRegistry = TypeRegistry::getSharedInstance();
+            trigger_deprecation('doctrine/mongodb-odm', '2.17', 'Using ClassMetadata without a TypeRegistry is deprecated. Inject the TypeRegistry instance from the DocumentManager via $classMetadata->setTypeRegistry($configuration->getTypeRegistry()).');
```
IonBazan:
```suggestion
            trigger_deprecation('doctrine/mongodb-odm', '2.18', 'Using ClassMetadata without a TypeRegistry is deprecated. Inject the TypeRegistry instance from the DocumentManager via $classMetadata->setTypeRegistry($configuration->getTypeRegistry()).');
```

src/Types/Type.php
```diff
@@ -155,133 +107,130 @@ public function closureToPHP(): string
 
     /**
      * Register a new type in the type map.
+     *
+     * @deprecated Use {@see TypeRegistry::register()} instead
      */
     public static function registerType(string $name, string $class): void
     {
-        self::$typesMap[$name] = $class;
+        trigger_deprecation('doctrine/mongodb-odm', '2.17', 'Type::registerType() is deprecated, use $typeRegistry->register() instead.');
```
IonBazan: Bump all remaining deprecation versions

src/Types/Type.php
```diff
@@ -155,133 +107,130 @@ public function closureToPHP(): string
...
+        trigger_deprecation('doctrine/mongodb-odm', '2.17', 'Type::getType() is deprecated, use $typeRegistry->get() instead.');
 
-        return self::$typeObjects[$type] ??= new (self::$typesMap[$type]);
+        return TypeRegistry::getSharedInstance()->get($type);
     }
 
     /**
      * Get a Type instance based on the type of the passed php variable.
      *
+     * @deprecated Use {@see TypeGuesser::guessTypeFromValue()} instead
```
IonBazan, line 137: `TypeGuesser` is marked as internal

src/UnitOfWork.php
```diff
@@ -758,12 +758,12 @@ private function computeOrRecomputeChangeSet(ClassMetadata $class, object $docum
 
                 // skip equivalent date values
                 if (isset($class->fieldMappings[$propName]['type']) && $class->fieldMappings[$propName]['type'] === 'date') {
-                    $dateType = Type::getType('date');
+                    $dateType = $class->getFieldType($propName);
                     assert($dateType instanceof DateType);
```
IonBazan: This will fail if someone overrides default `date` type which is encouraged in the docs added.

src/Persisters/DocumentPersister.php
```diff
@@ -1104,7 +1104,7 @@ private function convertToDatabaseValue(string $fieldName, $value, ?ClassMetadat
                 $value = $value->value;
             }
 
-            return Type::convertPHPToDatabaseValue($value);
+            return (new TypeGuesser($this->dm->getConfiguration()->getTypeRegistry()))->convertToDatabaseValue($value);
```
IonBazan: New instance will be created for every query value, recursively for arrays. The aggregation stages do the same.

IonBazan (changes requested):

src/Aggregation/Stage/AbstractBucket.php
GromNaN: Applied. convertExpression() now expands an Expr value through Expr::convertExpression() before converting it, and the conversion goes through the new DocumentManager::getTypeGuesser() instead of building a TypeGuesser each time. Added BucketTest::testGroupByExpandsExprValue to cover an Expr passed to groupBy.

src/Mapping/ClassMetadata.php
GromNaN: Agreed, two document managers that share the same metadata cache while having different type registries is a potential consistency issue. We consider shared cross-manager caches out of scope for this PR and do not support them, so no change here.

UPGRADE-2.17.md
GromNaN: Moved. The TypeRegistry upgrade instructions now live in a new UPGRADE-2.18.md document, preserving its existing 2.18 content, and UPGRADE-2.17.md is left identical to the 2.17.x branch.

src/Mapping/ClassMetadata.php
GromNaN: Done, the deprecation now reports 2.18 to match the project convention.

src/Types/Type.php
GromNaN: Done, all deprecations introduced by this PR now report 2.18: the static Type::* methods and the ClassMetadata registry fallback. Pre-existing 2.16 and 2.17 deprecations from other features were left untouched.

src/Types/Type.php
GromNaN: Agreed that this is inconsistent. TypeGuesser is intentionally internal: the guessing path is now centralized behind the also internal DocumentManager::getTypeGuesser(), and the static methods stay deprecated without exposing an internal class as a public recommendation. We are finalizing the wording of these two deprecations before merge.

src/UnitOfWork.php
GromNaN: Good catch. I removed the assert($dateType instanceof DateType), which would throw when the date type is overridden. The loose comparison now only applies when both converted values are MongoDB\\BSON\\UTCDateTime.

src/Persisters/DocumentPersister.php
GromNaN: Fixed. TypeGuesser is no longer instantiated per value. It is now held and provided once by DocumentManager::getTypeGuesser(), and reused by all aggregation stages and the persister.

GromNaN (commented):
Thanks for the careful review, Ion. I replied to each thread. All the code points are implemented in commit 8a6ef670: the UTCDateTime-scoped date comparison in UnitOfWork, the DocumentManager::getTypeGuesser() consolidation, the Expr expansion in the bucket stage with a test, and the deprecation versions bumped to 2.18. The upgrade instructions moved to UPGRADE-2.18.md. The only item still being finalized is the wording of the two deprecations that referenced TypeGuesser; happy to take your input on that.

src/Types/Type.php
GromNaN: I'll make `DocumentManager::getTypeGuesser()` and the `TypeGuesser` class part of the public API (non-internal).

GromNaN: Resolved: TypeGuesser and DocumentManager::getTypeGuesser() are no longer marked @internal, so the deprecations now point to a public replacement. The deprecation messages and the UPGRADE-2.18.md mapping were updated accordingly in commit e2f75b68. Thanks for flagging the inconsistency.

tests/Tests/Types/TypeGuesserTest.php
GromNaN: Switching to UUID v4.

src/Types/TypeRegistry.php
GromNaN: Thanks for the suggestion, I initially missunderstood the point. Since we already use the Reflection on the class, calling `isInstantiable` doesn't hurt.

src/UnitOfWork.php
GromNaN: Now checking if both before/after values are instances of UTCDateTime.

src/Mapping/ClassMetadata.php
```diff
@@ -932,6 +937,24 @@ public static function getReferenceFieldName(string $storeAs, string $pathPrefix
         return ($pathPrefix ? $pathPrefix . '.' : '') . self::getReferencePrefix($storeAs) . 'id';
     }
 
+    /**
+     * Inject the TypeRegistry instance, used for field transformation and type detection.
+     */
+    public function setTypeRegistry(TypeProvider $types): void
+    {
+        $this->typeRegistry = $types;
+    }
```
IonBazan, line 946: Inconsistent naming between method, property, class name and comments. Let's choose a single term for this, specially that `TypeRegistry` is a different class.

src/Types/TypeGuesser.php
```diff
@@ -0,0 +1,67 @@
...
+use DateTimeInterface;
+use Symfony\Component\Uid\Uuid;
+
+use function gettype;
+use function is_object;
+
+/**
+ * Determines the mapping type to use for a value that has no declared type.
+ */
+final class TypeGuesser
```
IonBazan, line 17: Should we allow customizing the type guesser too?

src/Types/TypeRegistry.php
```diff
@@ -0,0 +1,253 @@
...
+     * container service IDs providing them. Types are resolved on first access and cached.
+     *
+     * @param array<string, Type|class-string<Type>>|ContainerInterface $types
+     * @param array<string, string>|null                                $serviceIds Map of type names to container
+     *                                                                              service IDs. Required when passing
+     *                                                                              a container, in which case an empty
+     *                                                                              map means no types beyond the
+     *                                                                              built-in ones.
+     */
+    public function __construct(array|ContainerInterface $types = [], ?array $serviceIds = null)
```
IonBazan: Perhaps I'm missing something but why do we need to provide `$serviceIds`? Can we encourage to provide sub-containers containing only the types instead?

src/Types/TypeRegistry.php
```diff
@@ -0,0 +1,253 @@
...
+    private static ?TypeRegistry $sharedInstance = null;
+
+    /**
+     * Creates a registry pre-populated with all built-in types. Additional types passed via
+     * {@param $types} are registered on top; if a name matches a built-in type it is
+     * overridden rather than re-registered.
+     *
+     * A {@see ContainerInterface} can be passed instead of an array to lazy-load type instances
+     * from a service container. In that case, {@param $serviceIds} maps type names to the
+     * container service IDs providing them. Types are resolved on first access and cached.
```
IonBazan: Since types are cached, is it safe to reuse TypeRegistry instance between requests/queue jobs etc? I'm also wondering what's the performance impact of this change and whether we should allow injecting an instance of cache interface instead, so it works like ClassMetadata.

src/Mapping/ClassMetadata.php
```diff
@@ -932,6 +937,24 @@ public static function getReferenceFieldName(string $storeAs, string $pathPrefix
...
+
+    private function getTypeRegistry(): TypeProvider
+    {
+        if (! isset($this->typeRegistry)) {
+            $this->typeRegistry = TypeRegistry::getSharedInstance();
+            trigger_deprecation('doctrine/mongodb-odm', '2.18', 'Using ClassMetadata without a TypeRegistry is deprecated. Inject the TypeRegistry instance from the DocumentManager via $classMetadata->setTypeRegistry($configuration->getTypeRegistry()).');
```
IonBazan: Now that #3055 and other PRs are targeting the same release, I wonder if it wouldn't be better to explicitly target 2.19 instead. This should help diagnose potential issues easier, since both changes affect the core, metadata and UoW a lot.

src/Types/TypeRegistry.php
GromNaN: Good catch, that's an artefact of the duplication of the logic from DBAL. The `ContainerInterface` is not iterable and does not expose the list of service ids. In DBAL, the full list is used in [`AbstractPlatform`](https://github.com/doctrine/dbal/blob/205fdf552ffafe33c1d12a41a2d6c7217666c32c/src/Platforms/AbstractPlatform.php#L174).
This design was suggested by https://github.com/doctrine/dbal/pull/7342#issuecomment-4197441808

The method `Type::getTypesMap` is never used in MongoDB ODM or its bundle. But I must keep this deprecated method. Also, it's useful to show the list of type in the Symfony profiler and a debug command (but that's not implemented yet).

GromNaN: If we use [`Symfony\Contracts\Service\ServiceProviderInterface`](https://github.com/symfony/symfony/blob/8.2/src/Symfony/Contracts/Service/ServiceProviderInterface.php), we can leverage the `getProvidedServices` method.

src/Types/TypeGuesser.php
GromNaN: That would be smart as a follow-up. `ClassMetadata::validateAndCompleteTypedFieldMapping()` should also be in this class so that both logics to map a value or a field definition to a type class would be collocated and customizable.

src/Mapping/ClassMetadata.php
GromNaN: I agree. Let's keep `TypeRegistry` for the concrete implementation as it's the name already in released versions of DBAL. And rename everything else to `TypeProvider`.

src/Mapping/ClassMetadata.php
alcaeus: I'd vote for including the type registry in 2.18.x alongside the already-merged #3051 (extracting `DocumentRegistry` from UoW). We can then continue refactoring UoW and persisters in 2.19 and following versions.

src/Types/TypeRegistry.php
GromNaN: Thanks for the careful review, you hit on the right two axes (lifetime across requests/queue jobs, and where the caching boundary should sit). Here is how the design addresses them, and what changed in the PR.

For the "safe to reuse" question: the TypeRegistry only memoizes what it must. Class-name and built-in types are resolved once and held in memory, which is exactly the behavior we want for long-running reuse. Service-backed types are not memoized by the registry: each access delegates to the ServiceLocator behind it, so instance identity (and any reset behavior) is governed by the container, not by the registry. A `shared: false` or `reset_on_request` type service is therefore not pinned to a stale instance across HTTP requests or Messenger messages.

For the lifecycle parallel with ClassMetadata: the registry is injected into each ClassMetadata at runtime by ClassMetadataFactory, in both `wakeupReflection()` and `initializeReflection()`. The registry property is excluded from serialization because it is not in the `__sleep()` allowlist, so it never rides along in the serialized metadata cache and is re-injected on use.

For the performance and "inject a cache interface" question: I do not think a PSR cache is a good fit here, and it would diverge from why the metadata cache works. Type instances are few, instantiated once, and frequently non-serializable (they can hold a client or configuration object), whereas metadata is a serializable array of class information. The lazy resolution already makes the lookup O(1) after first use.

What was modified in the PR in response to this discussion:

- The TypeRegistry no longer memoizes the service branch: it delegates to the ServiceLocator on each call, while class-name and built-in types keep their in-memory memoization.
- The bundle now documents the opt-in behavior: once a custom type is declared as a service (service: config entry or the `#[AsFieldType]` attribute), that document manager uses its scoped registry and stops using the global one. The global registry is kept for backward compatibility with projects that still register types through the global `Type::register()`.
- The tests were adapted to pin these semantics: shared versus non-shared resolution, propagation of the service provider exception, and non-eager service resolution during iteration.

src/Mapping/ClassMetadata.php
GromNaN: I updated to 2.19, but I can revisit.
I don't have a strong opinion on grouping major changes in a release. It's still possible to bisect using commit hashes.
Something we can do is release 2.18 as it is now (modulo the pending PRs), merge this and release 2.19, then other work on the UoW can go to 2.20.
Or we bundle everything...

src/Types/TypeRegistry.php
IonBazan: Thanks for explanation - good to add tests covering these edge cases (if not done yet, sorry my human context window is quite narrow 😂).  This should prevent future optimizations from introducing regressions

IonBazan: https://github.com/doctrine/mongodb-odm/pull/2966/changes#diff-a09dc60c9849279eeac054b9a2efa0a686b5581f9a3272855712639ecbfda6dbR164 - this line still memoizes the resolved container service. Let's add explicit test to cover this scenario (eg. resolved service changes between container get calls).

src/Types/TypeRegistry.php
```diff
@@ -0,0 +1,208 @@
...
+
+        if (isset($this->typesMap[$name])) {
+            return $this->typeObjects[$name] = new ($this->typesMap[$name])();
+        }
+
+        if ($this->services?->has($name)) {
+            try {
+                $type = $this->services->get($name);
+            } catch (NotFoundExceptionInterface) {
+                throw InvalidTypeException::invalidServiceTypeNotFound($name, $name);
```
IonBazan: Because we already checked `has()`, the only way `get()` throws `NotFoundExceptionInterface` is when one of service dependencies could not be resolved. This exception obscures this fact and makes it harder to debug. Adjust exception message, pass the `$previous` to fix that. We are also passing `$name` twice here.

src/Types/TypeRegistry.php
```diff
@@ -0,0 +1,208 @@
...
+                yield $name => $this->get($name);
+            }
+        }
+    }
+
+    /** @internal Do not use this method. */
+    public static function getSharedInstance(): TypeRegistry
+    {
+        return self::$sharedInstance ??= new self();
+    }
```
IonBazan: Maybe we could move it to UoW or somewhere else to prevent this from being used?

UPGRADE-2.18.md
```diff
@@ -78,3 +78,50 @@ direction and now produces the correct `$meta` expression.
...
+- `Type::hasType($name)` → `TypeRegistry::has($name)`
+- `Type::addType($name, $class)` → `TypeRegistry::register($name, $class)`
+- `Type::registerType($name, $class)` → `TypeRegistry::register($name, $class)`
+- `Type::overrideType($name, $class)` → `TypeRegistry::register($name, $class)`
+- `Type::getTypesMap()` → iterate over the `TypeRegistry`
+- `Type::getTypeFromPHPVariable($value)` → `DocumentManager::getTypeGuesser()->guessTypeFromValue($value)`
+- `Type::convertPHPToDatabaseValue($value)` → `DocumentManager::getTypeGuesser()->convertToDatabaseValue($value)`
+
+You can set and get the type registry on `Doctrine\ODM\MongoDB\Configuration`
+using `getTypeRegistry()` and `setTypeRegistry()`. Types registered on a registry
```
IonBazan:
```suggestion
using `getTypeProvider()` and `setTypeProvider()`. Types registered on a registry
```

Apply same to other instances in this file.

src/Types/TypeRegistry.php
```diff
@@ -0,0 +1,208 @@
...
+     *
+     * @throws InvalidTypeException
+     */
+    public function getIterator(): Generator
+    {
+        yield from $this->typeObjects;
+
+        $serviceNames = $this->services?->getProvidedServices() ?? [];
+
+        // Resolving adds to $this->typeObjects, which is why each name is tracked in $seen instead.
```
IonBazan: $seen has been removed - update comment accordingly

tests/Tests/Types/TypeRegistryTest.php
```diff
@@ -0,0 +1,273 @@
...
+use stdClass;
+use Symfony\Contracts\Service\ServiceProviderInterface;
+
+use function array_combine;
+use function array_fill;
+use function array_keys;
+use function count;
+use function iterator_to_array;
+
+class TypeRegistryTest extends BaseTestCase
```
IonBazan:
```suggestion
class TypeRegistryTest extends TestCase
```

We don't need Mongo to test this.

src/Types/TypeRegistry.php
GromNaN: I'll rename it to `getDeprecatedSharedInstance` to be more explicit.

src/Types/TypeRegistry.php
GromNaN: I removed the catch entirely, so the provider exception now bubbles up as-is, and phpstan passes without any workaround.

src/Types/TypeRegistry.php
GromNaN: Renamed to getDeprecatedSharedInstance() and kept it @internal, since the method is new and only retained for backward compatibility.

UPGRADE-2.18.md
GromNaN: Aligned UPGRADE-2.18.md to use getTypeProvider() and setTypeProvider() consistently.

tests/Tests/Types/TypeRegistryTest.php
GromNaN: TypeRegistryTest now extends TestCase directly, since these tests do not require Mongo.

src/Mapping/ClassMetadata.php
GromNaN: The naming is now consistently typeProvider across the property, the accessor methods and the TypeProvider interface.

src/Mapping/ClassMetadata.php
GromNaN: Kept the target release at 2.18, so the deprecations and the UPGRADE note now report 2.18.

src/Types/TypeRegistry.php
GromNaN: I forgot to push the changes while posting the comment. Fixed.

IonBazan (approved):

alcaeus (approved):
Great work @GromNaN!

Issue #1837, opened 2018-07-31
Using references on embedded document field


MatTheCat:
|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

The query builder should allow queries like

```
->field('memberships.organization')->references($organization)
```

Currently I'm forced to know how the reference is stored and then write

```
->field('memberships.organization.id')->equals(new \MongoId($organization))
```

This is because `Doctrine\ODM\MongoDB\Query\Expr::getReferenceMapping` only checks field mapping.

malarzm:
I was pretty sure we had a similar request in the past but as I can't find I'm gonna keep this open. Scheduling for 2.x, but if @MatTheCat you'd like to tackle this yourself it might find a way into 1.3 :)

MatTheCat:
I did a quick search but couldn't find anything. As I have a workaround and little time I don't think I'll be able to help :confused:

alcaeus:
> I was pretty sure we had a similar request in the past but as I can't find I'm gonna keep this open.

You were thinking of #1723, which dealt with a similar issue but without the `references` call: https://github.com/doctrine/mongodb-odm/issues/1723#issuecomment-356531034

DuaelFr:
Hi! Duael from 2025 reporting that this is still an issue ;)
Here is the current workaround:

```php
    $query->field('myProperty.mySubField')->equals([
      '$ref' => $this->documentManager->getClassMetadata(MyTargetDocument::class)->getCollection(),
      '$id'  => new ObjectId($targetDocument->getId()),
    ]);
```

It would be very nice if the field mapping was able to traverse this kind of key.

alcaeus:
Thanks for the feedback @DuaelFr! Which version are you currently using? With 2.13.0 we improved handling of preparing queries on embedded documents (see #2827 for details). I would expect that PR to fix this but it could very well be that we missed it. If you can let me know whether that is still broken in 2.13+ I'd be happy to take a look. To confirm, you have an reference in an embedded document, and the query builder for the containing document does not correctly handle the reference?

DuaelFr:
Hi @alcaeus! Thank you for your quick answer.
I was using 2.11.x yesterday (some kind of ext-mongodb version issue)  so I upgraded and tried again but it's all the same.

Here is my setup right now:
```shell
$ composer show doctrine/mongo*
[...]

PR #2969, opened 2025-12-12
Always prepare full references when querying

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | Related to #1837 

#### Summary

I noticed that the changes made in #2827 always created partial queries for references. This can lead to wrong results when using discriminated references where identifiers are reused across collections (e.g. when using ascending numeric identifiers). This PR fixes this, bringing the behaviour of `Expr::equals` in line with `Expr::references` and `Expr::includesReferenceTo`, essentially making those methods unnecessary.

This fix required some changes in tests where we made assertions on query shapes. I made sure that the changes don't introduce a regression with shard keys, and I'm sure that it doesn't. While looking at this, I noticed that the support for references as shard keys may actually not work as expected due to how indexes work when the index value is a document. In order to match a document, both key order and values need to be identical, so if a document contains a value of `{ foo: 'bar', bar: 'baz' }`, using `{ bar: 'baz', foo: 'bar' }` will not match despite the documents having the same PHP representation (and being equal for all intents and purposes). This may also mean that the typical reference queries of `{ "bar.$id": "..." }` may not necessarily be able to use the underlying index of the shard key. I'm not particularly worried about this, but wanted to mention it in case someone trips over this.

I'll be following up this PR with another one to deprecate `references` and `includesReferenceTo` in 2.16.x for subsequent removal in 3.0.

### DIFF
--- a/src/Persisters/DocumentPersister.php
+++ b/src/Persisters/DocumentPersister.php
@@ -41,7 +41,6 @@ use stdClass;
 
 use function array_combine;
 use function array_fill;
-use function array_intersect_key;
 use function array_key_exists;
 use function array_keys;
 use function array_map;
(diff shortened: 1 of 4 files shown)


src/Persisters/DocumentPersister.php
```diff
@@ -1610,36 +1609,11 @@ private function prepareReference(string $fieldName, object $value, array $mappi
...
-
-        if ($mapping['type'] === ClassMetadata::MANY) {
-            return [[$fieldName, ['$elemMatch' => array_intersect_key($reference, $keys)]]];
-        }
-
-        return array_map(
-            static fn ($key) => [$fieldName . '.' . $key, $reference[$key]],
-            array_keys($keys),
-        );
+        return $mapping['type'] === ClassMetadata::MANY
```
GromNaN: I'd rather keep a regular `if` instead of the ternary, for readability.

GromNaN (approved):
I had difficulty understanding how it worked, but ultimately it is much simpler.

src/Persisters/DocumentPersister.php
ranfysvalle02: Here’s a suggestion you might want to try:

```php
if ($mapping['type'] === ClassMetadata::MANY) {
    return [[$fieldName, ['$elemMatch' => $reference]]];
} else {
    return array_map(static fn($key) => [$fieldName . '.' . $key, $reference[$key]], array_keys($reference));
}
```

Though I do see the PR has already been approved

alcaeus: @GromNaN updated, and I even added some explanatory comments while I was at it.

src/Persisters/DocumentPersister.php
```diff
@@ -1606,40 +1605,24 @@ private function isInTransaction(array $options): bool
...
-                }
-
-                break;
-
-            default:
-                throw new InvalidArgumentException(sprintf('Reference type %s is invalid.', $mapping['storeAs']));
-        }
-
+        // For other ReferenceMany fields, we need to use $elemMatch to find a single array element that matches all
+        // fields
```
GromNaN: Nit: this line it too long, you should move some words to the 2nd line. Or make it a single line.

src/Persisters/DocumentPersister.php
```diff
@@ -1606,40 +1605,24 @@ private function isInTransaction(array $options): bool
...
         if ($mapping['type'] === ClassMetadata::MANY) {
-            return [[$fieldName, ['$elemMatch' => array_intersect_key($reference, $keys)]]];
+            return [[$fieldName, ['$elemMatch' => $reference]]];
         }
 
+        // For ReferenceOne fields, we can use multiple conditions on individual fields, prefixed with the field name
+        // of the reference
```
GromNaN:
```suggestion
        // For ReferenceOne fields, we can use multiple conditions on individual fields,
        // prefixed with the field name of the reference
```

src/Persisters/DocumentPersister.php
alcaeus: Done. Couldn't remember whether we do 80 or 120 characters and guessed wrong.

PR #2972, opened 2025-12-17
Don't include static properties in skipped properties

MidnightDesign:
Otherwise, `skipLazyInitialization()` is called on the reflection of static properties, resulting in an error.

|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | 

#### Summary

`NativeLazyObjectFactory#getSkippedProperties()` previously returned static properties, resulting in a `ReflectionException: Can not use skipLazyInitialization on static property (...)` when `skipLazyInitialization()` was then called on it.

### DIFF
--- a/src/Proxy/Factory/NativeLazyObjectFactory.php
+++ b/src/Proxy/Factory/NativeLazyObjectFactory.php
@@ -101,6 +101,10 @@ class NativeLazyObjectFactory implements ProxyFactory
                 continue;
             }
 
+            if ($property->isStatic()) {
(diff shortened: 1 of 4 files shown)


frenchcomp:
Thanks, I have the same issue and I writing a similar fix. 
It's works for me (Without I have the error "Can not use skipLazyInitialization on static property")

GromNaN (approved):
Thank you @MidnightDesign.

PR #2976, opened 2025-12-26
[3.0] Remove annotations support

IonBazan:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | yes
| Fixed issues | 

#### Summary

Removes `Annotations` and adds types to `Attribute` namespace. Most attributes were marked as `final` and `readonly` with following exceptions:
- ~~`AbstractField` cannot be `readonly`~~
- ~~`SearchIndex` was not marked as `@final` earlier~~
- ~~`VectorSearchIndex` was not marked as `@final` earlier~~
- ~~`Validation` was not marked as `@final` earlier~~


src/Mapping/Attribute/AbstractField.php
```diff
@@ -4,46 +4,17 @@
 
 namespace Doctrine\ODM\MongoDB\Mapping\Attribute;
 
-use function class_alias;
-
 /** @internal */
 abstract class AbstractField
```
IonBazan, line 8: `AbstractField` cannot be `readonly` because `EmbedMany`/`EmbedOne` and `ReferenceMany`/`ReferenceOne` have a property `$embedded` and `$reference` with default value. Therefore they cannot be `readonly` themself. Mixing non-readonly and readonly classes in inheritance chain is not supported, so classes like `Id`, `Filename`, etc cannot be marked as `readonly` either.
Although I didn't find ant usages of `$embedded` nor `$reference`, I am not sure if we can remove this property.

IonBazan: Fixed by adding conditions in `AttributeDriver`:
https://github.com/IonBazan/mongodb-odm/blob/ecd29d247a8b2a3fa85a20e6975a138b9a333624/src/Mapping/Driver/AttributeDriver.php#L218-L228

I believe these removed properties should be marked as `@internal` or `@deprecated` in 2.16.x

GromNaN: I would not mark classes as readonly. This is restrictive for making changes to these classes.  Properties can be readonly.

IonBazan: Thanks for the suggestion - changed it in 967e2ec88029c581e3027b1644be04310207a898

src/Mapping/Attribute/ReadPreference.php
```diff
@@ -6,20 +6,11 @@
...
 {
-    /** @var string */
-    public $value;
-
-    /** @var string[][]|null */
-    public $tags;
-
     /** @param string[][]|null $tags */
-    public function __construct(string $value, ?array $tags = null)
+    public function __construct(public readonly string $value, public readonly ?array $tags = null)
```
GromNaN:
```suggestion
    public function __construct(
        public readonly string $value,
        public readonly ?array $tags = null,
    )
```

GromNaN (commented):
Could you also update the UPGRADE file please?

GromNaN (approved):
Thank you @IonBazan

PR #2978, opened 2026-01-09
Add types in tests

IonBazan:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

Makes entire code base fully typed by covering test cases too. Mostly `mixed` and union/intersection types were added as we dropped support for older PHP versions in 3.0.


alcaeus (approved):

tests/Tests/Functional/CustomTypeTest.php
```diff
@@ -99,11 +96,11 @@ public function testTypeFromPHPVariable(): void
     public function testNotOverridingClosureToPHPIsDeprecated(): void
     {
         $type = Type::getType('custom_type_without_closure_to_php');
+        $this->expectUserDeprecationMessage('Since doctrine/mongodb-odm 2.16: The method Type::closureToPHP() will change its default implementation in 3.0 to use convertToPHPValue(). Override this method if you need custom behavior before upgrading to 3.0 or use the trait ClosureToPHP to get the upcoming behavior now.');
```
IonBazan, line 99: @GromNaN How should we proceed with this deprecation message in v3.0?

GromNaN: We have the [`CaptureDeprecationMessages` trait](https://github.com/doctrine/mongodb-odm/blob/2.16.x/tests/Tests/CaptureDeprecationMessages.php) for testing deprecations as tests must run with PHPUnit 10 for PHP 8.1 compatibility, but in this version the error handler wasn't good.

IonBazan: This PR is targeting `3.0.x` so we are only supporting PHP 8.4+ with PHPUnit 12 and `CaptureDeprecationMessages` trait has been removed here but the deprecation message still remains in 3.0.x branch. Should it be removed?

GromNaN: Oh, yes. I was completely off topic. You have to change the implementation of this method in the Type class to use the one from the trait. And remove all uses of the trait (which will be deprecated later).

GromNaN: Oh, yes. I was completely off topic. You have to change the implementation of this method in the Type class to use the one from the trait. And remove all uses of the trait (which will be deprecated later).

GromNaN: I'll open an other PR to fix this.

GromNaN:
Thank you @IonBazan

PR #2982, opened 2026-01-16
Make `PersistentCollection` implement `Selectable`

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | Fix #2981

#### Summary

Implementing ReadableCollection without Selectable is deprecated https://github.com/doctrine/collections/pull/518

PR #2985, opened 2026-01-16
Drop support for `doctrine/collections` v1

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | yes
| Fixed issues | https://github.com/doctrine/mongodb-odm/pull/2982#discussion_r2699383799

#### Summary

doctrine/collections v1 can't be installed already because of the PHP version compatibility.

`doctrine/collections` 3.0 adds return types, this PR ensure we have the correct types documented.

- `Collection::add()` returns `void`, we currently return `true`. BC break will be done in mongodb-odm 3.0
- The `ArrayCollection::matching(): Collection` method is more restrictive than `Selectable::matching(): ReadableCollection`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
     "require": {
         "php": "^8.1",
         "ext-mongodb": "^1.21 || ^2.0",
-        "doctrine/collections": "^1.5 || ^2.0",
+        "doctrine/collections": "^2.1",
         "doctrine/event-manager": "^1.0 || ^2.0",
         "doctrine/instantiator": "^1.1 || ^2",
         "doctrine/persistence": "^3.2 || ^4",
--- a/src/PersistentCollection/PersistentCollectionTrait.php
+++ b/src/PersistentCollection/PersistentCollectionTrait.php
@@ -4,7 +4,6 @@ declare(strict_types=1);
 
 namespace Doctrine\ODM\MongoDB\PersistentCollection;
 
-use BadMethodCallException;
 use Closure;
 use Doctrine\Common\Collections\Collection as BaseCollection;
 use Doctrine\Common\Collections\Criteria;
@@ -26,7 +25,6 @@ use function array_values;
 use function count;
 use function get_class;
 use function is_object;
-use function method_exists;
 use function sprintf;
 
 /**
@@ -763,10 +761,6 @@ trait PersistentCollectionTrait
      */
     public function findFirst(Closure $p)
     {
-        if (! method_exists($this->coll, 'findFirst')) {
-            throw new BadMethodCallException('findFirst() is only available since doctrine/collections v2');
-        }
-
         return $this->coll->findFirst($p);
     }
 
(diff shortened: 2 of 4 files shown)


src/PersistentCollection/PersistentCollectionTrait.php
```diff
@@ -780,14 +788,21 @@ public function reduce(Closure $func, $initial = null)
...
         if (! $this->coll instanceof Selectable) {
             throw new LogicException('The backed collection must implement Selectable to use matching().');
         }
 
-        return $this->coll->matching($criteria);
+        $coll = $this->coll->matching($criteria);
+
+        if (! $coll instanceof BaseCollection) {
+            throw new LogicException(sprintf('The matching() method of the backed collection must return an instance of "%s".', BaseCollection::class));
+        }
```
GromNaN, line 794: The default implementation of `PersistableCollection` extends the `ArrayCollection` that has a more specific return type than `Selectable::matching(): ReadableCollection`
https://github.com/doctrine/collections/blob/a8353e2e0cb9db12ceb87c8e76114bd885b0170f/src/ArrayCollection.php#L386

src/PersistentCollection/PersistentCollectionTrait.php
```diff
@@ -414,6 +414,14 @@ public function set($key, $value)
         $this->doSet($key, $value, false);
     }
 
+    /**
+     * Adds an element at the end of the collection.
+     *
+     * @param mixed $element The element to add.
+     * @phpstan-param T $element
+     *
+     * @return true The return value is kept for BC reasons, but will be void in doctrine/mongodb-odm 3.0.
```
greg0ire, line 421: I think it might make sense to do a conditional bc break (based on the version of `doctrine/collections`), as in https://github.com/doctrine/orm/pull/12341/changes#diff-9634c9dcbc64c9326048aaa0df50b1a5dea4214b83b56ccb5ef676bf076841f6

composer.json
```diff
@@ -28,7 +28,7 @@
     "require": {
         "php": "^8.1",
         "ext-mongodb": "^1.21 || ^2.0",
-        "doctrine/collections": "^1.5 || ^2.0",
+        "doctrine/collections": "^2.1",
```
GromNaN, line 31: doctrine/collections 1 is already not compatible with PHP 8, so this changing nothing.

src/PersistentCollection/PersistentCollectionTrait.php
GromNaN: Thanks for the feedback.
I would have to put all the methods of the interface in the trait in order to have compatibility with both doctrine/collections 2 and 3.
I wonder if I should extend `AbstractLazyCollection` instead, and just use a conditional trait for the `add` method.

greg0ire: Maybe yes. I think `PersistentCollection` 49fb4e48a91a3d45b5437dbb46fb4d7ddf257fe4 was introduced 3 years prior to `AbstractLazyCollection` (https://github.com/doctrine/collections/pull/19)

Although ODM wasn't explicitly mentioned, I think it might have been hinted at in https://github.com/doctrine/orm/pull/882#issuecomment-30959228

GromNaN: I've open an issue, I want this PR to get merged without support for doctrine/collections v3: 
- https://github.com/doctrine/mongodb-odm/issues/2986

alcaeus (approved):

PR #2988, opened 2026-01-29
Fix ClassMetadata type for `SearchIndexMapping`

paulinevos:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

#### Summary
A `SearchIndexMapping` can have either `SearchIndexDefinition` or `VectorSearchIndexDefinition` for `definition`. Currently it is missing the latter.


GromNaN (approved):
Thanks.

We could add a test that uses this method so that PHPStan would validate the input array. But that's not a requirement.

PR #2992, opened 2026-03-04
Add Pagination classes to ODM

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | 

#### Summary

This PR introduces four new paginator classes. We introduce two styles of paginator: offset-based and cursor-based. Both paginator styles come in flavours for aggregation pipeline and queries.

##### Offset-based pagination

Offset-based paginators take a page number and number of items per page (defaults to 24), along with a builder for aggregation or queries. It is `Countable` and returns the number of pages when counted. It provides an iterator for the current page. To accomplish this, it uses the `count` operation for queries, and `$numDocuments` for aggregation pipeline. When fetching items, it inserts the correct `skip` and `limit` options/stages.

##### Cursor-based pagination

Cursor-based pagination takes an ID where to continue paginating, the number of items per page, and an optional field name for the cursoring field (an `ObjectId` value in the `_id` field is usually a good choice). Together with the builder for aggregation or queries you can iterate over the items of the current "page". Iterating the paginator multiple times always returns the same items. You are responsible for extracting the ID of the last item, which is used when creating the paginator for the next page.

##### Todo

- [ ] Write documentation
- [ ] Add API to get next page from a cursor-based paginator?

### DIFF
--- a/src/Pagination/AggregationOffsetPaginator.php
+++ b/src/Pagination/AggregationOffsetPaginator.php
@@ -1,0 +1,58 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ODM\MongoDB\Pagination;
+
+use Countable;
+use Doctrine\ODM\MongoDB\Aggregation\Builder;
+use Doctrine\ODM\MongoDB\Iterator\Iterator;
+
+use function ceil;
+use function iterator_to_array;
+
+/**
+ * @template-covariant TValue
+ * @template-extends Paginator<TValue>
+ */
+final class AggregationOffsetPaginator extends Paginator implements Countable
+{
+    private int $pageCount;
+
+    /** @psalm-param positive-int $page */
+    public function __construct(
+        private readonly Builder $aggregation,
+        public readonly int $page,
+        public readonly int $perPage = 24,
+    ) {
+    }
+
+    public function count(): int
+    {
+        return $this->pageCount ??= $this->getNumberOfPages();
+    }
+
+    private function getNumberOfPages(): int
+    {
+        $builder    = clone $this->aggregation;
+        $results    = $builder
+            ->hydrate(null)
+            ->count('numDocuments')
+            ->getAggregation()
+            ->getIterator();
+        $numResults = iterator_to_array($results)[0]['numDocuments'] ?? 0;
+
+        return (int) ceil($numResults / $this->perPage);
+    }
+
+    /** @return Iterator<TValue> */
+    protected function getResultsForCurrentPage(): Iterator
+    {
+        $builder = clone $this->aggregation;
+        $builder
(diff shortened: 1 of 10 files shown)


src/Pagination/AggregationOffsetPaginator.php
```diff
@@ -0,0 +1,58 @@
...
+    public function count(): int
+    {
+        return $this->pageCount ??= $this->getNumberOfPages();
+    }
+
+    private function getNumberOfPages(): int
+    {
+        $builder    = clone $this->aggregation;
+        $results    = $builder
+            ->hydrate(null)
```
GromNaN, line 39: What's the difference between `hydrate(null)` and `hydrate(false)` used in the query paginators?

src/Pagination/QueryOffsetPaginator.php
```diff
@@ -0,0 +1,55 @@
...
+
+    private function getNumberOfPages(): int
+    {
+        $builder = clone $this->query;
+
+        return (int) ceil($builder
+            ->hydrate(false)
+            ->count()
+            ->getQuery()
+            ->execute() / $this->perPage);
```
GromNaN, line 42: This is immediately forgetting the exact number of pages. I think it would be better to make this value accessible with a method such as `getNumberOfResults()`.

GromNaN (commented):
You decided not using the `$facet` stage, as [done in Api Platform](https://github.com/api-platform/core/blob/main/src/Doctrine/Odm/Paginator.php).

src/Pagination/AggregationOffsetPaginator.php
alcaeus: The type, probably nothing else. We should eventually have a `string|false` type to control this.

src/Pagination/QueryOffsetPaginator.php
alcaeus: We can add something like this. Since we're discussing this, how do you feel about the paginator count being the number of pages? I felt that made more sense than returning the number of results (or worse, the number of items on the current page). If we don't want to change the `Countable` behaviour, I would use something like `countResults` (or similar) and use that to return the number of results, storing them in a private property as well.

alcaeus: Following up on this. The logic is in the private method only called from `count()`, which caches the page count. Do you think it's relevant to store also the number of results?

GromNaN: As it's iterable, the count should be the number of items of the iteration. But this is distinct from Doctrine ORM which returns the total number of results for `count()`: https://github.com/doctrine/orm/blob/27c33cf88ddeeb4b0df8bc574f218cbb3d6f7eac/src/Tools/Pagination/Paginator.php#L92

PR #3000, opened 2026-03-26
PHPORM-254: Fix failing sharding tests

alcaeus:
#### Summary

Sharding tests failed with an error message similar to this:
> MongoDB\Driver\Exception\BulkWriteException: Transaction eb207f73-80ec-4bb0-8ed9-7e23da50315f - 47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU= -  - :249 was aborted on statement 2 due to: an error from cluster data placement change :: caused by :: Encountered error from localhost:27217 during a transaction :: caused by :: sharding status of collection doctrine_odm_tests.users is not currently known and needs to be recovered

This occurred if collections were implicitly created from within a transaction. Since this is not a typical use case, we can work around this by explicitly creating collections before flushing the document manager.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -110,14 +110,13 @@ jobs:
             proxy: "native"
             remove-optional-dependencies: true
           # Test with a sharded cluster
-          # Currently disabled due to a bug where MongoDB reports "sharding status unknown"
-#          - topology: "sharded_cluster"
-#            php-version: "8.2"
-#            mongodb-version: "6.0"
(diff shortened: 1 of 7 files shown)


GromNaN (commented):
Well done!

How did you decide which collections to create and which tests to include?

alcaeus:
> How did you decide which collections to create and which tests to include?

I looked through the failing tests and added all document classes. Didn't bother filtering them any further.

GromNaN (approved):

PR #3004, opened 2026-04-08
Fix wrong class metadata used if adding a reference using query builder

attrib:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | 

#### Summary

We noticed that when using a reference and then adding a complex expression afterwards. Inside the expression the class meta data of the reference object is used and not from the original document.

Added a tests to show this behaviour. Without the fix this will throw a fatal error `Object of class MongoDB\BSON\UTCDateTime could not be converted to int`

That we have two fields with the same name on two models, but with different types is kinda a fuck up on our side. But I think the ODM still should support this.

If you do not use the `$or` and do only a `$gte`, then it works without the fix. Or what we did in our case as a workaround, adding the equals after the expression.

The fix is small, but getting the test to work was some work.

### DIFF
--- a/src/Query/Expr.php
+++ b/src/Query/Expr.php
@@ -1450,6 +1450,7 @@ class Expr
 
         $convertedQuery = [];
         foreach ($query as $key => $value) {
+            unset($fieldMetadata);
             if (is_string($key) && $classMetadata->hasAssociation($key)) {
                 $targetDocument = $classMetadata->getAssociationTargetClass($key);
 
--- a/tests/Documents/Profile.php
+++ b/tests/Documents/Profile.php
@@ -4,8 +4,10 @@ declare(strict_types=1);
 
 namespace Documents;
 
+use DateTimeInterface;
 use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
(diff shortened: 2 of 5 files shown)


attrib:
Adapted failing test and fixed CS issues

src/Query/Expr.php
GromNaN, line 1464: Should we also unset the variable here?
```suggestion
                unset($fieldMetadata);
                continue;
```

GromNaN (commented):
Thank you very much for reporting the bug and for providing a fix.

I’m not very familiar with this code part, do you have any thoughts on the question below?

src/Query/Expr.php
malarzm: I think it'd be easier to unset it at the beginning of the loop :)

attrib: sounds about right. moved it to top of the loop

GromNaN (approved):
LGTM. It should be merged into 2.16.x as this is a bugfix.

SenseException (approved):

GromNaN:
Thank you @attrib

PR #3005, opened 2026-05-19
PHPORM-257: Consolidate per-collection writes into a single bulkWrite per commit phase

alcaeus:
[PHPORM-257](https://jira.mongodb.org/browse/PHPORM-257)

Part 1 of the bulkWrite refactor: every write against the same MongoDB collection that a single `flush()` produces is now consolidated into one `Collection::bulkWrite()` call per commit phase. Upserts, updates, deletes, and collection-level (`$push`, `$pull`, `$set`, `$addToSet`, …) writes that target the same parent collection all share that bulk — reducing the driver call count from one call per document to one per collection per phase. Inserts were already batched via `insertMany`; the structural change for inserts is that they now flow through the same per-collection `bulkWrite` path as all other write kinds rather than a separate `insertMany` call, so a flush that mixes inserts with updates or deletes on the same collection now produces a single command instead of two. Multi-collection consolidation via a future driver-level `bulkWriteCommand` is **out of scope** and will be addressed in a follow-up ticket.

## What changed

- New `Doctrine\ODM\MongoDB\Persisters\BulkWriteQueue` — a per-UnitOfWork accumulator keyed by `MongoDB\Collection` object identity. Both `DocumentPersister` and `CollectionPersister` push ops here; `UnitOfWork::doCommit()` flushes the queue at the end of each phase.
- `DocumentPersister::executeInserts()`, `executeUpserts()`, `update()` and `delete()` now build payloads and queue `insertOne` / `updateOne` / `deleteOne` ops onto the `BulkWriteQueue` instead of calling `insertMany` / `updateOne` / `deleteOne` on the collection directly.
- `CollectionPersister::executeQuery()` queues against the parent document's collection so embedded-collection writes and the parent's write share a single bulk.
- `UnitOfWork::doCommit()` runs each phase in three passes per collection: build & queue → `BulkWriteQueue::flush()` → dispatch `postPersist` / `postUpdate` / `postRemove`. Per-document `preUpdate` still fires before the queueing pass so listeners can still mutate the change set.
- Identifier / original-data cleanup for deletions moved from `executeDeletions()` to `dispatchPostRemoveEvents()`. Without this move, a `TransientTransactionError` retry from `withTransaction()` would have lost the identifiers needed to rebuild the delete filter on the second attempt.

## Trade-offs and deliberate omissions

- **Versioned and lockable documents bypass the consolidated bulk.** Each such document still gets its own one-op `bulkWrite` so `LockException` can be thrown precisely against the matched / modified / deleted count for that document. The MongoDB driver's `BulkWriteResult` aggregates counts across a batch, so per-op `LockException` for a versioned doc inside a multi-op batch would require post-hoc reconstruction. Once a server-side `bulkWriteCommand` exposes per-op write errors, this fence can be lifted.
- **Lifecycle event ordering is BC-different.** `postPersist`, `postUpdate` and `postRemove` now fire after the whole phase's bulkWrite, not interleaved with each document's write. Documented in `UPGRADE-3.0.md`.
- **Empty-modifier upsert retry has been inlined.** Where the 2.x persister relied on a `BulkWriteException` retry to handle "Mod on `_id` not allowed", we now detect the empty-modifier case at build time and emit `['$set' => new stdClass()]` directly. See `SERVER-12266`.
- **GridFS deletes still bypass the queue.** GridFS deletes touch two collections and are not expressible as a single bulkWrite on the parent collection. They flow through `Bucket::delete()` unchanged.
- **Multi-collection `bulkWriteCommand` consolidation** is explicitly deferred. A follow-up ticket will pick this up once the driver lands the server-side command.
- **Performance benchmarks.** A `BulkWritePerformanceTest` was added under `tests/Tests/Performance/`; it is excluded from the default suite via the `performance` group. The numbers should be tracked over time, but no regression gate is set in CI.

## Test plan

- [x] `vendor/bin/phpunit tests/Tests/Persisters/BulkWriteQueueTest.php` — 12 new unit tests covering op order, callback timing, per-collection grouping, exception propagation, and per-collection options forwarding.
- [x] `vendor/bin/phpunit tests/Tests/Functional/BulkWriteCommitTest.php` — 7 new functional tests asserting "one bulkWrite per (collection × phase)" via `CommandLogger`.
- [x] `vendor/bin/phpunit tests/Tests/Functional/BulkWriteLifecycleEventTest.php` — 2 new functional tests covering `preUpdate`/`postUpdate` ordering and `postPersist` non-firing on bulkWrite failure.
- [x] `vendor/bin/phpunit tests/Tests/UnitOfWorkCommitConsistencyTest.php tests/Tests/UnitOfWorkTransactionalCommitConsistencyTest.php` — the primary BC gate, both pass without source modification.
- [x] `vendor/bin/phpunit tests/Tests/Functional/LockTest.php tests/Tests/Functional/VersionTest.php` — pass unchanged.
- [x] `vendor/bin/phpunit tests/Tests/Functional/CommitImprovementTest.php` — pass unchanged.
- [x] `vendor/bin/phpunit tests/Tests/Functional/CollectionPersisterTest.php` — command-count assertions updated for the new consolidation; tests that exercise `CollectionPersister` directly now also drain the queue via a small `flushBulkWriteQueue()` helper.
- [x] `vendor/bin/phpunit tests/Tests/Functional/DocumentPersisterTest.php tests/Tests/UnitOfWorkTest.php` — write-concern / transactional-options mocks now expect `bulkWrite` instead of `insertMany` / `updateOne` / `deleteOne`.
- [x] `vendor/bin/phpunit --group performance tests/Tests/Performance/BulkWritePerformanceTest.php` — benchmarks run, no assertion failures.
- [x] Full suite: 3022 tests, 7 errors + 2 failures **all pre-existing on this branch** (Atlas Search, Vector Search, ReadPreference hint dataset #2, SchemaManager wait-for-search-index — all unrelated to this refactor and gated on Atlas-only features).
- [x] `vendor/bin/phpstan analyse` — no new errors introduced; the 25 baseline errors all predate this PR and relate to PHP 8.4+ Reflection methods not visible on 8.3.

## Notes for reviewers

- The order of post-events relative to *other documents'* writes has shifted from interleaved to "phase-completed first". Listeners that rely on the older order will need updating; this is documented in `UPGRADE-3.0.md`.
- The bulk-level options (writeConcern, session, …) are captured per-collection at the *first* op pushed and reused for every subsequent op on that collection in the phase. For mixed-class collections (inheritance), all classes resolve to the same per-class write concern via `ClassMetadata::getWriteConcern()` so this is safe in practice; reviewers should still sanity-check the assumption.
- The `CollectionPersister` direct-call tests now drain the queue via `flushBulkWriteQueue()`. This helper is purely a test convenience; production callers always drain via `UnitOfWork::doCommit()`.
- [ ] Consider extracting the empty-modifier upsert workaround into `PersistenceBuilder` so the persister logic stays focused on dispatch. Deferred — out of scope for this PR.

### DIFF
--- a/UPGRADE-3.0.md
+++ b/UPGRADE-3.0.md
@@ -57,3 +57,43 @@ to retain the functionality.
 
 `Doctrine\ODM\MongoDB\Event\OnClearEventArgs`' methods `getDocumentClass` and 
 `clearsAllDocuments` have been removed.
+
+## Flush now consolidates writes per collection via `bulkWrite`
+
+`UnitOfWork::commit()` previously issued an individual `insertMany`,
+`updateOne` or `deleteOne` call per scheduled document. As of 3.0, every
+scheduled write against the same MongoDB collection is now consolidated
+into a single `Collection::bulkWrite()` call per commit phase (upserts,
+inserts, updates, deletes).
+
+The visible effects:
+
+- **Lifecycle event ordering.** Within a phase, `postPersist`,
+  `postUpdate` and `postRemove` now fire **after** the entire phase's
+  `bulkWrite` has succeeded, rather than interleaved per document.
+  Listeners that relied on observing one document's post-event before
+  another document in the same phase had been written must be reviewed.
+  The per-document `preUpdate` callback still fires *before* the
+  bulkWrite, so listeners can still mutate the change set there.
+- **`LockException` semantics.** Versioned and lockable documents still
+  raise `LockException` with the same timing as before. Internally, the
+  refactor bypasses the consolidated bulk for these documents and issues
+  a single-op `bulkWrite` per such document so the matched / modified /
+  deleted counts can be inspected precisely.
+- **Partial-failure exceptions.** A failed write surfaces as
+  `MongoDB\Driver\Exception\BulkWriteException` (with `getWriteResult()`
+  describing every op in the batch), where the 2.x ODM might have
+  surfaced the same condition through the analogous error from
+  `insertMany` / `updateOne` / `deleteOne`. The exception class and
+  hierarchy are unchanged; only the aggregation of error info inside
+  the result is broader.
+- **Command count.** Tools and tests that count emitted MongoDB
+  commands (e.g. via `Doctrine\ODM\MongoDB\APM\CommandLogger`) will see
+  fewer commands per `flush()`. For example, persisting 100 documents
+  of the same class now emits one `insert` command instead of one per
+  document.
+
+The internal `DocumentPersister` and `CollectionPersister` classes are
+`@internal` and `final`; their public method names are unchanged but the
+queueing model now lives in `Doctrine\ODM\MongoDB\Persisters\BulkWriteQueue`.
+Applications must not extend these classes.
--- a/UPGRADE-3.0.md
+++ b/UPGRADE-3.0.md
@@ -60,11 +60,34 @@ to retain the functionality.
 
 ## Flush now consolidates writes per collection via `bulkWrite`
 
-`UnitOfWork::commit()` previously issued an individual `insertMany`,
-`updateOne` or `deleteOne` call per scheduled document. As of 3.0, every
-scheduled write against the same MongoDB collection is now consolidated
-into a single `Collection::bulkWrite()` call per commit phase (upserts,
-inserts, updates, deletes).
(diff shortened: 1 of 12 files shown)


alcaeus:
Closing -- I don't intend to continue on with this approach.

PR #3008, opened 2026-06-16
Fix CollectionPersister::delete() param type in test instead of baseline

### LINKED ISSUE #2796: CollectionPersister param type error
PHPStan reports a legitimate type error when passing a `Collection` to the `CollectionPersister`.

I added the error the baseline for the time being: https://github.com/doctrine/mongodb-odm/commit/f66e9e4038b2036a388dc33bcc9af985b04278f8

It seems that the `CollectionPersister` is only able to handle list of `PersistentCollectionInterface` while field properties are usually using `array|Collection` type.

GromNaN:
Fixes #2796.

The baselined PHPStan error came from the test passing chained document property accesses (typed `Collection|array`) to `CollectionPersister::delete()`, which requires `PersistentCollectionInterface[]`.

The signature is correct and the analysis should not be weakened: the only caller, `DocumentPersister::handleCollections()`, builds the list from `UnitOfWork::getScheduledCollections()`, which yields `PersistentCollection` instances. So the fix belongs in the test, not the production type.

The multi-collection `delete()` calls now route their arguments through an asserting helper that narrows `mixed` to `PersistentCollectionInterface`, and the baseline entry is removed.

### DIFF
--- a/tests/Tests/Functional/CollectionPersisterTest.php
+++ b/tests/Tests/Functional/CollectionPersisterTest.php
@@ -89,7 +89,7 @@ class CollectionPersisterTest extends BaseTestCase
         $this->logger->clear();
         $persister->delete(
             $user,
-            $this->persistentCollections([$user->categories[0]->children[0]->children, $user->categories[0]->children[1]->children]),
+            $this->assertPersistentCollections([$user->categories[0]->children[0]->children, $user->categories[0]->children[1]->children]),
             [],
         );
         self::assertCount(1, $this->logger, 'Deletion of several embedded-many collections of one document requires one query');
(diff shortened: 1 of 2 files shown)


tests/Tests/Functional/CollectionPersisterTest.php
```diff
@@ -135,11 +137,9 @@ public function testDeleteNestedEmbedManyAndNestedParent(): void
         $this->logger->clear();
         $firstCategoryChildren = $user->categories[0]->children;
         self::assertInstanceOf(PersistentCollectionInterface::class, $firstCategoryChildren);
```
malarzm: You could remove this one as well

tests/Tests/Functional/CollectionPersisterTest.php
```diff
@@ -268,6 +268,22 @@ private function getCollectionPersister(): CollectionPersister
...
     }
 
+    /**
+     * Asserts that each managed collection is a PersistentCollection and narrows the type for the persister.
+     *
+     * @param array<mixed> $collections
+     *
+     * @return list<PersistentCollectionInterface<array-key, object>>
+     */
+    private function persistentCollections(array $collections): array
```
malarzm: Maybe `assertPersistentCollections`?

malarzm (approved):

tests/Tests/Functional/CollectionPersisterTest.php
GromNaN: Good point. I've applied the suggested changes.

PR #3009, opened 2026-06-16
Deprecate external write access to read-only metadata properties

alcaeus:
This PR bumps the minimum PHP requirement to 8.4 in order to leverage property hooks to emit deprecation notices for properties when they are written from outside the `ClassMetadata` class. This will allow us to use asymmetric visibility (as in `public private(set)` for those properties in 3.0, codifying the previously documented read-only behaviour.

Note that this does not work for array properties, as `$foo->bar[]` doesn't go through the `set` hook, but rather reads the property by reference to replace it. Adding a `set` hook to those properties results in a warning about indirect property sets, which we don't want either. For that reason, I've left these properties as they are and will tackle them separately.

The `isView` property was previously private, so this PR deprecates the `isView()` method entirely and uses asymmetric visibility to make the property public for read access.

### DIFF
--- a/src/Mapping/ClassMetadata.php
+++ b/src/Mapping/ClassMetadata.php
@@ -846,9 +846,9 @@ use const PHP_VERSION_ID;
     }
 
     /**
-     * READ-ONLY: Whether this class describes the mapping of a database view.
+     * Whether this class describes the mapping of a database view.
      */
-    private bool $isView = false;
+    public private(set) bool $isView = false;
 
     /**
      * READ-ONLY: Whether this class describes the mapping of a gridFS file
@@ -1112,7 +1112,7 @@ use const PHP_VERSION_ID;
      */
     public function invokeLifecycleCallbacks(string $event, object $document, ?array $arguments = null): void
     {
-        if ($this->isView()) {
+        if ($this->isView) {
             return;
         }
 
@@ -2366,8 +2366,11 @@ use const PHP_VERSION_ID;
         return $this->rootClass;
     }
 
+    /** @deprecated Use the isView property instead */
     public function isView(): bool
     {
+        trigger_deprecation('doctrine/mongodb-odm', '2.18', 'The %s::isView() method is deprecated and will be removed in version 3.0. Use the isView property instead.');
+
         return $this->isView;
     }
 
--- a/src/Mapping/ClassMetadata.php
+++ b/src/Mapping/ClassMetadata.php
@@ -50,6 +50,7 @@ use function assert;
 use function class_exists;
 use function constant;
 use function count;
+use function debug_backtrace;
 use function enum_exists;
 use function extension_loaded;
 use function in_array;
@@ -63,6 +64,7 @@ use function strtolower;
 use function strtoupper;
 use function trigger_deprecation;
 
+use const DEBUG_BACKTRACE_IGNORE_ARGS;
 use const PHP_VERSION_ID;
 
 /**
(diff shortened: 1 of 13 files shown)


src/Mapping/ClassMetadata.php
```diff
@@ -490,76 +492,99 @@
...
+    // phpcs:disable PSR2.Classes.PropertyDeclaration.ScopeMissing
     /**
      * READ-ONLY: The name of the mongo database the document is mapped to.
      *
      * @var string|null
      */
-    public $db;
+    public $db {
+        set => $this->setPropertyValue('db', $value);
+    }
```
GromNaN, line 505: You can use a hack to only trigger this deprecation notices with PHP 8.4+. Create a trait with 2 implementations: one of PHP 8.4+ using hooks, and the other for older PHP versions. The traite is loaded conditionally depending on the PHP version.

```php
if (PHP_VERSION_ID >= 80400) {
    require_once __DIR__.'/ClassMetadataHookedPropertiesTrait.php');
} else {
    require_once __DIR__.'/ClassMetadataLegacyPropertiesTrait.php');
}

class ClassMetadata {
    use ClassMetadataPropertiesTrait;
}
```

GromNaN (changes requested):
Could you benchmark the performance impact of using hooks?

> This PR bumps the minimum PHP requirement to 8.4

There is still 17% of installs with PHP 8.1-8.3. I don't want to leave them with unmaintained versions of the ODM.

src/Mapping/ClassMetadata.php
GromNaN: You don't need to trigger the deprecation for PHP < 8.4, because MongoDB ODM 3.0 will require at lease PHP 8.4.

alcaeus:
> Could you benchmark the performance impact of using hooks?

Here's a quick benchmark for the raw property access I threw together:
```
    benchSetUnhookedProperty................I4 - Mo0.200μs (±6.58%)
    benchSetGetHookedProperty...............I4 - Mo0.203μs (±7.55%)
    benchSetHookedProperty..................I4 - Mo0.512μs (±3.94%)
    benchSetHookedPropertyWithMethod........I4 - Mo0.715μs (±3.33%)
```

All benchmarks set a value on a property -- the first two are unhooked and only with a `get` hook. The third property has a `set` hook, whereas the last property contains a `set` hook that invokes a method call. So there is a performance penalty with this, although caching should make this penalty negligible in production. I'll work on metadata loading benchmarks once we've settled on a way forward here.

> There is still 17% of installs with PHP 8.1-8.3. I don't want to leave them with unmaintained versions of the ODM.

Sure. Note there's also one case of asymmetric visibility for the (currently private) `isView` property which allows us to deprecate the `isView()` method. We can either make the property public in general (opening up the write loophole we've had with all other properties) and unconditionally deprecate `isView()`, or we can go the split approach and only deprecate in 8.4. What would you prefer?

src/Mapping/ClassMetadataHookedPropertiesTrait.php
```diff
@@ -0,0 +1,370 @@
...
+
+declare(strict_types=1);
+
+namespace Doctrine\ODM\MongoDB\Mapping;
+
+use Doctrine\ODM\MongoDB\Id\IdGenerator;
+use Doctrine\ODM\MongoDB\Mapping\Attribute\TimeSeries;
+use ReflectionClass;
+
+trait ClassMetadataPropertiesTrait
```
GromNaN, line 17: The trait must be flagged as `@internal`.

GromNaN (commented):
LGTM once the `@internal` annotation is added.

GromNaN (approved):

bronsonstowers-ai (commented):
✅

bronsonstowers-ai (commented):
✅

PR #3010, opened 2026-06-16
Fix excludeSubPaths dropping sibling collections with shared prefix

### LINKED ISSUE #2754: CollectionPersister::excludeSubPaths incorrectly filters sibling EmbedMany collections with shared prefix names, preventing persistence
### Bug Report

|    Q               |   A
|------------ | ------
| BC Break     | no
| Version        | 2.8.2

#### Summary

Persisting changes to multiple embedded collections within the same embedded document fails for collections whose field names share a common prefix. The collection with the longer name that starts with the shorter name is ignored during persistence due to incorrect path filtering in `CollectionPersister::excludeSubPaths`.

#### Current behavior

When an embedded document (e.g., `VariantVariant` embedded within `Product`) contains two or more `EmbedMany` collections whose field names share a common prefix (e.g., `supplier_factory_codes` and `supplier_factory_codes_v2`), and both collections are modified (e.g., by replacing their instances) within the same UnitOfWork cycle, only the changes to the collection with the shorter field name prefix (e.g., `supplier_factory_codes`) are persisted to MongoDB upon calling `$documentManager->flush()`. Changes to the collection with the longer, prefixed name (e.g., `supplier_factory_codes_v2`) are silently ignored.

This behavior was also observed with other collection pairs like `options` and `optionsList` within the same embedded document.

#### How to reproduce

1.  **Define Mappings:**
    * Create a top-level document mapping (e.g., `Product`) that embeds a collection of documents (e.g., `variants` embedding `VariantVariant`).
        ```xml
        <mapped-superclass name="Product" ...>
            ...
            <embed-many field="variants" target-document="VariantVariant"
                        collection-class="VariantVariants" />
            ...
        </mapped-superclass>
        ```
    * Define the embedded document mapping (`VariantVariant`) containing at least two `EmbedMany` collections where one field name is a prefix of the other.
        ```xml
        <embedded-document name="VariantVariant" >
            ...
            <embed-many field="supplier_factory_codes" field-name="supplierFactoryCodes" target-document="SupplierFactoryCode"
                        collection-class="SupplierFactoryCodes" />

            <embed-many field="supplier_factory_codes_v2" field-name="supplierFactoryCodesV2" target-document="SupplierFactoryCode"
                        collection-class="SupplierFactoryCodes" />
            ...
        </embedded-document>
        ```

2.  **Execute Code:**
    * Load a `Product` entity containing at least one `VariantVariant`.
    * Get a specific `VariantVariant` instance from the `variants` collection.
    * Replace the instance of *both* collections within that `VariantVariant`. Example:
        ```php
        $variant = $product->getVariants()->first(); // Get an embedded variant
        $newCodes = new SupplierFactoryCodes(/* ... some data ... */);
        $newCodesV2 = new SupplierFactoryCodes(/* ... some different data ... */);

        // Replace both collection instances
        $variant->setSupplierFactoryCodes($newCodes);
        $variant->setSupplierFactoryCodesV2($newCodesV2);

        // Persist changes
        $documentManager->flush();
        ```

3.  **Observe Result:** Check the MongoDB document for the updated variant. Only the `supplierFactoryCodes` field will reflect the changes from `$newCodes`. The `supplierFactoryCodesV2` field will remain unchanged (or as it was before the flush).

4.  **Identify Bug Location:** The issue stems from `Doctrine\ODM\MongoDB\Persisters\CollectionPersister::excludeSubPaths`.
    * This method receives the paths to be persisted, e.g., `['variants.0.supplier_factory_codes', 'variants.0.supplier_factory_codes_v2']` (simplified example).
    * It sorts the paths: `['variants.0.supplier_factory_codes', 'variants.0.supplier_factory_codes_v2']`.
    * It iterates and compares the current path with the last added path using `strpos($paths[$i], $lastUniquePath) === 0`.
    * In the example, `$lastUniquePath` is `'variants.0.supplier_factory_codes'`.
    * When `$paths[$i]` is `'variants.0.supplier_factory_codes_v2'`, `strpos` returns `0` because the string starts with the `$lastUniquePath`.
    * The `continue;` statement is executed, incorrectly discarding the path `'variants.0.supplier_factory_codes_v2'` as if it were a sub-path of the first, thus preventing its persistence.

#### Expected behavior

All modified embedded collections within the same embedded document should be persisted correctly during `flush()`, regardless of whether their field names share common prefixes.

The `CollectionPersister::excludeSubPaths` method should correctly differentiate between true sub-document paths (e.g., `address.street` being a sub-path of `address`) and sibling fields that happen to share a prefix (e.g., `supplierFactoryCodes` and `supplierFactoryCodesV2`).

#### Images

Image

GromNaN:
Fixes #2754.

`CollectionPersister::excludeSubPaths()` removes nested sub-paths from a batched collection update so that a parent path and its sub-path are never updated in the same query (which MongoDB rejects). It used a raw string-prefix test:

```php
if (strpos($paths[$i], $lastUniquePath) === 0) {
```

For sibling `embedMany` fields whose names share a prefix (e.g. `supplier_factory_codes` and `supplier_factory_codes_v2`, or `codes` and `codesV2`), the longer path starts with the shorter one, so it was wrongly treated as a sub-path, excluded from the update, and **never persisted**.

The fix only treats a path as a sub-path when it equals the prefix or is followed by the `.` separator:

```php
if ($paths[$i] === $lastUniquePath || str_starts_with($paths[$i], $lastUniquePath . '.')) {
```

As a side effect, sibling collections using the `pushAll` strategy are now correctly batched into a single query instead of two. The existing `pushAll` test was updated to reflect this and extended to keep covering the genuine parent/sub-path case, which must still split into two queries.

A regression test (`GH2754Test`) replaces two sibling embed-many collections with a shared name prefix in one flush and asserts both are persisted.

### DIFF
--- a/tests/Tests/Functional/Ticket/GH2754Test.php
+++ b/tests/Tests/Functional/Ticket/GH2754Test.php
@@ -13,64 +13,76 @@ class GH2754Test extends BaseTestCase
 {
     public function testSiblingCollectionsWithSharedPrefixArePersisted(): void
     {
-        $document           = new GH2754Document();
-        $document->embedded = new GH2754Embedded();
-        $document->embedded->codes->add(new GH2754Item('a'));
-        $document->embedded->codesV2->add(new GH2754Item('b'));
-        $this->dm->persist($document);
+        $user          = new GH2754User();
+        $user->profile = new GH2754Profile();
+        $user->profile->friends->add(new GH2754Friend('alice'));
+        $user->profile->friendsPending->add(new GH2754FriendRequest('bob'));
+        $this->dm->persist($user);
         $this->dm->flush();
         $this->dm->clear();
 
-        $document = $this->dm->find(GH2754Document::class, $document->id);
-        self::assertNotNull($document);
+        $user = $this->dm->find(GH2754User::class, $user->id);
+        self::assertNotNull($user);
 
-        // Replace both sibling collections within the same flush. The field name "codes"
-        // is a prefix of "codesV2", which previously caused the latter to be dropped.
-        $document->embedded->codes   = new ArrayCollection([new GH2754Item('c')]);
-        $document->embedded->codesV2 = new ArrayCollection([new GH2754Item('d'), new GH2754Item('e')]);
+        // Replace both sibling collections within the same flush. The field name "friends"
+        // is a prefix of "friendsPending", which previously caused the latter to be dropped.
+        $user->profile->friends        = new ArrayCollection([new GH2754Friend('carol')]);
+        $user->profile->friendsPending = new ArrayCollection([new GH2754FriendRequest('dave'), new GH2754FriendRequest('erin')]);
         $this->dm->flush();
         $this->dm->clear();
 
-        $document = $this->dm->find(GH2754Document::class, $document->id);
-        self::assertNotNull($document);
-        self::assertCount(1, $document->embedded->codes);
-        self::assertCount(2, $document->embedded->codesV2, 'Sibling collection with prefixed name must be persisted');
-        self::assertSame('d', $document->embedded->codesV2[0]->name);
-        self::assertSame('e', $document->embedded->codesV2[1]->name);
+        $user = $this->dm->find(GH2754User::class, $user->id);
+        self::assertNotNull($user);
+        self::assertCount(1, $user->profile->friends);
+        self::assertCount(2, $user->profile->friendsPending, 'Sibling collection with prefixed name must be persisted');
+        self::assertSame('dave', $user->profile->friendsPending[0]->name);
+        self::assertSame('erin', $user->profile->friendsPending[1]->name);
     }
 }
 
 #[ODM\Document]
-class GH2754Document
+class GH2754User
 {
     /** @var string|null */
     #[ODM\Id]
     public $id;
 
-    #[ODM\EmbedOne(targetDocument: GH2754Embedded::class)]
-    public ?GH2754Embedded $embedded = null;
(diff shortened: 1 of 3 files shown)


tests/Tests/Functional/Ticket/GH2754Test.php
```diff
@@ -0,0 +1,82 @@
...
+        self::assertNotNull($document);
+        self::assertCount(1, $document->embedded->codes);
+        self::assertCount(2, $document->embedded->codesV2, 'Sibling collection with prefixed name must be persisted');
+        self::assertSame('d', $document->embedded->codesV2[0]->name);
+        self::assertSame('e', $document->embedded->codesV2[1]->name);
+    }
+}
+
+#[ODM\Document]
+class GH2754Document
```
paulinevos: My only gripe with this test is because of the naming of the documents it's hard to discern which one is which, making it difficult to understand the original edge case at first glance. Would it be possible to rename the documents to something that might be a real-world scenario? 

Maybe something like:
```
User
|___
       Profile
       |____
            friends: EmbedMany(Friend) 
            friendRequests: EmbedMany(FriendRequest)
```

or something like that

GromNaN: Done, great call — the User/Profile/Friend naming makes the scenario much clearer. I kept `friends`/`friendsPending` rather than `friendRequests` so that one field name stays a literal prefix of the other, which is what triggers the bug.

paulinevos (approved):

PR #3011, opened 2026-06-16
Deprecate COLLECTION_PER_CLASS inheritance type

### LINKED ISSUE #2742: Deprecate `COLLECTION_PER_CLASS` inheritance
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | no
| RFC         | yes
| BC Break    | yes

#### Summary

The the `COLLECTION_PER_CLASS` inheritance means there is 1 collection per Document class. Which is not different from how all the Document classes works.

Getting the example from https://github.com/doctrine/mongodb-odm/pull/2726

```php
#[ODM\InheritanceType('COLLECTION_PER_CLASS')]
class Section {}

class Tag extends Section {}
```

If you run a query on the `Section` repository, the `section` collection is queried. There is no way to get a `Tag` or `Category` object in this way.

If you run a query on the `Tag` repository, the `tag` collection is queried. There is no way to get a `Category` object in this way. The discriminator field is not necessary.

This feature has an issue as we cannot guarantee that 2 documents doesn't share the same `_id`, as they are in distinct collections. https://github.com/doctrine/mongodb-odm/pull/2725#discussion_r1940947597

This feature has almost no documentation: https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/2.9/reference/inheritance-mapping.html#collection-per-class-inheritance

GromNaN:
Fixes #2742.

The `COLLECTION_PER_CLASS` inheritance type persists one collection per class, which is already the default behavior for any document. It provides no additional behavior, has almost no documentation, and cannot guarantee `_id` uniqueness across the distinct collections of a hierarchy ([context](https://github.com/doctrine/mongodb-odm/pull/2725#discussion_r1940947597)).

Deprecated with no replacement. The migration is to remove the `InheritanceType` mapping; each class remains mapped to its own collection.

- Runtime deprecation in `ClassMetadataFactory::loadMetadata()` (modeled on the `NOTIFY` deprecation)
- `@deprecated` on `ClassMetadata::INHERITANCE_TYPE_COLLECTION_PER_CLASS` and `isInheritanceTypeCollectionPerClass()`
- Deprecation note in the XSD enumeration
- `UPGRADE-2.17.md` and docs notes
- Migrated `User`/`VersionedUser` fixtures off `COLLECTION_PER_CLASS`; `#[IgnoreDeprecations]` on the remaining intentional usages

### DIFF
--- a/src/Mapping/ClassMetadata.php
+++ b/src/Mapping/ClassMetadata.php
@@ -2128,6 +2128,8 @@ use const PHP_VERSION_ID;
      */
     public function isInheritanceTypeCollectionPerClass(): bool
     {
+        trigger_deprecation('doctrine/mongodb-odm', '2.17', 'The method %s() is deprecated with no replacement.', __FUNCTION__);
+
         return $this->inheritanceType === self::INHERITANCE_TYPE_COLLECTION_PER_CLASS;
     }
 
--- a/src/Mapping/ClassMetadata.php
+++ b/src/Mapping/ClassMetadata.php
@@ -429,6 +429,8 @@ use const PHP_VERSION_ID;
     /**
      * COLLECTION_PER_CLASS means the class will be persisted according to the rules
      * of <tt>Concrete Collection Inheritance</tt>.
+     *
+     * @deprecated since 2.17 with no replacement. Each class is already mapped to its own collection.
      */
     public const INHERITANCE_TYPE_COLLECTION_PER_CLASS = 3;
 
@@ -2121,6 +2123,8 @@ use const PHP_VERSION_ID;
 
     /**
      * Checks whether the mapped class uses the COLLECTION_PER_CLASS inheritance mapping strategy.
+     *
+     * @deprecated since 2.17 with no replacement.
      */
     public function isInheritanceTypeCollectionPerClass(): bool
     {
--- a/tests/Documents/VersionedUser.php
+++ b/tests/Documents/VersionedUser.php
@@ -7,7 +7,6 @@ namespace Documents;
 use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
 
 #[ODM\Document(collection: 'users')]
-#[ODM\InheritanceType('COLLECTION_PER_CLASS')]
 class VersionedUser extends User
 {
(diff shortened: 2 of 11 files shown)


alcaeus (approved):

IonBazan (approved):

src/Mapping/ClassMetadata.php
```diff
@@ -2121,6 +2123,8 @@ public function isInheritanceTypeSingleCollection(): bool
 
     /**
      * Checks whether the mapped class uses the COLLECTION_PER_CLASS inheritance mapping strategy.
+     *
+     * @deprecated since 2.17 with no replacement.
```
malarzm, line 2127: `trigger_deprecation` is missing

tests/Documents/VersionedUser.php
```diff
@@ -7,7 +7,6 @@
 use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
 
 #[ODM\Document(collection: 'users')]
-#[ODM\InheritanceType('COLLECTION_PER_CLASS')]
```
malarzm, line 10: I think that in the past we were discouraging extending an already mapped document (you should extend a mapped superclass). I guess it's ok now given no test broke but I'm leaving this in case it'll ring someone else's bell too

src/Mapping/ClassMetadata.php
GromNaN: Done. Also refactored `ClassMetadataFactory` to compare against the constant directly to avoid calling the deprecated method internally.

tests/Documents/VersionedUser.php
GromNaN: I think this constraint comes from the XML/YAML mapping era: each class needed its own external file, so the parent had to be explicitly declared as a `MappedSuperclass` so the driver knew to load it.

With PHP attributes, `ReflectionClass::getProperties()` returns all properties including inherited ones, so the driver reads their ODM attributes regardless of which class declared them. A plain abstract (or concrete) parent class without `#[ODM\MappedSuperclass]` works fine — no separate file to load.

PR #3012, opened 2026-06-18
Remove COLLECTION_PER_CLASS inheritance type

GromNaN:
Removes the `COLLECTION_PER_CLASS` inheritance type, deprecated in 2.17 (#3011, issue #2742). This branch merges the 2.17 deprecation then removes the feature, to assess whether the removal is feasible for 3.0.

Every document class is already mapped to its own collection, so this inheritance type provided no additional behavior and could not guarantee `_id` uniqueness across the distinct collections of a hierarchy.

`ClassMetadata::INHERITANCE_TYPE_COLLECTION_PER_CLASS` and `ClassMetadata::isInheritanceTypeCollectionPerClass()` are removed. The mapping drivers now throw a `MappingException` for an unknown inheritance type instead of relying on an undefined constant.

Migration: remove the `InheritanceType` attribute/annotation (or the `inheritance-type` XML attribute) from the affected classes. Each class remains mapped to its own collection.


GromNaN:
Replaced by #3022

Issue #2986, opened 2026-01-22
Add support for `doctrine/collections` v3


GromNaN:
Doctrine Collections v3 adds return types, which is a breaking change for our PersistentCollection implementation.

We need to support both doctrine/collection v2 and v3 simultaneously in order to ensure a smooth upgrade path.

2 possibles approaches:
1. Create a conditional trait that declare all the methods from the interface. Usually, adding return types is a breaking change, but when this return types are added conditionally based on the major version of an other package, this is acceptable.
2. Leverage the `AbstractLazyCollection` to replace the `PersistentCollectionTrait`; abstract class already define the methods with return types depending on the version. But after some investigations, we still need to override these methods to support change detection (call the `->changed()` method).

GromNaN:
May be a duplicate of:
- https://github.com/doctrine/mongodb-odm/issues/2850

GromNaN:
Done in https://github.com/doctrine/mongodb-odm/pull/3014

PR #3014, opened 2026-06-18
Add compatibility with doctrine/collections v3, add SortDirection support in Query Builder

IonBazan:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #2986 

#### Summary

  - Expands `doctrine/collections` constraint from `^2.1` to `^2.1 || ^3.1`
  - Adds `PersistentCollectionCompatibility` trait (renamed from `PersistentCollectionImplementationTrait`) that conditionally defines `add()` with `bool` return (v2) or `void` return (v3) using a runtime
  version check via `defined(Criteria::class . '::ASC')`
  - Adds explicit PHP type declarations to all methods in `PersistentCollectionTrait` and `PersistentCollectionInterface` to satisfy v3's fully-typed `Collection`/`ReadableCollection`/`ArrayAccess`
  interfaces; removes now-redundant `#[ReturnTypeWillChange]` attributes
  - Fixes `QueryExpressionVisitor::walkValue()` missing `: mixed` return type (caught by the same type-declaration pass)

  ## Notable design decisions

  `add()` is the only method whose return type changed between v2 (`bool`) and v3 (`void`); all other methods were untyped in v2 and only gained types in v3. The conditional-trait approach isolates that
  single incompatibility. `filter()` and `map()` declare `): Collection` as their return type — identical to how `AbstractLazyCollection` handles it in doctrine/collections itself.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
     "require": {
         "php": "^8.1",
         "ext-mongodb": "^1.21 || ^2.0",
-        "doctrine/collections": "^2.1",
+        "doctrine/collections": "^2.1 || ^3.0",
         "doctrine/event-manager": "^1.0 || ^2.0",
         "doctrine/instantiator": "^1.1 || ^2",
         "doctrine/persistence": "^3.2 || ^4",
--- a/composer.json
+++ b/composer.json
@@ -39,6 +39,7 @@
         "symfony/console": "^5.4 || ^6.4 || ^7.0 || ^8.0",
         "symfony/deprecation-contracts": "^2.2 || ^3.0",
         "symfony/polyfill-php84": "^1.33",
+        "symfony/polyfill-php86": "^1.37",
         "symfony/var-dumper": "^5.4 || ^6.4 || ^7.0 || ^8.0",
         "symfony/var-exporter": "^6.4.1 || ^7.0 || ^8.0"
     },
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
     "require": {
         "php": "^8.1",
         "ext-mongodb": "^1.21 || ^2.0",
-        "doctrine/collections": "^2.1 || ^3.0",
+        "doctrine/collections": "^2.1 || ^3.1",
         "doctrine/event-manager": "^1.0 || ^2.0",
         "doctrine/instantiator": "^1.1 || ^2",
         "doctrine/persistence": "^3.2 || ^4",
--- a/src/PersistentCollection/PersistentCollectionCompatibility.php
+++ b/src/PersistentCollection/PersistentCollectionCompatibility.php
@@ -7,43 +7,304 @@ namespace Doctrine\ODM\MongoDB\PersistentCollection;
 use Closure;
 use Doctrine\Common\Collections\Collection as BaseCollection;
 use Doctrine\Common\Collections\Criteria;
-use Doctrine\Common\Collections\Selectable;
-use Doctrine\ODM\MongoDB\DocumentManager;
-use Doctrine\ODM\MongoDB\Mapping\ClassMetadata;
-use Doctrine\ODM\MongoDB\MongoDBException;
-use Doctrine\ODM\MongoDB\UnitOfWork;
-use Doctrine\ODM\MongoDB\Utility\CollectionHelper;
-use LogicException;
-use ReturnTypeWillChange;
 use Traversable;
 
-use function array_combine;
-use function array_diff_key;
-use function array_map;
-use function array_udiff_assoc;
-use function array_values;
-use function count;
-use function get_class;
-use function is_object;
-use function sprintf;
-
 /**
- * Compatibility trait for PersistentCollection between collections 2 and 3.
+ * Compatibility trait for PersistentCollection between doctrine/collections v2 and v3.
+ *
+ * The `add()` return type changed from bool (v2) to void (v3). All other Collection/ReadableCollection
+ * methods gained type declarations in v3. Both branches are duplicated here so that the version in
+ * use does not impose new type constraints on classes that extend PersistentCollection.
+ *
+ * @internal
  * @template TKey of array-key
  * @template T of object
  */
-if (defined(Criteria::class . '::ASC')) {
-    // collections 2
-    /** @internal */
+if (! defined(Criteria::class . '::ASC')) {
+    // collections 3 — full type declarations required by the interface
+    /**
+     * @internal
+     * @template TKey of array-key
+     * @template T of object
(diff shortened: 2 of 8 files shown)


src/PersistentCollection/PersistentCollectionCompatibility.php
```diff
@@ -0,0 +1,75 @@
...
+
+            return true;
+        }
+    }
+} else {
+    // collections 3
+    /** @internal */
+    trait PersistentCollectionCompatibility
+    {
+        abstract private function doAdd(mixed $value, bool $arrayAccess): bool;
```
GromNaN: `private` methods of an trait are part of its API. This new method is exposed and removing it will be a BC break.

src/PersistentCollection/PersistentCollectionInterface.php
```diff
@@ -28,122 +28,103 @@ interface PersistentCollectionInterface extends Collection, Selectable
 {
     /**
      * Sets the document manager and unit of work (used during merge operations).
-     *
-     * @return void
      */
-    public function setDocumentManager(DocumentManager $dm);
+    public function setDocumentManager(DocumentManager $dm): void;
```
GromNaN: Adding the return type to the interface is a breaking change. Implementations may not use the trait (they can use decoration). This types are added in 3.0 only by https://github.com/doctrine/mongodb-odm/pull/2939

src/PersistentCollection/PersistentCollectionTrait.php
```diff
@@ -369,115 +370,98 @@ public function exists(Closure $p)
...
-    public function getIterator()
+    public function getIterator(): Traversable
     {
         $this->initialize();
 
         return $this->coll->getIterator();
     }
 
-    public function map(Closure $func)
+    public function map(Closure $func): BaseCollection
```
GromNaN: This is a breaking change: the class that use this trait can be extended and the methods overridden without return type.

GromNaN (changes requested):
Given the BC issues highlighted below, I think the best course of action is to duplicate all the methods from the doctrine/collections interfaces into the compatibility trait.

src/PersistentCollection/PersistentCollectionCompatibility.php
```diff
@@ -0,0 +1,582 @@
...
+            $this->doSet($offset, $value, true);
+        }
+
+        public function offsetUnset(mixed $offset): void
+        {
+            $this->doRemove($offset, true);
+        }
+    }
+} else {
+    // collections 2 — no doctrine/collections type declarations
```
GromNaN:
```suggestion
    // doctrine/collections 2 — no type declarations
```

src/PersistentCollection/PersistentCollectionTrait.php
```diff
@@ -99,33 +104,33 @@ trait PersistentCollectionTrait
      */
     private array $hints = [];
 
-    public function setDocumentManager(DocumentManager $dm)
+    public function setDocumentManager(DocumentManager $dm): void
```
GromNaN: This new return types are not necessary for doctrine/collection v3 compatibility. They are a breaking change that we do in doctrine/mongodb-odm v3 only.

src/PersistentCollection/PersistentCollectionTrait.php
```diff
@@ -649,15 +354,44 @@ public function __clone()
...
     /**
      * Actual logic for adding an element to the collection.
      *
-     * @param mixed $value
-     * @param bool  $arrayAccess
-     *
      * @return true
      */
-    private function doAdd($value, $arrayAccess)
+    protected function doAdd(mixed $value, bool $arrayAccess): bool
```
GromNaN: I don't think you have to change this visiblity: the methods from the trait can call private methods.

GromNaN (changes requested):

src/PersistentCollection/PersistentCollectionCompatibility.php
```diff
@@ -0,0 +1,586 @@
...
+/**
+ * Compatibility trait for PersistentCollection between doctrine/collections v2 and v3.
+ *
+ * @internal
+ *
+ * @template TKey of array-key
+ * @template T of object
+ */
+if (! defined(Criteria::class . '::ASC')) {
+    // doctrine/collections 3 — full type declarations required by the interface
```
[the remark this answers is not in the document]
GromNaN: The suggested code is more accurate but less efficient. Let's ignore this suggestion.

composer.json
```diff
@@ -28,7 +28,7 @@
     "require": {
         "php": "^8.1",
         "ext-mongodb": "^1.21 || ^2.0",
-        "doctrine/collections": "^2.1",
+        "doctrine/collections": "^2.1 || ^3.0",
```
GromNaN: The [PHPUnit (lowest, 8.1, 6.0, 1.21.0, server, lazy-ghost)](https://github.com/doctrine/mongodb-odm/actions/runs/28035223713/job/82986610621?pr=3014#logs) job tests `doctrine/collections (2.2.0)`, which is fine. Because "doctrine/orm 3.7.x-dev requires doctrine/collections (^2.2 || ^3.1) "

GromNaN (approved):
The issue reported by PHPStan is unrelated.

SenseException (approved):

IonBazan:
I believe the issue reported by PHPStan is valid. I will need to port these changes too: https://github.com/doctrine/orm/pull/12449. That's because doctrine/collections support `SortDirection` since 3.1. Let me fix that later today.

GromNaN:
You can mark compatibility with doctrine/collections < 3.1 for now

src/Query/Builder.php
```diff
@@ -1474,16 +1475,18 @@ public function snapshot(bool $bool = true): self
...
     public function sort($fieldName, $order = 1): self
     {
         $this->query['sort'] ??= [];
         $fields                = is_array($fieldName) ? $fieldName : [$fieldName => $order];
 
         foreach ($fields as $fieldName => $order) {
-            if (is_string($order)) {
+            if ($order instanceof SortDirection) {
+                $order = $order === SortDirection::Ascending ? 1 : -1;
+            } elseif (is_string($order)) {
```
IonBazan, line 1489: I decided to only change Query Builder for now to handle collections v3.1 logic but full `SortDirection` support will require a separate PR and changes in other places too.

GromNaN (approved):
Thank you for the changes made to ensure BC, and for adding support for SortDirection in the query builder.

Issue #1482, opened 2016-08-03
Change detection doesn't work with objects and custom types


MidnightDesign:
I've got a property `$foo` on my document that holds a mutable object. That object is converted to an array by a custom type when persisting the document. Unfortunately, when I change a property on that mutable object, `$foo` isn't recognized as having changed because of [this condition](https://github.com/doctrine/mongodb-odm/blob/master/lib/Doctrine/ODM/MongoDB/UnitOfWork.php#L731-L741):

```php
if ($orgValue === $actualValue) {
    // ...
   continue;
    // ...
}
```

It's still the same instance, and so the contition evaluates to `true`.

Is there a better way to detect changes? Maybe something like comparing two `spl_object_hash()`es?

malarzm:
`spl_object_hash()` gave us enough headache with embedded documents since they can be reused, I'm not sure I would like to follow that way... As a temporary solution I'd recommend using immutable objects while we try to figure out a way to fix the bug.

MidnightDesign:
Thanks, I already changed it to be immutable.

malarzm:
Just to update, so far I have no idea how this could be introduced in a bugfix manner... 

Offhand I recalled there was #976 (but that'd be feature) and also I have an idea for 2.0 that could fix the issue at hand, namely I'm pondering over using database values for preparing change sets.

wesnick:
Hi!, happy 1 year anniversary for this issue.  This particular issue has long vexed me in one of my projects.  We use value objects extensively and in our world a value object with the same values should evaluate to be equal to a different value object with the same values.

I think ODM should make use of an interface offered in Common that mimics an interface never accepted into PHP core: https://github.com/doctrine/common/blob/master/lib/Doctrine/Common/Comparable.php

This code block could check if $orgValue implements Comparable, if so, use the compareTo($other) method.  If not, default to === check.

malarzm:
This is actually interesting, I had no idea such interface exists :) Given our current plans for 2.0 it indeed might be feasible to add as a feature earlier. @alcaeus @jmikola any downsides coming to your mind? From my perspective it can get into our way later if we'd like to remove the `Comparable` approach.

jmikola:
The Comparable interface looks like a good option here. I recall @guilhermeblanco trying to get this into core some time ago, but I wasn't aware it was added to Doctrine Common. Let's confirm if ORM is using it to address the same use case here.

alcaeus:
[...]

PR #3017, opened 2026-06-22
Allow custom type objects to trigger value updates

Artusamak:
Custom types are not correctly supported due to the fact that `UnitOfWork::computeOrRecomputeChangeSet()` uses` === `operations without checking the objects inner values.
This leads to missed update values.

The added test `testValueObjectChangeSets` is red if the changeset detection fragment is not altered.

There is also a `@TODO` added to track the suggestion to use `\Doctrine\Common\Comparable` interface. The problem is that this interface is part of doctrine/commons which is not required by the package at the moment.

|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #1482


tests/Documents/Functional/DocumentWithCustomTypeValueObject.php
```diff
@@ -0,0 +1,17 @@
...
+namespace Documents\Functional;
+
+use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
+
+#[ODM\Document]
+class DocumentWithCustomTypeValueObject {
+
+  /** @var string|null */
+  #[ODM\Id]
+  public $id;
```
GromNaN: Nitpick: you should use typed properties when adding examples.
```suggestion
  #[ODM\Id]
  public ?string $id;
```

tests/Documents/Functional/ValueObjectChild.php
```diff
@@ -0,0 +1,12 @@
...
+namespace Documents\Functional;
+
+class ValueObjectChild {
+
+  public function __construct(
+    public int $prop1,
+    public int $prop2,
+  ) {}
+
+}
```
GromNaN:
```suggestion
}

```

src/UnitOfWork.php
```diff
@@ -765,16 +765,14 @@ private function computeOrRecomputeChangeSet(ClassMetadata $class, object $docum
...
-                        continue;
-                    }
+                // Target object behaviors.
+                if (is_object($actualValue)) {
+                  if ($actualValue instanceof PersistentCollectionInterface && (! $actualValue->isDirty() && ! $this->isCollectionScheduledForDeletion($actualValue))) {
+                    // consider dirty collections as changed as well
+                    continue;
+                  }
+                  // @TODO: Consider using Comparable interface to delegate
+                  // object comparaisons.
```
GromNaN: We don't leave todo in the codebase. Is it something you want to address before the PR is merged?

GromNaN (commented):
Could you have a look at the failing tests?

src/UnitOfWork.php
Artusamak: Yes!

Artusamak:
You've been faster light for review! Thanks. 

Is it considered to add a `.editorfile` matching your code style expectations? Fixing manually is not the easiest solution 😅.

Artusamak:
Thanks to SPL there is now a way to detect the object content update which should do the trick.

src/UnitOfWork.php
```diff
@@ -48,6 +48,7 @@
 use function method_exists;
 use function preg_match;
 use function serialize;
+use function spl_object_hash;
```
GromNaN, line 51: `spl_object_hash($a) === spl_object_hash($b)` is functionally identical to `$a === $b` for two simultaneously live objects ([PHP docs](https://www.php.net/spl_object_hash): the hash can only be reused after the object is destroyed). This import can be dropped.

src/UnitOfWork.php
```diff
@@ -765,16 +766,19 @@ private function computeOrRecomputeChangeSet(ClassMetadata $class, object $docum
...
+                          // consider dirty collections as changed as well
+                          continue;
                     }
 
-                    if (! $actualValue->isDirty() && ! $this->isCollectionScheduledForDeletion($actualValue)) {
-                        // consider dirty collections as changed as well
+                    // If object content has not been modified, nothing to do.
+                    if (is_object($orgValue) && spl_object_hash($actualValue) === spl_object_hash($orgValue)) {
                         continue;
                     }
```
GromNaN, line 779: This restructuring introduces a regression for dirty `ReferenceMany` collections. When a `PersistentCollection` is dirty but its instance is the same object stored in `$originalDocumentData` (`$orgValue === $actualValue` — the normal case), the new code falls through to line 777 and executes `continue`, silently skipping the dirty collection instead of detecting it as changed.

`EmbedMany` collections are unaffected by luck — they go through the association changeset loop (~line 870). `ReferenceMany` collections are explicitly excluded there (`if (isset($mapping['reference'])) { continue; }`), so this block is their only detection path.

PR #3021 adds a test that demonstrates this: it passes on `2.17.x` and fails with this patch applied.

Also, the comment `// consider dirty collections as changed as well` is now on the `continue` of the **clean** collection branch — the opposite of what it says.

Since `spl_object_hash` is equivalent to `===` for live objects, it does not help with value comparison either. I'd suggest keeping the original structure and adding a `==` check as a new `elseif`. PHP's `==` compares objects by value: same class and all properties (public **and** private) equal — https://3v4l.org/sBNOH:

```php
// skip if value has not changed
if ($orgValue === $actualValue) {
    if (! $actualValue instanceof PersistentCollectionInterface) {
        continue;
    }

    if (! $actualValue->isDirty() && ! $this->isCollectionScheduledForDeletion($actualValue)) {
        // consider dirty collections as changed as well
        continue;
    }
} elseif (is_object($orgValue) && is_object($actualValue) && $orgValue == $actualValue) {
    continue;
}
```

This preserves the dirty-collection behaviour and fixes the value object case without any new dependency. The `TODO` on `Comparable` remains valid for edge cases where `==` is not sufficient, but is not a blocker.

tests/Tests/Functional/CustomTypeTest.php
```diff
@@ -216,3 +243,7 @@ public function convertToPHPValue($value): ?Language
 class CustomTypeWithoutClosureToPHP extends Type
 {
 }
+
+class CustomValueObjectChild extends Type
```
GromNaN, line 247: `CustomValueObjectChild` implements no `convertToPHPValue`/`convertToDatabaseValue`, so the test doesn't quite cover a realistic custom type scenario where the type actually converts values between PHP and MongoDB.

tests/Tests/Functional/CustomTypeTest.php
```diff
@@ -54,6 +57,30 @@ public function testCustomTypeValueConversions(): void
...
+
+        $this->uow->commit();
+        $valueChild  = new ValueObjectChild(12, 12);
+        $root->child = $valueChild;
+
+        $this->uow->computeChangeSets();
+        $changeSet = $this->uow->getDocumentChangeSet($root);
+
+        self::assertArrayHasKey('child', $changeSet);
+        self::assertEquals('12', $changeSet['child'][1]->prop1);
```
GromNaN, line 81: `prop1` is declared `int`, so this should compare against `12` (no quotes). The assertion should use `assertSame`.

GromNaN (changes requested):
Thanks for the contribution and the test that clearly reproduces the issue! I have a few comments below.

PR #3020, opened 2026-06-24
Merge 2.17.x into 3.0.x

IonBazan:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | See #3014 

#### Summary

Merges 2.17.x into 3.0.x and removes compatibility layer for doctrine/collections v2 support. Changes are contained in the merge commit to preserve original methods' authors credits.

### DIFF
--- a/src/PersistentCollection/PersistentCollectionTrait.php
+++ b/src/PersistentCollection/PersistentCollectionTrait.php
@@ -7,7 +7,6 @@ namespace Doctrine\ODM\MongoDB\PersistentCollection;
 use Closure;
 use Doctrine\Common\Collections\Collection as BaseCollection;
 use Doctrine\Common\Collections\Criteria;
-use Doctrine\Common\Collections\Selectable;
 use Doctrine\ODM\MongoDB\DocumentManager;
 use Doctrine\ODM\MongoDB\Mapping\ClassMetadata;
 use Doctrine\ODM\MongoDB\MongoDBException;
@@ -441,7 +440,7 @@ trait PersistentCollectionTrait
      * @param mixed $element The element to add.
      * @phpstan-param T $element
      */
-    public function add($element): void
+    public function add(mixed $element): void
     {
         $this->doAdd($element, false);
     }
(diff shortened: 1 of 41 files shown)


src/PersistentCollection/PersistentCollectionTrait.php
```diff
@@ -775,10 +774,6 @@ public function matching(Criteria $criteria): BaseCollection
     {
         $this->initialize();
 
-        if (! $this->coll instanceof Selectable) {
-            throw new LogicException('The backed collection must implement Selectable to use matching().');
-        }
```
GromNaN, line 780: This change must be reverted. In `doctrine/collections: v2`, the `Collection` interface does not extends `Selectable`.

We can remove support for `doctrine/collections: v2` in the 3.0.x branch later.

IonBazan: Support for collections v2 is already being removed in this PR. ODM v3 will only support collections v3.1

GromNaN (approved):
I've added follow-up commits to fix compatibility with changes already present on `3.0.x`:

- **efbf6ef6** — Fix Symfony version constraint in CI, improving on the fix from doctrine/mongodb-odm@713bdd76.

- **14c823e7** — Drop the `doRemove()` early-return guard that skipped `changed()` when the underlying `remove()` returned a falsy value. On `3.0.x` the test was intentionally updated to expect `isDirty() === true` regardless of the return value, reflecting the design decision that `PersistentCollection::remove()` always marks the collection dirty.

- **46bd6a56** — Fix `'>= 2.2'` → `'>= 2.2.0'` in `RequiresPhpExtension` (PHPUnit 12 requires a full three-part version string), and use `ReflectionProperty` to modify `ClassMetadata::$collection` in the test (`private(set)` asymmetric visibility in PHP 8.4 prevents direct assignment from outside the class).

The last two fixes are specific to 3.0 branch constraints. Backporting is possible but not required.

PR #3022, opened 2026-06-29
Remove COLLECTION_PER_CLASS inheritance type

### LINKED ISSUE #2742: Deprecate `COLLECTION_PER_CLASS` inheritance

GromNaN:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | Fixes #2742. Deprecation introduced in #3011

#### Summary

Removes the `COLLECTION_PER_CLASS` inheritance type, deprecated in 2.17. Each document class is already mapped to its own collection, so the type provided no additional behavior. The mapping drivers now throw a `MappingException` when an unknown inheritance type is configured.


docs/en/reference/inheritance-mapping.rst
```diff
@@ -182,55 +180,3 @@ discriminator field:
           <document name="Documents\Employee">
           </document>
         </doctrine-mongo-mapping>
-
-.. _collection_per_class_inheritance:
-
-Collection Per Class Inheritance
```
malarzm, line 188: I think it'd be good to keep a section and showing you can just extend documents and change the collection

GromNaN: Good point. There was already a section on `MappedSuperclass`, but I've revised the doc to make it clear that PHP inheritance just works.

src/Mapping/Driver/XmlDriver.php
```diff
@@ -172,7 +173,12 @@ public function loadMetadataForClass(string $className, \Doctrine\Persistence\Ma
 
         if (isset($xmlRoot['inheritance-type'])) {
             $inheritanceType = (string) $xmlRoot['inheritance-type'];
-            $metadata->setInheritanceType(constant(ClassMetadata::class . '::INHERITANCE_TYPE_' . $inheritanceType));
+            $constantName    = ClassMetadata::class . '::INHERITANCE_TYPE_' . $inheritanceType;
+            if (! defined($constantName)) {
+                throw MappingException::invalidInheritanceType($className, $inheritanceType);
+            }
+
+            $metadata->setInheritanceType(constant($constantName));
```
GromNaN, line 181: In a follow-up PR, we can add support for using the const value in the `InheritanceType` attribute.

GromNaN:
PR rebased after upmerging https://github.com/doctrine/mongodb-odm/pull/3011

paulinevos (approved):

IonBazan (approved):

PR #3024, opened 2026-07-01
Add `autoEmbed` support for vector search indexes and `$vectorSearch` stage

GromNaN:
Adds support for MongoDB's [automated embedding](https://www.mongodb.com/docs/vector-search/crud-embeddings/automated-embedding/) feature:

- New `autoEmbed` field type in `VectorSearchIndexField` (alongside `vector` and `filter`), with `modality` and `model` fields
- New constant `ClassMetadata::VECTOR_AUTOEMBEDDING_MODALITY_TEXT`
- XML driver support via `<auto-embed-field path="..." modality="..." model="..."/>`
- New `query(string)` and `model(string)` methods on the `$vectorSearch` aggregation stage, allowing text queries to be embedded automatically at query time

## Usage

```php
// Index definition
#[VectorSearchIndex(fields: [
    [
        'type' => 'autoEmbed',
        'path' => 'content', 'modality' => 'text',
        'model' => 'voyage-4-large',
    ],
    ['type' => 'filter', 'path' => 'category'],
])]

// Query
$dm->createAggregationBuilder(Article::class)
    ->vectorSearch($persister)
        ->index('default')
        ->path('content')
        ->query('semantic search text')
        ->model('voyage-4-light')
        ->limit(10);
```

PR #3028, opened 2026-07-03
Add `--wait` option to `odm:schema:create` and `odm:schema:update`

### LINKED ISSUE #3026: `odm:schema:create` has no option to wait for search/vector indexes to become READY
`SchemaManager::waitForSearchIndexes()` exists but isn't exposed by `odm:schema:create`/`update`. Since these commands create search indexes via `createSearchIndexes()`/`updateSearchIndexes()`, which only submit the definition, scripts that need the index queryable (CI, deploys, seed scripts) have to reimplement the create logic to add a wait.

This matters more with `autoEmbed` (#3024): the initial embedding backfill runs through Voyage AI and can take minutes on non-trivial collections, vs. seconds for regular Atlas Search indexes.

### Proposal

Add `--wait` (and optionally `--wait-ms`, defaulting to `waitForSearchIndexes()`'s own default) to `odm:schema:create` and `odm:schema:update`. When set, call `\$sm->waitForSearchIndexes(\$classNames, \$waitMs)` after submitting definitions.

Happy to PR — flagging first to confirm the flag shape (single `--wait` vs. `--wait` + `--wait-ms`).

GromNaN:
Closes #3026.

Search indexes are built asynchronously by the server: `odm:schema:create` and `odm:schema:update` return before the index is queryable. Scripts that need the index to be ready (CI seeding, deployments, integration tests) had to reimplement the wait around `SchemaManager::waitForSearchIndexes()`. This matters more with `autoEmbed` (#3024), where the initial embedding backfill through Voyage AI can take minutes on non-trivial collections.

Both commands now accept a `--wait` option with an optional value:

```
$ bin/mongodb odm:schema:create --search-index --wait
$ bin/mongodb odm:schema:update --wait=1minute
$ bin/mongodb odm:schema:create --search-index --wait="30 seconds"
$ bin/mongodb odm:schema:create --search-index --wait=5000
```

The value is a duration string parsable by `strtotime()` (`30 seconds`, `1minute`, `1 hour`) or a positive integer of milliseconds. When `--wait` is passed without a value, it falls back to the `SchemaManager` default of 10 seconds. The wait is skipped automatically when no mapped class declares a search index.

Documentation has been updated in `docs/en/reference/search-indexes.rst` and the vector-search cookbook, including a note that waiting reports index readiness only, not that the background indexer has caught up with documents inserted after index creation.

### DIFF
--- a/docs/en/cookbook/vector-search.rst
+++ b/docs/en/cookbook/vector-search.rst
@@ -151,7 +151,9 @@ the index to be ready using the following code:
 
 From the command line, ``odm:schema:create`` and ``odm:schema:update`` accept
 a ``--wait`` option that submits the search index definitions and waits for
-them to become queryable before returning:
+them to become queryable before returning. The value is a maximum wait
+duration: the command returns as soon as the indexes are ready, without
+waiting out the rest of the duration.
 
 .. code-block:: console
 
@@ -159,7 +161,7 @@ them to become queryable before returning:
 
 The value accepts a duration such as ``30 seconds``, ``1minute`` or
 ``1 hour``, or a positive integer number of milliseconds. When ``--wait`` is
-passed without a value, a default timeout of 10 seconds is used.
+passed without a value, a default timeout of 5 minutes is used.
 
 .. note::
 
--- a/docs/en/cookbook/vector-search.rst
+++ b/docs/en/cookbook/vector-search.rst
@@ -141,13 +141,34 @@ in search results.
     $schemaManager->createDocumentSearchIndexes(Guide::class);
 
 
-If the vector search index created after inserting documents, the index is
+If the vector search index is created after inserting documents, the index is
 marked as "READY" when all existing documents are indexed. You can wait for
 the index to be ready using the following code:
 
 .. code-block:: php
 
     $schemaManager->waitForSearchIndexes([Guide::class]);
+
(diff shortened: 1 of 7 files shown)


GromNaN:
I'm now thinking that the command should not return 0 when the `--wait` option is used and the wait timed out.

Edit: this is already the case. In case of timeout, a `MongoDBException` is thrown, and the command return code is 255.

docs/en/cookbook/vector-search.rst
```diff
@@ -141,14 +141,35 @@ in search results.
...
 
+From the command line, ``odm:schema:create`` and ``odm:schema:update`` accept
+a ``--wait`` option that submits the search index definitions and waits for
+them to become queryable before returning:
+
+.. code-block:: console
+
+    $ php mongodb.php odm:schema:create --search-index --wait=1minute
+
+The value accepts a duration such as ``30 seconds``, ``1minute`` or
```
paulinevos, line 160: I think this could be clarified as it seems to be a _maximum_ duration. I'm assuming that the script stops when the index is ready and doesn't wait out the remainder of the duration.

GromNaN: Yes, that's a wait *timeout*, can you update the description and merge?

paulinevos (approved):

PR #3033, opened 2026-08-12
Add Doctrine MongoDB ODM agent skill for AI coding assistants

paulinevos:
## Summary

- Adds an Agent Skill (`skills/mongodb-odm/`) for end users of `doctrine/mongodb-odm`, following the standard `skills/<name>/SKILL.md` layout.
- Covers the common workflows (mapping, references/embedding, querying, aggregation) plus the advanced topics requested: Atlas Search & Vector Search, Queryable Encryption, and combining Doctrine ORM with MongoDB ODM in one app.
- Suggests the official MongoDB MCP server for live cluster/index inspection when debugging data or search/vector index issues.
- All content is grounded directly in this repo's own `docs/en/` RST sources for the 2.16 release, with a couple of small corrections applied for 2.16-vs-3.0 drift (proxy terminology, the deprecated `UUID` id strategy, per-class `DocumentManager::clear()`). The shipped skill files do not reference `docs/` at all — they are self-contained.

## Files added

```
skills/mongodb-odm/
├── SKILL.md
└── references/
    ├── setup-architecture.md
    ├── mapping.md
    ├── references-embedding.md
    ├── querying-aggregation.md
    ├── indexes-search.md
    ├── vector-search.md
    ├── queryable-encryption.md
    ├── hybrid-orm-odm.md
    └── storage-transactions-lifecycle.md

tests/Skills/
├── README.md
└── mongodb-odm/evals.json
```

## Packaging

`.gitattributes` export-ignores `/docs`, `/tests`, `/tools` and `/benchmark`, but **not** `/skills`. That is deliberate: the skill ships in the Composer dist package, so a consuming project can point its assistant at `vendor/doctrine/mongodb-odm/skills/`. The evals under `tests/` stay out of the dist, as they are development-time tooling.

## Evals

`tests/Skills/mongodb-odm/evals.json` holds 11 evals — at least one per reference file. Each pairs a realistic user prompt with a prose description of what a correct answer must contain. `tests/Skills/README.md` documents how to run them.

There is no automated runner and these do not run in CI: grading a prose answer against a prose expectation needs a model, and therefore an API key, which we don't want to require from contributors.

## Benchmark results

Method: each eval's `expected_output` was first decomposed into a fixed, numbered list of atomic assertions (143 in total across the 11 evals). Every eval was then run twice — once in a workspace with the skill installed, once in an identical workspace without it — in fresh non-interactive sessions with the same model (Sonnet) and the same read-only tool access, sending the prompt verbatim with no follow-ups. Both arms were graded blind against the same fixed assertion list. Runs in the skill arm were verified to actually invoke the skill and read its reference files.

Eval 11 was added later and run under the same two-arm method, with one difference worth stating: its skill arm was pointed at the skill directory and told to load `SKILL.md` plus the reference files its routing table selects, rather than relying on autonomous invocation. It tests retrieval and use of the content, not the trigger.

| # | Eval | With skill | Without skill |
|---|---|---|---|
| 1 | `ecommerce-mapping-and-query` | 20/20 (100%) | 20/20 (100%) |
| 2 | `atlas-vector-search-semantic-search` † | 9/9 (100%) | 2/9 (22%) |
| 3 | `queryable-encryption-searchable-ssn` | 8/8 (100%) | 2/8 (25%) |
| 4 | `hybrid-orm-odm-cross-reference` | 10/10 (100%) | 4/10 (40%) |
| 5 | `n-plus-one-reference-priming` † | 6/6 (100%) | 3/6 (50%) |
| 6 | `bidirectional-reference-inverse-side-trap` | 9/9 (100%) | 9/9 (100%) |
| 7 | `optimistic-locking-concurrent-writes` | 9/9 (100%) | 7/9 (78%) |
| 8 | `atlas-search-index-over-embedded-collection` | 15/15 (100%) | 3/15 (20%) |
| 9 | `single-collection-inheritance-legacy-documents` †‡ | 12/12 (100%) | 7/12 (58%) |
| 10 | `batch-import-memory-growth` †‡ | 10/10 (100%) | 6/10 (60%) |
| 11 | `query-builder-maintenance-command-generation` | 15/15 (100%) | 7/15 (47%) |
| | **Overall (assertion-weighted)** | **123/123 (100%)** | **70/123 (56.9%)** |

† Re-measured after a follow-up pass on the gaps below. These four rows use assertion lists re-derived from the same `expected_output`, at a coarser grain than the originals (9/6/12/10 vs 19/9/14/15), because the original lists were not kept. Directionally comparable; not digit-comparable with the six rows above them, and the coarser grain mechanically flatters both arms' percentages. ‡ Prompt rewritten — see below.

### What the skill actually fixes

Without it, the model invents APIs that don't exist and misses platform constraints:

- **Vector search** — bypassed `#[VectorSearchIndex]` for a raw `createSearchIndexes` command, then used a raw `$vectorSearch` pipeline array while explicitly denying the Aggregation Builder has a `vectorSearch()` helper (it does, since 2.13).
- **Queryable encryption** — mapped the SSN as a plain `#[Field]`, never reaching `#[Encrypt(queryType: EncryptQuery::Equality)]`, and hand-rolled `driverOptions` instead of `Configuration::setAutoEncryption()`/`setKmsProvider()`.
- **Atlas Search** — invented a property-level `#[SearchField]` attribute, claimed dynamic mapping *does* recurse into embedded arrays (it doesn't), used the PHP property name instead of the database field name, and guessed the CLI flag as `--search-indexes`.
- **Query Builder update/read semantics** (eval 11) — denied that `Builder::updateMany()` can take an aggregation-pipeline update at all (`Builder::pipeline()` has existed since 2.15), wrote a comment justifying a deliberate drop to the raw `MongoDB\Collection`, reached for `hydrate(false)` instead of `readOnly()` for an unpersistable snapshot, and called `returnNew(true)` without knowing that `findAndUpdate` otherwise returns the pre-update document.
- **Priming** (eval 5) — invented a `#[ReferencePrime(['field' => 'author'])]` attribute and a `prime: true` option on `#[ReferenceOne]`, neither of which exists, and never reached the real query-builder `field('author')->prime(true)` or the hydration requirement.
- **Legacy discriminators** (eval 9) — stated that "the ODM does not have built-in 'default discriminator value' support the way some other ORMs do, so backfilling is the recommended approach", while `#[DefaultDiscriminatorValue('person')]` sat in its own code block two paragraphs below, and pushed a raw `updateMany` backfill as the "cleanest long-term" fix.
- **ORM/ODM bridge** — produced its own architecture instead of the cookbook's `postLoad` + `getReference()` pattern, using `find()` and omitting the `instanceof` guard.

### Evals that do not discriminate

Evals 1 and 6 score identically in both arms. Common mapping/querying and the inverse-side trap are well represented in model training data, so they are **not** evidence for the skill. They are kept deliberately as a floor — to catch a regression where a change to the skill makes a previously correct answer worse — and `tests/Skills/README.md` says so.

A first draft of eval 11 scored 12/12 against 11/12 for the same reason: it handed the model a broken script and asked it to diagnose the symptoms, and recognition of a pasted bug turns out to be far easier than avoiding it while generating. Rewriting the same facts as a from-scratch "build this command" prompt moved the control arm from 11/12 to 7/15. Generation prompts discriminate; diagnosis prompts mostly don't.

### The gap-closing pass

An earlier revision of this PR logged 7 skill-arm failures across evals 2, 5, 9 and 10. Every fact
behind them was already in the reference files, so none were content gaps. They resolved into four
different causes, which is why they needed four different fixes:

1. **A routing bug (3 of the 7, all in eval 10).** The batching, `clear()`, `merge()` and
   detached-document facts live in `setup-architecture.md`, but that file's routing row read
   "Bootstrapping ODM, debugging persist/flush behavior, custom repositories". A 500k-row CSV memory
   problem reads like *storage and lifecycle*, so the model loaded
   `storage-transactions-lifecycle.md`, which does not contain them. Fixed by naming the symptoms in
   the routing row and adding an explicit "not here" pointer at the top of the storage file.
2. **No rule that prerequisites must reach the answer (2 of the 7, evals 2 and 9).**
   `vector-search.md` states the 2.13+ requirement in its opening line and `mapping.md` flags the
   `COLLECTION_PER_CLASS` deprecation twice. The model read them and did not repeat them: an
   output-shaping gap, not a retrieval one. Fixed with a constraint in `SKILL.md` requiring version
   prerequisites and deprecations to be stated in the answer, on the grounds that the user cannot see
   these files.
3. **One thin explanation (2 of the 7, eval 5).** The priming section gave the fix without the
   mechanism, and so did the answer. It now names the N+1 problem and explains that
   `#[ReferenceOne]` hydrates as an uninitialized proxy and `#[ReferenceMany]` as an uninitialized
   `PersistentCollection`, so touching either fires a query invisibly.
4. **Two bad assertions (evals 9 and 10).** Both required the answer to warn against a mistake the
   user had not made — the `COLLECTION_PER_CLASS` deprecation when the user never proposed it, and
   "don't flush per row" when the user's code did not. The model avoided both traps correctly and
   simply declined to lecture about the road not taken, which is the right behaviour; padding every
   answer with deprecation notices for unused APIs would be worse. Fixed in the prompts rather than
   the skill: eval 9's user now has a colleague recommending `COLLECTION_PER_CLASS`, and eval 10's
   user now says they flush after every row and explains why they thought that was safe. Both facts
   became directly responsive, and both arms were re-run.

**One eval was asserting deprecated API.** Eval 10 required "the 2.16 per-class
`$dm->clear(Product::class)`". `DocumentManager::clear($objectName)` has been deprecated since
**2.4** and is unsupported in 3.0 (`src/DocumentManager.php:722-731`); it is not a 2.16 feature. The
skill said so too (`setup-architecture.md`), most likely because the `onClear` event args genuinely
are new in 2.16 and the version attached itself to the neighbouring feature. Both are corrected, and
the assertion is inverted: the model should *not* reach for that argument. The skill arm's original
12/15 on eval 10 was therefore mis-scored — one of its three "failures" was the skill arm being
right.

### Known gaps in the skill arm

With the above applied, all four re-measured evals score 100% on their assertion lists. Two defects
that the assertion lists do not capture are logged rather than papered over:

- On eval 10 the skill arm justified batching partly with "MongoDB doesn't have SQL-style
  transactions in the way that motivates flush-per-row". MongoDB has had multi-document transactions
  since 4.0 and ODM supports them — `storage-transactions-lifecycle.md` §5 documents exactly that.
  The advice is right; the reason given is wrong.
- On eval 2 the skill arm wrote `->distinct(...)->getQuery()->execute()->toArray()` in an earlier
  run. `Query::execute()` for a distinct query returns a plain array
  (`src/Query/Query.php:514-520`), so `toArray()` is a fatal error. Three of four runs across both
  arms made this call before `querying-aggregation.md` was amended to state the return type.

## Test plan

- [x] Skill triggers on Doctrine MongoDB ODM prompts and loads its reference files (verified via tool-use trace)
- [x] 10 evals run in both arms and graded against fixed assertion lists
- [ ] Spot-check a couple of reference files against the live `2.16` docs for accuracy
- [ ] Confirm PHP code samples are syntactically valid (`php -l`)

### DIFF
--- a/skills/mongodb-odm/SKILL.md
+++ b/skills/mongodb-odm/SKILL.md
@@ -1,0 +1,207 @@
+---
+name: mongodb-odm
+description: Implementation guide for doctrine/mongodb-odm (Doctrine MongoDB ODM), the PHP object-document mapper for MongoDB. Use whenever the user is writing, debugging, or reviewing PHP code that maps classes to MongoDB collections with Doctrine — mentions of "Doctrine MongoDB ODM", "mongodb-odm", "DocumentManager", "#[Document]"/"@Document", "#[ReferenceOne]"/"#[EmbedOne]", "ODM query builder", "aggregation builder" in a Doctrine/PHP context, "GridFS with Doctrine", "MongoDB queryable encryption PHP", "MongoDB vector search PHP", or combining Doctrine ORM and MongoDB ODM in one app should all trigger this skill. Also use it to correct common mistakes: treating MongoDB ODM like Doctrine ORM (assuming SQL-style joins, automatic cascading, or table-based migrations), calling flush() after every mutation, storing dates as strings, or assuming Vector Search/Queryable Encryption work outside MongoDB Atlas.
+license: MIT
+---
+
+# Doctrine MongoDB ODM
+
+Implementation skill for `doctrine/mongodb-odm` 2.16/2.17, the PHP object-document mapper that persists plain PHP classes to MongoDB. It exists to prevent the mistakes that come from treating it like a SQL ORM: expecting automatic cascading, joining collections like tables, flushing on every mutation, or assuming every advanced MongoDB feature (Vector Search, Queryable Encryption) works outside Atlas.
+
+## Core Workflow
+
+1. Identify the layer: mapping (a document class), persistence (`DocumentManager`), querying (Query Builder / Aggregation Builder), or an advanced/infra concern (search, encryption, sharding, transactions).
+2. Confirm the class has `#[Document]` (or `#[EmbeddedDocument]`/`#[MappedSuperclass]`) and an `#[Id]` property.
+3. Decide **embed vs. reference** deliberately (`references-embedding.md`) — don't default to references out of ORM habit when the data is really owned/composed.
+4. Batch mutations into a single unit of work and call `flush()` once, not after every `persist()`/`remove()` (`setup-architecture.md`).
+5. Use the Query Builder or Aggregation Builder for anything beyond `find()`/`findBy()` — never hand-build a raw driver query when the fluent API covers it (`querying-aggregation.md`).
+6. Validate with the consuming project's own test suite/static analysis, and use `odm:schema:create`/`odm:schema:update` to keep indexes and search indexes in sync with the mapping.
+
+## Suggest the MongoDB MCP Server
+
+Many ODM bugs are really questions about what's in the database — current document shape, whether an index exists yet, whether a vector/search index finished building. None of that is visible from PHP source. When debugging data/query issues or setting up Atlas Search/Vector Search/sharding, suggest connecting the official **MongoDB MCP Server**:
+
+```bash
+npx -y mongodb-mcp-server@latest --readOnly
+```
+
+```json
+{
+  "mcpServers": {
+    "MongoDB": {
+      "command": "npx",
+      "args": ["-y", "mongodb-mcp-server@latest", "--readOnly"],
+      "env": { "MDB_MCP_CONNECTION_STRING": "mongodb://localhost:27017/myDatabase" }
+    }
+  }
+}
+```
+
+It also accepts Atlas API service account credentials (`MDB_MCP_API_CLIENT_ID`/`MDB_MCP_API_CLIENT_SECRET`) for cluster/project-level operations instead of a connection string. Prefer environment variables over CLI args for credentials.
+
+## Reference Guide
+
+| Topic | Reference file | Load when |
+|---|---|---|
+| Installation, `DocumentManager`, UnitOfWork, repositories, console commands, `clear()`/`detach()`/`merge()`, document states | `references/setup-architecture.md` | Bootstrapping ODM, debugging persist/flush behavior, custom repositories, **batch imports, long-running CLI scripts, rising memory or slowdown over a loop, detached documents** |
+| `#[Document]`/`#[Field]`/`#[Id]`, field types, inheritance | `references/mapping.md` | Defining or modifying a document class |
+| `#[ReferenceOne]`/`#[ReferenceMany]`, `#[EmbedOne]`/`#[EmbedMany]`, bidirectional refs, trees, priming | `references/references-embedding.md` | Modeling relationships, fixing N+1 patterns |
+| Query Builder, findAndModify, upserts, geospatial queries, filters, Aggregation Builder | `references/querying-aggregation.md` | Writing any query or aggregation pipeline |
+| Standard indexes, Atlas Search indexes, simple keyword search | `references/indexes-search.md` | Adding indexes, Atlas full-text search |
+| Vector Search (Atlas) | `references/vector-search.md` | Embeddings, semantic search, `$vectorSearch`, hybrid text+vector search |
+| Queryable Encryption (Atlas/Enterprise) | `references/queryable-encryption.md` | Encrypting sensitive fields, `#[Encrypt]` |
+| Combining Doctrine ORM (SQL) and MongoDB ODM | `references/hybrid-orm-odm.md` | An app persists some data to SQL and some to MongoDB and needs them to interoperate |
+| GridFS, capped collections, sharding, transactions/locking, change tracking, storage strategies, events, schema migration, time series | `references/storage-transactions-lifecycle.md` | File storage, sharding, locking, lifecycle callbacks, evolving a mapping |
+
+## Constraints
+
+### MUST DO
+
+- Put `#[Document]` (or `#[EmbeddedDocument]`) plus `#[Id]` on every persisted class; `#[Id]` properties must never be `readonly` (the generator sets it externally).
+- Use `DateTime`/`DateTimeImmutable` (preferably immutable) for dates — never store a date as a string.
+- Batch changes and `flush()` once per logical unit of work (typically 0–2 times per HTTP request).
+- Set `cascade` explicitly on `#[ReferenceOne]`/`#[ReferenceMany]` when you need cascading persist/remove — not automatic for references (it *is* automatic for embeds).
+- Use `Doctrine\Common\Collections\Collection`/`ArrayCollection` (never a plain `array`) for `#[ReferenceMany]`/`#[EmbedMany]` properties.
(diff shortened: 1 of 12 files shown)


GromNaN:
> Common mapping & querying 7/7 (100%)

This is an indicator this skill is not necessary. You should identify more complex questions.

I think you should commit the eval files into the `tests/` directory.

skills/mongodb-odm/references/hybrid-orm-odm.md
```diff
@@ -0,0 +1,106 @@
+# Combining Doctrine ORM and MongoDB ODM (Hybrid Persistence)
+
+Grounded in `docs/en/cookbook/blending-orm-and-mongodb-odm.rst` and `mapping-classes-to-orm-and-odm.rst`. (`resolve-target-document-listener.rst` resolves an interface to a Document class *within* the ODM only — unrelated to ORM interop.)
```
GromNaN: The `./docs` directory is excluded from the ZIP package (used as dist package by Composer). How is this skill installed? It can't reference docs files if it's installed with composer. Otherwise the skill might be excluded from the dist zip too.

paulinevos: I think it's not referencing as much as documenting its reasoning, which is kind of annoying tbh. I'm gonna add to my guidelines for it to not do that. It's not relevant for usage

tests/Skills/mongodb-odm/evals.json
GromNaN, line 1: The `evals` file is relevant. It should contain at least as many tests as there are reference files.
Can we add a `tests/kills/README.md` file that documents how to run the `evals` file? I don't think we can run it in GitHub Actions because it requires a model API key.

skills/mongodb-odm/SKILL.md
```diff
@@ -0,0 +1,208 @@
+---
+name: mongodb-odm
+description: Implementation guide for doctrine/mongodb-odm (Doctrine MongoDB ODM), the PHP object-document mapper for MongoDB. Use whenever the user is writing, debugging, or reviewing PHP code that maps classes to MongoDB collections with Doctrine — mentions of "Doctrine MongoDB ODM", "mongodb-odm", "DocumentManager", "#[Document]"/"@Document", "#[ReferenceOne]"/"#[EmbedOne]", "ODM query builder", "aggregation builder" in a Doctrine/PHP context, "GridFS with Doctrine", "MongoDB queryable encryption PHP", "MongoDB vector search PHP", or combining Doctrine ORM and MongoDB ODM in one app should all trigger this skill. Also use it to correct common mistakes: treating MongoDB ODM like Doctrine ORM (assuming SQL-style joins, automatic cascading, or table-based migrations), calling flush() after every mutation, storing dates as strings, or assuming Vector Search/Queryable Encryption work outside MongoDB Atlas.
+license: MIT
+metadata:
+  version: "1.2.0"
```
GromNaN: What is this version?

paulinevos: No idea lol, it's gone now

GromNaN:
Could you iterate again on the skill to reach 100% success on the eval?

GromNaN (approved):

PR #3034, opened 2026-08-23
Catch SearchNotSupportedException in SchemaManager

GromNaN:
mongodb/mongodb 2.2.0 introduces a dedicated `SearchNotSupportedException` for servers that don't support Atlas Search / Vector Search, replacing the generic `CommandException` previously used for this case (see mongodb/mongo-php-library@fac7ebc93fe9b47ab6cd32b1bb87e7f6c1e3af2d). This catches the new exception with the same handling as before, while keeping the `CommandException` handling for older versions of mongodb/mongodb.

The error detection logic is backported from `MongoDB\Exception\SearchNotSupportedException::isSearchNotSupportedError()` into `SchemaManager::isSearchIndexCommandException()`, so both code paths recognize the same set of server errors.

Related: #2671, #2787

### DIFF
--- a/tests/Tests/SchemaManagerTest.php
+++ b/tests/Tests/SchemaManagerTest.php
@@ -1499,7 +1499,7 @@ EOT;
 
     private function createSearchIndexCommandExceptionForOlderServers(): CommandException
     {
-        return new CommandException('Unrecognized pipeline stage name: \'$listSearchIndexes\'', 40234);
+        return new CommandException('Unrecognized pipeline stage name: \'$listSearchIndexes\'', 40324);
     }
 
     private function createIndexIterator(array $indexes = []): Iterator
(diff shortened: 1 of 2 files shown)


tests/Tests/SchemaManagerTest.php
```diff
@@ -1499,7 +1499,7 @@ private function createSearchNotEnabledCommandException(): CommandException
 
     private function createSearchIndexCommandExceptionForOlderServers(): CommandException
     {
-        return new CommandException('Unrecognized pipeline stage name: \'$listSearchIndexes\'', 40234);
+        return new CommandException('Unrecognized pipeline stage name: \'$listSearchIndexes\'', 40324);
```
paulinevos, line 1502: What about the old error code this is replacing? Is that still handled somewhere?

GromNaN: This error code was not checked. It was incorrect:

| Code | Meaning | Typical cause |
| ------------- | ------------- | ------------- |
| 40234 | The field `<field>` must be an accumulator object | A $group, $bucket, or $bucketAuto output field is not an object containing an accumulator such as $sum or $avg. |
| 40324 | Unrecognized pipeline stage name: `<stage>` | The pipeline contains an unknown, misspelled, unsupported, or unavailable stage. |

paulinevos (approved):

Issue #1536, opened 2016-12-15
Improve index documentation


malarzm:
While answering http://stackoverflow.com/questions/41172585/index-not-created/41173619 I found that it's actually hard to find anything about creating indexes in the docs (there's small paragraph by the end of [Embedded Indexes](http://docs.doctrine-project.org/projects/doctrine-mongodb-odm/en/latest/reference/indexes.html#embedded-indexes)). Also [Console Commands](http://docs.doctrine-project.org/projects/doctrine-mongodb-odm/en/latest/reference/console-commands.html) chapter seems to be reachable only from that very sentence in indexes, might be good to expose them on the front page as well.

olivmaurel:
Hi, I can update the documentation if it's still needed

malarzm:
@olivmaurel that'd be awesome!

alcaeus:
Bad bot...

PR #3035, opened 2026-08-27
Improve index documentation

TrueFurina:
## Summary

Per #1536: index creation was hard to find (a small paragraph at the end of Embedded Indexes) and the Console Commands chapter was only reachable from that sentence.

## Changes

- `docs/en/reference/indexes.rst`: added an **introductory "how to create indexes"** block right after the chapter overview — shows `odm:schema:create --index` and `odm:schema:update` console usage before diving into mapping examples, so readers creating indexes can find the commands immediately
- `docs/en/index.rst`: new **"Useful reference pages"** section on the front page linking to both `Indexes` and `Console Commands`, so the console chapter is reachable from the documentation home

## Acceptance criteria

- [x] Creating indexes is discoverable at the start of the chapter, not only at the end
- [x] Console Commands is exposed on the front page

### DIFF
--- a/docs/en/index.rst
+++ b/docs/en/index.rst
@@ -21,9 +21,3 @@ Getting Started
 The best way to get started is with the :doc:`Setup <reference/introduction#setup>` section
 in the introduction tutorial. Use the sidebar to browse other tutorials and documentation
 for the Doctrine PHP MongoDB ODM.
-
-Useful reference pages
-----------------------
-
-- :doc:`Indexes <reference/indexes>` — how to define indexes on your mapped documents and create them
-- :doc:`Console Commands <reference/console-commands>` — ``odm:schema:create``, ``odm:schema:update`` and other command-line tools for schema and index management
--- a/docs/en/index.rst
+++ b/docs/en/index.rst
@@ -21,3 +21,9 @@ Getting Started
 The best way to get started is with the :doc:`Setup <reference/introduction#setup>` section
 in the introduction tutorial. Use the sidebar to browse other tutorials and documentation
 for the Doctrine PHP MongoDB ODM.
+
+Useful reference pages
+----------------------
+
+- :doc:`Indexes <reference/indexes>` — how to define indexes on your mapped documents and create them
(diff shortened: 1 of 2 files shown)


docs/en/index.rst
```diff
@@ -1,23 +1,29 @@
...
+
+The best way to get started is with the :doc:`Setup <reference/introduction#setup>` section
+in the introduction tutorial. Use the sidebar to browse other tutorials and documentation
+for the Doctrine PHP MongoDB ODM.
+
+Useful reference pages
+----------------------
+
+- :doc:`Indexes <reference/indexes>` — how to define indexes on your mapped documents and create them
+- :doc:`Console Commands <reference/console-commands>` — ``odm:schema:create``, ``odm:schema:update`` and other command-line tools for schema and index management
```
GromNaN: From every page, including the [index](https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/latest/index.html), the left bar lists all the pages. I don’t think we need an additional list here.

docs/en/reference/indexes.rst
GromNaN, line 1: The diff shows a full change of the file. What did you modify?

GromNaN (changes requested):
Thank you for looking at this issue.

GromNaN:
Thanks for the documentation improvements. I pushed a follow-up commit (112e3a8b, force-pushed over the earlier 29ff2c02) to your branch:

- **Line endings**: `docs/en/reference/indexes.rst` had been rewritten with CRLF line endings, which turned a ~20 line addition into a full-file rewrite and would have broken `git blame`. It is back to LF, so the diff is now 16 insertions / 7 deletions.
- **Duplication**: the new intro repeated the `odm:schema:create --index` instruction and its console block that already existed in the "Embedded Indexes" section. I removed the later copy since the intro now covers it.
- **Scope**: I reverted the changes to `docs/en/index.rst`, so the PR is now limited to the indexes page.

The wording you added on the indexes page is unchanged.

GromNaN (approved):

GromNaN:
Thank you @TrueFurina

Issue #3038, opened 2026-09-02
Stack overflow ("Maximum call stack size ... Infinite recursion?") after upgrade to 2.17.0


brpauwels:
### Bug Report

|    Q        |   A
|------------ | ------
| Version     | 2.17.0
| Previous version |	2.16.x
| PHP version | 8.5
| ext-mongodb | 2.4.1

#### Summary

After a rebuild that pulled in 2.17.0, one of our API endpoints started failing consistently with:

Uncaught PHP Exception Error: "Maximum call stack size of 8339456 bytes
(zend.max_allowed_stack_size - zend.reserved_stack_size) reached. Infinite recursion?"

Downgrading the library to 2.16.3 resolves it completely.

GromNaN:
Thanks for the report. Without a stack trace it is hard to tell which code path loops, so could you provide a few extra details?

1. The full stack trace of the error. The frames that repeat are what identify the cycle.
2. Does your application register a custom error or deprecation handler, or a logger that writes through the ODM (for example a Monolog handler storing logs in MongoDB via a document)? 2.17 emits new deprecations at runtime, and a handler that itself uses the ODM could re-enter the mapping layer.
3. Which major version of `doctrine/collections` is installed (`composer show doctrine/collections`)? 2.17 added compatibility with v3 and moved several `PersistentCollection` methods into a new trait.
4. If you configure `setPersistentCollectionDir()`, does clearing that directory change anything? Stale generated collection classes from 2.16 would still be loaded.
5. A minimal reproduction (mapping plus the code hitting the endpoint) would help a lot.

For context, the changes between 2.16.3 and 2.17.0 that touch runtime code paths are:

- #3009, which moved `ClassMetadata` properties to property hooks on PHP 8.4+ and deprecates external writes
- #3014, which added `doctrine/collections` 3 compatibility
- #3015 and #3004, which are much smaller

While triaging this I also found a related bug: on PHP 8.4+ the mapping drivers write metadata properties directly, so loading metadata for an embedded document, a mapped superclass, a GridFS file or a query result document triggers the ODM's own deprecation. I will open a separate issue for that.

GromNaN:
Update: the self-triggered deprecation I mentioned above is being reverted in #3040. If your application converts deprecations into exceptions or logs them through a handler that itself uses the ODM, that is a plausible source of unbounded recursion introduced in 2.17, and #3040 would fix it.

[...]

PR #3039, opened 2026-09-02
Avoid triggering the metadata write deprecation from the mapping drivers

GromNaN:
The deprecation added in 2.17 for external writes to `ClassMetadata` properties is triggered by the ODM itself. The attribute driver, the XML driver and `ClassMetadataFactory` write `isMappedSuperclass`, `isEmbeddedDocument`, `isQueryResultDocument`, `isFile` and `isEncrypted` directly, and `ClassMetadata::setPropertyValue()` only allows writes coming from `ClassMetadata.php`.

On PHP 8.4+, loading the metadata of any embedded document, mapped superclass, GridFS file or query result document emits a deprecation, for example:

```
Since doctrine/mongodb-odm 2.17: Writing to property Doctrine\ODM\MongoDB\Mapping\ClassMetadata::isEmbeddedDocument is deprecated and will be removed in version 3.0.
```

This adds `markAsMappedSuperclass()`, `markAsEmbeddedDocument()`, `markAsQueryResultDocument()`, `markAsFile()` and `markAsEncrypted()`, following the naming of the existing `markViewOf()` and `markReadOnly()`, and uses them in the drivers and the factory. Custom drivers now have a replacement API for these writes.

Found while triaging #3038.

### DIFF
--- a/src/Mapping/ClassMetadata.php
+++ b/src/Mapping/ClassMetadata.php
@@ -2021,6 +2021,46 @@ if (PHP_VERSION_ID >= 80400) {
         $this->isReadOnly = true;
     }
 
+    /**
+     * Marks this class as a mapped superclass.
+     */
+    public function markAsMappedSuperclass(): void
+    {
+        $this->isMappedSuperclass = true;
+    }
+
+    /**
+     * Marks this class as an embedded document.
+     */
+    public function markAsEmbeddedDocument(): void
+    {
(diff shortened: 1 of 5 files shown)


GromNaN:
Superseded by #3040, which reverts #3009 instead of adding one-off setters. Adding `markAs*()` methods now would create public API that has to be reworked anyway once a metadata builder carries these operations.

PR #3043, opened 2026-09-04
Create missing search indexes in updateDocumentSearchIndexes()

bbarhoum1:
### Problem

`SchemaManager::updateDocumentSearchIndexes()` drops the search indexes that are no longer mapped, then calls `Collection::updateSearchIndex()` for **every** mapped index. That command fails when the index does not exist, so `odm:schema:update` errors out on any database that does not already carry the mapped search indexes:

```
No index with name globalSearch exists in namespace mydb.MyCollection
```

Two ways to hit it:

* a fresh database — the mapped search indexes have never been created;
* a database where `odm:schema:create --search-index` was never run, or where a search index was dropped by hand.

Because `Schema\UpdateCommand` wraps its phases in a single try/catch, the failure also skips the search-index phase for every class after the failing one.

Reproduced on 2.17.1 and on `2.18.x` (`90c4690d`) against `mongodb/mongodb-atlas-local`: with the mapped search index absent, `odm:schema:update` exits 255; running `odm:schema:create --search-index` first makes the subsequent update succeed. So today `schema:update` cannot converge on its own — it only works if a create ran before it.

### Change

Mapped search indexes that the collection does not have are collected and created with `createSearchIndexes()` instead of being passed to `updateSearchIndex()`; existing ones keep being updated in place. The create path mirrors `createDocumentSearchIndexes()`, including its check that every requested name was reported back as created, so a name the server silently ignores still raises `SchemaException::missingSearchIndex()`.

### Tests

* `testUpdateDocumentSearchIndexesCreatesMissingSearchIndexes` — a mapped index that does not exist is created, `updateSearchIndex()` is not called for it.
* `testUpdateDocumentSearchIndexesMissingIndexNotCreatedError` — a missing index the server does not report back raises `SchemaException`.
* `testUpdateDocumentSearchIndexes` gains `expects($this->never())->method('createSearchIndexes')`, pinning that an existing index is still updated rather than re-created.

Both new tests fail on `2.18.x` without the change. Full suite green locally (3052 tests, 74 skipped — fail points and Atlas-only features unavailable on a single node); `phpcs` clean.

Targeted at `2.18.x` since that is the default branch — happy to rebase onto an older maintenance branch if you prefer.


GromNaN:
Apologies for the state of this pull request, and for the disruption.

While rebasing the branch onto the latest `2.18.x`, an erroneous force push was executed from the wrong working tree. The branch was moved to the bare tip of the target branch, which made the diff empty and caused GitHub to auto-close this PR. That was a mistake on the maintainer side, not a problem with the changes.

The same changes, rebased onto the latest `2.18.x`, have been recreated in a new pull request to preserve the work:

* https://github.com/doctrine/mongodb-odm/pull/3054

Thank you for the contribution, and sorry for the extra work.

Issue #2224, opened 2020-09-30
Update indexes fails if there already exists same index with different name.


endrsmar:
### Bug Report
Running the `doctrine:mongodb:schema:update` command fails with error `Index with name: <index_name> already exists with a different name` if there already exists an index with same keys and options but different name on the collection.

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.1.2

#### Summary

Because name of an index is ignored when checking if index should be dropped in SchemaManager (PR https://github.com/doctrine/mongodb-odm/pull/2114) and because `mongodb` prevets creating duplicate indexes with different names, running schema update after renaming index can lead to failure.

#### Current behavior

Schema update fails with `Index with name: <index_name> already exists with a different name`.

#### How to reproduce

* Create entity mapping with an index
* run `doctrine:mongodb:schema:update`
* rename the index in the mapping
* run `doctrine:mongodb:schema:update`

#### Expected behavior

Drop the existing index, create a new one with the correct name.

malarzm:
@endrsmar thanks for the report! As you can see we are currently struggling with time we can devote to ODM so would you like to work a on fix for your problem? I can offer my help on our Slack's `#mongodb-odm` channel but otherwise I don't know when I'll be able to look into the fixing the issue personally.

bbarhoum1:
Still reproducible on `2.18.x` (`90c4690d`). Opened #3044 with a fix and tests.

Rather than re-introducing the name comparison that #2114 removed, it compares the name **only when the class metadata sets one** — a server-generated name stays ignored (the existing `indexNameAutogenerated` case still expects `true`), while an explicitly mapped name that differs marks the index as not equivalent, so it goes through the normal drop-and-recreate path.

PR #3044, opened 2026-09-04
Respect an explicitly mapped index name when comparing indexes

bbarhoum1:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #2224

### Problem

Since #2114 the index name is excluded from `isEquivalentIndexOptions()` entirely. That is correct for a name the *server* generated — the mapping does not state one, so a difference carries no meaning — but it also means an index whose name the mapping **does** state is judged equivalent to a live index that holds the same key spec under a different name.

`updateDocumentIndexes()` consequently keeps that index instead of dropping it, and the `createIndex()` that immediately follows collides with it:

```
Index already exists with a different name: <old name>
```

`odm:schema:update` exits with an error, and since `Schema\UpdateCommand` wraps its phases in a single try/catch, every class after the failing one is left unprocessed. Renaming a mapped index is unrecoverable without dropping the old index by hand — which is what #2224 reports, still reproducible on `2.18.x` (`90c4690d`).

Same root cause makes it impossible to adopt an index that was created outside the mapping (via the Atlas UI, say) under a name of the mapping's choosing.

### Change

The name is compared only when the class metadata sets one:

```php
if (
    isset($documentIndexOptions['name']) &&
    ($mongoIndexOptions['name'] ?? null) !== $documentIndexOptions['name']
) {
    return false;
}
```

`ClassMetadata::addIndex()` stores the mapping's options verbatim, so `name` is present exactly when the mapping declares it. #2114's behaviour is therefore untouched for generated names — the existing `indexNameAutogenerated` case still expects `true` — while a rename now takes the normal drop-and-recreate path.

### Tests

Three cases added to `dataIsMongoIndexEquivalentToDocumentIndex`:

* `indexNameMappedAndSame` → equivalent
* `indexNameMappedAndDifferent` → not equivalent (the rename in #2224)
* `indexNameMappedAndServerGenerated` → not equivalent (mapping names an index the server named itself)

The latter two fail on `2.18.x` without the change. The drop-then-create path itself is already covered by `testUpdateDocumentIndexesShouldDeleteUnmappedIndexesBeforeCreatingMappedIndexes`, which is why no shared fixture document was given a named index. Full suite green locally (3053 tests, 74 skipped — fail points and Atlas-only features unavailable on a single node); `phpcs` clean.

Targeted at `2.18.x` since that is the default branch — happy to rebase onto an older maintenance branch if you prefer.


GromNaN (commented):
This change is a correct fix for #2224: by comparing the index name only when the mapping explicitly declares one, a mapped rename is no longer judged equivalent to the live index and takes the normal drop-and-recreate path, instead of making `odm:schema:update` fail with `Index already exists with a different name`.

A lighter alternative worth considering as a follow-up is `collMod`. It can change several index properties on an existing index in place, without dropping and rebuilding it: `expireAfterSeconds` (TTL), `hidden`, `unique` and `prepareUnique` (see the "Change Index Properties" section: https://www.mongodb.com/docs/manual/reference/command/collMod/#change-index-properties). `updateDocumentValidator()` already relies on `collMod` for schema updates, so the same mechanism could be extended to the index options that #3045, #3047 and this PR now detect as non-equivalent.

This does not apply to the rename case itself: MongoDB cannot rename an existing index in place, so dropping and recreating remains required for #2224, and the approach taken here is the right one. The `collMod` path would only avoid the rebuild for option-value changes.

One process note: the PR is currently conflicting. Since #3045 and #3047 landed, the docblock of `isEquivalentIndexOptions()` was updated in `2.18.x`, so the branch needs a rebase to become mergeable.

GromNaN (approved):

GromNaN:
Thank you @bbarhoum1

PR #3045, opened 2026-09-04
Detect a changed expireAfterSeconds when comparing indexes

bbarhoum1:
### Problem

`isEquivalentIndexOptions()` never compares the *value* of `expireAfterSeconds`. Its presence is caught — the option is not in `ALLOWED_MISSING_INDEX_OPTIONS`, so `indexOptionsAreMissing()` detects a TTL being added or removed — but once both sides carry it, any retention window is considered equivalent to any other.

So changing a TTL in the mapping leaves the live index in place, and the `createIndex()` that follows collides with it:

```
An equivalent index already exists with the same name but different options.
Requested index: { …, expireAfterSeconds: 5184000 }, existing index: { …, expireAfterSeconds: 2592000 }
```

`odm:schema:update` exits with an error and, because `Schema\UpdateCommand` wraps its phases in a single try/catch, every class after the failing one is left unprocessed. Retuning the TTL of a mapped index is unrecoverable without touching the index by hand. Reproduced on 2.17.1 and `2.18.x` (`90c4690d`).

### Change

The value is compared when both sides have it, so the index takes the normal drop-and-recreate path:

```php
if (
    isset($mongoIndexOptions['expireAfterSeconds'], $documentIndexOptions['expireAfterSeconds']) &&
    $mongoIndexOptions['expireAfterSeconds'] !== $documentIndexOptions['expireAfterSeconds']
) {
    return false;
}
```

### One question for reviewers

This makes a collection whose TTL has drifted from the mapping **rebuild** that index on the next `odm:schema:update`, where previously the command failed. An alternative is `collMod`, which changes the window as metadata with no rebuild — the same mechanism `updateDocumentValidator()` already uses — and would need `updateDocumentIndexes()` to treat a TTL-only difference as "modify" rather than "drop and recreate".

I went with the minimal correctness fix here, but I'm happy to follow up with the `collMod` variant (or replace this PR with it) if that is the direction you'd rather take.

### Tests

Four cases added to `dataIsMongoIndexEquivalentToDocumentIndex`: TTL only in the Mongo index, only in the document index, present in both with different values, present in both with the same value. The mismatch case fails on `2.18.x` without the change; the presence cases document what `indexOptionsAreMissing()` already covers. Full suite green locally (3054 tests, 74 skipped — fail points and Atlas-only features unavailable on a single node); `phpcs` clean.

Targeted at `2.18.x` since that is the default branch — happy to rebase onto an older maintenance branch if you prefer. Note the docblock hunk overlaps with #3044, so whichever lands second will need a trivial rebase.


src/SchemaManager.php
```diff
@@ -968,6 +969,15 @@ private function isEquivalentIndexOptions(IndexInfo $mongoIndex, array $document
             }
         }
 
+        /* A TTL index reports expireAfterSeconds on both sides, so a changed retention window is
+         * not caught by indexOptionsAreMissing() and has to be compared by value. */
+        if (
+            isset($mongoIndexOptions['expireAfterSeconds'], $documentIndexOptions['expireAfterSeconds']) &&
+            $mongoIndexOptions['expireAfterSeconds'] !== $documentIndexOptions['expireAfterSeconds']
+        ) {
```
GromNaN, line 977: It could also detect when this property is added/removed?
```suggestion
        if (($mongoIndexOptions['expireAfterSeconds'] ?? null) !== ($documentIndexOptions['expireAfterSeconds'] ?? null)) {
```

bbarhoum1: add/remove was already detected (indexOptionsAreMissing() runs earlier and expireAfterSeconds isn't in           
  ALLOWED_MISSING_INDEX_OPTIONS)

GromNaN (approved):
Thanks for the fix and the clear write-up. The self-contained null-coalescing form is the right call: it keeps the check correct even if expireAfterSeconds ever moves into ALLOWED_MISSING_INDEX_OPTIONS, and it preserves a TTL of 0.

A follow-up PR will extend value comparison to the other index options that share the same gap (bucketSize, collation, hidden, storageEngine, wildcardProjection, 2dsphereIndexVersion), so this one can stay focused on the TTL regression.

PR #3046, opened 2026-09-10
PHPORM-472: Gate auto-embedding test on MongoDB 8.2 instead of Atlas

paulinevos:
## Summary

PHPORM-472 — Auto embedding in Community Vector Search is available on Community Server from MongoDB 8.2. Per the MQL spec, `\$vectorSearch` and auto-embedding are **not** Atlas-only (only `\$search`/`\$searchMeta` carry that restriction); they are version-gated (`minVersion: 8.2` for the `query`/`model` arguments). So the vector search tests are gated on server version, with the `atlas` group retained so they also run on the Atlas track.

## Changes

- `tests/Tests/BaseTestCase.php`: add `requireMongoDB82()` helper (matches the existing `requireMongoDB63` convention).
- `tests/Tests/Functional/VectorSearchTest.php`:
  - Class-level `#[Group('atlas')]` and `#[Group('vector-search')]` (per review — class-level groups, both groups).
  - Renamed `testAtlasVectorSearch` → `testVectorSearch`, `testAtlasAutoEmbedding` → `testAutoEmbedding`, `testAtlasVectorSearchWithBinaryType` → `testVectorSearchWithBinaryType`.
  - Gate all three on `requireMongoDB82` (skip below 8.2).
  - Catch `SearchNotSupportedException` (skip when the server has no search engine) in `testVectorSearch` and `testAutoEmbedding`; keep the `"not registered"` catch in `testAutoEmbedding` for servers with search but no registered embedding model.
- `src/SchemaManager.php`: declare `@throws SearchNotSupportedException` on `createDocumentSearchIndexes` (propagated from the library) so PHPStan accepts the catch.
- `.github/workflows/atlas-ci.yml`: `--group atlas` (single invocation) — the vector-search tests are in the `atlas` group, so this selects them.
- `.github/workflows/continuous-integration.yml`: add a Community Server 8.2 matrix entry with `local-atlas: true` (search engine enabled), and a conditional `--group vector-search` step that runs the vector-search tests on that job so they also run on Community 8.2 — not only on the Atlas track.

## Behavior

- Atlas track (`mongodb-atlas-local` + `VOYAGE_API_KEY`): all three tests pass (on branches with the secret; fork PRs skip `testAutoEmbedding`).
- Community 8.2 (`local-atlas`, search enabled): `testVectorSearch` and `testVectorSearchWithBinaryType` pass; `testAutoEmbedding` skips without `VOYAGE_API_KEY`.
- Community < 8.2: tests are excluded via the `atlas` group (`--exclude-group=atlas`) and the conditional vector-search step does not run.

A plain community `mongod` (no search engine) throws `SearchNotSupportedException`, which the tests catch and skip — vector search is supported on Community 8.2 but requires the server to be started with the search engine (`mongodb-atlas-local`).

## CI

All checks pass (fork PRs don't receive `VOYAGE_API_KEY`, so `testAutoEmbedding` skips here; it passes on branches with the secret):
- Atlas CI — ✅ (8.1 / 8.2 / 8.4)
- Continuous Integration — ✅ (8.2 `local-atlas` job runs the vector-search tests: `testVectorSearch` and `testVectorSearchWithBinaryType` pass, `testAutoEmbedding` skips)
- Coding Standards — ✅
- Static Analysis — ✅

## Still pending for PHPORM-472

- `testAutoEmbedding` passing (not skipping) on a Community 8.2 CI job needs `VOYAGE_API_KEY` + `EMBEDDING_PROVIDER_ENDPOINT` to reach the container; `drivers-evergreen-tools`' `local-atlas` mode does not currently forward those env vars. An upstream PR to `mongodb-labs/drivers-evergreen-tools` would let it pass too.
- Reference docs are handled by the docs team.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -117,6 +117,17 @@ jobs:
             dependencies: "highest"
             symfony-version: ""
             proxy: "lazy-ghost"
+          # Test against Community Server 8.2, which introduces vector search
+          # and auto-embedding support (PHPORM-472). Vector search needs the
+          # search engine, enabled via mongodb-atlas-local (local-atlas mode).
+          - topology: "server"
+            php-version: "8.4"
+            mongodb-version: "8.2"
+            driver-version: "stable"
+            dependencies: "highest"
+            symfony-version: ""
+            proxy: "native"
+            local-atlas: "true"
 
     steps:
       - name: "Checkout"
@@ -180,9 +191,25 @@ jobs:
         with:
           version: ${{ matrix.mongodb-version }}
           topology: ${{ matrix.topology }}
+          # matrix.local-atlas is only set for the 8.2 entry; an empty string
+          # is treated as false by the action, so do not default it to 'false'
+          # (a non-empty string would be read as truthy and enable local-atlas).
+          local-atlas: ${{ matrix.local-atlas }}
 
       - name: "Run PHPUnit"
         run: "vendor/bin/phpunit --exclude-group=atlas ${{ matrix.dependencies == 'lowest' && '--do-not-fail-on-deprecation --do-not-fail-on-warning --do-not-fail-on-notice' || '' }}"
+        env:
+          DOCTRINE_MONGODB_SERVER: ${{ steps.setup-mongodb.outputs.cluster-uri }}
+          USE_LAZY_GHOST_OBJECT: ${{ matrix.proxy == 'lazy-ghost' && '1' || '0' }}
+          USE_NATIVE_LAZY_OBJECT: ${{ matrix.proxy == 'native' && '1' || '0' }}
+          CRYPT_SHARED_LIB_PATH: ${{ steps.setup-mongodb.outputs.crypt-shared-lib-path }}
+
+      # Run the vector-search tests on Community Server 8.2 (local-atlas), where the
+      # search engine is enabled. They are excluded from the run above by the
+      # "atlas" group, so select them explicitly here.
+      - name: "Run vector search tests"
+        if: ${{ matrix.local-atlas == 'true' }}
+        run: "vendor/bin/phpunit --group vector-search"
         env:
           DOCTRINE_MONGODB_SERVER: ${{ steps.setup-mongodb.outputs.cluster-uri }}
           USE_LAZY_GHOST_OBJECT: ${{ matrix.proxy == 'lazy-ghost' && '1' || '0' }}
--- a/tests/Tests/Functional/VectorSearchTest.php
+++ b/tests/Tests/Functional/VectorSearchTest.php
@@ -11,15 +11,17 @@ use Documents\VectorEmbedding;
 use MongoDB\BSON\Binary;
 use MongoDB\Driver\Exception\CommandException;
 use MongoDB\Driver\WriteConcern;
+use MongoDB\Exception\SearchNotSupportedException;
 use PHPUnit\Framework\Attributes\Group;
 use PHPUnit\Framework\Attributes\RequiresPhpExtension;
 
 use function str_contains;
 
 #[Group('atlas')]
+#[Group('vector-search')]
 class VectorSearchTest extends BaseTestCase
 {
-    public function testAtlasVectorSearch(): void
+    public function testVectorSearch(): void
     {
         // Create the collection by ensuring the schema
         $schemaManager = $this->dm->getSchemaManager();
@@ -50,7 +52,11 @@ class VectorSearchTest extends BaseTestCase
         $this->dm->flush(['writeConcern' => new WriteConcern(WriteConcern::MAJORITY)]);
 
         // Index must be created after data insertion, so the index status is not immediately "READY"
-        $schemaManager->createDocumentSearchIndexes(VectorEmbedding::class);
+        try {
+            $schemaManager->createDocumentSearchIndexes(VectorEmbedding::class);
+        } catch (SearchNotSupportedException) {
+            $this->markTestSkipped('Vector search requires a server with search support (Community Server 8.2+ or Atlas)');
+        }
 
         // Wait for the search index to be ready (Atlas Local needs time to build the index)
         $schemaManager->waitForSearchIndexes([VectorEmbedding::class]);
@@ -104,7 +110,7 @@ class VectorSearchTest extends BaseTestCase
         }
     }
 
-    public function testAtlasAutoEmbedding(): void
+    public function testAutoEmbedding(): void
     {
         $schemaManager = $this->dm->getSchemaManager();
         $schemaManager->createDocumentCollection(AutoEmbeddingArticle::class);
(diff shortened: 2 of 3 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -117,6 +117,15 @@ jobs:
             dependencies: "highest"
             symfony-version: ""
             proxy: "lazy-ghost"
+          # Test against Community Server 8.2, which introduces vector search
+          # and auto-embedding support (PHPORM-472).
```
paulinevos: Remove comment; not relevant for the future.

tests/Tests/Functional/VectorSearchTest.php
```diff
@@ -16,9 +16,9 @@
 
 use function str_contains;
 
-#[Group('atlas')]
 class VectorSearchTest extends BaseTestCase
 {
+    #[Group('atlas')]
```
paulinevos: I don't think these make sense actually, because vector search is also supported on community... I think we can remove this gate altogether and just require server >= 8.2.

tests/Tests/Functional/VectorSearchTest.php
```diff
@@ -11,16 +11,19 @@
...
 class VectorSearchTest extends BaseTestCase
 {
-    public function testAtlasVectorSearch(): void
+    #[Group('vector-search')]
+    public function testVectorSearch(): void
     {
+        $this->requireMongoDB82('Vector search requires MongoDB 8.2 or higher');
+
         // Create the collection by ensuring the schema
         $schemaManager = $this->dm->getSchemaManager();
```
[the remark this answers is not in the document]
paulinevos: Isn't the point that the community server does support it now

tests/Tests/Functional/VectorSearchTest.php
```diff
@@ -11,16 +11,19 @@
 use MongoDB\BSON\Binary;
 use MongoDB\Driver\Exception\CommandException;
 use MongoDB\Driver\WriteConcern;
+use MongoDB\Exception\SearchNotSupportedException;
 use PHPUnit\Framework\Attributes\Group;
 use PHPUnit\Framework\Attributes\RequiresPhpExtension;
 
 use function str_contains;
 
-#[Group('atlas')]
```
GromNaN, line 19: Groups on the class doesn't work? Why duplicating the attribute on every methods of the class?

tests/Tests/Functional/VectorSearchTest.php
```diff
@@ -11,16 +11,19 @@
...
 class VectorSearchTest extends BaseTestCase
 {
-    public function testAtlasVectorSearch(): void
+    #[Group('vector-search')]
```
GromNaN: This test also works on atlas. You can add both groups?

tests/Tests/BaseTestCase.php
```diff
@@ -270,6 +270,11 @@ protected function requireMongoDB63(string $message): void
         $this->requireVersion($this->getServerVersion(), '6.3.0', '<', $message);
     }
 
+    protected function requireMongoDB82(string $message): void
```
GromNaN: This could be `requireAtlasOrServerVesion82`

tests/Tests/Functional/VectorSearchTest.php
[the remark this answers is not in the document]
paulinevos: (update: yes but if we run locally it might fail even if on CI this won't be a problem because CI uses community server)

.github/workflows/atlas-ci.yml
```diff
@@ -94,7 +94,7 @@ jobs:
           SYMFONY_REQUIRE: ${{ matrix.symfony-version }}
 
       - name: "Run PHPUnit with Atlas Local"
-        run: "vendor/bin/phpunit --group atlas"
+        run: "vendor/bin/phpunit --group atlas && vendor/bin/phpunit --group vector-search"
```
GromNaN: The "atlas" group is for tests that run on atlas, that should include the vector-search tests.

tests/Tests/Functional/VectorSearchTest.php
paulinevos: Because Claude thought at a first iteration that the other tests should only be run on Atlas until I crrected it

tests/Tests/Functional/VectorSearchTest.php
```diff
@@ -50,7 +54,11 @@ public function testAtlasVectorSearch(): void
         $this->dm->flush(['writeConcern' => new WriteConcern(WriteConcern::MAJORITY)]);
 
         // Index must be created after data insertion, so the index status is not immediately "READY"
-        $schemaManager->createDocumentSearchIndexes(VectorEmbedding::class);
+        try {
+            $schemaManager->createDocumentSearchIndexes(VectorEmbedding::class);
+        } catch (SearchNotSupportedException) {
+            $this->markTestSkipped('Vector search requires search support, available on Atlas or an Atlas Local deployment (MongoDB 8.2+)');
```
GromNaN: Hum, the goal is to run the tests on community 8.2+, right? Then this skip message should say that. I understood that vector search did not require search anymore.

tests/Tests/Functional/VectorSearchTest.php
```diff
@@ -11,16 +11,20 @@
...
 use function str_contains;
 
 #[Group('atlas')]
+#[Group('vector-search')]
 class VectorSearchTest extends BaseTestCase
 {
-    public function testAtlasVectorSearch(): void
+    public function testVectorSearch(): void
     {
+        $this->requireMongoDB82('Vector search requires MongoDB 8.2 or higher');
```
GromNaN: This is false, because vector search works on older versions of atlas/atlas-local.

GromNaN (changes requested):
This whole changes are confusing.
The ticket says we can run vector search tests on new server versions (community 8.2+). It should not add restrictions for Atlas or Atlas Local.

tests/Tests/Functional/VectorSearchTest.php
paulinevos: I think Claude conflated "community server" with "atlas local" which then confused me into thinking I got wrong what "community server" was the entire time.

.github/workflows/continuous-integration.yml
```diff
@@ -117,6 +117,17 @@ jobs:
             dependencies: "highest"
             symfony-version: ""
             proxy: "lazy-ghost"
+          # Test against Community Server 8.2, which introduces vector search
+          # and auto-embedding support (PHPORM-472). Vector search needs the
+          # search engine, enabled via mongodb-atlas-local (local-atlas mode).
+          - topology: "server"
+            php-version: "8.4"
+            mongodb-version: "8.2"
```
GromNaN, line 125: Instead, you can update all the `mongodb-version: "8.0"` to `mongodb-version: "8.2"`. This way, vector search is tested in all typologies and configurations.

paulinevos: But we don't want to run anything on older versions then?

GromNaN: The default matrix run on `mongodb-version: "8.0", "7.0", "6.0"`. This is already a good coverage.

PR #3048, opened 2026-09-11
Add `SortDirection` support to query and aggregation sort orders

GromNaN:
Add support for the global `SortDirection` enum (PHP 8.6 natively, or `symfony/polyfill-php86` for PHP 8.1 and later) everywhere a sort direction is accepted, alongside the existing `asc`/`desc` strings and int (`1`/`-1`) values. Float values such as `1.0` and `-1.0` are accepted as well. Strings are not deprecated.

Sort values are normalized by a single helper, `Doctrine\ODM\MongoDB\Utility\SortHelper::normalizeSortDirections()`, which maps `SortDirection::Ascending` and `SortDirection::Descending` to 1 and -1.

Covered entry points:
- `DocumentRepository::findBy()` and `findOneBy()` orderBy, via `DocumentPersister::prepareSort()`
- Query Builder `sort()`
- `Expr::sort()` (the `$sort` operator of a `$push` update)
- Aggregation stages `Sort`, `Search`, `SetWindowFields::sortBy()` and `Fill::sortBy()`

The `instanceof SortDirection` guard is safe when the enum is not installed, so `symfony/polyfill-php86` stays a regular dependency.

### Stricter validation of sort values

Centralizing the normalization surfaced inconsistencies in how invalid values were handled: query sorts passed unknown strings through to the server, which rejected them with a confusing error, while aggregation sorts silently converted any unknown string to `-1`. All entry points now fail fast instead:

- Any sort value other than `1`, `-1`, `asc`/`desc`, a `SortDirection` case, an allowed `$meta` keyword or a `['$meta' => ...]` expression throws an `InvalidArgumentException` naming the field and the invalid value.
- The aggregation `$sort` stage accepts the `textScore`, `searchScore` and `vectorSearchScore` keywords and wraps them as `$meta` expressions. Previously only `textScore` was recognized; `searchScore` was silently converted to a `-1` direction.
- The Query Builder `sort()` also accepts the `textScore` keyword, consistent with the sort `find()` supports.
- `$meta` expressions are validated: only a single `['$meta' => keyword]` entry whose keyword is allowed on this entry point is accepted.

No working usage changes behavior: every value that now throws was previously either rejected by the server or silently produced a wrong sort.

### Bug fixes

- Calling `sort($field)` without an order in the Aggregation Builder, `Expr::sort()` and `Search::sort()` now defaults to ascending. Previously `null` was sent to the server, which rejected it.
- `DocumentPersister::prepareSort()` no longer triggers a `TypeError` for integer-like field names such as `'2024'` under `strict_types`.
- XML mapping `sort` orders (`order="1"`, `order="-1"`, `order="asc"`, `order="desc"`) are now normalized to the integers `1` and `-1` at the metadata driver level. They previously produced a string sort value that the server rejected or that sorted in the wrong direction.


src/Persisters/DocumentPersister.php
```diff
@@ -921,43 +922,22 @@ public function prepareProjection(array $fields): array
...
      *
-     * @param array<string, int|string|array<string, string>> $fields
+     * @param array<string, int|string|SortDirection|array<string, string>> $fields
+     * @param list<string>                                                  $allowedMetaSort
      * @phpstan-param SortShape $fields
      *
      * @phpstan-return array<string, -1|1|SortMeta>
      */
-    public function prepareSort(array $fields): array
+    public function prepareSort(array $fields, array $allowedMetaSort = []): array
```
GromNaN, line 935: Final class, I can add a parameter.

alcaeus (approved):

src/Aggregation/Stage/Fill.php
```diff
@@ -65,22 +64,16 @@ public function partitionByFields(string ...$fields): static
...
         $fields = is_array($fieldName) ? $fieldName : [$fieldName => $order ?? 1];
 
-        foreach ($fields as $fieldName => $order) {
-            if (is_string($order)) {
-                $order = strtolower($order) === 'asc' ? 1 : -1;
-            }
-
-            $this->sortBy[$fieldName] = $order;
-        }
+        $this->sortBy = SortHelper::normalizeSortDirections($fields);
```
IonBazan: This overwrites previously set sortBy which is a breaking change: `->sortBy('a')->sortBy('b', 'desc')` will produce `{"b":-1}` instead of `{"a":1,"b":-1}` like before.

src/Aggregation/Stage/Search.php
```diff
@@ -173,28 +173,18 @@ public function searchAfter(string $searchAfter): static
...
-                if (in_array($order, $allowedMetaSort, true)) {
-                    $order = ['$meta' => $order];
-                } else {
-                    $order = strtolower($order) === 'asc' ? 1 : -1;
-                }
-            }
-
-            $this->sort[$fieldName] = $order;
-        }
+        $this->sort = array_replace($this->sort, SortHelper::normalizeSortDirections($fields, $allowedMetaSort));
```
IonBazan, line 187: This is different from `Fill::sortBy()` and it's correct.

tests/Tests/Query/BuilderTest.php
```diff
@@ -712,6 +713,8 @@ public static function provideSortOrders(): array
             ['desc', -1],
             ['ASC', 1],
             ['DESC', -1],
+            [SortDirection::Ascending, 1],
+            [SortDirection::Descending, -1],
```
IonBazan:
```suggestion
            [SortDirection::Descending, -1],
            ['-1', -1],
            ['1', 1],
            [1.0, 1],
            [-1.0, -1],
```

Add these as additional regression coverage. I believe XmlDriver can produce such `string` `'1'` value when using `<sort field="x" order="-1"/>`. This will now throw exception.

src/Aggregation/Stage/Sort.php
```diff
@@ -6,53 +6,48 @@
...
  */
 class Sort extends Stage
 {
+    /**
+     * Sort keywords that are converted to a $meta expression instead of a
+     * sort direction.
+     *
+     * @see https://www.mongodb.com/docs/manual/reference/operator/aggregation/meta/
+     */
+    public const ALLOWED_META_SORTS = ['textScore', 'searchScore', 'vectorSearchScore'];
```
IonBazan, line 33: How about `indexKey`?

src/Utility/SortHelper.php
```diff
@@ -0,0 +1,67 @@
...
+        foreach ($fields as $fieldName => $order) {
+            $normalized[$fieldName] = match (true) {
+                $order === 1, $order === -1 => $order,
+                $order instanceof SortDirection => $order === SortDirection::Ascending ? 1 : -1,
+                is_array($order) => $order,
+                is_string($order) && in_array($order, $allowedMetaSort, true) => ['$meta' => $order],
+                is_string($order) && strtolower($order) === 'asc' => 1,
+                is_string($order) && strtolower($order) === 'desc' => -1,
+                default => throw new InvalidArgumentException(sprintf(
+                    'Invalid sort order %s for field "%s". Allowed values are 1, -1, "asc", "desc" and SortDirection cases.',
```
IonBazan: We could add info about the `$meta` here too.

src/Query/Builder.php
```diff
@@ -1483,14 +1482,8 @@ public function sort($fieldName, $order = 1): self
...
 
-        foreach ($fields as $fieldName => $order) {
-            if ($order instanceof SortDirection) {
-                $order = $order === SortDirection::Ascending ? 1 : -1;
-            } elseif (is_string($order)) {
-                $order = strtolower($order) === 'asc' ? 1 : -1;
-            }
-
-            $this->query['sort'][$fieldName] = (int) $order;
+        foreach (SortHelper::normalizeSortDirections($fields) as $fieldName => $order) {
```
IonBazan: Do we need to pass `$allowedMetaSort` here too?

src/Aggregation/Builder.php
```diff
@@ -628,15 +629,15 @@ public function skip(int $skip): Stage\Skip
...
      * @phpstan-param SortShape|string $fieldName Field name or array of field/order pairs
      */
     public function sort($fieldName, $order = null): Stage\Sort
     {
-        $fields = is_array($fieldName) ? $fieldName : [$fieldName => $order];
-        // fixme: move to sort stage
-        $stage = new Stage\Sort($this, $this->getDocumentPersister()->prepareSort($fields));
+        $fields = is_array($fieldName) ? $fieldName : [$fieldName => $order ?? 1];
+        // The persister is needed here to map PHP field names to database field names
+        $stage = new Stage\Sort($this, $this->getDocumentPersister()->prepareSort($fields, Stage\Sort::ALLOWED_META_SORTS));
```
IonBazan, line 640: We already normalise this in the `Stage\Sort` constructor too. Is it still required to normalise here?

IonBazan (changes requested):

src/Aggregation/Stage/Fill.php
GromNaN: Good catch! Thank you for your keen eye.

tests/Tests/Query/BuilderTest.php
GromNaN: Test cases added.

src/Aggregation/Stage/Sort.php
GromNaN: Added

src/Utility/SortHelper.php
GromNaN: Added

src/Aggregation/Builder.php
GromNaN: `DocumentPersister::prepareFieldName` calls `prepareFieldName` to map the field names. This can't be done inside the Sort class.

src/Aggregation/Stage/Sort.php
IonBazan: Sorry I think I was wrong here.
`indexKey` can't be used to sort. It only works in projections. MongoDB rejects it in both `find()` and `$sort`:
```
Illegal $meta sort: $meta: "indexKey"
```

So `$qb->sort('x', 'indexKey')` now builds a query that always fails on the server. On 2.18.x it became a `-1` direction and the query ran.

src/Utility/SortHelper.php
```diff
@@ -0,0 +1,84 @@
...
+     * @throws InvalidArgumentException
+     */
+    private static function normalizeSortOrder(mixed $order, array $allowedMetaSort, int|string $fieldName): int|array
+    {
+        if (is_array($order)) {
+            return $order;
+        }
+
+        if (is_string($order) && in_array($order, $allowedMetaSort, true)) {
+            return ['$meta' => $order];
```
IonBazan, line 68: Arrays are returned without any check, so the new validation only covers strings. These all pass through and then fail on the server:
```php
$qb->sort('f', ['$meta' => 'nonExistentScore']);
$qb->sort('f', ['bogus' => 'nonsense']);
$qb->sort('f', []);
```

It also makes the stages inconsistent. `Fill::sortBy('f', 'searchScore')` correctly throws, but `Fill::sortBy('f', ['$meta' => 'searchScore'])` is accepted, even though `$fill` doesn't support `$meta `at all. And for the query builder this is a new regression: 2.18.x turned these arrays into `1`, so the query ran.

We should only accept an array if it has a single `$meta` key whose value is in `$allowedMetaSort`.

Edit: The comment location is wrong - it relates to line 67

tests/Tests/Query/BuilderTest.php
IonBazan: I prepared a test for mapping here: https://github.com/IonBazan/mongodb-odm/commit/1c97ec6938f013ade43d064d947de33802251a58 - looks like this was silently failing for `1` on owning side:
`server error: Illegal key in $sort specification: position: "1"`
and swapping to descending on inverse side on 2.18.x. That means it was never actually working. 

Now I wonder if this should be fixed at the metadata mapper layer instead. I believe you can revert the `'1'`/`'-1'` support for now so that behaviour stays same and we can fix that separately.

Sorry again for the confusion.

IonBazan (approved):

tests/Tests/Query/BuilderTest.php
GromNaN: Thanks for the test, it helped.

I reverted the `'1'`/`'-1'` string support in `SortHelper`, so these values are rejected again. Instead, the XML driver now normalizes the mapped `sort` orders to `1`/`-1` at the metadata layer, and I imported your mapping test for the owning and the inverse side. Both pass.

IonBazan (approved):
LGTM

PR #3049, opened 2026-09-13
Improve performance test suite

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | N/A

#### Summary

The benchmark suite only covered a handful of shapes for hydration, load and store, and had no coverage at all for metadata loading, the query builder or the aggregation builder. This PR fills those gaps.

- Metadata loading: cold loading (driver parsing + `ClassMetadataFactory` construction, single class and the whole fixture set) vs. cache-hit lookups, so the two costs are told apart.
- Query builder: `Builder`/`Expr` construction in isolation vs. execution (find, hydration disabled, reference priming).
- Aggregation builder: construction in isolation vs. execution (hydrated/raw results, `$lookup`).
- Hydration: nested and larger embed-many collections, a polymorphic (discriminated) embed-many, and a broader set of built-in types.
- Load: identity-map fast path, `findOneBy()`/`findBy()` as distinct persister paths from `find()`, lazy collection initialization.
- Store: updates (scalar and embedded collection), `computeChangeSets()` in isolation, batch insert, removal.
- A wide/deeply-nested document shape mirroring the "small_doc"/"large_doc_nested" fixtures from the MongoDB driver benchmark corpus, to see how document scale affects hydrate/store/load independent of association type.

New fixtures live under `benchmark/Fixtures` rather than `tests/Documents`, so metadata-loading benchmarks stay unaffected by unrelated changes to the test suite's document fixtures. `BaseBench::createMetadataDriverImpl()` is now resolved via `static::` instead of `self::`, which was the one behavioural change needed to let subclasses point at that directory.

#### AI Notice

<details>
<summary>This PR was built with Claude Code — click to expand the session recap.</summary>

I described the gaps in the existing benchmark suite (no coverage for metadata loading, the query builder or the aggregation builder; light coverage for hydration/load/store) and asked Claude to propose a list of benchmarks to fill them before implementing anything.

It surveyed `benchmark/BaseBench.php` and the three existing `Document/*Bench.php` files, then the relevant library surface (`ClassMetadataFactory`, `Query\Builder`, `Aggregation\Builder`, `UnitOfWork`), and came back with ~26 candidate benchmarks grouped by area, flagging which ones it thought were highest-value for the DocumentManager/UnitOfWork work I mentioned I'm doing next.

I asked it to implement all of them, pointing it at a folder of MongoDB driver benchmark corpus sample data (`small_doc.json`, `large_doc_nested.json`) for the document-scale benchmarks, and asked it to use separate fixtures rather than the test suite's `Documents` classes for metadata-loading benchmarks specifically, so those wouldn't be affected by unrelated test fixture changes later.

It built a new `benchmark/Fixtures` document set (simple/inheritance/embedded/referenced/polymorphic shapes, plus flat and deeply-nested documents mirroring the sample data), copied the two JSON files into the repo, and wrote the new benchmark classes. It needed one small change to `BaseBench` (`self::` → `static::` for the metadata driver hook) so subclasses could point at the new fixtures directory. Along the way it ran the actual benchmark suite against a local MongoDB container and iterated on a few mapping mistakes it hit (an `array`-typed property that should have been a `Collection`, an `id` typed as `ObjectId` instead of `string`, a couple of aggregation builder method-chaining errors) until all 51 benchmark subjects ran clean, then checked PHPCS and PHPStan.

Finally I asked it to split the work into independent commits for easier review and draft this description, using my recent merged PRs as a style reference.
</details>


alcaeus:
atlas-local CI failure is unrelated:
> The action 'Create MongoDB Atlas Local' has timed out after 5 minutes.

GromNaN:
I've noticed significant variations in job execution times in GitHub Actions. It's not possible to meaningfully compare one job to another.
Accessing performance results for jobs isn't easy, and, more importantly, it's not centralized.
Aside from verifying that the benchmark code works, I don't see the current value in running these tests in CI.

alcaeus:
> Aside from verifying that the benchmark code works, I don't see the current value in running these tests in CI.

Yeah, the advantage is currently confirming that nothing broke. I'm thinking about how we can use the tests to protect against regressions -- the only way I can think of is checking out the latest stable release, running benchmarks to store a baseline, then running against the current patch. Since this is a considerable effort and will bring its own challenges I haven't bothered with this yet. I did want to expand on the performance tests since I'll be working on some larger refactoring that might affect performance in near future; those PRs would do well by at least confirming that there are no regressions.

benchmark/Document/LoadDocumentBench.php
```diff
@@ -86,6 +97,32 @@ public function benchLoadReferenceMany(): void
...
+    {
+        // Warm the identity map, then load again without an intervening
+        // clear() so the second call hits UnitOfWork::tryGetById().
+        $this->loadDocument();
+        $this->loadDocument();
+    }
+
+    public function benchLoadDocumentByQuery(): void
+    {
+        $this->getDocumentManager()->clear();
```
GromNaN: The clear execution is part of the measured time, but the method description indicates you want to benchmark only the query.

alcaeus: Fixed — `findOneBy()`/`findBy()` always query MongoDB regardless of identity-map state, so `clear()` was unnecessary here and removed. `benchLoadReferenceManyCollectionInitialization` genuinely needed a cold identity map to force a real lazy-collection init, so instead of a timed `clear()` it now reads from a pool of pre-seeded documents.

alcaeus:
Pushed a follow-up round: fixed PHP 8.1 CI failure, addressed @GromNaN's `clear()` timing comment (reply above), fixed a rev/iteration bug where revolutions 2+ were silently measuring "skipped hydration" instead of a real load, and stopped benchmarking against the kitchen-sink `Documents\User` in favor of a dedicated lean fixture — a couple of benchmarks dropped from ~19ms to ~2-4ms as a result of that last one. All 51 subjects pass, PHPCS/PHPStan clean.

<details>
<summary>Full details</summary>

- **PHP 8.1 fix**: trait constants require PHP 8.2+; `MappingBenchTrait` used one. Inlined the value instead — CI on 8.1 was failing on this.
- **Reworked revs/iterations across the DB-bound benchmarks.** 100 revs mostly compensates for `microtime()` resolution noise on sub-microsecond in-memory work; once a call is a real MongoDB round trip it already dwarfs that noise, so high revs was just adding CI time. Worse, it was actively wrong for any benchmark reading the same document repeatedly: `UnitOfWork::getOrCreateDocument()` skips hydration entirely for a document that's already managed and initialized, so revolutions 2+ were silently measuring "driver round trip plus skipped hydration," not a real load. Benchmarks in that category (`LoadDocumentBench`, the hydrated subjects in `ExecuteQueryBench`/`ExecuteAggregationBench`/`LargeDocumentBench`) are now pinned to a single un-warmed-up revolution per iteration — each iteration already runs in its own process, so that alone guarantees a cold identity map — with iterations raised to recover statistical power. Everything else got a rev count sized for a network round trip rather than an in-memory call.
- **Stopped benchmarking against `Documents\User`.** It's the test suite's kitchen-sink fixture with ten `EmbedMany`/`ReferenceMany` fields, and its generated hydrator unconditionally constructs a `PersistentCollection` for every one of them on every hydration, regardless of whether that field has data. That overhead dominated `User`'s hydration cost and scaled with document count, so benchmarks touching more than one document (e.g. `benchExecuteAggregationHydrated`, `benchExecuteFindQueryWithReferencePriming`) were measuring `User`'s structural weight, not aggregation or reference priming. `HydrateDocumentBench`, `LoadDocumentBench`, `StoreDocumentBench`, `ExecuteQueryBench` and `ExecuteAggregationBench` now use `Fixtures\RichDocument` instead, which has exactly one of each association type. Both `benchExecuteAggregationHydrated` and `benchExecuteFindQueryWithReferencePriming` dropped from ~19ms to ~2-4ms as a result.
- Also found and fixed a real bug in `benchUpdateDocumentWithEmbedMany`: it mutated an embedded field to the same hardcoded value on every revolution, so after the first revolution the value already matched and `flush()` silently no-opped for the rest — understating that benchmark by about an order of magnitude. It now writes a fresh value each time.

</details>

GromNaN (approved):

PR #3050, opened 2026-09-14
Cleanup PHPStan baseline

alcaeus:
Works through the PHPStan baseline in four tiers, ordered by risk (bug-hunt first, mechanical typing last), fixing what's safe to fix now and leaving a scoped list of what still needs a design decision.

Net result: **211 → 75 baseline entries** (1267 → 451 lines), two real bugs fixed along the way.

## What changed

**Tier 1 — bug hunt** (entries that PHPStan flags as "always true/false/unreachable", since those are the ones most likely to hide a real bug rather than a false positive):
- `PersistentCollectionTrait::__clone()` dropped a redundant `is_object()` guard around cloning `$coll` — the property is non-nullable, so the check could never be false.
- `DocumentPersister::createReferenceManyWithRepositoryMethodCursor()` dropped a redundant `assert(is_array($primers))` left over from pre-PHP8 `array_combine()` behavior (which used to return `false` on mismatched array lengths; on PHP 8+ it always returns `array` or throws).
- Removed 3 baseline entries that no longer reproduced at all — stale from files that had moved/been rewritten since the baseline was last generated.
- Confirmed the rest of this category (7 entries) are legitimate false positives — deprecation/BC-compat shims and one `@final`-trust situation in `ClassMetadata` — and left them baselined as-is.

**Tier 2 — mechanical cleanup** (near-zero-risk: unused property types, write-only properties, stale `@throws`, etc.):
- Found and fixed a real bug while investigating a "never read" property: `Aggregation\Stage\Search\Range::lt()`/`lte()` were copy-pasted from `gt()`/`gte()` and wrote to `$includeLowerBound` instead of `$includeUpperBound`. Mixing an exclusive/inclusive lower bound with an exclusive/inclusive upper bound (e.g. `->gt(5)->lte(10)`) silently produced the wrong Atlas Search operator. Added a regression test (`SearchTest::provideRangeBuilders` — "Range with mixed bound inclusivity") that fails against the old code.
- ~15 dead-code/unused-type removals: `GeoWithin::$relation` (copy-paste leftover from `GeoShape`, never wired up), stale `@throws` tags, `static::` → `self::` for private method calls, an over-wide return type on `ClassMetadata::getBucketName()`, a no-op statement in `FunctionalTest`.

**Tier 3 — typing sweep** (the bulk of the baseline: `missingType.generics`/`missingType.iterableValue`/`missingType.parameter`, ~114 entries), split into 3 commits:
- Batch 1: `Iterator<TValue>` and `PersistentCollectionInterface<TKey, T>` generics across the persister layer, plus `ReflectionClass<T>` and `InternalProxy<T>`.
- Batch 2: array/parameter types across the Aggregation search operators (`GeoShape`/`GeoWithin`/`Near`/`Fill`/`VectorSearch`/`UnionWith`/the 11 `Compounded*` search classes, fixed via one shared trait).
- Batch 3: parameter/return types on test data-provider methods.
- Plus a follow-up commit removing one baseline entry that batch 1 had already fixed in code but the cleanup script missed.

**Final pass — Tier 4 triage:** went through all 75 remaining entries looking for anything fixable without a design decision. Found and removed one more stale cluster (5 entries on `PersistentCollection.php`, left over from the same earlier refactor as the Tier 1 stale entries). Everything else was investigated and confirmed to need a real decision — see below.

## What was deliberately left alone, and why

- **`Query::getIterator()`/`makeIterator()`/`$iterator` (3 entries).** I tried adding `Iterator<object>` here in Tier 3. That let PHPStan "prove" `Query::getSingleResult()` never returns `null` — which is wrong: `Iterator::current()` can genuinely yield nothing on an empty result, and the `Iterator<TValue>` interface's contract doesn't model that possibility (it declares `current(): TValue` with no allowance for the exhausted/empty case). Reverted; fixing this properly means revisiting that shared interface contract, not just this one call site.
- **`DateImmutableType`'s unreachable `throw`.** PHPStan can prove it's dead (`DateTimeInterface` only has two possible concrete implementations), but per your call this stays as a defensive fallback.
- **11× "unable to resolve template type T" on `getClassMetadata()`.** Root cause: `targetDocument` in the mapping array is typed `class-string`, not `class-string<T>`, so PHPStan has nothing to bind the template to. Fixing this means changing how mapping arrays carry type information, not a local fix. `ReferencePrimer::addManyReferences()`'s by-ref type mismatch traces back to this same root cause.
- **7 entries around `ClassMetadata`'s `FieldMapping` array shape** (`mapField()`, `checkDuplicateMapping()`, `setVersionMapping()`, `typeRequirementsAreMet()`, plus 3 in `XmlDriver`). The shape declares required keys that several call sites only partially populate. Narrowing it safely needs a pass over every caller, not a quick annotation fix.
- **3× `generics.variance` + `DocumentManager`'s `method.childReturnType`/`new.static`.** Real API/LSP design questions about the `ClassMetadata`/`DocumentManager` class hierarchy.
- **6× `HelperInterface::getDocumentManager()` not found** in the console commands. Needs a proper custom helper interface — an API addition, not a typing fix.
- **6× `method.notFound` in `EncryptedFieldsMapGeneratorTest`'s anonymous test double + 3× `argument.type` in `ObjectCastPropertyAccessor`/`RawValuePropertyAccessor`/`StaticProxyFactory`.** The latter three are ProxyManager stub signature mismatches (closure params declared optional in the stub, required in our actual closures) — workable via closure-signature adjustments, but that touches proxy-initialization internals I'd want a second pair of eyes on.
- **Everything else** (~25 entries): confirmed legitimate and permanent — version-compat guards (`doctrine/collections` v2/v3, `doctrine/persistence` constructor signatures across versions), environment-dependent driver classes (`MongoDB\Model\*` iterator classes, `ext-mongodb` 2.2+ features), test bootstrap constants, dynamic `BSONDocument` property access, and deliberate "test invalid input" cases (spot-checked `BuilderTest::testPrimeRequiresBooleanOrCallable()` and `DocumentManagerTest`'s partial-mapping test — both intentionally pass malformed data to exercise a guard clause, not bugs).

## Trade-offs and things NOT done

- I did not touch the `Iterator<TValue>` interface's `current()` contract to properly model "may yield nothing." That's the correct fix for the `Query.php` entries above but is a shared-interface change with wider blast radius than this cleanup pass intended.
- I did not attempt the `ClassMetadata`/`FieldMapping` array-shape narrowing. It's the single largest remaining cluster (7 entries) but requires auditing every caller to avoid silently narrowing the type in a way that breaks a legitimate partial-mapping use case.
- I did not add a custom Symfony Console helper interface to fix the 6 `getDocumentManager()` entries, since that's a public API surface addition.

## Suggested follow-up (separate tickets)

- [ ] Revisit the `Iterator<TValue>` contract to model "no current element" (`Query.php`'s 3 entries depend on this).
- [ ] Audit `ClassMetadata`'s `FieldMapping` array shape against all callers and either tighten call sites or relax the shape's required keys (7 entries).
- [ ] Decide on `ClassMetadata`/`DocumentManager` generics variance and the `getMetadataFactory()` LSP mismatch (4 entries) — likely a 3.0 concern given `ClassMetadata` is `@final`-annotated but not literally final for BC.
- [ ] Design a typed helper interface for the console commands to replace `HelperInterface::getDocumentManager()` (6 entries).
- [ ] Decide whether to invest in reconciling the ProxyManager closure-signature stub mismatches (3 entries) or leave them baselined long-term.

## Verification

Every commit was verified independently with:
- `vendor/bin/phpstan analyse` with the baseline temporarily disabled (`includes:` line removed via a scratch config), to confirm fixes actually resolve the underlying error rather than just happening to still match the baseline's regex.
- Full `phpstan analyse` with the real baseline, cache cleared, confirming a clean run after each batch.
- Full `vendor/bin/phpunit` run after each batch. The suite consistently shows the same 11 errors / 4 failures / 64 skips across every run in this PR — all pre-existing environment limitations (no Atlas Search or replica set available locally), confirmed unrelated to any change here by checking they reproduce identically before and after.

<details>
<summary>Full session recap (click to expand) — how each decision was made</summary>

### Starting point

Asked to look at the PHPStan baseline and come up with a strategy to fix issues, prioritizing some and leaving others for later, with a specific note: errors about "impossible types" in tests should generally be kept in the baseline unless completely wrong.

I read the baseline in full (211 entries across 1267 lines — bloated mostly by long inline regexes, not actual scale) and grouped it by `identifier`:

| Count | Identifier | Nature |
|---|---|---|
| 46 | `missingType.iterableValue` | missing `array<...>` value types |
| 44 | `missingType.generics` | missing template args |
| 24 | `missingType.parameter` | untyped params |
| 15 | `argument.type` | real type mismatches |
| 13 | `method.notFound` | calls to undeclared interface methods |
| ~20 singles | `booleanAnd.alwaysFalse`, `deadCode.unreachable`, `notIdentical.always*`, `instanceof.alwaysTrue`, `arguments.count`, `match.unhandled` | candidates for real bugs |
| rest | misc | mixed |

I specifically checked the two `staticMethod.impossibleType` entries in `BinaryUuidTypeTest.php` you'd flagged interest in: both are on closures that `eval()` generated code (`$type->closureToMongo()`), which PHPStan can't see through — it assumes the closure's return value is always `null` (its last known static value before the `eval()`), making `assertSame($binaryUuid, null)` look "always false." That's a real PHPStan blind spot around the deprecated eval-based code path, not a bug. Verdict: keep baselined — this became the template for how I treated the rest of that category throughout the whole session.

Proposed 4 tiers ordered by risk: bug-hunt → free wins → mechanical typing sweep → design-judgment items. You approved starting with Tier 1.

### Tier 1: bug hunt

Went through every "always true/false/unreachable" entry. Where I could, I ran PHPStan directly on the isolated file/cluster with the baseline's `includes:` line stripped out (via a scratch copy of the config), to see the *actual* current error state rather than trust the baseline's stale regex matches. This turned out to be essential: I found that several entries no longer reproduced at all, because the code they described had moved to a different file (`PersistentCollection.php` → `PersistentCollectionTrait.php` + `PersistentCollectionCompatibility.php`) during a refactor, and nobody had regenerated the baseline since. That "reproduce fresh, don't trust the baseline's file path" check became my standard verification method for the rest of the session — a plain `[OK] No errors` from the real config with baseline included isn't proof anything's still needed, since a stale baseline entry sitting on the wrong file doesn't get flagged as unmatched (the config disables `reportUnmatchedIgnoredErrors`).

I nearly went down a rabbit hole here suspecting PHPStan's baseline `path:` matching was buggy (an experiment where I deliberately broke a path but the error still didn't reproduce). Root cause turned out to be: analysing a single file (like `PersistentCollectionTrait.php`) alone doesn't give PHPStan enough context to resolve `$this`'s concrete type in the trait, so no error fires at all *regardless* of the baseline — the "bug" was in my test methodology, not PHPStan. Lesson learned: use whole-project analysis when verifying trait-based code.

Findings, one by one:
- `ClassMetadata.php`'s `booleanAnd.alwaysFalse` + `notIdentical.alwaysFalse` + `notIdentical.alwaysTrue` — all three land on the exact same line (`static::class !== self::class && ...`) and share one root cause: the class carries an `@final` docblock annotation (aspirational — it's deliberately not literally `final` yet, for BC with existing subclasses) which PHPStan trusts completely. Not a bug — kept baselined.
- `OnClearEventArgs::__construct()`'s `arguments.count` — a deliberate `method_exists()`-gated call into a differently-shaped parent constructor, for supporting multiple `doctrine/persistence` versions. Same pattern as two other methods in the same file. Not a bug.
- `XmlDriver`/`CreateCommand`/`DropCommand`'s three `match.unhandled` — checked the XSD schema for `<encrypt>` and confirmed the match arms are exhaustive per the schema (for XmlDriver) or per the class's own constant list (for the two commands, which already carry `@phpstan-ignore` comments acknowledging this exact class of false positive elsewhere in the same files). Not a bug.
- `PipelineUpdateTest`'s `deadCode.unreachable` — code after `markTestSkipped()`, a completely normal PHPUnit idiom for keeping a test's intended-future body around while it's skipped (references a tracked upstream issue). Not a bug.
- `PersistentCollectionTrait`'s `instanceof.alwaysTrue` (Selectable) — genuine `doctrine/collections` v2/v3 compatibility guard; `composer.json` allows both. Not a bug, correctly scoped already.
- `DateImmutableType`'s `deadCode.unreachable` — PHPStan can prove it (only two concrete `DateTimeInterface` implementations exist), but I asked you directly whether to simplify or keep as defensive code, since removing it changes intent, not just satisfies the type checker. You chose to keep it.
- `PersistentCollection.php`'s `is_object()`/`method_exists()` (findFirst/reduce) entries — traced these to the exact same "moved during refactor" pattern: the `is_object()` check *had* moved to `PersistentCollectionTrait::__clone()`, but there it's genuinely dead (the property is non-nullable), so I fixed the code and dropped the baseline entry. The two `method_exists()` entries (findFirst/reduce) were fully stale — that code doesn't exist anywhere anymore under any name, the compatibility mechanism was rewritten entirely into `PersistentCollectionCompatibility.php` using a different technique.
- `DocumentPersister`'s `is_array()` — an `assert()` guarding an `array_combine()` result; on PHP 8+ that function always returns an array (never `false`), so the assert can never fail. Removed.
- `Query.php`'s `notIdentical.alwaysTrue` — confirmed via full-project analysis that it simply doesn't reproduce anymore. Stale, removed.

Verified everything with targeted PHPUnit runs (`PersistentCollectionTest`, `PersistentCollectionCloneTest`, `ReferencePrimerTest`) before committing.

### Tier 2: mechanical cleanup

Went through the "free win" categories (unused property types, write-only properties, stale `@throws`, etc.) file by file. Most were exactly what they looked like — but two stood out:

- `Aggregation\Builder::getPipeline()`'s `parameter.notFound` (`$applyFilters`) — the parameter is *deliberately* commented out of the actual signature (`/* bool $applyFilters = true */`) as a soft-deprecation technique (still callable via `func_get_args()`), with a `phpcs:disable` comment right above acknowledging exactly this pattern. The `@param` tag documents it for callers even though PHPStan can't see it as a real parameter. Recognized this as the same category as Tier 1's `@final`-trust and BC-shim findings — legitimate, not a "free win," left alone.
- `Search\Range::$includeUpperBound` "never read, only written" — investigating *why* it was never read (rather than just accepting "unused, delete it") turned up a real bug: `lt()`/`lte()` were literally copy-pasted from `gt()`/`gte()` and never had their `$includeLowerBound` references changed to `$includeUpperBound`. Since this was a behavior change rather than dead-code removal, I stopped and asked before touching it, laying out exactly what breaks (`->gt(5)->lte(10)` silently returns `gte`/`lte` — wrong operator — because both bounds shared one flag). You approved fixing it as its own commit. I added a "mixed bound inclusivity" test case and confirmed it fails against the pre-fix code and passes after, since the existing test suite's bound-inclusivity cases happened to only ever combine same-inclusivity bounds (both inclusive or both exclusive), which is exactly why the aliasing bug had gone unnoticed.
- `GeoWithin::$relation` — by contrast, this one *was* just dead: no setter, never read, and cross-checking against `GeoShape.php` (which has a real, wired-up `relation` concept) confirmed this was a copy-paste leftover from a different search operator, not a missing feature. Atlas Search's `geoWithin` operator has no `relation` parameter in its actual API. Deleted outright.

Two commits: the bug fix (with test) separately from the mechanical cleanup, per your instruction to split when necessary.

### Tier 3: the typing sweep

This was the largest, most mechanical tier (~114 entries). I worked it in three batches, verifying each with the "disable baseline, rerun" method plus targeted PHPUnit runs before moving to the next batch and before removing baseline entries.

Two things went wrong here worth recording:

1. **Over-widening `Query.php`'s `Iterator<TValue>`.** I initially gave `Query::getIterator()` (and `Aggregation::getIterator()`, which shares the same underlying pattern) `Iterator<array<string, mixed>|object>`, matching the file's own existing `execute()` docblock. This was fine for `Aggregation.php` (no test cared), but for `Query.php` it broke ~18 call sites across `ReferencePrimerTest`/`QueryTest` that call methods like `->getAccount()` directly on iterated results — because now PHPStan correctly saw the array half of the union and flagged "Cannot call method on array|object." Narrowing to bare `object` (matching the dominant real-world usage, since `hydrate` defaults to `true`) fixed those, but then surfaced a *second*, subtler issue: PHPStan started "proving" `getSingleResult()` could never return `null`, based on `Iterator<object>::current()` never being able to return a falsy value — except `current()` *can* return nothing when the result set is empty, and the `Iterator<TValue>` interface doesn't model that. Rather than force a type that's technically satisfies-the-linter-but-wrong, I reverted all of `Query.php`'s changes for this tier and left its 3 entries baselined, with a clear note on why (this needs a real interface-contract fix, not a local annotation).

2. **Doc-comment-after-attributes placement.** Repeatedly (in `SetWindowFieldsTest`, then again more visibly in `SearchTest`'s 6 methods), I placed a `/** @param ... */` doc comment directly above a method but *after* a stack of `#[DataProvider(...)]` attributes, and PHPStan silently ignored it — no error, but also no effect, so my "fix" did nothing and I initially didn't notice because I was reading source rather than rerunning PHPStan immediately. Once I caught it (by insisting on rerunning phpstan-without-baseline after *every* edit rather than trusting the diff), the fix was to move the doc comment *before* the attribute stack instead. Same issue showed up in a different shape for `tests/Documents/Encryption/Client.php`'s promoted constructor property: an inline `@var` between an attribute and the promoted property was silently ignored, and needed to become a `@param` on the constructor's own docblock instead.

Batch 1 covered `Iterator<TValue>` and `PersistentCollectionInterface<TKey, T>` generics across the persister layer — mostly parameterizing with the widest safe bound (`array-key, object`) since the concrete document class isn't known at these generic call sites. Also fixed `AttributeDriver`/`AttributeReader::getClassAttributes()`'s `ReflectionClass` generic, which needed its own fresh `@template T` (not a hardcoded `object`) because one call site passes in a `ReflectionClass<T>` coming from `ClassMetadata`'s own covariant template — hardcoding `object` there produced a variance error, which I sidestepped by reconstructing a fresh `ReflectionClass` at that one call site instead of touching `ClassMetadata`'s (separately tracked, Tier 4) covariance issue.

Batch 2 covered the Aggregation search operators. The most efficient single fix here was `CompoundedSearchOperatorTrait::__construct()`'s untyped `...$args` — one `@param mixed ...$args` annotation on the shared trait resolved the identical finding on all 11 `Compounded*` subclasses at once, since PHPStan was reporting the same trait-inlined error separately for every consumer. Also had to walk back an over-narrow `SearchIndex::$fields` type once `AbstractMappingDriverTestCase`'s fixture data showed real several-level-deep nested structures that a two-level-deep annotation rejected — went with permissive `array<string, mixed>` instead.

Batch 3 covered test data-provider methods, mostly straightforward once I'd learned the attribute-ordering lesson from batch 2.

Found and fixed one leftover in a follow-up commit: `DocumentPersister::isInTransaction()`'s baseline entry, which batch 1 had already fixed in code but my removal script's identifier list didn't happen to catch.

### Tier 4: final triage

Went through all 75 remaining entries looking specifically for anything fixable *without* a design decision, per your ask to continue with "items that don't need any input." For each candidate cluster I did enough investigation to be confident about which bucket it belonged in, rather than guessing from the message text alone:

- The 11-entry `argument.templateType` cluster and the `ReferencePrimer` by-ref mismatch: traced both to the same root cause (`class-string` without a template parameter on `targetDocument`), confirming there's no local fix — would need to change how mapping arrays carry type information.
- The `ClassMetadata`/`XmlDriver` `FieldMapping` array-shape entries: read enough of `ClassMetadata::mapField()`/`checkDuplicateMapping()` to confirm the required-keys shape is stricter than what real call sites populate, meaning any fix needs a caller-by-caller audit rather than a type-widening.
- Vendor stub mismatches (`ObjectCastPropertyAccessor`, `RawValuePropertyAccessor`, `StaticProxyFactory`): confirmed via the exact error text that these are all instances of one underlying issue — ProxyManager's `GhostObjectInterface::setProxyInitializer()` stub declares its closure parameter as optional where our real implementations declare it required — workable, but touches proxy-initialization code I felt warranted a second reviewer rather than a solo judgment call.
- Spot-checked two of the remaining test-only `argument.type` entries (`BuilderTest::testPrimeRequiresBooleanOrCallable()`, `DocumentManagerTest`'s partial-mapping test) to confirm the pattern holds across the rest: these are all deliberate "pass bad/partial input, assert it's handled" tests, not bugs.
- Found one more stale cluster this way: `PersistentCollection.php` still had 5 baseline entries (`property.phpDocType`/`class.notFound`/`argument.templateType`) referencing a `$hints`/`$mapping` property and `Hints`/`FieldMapping` type names that don't exist anywhere in that file anymore — same "moved during refactor, baseline never regenerated" pattern as Tier 1's finding on the exact same file. Removed.

Confirmed the final state with a full phpstan run (clean) and a full phpunit run (same 11 errors/4 failures/64 skips as every prior run in this PR, all pre-existing environment gaps).

</details>


src/Aggregation/Stage/Search/GeoShape.php
```diff
@@ -25,9 +25,10 @@ class GeoShape extends AbstractSearchOperator implements ScoredSearchOperator
     private array $path      = [];
     private string $relation = '';
 
+    /** @var LineString|Point|Polygon|MultiPolygon|array<mixed>|null */
     private LineString|Point|Polygon|MultiPolygon|array|null $geometry = null;
 
-    /** @param LineString|Point|Polygon|MultiPolygon|array|null $geometry */
+    /** @param LineString|Point|Polygon|MultiPolygon|array<mixed>|null $geometry */
```
GromNaN: This can probably be more specific about the array-shape that is expected for all the geomeries?

src/Mapping/Driver/AttributeDriver.php
```diff
@@ -84,7 +84,7 @@ public function loadMetadataForClass($className, PersistenceClassMetadata $metad
         assert($metadata instanceof ClassMetadata);
         $reflClass = $metadata->getReflectionClass();
 
-        $classAttributes = $this->getClassAttributes($reflClass);
+        $classAttributes = $this->getClassAttributes(new ReflectionClass($reflClass->getName()));
```
GromNaN: `$reflClass` is already an instance of `ReflectionClass`. Could you improve getReflectionClass instead?

src/PersistentCollection/PersistentCollectionTrait.php
```diff
@@ -359,9 +358,7 @@ public function __sleep()
      */
     public function __clone()
     {
-        if (is_object($this->coll)) {
-            $this->coll = clone $this->coll;
-        }
+        $this->coll = clone $this->coll;
```
GromNaN, line 361: Can you check that `$coll` is not nullable?

tests/Tests/Functional/FunctionalTest.php
```diff
@@ -910,7 +910,7 @@ public function testReplaceEntireGroupsArray(): void
 
         // Issue is collection must be initialized
         $groups = $user->getGroups();
-        $groups[0]; // initialize collection
+        $groups->offsetGet(0); // initialize collection
```
GromNaN, line 913: What error is this fixing?

src/Aggregation/Stage/Search/GeoShape.php
alcaeus: Good call — tightened this to a shared `GeoJsonGeometry` type (`array{type: string, coordinates?: mixed, crs?: mixed, bbox?: mixed}`) instead of bare `array<mixed>`. That matches what `GeoJson\Geometry::jsonSerialize()`/`GeoJson::jsonSerialize()` actually produce — `type` is always present, `coordinates` comes from the `Geometry` subclass, `crs`/`bbox` are optional from the base class. This is the shape someone would pass in as an alternative to constructing a `Point`/`Polygon`/etc. object. Applied the same type to `GeoWithin.php` and `Near.php` too, since they have the identical pattern.

src/Mapping/Driver/AttributeDriver.php
alcaeus: Good point — reverted this rather than keep the workaround. Turns out the `<T>` generic on `getClassAttributes()` can't be added cleanly without hitting `ClassMetadata`'s covariance issue (the same one behind the 3 already-baselined `generics.variance` entries): after `assert($metadata instanceof ClassMetadata)`, PHPStan narrows to a covariant, unbound T, which can't satisfy an invariant `ReflectionClass<T>` parameter regardless of what `getReflectionClass()` declares. Rather than paper over that with a reconstructed `ReflectionClass`, I've put this back to the original code and baseline entry — it can go in properly once there's a decision on `ClassMetadata`'s covariance.

src/PersistentCollection/PersistentCollectionTrait.php
alcaeus: Confirmed non-nullable: `private BaseCollection $coll;` (line 78), no default value. It's actually guaranteed twice over — the type declaration rules out `null`, and since it's an uninitialized typed property with no default, PHP itself throws an `Error` if it's ever read before `__construct()` sets it, so there's no silent-null path either. That's exactly why PHPStan flagged the old `is_object($this->coll)` check as always-true.

tests/Tests/Functional/FunctionalTest.php
alcaeus: `expr.resultUnused`: `Expression "$groups[0]" on a separate line does not do anything.` PHPStan can't see that `ArrayAccess::offsetGet()` on a lazy `PersistentCollection` has the side effect of triggering initialization, so a bare `$groups[0];` statement with its result discarded looks like dead code to it. `$groups->offsetGet(0);` is behaviorally identical (array access on an object desugars to exactly that call) — PHPStan just doesn't flag a discarded method-call return the way it flags a bare expression.

GromNaN (approved):

GromNaN:
Thanks @alcaeus

PR #3051, opened 2026-09-15
Move UnitOfWork's identity map and per-document state into DocumentManager

alcaeus:
This starts moving `UnitOfWork`'s long-lived, session-scoped state — the identity map, per-document persistence state, original data snapshots, and parent associations — into `DocumentManager`, via a new internal `DocumentRegistry`. `UnitOfWork` keeps its flush-scoped responsibilities and now delegates to `DocumentManager` for everything session-scoped instead of holding that state itself.

This is a first step towards a clearer separation of concerns between `DocumentManager` (the long-lived session) and `UnitOfWork` (transaction/flush handling), not a complete restructuring — `UnitOfWork` still owns scheduling, change-set computation, and flush execution. Fully backwards compatible: all previous `UnitOfWork` methods for this state still work, now deprecated in favor of their `DocumentManager` equivalents.

<details>
<summary>More details</summary>

### New API

- `Doctrine\ODM\MongoDB\Registry\DocumentRegistry` (internal): tracks documents by instance (persistence state, identifier, original data, parent association) and by class + identifier (the identity map), and keeps both in sync via `track()`/`stopTracking()`.
- `Doctrine\ODM\MongoDB\Registry\ManagedObjectState`, `PersistenceState` (enum), `ParentAssociation`: internal value objects backing the registry.
- `DocumentManager` gains `@internal` methods that proxy to the registry: `getParentAssociation()`, `setParentAssociation()`, `getOriginalDocumentData()`, `setOriginalDocumentData()`, `setOriginalDocumentProperty()`, `getDocumentIdentifier()`, `getById()`, `tryGetById()`, `containsId()`, `isInIdentityMap()`, `removeFromIdentityMap()`, `getIdentityMap()`, `getManagedDocumentsCount()`, and `getDocumentRegistry()`.

### Deprecated

The corresponding `UnitOfWork` methods that used to hold this state directly are deprecated in favor of the `DocumentManager` equivalents above: `UnitOfWork::STATE_*` constants, `getParentAssociation()`/`setParentAssociation()`, `getOriginalDocumentData()`/`setOriginalDocumentData()`/`setOriginalDocumentProperty()`, `getDocumentIdentifier()`, `getById()`/`tryGetById()`/`containsId()`, `isInIdentityMap()`/`removeFromIdentityMap()`, `getIdentityMap()`, and `size()`.

### Deliberately not done here

- `UnitOfWork::scheduleForInsert()`/`scheduleForUpsert()`/`scheduleForDelete()` and related scheduling logic are untouched — they mix flush-scoped state with the identity map in ways that need more careful handling than a mechanical move.
- No change to `PersistentCollection`'s dependency on `UnitOfWork` for lazy loading and orphan/collection scheduling.
- No attempt yet to extract flush/commit orchestration out of `UnitOfWork` — this PR only relocates ownership of already-mostly-static state.

These are natural candidates for follow-up PRs.

</details>


src/DocumentManager.php
```diff
@@ -733,6 +747,192 @@ public function clear($objectName = null): void
...
+     * Gets the registry tracking long-lived per-document state and the
+     * identity map for this DocumentManager.
+     *
+     * Only reach for this directly when you need lower-level access than
+     * the methods below provide (e.g. from UnitOfWork, which caches the
+     * result of this method); most callers should prefer those instead.
+     *
+     * @internal
+     */
+    public function getDocumentRegistry(): DocumentRegistry
```
GromNaN, line 750: This single getter should be enough.

Instead of adding all this internal public methods to the DocumentManager, you can call the methods of the DocumentRegistry directly?

The DocumentManager should be the public facade. Even if the `@internal` annotation is explicit, it is not a hard restriction for using this methods. It's also more methods if you want to decorate this class.

alcaeus: These methods were public on the UnitOfWork, but accessed state that was session-scoped (i.e. not invalidated after `flush`). Since some of that logic was called from other classes (e.g. persisters), I moved the methods to the DocumentManager since that now owns the session-scoped data (the goal is to have UoW only own transient data that disappears on flushing).
One alternative would be to move some of those methods (e.g. `setParentAssociation`) to the document registry and letting persisters and the UoW work on the document registry. The other alternative would be to leave these method on the UoW, not deprecate them, then deal with that issue later to avoid widening the DocumentManager API.

I agree with not adding to the DocumentManager API as that's the central entry point for users, but I'd also like to encapsulate as much of that logic as possible.

src/Registry/ManagedObjectState.php
```diff
@@ -0,0 +1,38 @@
...
+ * Bundles the long-lived state tracked for a single document while it is
+ * known to the DocumentManager: its persistence state, its identifier, the
+ * snapshot of data used for dirty checking at flush time, and (for embedded
+ * documents) the parent association it was found through.
+ *
+ * @internal This class is not part of the public API and is subject to change.
+ */
+final class ManagedObjectState
+{
+    public ?ParentAssociation $parentAssociation = null;
```
GromNaN: Maybe?
```suggestion
    public readonly ?ParentAssociation $parentAssociation = null;
```

You can also use CPP.

src/Repository/DocumentRepository.php
```diff
@@ -128,7 +128,7 @@ public function find($id, int $lockMode = LockMode::NONE, ?int $lockVersion = nu
         }
 
         // Check identity map first
-        $document = $this->uow->tryGetById($id, $this->class);
+        $document = $this->dm->tryGetById($id, $this->class);
```
GromNaN:
```suggestion
        $document = $this->dm->getDocumentRegistry()->tryGetById($id, $this->class);
```

src/DocumentManager.php
GromNaN: What I understand is that you want the UoW instance to be recreated with every transaction (or session?). This means we can no longer inject it everywhere, as it is currently done.
Whereas the new DocumentRegistry is indeed a single instance, linked to the DocumentManager.

So, given that all the methods you propose adding to DocumentManager are actually enriched proxies to DocumentRegistry, I think it's better to add more methods to DocumentRegistry and use that instance directly.

src/Registry/ManagedObjectState.php
alcaeus: Used CPP, but didn't make it readonly. I wanted to encapsulate this kind of data, but since `setParentAssociation` is called from multiple places I didn't want to risk breakage. This will be refactored further as I start changing the UnitOfWork workflow.

src/DocumentManager.php
alcaeus: Removed all those new methods that were just proxies and replaced them with direct access to the DocumentRegistry.

src/Registry/DocumentRegistry.php
```diff
@@ -0,0 +1,321 @@
...
+/**
+ * Tracks every document known to a DocumentManager for the lifetime of that
+ * manager (until it is cleared), from two directions:
+ *
+ *  - by document instance, via {@see ManagedObjectState} (persistence state,
+ *    identifier, original data, parent association);
+ *  - by class and (serialized) identifier, to find the managed instance for
+ *    a given database identity without loading it twice.
+ *
+ * @internal This class is not part of the public API and is subject to change.
```
GromNaN: Usually not specified.

```suggestion
 * @internal
```

src/Registry/DocumentRegistry.php
```diff
@@ -0,0 +1,321 @@
...
+    public function clear(): void
+    {
+        $this->objectStates = new SplObjectStorage();
+        $this->identityMap  = [];
+    }
+
+    /**
+     * Sets the parent association for a given embedded document.
+     *
+     * @internal
```
GromNaN: The class is already internal, no need to repeat on some methods.

tests/Tests/Registry/PersistenceStateTest.php
```diff
@@ -0,0 +1,22 @@
...
+    {
+        self::assertCount(4, PersistenceState::cases());
+        self::assertNotSame(PersistenceState::New, PersistenceState::Managed);
+        self::assertNotSame(PersistenceState::New, PersistenceState::Detached);
+        self::assertNotSame(PersistenceState::New, PersistenceState::Removed);
+        self::assertNotSame(PersistenceState::Managed, PersistenceState::Detached);
+        self::assertNotSame(PersistenceState::Managed, PersistenceState::Removed);
+        self::assertNotSame(PersistenceState::Detached, PersistenceState::Removed);
+    }
+}
```
GromNaN: This test is not covering anything. It can be safely removed.

tests/Tests/DocumentManagerTest.php
```diff
@@ -76,6 +76,11 @@ public function testGetUnitOfWork(): void
         self::assertInstanceOf(UnitOfWork::class, $this->dm->getUnitOfWork());
     }
 
+    public function testGetDocumentRegistryReturnsTheSameInstanceEveryTime(): void
+    {
+        self::assertSame($this->dm->getDocumentRegistry(), $this->dm->getDocumentRegistry());
```
GromNaN: That's a bit of tautology. I don't think this is a requirement.

src/DocumentManager.php
```diff
@@ -140,19 +141,29 @@ class DocumentManager implements ObjectManager
     /** @var ProxyClassNameResolver&ClassNameResolver  */
     private ProxyClassNameResolver $classNameResolver;
 
+    /**
+     * Tracks every document known to this DocumentManager, by instance and
+     * by class + identifier, for the manager's lifetime (until it is
+     * cleared).
+     *
+     * @internal
```
GromNaN: No need for `@internal` annotation on a private property. The comment can be removed, it duplicates the getter description.

GromNaN (commented):
It's good to me after trimming some comments.

src/Registry/DocumentRegistry.php
alcaeus: Fair point, removed the comment.

src/Registry/DocumentRegistry.php
alcaeus: Removed.

tests/Tests/Registry/PersistenceStateTest.php
alcaeus: Yeah, fair point.

tests/Tests/DocumentManagerTest.php
alcaeus: I agree. In some cases it may make sense, but in this case we can disregard it. Removed.

src/DocumentManager.php
alcaeus: Removed the comment.

src/Registry/DocumentRegistry.php
```diff
@@ -0,0 +1,319 @@
...
+     */
+    public function getOriginalDocumentData(object $document): array
+    {
+        $objectState = $this->getObjectState($document);
+
+        return $objectState !== null ? $objectState->originalData ?? [] : [];
+    }
+
+    /**
+     * @internal
```
GromNaN: There are unnecessary internal annotations remaining on other methods

src/UnitOfWork.php
```diff
@@ -323,31 +279,49 @@ public function getPersistenceBuilder(): PersistenceBuilder
     /**
      * Sets the parent association for a given embedded document.
      *
-     * @internal
+     * @deprecated Use {@see DocumentRegistry::setParentAssociation()} instead.
```
GromNaN: This method was internal. You can remove it directly.

src/UnitOfWork.php
```diff
@@ -1585,40 +1563,44 @@ public function getDocumentState(object $document, ?int $assume = null): int
...
+            PersistenceState::Removed => self::STATE_REMOVED,
+        };
+    }
+
     /**
      * Removes a document from the identity map. This effectively detaches the
      * document from the persistence management of Doctrine.
      *
-     * @internal
+     * @deprecated Use {@see DocumentRegistry::removeFromIdentityMap()} instead.
```
GromNaN: Also internal, can be removed directly.

src/UnitOfWork.php
```diff
@@ -1585,40 +1563,44 @@ public function getDocumentState(object $document, ?int $assume = null): int
...
-
-        return false;
+        return $this->documentRegistry->removeFromIdentityMap($class, $document);
     }
 
     /**
      * Gets a document in the identity map by its identifier hash.
      *
-     * @internal
+     * @deprecated Use {@see DocumentRegistry::getById()} instead.
```
GromNaN: Also internal.

GromNaN (commented):
There are some remaining `@internal` annotations on methods in internal classes. 
You can also remove the methods that were internal, instead of deprecating them; especially since the depreciation message suggests the the internal method.

src/Registry/DocumentRegistry.php
alcaeus: Removed them all.

src/UnitOfWork.php
alcaeus: Removed all previously internal, now deprecated methods. Cleaned up the tests and also documented this in UPGRADE-2.18.md.

src/Registry/PersistenceState.php
```diff
@@ -0,0 +1,25 @@
...
+declare(strict_types=1);
+
+namespace Doctrine\ODM\MongoDB\Registry;
+
+/**
+ * The persistence state of a document tracked by the DocumentManager.
+ *
+ * @internal
+ */
+enum PersistenceState
```
IonBazan, line 12: Can we move this to https://github.com/doctrine/persistence? I could see a similar change happening for ORM.
If you decide to keep it only here, I believe it should not be in `Registry` namespace.

src/Registry/DocumentRegistry.php
```diff
@@ -0,0 +1,298 @@
...
+ *  - by document instance, via {@see ManagedObjectState} (persistence state,
+ *    identifier, original data, parent association);
+ *  - by class and (serialized) identifier, to find the managed instance for
+ *    a given database identity without loading it twice.
+ *
+ * @internal
+ *
+ * @phpstan-import-type FieldMapping from ClassMetadata
+ */
+final class DocumentRegistry implements Countable
```
IonBazan, line 30: It would be great if we could extract at least part of this class to persistence too.

src/Registry/PersistenceState.php
GromNaN: It's internal at this point. Unless we have a cross-orm compatibility need, I would not push to move things to a common package.

IonBazan: Agreed. Let's propose that for persistence v5 as a replacement for `UnitOfWork::STATE_*` constants.

IonBazan (approved):

alcaeus:
@IonBazan There's a lot more coming, and I purposely avoided extracting to Persistence prematurely. ClassMetadata and a bunch of other things will be changing soon. Once it's done, we can discuss extracting functionality to doctrine/persistence if the ORM can (and wants to) leverage it as well.

PR #3053, opened 2026-09-16
Improve UnitOfWork and ClassMetadata test coverage

alcaeus:
This PR adds targeted test coverage for `UnitOfWork` and `ClassMetadata`, the two areas we're actively working on, based on gaps that survived the merged codecov.io report (multi-session, multi-config) rather than the false positives a single local run would show (e.g. untested proxy factories turned out to just be config-excluded, not actually untested).

- **`UnitOfWork::withTransaction()`**: added tests for the `commitTransaction` retry branches — a transient error restarting the whole transaction, an `UnknownTransactionCommitResult` error retrying just the commit, retry exhaustion, and the `MaxTimeMSExpired` (code 50) exclusion. The existing failpoint-based tests only ever failed the callback's write commands, never the `commitTransaction` command itself, so this whole branch was untested.
- **`UnitOfWork::commit()`**: added a test for the reentrancy guard (`flush()` called from an event listener). Writing this test surfaced a real bug — see below.
- **`UnitOfWork::merge()` / `doMerge()`**: had no dedicated coverage for its edge branches. Added tests for the removed-managed-copy guard, the optimistic-lock version-mismatch check, initializing an uninitialized document before merging, and reattaching a detached, non-cascade-merged reference by identity instead of field-merging it.
- **`ClassMetadata`**: added tests for `setDiscriminatorField()`'s deprecated array-argument BC path and for `checkDuplicateMapping()`'s throw when two plain fields map to the same database field name.

Along the way, testing the reentrancy guard exposed a genuine bug: the guard threw *before* entering the `try`/`finally` that decrements `commitsInProgress`, so a nested `commit()` call leaked the counter permanently — any later, unrelated `flush()` on that `DocumentManager` would then throw `commitInProgress()` too. Fixed in a separate, small commit with its own regression test.

Sharding (`SchemaManager`) is a separate piece of work we're tackling next, since it genuinely needs a sharded-cluster CI leg rather than more unit tests.

<details>
<summary>Longer context: how we got here, and what's deliberately out of scope</summary>

We'd been analysing a local `coverage.xml` and initially flagged the proxy factories (`LazyGhostProxyFactory`, `StaticProxyFactory`) and most of `SchemaManager` as untested. Once codecov.io reporting was enabled (#3052) and merged across the CI matrix, most of that turned out to be a measurement artifact: `phpunit.xml` only exercises one lazy-object strategy per run, and sharding only runs against a sharded cluster, not the replica set used locally. The merged report (25 sessions, 85.81%) cleared those false positives and left a short, real list — the items covered by this PR.

We also considered adopting PHPUnit `#[CoversClass]` attributes to scope coverage per test. Decided against it: this codebase's tests are almost entirely integration-style against a real `mongod`/`mongos` connection, and a single test routinely exercises a dozen classes at once. Forcing `#[CoversClass]` here would mean either annotating tests with long class lists for no real filtering benefit, or enabling strict-coverage mode and breaking most of the suite. Worth revisiting narrowly for a genuinely isolated unit-test class in the future, but not a repo-wide change.

Local dev note: `docker-compose.override.yaml` didn't pass `--setParameter enableTestCommands=1`, so `configureFailPoint` — the mechanism the transaction-retry tests (and several pre-existing ones) rely on — was silently rejected by the local MongoDB container. Fixed as part of this work; if you're pulling this branch down, recreate your local container (`docker compose down && docker compose up -d`) to pick up the change.

Deliberately left out of scope for this PR:
- [ ] `SchemaManager` sharding coverage — needs the sharded-cluster CI leg, tracked separately.
- [ ] `DocumentPersister::guardMissingShardKey()` / the `executeUpserts()` `BulkWriteException` path — adjacent to sharding, not metadata/UnitOfWork.
- [ ] `ClassMetadata::fieldNotAllowedForGridFS()` — also untested, spotted during this pass, but outside the discriminator/duplicate-mapping cluster this PR targets. Worth a small follow-up.

</details>


.github/workflows/continuous-integration.yml
```diff
@@ -77,13 +77,15 @@ jobs:
             dependencies: "highest"
             symfony-version: ""
             proxy: "proxy-manager"
+            topology: "server"
```
GromNaN: Could you move this key first, to keep all configs in a consistent order.
Damn, I feel like @jmikola with the PedantryTest.

src/UnitOfWork.php
```diff
@@ -447,6 +447,10 @@ public function commit(array $options = []): void
 
         $this->commitsInProgress++;
         if ($this->commitsInProgress > 1) {
+            // The guarded try/finally below never runs for this call, so undo
+            // the increment here to avoid leaking commitsInProgress.
+            $this->commitsInProgress--;
```
GromNaN: Instead, move `$this->commitsInProgress++;` after the `if` and adjust the comparison.

.github/workflows/continuous-integration.yml
alcaeus: I was thinking about placing it first except there are some places where it isn't already the first key, and it would cause a two-line diff 😅 

No worries though, I'll change it for consistency.

src/UnitOfWork.php
alcaeus: Good point, done.

.github/workflows/continuous-integration.yml
alcaeus: Moved -- I did end up moving the `topology` to the top of the matrix as well to have the same variable order everywhere. I'll probably end up hating myself for that decision if this changes the name of the checks we're requiring on pull requests.

src/UnitOfWork.php
```diff
@@ -445,11 +445,11 @@ public function commit(array $options = []): void
             return; // Nothing to do.
         }
 
-        $this->commitsInProgress++;
         if ($this->commitsInProgress > 1) {
```
GromNaN:
```suggestion
        if ($this->commitsInProgress > 0) {
```

alcaeus: 🤦 Yeah, that's what I get for making changes in the wrong branch, then doing them again.
That said, I would've expected a test to fail as a result. I'll make sure that actually happens.

GromNaN: worktree ftw

GromNaN (approved):

PR #3054, opened 2026-09-17
Create missing search indexes in updateDocumentSearchIndexes()

GromNaN:
This recreates the closed pull request #3043, which was closed by an accidental force push on its branch. The changes are identical, rebased onto the latest `2.18.x`.

### Problem

`SchemaManager::updateDocumentSearchIndexes()` drops the search indexes that are no longer mapped, then calls `Collection::updateSearchIndex()` for **every** mapped index. That command fails when the index does not exist, so `odm:schema:update` errors out on any database that does not already carry the mapped search indexes:

```
No index with name globalSearch exists in namespace mydb.MyCollection
```

Two ways to hit it:

* a fresh database, the mapped search indexes have never been created;
* a database where `odm:schema:create --search-index` was never run, or where a search index was dropped by hand.

Because `Schema\UpdateCommand` wraps its phases in a single try/catch, the failure also skips the search-index phase for every class after the failing one.

Reproduced on 2.17.1 and on `2.18.x` against `mongodb/mongodb-atlas-local`: with the mapped search index absent, `odm:schema:update` exits 255; running `odm:schema:create --search-index` first makes the subsequent update succeed. So today `schema:update` cannot converge on its own, it only works if a create ran before it.

### Change

Mapped search indexes that the collection does not have are collected and created with `createSearchIndexes()` instead of being passed to `updateSearchIndex()`; existing ones keep being updated in place. The create path mirrors `createDocumentSearchIndexes()`, including its check that every requested name was reported back as created, so a name the server silently ignores still raises `SchemaException::missingSearchIndex()`.

### Tests

* `testUpdateDocumentSearchIndexesCreatesMissingSearchIndexes`, a mapped index that does not exist is created, `updateSearchIndex()` is not called for it.
* `testUpdateDocumentSearchIndexesMissingIndexNotCreatedError`, a missing index the server does not report back raises `SchemaException`.
* `testUpdateDocumentSearchIndexes` gains `expects($this->never())->method('createSearchIndexes')`, pinning that an existing index is still updated rather than re-created.

Both new tests fail on `2.18.x` without the change. PHPStan and PHPCS are clean.


GromNaN:
LGTM. Thank you @bbarhoum1

alcaeus (approved):

bbarhoum1:
> LGTM. Thank you @bbarhoum1

No problem. Thank you @GromNaN

Issue #2834, opened 2025-09-25
DocumentChangeSet not containing original document values if change set is recomputed


Nilz11:
### Bug Report

Seems to be since 10 years unchanged behaviour. Affected versions 2.x and 3.x

#### Summary

When I recompute the change set, I would expect that it contains the original document data. 

So for example I change a field recomputeChangeSet and fetching this change set afterwards contains the updated value as the old value and not the original database value.

I tracked this down to UnitOfWork which does set $this->originalDocumentData[$oid] = $actualData; ~ line 850 depending on the version.

#### Current behavior

It contains the updated value as the old value and not the original value.

#### How to reproduce

Short example i made:
```

class ModifyListener
{
    public function onFlush(OnFlushEventArgs $eventArgs)
    {
        $uow = $eventArgs->getDocumentManager()->getUnitOfWork();

        foreach ($uow->getScheduledDocumentUpdates() as $document) {
            $document->setFirstName('Change in listener');

            $uow->recomputeSingleDocumentChangeSet(
                $eventArgs->getDocumentManager()->getClassMetadata(Customer::class),
                $document
            );
        }
    }
}

class TrackingListener
{
    public function onFlush(OnFlushEventArgs $eventArgs)
    {
        $uow = $eventArgs->getDocumentManager()->getUnitOfWork();

        foreach ($uow->getScheduledDocumentUpdates() as $document) {

            $cs = $uow->getDocumentChangeSet($document);

            var_dump($cs);
        }
    }
}

$customer = (new Customer())
->setFirstName('Insert');

$dm->persist($customer);
$dm->flush();

$dm->getEventManager()->addEventListener('onFlush', new ModifyListener());
$dm->getEventManager()->addEventListener('onFlush', new TrackingListener());

$customer->setFirstName('Update');

$dm->flush();

```

This outputs using var_dump:

```

array(1) {
  ["firstName"]=>
  array(2) {
    [0]=>
    string(6) "Update"
    [1]=>
    string(18) "Change in listener"
  }
}

```

#### Expected behavior

I would expect that the change set is:

```
array(1) {
  ["firstName"]=>
  array(2) {
    [0]=>
    string(6) "Insert"
    [1]=>
    string(18) "Change in listener"
  }
}
```

PR #3055, opened 2026-09-21
Extract stateless ChangeSetComputer for changeset field diffing

alcaeus:
This is the second layer of the ongoing DocumentManager/UnitOfWork/Persister structural refactor (see #3051 for the first layer, `dm-object-state`). It extracts field-level changeset computation out of `UnitOfWork::computeOrRecomputeChangeSet()` into a new stateless `ChangeSets\ChangeSetComputer`, backed by a new `ChangeSets\ChangeSet` value object that replaces the ad-hoc `array{0: mixed, 1: mixed}` tuple shape internally.

- Added `ChangeSets\ChangeSet`, a value object describing a document's pending field changes (public `$document`, replacing the removed `setNewValue()` mutator)
- Added `ChangeSets\ChangeSetComputationRequest` / `ChangeSetComputationResult` DTOs for batched changeset computation
- Added `ChangeSets\ChangeSetComputer`, wired into `UnitOfWork`, keyed by document instance (`computeOrRecomputeChangeSet()` removed as unneeded)
- `UnitOfWork::getChangeSet(object $document): ChangeSet` is a new, purely additive public method; the existing `getDocumentChangeSet(): array` is unchanged
- Documented in `UPGRADE-2.18.md`

Deliberately unchanged / out of scope:
- Association-change walking (`computeAssociationChanges()`) and scheduling remain in `UnitOfWork` for this layer; both are candidates for further extraction once a scheduling abstraction exists
- [ ] doctrine/mongodb-odm#2834 (wrong "old value" reported when recomputing a changeset after a `preUpdate` listener mutation) is a known, pre-existing bug this refactor does not touch — left for its own follow-up layer with an opt-in `Configuration` flag, so the structural change and the behavior fix can be reviewed and released independently

No public API signatures changed; this PR is purely structural.


src/ChangeSets/ChangeSetComputationRequest.php
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * One document's input to {@see ChangeSetComputer::computeChangeSets()}.
+ *
+ * @internal
+ */
+final class ChangeSetComputationRequest
+{
+    /**
+     * @param ClassMetadata<object>     $class
```
GromNaN: Can you improve type definition?
```suggestion
     * @template T of object = object
     * @param ClassMetadata<T>     $class
     * @param T $object
```

src/UnitOfWork.php
```diff
@@ -410,11 +413,12 @@ function (Session $session) use ($options): void {
                 }
             }
 
+            $this->documentChangeSets = new SplObjectStorage();
```
GromNaN, line 416: I remember doing an benchmark on this. `SplObjectStorage` was significantly slower than an array indexed by `spl_object_id`.

src/ChangeSets/ChangeSetComputer.php
```diff
@@ -0,0 +1,368 @@
...
+        $isNewDocument = $request->originalData === null;
+
+        if ($isNewDocument || $changeSet->isEmpty()) {
+            return new ChangeSetComputationResult($changeSet, $isNewDocument);
+        }
+
+        [$orphansToRemove, $collectionsToDelete] = $this->determineScheduling($changeSet, $request->class);
+
+        if ($request->existingChangeSet !== null && $request->existingChangeSet !== $changeSet) {
+            $request->existingChangeSet->merge($changeSet);
```
IonBazan, line 68: merge() keeps the first pass's originalData, so a recompute after a preUpdate change now reports a different old value.

src/ChangeSets/ChangeSetComputer.php
```diff
@@ -0,0 +1,368 @@
...
+    public function computeChangeSet(ChangeSetComputationRequest $request, Closure $isCollectionScheduledForDeletion): ChangeSetComputationResult
+    {
+        $changeSet     = $this->computeChangeSets([$request], $isCollectionScheduledForDeletion)[$request->document];
+        $isNewDocument = $request->originalData === null;
+
+        if ($isNewDocument || $changeSet->isEmpty()) {
+            return new ChangeSetComputationResult($changeSet, $isNewDocument);
+        }
+
+        [$orphansToRemove, $collectionsToDelete] = $this->determineScheduling($changeSet, $request->class);
```
IonBazan: Under NOTIFY, `determineScheduling()` runs on the reused changeset, so collections recorded by `propertyChanged()` now get scheduled for deletion.

src/UnitOfWork.php
```diff
@@ -611,180 +632,94 @@ public function computeChangeSet(ClassMetadata $class, object $document): void
...
-
-                // ignore inverse side of reference relationship
-                if (isset($class->fieldMappings[$propName]['reference']) && $class->fieldMappings[$propName]['isInverseSide']) {
-                    continue;
-                }
-
-                $changeSet[$propName] = [null, $actualValue];
-            }
+        $actualData  = $this->getDocumentActualData($document);
+        $actualData  = $this->fixActualDataCollectionOwnership($class, $document, $actualData);
```
IonBazan: Collection ownership is now fixed for every field and before the read-only check. This can clone collections and write them back onto read-only documents and onto `notSaved` or inverse fields. Move it after the read-only return and limit it to the fields the old code touched.

src/UnitOfWork.php
```diff
@@ -2814,7 +2834,11 @@ public function propertyChanged($sender, $propertyName, $oldValue, $newValue): v
         }
 
         // Update changeset and mark document for synchronization
-        $this->documentChangeSets[$oid][$propertyName] = [$oldValue, $newValue];
+        if (! isset($this->documentChangeSets[$sender])) {
+            $this->documentChangeSets[$sender] = new ChangeSet($sender, $this->documentRegistry->getOrCreateObjectState($sender)->originalData ?? []);
+        }
+
+        $this->documentChangeSets[$sender]->recordChange($propertyName, $newValue);
```
IonBazan, line 2872: `$oldValue` from the notifier is now ignored and the old value is read from the snapshot (`null` when there is none). Listeners using NOTIFY tracking can see a different old value. Is this intended?

tests/Tests/ChangeSets/ChangeSetComputerTest.php
```diff
@@ -0,0 +1,573 @@
...
+        ));
+
+        self::assertTrue($changeSet->hasChangedField('groups'));
+    }
+
+    public function testDirtyPersistentCollectionWithUnchangedReferenceIsRecorded(): void
+    {
+        $class      = $this->getClassMetadata(User::class);
+        $document   = new User();
+        $collection = $this->createMock(PersistentCollectionInterface::class);
```
IonBazan, line 270: We only rely on the collection mocks which doesn't truly cover UoW changes. Let's add coverage to affected areas:
- recompute after a preUpdate mutation, asserting the old value either way so the #2834 change is intentional and visible
- NOTIFY plus a swapped collection
- a read-only document holding a cloned collection
- the clearDocumentChangeSet() behaviour of bubbling up embedded changes

tests/Tests/ChangeSets/ChangeSetComputerTest.php
```diff
@@ -0,0 +1,573 @@
...
+
+        $result = $this->computer->computeChangeSet(
+            new ChangeSetComputationRequest($class, $document, ['groups' => $collection], ['groups' => $collection], null, false, false),
+            static fn () => false,
+        );
+
+        self::assertSame([], $result->collectionsToDelete);
+    }
+
+    public function testComputeChangeSetMergesOntoExistingChangeSetUsingOnlyThisPassDiffForScheduling(): void
```
IonBazan, line 513: This test doesn't verify what its name says. `address` goes from `null` to a new value, so no orphan could be produced either way. Use a non-null old address in the existing changeset and assert it isn't orphaned twice.

tests/Tests/ChangeSets/ChangeSetComputerTest.php
```diff
@@ -0,0 +1,573 @@
...
+     * @return ClassMetadata<T>
+     *
+     * @template T of object
+     */
+    private function getClassMetadata(string $className): ClassMetadata
+    {
+        return self::getMetadataOnlyDocumentManager()->getClassMetadata($className);
+    }
+
+    private static function getMetadataOnlyDocumentManager(): DocumentManager
```
IonBazan, line 55: Very similar to what we already have in `BaseTestCase`. Let's merge them.

src/UnitOfWork.php
```diff
@@ -511,25 +515,42 @@ private function computeScheduleUpsertsChangeSets(): void
...
 
-        return $this->documentChangeSets[$oid] ?? [];
+    /**
+     * Gets the changeset for a document as a {@see ChangeSet} value object.
+     *
+     * Returns the same instance stored internally, so mutating it (e.g. via
+     * {@see ChangeSet::recordChange()}) is visible to the rest of the UnitOfWork
+     * without calling {@see self::setDocumentChangeSet()} back.
+     */
+    public function getChangeSet(object $document): ChangeSet
```
IonBazan, line 533: It returns a mutable object with inconsistent sharing:
- When a changeset exists, it returns the live internal instance, and the docblock invites mutating it through the `@internal recordChange()`.
- When none exists, it returns a detached new instance, and changes to it are silently lost.
- Either return a read-only view, or don't add the public method until we use it.

IonBazan (changes requested):

src/UnitOfWork.php
alcaeus: Here's the benchmark I ran before: https://gist.github.com/alcaeus/0fe9601ac220e4129b915d158f7ff434 — `isset`/`unset` actually favor `SplObjectStorage` (0.151μs vs 0.246μs/0.235μs for the array+`spl_object_id()` equivalent), `get` is roughly a wash (0.176μs vs 0.158μs). So I don't think this particular keying is a regression; happy to revisit if you've got a benchmark showing a different pattern is slower.

src/UnitOfWork.php
alcaeus: This is intentional. `ChangeSet` always derives old values from the original-data snapshot, never from what the caller passes into `propertyChanged()` — that's the value object's core contract. Reintroducing the old array-based semantics (reporting whatever the previous `propertyChanged()` call happened to pass as `$oldValue`) would take deliberate extra work at this point, unlike the already-flagged `originalData`-stamped-too-early issue elsewhere in this method, which is a much harder fix. Given change tracking NOTIFY is deprecated anyway, I'd rather leave this as-is than spend effort reintroducing behavior we don't want to keep long-term.

src/ChangeSets/ChangeSetComputer.php
alcaeus: Thanks for flagging this — I traced it through carefully: pass 1 diffs against the true snapshot and stores that changeset (anchored to the true original) as `documentChangeSets[$document]`. When `recomputeSingleDocumentChangeSet()` runs a second pass after a preUpdate mutation, its own transient changeset is anchored to the (already-stamped) pass-1 snapshot, but `merge()` only copies pass 2's *new* values onto the pass-1 changeset — `getOldValue()` always resolves through the pass-1 changeset's own snapshot, never pass 2's. So the externally visible old value should stay anchored to the true original throughout, and I confirmed this with an integration test reproducing #2834 — it passes on this branch. If you've got a scenario where it still comes out wrong, I'd appreciate a concrete repro so I can dig further; happy to add it as a regression test either way.

alcaeus:
Thank you for your reviews. I've had Claude made changes where possible. Let me know if you spot other issues.

src/UnitOfWork.php
GromNaN: Thanks for sharing the benchmark, Andreas! I re-ran it with Xdebug enabled and more revisions (50k x 10 iterations) to stress it a bit, and the time results line up with yours: `isset`/`unset` favor `SplObjectStorage`, and `get` is a wash.

I also compared memory (each structure measured in a separate process, no Xdebug): `SplObjectStorage` uses about 24 bytes more per entry, roughly +30-40% across 1k/10k/100k entries (e.g. +24 KB at 1000 entries). Not a concern for the UnitOfWork identity map at typical sizes, but worth keeping in mind if the map grows large.

src/ChangeSets/ChangeSetComputationRequest.php
alcaeus: Done.

src/ChangeSets/ChangeSetComputer.php
alcaeus: Done — `determineScheduling()` now skips to-many scheduling under NOTIFY, with a regression test.

src/UnitOfWork.php
alcaeus: Done — moved the fix after the read-only return and limited it to the fields the old diff loop actually touched.

src/UnitOfWork.php
alcaeus: Done — it now returns a detached clone in both cases.

tests/Tests/ChangeSets/ChangeSetComputerTest.php
alcaeus: Done — reworked with a non-null old address so it actually asserts the orphan.

tests/Tests/ChangeSets/ChangeSetComputerTest.php
alcaeus: Done — extracted a shared `BaseTestCase::createMetadataOnlyDocumentManager()` helper.

tests/Tests/ChangeSets/ChangeSetComputerTest.php
alcaeus: Added the NOTIFY+swapped-collection case as part of the scheduling fix; the other three (preUpdate-recompute old value, read-only+cloned collection, clearDocumentChangeSet() bubbling) still need dedicated tests — want those now or as follow-up?

src/UnitOfWork.php
IonBazan: OK, worth mentioning in UPGRADING for anyone still stuck with NOTIFY tracking. Also, ensure it's covered by a test.

src/UnitOfWork.php
```diff
@@ -511,25 +515,45 @@ private function computeScheduleUpsertsChangeSets(): void
...
+     * Always returns an instance detached from the one (if any) stored
+     * internally: mutating the returned instance (e.g. via
+     * {@see ChangeSet::recordChange()}) has no effect on the UnitOfWork's own
+     * bookkeeping.
+     */
+    public function getChangeSet(object $document): ChangeSet
+    {
+        $changeSet = $this->documentChangeSets[$document] ?? null;
+
+        return $changeSet !== null ? clone $changeSet : new ChangeSet($document, []);
```
IonBazan, line 537: Can we measure the performance impact of this? We are extensively cloning this object and copying the arrays on each call.

src/UnitOfWork.php
```diff
@@ -611,180 +635,95 @@ public function computeChangeSet(ClassMetadata $class, object $document): void
...
+            $this->documentChangeSets[$document] ?? null,
+            $class->isChangeTrackingNotify(),
+            $recompute,
+        );
 
-            $gridFSMetadataProperty = null;
+        $result = $this->changeSetComputer->computeChangeSet($request, $this->isCollectionScheduledForDeletion(...));
+
+        if ($result->isNewDocument) {
+            // FIXME: this stamps $actualData as the document's "original" (i.e. persisted)
```
IonBazan, line 680: Let's create an issue

src/UnitOfWork.php
```diff
@@ -2785,13 +2836,13 @@ public function registerManaged(object $document, $id, array $data): void
...
-     * Clears the property changeset of the document with the given OID.
+     * Clears the property changeset of the given document.
      *
      * @internal
      */
-    public function clearDocumentChangeSet(int $oid): void
+    public function clearDocumentChangeSet(object $document): void
     {
-        $this->documentChangeSets[$oid] = [];
+        unset($this->documentChangeSets[$document]);
```
IonBazan, line 2845: Might be useful to clear changesets during `doDetach` as well, as they now contain `ChangeSet` objects.

src/ChangeSets/ChangeSetComputer.php
```diff
@@ -0,0 +1,379 @@
...
+        $isNewDocument = $request->originalData === null;
+
+        if ($isNewDocument || $changeSet->isEmpty()) {
+            return new ChangeSetComputationResult($changeSet, $isNewDocument);
+        }
+
+        [$orphansToRemove, $collectionsToDelete] = $this->determineScheduling($changeSet, $request->class, $request->isChangeTrackingNotify);
+
+        if ($request->existingChangeSet !== null && $request->existingChangeSet !== $changeSet) {
+            $request->existingChangeSet->merge($changeSet);
```
GromNaN: Blocking: the merged changeset is anchored on the listener-provided `originalData`, so any field the recompute adds reports a null old value, which produces a wrong `$inc`.

`UnitOfWork::setDocumentChangeSet()` builds `originalData` from only the fields present in the array it receives. For `PreUpdateEventArgs::setNewValue()` that array is the changeset as it stood at that moment, so the snapshot is incomplete by construction. Merging the fresh changeset onto that existing instance copies the fresh new values but keeps the existing instance's incomplete snapshot, so `ChangeSet::getOldValue()` returns null for every field that was not already in the changeset.

Reproduction, on a managed document with `count` mapped as `strategy: 'increment'`:
1. `$uow->setDocumentChangeSet($doc, ['username' => ['alice', 'bob']])`, which is what `setNewValue()` does
2. `$doc->setCount(5)`
3. `$uow->recomputeSingleDocumentChangeSet($class, $doc)`

On 2.18.x: `getDocumentChangeSet()` gives `count old=1 new=5` and `prepareUpdateData()` gives `$inc: 4`.
On this branch: `count old=null new=5` and `$inc: 5`.

With `count = 1` and `count` set to 5, the document ends up at 6 instead of 5. `PersistenceBuilder::prepareUpdateData()` feeds `$old` to the increment diff (`IntType::diff()` returns `$new - $old`), so this is data corruption, not only a wrong reported value.

It is reachable when a class has a `#[ODM\PreUpdate]` callback that calls `setNewValue()` and a second field that differs by the time `LifecycleEventManager::preUpdate()` runs `recomputeSingleDocumentChangeSet()` after the callbacks.

The same null old value also affects `DocumentPersister::guardMissingShardKey()` (line 1518 compares `$dcs[$field][0] !== $dcs[$field][1]`), which then throws `shardKeyFieldCannotBeChanged` for an unchanged shard key, and it is what later listeners and the embedded cascade read through `PreUpdateEventArgs::getOldValue()`.

This is distinct from #2834: there the old value was wrong, here it becomes null for fields the fresh diff used to supply correctly, because the previous code composed `$fresh + $existing` and the fresh entry won.

Anchoring the merge on the fresh changeset's snapshot, or extending the existing snapshot from `$request->originalData`, would fix it. Nothing covers `setNewValue()` followed by a recompute today (only `tests/Tests/Events/PreUpdateEventArgsTest.php:71` calls `setNewValue`), so a regression test is needed.

src/UnitOfWork.php
```diff
@@ -511,25 +515,45 @@ private function computeScheduleUpsertsChangeSets(): void
...
     {
-        $this->documentChangeSets[spl_object_id($document)] = $changeset;
+        $originalData = [];
+        $newValues    = [];
+        foreach ($changeset as $field => [$oldValue, $newValue]) {
+            $originalData[$field] = $oldValue;
+            $newValues[$field]    = $newValue;
+        }
+
+        $this->documentChangeSets[$document] = new ChangeSet($document, $originalData, $newValues);
```
GromNaN: Companion to the comment on `ChangeSetComputer::computeChangeSet()`. The `originalData` built here only covers the fields present in `$changeset`, and `PreUpdateEventArgs::setNewValue()` passes the changeset as it stood at that moment, so this snapshot is incomplete by construction. Any later merge that anchors on this instance, which is what the computer does, then reports null old values for the fields the merge adds.

Seeding `originalData` from the document's real snapshot (`DocumentRegistry` original data) instead of from the passed array would keep old values correct regardless of which fields the listener happened to touch.

src/ChangeSets/ChangeSet.php
```diff
@@ -0,0 +1,132 @@
...
+    {
+        return array_keys($this->newValues);
+    }
+
+    public function hasChangedField(string $field): bool
+    {
+        return array_key_exists($field, $this->newValues);
+    }
+
+    public function getOldValue(string $field): mixed
```
GromNaN: `?? null` makes "field not in the snapshot" indistinguishable from "the old value was null". That is what turns the incomplete snapshot described in the other comments into a silent null instead of a visible failure. Worth either reading from the full snapshot or handling the missing case explicitly.

src/ChangeSets/ChangeSetComputer.php
```diff
@@ -0,0 +1,379 @@
...
+            }
+
+            // Under NOTIFY change tracking, a to-many field is never diffed against
+            // its snapshot here in the first place (buildManagedChangeSet() records
+            // it only via propertyChanged(), which has no old collection to compare
+            // against) — so when $changeSet is the NOTIFY-reused, cross-pass instance
+            // (see computeChangeSet()), a to-many field it carries may simply be a
+            // propertyChanged()-recorded change unrelated to a collection swap in
+            // this pass, and must not be scanned for a collection to delete.
+            if ($isChangeTrackingNotify) {
```
GromNaN: The guard sits after the single-valued branches (lines 95 to 114), so those still scan the reused changeset, and the reasoning in the comment above applies to them as well. `determineScheduling()` now iterates the changeset's field names instead of diffing `$actualData`, and under NOTIFY a reused changeset carries fields that `propertyChanged()` recorded, possibly in an earlier pass.

For a to-one embed or a to-one reference with `orphanRemoval`, a value changed and then restored to its original within the same flush stays in the changeset, and the old value, which is also the current value, gets scheduled for orphan removal. The previous code did not do that, because its scheduling came from the `$actualData` diff, where equal values short-circuit.

No test document combines NOTIFY with `orphanRemoval`, so this is unverified. Either extend the guard to the to-one branches or add a test.

src/UnitOfWork.php
```diff
@@ -2785,13 +2836,13 @@ public function registerManaged(object $document, $id, array $data): void
     }
 
     /**
-     * Clears the property changeset of the document with the given OID.
+     * Clears the property changeset of the given document.
      *
      * @internal
      */
-    public function clearDocumentChangeSet(int $oid): void
+    public function clearDocumentChangeSet(object $document): void
```
GromNaN: Backward compatibility: this changes a public signature from `int $oid` to `object $document`. It is marked `@internal`, but the PR description states that no public API signatures changed, and `UPGRADE-2.18.md` does not mention it while it does list the other internal `UnitOfWork` methods that were removed. Callers pass `spl_object_id()` today, this repository's own `tests/Tests/Mocks/PreUpdateListenerMock.php` did, so the change deserves an UPGRADE entry.

Note that the behavior also changes from storing `[]` to `unset()`, which flips `isset($this->documentChangeSets[$document])` from true to false for the entry.

GromNaN (commented):
Reviewed with attention to correctness, backward compatibility, performance and security. One blocking item, plus a rebase note.

Verified on this branch: `tests/Tests/ChangeSets` passes (38 tests), `UnitOfWorkTest` passes (41 tests), `phpcs` is clean, and a full suite run shows no new failures versus 2.18.x. That run is not conclusive for the findings below: the MongoDB server expected on port 27017 was not reachable from the driver, so every functional test errored on connection, and the regression in `ChangeSetComputer::computeChangeSet()` is not covered by any test. It is reproduced with a standalone script, see the comment on that method.

Structure: the extraction is clean, `ChangeSetComputer` really is stateless, the request/result DTOs are a good call, and the added `FIXME` comments about the premature snapshot are useful.

Security: nothing to report. No new input surface, no value logging, and `ChangeSet::assertFieldChanged()` only exposes the document class name, which callers could already obtain.

UPGRADE-2.18.md
```diff
@@ -1,5 +1,14 @@
 # UPGRADE FROM 2.17 to 2.18
 
+## New `ChangeSets\ChangeSet` value object
+
+`UnitOfWork::getChangeSet(object $document): ChangeSet` is a new method
+returning a `Doctrine\ODM\MongoDB\ChangeSets\ChangeSet` object describing a
+document's pending field changes, as an alternative to the existing
+`getDocumentChangeSet(): array`, which is unaffected and continues to return
```
GromNaN: Two behavior changes make this statement inaccurate, so either fix them or document them:
1. After `PreUpdateEventArgs::setNewValue()` followed by a recompute, fields that the recompute adds now report a null old value. See the comment on `ChangeSetComputer::computeChangeSet()`.
2. `UnitOfWork::bubbleUpEmbeddedChildChanges()` used to record `[$value, $value]` for a changed embedded child, so old and new were identical, and now records the snapshot value as the old value.

The `array<string, array{0: mixed, 1: mixed}>` shape is unchanged, the values inside it are not.

src/UnitOfWork.php
```diff
@@ -511,25 +515,45 @@ private function computeScheduleUpsertsChangeSets(): void
      * Gets the changeset for a document.
      *
      * @return array{property: array{0: mixed, 1: mixed}}
-     * @phpstan-return array<string, ChangeSet>
+     * @phpstan-return LegacyChangeSetArray
      */
     public function getDocumentChangeSet(object $document): array
     {
-        $oid = spl_object_id($document);
+        return $this->getChangeSet($document)->toArray();
```
GromNaN: Performance: this clones the stored `ChangeSet` on every call, but `toArray()` already returns a fresh array, so the clone is pure overhead here. `getDocumentChangeSet()` is called once per document by `PersistenceBuilder` for inserts, updates and upserts, recursively for every embedded document, several times per document by `LifecycleEventManager`, and once by `guardMissingShardKey()`.

`$this->documentChangeSets[$document]?->toArray() ?? []` returns the same thing without the clone. Cloning only matters for the public `getChangeSet()`, where returning a detached instance is intentional.

src/UnitOfWork.php
```diff
@@ -1163,7 +1162,7 @@ private function executeUpdates(ClassMetadata $class, array $documents, array $o
         foreach ($documents as $oid => $document) {
             $this->lifecycleEventManager->preUpdate($class, $document, $options['session'] ?? null);
 
-            if (! empty($this->documentChangeSets[$oid]) || $this->hasScheduledCollections($document)) {
+            if (! $this->getChangeSet($document)->isEmpty() || $this->hasScheduledCollections($document)) {
```
GromNaN: Performance: this allocates a cloned `ChangeSet` per document and per update, where the previous code did an array lookup. An internal check that does not clone would be enough, for example `isset($this->documentChangeSets[$document]) && ! $this->documentChangeSets[$document]->isEmpty()`.

src/ChangeSets/ChangeSetComputer.php
```diff
@@ -0,0 +1,379 @@
...
+     * the caller needs to schedule as a consequence (orphan removal,
+     * collection deletion), merging onto an existing changeset when one is
+     * given. Scheduling is derived from this pass's own diff, before that
+     * merge happens.
+     *
+     * @param Closure(PersistentCollectionInterface<array-key, object>): bool $isCollectionScheduledForDeletion
+     */
+    public function computeChangeSet(ChangeSetComputationRequest $request, Closure $isCollectionScheduledForDeletion): ChangeSetComputationResult
+    {
+        $changeSet     = $this->computeChangeSets([$request], $isCollectionScheduledForDeletion)[$request->document];
```
GromNaN: This routes the single-document case through the batch entry point, so every call allocates a one-entry `SplObjectStorage`, hashes the document for the lookup, and the caller also allocates `$this->isCollectionScheduledForDeletion(...)` as a closure per document. It runs once per document per flush, and the batch API has no caller in this layer yet, so computing directly here would avoid the per-document overhead.

src/UnitOfWork.php
```diff
@@ -800,30 +739,56 @@ private function computeOrRecomputeChangeSet(ClassMetadata $class, object $docum
...
+                $this->documentChangeSets[$parentDocument] = new ChangeSet($parentDocument, $originalData ?? []);
+            }
 
-                if (! $isNewDocument) {
-                    $this->scheduleForUpdate($document);
-                }
+            if (! $this->documentChangeSets[$parentDocument]->hasChangedField($mapping['fieldName'])) {
+                // instance of $value is the same as it was previously otherwise there would be
+                // change set already in place
+                $this->documentChangeSets[$parentDocument]->recordChange($mapping['fieldName'], $value);
```
GromNaN: Observable change worth documenting: this used to record `[$value, $value]`, so the old and new values were identical for a bubbled embedded change, and now the old value is the snapshot value. That is arguably more correct, but it changes what `getDocumentChangeSet()` and `PreUpdateEventArgs::getOldValue()` report for embedded fields.

src/ChangeSets/ChangeSetComputer.php
```diff
@@ -0,0 +1,379 @@
...
+        // to diff against, so every current field value is recorded as a change
+        // and the whole document becomes an INSERT.
+        if ($request->originalData === null) {
+            return $this->buildInsertChangeSet($class, $document, $actualData);
+        }
+
+        // Read-only documents never produce a changeset: whatever their fields
+        // say now is irrelevant. We just echo back whatever changeset (if any)
+        // the caller already had.
+        if ($class->isReadOnly) {
```
GromNaN: `UnitOfWork::applyChangeSet()` returns early for a read-only class that already has a snapshot (`UnitOfWork.php:660`), so this branch is unreachable from the UnitOfWork. Either drop it or note that it only matters for direct use of the computer.

src/ChangeSets/ChangeSet.php
```diff
@@ -0,0 +1,132 @@
...
+ * @phpstan-type LegacyChangeSet array{mixed, mixed}
+ * @phpstan-type LegacyChangeSetArray array<string, LegacyChangeSet>
+ */
+final class ChangeSet
+{
+    /**
+     * @param array<string, mixed> $originalData
+     * @param array<string, mixed> $newValues
+     */
+    public function __construct(
```
GromNaN: Since `recordChange()` and `merge()` are `@internal`, exposing `$newValues` on the public constructor is inconsistent: it is the only way for outside code to build a changeset that carries changes. An internal named constructor would match the rest of the API.

Unrelated small point on the same class: `applyToOriginalData()` (line 102) has no caller outside the tests.

src/UnitOfWork.php
GromNaN: Good that the premature snapshot is documented rather than left silent, but these are pre-existing issues rather than something this layer introduces, and a `FIXME` inside a hot method tends to linger. A follow-up ticket would be better, and it could also host #2834 since both touch the same lifecycle.

tests/Tests/BaseTestCase.php
```diff
@@ -174,6 +175,36 @@ protected static function createTestDocumentManager(): DocumentManager
...
+    {
+        $config = new Configuration();
+        $config->setProxyDir(sys_get_temp_dir());
+        $config->setProxyNamespace('Proxies');
+        $config->setHydratorDir(sys_get_temp_dir());
+        $config->setHydratorNamespace('Hydrators');
+        $config->setPersistentCollectionDir(sys_get_temp_dir());
+        $config->setPersistentCollectionNamespace('PersistentCollections');
+        $config->setDefaultDB('doctrine_odm_metadata_only_test');
+        $config->setMetadataDriverImpl(static::createMetadataDriverImpl());
```
GromNaN: Rebase note: this branch predates `0c8f92f0` (TypeRegistry), so this new helper does not inject a `TypeProvider` while `getConfiguration()` now does. Worth rebasing and adding `$config->setTypeProvider(new TypeRegistry())` so the helper does not fall back to the deprecated shared instance.

src/ChangeSets/ChangeSetComputer.php
alcaeus: Good catch, thanks for the thorough repro. Root cause confirmed: `setDocumentChangeSet()` built `originalData` only from the fields present in the array it received, so `setNewValue()`'s truncated changeset became a permanently-incomplete snapshot for any later merge.

Fixed at the source in dae43437: `setDocumentChangeSet()` now seeds `originalData` from the document's real `ManagedObjectState` snapshot instead. Added a regression test (`UnitOfWorkTest::testRecomputeAfterSetNewValuePreservesOldValueForFieldAddedByRecompute`) reproducing your exact scenario (increment-strategy field, `setDocumentChangeSet()` via `setNewValue()`-style call, then a second field changed and picked up by recompute) and asserting the old value stays correct.

src/UnitOfWork.php
alcaeus: Fixed together with the companion comment on `ChangeSetComputer::computeChangeSet()` — see dae43437. `originalData` is now seeded from the document's real snapshot (`DocumentRegistry`/`ManagedObjectState`) rather than from the fields present in the passed array.

Left a TODO on both `setDocumentChangeSet()` and `ChangeSet`'s class docblock noting this is still a copy-at-construction-time fix rather than a structural one — filed doctrine/mongodb-odm#3065 to track the broader "originalData copied vs. live" issue, which this is part of.

src/ChangeSets/ChangeSet.php
alcaeus: Left as-is: with the `setDocumentChangeSet()` fix (dae43437), `originalData` is now always the document's real, full snapshot, so a missing key here is a legitimate "no prior value" state (e.g. a field added to the schema after the snapshot was taken), not a symptom of a truncated snapshot anymore. Added a TODO on the class docblock flagging the copy-vs-live-reference question for the future (see doctrine/mongodb-odm#3065), but didn't change this line itself.

src/ChangeSets/ChangeSetComputer.php
alcaeus: Fixed in 32694992: both single-valued branches now check `$orphan !== $changeSet->getNewValue($propName)` before scheduling, so a field present in the changeset with old === new (the revert-within-the-same-flush case you describe) is skipped. Applied to both tracking modes for the invariant to be explicit rather than NOTIFY-specific.

Added `ChangeSetComputerTest::testNotifyReusedChangeSetDoesNotOrphanEmbedRevertedToOriginalValue` covering exactly the scenario you described (NOTIFY, single-valued embed changed then changed back within the same pass).

src/UnitOfWork.php
alcaeus: Documented in UPGRADE-2.18.md (601b3846): the signature change (`int $oid` → `object $document`) and the `unset()`-vs-`[]` behavior change (flipping `isset()`) are both called out now, alongside the other internal-method changes already listed there.

UPGRADE-2.18.md
alcaeus: Rewrote this section in 601b3846 to drop the "purely additive" framing and document both behavior changes you flagged: NOTIFY's old value always coming from the snapshot now, and `bubbleUpEmbeddedChildChanges()` reporting the real snapshot value instead of `[$value, $value]`.

src/UnitOfWork.php
alcaeus: Fixed in 33670a8e: `getDocumentChangeSet()` now reads `$this->documentChangeSets[$document]?->toArray() ?? []` directly instead of going through `getChangeSet()`'s clone.

src/UnitOfWork.php
alcaeus: Fixed in 33670a8e alongside the `getDocumentChangeSet()` clone: `executeUpdates()` now reads `$this->documentChangeSets[$document]` directly and checks `isEmpty()` on that, without cloning.

src/ChangeSets/ChangeSetComputer.php
alcaeus: Fixed in 37c06ece: `computeChangeSet()` now calls the private `computeSingleChangeSet()` directly instead of routing through `computeChangeSets()`, avoiding the per-call `SplObjectStorage` allocation.

src/UnitOfWork.php
alcaeus: Documented in UPGRADE-2.18.md (601b3846), alongside the NOTIFY old-value change.

src/ChangeSets/ChangeSetComputer.php
alcaeus: Kept the branch — dropping it risks wrong behavior for any future/direct caller of this stateless computer outside `UnitOfWork`. Tightened the comment in cf96949e to say explicitly that it's unreachable via `UnitOfWork::applyChangeSet()` (which returns early for that case) and only matters for a direct caller.

src/ChangeSets/ChangeSet.php
alcaeus: Fixed in 098836ee: the 3-arg constructor is gone; `setDocumentChangeSet()` (the only production caller that needed to carry changes at construction) now uses a new `@internal` named factory, `ChangeSet::fromChanges()`. The public constructor is 2-arg only now (`$document`, `$originalData`).

Left `applyToOriginalData()` as-is per your framing of it as a passing observation rather than something to remove.

src/UnitOfWork.php
alcaeus: Filed doctrine/mongodb-odm#3065, noting it can host #2834 too as you suggested. Trimmed both `FIXME`s down to a one-line pointer at the issue in cf96949e.

tests/Tests/BaseTestCase.php
alcaeus: Fixed in 14150e34: `createMetadataOnlyDocumentManager()` now calls `$config->setTypeProvider(new TypeRegistry())` like `getConfiguration()` does, so it no longer falls back to the deprecated shared instance.

src/UnitOfWork.php
alcaeus: Done in 601b3846: added a note in UPGRADE-2.18.md for NOTIFY tracking, and a regression test (`UnitOfWorkTest::testNotifyTrackingReportsOldValueFromSnapshotNotFromNotifier`, in dae43437) asserting the reported old value comes from the snapshot even when `propertyChanged()` is called with a different `$oldValue`.

src/UnitOfWork.php
alcaeus: Filed doctrine/mongodb-odm#3066 to track measuring this properly. In the meantime (33670a8e), the two internal call sites that didn't actually need a detached copy — `getDocumentChangeSet()` and the `isEmpty()` check in `executeUpdates()` — now read `$documentChangeSets` directly instead of going through `getChangeSet()`, which should cover the hot paths that mattered most. `getChangeSet()` itself still clones, since that's part of its public "detached instance" contract.

src/UnitOfWork.php
alcaeus: Done in d97a269c: `doDetach()` now also unsets the document's entry in `documentChangeSets`, alongside the other scheduling state it already cleared.

alcaeus:
Pushed 9 commits addressing @GromNaN's review and the remaining open threads from @IonBazan's earlier one:

- **Blocking**: fixed the `$inc`-corrupting old-value bug — `setDocumentChangeSet()` now seeds its `ChangeSet`'s `originalData` from the document's real snapshot instead of only the fields the caller happened to pass (dae43437), with a regression test reproducing the repro from the review.
- Fixed a related correctness bug: NOTIFY tracking could wrongly schedule orphan removal for a single-valued embed/reference reverted to its original value within the same flush (32694992).
- Documented the observable behavior changes in `UPGRADE-2.18.md` (NOTIFY old-value source, bubbled embedded-child old value, `clearDocumentChangeSet()`'s signature/behavior) (601b3846).
- Perf: dropped unnecessary `ChangeSet` clones on `getDocumentChangeSet()`/`executeUpdates()`, and `computeChangeSet()` no longer routes through the batch `SplObjectStorage` entry point (33670a8e, 37c06ece).
- Cleanup: `ChangeSet`'s changes-carrying constructor is `@internal`-only now (098836ee); `doDetach()` also clears a document's changeset (d97a269c); fixed a stale rebase gap in `BaseTestCase::createMetadataOnlyDocumentManager()` (14150e34).
- Filed doctrine/mongodb-odm#3065 (premature `originalData` snapshot, can also host #2834) and doctrine/mongodb-odm#3066 (measuring `getChangeSet()`'s clone cost) for the two follow-ups that were asked for as separate issues rather than inline fixes.

Replied inline on every thread. All local tests (`ChangeSets`, `UnitOfWorkTest`, `Events`, full suite), `phpcs`, and `phpstan` are green. Re-requested review from both of you.

src/UnitOfWork.php
```diff
@@ -511,25 +515,65 @@ private function computeScheduleUpsertsChangeSets(): void
...
+        // truncated snapshot here would make a later merge onto this
+        // instance (see ChangeSetComputer::computeChangeSet()) report a
+        // null old value for every field the merge adds.
+        //
+        // TODO: originalData is still a snapshot copied at construction
+        // time rather than a live view of the document's actual state, so
+        // this class of desync is only closed at this one call site, not
+        // structurally. Revisit alongside the premature-originalData-stamp
+        // issue tracked in applyChangeSet().
+        $originalData = $this->documentRegistry->getOrCreateObjectState($document)->originalData ?? [];
```
GromNaN: Blocking: this seeding replaces one wrong old value with another in the `setNewValue()` case it was meant to fix.

`objectState->originalData` is not the persisted state at this point. Earlier in the same flush, `computeChangeSets()` already stamped it with the pre-update in-memory data (`applyChangeSet()`, the `originalData = $actualData` assignment). So for any field the passed changeset carries, the value seeded here is the value the document had when the flush started, not the value stored in MongoDB. For those fields the passed changeset is the authoritative source, because the field diff built it against the persisted snapshot.

Scenario following the real flush order. The document is loaded with `count = 1`, the application sets `count = 5`, `computeChangeSets()` runs, then a preUpdate subscriber calls `setNewValue('count', 7)`:
- 2.18.x: `count old=1 new=7` and `$inc: 6`, so the document ends at 7. Correct.
- This head: `count old=5 new=7` and **`$inc: 2`**, so the document ends at 3.

The same happens for plain fields. In `tests/Tests/Events/PreUpdateEventArgsTest::testChangeSetIsUpdated`, `body` is null in the database, gets set to `'Body'`, then a subscriber overrides it through `setNewValue('body', 'Changed')`. On 2.18.x the changeset reports `body old=null new='Changed'`. On this head it reports `body old='Body' new='Changed'`, where `'Body'` was never the persisted value. That test only asserts the persisted value, so it passes either way.

The passed changeset should win over the snapshot for the fields it carries:

```php
$originalData = $this->documentRegistry->getOrCreateObjectState($document)->originalData ?? [];
$newValues    = [];
foreach ($changeset as $field => [$oldValue, $newValue]) {
    $originalData[$field] = $oldValue;
    $newValues[$field]    = $newValue;
}
```

I applied that on a scratch copy and ran three scenarios: the one from my previous review (a recompute adds a field after `setNewValue()`), this flush-order case, and the `testChangeSetIsUpdated` case. All three then match 2.18.x, and `tests/Tests/ChangeSets` still passes. Both halves are needed: keeping the snapshot as the base is what fixes the missing-field case, and letting the changeset overwrite it is what keeps the persisted old values.

src/UnitOfWork.php
```diff
@@ -511,25 +515,65 @@ private function computeScheduleUpsertsChangeSets(): void
...
+    /**
+     * Gets the changeset for a document as a {@see ChangeSet} value object.
+     *
+     * Always returns an instance detached from the one (if any) stored
+     * internally: mutating the returned instance (e.g. via
+     * {@see ChangeSet::recordChange()}) has no effect on the UnitOfWork's own
+     * bookkeeping. That guarantee costs a clone on every call; see #3066
+     * for whether that is worth optimizing further. Callers that don't need
+     * a detached ChangeSet (e.g. only checking whether one is empty) should
+     * read {@see self::$documentChangeSets} directly instead.
```
GromNaN: External callers cannot read `self::$documentChangeSets`, it is private, so this advice only applies to code inside `UnitOfWork`. For a public method's docblock, pointing inspect-only callers at `getDocumentChangeSet()` would be clearer.

src/ChangeSets/ChangeSetComputer.php
```diff
@@ -0,0 +1,397 @@
...
+            return false;
+        }
+
+        $dateType      = $class->getFieldType($propName);
+        $dbOrgValue    = $dateType->convertToDatabaseValue($orgValue);
+        $dbActualValue = $dateType->convertToDatabaseValue($actualValue);
+
+        // Loose comparison is only safe when both values are UTC dates. A custom
+        // type overriding "date" may produce a different database representation.
+        if (! $dbOrgValue instanceof UTCDateTime || ! $dbActualValue instanceof UTCDateTime) {
```
GromNaN: When a `date`-mapped field resolves to a custom type whose database value is not a `UTCDateTime`, this returns false, which means "not equivalent". The field is then recorded as changed and the document is written on every flush even when nothing actually changed. Falling back to a `===` comparison of the two database values would avoid that write amplification.

GromNaN (commented):
Re-review of the fixes. Ten of the fourteen comments from my previous review are resolved, three are resolved by documenting a deliberate decision, and the fix for the blocking one is incomplete: it removes the null old value in the case I reported, but replaces the old value with a different wrong value in the more common `setNewValue()` case. Details in the comment on `setDocumentChangeSet()`.

Resolved:
- `getDocumentChangeSet()` no longer clones the stored `ChangeSet`, and `executeUpdates()` reads the stored instance directly.
- `computeChangeSet()` no longer routes the single-document case through the batch entry point.
- `ChangeSet`'s constructor no longer exposes the changes, replaced by an `@internal` `fromChanges()` factory.
- The premature-snapshot `FIXME`s now point at #3065, which exists.
- `clearDocumentChangeSet()`'s signature and `unset()` semantics, plus the two value changes of `getDocumentChangeSet()` (NOTIFY old value source, embedded bubbles), are documented in `UPGRADE-2.18.md`. That section reads accurately against the code.
- The read-only branch in `ChangeSetComputer` is now documented as only reachable by a direct caller.
- `BaseTestCase` injects a `TypeRegistry`.
- Single-valued orphan scheduling now compares old against new, with a test for the NOTIFY revert case.

Still open, both minor:
- `ChangeSet::applyToOriginalData()` still has no caller outside `ChangeSetTest`.
- `getChangeSet()`'s docblock points callers at a private property.

Verified on this head: `tests/Tests/ChangeSets` passes (40 tests), `phpcs` is clean on the changed files, and the scenarios described in the comment on `setDocumentChangeSet()` were run against 2.18.x, the previous head and this head (scripts kept outside the repository). I could not run the functional suite: no MongoDB server is reachable from the driver in my environment, so `PreUpdateEventArgsTest::testChangeSetIsUpdated` never executed, and it only asserts the persisted value, so it would not detect the issue anyway.

PR #3056, opened 2026-09-21
Switch PersistenceBuilder to ChangeSet

alcaeus:
Third layer of the same refactor stack, based on `changeset-computer` (#3055 — merge that first). Switches `PersistenceBuilder::prepareInsertData()`, `prepareUpdateData()`, and `prepareUpsertData()` from the legacy tuple/array changeset shape to the `ChangeSets\ChangeSet` value object introduced in the previous layer, extracting per-method helpers along the way for readability, and brings the class's test coverage up to match.

- `prepareInsertData()`, `prepareUpdateData()`, `prepareUpsertData()` now consume `ChangeSet` instead of the raw array shape
- Extracted smaller private helpers out of each of those methods
- Fixed `resolveDiscriminatorValue()`'s return type
- Expanded `PersistenceBuilderTest` from 12 to 46 tests, raising `PersistenceBuilder` coverage from 37.6%/22.7% (line/method) to 97.6%/90.9% — covers null-field handling, the `increment` strategy, embed-one/embed-many (atomic and per-element), scheduled-collection updates, discriminator resolution, and nested-collection wrapping
- Added three test-only fixtures to reach otherwise-unreachable branches: `Documents\UnmappedSubProject` and `Documents\Functional\Ticket\GH683\UnmappedEmbedded` (subclasses deliberately left out of their parent's discriminator map, to exercise `MappingException::unlistedClassInDiscriminatorMap`), and `Documents\Functional\OpenDiscriminatorDocument` (a discriminator field with no map, to exercise the class-name fallback)
- Added a defensive `UnexpectedValueException` guard at the end of `prepareInsertData()`'s and `prepareUpdateData()`'s field-mapping branches, consistent with the existing convention in `prepareEmbeddedDocumentValue()`. Getting insert's guard to be safely exhaustive required extracting `applyManyFieldInsert()`, since the old inline condition there was only ever a partial match (owning-side, non-empty-or-`storeEmptyArray`, non-`addToSet`) — a straight `throw` after it would have broken any insert of a document with an inverse-side collection or a populated `addToSet` field. Caught by the expanded test suite
- Added the missing native return types across the class (`array`, `array|object`, or `mixed` where a branch can return a bare scalar reference id and can't be narrowed further)

Deliberately unchanged / out of scope:
- `applyScalarFieldUpdate`/`applyScalarFieldUpsert` and `applyEmbedOneFieldUpdate`/`applyEmbedOneFieldUpsert` are near-duplicate pairs that only differ in which sibling `prepare*` method they recurse into (and, for the scalar pair, the `empty($mapping['id'])` guard, which reflects a real behavioral difference — see that method's docblock). Left unmerged deliberately: unifying them means threading a closure through, which trades a small amount of duplication for an extra layer of indirection in an `@internal` class already covered by tests
- 5 lines remain uncovered: a `switch` `default` arm in `prepareEmbeddedDocumentValue()` that's unreachable via any valid mapping, a discriminator-map-without-`targetDocument` combination on an embedded field that no existing fixture models, and `preparePersistentCollection()`'s first branch, which its only caller never actually exercises (it never passes an already-persistent collection). Not chased further — real fixtures for these are either impossible or not worth manufacturing


src/Persisters/PersistenceBuilder.php
```diff
@@ -123,95 +196,214 @@ public function prepareInsertData($document)
...
+    /**
+     * An INCREMENT-strategy field is written as the delta between old and
+     * new value via `$inc`, so concurrent writers accumulate rather than
+     * clobber each other; every other scalar field is simply overwritten.
+     *
+     * @param array<string, mixed> $updateData
+     * @phpstan-param ClassMetadata<object> $class
+     * @phpstan-param FieldMapping $mapping
+     */
+    private function applyScalarFieldUpdate(array &$updateData, ClassMetadata $class, array $mapping, mixed $old, mixed $new): void
```
GromNaN: Cross-reference to #3055: `$old` here, and in `applyScalarFieldUpsert()`, now comes from `ChangeSet::getOldValue()`, which reads the changeset's `originalData`. That makes the `$inc` operator depend on the old-value issue raised on #3055 (`UnitOfWork::setDocumentChangeSet()` seeding a snapshot that is not the persisted state). In the scenario described there, the changeset reports `5` where the database holds `1`, so this method emits `$inc: 2` where `$inc: 6` is correct, and the document lands on the wrong value.

The new `testPrepareUpdateDataIncrementStrategyComputesDelta()` does not catch it, because it never goes through `PreUpdateEventArgs::setNewValue()`. Once the fix lands on #3055, a test here covering the increment delta after a preUpdate listener pushed a change through the event args would protect this method, since this is where a wrong old value becomes a wrong write.

GromNaN (commented):
Reviewed the incremental diff against the `changeset-computer` branch (layer 3 of the stack), with attention to correctness, backward compatibility, performance and security.

Verdict: the conversion is faithful. I compared each extracted helper against the code it replaces, focusing on the parts most likely to drift:

- `applyManyFieldInsert()`: the three negated guards (inverse side, empty without `storeEmptyArray`, `addToSet` without `storeEmptyArray`) reproduce the old compound `elseif` condition exactly, including the fact that all three cases silently skip the field rather than erroring.
- `applyManyFieldUpdate()`: the four-branch order is unchanged, and `$old` still comes from the same source as the removed tuple's `[0]`.
- `applyNullFieldUpdate()`, `applyNullFieldUpsert()`, the `empty($mapping['id'])` guard in `applyScalarFieldUpsert()` and the discriminator handling (now in `resolveDiscriminatorValue()`, still only used by insert and upsert) are all preserved.
- Return types: `prepareEmbeddedDocumentValue()` never returned null despite the old docblock, so `array|object` is correct and the new empty-result test pins it; `prepareAssociatedCollectionValue()` always returns an array.
- The `object` type hints added on the embed-one and reference-one helpers do not narrow anything in practice: `DocumentManager::createReference()` already requires an object, so an unsupported scalar fails at the same point as before.

Two notes, neither blocking:

1. `prepareInsertData()`, `prepareUpdateData()` and `prepareUpsertData()` now call `getChangeSet()`, which builds a detached clone on every call, once per document and recursively per embedded document. That is not a regression versus the state this branch starts from, since `getDocumentChangeSet()` built a whole array instead, but the write path only reads here, and #3066 tracks whether the clone is worth removing.
2. This branch is where a wrong old value turns into a wrong write. See the inline comment for the dependency on #3055.

Test coverage: I could not execute `PersistenceBuilderTest` (no MongoDB server reachable from the driver, and it extends `BaseTestCase`). `phpcs` is clean on both changed files. PHPStan reports nothing on the files this PR touches (three covariance errors appear in `Tools/Console/Command/CommandCompatibility.php` only under my scratch harness's autoloading, on a file this PR does not modify and which is clean on the base branch).

